PR #1, opened 2018-08-07
proto: add daemon proto

jazzdan:


maiamcc:
oh, you should go ahead and generate the proto files (/add a fabfile?)

PR #3, opened 2018-08-07
cli: skeleton for cli

maiamcc:

Please review the following commits I made in branch maiamcc/cli-skeleton:

ba7c666936d46833fc991ed4af4eaf89e9e74b9d (2018-08-07 12:37:39 -0400)
cli: skeleton for cli


cli/cli.go
```diff
@@ -0,0 +1,36 @@
+package cli
```
jazzdan: Like we talked about at lunch, if we followed the official go conventions this would live in `internal/cli` because this isn't designed to be used outside of this project.

maiamcc: @jazzdan @nicks thoughts on this new directory structure? am i missing something?

PR #4, opened 2018-08-07
k8s: implement Apply

jazzdan:


internal/k8s/apply.go
```diff
@@ -0,0 +1,29 @@
...
+	"os"
+	"os/exec"
+	"path/filepath"
+)
+
+func Apply(ctx context.Context, rawYAML string) error {
+	// TODO(dmiller) validate that the string is YAML and give a good error
+	dir, err := ioutil.TempDir("", "tilt")
+	if err != nil {
+		log.Fatal(err)
```
nicks: return the error?

internal/k8s/apply.go
```diff
@@ -0,0 +1,29 @@
...
+	}
+	defer os.RemoveAll(dir)
+
+	tmpf := filepath.Join(dir, "tilt.yaml")
+	fmt.Println(tmpf)
```
nicks: :eyes:

jazzdan: whoops will remove

internal/k8s/apply.go
```diff
@@ -0,0 +1,29 @@
...
+	contents := []byte(rawYAML)
+	if err := ioutil.WriteFile(tmpf, contents, 0666); err != nil {
+		log.Fatal(err)
```
nicks: return err?

jazzdan:
btw, I tested this with the following script:

```go
package main

import (
	"context"
	"fmt"
	"os"

)

func main() {
	yaml := `apiVersion: v1
kind: Service
metadata:
  name: hello-kubernetes
spec:
  type: LoadBalancer
  ports:
  - port: 80
    targetPort: 8080
  selector:
    app: hello-kubernetes
---
apiVersion: extensions/v1beta1
kind: Deployment
metadata:
  name: hello-kubernetes
spec:
  replicas: 3
  template:
    metadata:
      labels:
        app: hello-kubernetes
    spec:
      containers:
      - name: hello-kubernetes
        image: paulbouwer/hello-kubernetes:1.4
        ports:
        - containerPort: 8080`

	err := k8s.Apply(context.Background(), yaml)
	if err != nil {
		fmt.Println(err.Error())
		os.Exit(1)
	} else {
		fmt.Println("It worked!")
	}
}
```

PR #8, opened 2018-08-07
proto: commit generated go files

jazzdan:
Also add go to Makefile


Makefile
```diff
@@ -1,6 +1,9 @@
 .PHONY: all proto
 
-all: proto
+all: proto go
 
 proto:
-	protoc --go_out=plugins=grpc:../../../ -I. proto/*.proto
+	protoc --go_out=plugins=grpc:../../../ -I. internal/proto/*.proto
+
+go:
```
nicks: i would usually call this `install`?

Makefile
```diff
@@ -1,6 +1,9 @@
 .PHONY: all proto
 
-all: proto
+all: proto go
```
nicks: by convention, `all` usually only compiles things and doesn't create binaries, see:
https://www.gnu.org/prep/standards/html_node/Standard-Targets.html#Standard-Targets

nicks: (tho i'm not sure i care that much)

PR #11, opened 2018-08-08
tilt: basic tiltfile loading

landism:


internal/tiltfile/tiltfile.go
```diff
@@ -0,0 +1,31 @@
...
+	"log"
+)
+
+type Tiltfile struct {
+	globals  skylark.StringDict
+	filename string
+	thread   *skylark.Thread
+}
+
+func Load(filename string) Tiltfile {
```
nicks: in Go, it's idiomatic to return a (Tiltfile, error) and propagate the error to the caller, so that the caller can decide what to do with it

(then the error-printing code would go in main.go or in the base cli handlers)

PR #13, opened 2018-08-08
proto: proto supports building docker images

jazzdan:


internal/proto/daemon.proto
```diff
@@ -13,18 +13,32 @@ service Daemon {
     // rpc WatchService(WatchServiceRequest) returns (WatchServiceReply) {}
 }
 
-// Q: Is this sufficient?
-// Q: How will we uniquely identify a service?
-// ie, what's a service's "name"?
 message Service {
     string k8s_yaml = 1;
-    //repeated Cmd steps = 2;
+    string dockerfile_text = 2;
```
maiamcc: i'd say clarify in variable name OR in comment that this is the base dockerfile?

PR #14, opened 2018-08-08
tiltd: CreateService applies yaml

maiamcc:
Hello @jazzdan, @landism,

Please review the following commits I made in branch maiamcc/apply-yaml:

4a03afb1fa3026cc3e5a11faba9c8d74d006027e (2018-08-08 14:53:57 -0400)
tiltd: CreateService applies yaml


internal/cli/up.go
```diff
@@ -30,14 +29,15 @@ func (c *upCmd) run(args []string) error {
...
 
-	dCli, err := tiltd_client.NewDaemonClient()
-	if err != nil {
-		return err
-	}
-	err = dCli.CreateService(ctx, "blahblahblah")
-	if err != nil {
-		return err
-	}
+	//dCli, err := tiltd_client.NewDaemonClient()
```
jazzdan: 👀

internal/k8s/apply.go
```diff
@@ -12,5 +13,9 @@ func Apply(ctx context.Context, rawYAML string) error {
 	r := bytes.NewReader([]byte(rawYAML))
 	c.Stdin = r
 
+	// TODO: handle output more cleanly
+	c.Stdout = os.Stdout
```
jazzdan: Do we want this printing to the screen?

.python-version
```diff
@@ -0,0 +1 @@
+2.7.14
```
jazzdan: Why do we need python?

internal/k8s/apply.go
nicks: ya, i'm also not totally clear on what you're trying to accomplish here. You can pull Stderr out of an error if that's what you're looking for

internal/cli/up.go
maiamcc: ...i know this is awkward but tbh i'm gonna leave it cuz it makes the code easier to test

jazzdan: You could make a new `cmd/` binary to test and not commit if that's what you're looking for.

.python-version
maiamcc: i have NO IDEA, but something in the `kubectl` code we're executing relies on it:
```Unable to connect to the server: error executing access token command "/Users/maia/google-cloud-sdk/bin/gcloud config config-helper --format=json": err=exit status 127 output= stderr=pyenv: python2: command not found

The `python2' command exists in these Python versions:
  2.7.14```

however, i don't think pyenv needs to care about it unless your pyenv happens to be borked from the last project (like mine). fixed and removed this file.

PR #16, opened 2018-08-08
tilt: pass service defs to daemon

landism:


internal/tiltfile/tiltfile.go
```diff
@@ -25,3 +26,34 @@ func Load(filename string) (*Tiltfile, error) {
 
 	return &Tiltfile{globals, filename, thread}, nil
 }
+
+func (tiltfile Tiltfile) GetServiceConfig(serviceName string) (*string, error) {
```
jazzdan: String pointer! Exciting :)

internal/tiltfile/tiltfile_test.go
```diff
@@ -34,8 +35,72 @@ def blorgly_frontend():
...
 	defer os.Remove(file)
 	tiltConfig, err := Load(file)
-	for _, s := range([]string{"blorgly", "blorgly_backend", "blorgly_frontend"}) {
+	for _, s := range []string{"blorgly", "blorgly_backend", "blorgly_frontend"} {
 		assert.Contains(t, tiltConfig.globals, s)
 	}
 	assert.Nil(t, err)
 }
+
+func TestGetServiceConfig(t *testing.T) {
```
jazzdan: yasss tests!

internal/tiltfile/tiltfile.go
```diff
@@ -25,3 +26,34 @@ func Load(filename string) (*Tiltfile, error) {
...
+	f, ok := tiltfile.globals[serviceName]
+
+	if !ok {
+		return nil, errors.New(fmt.Sprintf("%v does not define a global named '%v'", tiltfile.filename, serviceName))
```
nicks: fwiw, there's a shortcut function `fmt.Errorf`

internal/tiltfile/tiltfile.go
nicks: ya, this seems non-idiomatic. most functions like this return (string, error) and return the empty string on error

PR #20, opened 2018-08-09
build: build base docker image

jazzdan:


jazzdan:
I'm going to have to do something to get these tests to work in CI.

jazzdan:
(but for now I'm just not going to run them)

internal/build/docker_test.go
```diff
@@ -0,0 +1,154 @@
...
+	repo *temp.TempDir
+	dcli *client.Client
+}
+
+func newTestFixture(t *testing.T) *testFixture {
+	repo, err := temp.NewDir(t.Name())
+	if err != nil {
+		t.Fatalf("Error making temp dir: %v", err)
+	}
+	dcli, err := client.NewEnvClient()
```
nicks: hmmm...i'm skeptical that our tests should be using a real docker client

internal/build/docker_test.go
```diff
@@ -0,0 +1,154 @@
...
+	}
+}
+
+func (f *testFixture) newBuilderForTesting() *localDockerBuilder {
+	return &localDockerBuilder{
+		dcli: f.dcli,
+	}
+}
+
+func (f *testFixture) assertFileInImage(tag string, path string) {
```
nicks: this function is a good reason why it might make more sense to use a fake docker client

internal/build/docker.go
```diff
@@ -0,0 +1,104 @@
...
+	"io"
+	"regexp"
+	"strings"
+
+	"github.com/docker/docker/api/types"
+	"github.com/docker/docker/client"
+)
+
+type Mount struct {
+	Repo          LocalGithubRepo
```
maiamcc: maybe a TODO here to type this better/more generic?

internal/build/docker.go
```diff
@@ -0,0 +1,104 @@
...
+	ContainerPath string
```
maiamcc: nit: i always read this as "path TO container" rather than "path IN container", and so I used `RemotePath` in my code to disambiguate. If you want 🤷‍♀️

internal/build/docker.go
```diff
@@ -0,0 +1,104 @@
...
+}
+
+func tarFromDockerfile(df string) (*bytes.Reader, error) {
+	buf := new(bytes.Buffer)
+	tw := tar.NewWriter(buf)
+	defer tw.Close()
+
+	tarHeader := &tar.Header{
+		Name: "Dockerfile",
+		Size: int64(len(df)),
```
maiamcc: usually won't make a difference but technically we want `len([]byte(df))`, no?

internal/build/docker.go
```diff
@@ -0,0 +1,104 @@
...
+	dockerFileTarReader := bytes.NewReader(buf.Bytes())
+
+	return dockerFileTarReader, nil
+}
+
+func getTagFromOutput(output string) string {
+	re := regexp.MustCompile("Successfully built ([[:alnum:]]*)")
+	res := re.FindStringSubmatch(output)
+	if len(res) != 2 {
+		panic("fml")
```
maiamcc: ...maybe we WILL want proper error handling here some day? idk.

internal/build/hello
```diff
@@ -0,0 +1 @@
+hi hello
```
maiamcc: 👀

internal/build/docker_test.go
maiamcc: let's def scheme about ways we could usefully mock this out! that would be EXCELLENT -- but also imo the thing here that most needs testing is that our code plays nice with the docker client and does what we expect it to?

internal/build/docker.go
```diff
@@ -0,0 +1,104 @@
...
+		})
+
+	defer imageBuildResponse.Body.Close()
+	output := &strings.Builder{}
+	_, err = io.Copy(output, imageBuildResponse.Body)
+	if err != nil {
+		return "", err
+	}
+
+	return getTagFromOutput(output.String()), nil
```
maiamcc: thinking about this more, i think a lot of what we're calling `tag` is actually `imageHash` or `digest` or something...?

internal/build/docker_test.go
jazzdan: If we used a fake docker client, wouldn't we be testing nothing? There's not a lot of logic in this code between us and the docker daemon.

internal/build/docker.go
maiamcc: mm yeah probs `digest`? https://windsock.io/explaining-docker-image-ids/

internal/build/docker_test.go
nicks: fair!

internal/build/docker_test.go
```diff
@@ -0,0 +1,154 @@
...
+	"io/ioutil"
+	"os"
+	"strings"
+	"testing"
+
+	"github.com/docker/docker/api/types"
+	"github.com/docker/docker/api/types/container"
+	"github.com/docker/docker/client"
+
+	"github.com/windmilleng/wat/os/temp"
```
nicks: ???

this seems wrong

internal/build/docker.go
jazzdan: Yeah, except in this case it's not remote at all. And since we're getting passed the Dockerfile we don't need to worry about the path to the dockerfile! :D

internal/build/docker.go
jazzdan: ooh, good catch. And it turns out that `len(myString)` is returning the number of bytes and not the number of characters or runes. So this is fine!

https://stackoverflow.com/questions/12668681/how-to-get-the-number-of-characters-in-a-string

internal/build/docker_test.go
jazzdan: Yeah this was a bit of a hack to get all that functionality. Do you think we should copy it in or move it to wmclient @nicks?

internal/build/docker.go
maiamcc: OMG WHAAAAT?!?

https://media0.giphy.com/media/26ufdipQqU2lhNA4g/giphy.gif

PR #21, opened 2018-08-09
regenerate code from .proto; add golang/protobuf to README

landism:


nicks:
we might need to vendor the protoc-gen-go tool itself...it looks like there are some problems?

landism:
turns out this was unneeded anyway

PR #23, opened 2018-08-09
landism/yaml generation

landism:
can now generate yaml from local commands, e.g.:
```
def blorg_backend():
  image = build_docker_image('Dockerfile', 'gcr.io/blorg-dev/blorg-backend:dev-landism')
  yaml = local('./populate_config_template.py argiuherg && cat k8s-conf.generated.yaml')
  return k8s_service(yaml, image)
```

### DIFF
--- a/internal/tiltfile/tiltfile.go
+++ b/internal/tiltfile/tiltfile.go
@@ -5,6 +5,8 @@ import (
 	"github.com/google/skylark"
 	"github.com/windmilleng/tilt/internal/proto"
 	"io/ioutil"
+	"os/exec"
+	"github.com/pkg/errors"
 )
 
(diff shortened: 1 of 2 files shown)


internal/tiltfile/tiltfile.go
```diff
@@ -3,8 +3,10 @@ package tiltfile
 import (
 	"fmt"
 	"github.com/google/skylark"
+	"github.com/pkg/errors"
```
nicks: did you intend to use plain "errors" from the standard lib?

landism: is it idiomatic to use fmt.Errorf on bare (non-format) strings?

PR #24, opened 2018-08-09
tilt: add tiltfile support for mount points

landism:
had been putting this off because I didn't understand this part of the spec

usage, e.g.:
```
  image = build_docker_image("%v", "docker tag")
  image.add_mount('/mount_points/1', git_repo('/mount_points/1/foo'))
```

### DIFF
--- a/internal/tiltfile/skylarkTypes.go
+++ b/internal/tiltfile/skylarkTypes.go
@@ -39,9 +39,15 @@ func (k8sService) Hash() (uint32, error) {
 	return 0, errors.New("unhashable type: k8sService")
 }
 
+type mount struct {
+	mount_point string
+	repo gitRepo
+}
+
 type dockerImage struct {
 	fileName skylark.String
 	fileTag  skylark.String
+	mounts	 []mount
 	cmds     []string
 }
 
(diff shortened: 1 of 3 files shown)


internal/tiltfile/tiltfile.go
```diff
@@ -122,10 +133,16 @@ func (tiltfile Tiltfile) GetServiceConfig(serviceName string) (*proto.Service, e
 		return nil, fmt.Errorf("internal error: k8sService.dockerFileTag was not a string in '%v'", service)
 	}
 
+	mounts := make([]*proto.Mount, 0, len(service.dockerImage.mounts))
+	for _, mount := range service.dockerImage.mounts {
+		repo := proto.Repo{RepoType: &proto.Repo_GitRepo{GitRepo: &proto.GitRepo{LocalPath: mount.repo.path}}}
```
landism: this is kinda gross, but I couldn't find a better way

internal/tiltfile/skylarkTypes.go
```diff
@@ -39,9 +39,15 @@ func (k8sService) Hash() (uint32, error) {
 	return 0, errors.New("unhashable type: k8sService")
 }
 
+type mount struct {
+	mount_point string
```
nicks: mount_point -> mountPoint

internal/tiltfile/tiltfile.go
nicks: ya, the go APIs for protobuf oneofs are awkward. (i am not a big fan of the Go API for generated protobuf in general)

PR #27, opened 2018-08-10
docker: use Digest instead of string

nicks:
Hello @maiamcc, @jazzdan,

Please review the following commits I made in branch nicks/ref:

64bdcea9cbbffd98c9117bcfad7a079886c14350 (2018-08-10 14:02:35 -0400)
docker: use Digest instead of string

### DIFF
--- a/internal/build/docker_test.go
+++ b/internal/build/docker_test.go
@@ -11,11 +11,12 @@ import (
 	"github.com/docker/docker/api/types"
 	"github.com/docker/docker/api/types/container"
 	"github.com/docker/docker/client"
+	digest "github.com/opencontainers/go-digest"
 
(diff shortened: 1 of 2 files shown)


internal/build/docker_test.go
```diff
@@ -118,11 +119,11 @@ func (f *testFixture) newBuilderForTesting() *localDockerBuilder {
 	}
 }
 
-func (f *testFixture) assertFileInImage(tag string, path string) {
+func (f *testFixture) assertFileInImage(tag digest.Digest, path string) {
```
maiamcc: should also rename variable "tag"?

jazzdan:
this was integrated in my pr

PR #29, opened 2018-08-10
tilt: pull daemon into main tilt binary

landism:
now instead of running `tiltd`, one runs `tilt daemon` (and `go install cmd/tilt` instead of `go install cmd/tilt && go install cmd/tiltd`)

this is mostly to reduce the chance that we're running inconsistent versions of `tilt` and `tiltd`


internal/tiltd/tiltd_server/tiltd.go
```diff
@@ -23,7 +23,7 @@ func (d *Daemon) CreateService(ctx context.Context, k8sYaml string) error {
 
 func RunDaemon(ctx context.Context) (*os.Process, error) {
 	// Relies on having the latest tiltd go install'd.
```
maiamcc: this comment is not obsolete yay!

PR #33, opened 2018-08-10
build: asserting files and contents in tests only executes container …

jazzdan:
…once

### DIFF
--- a/internal/build/docker_test.go
+++ b/internal/build/docker_test.go
@@ -2,6 +2,7 @@ package build
 
 import (
 	"context"
+	"fmt"
 	"io"
 	"io/ioutil"
 	"os"
@@ -35,23 +36,6 @@ func TestDigestFromSingleStepOutput(t *testing.T) {
 	}
 }
 
-func TestBuildBase(t *testing.T) {
-	f := newTestFixture(t)
-	defer f.teardown()
-	baseDockerFile := `FROM alpine
-	RUN touch hi
-	`
-
-	builder := f.newBuilderForTesting()
-
-	tag, err := builder.buildBaseWithMounts(context.Background(), baseDockerFile, "hi", []Mount{})
-	if err != nil {
-		t.Fatal(err)
-	}
-
-	f.assertFileInImageWithContents(tag, "hi", "")
-}
-
 func TestMount(t *testing.T) {
 	f := newTestFixture(t)
 	defer f.teardown()
@@ -74,9 +58,12 @@ func TestMount(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	// TODO(dmiller): this is slow
-	f.assertFileInImageWithContents(tag, "/src/hi/hello", "hi hello")
-	f.assertFileInImageWithContents(tag, "/src/sup", "my name is dan")
+	pcs := []pathContent{
+		pathContent{path: "/src/hi/hello", contents: "hi hello"},
+		pathContent{path: "/src/sup", contents: "my name is dan"},
+	}
+
+	f.assertFilesInImageWithContents(tag, pcs)
 }
 
 type testFixture struct {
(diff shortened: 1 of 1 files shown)


internal/build/docker_test.go
```diff
@@ -35,26 +36,6 @@ func TestDigestFromSingleStepOutput(t *testing.T) {
 	}
 }
 
-func TestBuildBase(t *testing.T) {
```
jazzdan: We don't really need this test any more, it's entirely covered by the other test.

internal/build/docker_test.go
```diff
@@ -132,11 +119,22 @@ func (f *testFixture) newBuilderForTesting() *localDockerBuilder {
...
+type pathContent struct {
+	path     string
+	contents string
+}
+
+func (f *testFixture) assertFilesInImageWithContents(tag digest.Digest, contents []pathContent) {
 	ctx := context.Background()
+	var cmd strings.Builder
+	for _, c := range contents {
+		cs := fmt.Sprintf("cat %s | grep \"%s\" || echo \"contents not found\"; ", c.path, c.contents)
```
maiamcc: hmm do we want to differentiate the error message between file not found and contents not found? (at very least, err should read `file not found or has wrong contents`). or can you maybe printf the file name or the bad contents into the `echo`? this doesn't give the user much debug info...

internal/build/docker_test.go
```diff
@@ -167,10 +165,10 @@ func (f *testFixture) assertFileInImageWithContents(tag digest.Digest, path stri
 	io.Copy(output, out)
 
 	if strings.Contains(output.String(), "No such file") {
-		f.t.Errorf("Expected to find file %s in container. Got: %s", path, output)
+		f.t.Errorf("Expected to find file in container. Got: %s", output)
```
maiamcc: think this conditional is useless now: this is the bad output from `cat`, and `grep` will eat the output of `cat`, right?

internal/build/docker_test.go
```diff
@@ -167,10 +165,10 @@ func (f *testFixture) assertFileInImageWithContents(tag digest.Digest, path stri
...
 	}
 
-	if !strings.Contains(output.String(), contents) {
-		f.t.Errorf("Expected file %s to have contents %s but got %s", path, contents, output)
+	if strings.Contains(output.String(), "contents not found") {
+		f.t.Errorf("Expected to find contents in container. Got: %s", output)
```
maiamcc: maybe: `Failed to find one or more expected files in container with output: %s`

internal/build/docker_test.go
```diff
@@ -74,9 +58,12 @@ func TestMount(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	// TODO(dmiller): this is slow
-	f.assertFileInImageWithContents(tag, "/src/hi/hello", "hi hello")
-	f.assertFileInImageWithContents(tag, "/src/sup", "my name is dan")
+	pcs := []pathContent{
```
maiamcc: hmm it feeeeels like we should a single struct that holds test data that we can both write files from and assert test behavior from...? any thoughts? seems awk to rewrite the same string multiple times :-/

internal/build/docker_test.go
jazzdan: I couldn't think of a way to do this, but in my experience getting the stdout printed to you is enough debug info.

PRs welcome! :)

internal/build/docker_test.go
jazzdan: yeah this would be better if we were doing table tests...

internal/build/docker_test.go
maiamcc: ` echo \"could not find file %s / did not have contents %s\"` and sprintf it in

PR #36, opened 2018-08-10
build: build one step locally

jazzdan:


nicks (approved):

internal/build/docker.go
```diff
@@ -95,6 +99,42 @@ func (l *localDockerBuilder) buildBaseWithMounts(ctx context.Context, baseDocker
 	return getDigestFromOutput(output.String())
 }
 
+func (l *localDockerBuilder) buildImageWithSteps(ctx context.Context, baseDigest digest.Digest, tag string, steps []Cmd) (digest.Digest, error) {
```
maiamcc: naming nit: consider `execStepsOnImage`, maybe? (i don't feel strongly tho 🤷‍♀️ )

internal/build/docker.go
```diff
@@ -95,6 +99,42 @@ func (l *localDockerBuilder) buildBaseWithMounts(ctx context.Context, baseDocker
...
+	imageWithSteps := baseDigest
+	for _, s := range steps {
+		resp, err := l.dcli.ContainerCreate(ctx, &container.Config{
```
maiamcc: guhhh it seems like there should be a way to create a single container and run all the steps on it in sequence rather than creating a container per step? maybe if you `&&`'d them all together into a single bash command?

internal/build/docker.go
```diff
@@ -95,6 +99,42 @@ func (l *localDockerBuilder) buildBaseWithMounts(ctx context.Context, baseDocker
...
+			Image: string(baseDigest),
```
maiamcc: but also if you do it this way, this should be `imageWithSteps` lol

internal/build/docker.go
```diff
@@ -95,6 +99,42 @@ func (l *localDockerBuilder) buildBaseWithMounts(ctx context.Context, baseDocker
...
+		select {
+		case err := <-errCh:
+			if err != nil {
+				return "", err
+			}
+		case <-statusCh:
+		}
+
+		id, err := l.dcli.ContainerCommit(ctx, containerID, types.ContainerCommitOptions{})
+		imageWithSteps = digest.Digest(id.ID)
```
maiamcc: if something goes wrong with the commit, `id` might be nil, no? assign this after you check the err

internal/build/docker_test.go
```diff
@@ -193,6 +213,46 @@ type pathContent struct {
 	contents string
 }
 
+func (f *testFixture) assertFileInImage(tag digest.Digest, path string) {
```
maiamcc: no need to make a separate func for this, just assert the file exists with contents `""`

internal/build/docker_test.go
```diff
@@ -132,6 +132,26 @@ func TestMountCollisions(t *testing.T) {
 	f.assertFilesInImageWithContents(tag, pcs)
 }
 
+func TestBuildOneStep(t *testing.T) {
+	f := newTestFixture(t)
+	defer f.teardown()
+	baseDockerFile := "FROM alpine"
```
maiamcc: we use this often enough, let's pull it out to the top of the file as a const?

internal/build/docker.go
```diff
@@ -95,6 +99,42 @@ func (l *localDockerBuilder) buildBaseWithMounts(ctx context.Context, baseDocker
...
+		}
+
+		statusCh, errCh := l.dcli.ContainerWait(ctx, containerID, container.WaitConditionNotRunning)
+		select {
+		case err := <-errCh:
+			if err != nil {
+				return "", err
+			}
+		case <-statusCh:
+		}
```
maiamcc: doesn't need to happen now, but i bet we can extract `createContainerWithCommand` into its own func?

internal/build/docker_test.go
maiamcc: ...then my lazy ass might prefer that you run the command `echo hello world > foo` so you can still use the existing func? 🤷‍♀️

internal/build/docker.go
jazzdan: yes good catch! 😅

PR #40, opened 2018-08-13
wire up builder

jazzdan:


internal/proto/server_adaptor.go
```diff
@@ -16,5 +17,45 @@ func NewGRPCServer(del tiltd.TiltD) *GRPCServer {
...
+		r = append(r, mountP2D(m))
+	}
+
+	return r
+}
+
+func mountP2D(mount *Mount) build.Mount {
+	return build.Mount{
+		// TODO(dmiller): convert to repo, we need a path
+		Repo:          build.LocalGithubRepo{LocalPath: ""},
```
jazzdan: We need to modify the proto  so that we can get passed the path of the repo with the various kinds of repos. I'm regretting making this generic because dealing with generics in go is so painful.

internal/tiltd/tiltd_server/tiltd.go
```diff
@@ -5,19 +5,55 @@ import (
...
+}
 
 var _ tiltd.TiltD = &Daemon{}
 
 func NewDaemon() (*Daemon, error) {
-	return &Daemon{}, nil
+	opts := make([]func(*client.Client) error, 0)
+	opts = append(opts, client.FromEnv)
+
+	// Use client for docker 17
```
jazzdan: We should extract this out in to a function.

internal/tiltd/tiltd_server/tiltd.go
```diff
@@ -5,19 +5,55 @@ import (
...
+	if err != nil {
+		return err
+	}
+
+	_, err = k8s.ParseYAMLFromString(k8sYaml)
+	if err != nil {
+		return err
+	}
+
+	// TODO(dmiller): not sure how to use this API
```
jazzdan: @nicks how should I be using this?

internal/proto/server_adaptor.go
```diff
@@ -16,5 +17,45 @@ func NewGRPCServer(del tiltd.TiltD) *GRPCServer {
...
+	r := []build.Cmd{}
+
+	for _, c := range cmds {
+		r = append(r, cmdP2D(c))
+	}
+
+	return r
+}
+
+func cmdP2D(cmd *Cmd) build.Cmd {
```
maiamcc: aren't we calling these `steps` now?

internal/proto/server_adaptor.go
maiamcc: sigh

internal/proto/server_adaptor.go
```diff
@@ -1,6 +1,7 @@
 package proto
 
 import (
+	"github.com/windmilleng/tilt/internal/build"
```
nicks: i'm really concerned about the dependency of the protobuf package on the build package. can we restructure the packages to avoid this dependency?

internal/tiltd/tiltd_server/tiltd.go
nicks: you need to iterate thru the entities and check that at least one has the image you want

internal/proto/server_adaptor.go
jazzdan: Steps are made up of a list of Cmds

internal/proto/server_adaptor.go
jazzdan: yes

internal/tiltd/tiltd_server/tiltd.go
```diff
@@ -44,17 +45,35 @@ func (d *Daemon) CreateService(ctx context.Context, k8sYaml string, dockerfile s
...
+			return err
+		}
+		if replaced {
+			didReplace = true
+		}
+		newK8sEntities = append(newK8sEntities, newK8s)
+	}
+
+	if !didReplace {
+		return fmt.Errorf("Expected to find a matching k8s config to replace, but didn't")
```
nicks: the error message here could be better. how about "Docker image missing from yaml: %s", dockerfileTag?

PR #41, opened 2018-08-13
tiltfile: use sh instead of bash to make Cmd

jazzdan:


jazzdan:
The reason being that a lot of barebones images don't have `bash`, they only have `sh.

PR #43, opened 2018-08-13
tilt: verify gofmt in ci

landism:


Makefile
```diff
@@ -16,3 +16,6 @@ test:
 
 ensure:
 	dep ensure
+
+verifyfmt:
+	bash -c 'diff <(go fmt ./...) <(echo -n)'
```
jazzdan: Hmm, does this not work on macOS?

```
> sh -c 'diff <(go fmt ./...) <(echo -n)'
sh: -c: line 0: syntax error near unexpected token `('
sh: -c: line 0: `diff <(go fmt ./...) <(echo -n)'
```

landism: I'd expect `<(foo)` is a bash feature, not an sh feature

(I'm on macOS)

PR #44, opened 2018-08-13
build: add entrypoint to docker image

maiamcc:
Hello @jazzdan, @nicks,

Please review the following commits I made in branch maiamcc/entrypoint:

595e65ab0d7927684a35e45a532527bc1ad2da20 (2018-08-13 17:05:08 -0400)
build: add entrypoint to docker image

### DIFF
--- a/internal/build/docker.go
+++ b/internal/build/docker.go
@@ -43,6 +43,12 @@ type Cmd struct {
 	Argv []string
 }
 
+// entrypointStr returns the Cmd in string form for use as an entrypoint,
+// e.g. 'ENTRYPOINT ["executable", "param1", "param2"]'
+func (c Cmd) entrypointStr() string {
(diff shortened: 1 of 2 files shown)


internal/build/docker.go
```diff
@@ -43,6 +43,12 @@ type Cmd struct {
 	Argv []string
 }
 
+// entrypointStr returns the Cmd in string form for use as an entrypoint,
+// e.g. 'ENTRYPOINT ["executable", "param1", "param2"]'
+func (c Cmd) entrypointStr() string {
```
jazzdan: should this be a pointer receiver?

maiamcc: why should it be a pointer receiver? we mostly don't pass cmd's around as pointers, and I'm loath to start b/c it'd just add extra `nil` checks. (i am passing the cmd as a pointer to `startContainer` but that's just b/c it's optional.)

PR #47, opened 2018-08-13
proto: generate protobufs inside a container

nicks:
Hello @landism, @pmt,

Please review the following commits I made in branch nicks/protobuild:

93945870680da1f19906d06d3c892b21215e786b (2018-08-13 19:14:33 -0400)
proto: generate protobufs inside a container


Makefile
```diff
@@ -3,7 +3,11 @@
 all: proto
 
 proto:
-	protoc --go_out=plugins=grpc:../../../ -I. internal/proto/*.proto
+	docker build -t tilt-protogen -f Dockerfile.protogen .
+	docker rm tilt-protogen || exit 0
```
landism: fine as-is, but could change this to `docker rm -f tilt-protogen` to avoid the mildly distracting `Error: No such container: tilt-protogen` in the output

nicks: sadly, that doesn't work.

```
$ docker rm -f tilt-protogen
Error: No such container: tilt-protogen
```

tho maybe we should complain to docker about it; it would be nice if it worked the same as `rm`

landism: oops, that's what I get for assuming it just ran `rm`

PR #49, opened 2018-08-14
tilt: pass output from daemon to cli via streaming rpc instead of stdout/stderr

landism:
I think we'll want to rename `stdoutStderrWriter` to `logger` or something, and change the interface to support logging to different levels, in line with what Pat and Han are working on, but this is working as-is and I wanted to get eyes on the `io.Writer`/grpc/goroutine stuff since that's more likely to be weird.


internal/proto/daemon.proto
```diff
@@ -53,6 +58,10 @@ message GitRepo {
 //     repeated ServiceStatus statuses = 1;
 // }
 
-message CreateServiceReply {}
+message CreateServiceReply {
+    oneof body {
+        Output output = 1;
```
maiamcc: is there a reason for this type switch? what might it be besides an `Output` object?

internal/tiltd/tiltd.go
```diff
@@ -4,12 +4,13 @@ import (
...
 	"fmt"
 	"strings"
+	"io"
 )
 
 const Port = 10000
 
 type TiltD interface {
-	CreateService(ctx context.Context, k8sYaml string, dockerFileText string, mounts []Mount, steps []Cmd, dockerfileTag string) error
+	CreateService(ctx context.Context, k8sYaml string, dockerFileText string, mounts []Mount, steps []Cmd, dockerfileTag string, stdoutWriter io.Writer, stderrWriter io.Writer) error
```
maiamcc: imo you could get away with calling these e.g. `stdout` rather than `stdoutWriter` (though i have no real justification behind this except that if I were reading this code, i'd assume `stdout` was a writer unless otherwise noted)

also, if you have multiple args of the same type, can notate them as: `... stdout, stderr io.Writer`

internal/tiltd/tiltd_server/tiltd.go
```diff
@@ -35,7 +36,7 @@ func NewDaemon() (*Daemon, error) {
 	return &Daemon{b: b}, nil
 }
 
-func (d *Daemon) CreateService(ctx context.Context, k8sYaml string, dockerfile string, mounts []tiltd.Mount, steps []tiltd.Cmd, dockerfileTag string) error {
+func (d *Daemon) CreateService(ctx context.Context, k8sYaml string, dockerfile string, mounts []tiltd.Mount, steps []tiltd.Cmd, dockerfileTag string, stdoutWriter io.Writer, stderrWriter io.Writer) error {
 	// TODO(maia): a real entrypoint here
 	digest, err := d.b.BuildDocker(ctx, dockerfile, mounts, steps, tiltd.Cmd{})
```
maiamcc: we should prooobably be streaming the user output from the build step too -- don't need to implement now, but leave a todo?

internal/proto/stdout_stderr_writer.go
```diff
@@ -0,0 +1,133 @@
...
+}
+
+func makeStdoutStderrWriterFromReaderWriters(
+	sendOutput func(Output),
+	stdoutReader io.ReadCloser,
+	stdoutWriter io.WriteCloser,
+	stderrReader io.ReadCloser,
+	stderrWriter io.WriteCloser) StdoutStderrWriter {
+	s := protoStdoutStderrWriter{}
+	s.stdoutReader, s.stdoutWriter, s.stderrReader, s.stderrWriter = stdoutReader, stdoutWriter, stderrReader, stderrWriter
```
maiamcc: imo more readable/idiomatic to do as a struct literal:
```
s := protoStdoutStderrWriter{
    stdoutReader: stdoutReader
    ...
}
```

internal/proto/stdout_stderr_writer.go
```diff
@@ -0,0 +1,133 @@
...
+		}
+	}
+
+	switch len(ret) {
+	case 0:
+		return nil
+	case 1:
+		return ret[0]
+	default:
+		return errors.NewAggregate(ret)
```
maiamcc: I believe you can safely call `errors.NewAggregate` on a list of errs of any length -- if len == 0, returns nil, and if len == 1, just returns that error.

internal/proto/stdout_stderr_writer.go
```diff
@@ -0,0 +1,133 @@
...
+func (s protoStdoutStderrWriter) Close() error {
+	ret := make([]error, 0)
+	for _, closer := range []io.Closer{s.stdoutWriter, s.stderrWriter} {
+		err := closer.Close()
+		if err != nil {
+			ret = append(ret, err)
+		}
+	}
+
+	stdoutErr, stderrErr := <-s.stdoutDone, <-s.stderrDone
```
maiamcc: hmmm concurrency is not my strong suit but i thiiiiink you're probably going to want to wrap this in a timeout somehow -- otherwise you risk blocking forever? cc @dbentley @nicks who actually know this stuff.

internal/proto/stdout_stderr_writer_test.go
```diff
@@ -0,0 +1,86 @@
...
+	buf := bytes.Buffer{}
+	sendOutput := func(output Output) {
+		fmt.Printf("writing '%v'\n", string(output.Stdout))
+		buf.Write(output.Stdout)
+	}
+	outputStream := MakeStdoutStderrWriter(sendOutput)
+	outputStream.GetStdoutWriter().Write([]byte("hello"))
+	outputStream.GetStdoutWriter().Write([]byte(" world"))
+	outputStream.Close()
+	assert.Equal(t, "hello world", string(buf.Bytes()))
```
maiamcc: or `buf.String()`

internal/proto/stdout_stderr_writer_test.go
```diff
@@ -0,0 +1,86 @@
...
+type failingReadCloser struct {
+	name string
+	underlying io.ReadCloser
+	readFails bool
+	closeFails bool
+}
+
+var _ io.ReadCloser = failingReadCloser{}
+
+func (f failingReadCloser) Close() error {
```
maiamcc: niiiiice!

internal/proto/stdout_stderr_writer_test.go
```diff
@@ -0,0 +1,86 @@
...
+	if f.closeFails {
+		return fmt.Errorf("closing %v fails!", f.name)
+	} else {
+		return f.underlying.Close()
```
maiamcc: can also attach a `*testing.T` to your `failingReadCloser` and `t.Fatal` on an UNexpected error, only return the expected error. then your test can just assert that there's a non-nil err, rather than string-parsing error messages.

internal/proto/stdout_stderr_writer_test.go
```diff
@@ -0,0 +1,86 @@
...
+)
+
+func TestOutputStreamSuccess(t *testing.T) {
+	buf := bytes.Buffer{}
+	sendOutput := func(output Output) {
+		fmt.Printf("writing '%v'\n", string(output.Stdout))
+		buf.Write(output.Stdout)
+	}
+	outputStream := MakeStdoutStderrWriter(sendOutput)
+	outputStream.GetStdoutWriter().Write([]byte("hello"))
```
maiamcc: check errors for `Write` calls?

internal/proto/stdout_stderr_writer_test.go
```diff
@@ -0,0 +1,86 @@
...
+		buf.Write(output.Stdout)
+	}
+	stdoutReader, stdoutWriter := io.Pipe()
+	stderrReader, stderrWriter := io.Pipe()
+	failingStdoutReader := failingReadCloser{name: "stdout", underlying: stdoutReader, readFails: true}
+	failingStderrReader := failingReadCloser{name: "stderr", underlying: stderrReader, readFails: true}
+	outputStream := makeStdoutStderrWriterFromReaderWriters(sendOutput, failingStdoutReader, stdoutWriter, failingStderrReader, stderrWriter)
+	outputStream.GetStdoutWriter().Write([]byte("hello"))
+	err := outputStream.Close()
+	for _, s := range []string{"reading stderr fails", "reading stdout fails"} {
```
maiamcc: ...wait why do the read errors show up in the close error? i'm confused.

internal/proto/stdout_stderr_writer_test.go
```diff
@@ -0,0 +1,86 @@
...
+	outputStream.GetStdoutWriter().Write([]byte("hello"))
+	err := outputStream.Close()
+	for _, s := range []string{"closing stderr fails", "closing stdout fails"} {
+		assert.True(t, strings.Contains(err.Error(), s), "error '%v' did not contain '%v'", err.Error(), s)
+	}
+}
+
+func TestOutputStreamReadError(t *testing.T) {
+	buf := bytes.Buffer{}
+	sendOutput := func(output Output) {
```
maiamcc: can pull this out of tests and put it up top as a common variable

internal/proto/stdout_stderr_writer.go
```diff
@@ -0,0 +1,133 @@
+package proto
```
nicks: it seems really weird for all this logic to live in the protobuf package...it doesn't really have anything to do with protos?

internal/proto/stdout_stderr_writer.go
```diff
@@ -0,0 +1,133 @@
...
+
+import (
+	"bytes"
+	"io"
+	"k8s.io/apimachinery/pkg/util/errors"
+	"os"
+)
+
+type StdoutStderrWriter interface {
+	GetStdoutWriter() io.Writer
```
nicks: in idiomatic Go, we usually don't include "Get" or the type name in the method name
https://golang.org/doc/effective_go.html#Getters
so this would really be called `Stdout() io.Writer`

internal/proto/stdout_stderr_writer.go
```diff
@@ -0,0 +1,133 @@
...
+	stdoutReader io.ReadCloser,
+	stdoutWriter io.WriteCloser,
+	stderrReader io.ReadCloser,
+	stderrWriter io.WriteCloser) StdoutStderrWriter {
+	s := protoStdoutStderrWriter{}
+	s.stdoutReader, s.stdoutWriter, s.stderrReader, s.stderrWriter = stdoutReader, stdoutWriter, stderrReader, stderrWriter
+
+	s.stdoutDone, s.stderrDone = make(chan error), make(chan error)
+
+	go func() {
```
nicks: these goroutines seem very complicated...let's talk a bit more about this in person

internal/proto/daemon.proto
landism: Theoretically before we were returning a `CreateServiceReply` that happened to contain no fields. This was an attempt to force myself to write the code so that we now kinda theoretically have either an `Output` or a non-output reply, but then I removed the non-output reply, since CreateServiceReply doesn't actually return anything, which leaves things weird. I'll undo that.

internal/tiltd/tiltd.go
landism: yeah, that's much prettier, thanks

internal/proto/stdout_stderr_writer.go
landism: yep, much

internal/proto/stdout_stderr_writer.go
landism: I worried a bit about that and then covered my eyes and rationalized it by "well, it's operating on `Output`s". I think that after the change you suggested, this is little enough code, and sufficiently `Output`-specific, it might actually be ok to leave in this package?

internal/proto/stdout_stderr_writer.go
landism: kinda embarrassed by how much less code your suggestion resulted in

internal/proto/stdout_stderr_writer.go
landism: So you can. Nice!

internal/proto/stdout_stderr_writer_test.go
landism: buffering :(

internal/tiltd/tiltd_server/tiltd.go
landism: how about a sticky?

internal/proto/stdout_stderr_writer.go
nicks: lol. Golang interfaces give you this great judo-power where it feels like you can get a lot more done by just...letting the objects lean into you rather than pushing them around

internal/proto/stdout_stderr_writer.go
```diff
@@ -0,0 +1,50 @@
...
+
+	stderr := output.GetStderr()
+	if stderr != nil {
+		_, err := os.Stderr.Write(stderr)
+		if err != nil {
+			errs = append(errs, err)
+		}
+	}
+
+	return errors.NewAggregate(errs)
```
nicks: this seems very specific to certain kubernetes patterns where they really do need an aggregate error. I don't think this is a good pattern to copy here.

it's generally easier to bail out at the first error, and easier to handle upstream (rather than make the caller figure out what to do with aggregates)

internal/proto/stdout_stderr_writer.go
nicks: :+1:

internal/k8s/apply.go
```diff
@@ -4,21 +4,25 @@ import (
...
+	c.Stdout = stdout
+
+	stderrBuf := &bytes.Buffer{}
+
+	c.Stderr = io.MultiWriter(stderrBuf, stderr)
 
 	err := c.Run()
 	if err != nil {
-		return fmt.Errorf("kubectl apply: %v\nstderr: %s", err, stderr.String())
+		return fmt.Errorf("kubectl apply: %v\nstderr: %s", err, stderrBuf.String())
```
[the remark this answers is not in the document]
landism: yep

PR #50, opened 2018-08-14
tilt: clean up error output

landism:
1. strip the grpc stuff when it's not relevant
2. don't repeat the error twice
3. don't print usage on failure

old:
```
$ tilt up blorgly_backend_local
2018/08/14 12:21:53 Running tiltd listening on 127.0.0.1:10000
fileName: "Dockerfile", fileTag: "gcr.io/blorg-dev/blorgly-devel-matt", cmds: [echo hi]
hello!Error: rpc error: code = Unknown desc = buildBaseWithMounts: ImageBuild: The command '/bin/sh -c cd $SRC_DIR; go build -o server; cp server /app/' returned a non-zero code: 1
Usage:
  tilt up <servicename> [flags]

Flags:
  -h, --help   help for up

rpc error: code = Unknown desc = buildBaseWithMounts: ImageBuild: The command '/bin/sh -c cd $SRC_DIR; go build -o server; cp server /app/' returned a non-zero code: 1
```

new:
```
$ tilt up blorgly_backend_local
2018/08/14 14:13:29 Running tiltd listening on 127.0.0.1:10000
fileName: "Dockerfile", fileTag: "gcr.io/blorg-dev/blorgly-devel-matt", cmds: [echo hi]
buildBaseWithMounts: ImageBuild: The command '/bin/sh -c cd $SRC_DIR; go build -o server; cp server /app/' returned a non-zero code: 1rpc error: code = Unknown desc = buildBaseWithMounts: ImageBuild: The command '/bin/sh -c cd $SRC_DIR; go build -o server; cp server /app/' returned a non-zero code: 1
```


internal/cli/cli.go
```diff
@@ -9,8 +9,9 @@ import (
 
 func Execute() {
 	rootCmd := &cobra.Command{
-		Use:   "tilt",
-		Short: "tilt is great, yo",
+		Use:          "tilt",
+		Short:        "tilt is great, yo",
```
maiamcc: ...we should probably fix this eventually lol

internal/cli/up.go
```diff
@@ -13,9 +17,10 @@ type upCmd struct{}
 
 func (c *upCmd) register() *cobra.Command {
 	cmd := &cobra.Command{
-		Use:   "up <servicename>",
-		Short: "stand up a service",
-		Args:  cobra.ExactArgs(1),
+		Use:           "up <servicename>",
+		Short:         "stand up a service",
+		Args:          cobra.ExactArgs(1),
+		SilenceErrors: true,
```
nicks: hmmm...it seems weird to use RunE (which means "handle errors upstream") and at the same use SilenceErrors (which means "don't handle errors upstream")

maybe we could switch the RunE to a Run, and do the error handling ourselves, like Pat did in the pets code?

internal/cli/cli.go
```diff
@@ -30,8 +30,11 @@ type tiltCmd interface {
 
 func addCommand(parent *cobra.Command, child tiltCmd) {
 	cobraChild := child.register()
-	cobraChild.RunE = func(_ *cobra.Command, args []string) error {
-		return child.run(args)
+	cobraChild.Run = func(_ *cobra.Command, args []string) {
+		err := child.run(args)
+		if err != nil {
+			log.Fatalf("Error: %v", err)
```
nicks: `log` isn't meant for user-facing output, use fmt.Fprintf/fmt.Printf and os.Exit

`log` is for when you have machine-readable logs that you might want json-formatted or timestamp-appended or whatever

PR #53, opened 2018-08-14
tiltfile sends entrypoint to daemon, daemon attaches entrypoint to docker image

maiamcc:


internal/proto/daemon.proto
```diff
@@ -18,7 +18,8 @@ message Service {
     string dockerfile_text = 2; // base, e.g. no code
     repeated Mount mounts = 3;
     repeated Cmd steps = 4;
-    string dockerfile_tag = 5; // e.g. gcr.io/blorgly/blorgly-dev
+    Cmd entrypoint = 5;
```
jazzdan: I'm pretty sure protobuf doesn't want you to change the meaning of a field after you add it. If the fifth field in a message is a string, it should always be a string. This is so definitions can be backwards compatible. So in this case we could instead add `Cmd entrypoint = 6` and leave `dockerfile_tag`.

Is that true @nicks?

maiamcc: yeahhhh i know but that seems more relevant after the API is like, shipped to the world and is in use. i figure when we're still all internal we can change the proto at will for the sake of cleanliness...?

internal/build/docker_test.go
```diff
@@ -334,6 +334,22 @@ type pathContent struct {
...
+		f.t.Fatal(err)
+	}
+
+	out, err := f.dcli.ContainerLogs(ctx, cId, types.ContainerLogsOptions{ShowStdout: true, ShowStderr: true})
+	if err != nil {
+		f.t.Fatal(err)
+	}
+
+	output := &strings.Builder{}
+	io.Copy(output, out)
```
nicks: ioutil.ReadAll would be a better choice here

internal/build/docker_test.go
```diff
@@ -334,6 +334,22 @@ type pathContent struct {
...
+	if err != nil {
+		f.t.Fatal(err)
+	}
+
+	out, err := f.dcli.ContainerLogs(ctx, cId, types.ContainerLogsOptions{ShowStdout: true, ShowStderr: true})
+	if err != nil {
+		f.t.Fatal(err)
+	}
+
+	output := &strings.Builder{}
```
nicks: you probably want a `defer out.Close()` here

internal/proto/daemon.proto
jazzdan: I won't die on this hill but I think it's worth practicing being correct. I actually would have expected the proto compiler to fail though, so maybe I'm wrong.

nicks: ya, never change the tag numbers. this is a good habit to get into.

the tag numbers don't necessarily need to be in ascending order; it's perfectly reasonable to do

```
Cmd entrypoint = 6;
string dockerfile_tag = 5; 
```

if field organization is what you're concerned about

internal/tiltfile/tiltfile.go
```diff
@@ -17,12 +17,16 @@ type Tiltfile struct {
 }
 
 func makeSkylarkDockerImage(thread *skylark.Thread, fn *skylark.Builtin, args skylark.Tuple, kwargs []skylark.Tuple) (skylark.Value, error) {
-	var dockerfileName, dockerfileTag skylark.String
-	err := skylark.UnpackArgs(fn.Name(), args, kwargs, "docker_file_name", &dockerfileName, "docker_file_tag", &dockerfileTag)
+	var dockerfileName, entrypoint, dockerfileTag skylark.String
```
nicks: out of curiosity, why are we unpacking these as skylark.Strings instead of as plain strings? it's not like we ever need them as skylark.Strings again, and it seems like we might as well let UnpackArgs do the unpacking work for us

landism: pretty much just ignorance - at the time I wrote this, I was unaware UnpackArgs could handle plain strings, and I thought we'd need them in the data structure as strings so that we could do other things like `print(docker_image)` from the Tiltfile. It turns out these were both unfounded.

internal/build/docker_test.go
maiamcc: fyi @jazzdan -- i didn't know about this thing, did you?!

internal/tiltfile/tiltfile.go
```diff
@@ -142,8 +152,14 @@ func (tiltfile Tiltfile) GetServiceConfig(serviceName string) (*proto.Service, e
...
+		dockerCmds = append(dockerCmds, toBashCmd(cmd))
 	}
 
-	return &proto.Service{K8SYaml: k8sYaml, DockerfileText: dockerFileText, Mounts: mounts, Steps: dockerCmds, DockerfileTag: dockerFileTag}, nil
+	return &proto.Service{K8SYaml: k8sYaml, DockerfileText: dockerFileText,
+		Mounts: mounts, Steps: dockerCmds, Entrypoint: entrypointCmd,
+		DockerfileTag: dockerFileTag}, nil
+}
+
+func toBashCmd(cmd string) *proto.Cmd {
```
landism: we're not using bash any more - maybe s/Bash/Shell?

internal/tiltfile/skylarkTypes.go
```diff
@@ -112,9 +114,9 @@ func (*dockerImage) Hash() (uint32, error) {
 func (d *dockerImage) Attr(name string) (skylark.Value, error) {
 	switch name {
 	case "file_name":
-		return d.fileName, nil
+		return skylark.String(d.fileName), nil
```
maiamcc: uhh @landism / @nicks am i doing this right?

nicks: :+1:

internal/build/docker_test.go
jazzdan: Nope, TIL!

PR #60, opened 2018-08-15
build: get digest from push output

jazzdan:
With this PR our YAML has the correct image and the container can pull the image from GKE.

### DIFF
--- a/internal/build/docker.go
+++ b/internal/build/docker.go
@@ -35,7 +35,13 @@ type localDockerBuilder struct {
 
 type Builder interface {
 	BuildDocker(ctx context.Context, baseDockerfile string, mounts []tiltd.Mount, steps []tiltd.Cmd, entrypoint tiltd.Cmd) (digest.Digest, error)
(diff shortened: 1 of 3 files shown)


jazzdan:
@nicks this test fails locally for me too but I can't see what I would have done to cause it, any ideas?

internal/build/docker.go
```diff
@@ -370,17 +376,35 @@ func readDockerOutput(reader io.Reader) (*json.RawMessage, error) {
...
 	}
 	if aux == nil {
-		return "", fmt.Errorf("getDigestFromOutput: No results found in docker output")
+		return "", fmt.Errorf("getDigestFromBuildOutput: No results found in docker output")
 	}
 	return getDigestFromAux(*aux)
 }
 
+func getDigestFromPushOutput(reader io.Reader) (digest.Digest, error) {
+	aux, err := readDockerOutput(reader)
```
maiamcc: fwiw i find it confusing to call this variable "aux" -- it only makes sense if you know the format of the docker output

PR #67, opened 2018-08-15
tilt: handle/log errors on docker build steps

landism:


internal/build/docker.go
```diff
@@ -195,10 +195,13 @@ func (l *localDockerBuilder) execStepsOnImage(ctx context.Context, baseDigest di
 	imageWithSteps := string(baseDigest)
 	for _, s := range steps {
 		cId, err := l.startContainer(ctx, imageWithSteps, &s)
+		if err != nil {
```
landism: I'm not sure if the ignored errs here and below were intentional or if I'm breaking things by checking them...I'm happy to change these back to ignored and add a comment for posterity

landism:
> LGTM! Add a test? :)

oh hey yeah that's a good idea

internal/build/docker.go
maiamcc: aah no you are totally correct, those should be handled, i messed up 😳

internal/build/docker.go
```diff
@@ -247,7 +250,23 @@ func (l *localDockerBuilder) startContainer(ctx context.Context, imgRef string,
 		if err != nil {
 			return "", err
 		}
-	case <-statusCh:
+	case status := <-statusCh:
+		if status.Error != nil {
+			return "", errors.New(status.Error.Message)
+		}
+		// TODO(matt) feed this reader into the logger
+		r, err := l.dcli.ContainerLogs(ctx, containerID, types.ContainerLogsOptions{ShowStdout: true, ShowStderr: true})
```
maiamcc: ah, whelp, i believe we already had a func implementing some of this? `startContainerWithOutput` -- so whenever you actually do something with the output, you should remember to update that func too to remove any redundancies.

PR #69, opened 2018-08-15
tilt: composite_service WIP

pmt:
it's underway!


nicks:
unit test?

internal/tiltfile/tiltfile.go
```diff
@@ -72,6 +72,16 @@ func runLocalCmd(thread *skylark.Thread, fn *skylark.Builtin, args skylark.Tuple
 	return skylark.String(out), nil
 }
 
+func makeSkylarkCompositeService(thread *skylark.Thread, fn *skylark.Builtin, args skylark.Tuple, kwargs []skylark.Tuple) (skylark.Value, error) {
+	var services []skylark.Value
+
+	for i := range args {
+		services = append(services, args[i])
```
landism: since we're not getting type checking from `unpackArgs`, we should probably return an error if `args[i]` isn't an instance of `k8sService`

pmt: Can you explain "instance of `k8sService`"? I'm confused by the wording - I'm not sure how k8sService exactly relates to the composite services we're expecting? Are they the same? :thinking:

landism: `k8sService` as in an instance of the thing returned by the `k8s_service` function in a Tiltfile

(if a user misunderstands the api makes a tiltfile that calls `composite_service("blorgly_backend", "blorgly_frontend")`, it'd be nice to have a good error message here, rather than whatever random failure we'd get later, and this seems like the place to check it)

PR #70, opened 2018-08-15
build: Add a more robust assert function

nicks:
Hello @maiamcc, @jazzdan,

Please review the following commits I made in branch nicks/assert:

455544fde43fc576d090038d0f4c8bd832541a86 (2018-08-15 17:35:58 -0400)
build: Add a more robust assert function

### DIFF
--- a/internal/build/docker.go
+++ b/internal/build/docker.go
@@ -232,7 +232,10 @@ func (l *localDockerBuilder) startContainer(ctx context.Context, imgRef string, 
 	config := &container.Config{Image: imgRef}
 	if cmd != nil {
 		config.Cmd = cmd.Argv
+	} else {
+		config.Cmd = []string{"sh", "-c", "# NOTE(nick): a fake entrypoint"}
 	}
+
 	resp, err := l.dcli.ContainerCreate(ctx, config, nil, nil, "")
 	if err != nil {
 		return "", err
--- a/internal/build/docker_test.go
+++ b/internal/build/docker_test.go
@@ -1,10 +1,10 @@
 package build
 
 import (
+	"archive/tar"
 	"bytes"
 	"context"
 	"fmt"
-	"github.com/stretchr/testify/assert"
 	"io"
 	"io/ioutil"
 	"log"
@@ -15,6 +15,8 @@ import (
 	"sync"
 	"testing"
 	"time"
+
+	"github.com/stretchr/testify/assert"
 
 	"github.com/docker/distribution/reference"
 	"github.com/docker/docker/api/types"
@@ -223,7 +225,7 @@ func TestBuildOneStep(t *testing.T) {
 	}
 
 	contents := []pathContent{
-		pathContent{path: "hi", contents: "hello"},
+		pathContent{path: "hi", contents: "hello\n"},
 	}
 	f.assertFilesInImageWithContents(string(digest), contents)
 }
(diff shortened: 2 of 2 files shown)


internal/build/docker.go
```diff
@@ -232,7 +232,10 @@ func (l *localDockerBuilder) startContainer(ctx context.Context, imgRef string,
 	config := &container.Config{Image: imgRef}
 	if cmd != nil {
 		config.Cmd = cmd.Argv
+	} else {
+		config.Cmd = []string{"sh", "-c", "# NOTE(nick): a fake entrypoint"}
```
maiamcc: nit: it's not an entrypoint, it's a cmd :P so maybe `a no-op cmd`

also, consider a comment about the fact that an entrypoint takes precedence, for folks unfamiliar with that bit of docker triva

internal/build/docker_test.go
```diff
@@ -424,20 +426,43 @@ func (f *testFixture) startContainerWithOutput(ctx context.Context, ref string,
 
 func (f *testFixture) assertFilesInImageWithContents(ref string, contents []pathContent) {
 	ctx := context.Background()
-	var cmd strings.Builder
+	cId, err := f.b.startContainer(ctx, ref, &tiltd.Cmd{Argv: []string{"sh"}})
```
maiamcc: huhhh i thought this func blocked until the container closed! how is the container still around for you to copy stuff?

internal/build/docker_test.go
```diff
@@ -424,20 +426,43 @@ func (f *testFixture) startContainerWithOutput(ctx context.Context, ref string,
...
+		if err != nil {
+			f.t.Fatal(err)
+		}
+
+		f.assertFileInTar(tar.NewReader(reader), c)
 	}
-	cmdToRun := tiltd.Cmd{Argv: []string{"sh", "-c", cmd.String()}}
+}
+
+func (f *testFixture) assertFileInTar(tr *tar.Reader, content pathContent) {
```
maiamcc: confusing var name: rather than `content`, maybe `expected` or `pc` or something?

internal/build/docker_test.go
```diff
@@ -424,20 +426,43 @@ func (f *testFixture) startContainerWithOutput(ctx context.Context, ref string,
...
+				f.t.Fatalf("Error reading tar file: %v", err)
+			}
 
-	if strings.Contains(output, "ERROR:") {
-		f.t.Errorf("Failed to find one or more expected files in container with output:\n%s", output)
+			if contents.String() != content.contents {
+				f.t.Errorf("Wrong contents in %q. Expected: %q. Actual: %q",
+					content.path, content.contents, contents.String())
+			}
+			return // we found it!
```
maiamcc: 🙌

internal/build/docker_test.go
maiamcc: and actually on second thought, have you verified that this works even if the container has an entrypoint on it that causes it to exit immediately?

nicks: hmmm...define "works"? like i can copy files out of the container, so it must work?

internal/build/docker_test.go
nicks: good idea

internal/build/docker.go
nicks: will do

PR #71, opened 2018-08-15
build: first pass at building docker img over an existing image

maiamcc:
FYI this is still missing functionality -- waiting on better asserts in tests to make sure that we can successfully un-set an entrypoint on an existing image (so that we can actually exec steps on it)


internal/build/docker.go
```diff
@@ -75,6 +74,14 @@ func (l *localDockerBuilder) BuildDocker(ctx context.Context, baseDockerfile str
 	return newDigest, nil
 }
 
+func (l *localDockerBuilder) BuildDockerFromExisting(ctx context.Context, existing digest.Digest,
```
jazzdan: I love how easy this function was to implement, high five team!

PR #72, opened 2018-08-15
tiltd: helper func to make shell Cmd's

maiamcc:
i was missing this just now writing tests, so I figured we should have it

### DIFF
--- a/internal/tiltd/tiltd.go
+++ b/internal/tiltd/tiltd.go
@@ -42,3 +42,7 @@ type Cmd struct {
 func (c Cmd) EntrypointStr() string {
 	return fmt.Sprintf("ENTRYPOINT [\"%s\"]", strings.Join(c.Argv, "\", \""))
 }
+
+func ToShellCmd(cmd string) Cmd {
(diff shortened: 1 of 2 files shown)


internal/tiltd/tiltd.go
```diff
@@ -42,3 +42,7 @@ type Cmd struct {
 func (c Cmd) EntrypointStr() string {
 	return fmt.Sprintf("ENTRYPOINT [\"%s\"]", strings.Join(c.Argv, "\", \""))
 }
+
+func ToShellCmd(cmd string) Cmd {
```
jazzdan: Would it make sense for this to live on the Cmd struct?

maiamcc: the thing i'm trying to avoid is the boilerplate of wrapping something in
```
myCmd := Cmd{Argv: []string{"blah"}}
```
and if this func lived on the `Cmd` struct I'd still have to do that :-/

maiamcc:
Made new PR: #77 b/c too many merge conflicts here. I'll land from that one.

maiamcc:
Annnnd landed #77, so closing this one.

PR #73, opened 2018-08-15
image: remove savetofs in favor of incremental add

nicks:
Hello @maiamcc,

Please review the following commits I made in branch nicks/add:

d9a2ac5719ab1a0a7398a518752ab00c0acd3fb4 (2018-08-15 18:13:56 -0400)
image: remove savetofs in favor of incremental add

### DIFF
--- a/internal/image/fs.go
+++ b/internal/image/fs.go
@@ -41,21 +41,17 @@ func historyFromFS(dir *dirs.WindmillDir) (map[refKey][]historyEntry, error) {
 	return result, nil
 }
 
-func historyToFS(dir *dirs.WindmillDir, entriesMap map[refKey][]historyEntry) error {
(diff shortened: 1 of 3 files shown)


internal/image/fs.go
```diff
@@ -41,21 +41,17 @@ func historyFromFS(dir *dirs.WindmillDir) (map[refKey][]historyEntry, error) {
 	return result, nil
 }
 
-func historyToFS(dir *dirs.WindmillDir, entriesMap map[refKey][]historyEntry) error {
-	file, err := dir.OpenFile(imagesPath, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, os.FileMode(0644))
+func addHistoryToFS(dir *dirs.WindmillDir, ref refKey, entry historyEntry) error {
```
maiamcc: seems like this would also make sense as a method on a `historyEntry`...?

nicks: maybe! tho i'd like to keep the serialize/deserialize code in the same place

PR #74, opened 2018-08-15
build: response to comments and spec references

nicks:
Hello @maiamcc,

Please review the following commits I made in branch nicks/asserts2:

b3e9c6f6776e70fe312dea72d36b320fe192f65c (2018-08-15 18:20:10 -0400)
build: response to comments and spec references

### DIFF
--- a/internal/build/container.go
+++ b/internal/build/container.go
@@ -1,0 +1,34 @@
+package build
+
+import (
+	"github.com/docker/docker/api/types/container"
+	digest "github.com/opencontainers/go-digest"
+	"github.com/windmilleng/tilt/internal/model"
+)
+
+// Get a container config to run a container with a given tilt command instead of
+// the existing entrypoint. If cmd is nil, we run nothing.
+func containerConfigRunCmd(imgRef digest.Digest, cmd model.Cmd) *container.Config {
+	config := containerConfig(imgRef)
+
+	// In Docker, both the Entrypoint and the Cmd are used to determine what
+	// process the container runtime uses, where Entrypoint takes precedence over
+	// command. We set both here to ensure that we don't get weird results due
+	// to inheritance.
+	//
+	// If cmd is nil, we use a fake cmd that does nothing.
+	//
+	// https://github.com/opencontainers/image-spec/blob/master/config.md#properties
+	if cmd.Empty() {
+		config.Cmd = []string{"sh", "-c", "# NOTE(nick): a fake cmd"}
+	} else {
+		config.Cmd = cmd.Argv
+	}
+	config.Entrypoint = []string{}
+	return config
+}
+
+// Get a container config to run a container as-is.
+func containerConfig(imgRef digest.Digest) *container.Config {
+	return &container.Config{Image: string(imgRef)}
+}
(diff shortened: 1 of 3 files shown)


nicks:
@maiamcc i think we need to walk thru the failing test tomorrow. I'm pretty sure this test *should* be failing, but i don't totally understand what you were trying to test here.

nicks:
(in particular, the problem test is TestEntrypoint
https://github.com/windmilleng/tilt/blob/master/internal/build/docker_test.go#L269
which is currently asserting that BuildDocker runs the entry point, which seems wrong)

internal/build/docker.go
```diff
@@ -229,12 +229,23 @@ func (l *localDockerBuilder) imageWithEntrypoint(ctx context.Context, d digest.D
 // startContainer starts a container from the given image ref, exec'ing a command if given.
 // Returns the container id iff the container successfully starts up; else, error.
 func (l *localDockerBuilder) startContainer(ctx context.Context, imgRef string, cmd *tiltd.Cmd) (cId string, err error) {
+
 	config := &container.Config{Image: imgRef}
+
+	// In Docker, both the Entrypoint and the Cmd are used to determine what
+	// process the container runtime uses, where Entrypoint takes precedence over
+	// container. We set both here to ensure that we don't get weird results due
```
maiamcc: believe you meant "entrypoint takes precedence over cmd"

nicks:
per our offline discussion, we realized that there's a problem where the same interface is trying to use the same API for two different use-cases: run a container without an entrypoint, and run a container with the pre-existing entrypoint. I rejiggered the API to support both.

internal/build/container.go
```diff
@@ -0,0 +1,34 @@
...
+
+	// In Docker, both the Entrypoint and the Cmd are used to determine what
+	// process the container runtime uses, where Entrypoint takes precedence over
+	// command. We set both here to ensure that we don't get weird results due
+	// to inheritance.
+	//
+	// If cmd is nil, we use a fake cmd that does nothing.
+	//
+	// https://github.com/opencontainers/image-spec/blob/master/config.md#properties
+	if len(cmd.Argv) > 0 {
```
maiamcc: we have a `cmd.Empty()` func now fwiw

internal/build/docker.go
```diff
@@ -239,16 +239,9 @@ func (l *localDockerBuilder) imageWithEntrypoint(ctx context.Context, img digest
...
-func (l *localDockerBuilder) startContainer(ctx context.Context, imgRef string, cmd *model.Cmd) (cId string, err error) {
-	config := &container.Config{Image: imgRef}
-	if cmd != nil {
-		config.Cmd = cmd.Argv
-	} else {
-		config.Cmd = []string{"sh", "-c", "# NOTE(nick): a fake entrypoint"}
-	}
-
+// startContainer starts a container from the given config.
+// Returns the container id iff the container successfully starts up; else, error.
```
maiamcc: s/successfully starts up/successfully runs

internal/build/container.go
nicks: done!

internal/build/docker.go
nicks: done

PR #75, opened 2018-08-16
tilt: move code out of tiltd, add --watch that just loops

landism:
Problems:
1. discussion with the Dans yesterday about the interface between the daemon and the builder for `--watch` got us to something approximating the new `BuildAndDeployer` interface
2. as I was implementing --watch, more logic and dependencies were going into tiltd.go, which already had too much code, and it felt weird that all the other packages had to reference `tiltd` to get `Mount`, `Cmd`, etc.

Solutions:
1. Move the various classes used to pass around service definitions into a new `model` package
2. Move the logic for combining all the work together into a new `engine` package
3. Create a new `BuildAndDeployer` interface at an attempt to draw a clean line in the code and reduce the chances of toe-stepping.

If I'd realized how big this diff would end up being, I'd have done them in separate PRs (so obviously I'm making up for the big PR by giving you even more to read in the PR's description). Almost all of the diff is just code-shifting, with the important things to review being upper.go and build_and_deployer.go (plus the package change).

I also lumped a dummy looping `--watch` implementation in the second commit, to motivate this PR.

### DIFF
--- a/internal/cli/up.go
+++ b/internal/cli/up.go
@@ -24,7 +24,7 @@ func (c *upCmd) register() *cobra.Command {
 		Args:  cobra.ExactArgs(1),
 	}
 
-	cmd.Flags().BoolVar(&c.watch, "watch", false, "any started services will be automatically redeployed when files in their repos change")
+	cmd.Flags().BoolVar(&c.watch, "watch", false, "any started services will be automatically rebuilt and redeployed when files in their repos change")
 
 	return cmd
 }
--- a/internal/cli/up.go
+++ b/internal/cli/up.go
@@ -3,8 +3,6 @@ package cli
 import (
 	"context"
 	"errors"
-	"log"
-
 	"github.com/spf13/cobra"
 	"github.com/windmilleng/tilt/internal/proto"
 	"github.com/windmilleng/tilt/internal/tiltd/tiltd_client"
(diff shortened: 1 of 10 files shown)


internal/cli/up.go
```diff
@@ -23,6 +24,8 @@ func (c *upCmd) register() *cobra.Command {
 		Args:  cobra.ExactArgs(1),
 	}
 
+	cmd.Flags().BoolVar(&c.watch, "watch", false, "any started services will be automatically redeployed when files in their repos change")
```
jazzdan: nit: they won't necessarily be redeployed, but they will definitely be updated with their new code.

internal/tiltd/tiltd.go
```diff
@@ -1,44 +1,9 @@
...
-)
-
 const Port = 10000
 
-type TiltD interface {
-	CreateService(ctx context.Context, k8sYaml string, dockerFileText string, mounts []Mount,
-		steps []Cmd, entryfile Cmd, dockerfileTag string, stdout io.Writer, stderr io.Writer) error
-	SetDebug(ctx context.Context, mode bool)
-}
+type TiltD interface{}
```
jazzdan: can we delete this now?

internal/cli/up.go
landism: In my mind, pushing the new code to production might not be an image or k8s deployment, but it is a service deployment. I don't know if that would be the common understanding.

internal/tiltd/tiltd.go
landism: ah, yeah, thanks

PR #76, opened 2018-08-16
build: add an assert that files are missing from a container

nicks:
Hello @maiamcc,

Please review the following commits I made in branch nicks/not:

de4ef6cc6ece40619307d0a88b1750f5f032ffee (2018-08-16 11:28:11 -0400)
build: add an assert that files are missing from a container

### DIFF
--- a/internal/build/docker_test.go
+++ b/internal/build/docker_test.go
@@ -91,11 +91,11 @@ func TestMount(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	pcs := []pathContent{
-		pathContent{path: "/src/hi/hello", contents: "hi hello"},
-		pathContent{path: "/src/sup", contents: "my name is dan"},
+	pcs := []expectedFile{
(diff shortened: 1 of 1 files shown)


internal/build/docker_test.go
```diff
@@ -441,6 +465,19 @@ func (f *testFixture) assertFilesInImageWithContents(ref string, contents []path
 	}
 }
 
+func (f *testFixture) assertFileNotInImage(ref string, path string) {
```
maiamcc: bluh it would be great if you could pass a single blob of expected info to a single `assertFiles` func and have a single `startContainer` call assert X files exist with contents Y, and Z files do not exist. because an additional `startContainer` call is non-trivial overhead and our test suite is already super slow as it is. (if this feels too finicky i'm happy to take it over)

nicks: done!

nicks: ah sorry now it's done! there were some merge issues.

nicks:
ya, i like that! flipped around some of the names.

PR #79, opened 2018-08-16
tilt: global debug flag -> per-request loglevel

landism:
1. use a `loglevel int` instead of `debug bool`, so we can have choose between basic stuff for users (e.g., "starting service") or spammy debug stuff (e.g., "if you ever see this, tell Han she was right about that bug") at runtime
2. attach the log level to the request context instead of setting it globally
3. add one sample usage: `logger.Get(ctx).Verbose("building and deploying")`


internal/logger/logger.go
```diff
@@ -0,0 +1,57 @@
...
+	l.write(VerboseLvl, format, a...)
+}
+
+func (l logger) Debug(format string, a ...interface{}) {
+	l.write(DebugLvl, format, a...)
+}
+
+func (l logger) write(level Level, format string, a ...interface{}) {
+	if l.level >= level {
+		fmt.Printf(format+"\n", a...)
```
nicks: maybe do a separate fmt.Println("") instead of concatenating strings?

internal/engine/upper.go
```diff
@@ -15,6 +16,7 @@ func UpService(ctx context.Context, service model.Service, watch bool, stdout io
 	if watch {
 		for {
 			// TODO(matt) actually wait for a file to change instead of building on a loop
+			logger.Get(ctx).Verbose("building and deploying")
```
nicks: do we expect this to be user-visible?

internal/proto/daemon.proto
```diff
@@ -65,9 +64,16 @@ message Output {
...
+enum LogLevel {
+    INFO = 0;
+    VERBOSE = 1;
+    DEBUG = 2;
+}
+
 message CreateServiceRequest {
     Service service = 1;
     bool watch = 2;
+    LogLevel log_level = 3;
```
nicks: i think it's important to be specific of who the expected audience of this log level is. How do I decide which level to use? how do i decide whether to send progress information to the log vs. the CreateServiceReply? there should be with written comments documenting the decision

internal/engine/upper.go
landism: yes, if the user runs with loglevel >= verbose

the message should change (like, at least include a service name), but this PR is more about allowing logging than polishing the UI

internal/proto/daemon.proto
landism: re: log vs CreateServiceReply - my plan is to hook the logger up to the CreateServiceReply. maybe the name "logger" is misleading if this is what we're using to report output to the user?

I've added comments to the `Logger` interface, which might be the least bad place to put them, but open to thoughts.

internal/logger/logger.go
landism: is there an advantage to that, or a matter of taste?

(fwiw, that option kinda goes away when this is hooked up to the grpc stream)

PR #80, opened 2018-08-16
service: manager tracks services

jazzdan:


internal/proto/server_adaptor.go
```diff
@@ -4,31 +4,38 @@ import (
...
 
-func NewGRPCServer() *GRPCServer {
-	return &GRPCServer{}
+func NewGRPCServer() *grpcServer {
+	sm := service.NewMemoryManager()
+	return &grpcServer{sm: sm}
 }
 
-var _ DaemonServer = &GRPCServer{}
+var _ DaemonServer = &grpcServer{}
```
jazzdan: this doesn't need to be exported.

internal/service/manager.go
```diff
@@ -0,0 +1,79 @@
...
+	_, servicePresent := m.services[n]
+	return servicePresent
+}
+
+func (m *memoryManager) List() []model.Service {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+
+	v := make([]model.Service, len(m.services))
+	fmt.Printf("%+v\n", len(v))
```
nicks: :eyes:

internal/service/manager.go
```diff
@@ -0,0 +1,79 @@
...
+
+	i := 0
+	for _, s := range m.services {
```
nicks: usually you would write this as
```
for i, s := range m.services
```
rather than manage `i` yourself haha

jazzdan: wait, but in this case isn't the first variable the key, not an int?

internal/service/manager.go
maiamcc: 👀

internal/service/manager.go
```diff
@@ -0,0 +1,79 @@
...
+		v[i] = s
```
maiamcc: why not just append?

internal/service/manager.go
nicks: oh nm ignore me, i forgot this is a map

internal/service/manager_test.go
```diff
@@ -0,0 +1,108 @@
...
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestAdd(t *testing.T) {
+	f := newServiceManagerFixture(t)
+
+	s := model.Service{Name: model.ServiceName("hello")}
+	err := f.sm.Add(s)
+	assert.NoError(t, err)
```
maiamcc: i don't think this stops function execution if there IS an error, which will result in confusing nil pointer shit further along in execution. should do:
```
if assert.NoError(t, err) {
    <rest of test>
}
```
(also, noting the func that threw the error (`assert.NoError(t, err, "ListServices")` is A+ helpful in debugging if you feel like doing it.)

internal/service/manager.go
jazzdan: The problem with appending is that you'd allocate a zero length slice up front and then you'd keep expanding it as you append. Not a big deal, but an easy optimization to make.

internal/service/manager_test.go
jazzdan: Good call!

PR #83, opened 2018-08-16
build: inc. build doesn't take its own entrypoint, preserves entrypoint of existing img

maiamcc:
Hello @jazzdan, @nicks,

Please review the following commits I made in branch maiamcc/inc-build-entrypoint:

bfdf19c6361cdb3fd3454e84a6540ada154c7eb9 (2018-08-16 15:05:06 -0400)
build: inc. build doesn't take its own entrypoint, preserves entrypoint of existing img

fb7b5b1037d2b626ea0771c92c3f8293d8afce31 (2018-08-16 14:56:09 -0400)
wip

### DIFF
--- a/internal/build/docker.go
+++ b/internal/build/docker.go
@@ -38,7 +38,7 @@ type localDockerBuilder struct {
 }
 
 type Builder interface {
-	BuildDocker(ctx context.Context, baseDockerfile string, mounts []model.Mount, steps []model.Cmd, entrypoint *model.Cmd) (digest.Digest, error)
+	BuildDocker(ctx context.Context, baseDockerfile string, mounts []model.Mount, steps []model.Cmd, entrypoint model.Cmd) (digest.Digest, error)
 	BuildDockerFromExisting(ctx context.Context, existing digest.Digest, mounts []model.Mount, steps []model.Cmd) (digest.Digest, error)
 	PushDocker(ctx context.Context, name reference.Named, dig digest.Digest) (digest.Digest, error)
 }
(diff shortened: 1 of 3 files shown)


internal/build/docker.go
```diff
@@ -38,8 +38,8 @@ type localDockerBuilder struct {
 }
 
 type Builder interface {
-	BuildDocker(ctx context.Context, baseDockerfile string, mounts []model.Mount, steps []model.Cmd, entrypoint model.Cmd) (digest.Digest, error)
-	BuildDockerFromExisting(ctx context.Context, existing digest.Digest, mounts []model.Mount, steps []model.Cmd, entrypoint model.Cmd) (digest.Digest, error)
+	BuildDocker(ctx context.Context, baseDockerfile string, mounts []model.Mount, steps []model.Cmd, entrypoint *model.Cmd) (digest.Digest, error)
```
nicks: when would you pass a nil entrypoint?

as a style guideline, Go encourages people to treat nil and the default value as equivalent. It really frowns on having two separate "zero" values that do different things. append is a good example, append([]int(nil), 1) and append([]int{}, 1) are equivalent

maiamcc: hmm that's fair, I was trying to differentiate the case of calling BuildDocker as the first time an image is build -- so it MUST be given a valid entrypoint (except also maybe it doesn't have to be given a valid entrypoint anymore) and calling it from within BuildDockerFromExisting, where it should never be passed an entrypoint bc the behavior of BuildDockerFromExisting is that it uses the existing img's entrypoint. but you're right that this is super confusing, let me rejigger it.

nicks: ya -- i might go the opposite direction, make this take a Cmd value (not a pointer), and return an error if you get an empty Cmd

nicks: (and ya, totally understand how you got here, probably in need of a helper function)

maiamcc: @nicks how's this? this org feels better to me b/c i'm trying to different `BuildDocker` the func that we use to build a docker image for the first time from `buildDocker`, the helper func that is used both when building an img from scratch or on top of an existing image... but lmk if this is confusing

PR #84, opened 2018-08-16
build/test: test prettification

maiamcc:
Hello @jazzdan, @nicks,

Please review the following commits I made in branch maiamcc/prettify-tests:

741c9285d5b4c84a3b380d0733993b3703c50a1f (2018-08-16 15:18:29 -0400)
build/test: test prettification

bfdf19c6361cdb3fd3454e84a6540ada154c7eb9 (2018-08-16 15:05:06 -0400)
build: inc. build doesn't take its own entrypoint, preserves entrypoint of existing img

fb7b5b1037d2b626ea0771c92c3f8293d8afce31 (2018-08-16 14:56:09 -0400)
wip

### DIFF
--- a/internal/build/docker_test.go
+++ b/internal/build/docker_test.go
@@ -86,7 +86,7 @@ func TestMount(t *testing.T) {
 		ContainerPath: "/src",
 	}
 
-	digest, err := f.b.BuildDocker(context.Background(), simpleDockerfile, []model.Mount{m}, []model.Cmd{}, model.Cmd{})
+	digest, err := f.b.BuildDocker(f.ctx, simpleDockerfile, []model.Mount{m}, []model.Cmd{}, model.Cmd{})
 	if err != nil {
 		t.Fatal(err)
 	}
@@ -115,7 +115,7 @@ func TestMultipleMounts(t *testing.T) {
 		ContainerPath: "goodbye_there",
 	}
 
-	digest, err := f.b.BuildDocker(context.Background(), simpleDockerfile, []model.Mount{m1, m2}, []model.Cmd{}, model.Cmd{})
+	digest, err := f.b.BuildDocker(f.ctx, simpleDockerfile, []model.Mount{m1, m2}, []model.Cmd{}, model.Cmd{})
 	if err != nil {
 		t.Fatal(err)
 	}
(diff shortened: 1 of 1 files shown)


internal/build/docker.go
```diff
@@ -38,8 +38,8 @@ type localDockerBuilder struct {
 }
 
 type Builder interface {
-	BuildDocker(ctx context.Context, baseDockerfile string, mounts []model.Mount, steps []model.Cmd, entrypoint model.Cmd) (digest.Digest, error)
-	BuildDockerFromExisting(ctx context.Context, existing digest.Digest, mounts []model.Mount, steps []model.Cmd, entrypoint model.Cmd) (digest.Digest, error)
+	BuildDocker(ctx context.Context, baseDockerfile string, mounts []model.Mount, steps []model.Cmd, entrypoint *model.Cmd) (digest.Digest, error)
```
nicks: when would you pass a nil entrypoint?

as a style guideline, Go encourages people to treat `nil` and the default value as equivalent. It really frowns on having two separate "zero" values that do different things. `append` is a good example, append([]int(nil), 1) and append([]int{}, 1) are equivalent

nicks: actually, i'll transfer this comment to the other pr and we can discuss it there; i didn't realize there were two

maiamcc: hmm that's fair, I was trying to differentiate the case of calling `BuildDocker` as the first time an image is build -- so it MUST be given a valid entrypoint (except also maybe it doesn't have to be given a valid entrypoint anymore) and calling it from within `BuildDockerFromExisting`, where it should never be passed an entrypoint bc the behavior of `BuildDockerFromExisting` is that it uses the existing img's entrypoint. but you're right that this is super confusing, let me rejigger it.

maiamcc: lol yeah whoops, i will also relocate

maiamcc:
@jazzdan @nicks try now. (def didn't force push this what are you talking about 🤐 )

nicks:
fwiw, i'm a big believer that all pushes should be forceful. it's important to assert your boundaries with git.

maiamcc:
😒

PR #85, opened 2018-08-16
tilt: --watch actually watches the filesystem

landism:


internal/engine/upper.go
```diff
@@ -2,24 +2,63 @@ package engine
...
 	"github.com/windmilleng/tilt/internal/logger"
 	"github.com/windmilleng/tilt/internal/model"
+	"github.com/windmilleng/tilt/internal/watch"
+
 	"io"
 )
 
-func UpService(ctx context.Context, service model.Service, watch bool, stdout io.Writer, stderr io.Writer) error {
-	bad, err := NewLocalBuildAndDeployer()
+type Upper struct {
```
landism: Upper/NewUpper are just to enable DI/testing

internal/engine/upper.go
```diff
@@ -2,24 +2,63 @@ package engine
...
-			buildToken, err = bad.BuildAndDeploy(ctx, service, buildToken)
-			if err != nil {
+			// TODO(matt) honor .gitignore / .dockerignore
+			// TODO(matt) buffer events a bit so that we're not triggering 10 builds when you change branches
+			select {
+			case err := <-watcher.Errors():
 				return err
+			case <-watcher.Events():
+				logger.Get(ctx).Verbose("file changed, rebuilding %v", service.Name)
+				buildToken, err = u.b.BuildAndDeploy(ctx, service, buildToken)
```
nicks: if you're auto-rebuilding every time, it seems like...builds will fail often? and we'll want to build them next time?

internal/engine/upper.go
```diff
@@ -2,24 +2,63 @@ package engine
...
+		for _, mount := range service.Mounts {
+			watcher.Add(mount.Repo.LocalPath)
+		}
+
 		for {
-			// TODO(matt) actually wait for a file to change instead of building on a loop
-			logger.Get(ctx).Verbose("building and deploying")
-			buildToken, err = bad.BuildAndDeploy(ctx, service, buildToken)
-			if err != nil {
+			// TODO(matt) honor .gitignore / .dockerignore
```
nicks: i'm going back and forth on whether the Upper should be responsible for this. i think the builder will also need to implement the rules to know what to send to the docker daemon? what do you think?

internal/engine/upper_test.go
```diff
@@ -0,0 +1,143 @@
...
+type fakeBuildAndDeployer struct {
+	startedServices []model.Service
+	cond            *sync.Cond
+}
+
+var _ BuildAndDeployer = &fakeBuildAndDeployer{}
+
+func (b *fakeBuildAndDeployer) BuildAndDeploy(ctx context.Context, service model.Service, token BuildToken) (BuildToken, error) {
+	b.startedServices = append(b.startedServices, service)
+	b.cond.L.Lock()
```
nicks: you don't need to hold the lock to broadcast. see:
https://golang.org/pkg/sync/#Cond.Broadcast

internal/engine/upper_test.go
```diff
@@ -0,0 +1,143 @@
...
+	if assert.NotNil(t, err) {
+		assert.Contains(t, err.Error(), "0 repos")
+	}
+}
+
+func TestUpper_UpWatchError(t *testing.T) {
+	f := newTestFixture(t)
+	mount := model.Mount{Repo: model.LocalGithubRepo{LocalPath: "/go"}, ContainerPath: "/go"}
+	service := model.Service{Name: "foobar", Mounts: []model.Mount{mount}}
+	f.b.cond.L.Lock()
```
nicks: let me come talk to you about these condition variables, i think you're confused about something

internal/engine/upper.go
landism: whoops, good catch

internal/engine/upper_test.go
landism: yeah, buffered channels are much prettier

internal/engine/upper.go
landism: I hadn't considered that the builder might be honoring the rules. I think that makes sense. It's probably also better for correctness that we aren't copying gitignored stuff (like build output) from the user's repo in the docker image.

PR #89, opened 2018-08-17
tilt: coalesce file change notifications

landism:
when we're notified of a file change, keep waiting for further file changes (potentially forever) until we've gone 200ms without seeing a file change, then send all the changes to `BuildAndDeploy` at once


internal/engine/upper.go
```diff
@@ -25,7 +35,34 @@ func NewUpper(manager service.Manager) (Upper, error) {
...
+	sleeper := time.Sleep
+	return Upper{b, watcherMaker, sleeper}, nil
+}
+
+//makes an attempt to read some events from `eventChan` so that multiple file changes that happen at the same time
+//from the user's perspective are grouped together.
+func (u Upper) coalesceEvents(eventChan chan fsnotify.Event) []fsnotify.Event {
+	var events []fsnotify.Event
+	doneWaitingForChanges := false
+	for !doneWaitingForChanges {
```
nicks: the double loop here seems overly complex. One of the great powers of `select` is to avoid this kind of nested-loops-to-handle-multiple-conditions

this would be a lot simpler as

```
timer := time.NewTimer(delay)
for true {
  select {
    case event := <- eventChan:
      timer.Reset(delay)
       ...
    case <-timer.C:
       ...
  }
}
```

internal/engine/upper.go
```diff
@@ -25,7 +35,34 @@ func NewUpper(manager service.Manager) (Upper, error) {
...
+		// keep grabbing changes until we've gone `watchBufferDurationInMs` without seeing a change
```
nicks: i'm not convinced that this is a good buffering strategy. the loop might never complete if there are continuous changes. 

one of the big problems that came up in live integration was codebases where the user accidentally watched a log directory so that the tool would just hang forever with no output or indication of why it was broken. you should talk to @jazzdan about how he ended up handling this.

jazzdan: Yeah we calculated the rate of file events, and if it was too high printed an error and crashed.

landism: I've made it so it'll wait a max of 10 seconds, and then just go with whatever it's seen so far.

internal/engine/upper.go
landism: yeah, that's much better. thanks.

internal/engine/upper.go
```diff
@@ -49,17 +92,23 @@ func (u Upper) Up(ctx context.Context, service model.Service, watchMounts bool,
...
 			case err := <-watcher.Errors():
 				return err
 			case event := <-watcher.Events():
-				logger.Get(ctx).Info("file changed, rebuilding %v", service.Name)
-				path, err := filepath.Abs(event.Name)
-				if err != nil {
-					return err
+				logger.Get(ctx).Info("files changed, rebuilding %v", service.Name)
+				events := []fsnotify.Event{event}
+				events = append(events, u.coalesceEvents(watcher.Events())...)
```
jazzdan: This feels weird to me. Instead of appending in this receive case, would it make more sense to make a new channel of coalesced events that this part of the code receives from?

internal/engine/upper.go
```diff
@@ -25,7 +42,33 @@ func NewUpper(manager service.Manager) (Upper, error) {
...
+	var events []fsnotify.Event
+	// keep grabbing changes until we've gone `watchBufferMinRestDuration` without seeing a change
+	minRestTimer := u.makeTimer(watchBufferMinRestDuration)
+
+	// but if we go too long before seeing a break (e.g., a process is constantly writing logs to that dir)
+	// then just send what we've got
+	timeout := u.makeTimer(watchBufferMaxDuration)
+
+	done := false
+	for !done {
```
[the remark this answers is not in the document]
landism: Yeah, I tried to do that first, but felt it was uglier to duplicate `return events` under both the second and third cases. I don't really have any strong arguments either way, though.

(and I learned you can't use `case a, b` in `select`s like you can `switch`es)

internal/engine/upper.go
landism: yeah, that's prettier

internal/engine/upper.go
```diff
@@ -25,7 +42,41 @@ func NewUpper(manager service.Manager) (Upper, error) {
...
+	return Upper{b, watcherMaker, time.After}, nil
+}
+
+//makes an attempt to read some events from `eventChan` so that multiple file changes that happen at the same time
+//from the user's perspective are grouped together.
+func (u Upper) coalesceEvents(eventChan <-chan fsnotify.Event) <-chan []fsnotify.Event {
+	ret := make(chan []fsnotify.Event)
+	go func() {
+		select {
+		case event := <-eventChan:
```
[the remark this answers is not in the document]
landism: Yeah, I was trying to ensure we could immediately close the channel when there are no events to read, so we could get on to the next iteration of the outer for loop, but that doesn't make any sense. I'm still working on my mental model for select and chans.

internal/engine/upper.go
```diff
@@ -25,7 +42,41 @@ func NewUpper(manager service.Manager) (Upper, error) {
...
+			minRestTimer := u.makeTimer(watchBufferMinRestDuration)
+
+			// but if we go too long before seeing a break (e.g., a process is constantly writing logs to that dir)
+			// then just send what we've got
+			timeout := u.makeTimer(watchBufferMaxDuration)
+
+			done := false
+			for !done {
+				select {
+				case event := <-eventChan:
```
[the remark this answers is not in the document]
landism: yep! hadn't seen that before. thanks.

PR #90, opened 2018-08-17
tilt: make logger talk to grpc, nix most references to stdout

landism:
not sure if it still makes sense for StdoutStderrWriter to still have two writers or be named such, but that can be a separate change anyway


internal/k8s/apply.go
```diff
@@ -6,19 +6,23 @@ import (
...
 
-func Apply(ctx context.Context, rawYAML string, stdout io.Writer, stderr io.Writer) error {
+func Apply(ctx context.Context, rawYAML string) error {
 	// TODO(dmiller) validate that the string is YAML and give a good error
 	c := exec.CommandContext(ctx, "kubectl", "apply", "-f", "-")
 	r := bytes.NewReader([]byte(rawYAML))
 	c.Stdin = r
 
-	c.Stdout = stdout
+	writer := logger.Get(ctx).Writer(logger.VerboseLvl)
```
nicks: this seems like something we should display to the user by default?

landism: sounds good

PR #93, opened 2018-08-17
add errcheck

jazzdan:


jazzdan:
oh yes, will do! Totally forgot about that.

jazzdan:
hmm ignoring stuff is annoying so @landism and I just decided to do this the right way and fix all of the errors, except the errors in generated files. Will do that on Monday.

PR #97, opened 2018-08-17
build: use exec instead of creating a new container on every step

nicks:
Hello @maiamcc, @jazzdan,

Please review the following commits I made in branch nicks/container:

59788d7d0dab5a7797a45e4f47e31ad219fb3a45 (2018-08-17 17:10:52 -0400)
build: use exec instead of creating a new container on every step

9b359d0a689d78b0ba8e5eab935af4e9d893b7c8 (2018-08-17 17:07:05 -0400)
build: add a benchmark


nicks:
before this change:
BenchmarkBuildTenSteps-4   	       1	12942196784 ns/op

after:
BenchmarkBuildTenSteps-4   	       1	2584671014 ns/op

internal/build/pool.go
```diff
@@ -0,0 +1,102 @@
...
+	}
+}
+
+// Claim a container, creating a new one if necessary.
+//
+// Once the container is claimed, no one is allowed to use it until it's released.
+//
+// TODO(nick): Right now, this always creates a new container.
+func (p containerPool) claimContainer(ctx context.Context, digest digest.Digest) (containerID, error) {
+	config := containerConfigRunCmd(digest, model.Cmd{Argv: []string{"tail", "-f", "/dev/null"}})
```
jazzdan: Why do we have to pass a Cmd here at all? Could we pass `nil`?

internal/build/pool.go
```diff
@@ -0,0 +1,102 @@
...
+	"fmt"
+	"log"
+
+	"github.com/docker/docker/api/types"
+	"github.com/docker/docker/client"
+	digest "github.com/opencontainers/go-digest"
+	"github.com/windmilleng/tilt/internal/model"
+)
+
+type containerPool struct {
```
jazzdan: Do we need a lock here?

nicks: added one!

internal/build/pool.go
```diff
@@ -0,0 +1,112 @@
...
+	if !ok {
+		return "", fmt.Errorf("container %s unknown to pool", baseDigest)
+	}
+
+	opts := types.ContainerCommitOptions{}
+	if !entrypoint.Empty() {
+		// Attach an entrypoint provided by the caller.
+		opts.Changes = []string{entrypoint.EntrypointStr()}
+	} else {
+		// When the container was claimed, we overwrote the entrypoint with tail -f /dev/null.
```
maiamcc: teeeechnically we overwrote `cmd` and  _un-set_ the entrypoint (http://bit.ly/2L92LuA). probs most accurate to say: "we overwrote the CMD and the ENTRYPOINT" or something.

internal/build/pool.go
```diff
@@ -0,0 +1,112 @@
...
+		}
+
+		if inspected.Config != nil && len(inspected.Config.Entrypoint) > 0 {
+			cmd := model.Cmd{Argv: inspected.Config.Entrypoint}
+			opts.Changes = []string{cmd.EntrypointStr()}
+		} else if inspected.Config != nil && len(inspected.Config.Cmd) > 0 {
+			cmd := model.Cmd{Argv: inspected.Config.Cmd}
+			opts.Changes = []string{cmd.EntrypointStr()}
+		} else {
+			cmd := model.Cmd{Argv: []string{"sh", "-c", "# NOTE(nick): no cmd found"}}
```
maiamcc: if you want: `model.ToShellCmd("# NOTE(nick): no cmd found")`

internal/build/pool.go
```diff
@@ -0,0 +1,112 @@
...
+		if err != nil {
+			return "", fmt.Errorf("error inspected image: %s", baseDigest)
+		}
+
+		if inspected.Config != nil && len(inspected.Config.Entrypoint) > 0 {
+			cmd := model.Cmd{Argv: inspected.Config.Entrypoint}
+			opts.Changes = []string{cmd.EntrypointStr()}
+		} else if inspected.Config != nil && len(inspected.Config.Cmd) > 0 {
+			cmd := model.Cmd{Argv: inspected.Config.Cmd}
+			opts.Changes = []string{cmd.EntrypointStr()}
```
maiamcc: should be setting as the cmd rather than the entrypoint, no?

internal/build/pool.go
```diff
@@ -0,0 +1,112 @@
...
+		}
+	}
+
+	commit, err := p.docker.ContainerCommit(ctx, string(cID), opts)
+	if err != nil {
+		return "", fmt.Errorf("releaseContainer#Commit: %v", err)
+	}
+
+	go func() {
```
maiamcc: do we care about guaranteeing that this will finish?

internal/build/docker.go
```diff
@@ -72,18 +73,19 @@ func (l *localDockerBuilder) buildDocker(ctx context.Context, df Dockerfile,
 		return "", fmt.Errorf("buildFromDfWithFiles: %v", err)
 	}
 
-	newDigest, err := l.execStepsOnImage(ctx, baseDigest, steps)
-	if err != nil {
-		return "", fmt.Errorf("execStepsOnImage: %v", err)
-	}
+	cID, err := l.pool.claimContainer(ctx, baseDigest)
```
maiamcc: maybe docstring: "start a container from base image so we can exec steps on it" or something

internal/build/docker.go
```diff
@@ -189,79 +191,87 @@ func (l *localDockerBuilder) buildFromDfWithFiles(ctx context.Context, df Docker
...
+	attached, err := l.dcli.ContainerExecAttach(ctx, string(execID), types.ExecStartCheck{})
+	if err != nil {
+		return fmt.Errorf("execInContainer#attach: %v", err)
 	}
+	defer attached.Close()
 
-	id, err := l.dcli.ContainerCommit(ctx, cId, opts)
+	// TODO(nick): feed this reader into the logger
+	buf := bytes.NewBuffer(nil)
+	_, err = io.Copy(buf, attached.Reader)
```
maiamcc: i know at some point we discussed a way to read from this buf until we encountered EOF -- would that let us simplify the `inspect` loop below?

internal/build/docker.go
```diff
@@ -189,79 +191,87 @@ func (l *localDockerBuilder) buildFromDfWithFiles(ctx context.Context, df Docker
...
+			return fmt.Errorf("execInContainer#inspect: %v", err)
+		}
+
+		if inspected.Running {
+			continue
+		}
+
+		status := inspected.ExitCode
+		if status != 0 {
+			return fmt.Errorf("Failed with exit code %d. Output:\n%s", status, buf.String())
```
maiamcc: todo: option to print output even if exit code = 0 (verbose/debug logging?)

internal/build/docker.go
```diff
@@ -189,79 +191,87 @@ func (l *localDockerBuilder) buildFromDfWithFiles(ctx context.Context, df Docker
...
-		}
-
-		id, err := l.dcli.ContainerCommit(ctx, cId, types.ContainerCommitOptions{})
-		if err != nil {
-			return "", fmt.Errorf("containerCommit: %v", err)
-		}
-		imageWithSteps = digest.Digest(id.ID)
+func (l *localDockerBuilder) execInContainer(ctx context.Context, cID containerID, cmd model.Cmd) error {
+	created, err := l.dcli.ContainerExecCreate(ctx, string(cID), types.ExecConfig{
+		Cmd: cmd.Argv,
```
maiamcc: tesseract code for this sets a few additional args, are those necessary? http://bit.ly/2Lbq86u

internal/build/container.go
```diff
@@ -6,6 +6,9 @@ import (
 	"github.com/windmilleng/tilt/internal/model"
 )
 
+type containerID string
```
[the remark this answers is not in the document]
nicks: it's a Go best practice to declare types for things like this, to make parameters more self-documenting. (e.g., any string wouldn't satisfy this method, it has to be a container id)

internal/build/container.go
```diff
@@ -6,6 +6,9 @@ import (
...
+type execID string
+
 // Get a container config to run a container with a given tilt command instead of
```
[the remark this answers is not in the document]
nicks: done

internal/build/docker.go
nicks: done

internal/build/docker.go
nicks: no, we don't want to attach stdout or tty here

internal/build/docker.go
nicks: i think you're confused. the existing code already reads from this buf until we encounter EOF. The inspect loop is a separate unrelated thing, because we have to wait until the exit code gets written.

internal/build/docker.go
nicks: there's already a TODO for that? before the loop

internal/build/pool.go
nicks: done

internal/build/pool.go
nicks: hmmm....we probably want to normalize on one, to reduce the number of special cases. if there's a particular case you're worried about, let's write a test for it; the implementation is less important.

internal/build/pool.go
nicks: no not really. this code is probably all going to get deleted and replaced soon anyway once we add re-use of containers

internal/build/container.go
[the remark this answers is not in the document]
maiamcc: huh, TIL

internal/build/docker.go
maiamcc: ...yes, i was indeed confused. that makes sense, thanks!

internal/build/docker.go
maiamcc: ah whoops

PR #101, opened 2018-08-20
tilt: service becomes array of services

pmt:
No longer a WIP!!!! I would love feedback


internal/tiltfile/tiltfile.go
```diff
@@ -114,6 +114,56 @@ func (tiltfile Tiltfile) GetServiceConfig(serviceName string) (*proto.Service, e
 		return nil, fmt.Errorf("error running '%v': %v", serviceName, err.Error())
 	}
 
+<<<<<<< HEAD
```
jazzdan: 😱

internal/tiltfile/tiltfile.go
```diff
@@ -114,11 +114,26 @@ func (tiltfile Tiltfile) GetServiceConfig(serviceName string) (*proto.Service, e
...
+		for _, cServ := range compService.cService {
+			return addToService(cServ)
+		}
+	case k8sService:
+		return addToService(service)
+	default:
+		return nil, fmt.Errorf("'%v' returned a '%v', but it needs to return a k8s_service or composite_service", serviceName, val.Type())
 	}
 
+	// what do I return here?
```
pmt: ??????? confused by how to return correctly

internal/cli/up.go
```diff
@@ -62,12 +62,13 @@ func (c *upCmd) run(args []string) error {
 		return err
 	}
 
-	req := proto.CreateServiceRequest{Service: service, Watch: c.watch, LogLevel: proto.LogLevel(logLevel())}
-
-	err = dCli.CreateService(ctx, req)
-	s, ok := status.FromError(err)
-	if ok && s.Code() == codes.Unknown {
-		return errors.New(s.Message())
+	for i := range service {
```
jazzdan: maybe rename this to `services` since it now contains many services, not just one?

internal/cli/up.go
```diff
@@ -62,12 +62,13 @@ func (c *upCmd) run(args []string) error {
...
+		req := proto.CreateServiceRequest{Service: service[i], Watch: c.watch, LogLevel: proto.LogLevel(logLevel())}
```
jazzdan: Do we think it would be more efficient to change the RPC so that the _server_ is responsible for handling multiple services? It might be more efficient since then the client would only have to send one RPC, but the difference is probably negligible.

internal/cli/up.go
```diff
@@ -62,12 +62,13 @@ func (c *upCmd) run(args []string) error {
...
+		err = dCli.CreateService(ctx, req)
+		s, ok := status.FromError(err)
+		if ok && s.Code() == codes.Unknown {
+			return errors.New(s.Message())
```
jazzdan: If we fail to create one service, do we want to fail to create all subsequent services?

internal/cli/up.go
pmt: Hmm. I don't know the answer to this question. Also wouldn't really know how to implement.

internal/cli/up.go
pmt: Is this rhetorical and a suggestion or is it a question :^)

jazzdan: It's an actual question! :)

internal/cli/up.go
jazzdan: No worries, I think this is fine for now. We can always make it more efficient later.

landism: yeah, efficiency aside, we need to do this on the server side, since the rpc won't return with `--watch`

internal/cli/up.go
landism: Aborting after one failure seems like the best option right now. If replicated is starting 30 services with a single `tilt up`, it's probably better to fail early and not make them scroll up to see if things actually worked.

pmt: Is ... that what it's doing? Sorry this sort of error handling is a little unique to me

internal/tiltfile/tiltfile.go
```diff
@@ -114,11 +114,32 @@ func (tiltfile Tiltfile) GetServiceConfig(serviceName string) (*proto.Service, e
...
+			return nil, err
+		}
+		return []*proto.Service{s}, nil
+
+	default:
+		return nil, fmt.Errorf("'%v' returned a '%v', but it needs to return a k8s_service or composite_service", serviceName, val.Type())
 	}
+}
 
+func addToService(service k8sService) (*proto.Service, error) {
```
landism: this isn't really adding anything any more - maybe rename to something like `skylarkServiceToProto`?

internal/cli/up.go
pmt: Do I leave it :sweat_smile:

PR #103, opened 2018-08-20
Initialize tracing

jazzdan:
The goal of this PR is to:

1) Init the tracer in both the cli (thing that reads the tiltfile) and the daemon correctly
2) vendor in all of the opentracing stuff


internal/tracer/tracer.go
```diff
@@ -0,0 +1,26 @@
...
+			Type:  "const",
+			Param: 1,
+		},
+		Reporter: &config.ReporterConfig{
+			LogSpans: true,
+		},
+	}
+	tracer, _, err := cfg.New("tilt", config.Logger(jaeger.StdLogger))
+	if err != nil {
+		log.Printf("ERROR: cannot init Jaeger: %v\n", err)
```
nicks: i'm pretty nervous about having an init() function that can fail...can we make this setup more explicit, so that the cli has to call it?

jazzdan: Sure! Good thought

jazzdan: Actually, in real life we won't be using a local jaeger/opentracing server. And we'll need the init to be tolerant of failure: we don't want tilt to fail if it can't transmit a span. So I think in the long term the `init()` approach with logging is what we want.

jazzdan: And, looking at the code south of `config.New` this can't really fail dynamically. The error is returned if the sampling config or the logging config is invalid. It's not based off of I/O or anything.

jazzdan: ... but on third thought, there's no harm in doing it the way you describe, so will do.

PR #105, opened 2018-08-20
build/bench: multiple commands as single step

maiamcc:
Hello @jazzdan, @nicks,

Please review the following commits I made in branch maiamcc/bench-ten-steps-in-one:

4c010ce53cc03b3dffe3470fa68369602cf59e37 (2018-08-20 15:16:43 -0400)
build/bench: multiple commands as single step


maiamcc:
want to know how much (if any) latency comes from the overhead of executing multiple steps -- if there's anything to be gained in executing steps as a single command. but if you don't think it's worth it i can skip it.

PR #106, opened 2018-08-20
build/logs: verbose build logs so can track progress

maiamcc:
Don't feel strongly about this PR, but helps me keep track of what's going on when I'm running `tilt up`


maiamcc:
Good call! @yuindustries do you have any objections to this diff? (Would you prefer these be `debug` logs, for instance?)

yuindustries:
This is great, @maiamcc — and I think it's great that it's just general output rather than debug. I'd love to have this merged so I can keep building on this.

PR #114, opened 2018-08-20
build: use the ImageBuild API instead of exec

nicks:
Hello @maiamcc, @jazzdan,

cf49aad01208f2ba1cef1b63877865c073d6a2f3 (2018-08-20 17:34:15 -0400)
build: use the ImageBuild API instead of exec

97c0891bdd41c38c2929579d077b2e31e5c7fbf4 (2018-08-20 17:30:14 -0400)
build: remove dead startContainer code


nicks:
right now this copies and pastes a lot of the tarball code, which didn't really expose the right API for what i needed to do. there were already TODOs in the code to create some better tarball builder api

internal/build/docker.go
```diff
@@ -78,29 +76,27 @@ func (l *localDockerBuilder) BuildDockerFromExisting(ctx context.Context, existi
...
+		stepsDf := DockerfileFromExisting(resultDigest)
+		for _, step := range steps {
+			stepsDf = stepsDf.Run(step)
+		}
 
-	logger.Get(ctx).Verbose("-- executing steps on base image")
-	for i, s := range steps {
-		logger.Get(ctx).Verbose("--- step %d/%d: %v", i+1, len(steps), s.Argv)
-		err := l.execInContainer(ctx, cID, s)
+		stepsDf = stepsDf.Entrypoint(entrypoint)
```
maiamcc: I think this will kick our asses if entrypoint is not nil, no? (if we're building from scratch, empty entrypoint should throw an error -- and i believe this will put some nonsense in the dockerfile, or else make the image un-runnable? and if we're building from an existing image, we expect to pass the `buildDocker` helper func an empty entrypoint and use the one from the existing image -- does this break that behavior?)

internal/build/docker.go
```diff
@@ -176,10 +172,46 @@ func (l *localDockerBuilder) PushDocker(ctx context.Context, ref reference.Named
...
+func (l *localDockerBuilder) buildWithDfOnly(ctx context.Context, df Dockerfile) (digest.Digest, error) {
+	span, ctx := opentracing.StartSpanFromContext(ctx, "daemon-buildWithDfOnly")
+	defer span.Finish()
+
+	archive, err := TarWithDfOnly(ctx, df)
+	if err != nil {
+		return "", err
+	}
+
+	imageBuildResponse, err := l.dcli.ImageBuild(
```
dbentley: An important, currently-unplanned task that I think is necessary for Replicated is streaming the output of the build steps.

Does this change preclude that? Does this API allow streaming the output of the running commands, as they run?

nicks: yep, that shouldn't change

nicks:
fixed the entrypoint bit! nice catch!

PR #115, opened 2018-08-20
engine: print stdout from docker

jazzdan:


internal/build/docker.go
```diff
@@ -375,6 +374,9 @@ func readDockerOutput(reader io.Reader) (*json.RawMessage, error) {
 			return nil, fmt.Errorf("decoding docker output: %v", err)
 		}
 
+		// TODO(Han): make me smarter! 🤓
+		stdout.Write([]byte(fmt.Sprintf("%+v\n", message)))
```
jazzdan: We pass the `io.Writer` all the way through, just to get it here and print out this ugly message. @yuindustries is going to make the output a lot better in the future, but for now we can see what's going on.

nicks:
hmmm...i thought the goal of @landism 's changes were to move away from having io.Writers and towards using context-attached loggers for output?

internal/build/docker.go
```diff
@@ -391,7 +396,7 @@ func readDockerOutput(reader io.Reader) (*json.RawMessage, error) {
 }
 
 func getDigestFromBuildOutput(reader io.Reader) (digest.Digest, error) {
-	aux, err := readDockerOutput(reader)
+	aux, err := readDockerOutput(reader, os.Stdout)
```
nicks: hmmm...this seems wrong?

jazzdan:
Yeah closing in favor of https://github.com/windmilleng/tilt/pull/117

PR #116, opened 2018-08-20
timing: a few simple tilt up cases

maiamcc:
Hello @jazzdan, @nicks,

the beginnings of a timing script that should hopefully be extensible n stuff


scripts/timing.py
```diff
@@ -0,0 +1,140 @@
...
+import string
+from subprocess import call
+
+RESULTS_BLOCKLTR = '''  ____                 _ _
+ |  _ \ ___  ___ _   _| | |_ ___ _
+ | |_) / _ \/ __| | | | | __/ __(_)
+ |  _ <  __/\__ \ |_| | | |_\__ \_
+ |_| \_\___||___/\__,_|_|\__|___(_)'''
+
+BLORGLY_BACKEND_DIR = os.path.join(os.environ['GOPATH'], 'src/github.com/windmilleng/blorgly-backend')
```
jazzdan: Should we be using blorg-backend for our benchmarks since @yuindustries discovered that blorgly-backend is kind of borked?

scripts/timing.py
```diff
@@ -0,0 +1,140 @@
...
+from collections import namedtuple
+import datetime
+import functools
+import os
+import random
+import string
+from subprocess import call
+
+RESULTS_BLOCKLTR = '''  ____                 _ _
+ |  _ \ ___  ___ _   _| | |_ ___ _
```
nicks: :tophat:

scripts/timing.py
```diff
@@ -0,0 +1,140 @@
...
+
+Case = namedtuple('Case', ['name', 'setup', 'test'])
+Result = namedtuple('Result', ['name', 'time_seconds'])
+
+tilt_up_called = False
+tilt_up_cmd = ["tilt", "up", SERVICE_NAME, '-d']
+
+
+def main():
+    print 'NOTE: this script doesn\'t install `tilt` for you, and relies on you having the ' \
```
nicks: wait are we on python3 yet

scripts/timing.py
nicks: not for this pr, but i wonder if this will eventually become a more OO class where the current backend dir and service names are fields of the class

yuindustries: Blorgly-Backend is not actually hooked up to the frontend. The original idea was for it to provide random strings, but it neither generates those strings nor seems to have ever been successfully tested communicating with the frontend when everyone's deployed to K8s. (It was tested on local, though, and I assume it works there.)

maiamcc: hmm so i feel like as long as it deploys (which it does), it doesn't matter if it works -- but also i don't have strong feelings and am happy to defer to others here

yuindustries: I think it's fine to use blorg-backend for now, since for milestone purposes I'd want anything related to output to be merged in sooner rather than later. And I agree with your logic. 

I myself ended up using blorg-backend for the Docker for Mac test (PITA though it is, with its Cockroach DB setup and such) because I couldn't figure out another way of verifying that "everything works."

README.md
```diff
@@ -7,11 +7,16 @@
 - `make`
 - **[go 1.10](https://golang.org/dl/)**
 - **protobuf 3.2**: `brew install protobuf` or install `protoc-3.2.0-[your_OS]` [from Github](https://github.com/google/protobuf/releases?after=v3.2.1)
+- Our Python scripts are in Python 3.6.0. To run them:
+  - **[pyenv](https://github.com/pyenv/pyenv#installation)**
+  - **python**: `pyenv install`
+  - if you're using GKE and get the error: "pyenv: python2: command not found", run:
```
maiamcc: fyi @nicks @jazzdan this is a potentially awkward result of having our scripts be in py3. I've got stuff working fine now, but if we think this is too much of a PITA for ppl trying to run the scripts, we can easily change them back to py2. thoughts?

PR #117, opened 2018-08-20
build: print output from docker to GRPC stream

jazzdan:


internal/build/docker.go
```diff
@@ -375,6 +375,9 @@ func readDockerOutput(reader io.Reader) (*json.RawMessage, error) {
 			return nil, fmt.Errorf("decoding docker output: %v", err)
 		}
 
+		// TODO(Han): make me smarter! 🤓
+		logger.Get(ctx).Info(fmt.Sprintf("%+v\n", message))
```
nicks: wonder if we need Infof and friends

internal/build/docker_test.go
```diff
@@ -39,7 +39,7 @@ func TestDigestFromSingleStepOutput(t *testing.T) {
 `
 
 	expected := digest.Digest("sha256:11cd0b38bc3ceb958ffb2f9bd70be3fb317ce7d255c8a4c3f4af30e298aa1aab")
-	actual, err := getDigestFromBuildOutput(bytes.NewBuffer([]byte(input)))
+	actual, err := getDigestFromBuildOutput(context.Background(), bytes.NewBuffer([]byte(input)))
```
nicks: i think there might be a helper method to create contexts with loggers for testing? if not, can we add one?

internal/build/docker.go
landism: whoops, maybe I just named it poorly
`func (l logger) Info(format string, a ...interface{})`

internal/build/docker_test.go
jazzdan: Oooh thanks for pointing that out!

PR #120, opened 2018-08-21
tilt: ignore .git, honor .gitignore

landism:
also log which files changed and caused a build to be kicked off

I'm on the fence on whether it's worth injecting a MatcherMaker to Upper to allow testing the matcher there. It seems like a fair amount of mess just to verify that it's calling it and using the bool return value (half the changes to that file are in service of the logging change). If someone asks for it, I'll add it.


internal/git/gitignore_test.go
```diff
@@ -0,0 +1,130 @@
...
+	"github.com/windmilleng/tilt/internal/testutils"
+)
+
+func TestGitIgnoreTester_Simple(t *testing.T) {
+	tf := newTestFixture(t, ".*.swp")
+	defer tf.TearDown()
+
+	g, err := NewGitIgnoreTester(testutils.CtxForTest(), tf.repoRoots[0].Path())
+	if err != nil {
+		t.Error(err)
```
maiamcc: can do the same in one line with `t.Fatal(err)`

internal/git/gitignore_test.go
```diff
@@ -0,0 +1,130 @@
...
+		return
+	}
+
+	ret, err := g.IsIgnored(tf.repoRoots[0].JoinPath("a", "b", ".foo.swp"), false)
+	assert.NoError(t, err)
```
maiamcc: not a big deal but i prefer:
```
if assert.NoError(t, err) {
    assert.True(t, ret)
}
```

b/c I only care about the result of the second assert if there's no error -- reporting the second when there IS an error is imo extra noise.

internal/engine/upper.go
```diff
@@ -135,17 +147,32 @@ func (u Upper) Up(ctx context.Context, services []model.Service, watchMounts boo
...
+					if !isIgnored {
+						changedPaths = append(changedPaths, path)
+					}
+				}
+				if len(changedPaths) > 0 {
+					var changedPathsToPrint []string
+					if len(changedPaths) > maxChangedFilesToPrint {
+						changedPathsToPrint = append(changedPaths[:maxChangedFilesToPrint], "...")
+					} else {
+						changedPathsToPrint = changedPaths
```
maiamcc: can't you just declare in ln 166 as `changedPathsToPrint := changedPaths`? then you can get rid of this extra `else`

internal/git/gitignore.go
```diff
@@ -0,0 +1,134 @@
...
+	p := path.Join(absRoot, ".gitignore")
+	i, err := gitignore.NewGitIgnore(p)
+	if err != nil {
+		_, err = os.Open(path.Join(absRoot, ".gitignore"))
+
+		pathError, ok := err.(*os.PathError)
+		//if the error is that file isn't there (ENOENT), then we don't need a warning, since that's a normal case
+		//if it's any other error, log a warning and pretend the file doesn't exist (matching git's behavior)
+		if ok && pathError.Err != syscall.ENOENT {
+			logger.Get(ctx).Info("warning: failed to open %V: %V", p, err)
```
maiamcc: these should be lowercase v i think

internal/git/gitignore.go
```diff
@@ -0,0 +1,134 @@
...
+// ignores files specified in .gitignore
+type gitIgnoreTester struct {
+	repoRoot      string
+	ignoreMatcher gitignore.IgnoreMatcher
+}
+
+var _ IgnoreTester = gitIgnoreTester{}
+
+func (i gitIgnoreTester) IsIgnored(f string, isDir bool) (bool, error) {
+	_, isChild := ospath.Child(i.repoRoot, f)
```
maiamcc: do we care about resolving symlinks here? o_0

internal/git/gitignore.go
```diff
@@ -0,0 +1,134 @@
...
+	for _, repoRoot := range repoRoots {
+		t, err := NewRepoIgnoreTester(ctx, repoRoot)
+		if err != nil {
+			return nil, err
+		}
+
+		testers = append(testers, t)
+	}
+
+	return compositeIgnoreTester{testers}, nil
```
maiamcc: fwiw i've heard it said as a good practice that you should always name the fields in your struct literals (e.g. `compositeIgnoreTester{testers: testers}`) so that your code doesn't break if, e.g. the fields on the struct get reordered.

internal/testutils/temp_dir_fixture.go
```diff
@@ -41,8 +41,10 @@ func (f *TempDirFixture) Path() string {
 	return f.dir.Path()
 }
 
-func (f *TempDirFixture) JoinPath(path string) string {
-	return filepath.Join(f.Path(), path)
+func (f *TempDirFixture) JoinPath(path ...string) string {
```
maiamcc: nice!

internal/git/gitignore_test.go
```diff
@@ -0,0 +1,130 @@
...
+type testFixture struct {
+	repoRoots []*testutils.TempDirFixture
+}
+
+// initializes `tf.repoRoots` to be an array with one dir per gitignore
+func newTestFixture(t *testing.T, gitignores ...string) *testFixture {
+	tf := testFixture{}
+	for _, gitignore := range gitignores {
+		tempDir := testutils.NewTempDirFixture(t)
+		f, err := os.Create(tempDir.JoinPath(".gitignore"))
```
maiamcc: ...can't you just do `tempDir.WriteFile(".gitignore", gitignore)`?

landism: yep, thanks

internal/git/gitignore.go
landism: That seems like a two-edged sword. The flipside is sometimes I *want* my code to break so that I can ensure I'm initializing the new field with an appropriate value. I don't have a good feel for whether that happens often enough to be worth worrying about.

internal/git/gitignore_test.go
landism: yep, thanks

internal/engine/upper.go
landism: Yep, that'd work.

It's not a big deal either way, but I generally prefer not changing variables' values / reusing variables when reasonable. This way the variable is only set once, and keeps its value. If I see code that declares a variable and then modifies it on the next line, I'll worry about whether they're changing it later in the function and when I get to the code that actually uses it, worry about whether it got mutated on the way.

Of course, even as written in this PR, it could get modified later, and the block is so short it doesn't really matter in this case.

internal/git/gitignore.go
landism: Good question. I'm gonna TODO that for now.

PR #121, opened 2018-08-21
tracing: more tracing

jazzdan:
note: for some reason cross grpc tracing isn't working. Looking in to that now but I may be at a loss.


jazzdan:
Here's what a sample trace looks like now on my computer:

screen shot 2018-08-21 at 11 30 30 am

PR #122, opened 2018-08-21
build/tests: in most cases docker tests should probs be overwriting files instead of appending

maiamcc:


nicks:
can you explain your thinking of why overwriting is better than appending?

i would expect appending to be better, because it preserves more information about what happened, which is better for testing. We used to have a big problem with double-executing commands by accident, and this made it easy to catch those problems.

maiamcc:
oh shoot -- @nicks you're totally right, i forgot that we're now checking the exact contents of files. in my head we were on the old strategy of `cat`ing the file and checking `strings.Contains(output, contents)`, which could return false positives if files could contain both the old contents and new contents strings. but we do things smarter now sooooo never mind 😅

PR #128, opened 2018-08-21
tilt: Upper supports watch with multiple services

pmt:


internal/engine/upper.go
```diff
@@ -52,138 +51,53 @@ func NewUpper(ctx context.Context, manager service.Manager, env k8s.Env) (Upper,
...
+				if len(event.files) > maxChangedFilesToPrint {
+					changedPathsToPrint = append(event.files[:maxChangedFilesToPrint], "...")
+				} else {
+					changedPathsToPrint = event.files
 				}
+
+				logger.Get(ctx).Infof("files changed. rebuilding %v. observed changes: %v", event.service.Name, changedPathsToPrint)
+
+				var err error
+				buildTokens[event.service.Name], err = u.b.BuildAndDeploy(
```
nicks: if there's a build error, you probably don't want to overwrite the old token?

internal/engine/up_watch.go
```diff
@@ -0,0 +1,254 @@
...
+	watchEventsStream, err := makeWatchEventsStream(ctx, watcherMaker, timerMaker, services)
+	if err != nil {
+		return nil, err
+	}
+
+	serviceChan := make(chan serviceFilesChangedEvent)
+	errs := make(chan error)
+	go func() {
+		// a map of service names to a list of any unprocessed changes
+		leftover := make(map[string]*serviceFilesChangedEvent)
```
nicks: talked about this offline, but i think this layer is unnecessary, and can be mostly deleted without changing behavior

internal/engine/up_watch.go
```diff
@@ -0,0 +1,254 @@
...
+				case err, ok := <-watcher.Errors():
+					if !ok {
+						close(events)
+						close(errs)
+						return
+					}
+					errs <- err
+				case fsEvents, ok := <-coalescedEvents:
+					if !ok {
+						close(events)
```
nicks: this is going to cause panics, because you have mutliple goroutines and they're all trying to close the same channel.

take a look at https://blog.golang.org/pipelines, particularly the `merge` function, which shows how to merge multiple channels into one channel

internal/engine/up_watch.go
```diff
@@ -0,0 +1,254 @@
...
+						close(errs)
+						return
+					}
+					var watchEvents []serviceSingleFileChangeEvent
```
nicks: i think this code will be simpler if you separate out the filtering from the multiplexing. merging multiple channels is kind of tricky. filtering/mapping is easy, because you can use a range on the input channel

for event := range myChannel {
  ...
  newChannel <- newEvent
}

internal/engine/upper.go
landism: yeah, that was unintentional. thanks.

(though theoretically even a failed build could generate artifacts that would speed up successive builds and maybe the contract should be we always returns a token, even if it's just the one that was passed in?)

internal/engine/up_watch.go
landism: per offline discussion, this starts getting uglier since we need to handle errors, and we can figure it out later

internal/engine/up_watch.go
landism: commented out and TODO'd

PR #129, opened 2018-08-21
engine: don't push to a central registry when using docker-for-mac

nicks:
Hello @jazzdan, @maiamcc, @yuindustries,

Please review the following commits I made in branch nicks/kube:

35fe94ac73de9915d0768ac13443455500f363a7 (2018-08-21 14:30:38 -0400)
engine: don't push to a central registry when using docker-for-mac

### DIFF
--- a/internal/build/docker.go
+++ b/internal/build/docker.go
@@ -26,7 +26,7 @@ import (
 )
 
 // Tilt always pushes to the same tag. We never push to latest.
-const pushTag = "wm-tilt"
+const tiltTag = "wm-tilt"
 
 var ErrEntrypointInDockerfile = errors.New("base Dockerfile contains an ENTRYPOINT, " +
 	"which is not currently supported -- provide an entrypoint in your Tiltfile")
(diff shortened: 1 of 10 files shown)


nicks:
deeply unhappy with the way this pr ended up, which seems to be mostly work-arounds for weird holes in how kubernetes and docker interop *sigh*

working on a better test for this

internal/build/docker.go
```diff
@@ -102,6 +103,21 @@ func (l *localDockerBuilder) buildDocker(ctx context.Context, df Dockerfile,
...
 }
 
+// Tag the digest with the given name and wm-tilt tag.
+func (l *localDockerBuilder) TagDocker(ctx context.Context, ref reference.Named, dig digest.Digest) (reference.NamedTagged, error) {
+	tag, err := reference.WithTag(ref, tiltTag)
+	if err != nil {
+		return nil, fmt.Errorf("TagDocker: %v", err)
+	}
+
+	err = l.dcli.ImageTag(ctx, dig.String(), tag.String())
```
jazzdan: should we put some spans around this?

internal/k8s/image.go
```diff
@@ -51,7 +30,21 @@ func InjectImagePullPolicy(entity K8sEntity, policy v1.PullPolicy) (K8sEntity, e
 //   to ensure that k8s fails hard if the image is missing from docker.
 //
 // Returns: the new entity, whether the image was replaced, and an error.
-func InjectImageDigest(entity K8sEntity, canonicalRef reference.Canonical, policy v1.PullPolicy) (K8sEntity, bool, error) {
+func InjectImageDigest(entity K8sEntity, injectRef reference.Named, policy v1.PullPolicy) (K8sEntity, bool, error) {
+	// NOTE(nick): For some reason, if you have a reference with a digest,
```
jazzdan: ughhhhhh good find

nicks:
(the test will probably be a separate pr)

internal/build/docker.go
nicks: i think a good rule of thumb is to only add spans where we have good reason to think we need them; the bigger risk is wasting a lot of engineering time creating spans we never use

PR #131, opened 2018-08-21
timing: raise an exception if tilt fails for some reason

nicks:
Hello @maiamcc,

Please review the following commits I made in branch nicks/timing:

a47f8e6609955e9f205f8010ff6e23ab8786c0ff (2018-08-21 15:44:08 -0400)
timing: raise an exception if tilt fails for some reason

### DIFF
--- a/scripts/timing.py
+++ b/scripts/timing.py
@@ -76,7 +76,7 @@ def make_case_tilt_up_again_no_change():
             print('Initial `tilt up` already called, no setup required')
             return
         print('Initial call to `tilt up`')
-        call(tilt_up_cmd)
+        call_or_error(tilt_up_cmd)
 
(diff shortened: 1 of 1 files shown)


scripts/timing.py
```diff
@@ -103,11 +103,20 @@ def time_call(cmd):
...
 
     return (end - start).total_seconds()
 
+def call_or_error(cmd):
+    """
+        Call the given command (a list of strings representing command and args),
+        raising an error if it fails.
+    """
+    return_code = call(cmd)
+    if return_code != 0:
```
maiamcc: yes thank you!

```
'Command %s exited with exit code %d'
```
?

or actually, given that this is python 3 now:
```
'Command %s exited with exit code %d'.format(cmd, return_code)
```

nicks: done!

PR #134, opened 2018-08-21
build: do the whole build in a single ImageBuild call. this should be simpler and perform better

nicks:
Hello @maiamcc, @jazzdan,

Please review the following commits I made in branch nicks/onefile:

45fed0cb0e76a550bdf08089f9c7686747993ea7 (2018-08-21 16:53:25 -0400)
build: do the whole build in a single ImageBuild call. this should be simpler and perform better

### DIFF
--- a/internal/build/docker.go
+++ b/internal/build/docker.go
@@ -61,6 +61,11 @@ func (l *localDockerBuilder) BuildDockerFromScratch(ctx context.Context, baseDoc
 	span, ctx := opentracing.StartSpanFromContext(ctx, "daemon-BuildDockerFromScratch")
 	defer span.Finish()
 
+	err := baseDockerfile.ForbidEntrypoint()
+	if err != nil {
+		return "", err
+	}
+
 	return l.buildDocker(ctx, baseDockerfile, MountsToPath(mounts), steps, entrypoint)
 }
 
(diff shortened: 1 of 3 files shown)


internal/build/docker.go
```diff
@@ -75,29 +80,29 @@ func (l *localDockerBuilder) BuildDockerFromExisting(ctx context.Context, existi
...
 }
 
-func (l *localDockerBuilder) buildDocker(ctx context.Context, df Dockerfile,
+func (l *localDockerBuilder) buildDocker(ctx context.Context, baseDockerfile Dockerfile,
 	paths []pathMapping, steps []model.Cmd, entrypoint model.Cmd) (digest.Digest, error) {
-	// Do a docker build with the new files.
-	resultDigest, err := l.buildFromDfWithFiles(ctx, df, paths)
+
+	df := baseDockerfile.AddAll()
+	toRemove, err := missingLocalPaths(paths)
```
jazzdan: Nice rename!

internal/build/docker.go
```diff
@@ -378,28 +261,19 @@ func tarContextAndUpdateDf(ctx context.Context, df Dockerfile, paths []pathMappi
 	}()
 
 	// TODO: maybe write our own tarWriter struct with methods on it, so it's clearer that we're modifying the tar writer in place
-	dnePaths, err := archiveIfExists(ctx, tw, paths)
+	err := archiveIfExists(ctx, tw, paths)
```
maiamcc: since this is no longer responsible for returning missing paths, can rename if you want

maiamcc: `archivePaths`? `makeContextArchive`? `tarContext`?

internal/build/tar.go
```diff
@@ -31,24 +31,18 @@ func archiveDf(ctx context.Context, tw *tar.Writer, df Dockerfile) error {
 	return nil
 }
 
-// archiveIfExists creates a tar archive of all local files in `paths` _if they exist_,
-// and returns a list of all PathMappings that point to nonexistent LocalPaths (we'll
-// use that later to add RM statements).
+// archiveIfExists creates a tar archive of all local files in `paths` _if they exist_.
```
maiamcc: likewise this comment is no longer really relevant.

internal/build/tar.go
```diff
@@ -109,6 +103,10 @@ func tarPath(ctx context.Context, tarWriter *tar.Writer, source, dest string) (d
 		if header.Typeflag == tar.TypeReg {
 			file, err := os.Open(path)
 			if err != nil {
+				// In case the file has been deleted since we last looked at it.
+				if os.IsNotExist(err) {
```
maiamcc: iiiiii am not SURE this is the right behavior? because we don't delete it in the container, we just don't touch it...

internal/build/tar.go
```diff
@@ -31,24 +31,18 @@ func archiveDf(ctx context.Context, tw *tar.Writer, df Dockerfile) error {
...
-// use that later to add RM statements).
+// archiveIfExists creates a tar archive of all local files in `paths` _if they exist_.
 // NOTE: modifies tw in place.
-func archiveIfExists(ctx context.Context, tw *tar.Writer, paths []pathMapping) (dnePaths []pathMapping, err error) {
+func archiveIfExists(ctx context.Context, tw *tar.Writer, paths []pathMapping) error {
 	span, ctx := opentracing.StartSpanFromContext(ctx, "daemon-archiveIfExists")
 	defer span.Finish()
 	for _, p := range paths {
-		dne, err := tarPath(ctx, tw, p.LocalPath, p.ContainerPath)
+		_, err := tarPath(ctx, tw, p.LocalPath, p.ContainerPath)
```
maiamcc: update docstring and return values for tarPath if we're not using it to send back nonexistent paths?

internal/build/docker.go
nicks: changed to archivePaths

internal/build/tar.go
nicks: updated

internal/build/tar.go
nicks: updated

internal/build/tar.go
nicks: i think you're confused, this is reading the file outside the container

PR #137, opened 2018-08-22
timing: benchmark builds from --watch

maiamcc:
note that process.terminate() will leave an orphan daemon running until we put better signal handling somewhere

### DIFF
--- a/scripts/timing.py
+++ b/scripts/timing.py
@@ -5,9 +5,9 @@ import datetime
 import functools
 import os
 import random
+import signal
 import string
 import subprocess
-import time
 from typing import List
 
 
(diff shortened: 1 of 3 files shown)


scripts/timing.py
```diff
@@ -20,47 +22,119 @@
...
           'blorgly-backend project in your $GOPATH (`github.com/windmilleng/blorgly-backend`)')
     print()
 
     os.chdir(BLORGLY_BACKEND_DIR)
 
     cases = [
-        make_case_tilt_up_once(),
-        make_case_tilt_up_again_no_change(),
-        make_case_tilt_up_again_new_file(),
+        # make_case_tilt_up_once(),
```
jazzdan: 👀

scripts/timing.py
```diff
@@ -69,7 +143,7 @@ def set_tilt_up_called():
                 functools.partial(time_call, tilt_up_cmd))
 
 
-def make_case_tilt_up_again_no_change():
+def make_case_tilt_up_again_no_change() -> Case:
```
jazzdan: types! feels good

scripts/timing.py
```diff
@@ -40,27 +53,88 @@ def main():
...
-            c.setup()
-            timetake = c.test()
+            print('~~ RUNNING CASE: {}'.format(c.name))
+            args = []
+            kwargs = {}
+
+            print('~~~~ setup: {}'.format(c.name))
+            ret = c.setup()
+            if ret is not None:
+                args = ret[0]
```
nicks: this API seems...weird.

specifically, the setup() and test() methods are tightly coupled, and using them correctly requires a lot of inside knowledge about how they're implemented.

possible alternatives:
1) is there a reason to have separate setup and test methods at all? i'm not sure that the separation has any upsides
2) make case a real class that keeps internal state

wdyt?

nicks: ya, not blocking or anything

PR #141, opened 2018-08-22
cli: error handling when the daemon exists unexpectedly. also fixes the case where you're talking to an old daemon

nicks:
Hello @pmt, @landism,

Please review the following commits I made in branch nicks/errors2:

f42b533e363e069fc7831e9b9b858b5c89022f0f (2018-08-22 13:13:45 -0400)
cli: error handling when the daemon exists unexpectedly. also fixes the case where you're talking to an old daemon

### DIFF
--- a/internal/cli/up.go
+++ b/internal/cli/up.go
@@ -1,9 +1,13 @@
 package cli
 
 import (
+	"bytes"
 	"context"
 	"errors"
+	"fmt"
 	"log"
+	"strings"
+	"sync"
 
 	opentracing "github.com/opentracing/opentracing-go"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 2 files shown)


nicks:
updated this a bit. the main way i'm testing this is to have the daemon panic, and make sure the parent process captures the panic() output

nicks:
hmmm...for some reason github didn't show me maia's message until just now. i don't care that much either way.

nicks:
as discussed off-list, this doesn't kill any new processes, or change anything about how/why processes are killed. it only changes what the processes print when they exit.

nicks:
(i'm fine with moving the daemon in-process, but gonna merge this for now because it unblocks the bugs i'm trying to diagnose)

PR #142, opened 2018-08-22
image: hash inputs to image

jazzdan:
only return image from history if service
 hash matches

~I'm not happy with how many arguments we're passing to things, and I'm not convinced this is the right API. I welcome all feedback on how to make this a bit more clean.~ -I feel better now, thanks Nick!


jazzdan:
@nicks that's a good point, and yeah I looked at it. We don't want to invalidate the cache on everything that a `Service` has, though I suppose we could pass that around and just grab the things we need. 

Unless anyone has other ideas I'll do that when I get back to this.

internal/image/history.go
```diff
@@ -89,25 +93,91 @@ func (h ImageHistory) load(ctx context.Context, name reference.Named, digest dig
...
+	Mounts         []model.Mount
+	Steps          []model.Cmd
+	Entrypoint     model.Cmd
 }
 
-func (h ImageHistory) AddAndPersist(ctx context.Context, name reference.Named, digest digest.Digest, checkpoint CheckpointID) error {
-	key, entry := h.load(ctx, name, digest, checkpoint)
+type HashedInputs = uint64
+
+func hashInputs(service model.Service) (HashedInputs, error) {
```
jazzdan: Will change this function name to `hashService`

internal/image/history.go
```diff
@@ -89,25 +93,91 @@ func (h ImageHistory) load(ctx context.Context, name reference.Named, digest dig
...
+	Entrypoint     model.Cmd
 }
 
-func (h ImageHistory) AddAndPersist(ctx context.Context, name reference.Named, digest digest.Digest, checkpoint CheckpointID) error {
-	key, entry := h.load(ctx, name, digest, checkpoint)
+type HashedService = uint64
+
+func hashService(service model.Service) (HashedService, error) {
+	hi := hash{
+		BaseDockerfile: build.Dockerfile(service.DockerfileText),
```
nicks: this is a dangerous approach. if someone adds new fields to Service, they won't know to update them here.

can you either:
1) hash the existing Service structure, but void out the fields you don't care about?
2) create some sort of hash method on Service itself?

internal/image/history.go
```diff
@@ -89,25 +93,91 @@ func (h ImageHistory) load(ctx context.Context, name reference.Named, digest dig
 		h.byName[key] = bucket
 	}
 
+	bucket.entries = append(bucket.entries, entry)
+	if entry.After(bucket.mostRecent.CheckpointID) {
+		bucket.mostRecent = entry
+	}
+}
+
+func (h ImageHistory) load(
```
maiamcc: is this replicating logic from `loadFromEntry`? maybe this should just assemble all the inputs into a `historyEntry` and call `loadFromEntry` on it?

internal/image/history.go
maiamcc: ooohhh hash method on the service++

jazzdan: yeah that's a great point @nicks, I like that idea.

internal/model/service.go
```diff
@@ -4,20 +4,40 @@ import (
...
+	"github.com/mitchellh/hashstructure"
 )
 
 type ServiceName string
 
 func (s ServiceName) String() string { return string(s) }
 
 type Service struct {
-	K8sYaml        string
+	K8sYaml        string `hash:"ignore"`
```
jazzdan: Check this out @nicks. I didn't realize this was a feature of the library!

internal/image/history.go
```diff
@@ -89,25 +117,45 @@ func (h ImageHistory) load(ctx context.Context, name reference.Named, digest dig
 		h.byName[key] = bucket
 	}
 
+	hi, err := service.Hash()
```
maiamcc: 👋

internal/image/history.go
```diff
@@ -117,7 +165,16 @@ func (h ImageHistory) MostRecent(name reference.Named) (digest.Digest, Checkpoin
 		return "", CheckpointID{}, false
 	}
 
+	hi, err := service.Hash()
```
maiamcc: 👋

internal/model/service.go
```diff
@@ -4,20 +4,40 @@ import (
...
 	DockerfileText string
 	Mounts         []Mount
 	Steps          []Cmd
 	Entrypoint     Cmd
```
maiamcc: hmm I'm not conviiiinced we need to enforce ALL of these being constant to use an image for the same service... off the top of my head i knew we def want dockerfile, we might want entrypoint for incremental build, and idk if we need mounts or steps...

internal/model/service.go
```diff
@@ -4,20 +4,40 @@ import (
...
-	DockerfileTag  string
-	Name           ServiceName
+	DockerfileTag  string      `hash:"ignore"`
+	Name           ServiceName `hash:"ignore"`
+}
+
+type HashedService = uint64
+
+type hash struct {
```
maiamcc: is this still being used?

internal/image/history.go
```diff
@@ -101,8 +99,8 @@ func (h ImageHistory) loadFromEntry(ctx context.Context, name reference.Named, e
 	}
 }
 
-// load takes checkpoint and a service definition and loads it in to the appropriate bucket in memory.
-func (h ImageHistory) load(
+// addInMemory takes checkpoint and a service definition and loads it in to the appropriate bucket in memory.
+func (h ImageHistory) addInMemory(
```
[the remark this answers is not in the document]
jazzdan: oh yes thanks for the reminder

internal/model/service.go
jazzdan: I'll adjust this in the next step where we actually make use of this code.

internal/model/service.go
nicks: :raised_hands:

PR #147, opened 2018-08-22
tilt: git_repo becomes local_git_repo per README

pmt:
see: https://github.com/windmilleng/tilt/pull/146/files

q: should I update the blorg and blorgly tiltfiles with these new names?


jazzdan:
FYI I'll take care of https://github.com/windmilleng/blorg-tilt/pull/1

PR #149, opened 2018-08-22
daemon: call daemon in-process instead of as child proc

maiamcc:
tried to strike a balance between speed, reasonableness of code, and ability to slot daemon back in as its own process if we decide to. i... don't know if i succeeded. feedback welcome.

in particular, @landism, objections to me having `GetServiceConfig` return `[]model.Service` rather than `[]*proto.Service`?


internal/cli/up.go
```diff
@@ -34,39 +39,43 @@ func (c *upCmd) register() *cobra.Command {
...
-	tf, err := tiltfile.Load("Tiltfile")
+	env, err := k8s.DetectEnv()
 	if err != nil {
-		return err
+		return fmt.Errorf("failed to detect kubernetes: %v", err)
 	}
 
-	serviceName := args[0]
-	services, err := tf.GetServiceConfig(serviceName)
+	serviceManager := service.NewMemoryManager()
```
nicks: fyi, this all got replaced by auto-generated DI code

internal/cli/up.go
```diff
@@ -34,39 +39,43 @@ func (c *upCmd) register() *cobra.Command {
...
-	defer func() {
-		err := proc.Kill()
-		if err != nil {
-			log.Fatalf("failed to shut down daemon: %v", err)
-		}
-	}()
-
-	dCli, err := tiltd_client.NewDaemonClient()
+	serviceName := args[0]
+	protoServices, err := tf.GetServiceConfig(serviceName)
```
nicks: +1 for changing GetServiceConfig to return domain objects

internal/cli/up.go
maiamcc: ha yeah i just found those conflicts, fixing now

internal/cli/up.go
maiamcc: will do in another diff, gonna land this asap so i don't get more weird merge shit

PR #150, opened 2018-08-22
demo: more output after Enter keypress

yuindustries:
Also nixes Cobra, since this is super simple.


cmd/tiltdemo/main.go
```diff
@@ -1,9 +1,37 @@
...
+func println(ms int, msg string) {
+	time.Sleep(time.Duration(ms) * time.Millisecond)
+	fmt.Println(msg)
+}
+
+func awaitInput(input byte) {
+	reader := bufio.NewReader(os.Stdin)
+	c, err := reader.ReadByte()
+	if err != nil {
+		fmt.Println(err)
```
dbentley: I'd fatal here.

cmd/tiltdemo/main.go
```diff
@@ -1,9 +1,37 @@
...
+		return
+	}
+
+	if c == input {
```
dbentley: you're going to return anyway, so I think you can just skip this.

PR #152, opened 2018-08-22
build: create a tag based on the digest, instead of using a constant tag

nicks:
Hello @jazzdan, @maiamcc,

Please review the following commits I made in branch nicks/tag:

0c9eae20dc621c9bd413410998299ba96d16148c (2018-08-22 16:49:26 -0400)
build: create a tag based on the digest, instead of using a constant tag

### DIFF
--- a/internal/build/docker.go
+++ b/internal/build/docker.go
@@ -26,9 +26,6 @@ import (
 	"github.com/opencontainers/go-digest"
 )
 
-// Tilt always pushes to the same tag. We never push to latest.
-const tiltTag = "wm-tilt"
-
 var ErrEntrypointInDockerfile = errors.New("base Dockerfile contains an ENTRYPOINT, " +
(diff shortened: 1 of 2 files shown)


internal/build/docker.go
```diff
@@ -366,3 +363,16 @@ func shouldRemoveImage() bool {
...
+	str := string(d)
+	if !strings.HasPrefix(str, "sha256:") {
+		return "", fmt.Errorf("Unrecognized digest type: %s", str)
+	}
+
+	withoutPrefix := str[len("sha256:"):]
+	if len(withoutPrefix) < 16 {
+		return "", fmt.Errorf("Digest too short: %s", str)
+	}
+	return fmt.Sprintf("tilt-%s", withoutPrefix[:16]), nil
```
maiamcc: ...can't you just do d.Encoded()?

nicks: ooh good point! fixed!

PR #153, opened 2018-08-22
demo: realistic output

yuindustries:
output-demo

### DIFF
--- a/cmd/tiltdemo/main.go
+++ b/cmd/tiltdemo/main.go
@@ -8,60 +8,108 @@ import (
 	"time"
 )
 
+// Colors
 var cGreen = "\033[32m"
+var cMagenta = "\033[35m"
 var cBlue = "\033[34m"
 var cReset = "\u001b[0m"
 
+// Snippets
+var linebreak = "\n"
+var stream = "    │ "
+var service = "app-frontend"
+
 func main() {
-	printlnColor(0, "Starting tilt!", cGreen)
-	println(20, "  → in context [docker-for-desktop]")
+	printlnColor(0, "Starting tilt…", cGreen)
+	println(20, "  → current-context [docker-for-desktop]")
+
+	firstPipeline()
 
 	newline()
-	printlnColor(100, "Building from scratch: [gcr.io/project/app-frontend]", cGreen)
-	println(100, "  → tarring context")
-	println(100, "  → sending to Docker daemon…")
-
-	newline()
-	printDockerBuild(3)
-
-	newline()
-	println(100, "  → tagged 2917b4065035")
-	printlnColor(200, "  → Done in 7.234s", cBlue)
-
-	newline()
-	printlnColor(100, "Deploying: [devel-hanyu-lb-blorg-be]", cGreen)
-	println(100, "  → parsing config YAML")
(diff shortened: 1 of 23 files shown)


cmd/tiltdemo/main.go
```diff
@@ -9,26 +9,67 @@ import (
...
 
 	awaitInput()
-
 	println(100, "Building [my-app-backend]")
-	println(500, "Build complete in 0.1232s")
+	printlnColor(200, "Build complete in 1.2342s")
 
 	awaitInput()
-
 	println(100, "Building [my-app-backend2]")
```
dbentley: Do we expect to print the build output for later builds?

cmd/tiltdemo/main.go
```diff
@@ -9,26 +9,67 @@ import (
...
-	println(500, "Build complete in 1.3532s")
+	printlnColor(200, "Build complete in 2.1332s")
```
dbentley: We probably need better terminology, as it's also pushing.

cmd/tiltdemo/main.go
```diff
@@ -9,26 +9,67 @@ import (
...
+	println(100, " → applying via kubectl")
+	println(100, "Successfully built 2917b4065035")
+	println(100, "Successfully tagged 2917b4065035")
+	printlnColor(200, "Build complete in 7.2342s")
 
 	awaitInput()
-
 	println(100, "Building [my-app-backend]")
-	println(500, "Build complete in 0.1232s")
+	printlnColor(200, "Build complete in 1.2342s")
```
dbentley: It'd be great to include how much was their build, vs. the rest.

cmd/tiltdemo/main.go
```diff
@@ -9,26 +9,67 @@ import (
...
+	newline()
+	println(100, "Parsing Kubernetes config YAML")
+	println(100, " → applying via kubectl")
+	println(100, "Successfully built 2917b4065035")
+	println(100, "Successfully tagged 2917b4065035")
+	printlnColor(200, "Build complete in 7.2342s")
 
 	awaitInput()
-
 	println(100, "Building [my-app-backend]")
```
dbentley: We should say what file was edited that's causing us to rebuild.

yuindustries:
tilt-ideal

cmd/tiltdemo/main.go
```diff
@@ -8,27 +8,101 @@ import (
...
+	printlnColor(200, "  → Done in 1.234s", cBlue)
+
+	newline()
+	printlnColor(100, "Deploying: [devel-hanyu-lb-blorg-be]", cGreen)
+	println(100, "  → parsing config YAML")
+	println(100, "  → applying config via kubectl")
+	println(100, "  → service created")
+	println(100, "  → deployment created")
+	printlnColor(200, "  → Done in 472ms", cBlue)
+	newline()
```
dbentley: I think my big remaining question is about terminology and timing for the overall pipeline run.

A user should see how long the entire thing took, and how much of that was their build (which we don't control), as a summary, without having to scroll.

yuindustries:
color code output

yuindustries:
color code output 2

PR #154, opened 2018-08-22
internal/tiltfile: composite_service takes an array

dbentley:


dbentley:
Sorry, realized it was as fast to type this as to assign it.

PR #158, opened 2018-08-23
tilt: use go vet instead of go tool vet, fix go vet errors

landism:
`make lint` was failing in master on my laptop, but passing in CI

When I SSHd into the CI box, I observed `go vet` was failing, as desired, while `go tool vet` was not. Both the CI box and my laptop have go 1.10.3. Maybe `go vet` execs something whose version can vary independently of go's version? Or else this seems like a bug. I haven't come up with an explanation for the difference in behavior, but https://golang.org/cmd/vet/ says:

> For testing and debugging vet can be run directly by invoking "go tool vet" or just running the binary. Run this way, vet might not have up to date information for imported packages.
>
> `go tool vet source/directory/*.go`
> vets the files named, all of which must be in the same package.
>
> `go tool vet source/directory`
> recursively descends the directory, vetting each package it finds.

Which seems better both in terms of simplicity and correctness, so we might as well go with that.


internal/engine/upper_test.go
```diff
@@ -130,7 +130,7 @@ func TestUpper_UpWatchFileChangeThenError(t *testing.T) {
 		assert.Equal(t, service, call.service)
 		assert.Equal(t, []string(nil), call.files)
 		fileRelPath := "fdas"
-		f.watcher.events <- watch.FileEvent{fileRelPath}
+		f.watcher.events <- watch.FileEvent{Path: fileRelPath}
```
jazzdan: 👀

PR #165, opened 2018-08-23
timing: get current k8s env

maiamcc:
not currently super necessary but we'll probs want this when outputting to csv etc.

### DIFF
--- a/scripts/timing.py
+++ b/scripts/timing.py
@@ -158,16 +158,9 @@ def call_or_error(cmd: List[str]):
 
 def get_k8s_env() -> K8sEnv:
     """Get current Kubernetes env. (or throw an exception)."""
-    p = subprocess.Popen(['kubectl', 'config', 'view', '-o', 'template', '--template={{ index . "current-context" }}'], stdout=subprocess.PIPE)
-    output, err = p.communicate()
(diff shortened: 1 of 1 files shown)


scripts/timing.py
```diff
@@ -149,6 +156,26 @@ def call_or_error(cmd: List[str]):
         raise Exception('Command {} exited with exit code {}'.format(cmd, return_code))
 
 
+def get_k8s_env() -> K8sEnv:
+    """Get current Kubernetes env. (or throw an exception)."""
+    p = subprocess.Popen(['kubectl', 'config', 'view', '-o', 'template', '--template={{ index . "current-context" }}'], stdout=subprocess.PIPE)
```
nicks: are you aware of 'kubectl config current-context'?

this seems...overly complicated

nicks: did you consider using subprocess.check_output? this is a very common pattern and it seems weird to reach for these lower-level primitives.

maiamcc: ...i did NOT consider that, no. i was lead astray by a stack overflow article 😳

maiamcc: nnnnno. again, stack overflow strugglebus. 😳

PR #166, opened 2018-08-23
watch: fix more data races on darwin

nicks:
Hello @jazzdan,

Please review the following commits I made in branch nicks/watch:

ed2bc70ef0e1eb02b69a0aebd66e5f3f00689d13 (2018-08-23 15:05:09 -0400)
watch: fix more data races on darwin


nicks:
we have always been at war with fsevents 👀

PR #170, opened 2018-08-23
build: remove previous mounts when building from existing

jazzdan:
`BuildDockerFromExisting` now takes the previous service's mounts, which it gets from the build token.

### DIFF
--- a/internal/build/docker_test.go
+++ b/internal/build/docker_test.go
@@ -378,7 +378,7 @@ func TestSelectiveAddFilesToExisting(t *testing.T) {
 		f.t.Fatal("filesToPathMappings:", err)
 	}
 
-	digest, err := f.b.BuildDockerFromExisting(f.ctx, existing, pms, []model.Cmd{})
+	digest, err := f.b.BuildDockerFromExisting(f.ctx, existing, pms, []model.Cmd{}, []model.Mount{})
 	if err != nil {
 		t.Fatal(err)
 	}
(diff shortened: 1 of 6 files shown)


jazzdan:
Note this makes things slightly slower in the iterative build case. Previously:

```
	tilt up 1x --> 1.11432 seconds
	tilt up again, no change --> 0.60320 seconds
	tilt up again, new file --> 12.66296 seconds
	watch build from changed file --> 5.25783 seconds
```

With this patch:

```
	tilt up 1x --> 0.64434 seconds
	tilt up again, no change --> 0.64482 seconds
	tilt up again, new file --> 11.46676 seconds
	watch build from changed file --> 8.24862 seconds
```

internal/build/dockerfile.go
```diff
@@ -11,10 +11,15 @@ import (
...
-// Docker builds.
-func DockerfileFromExisting(existing digest.Digest) Dockerfile {
-	return Dockerfile(fmt.Sprintf("FROM %s", existing.Encoded()))
+// as its base image with a FROM statement, and removes any previously applied mounts
+// This is necessary for iterative Docker builds.
+func DockerfileFromExisting(existing digest.Digest, prevMounts []model.Mount) Dockerfile {
+	df := strings.Builder{}
+	df.WriteString(fmt.Sprintf("FROM %s\n", existing.Encoded()))
+	for _, m := range prevMounts {
+		df.WriteString(fmt.Sprintf("RUN rm -rf %s\n", m.ContainerPath))
```
maiamcc: `dm.RmPaths(MountsToPath(prevMounts))`
(also wtf why is it called `MountsToPath`? I think that's a typo.)

internal/build/docker_test.go
```diff
@@ -492,6 +492,49 @@ func TestBuildDockerWithStepsFromExistingPreservesEntrypoint(t *testing.T) {
 	f.assertFilesInContainer(f.ctx, cID, expected)
 }
 
+func TestBuildDockerWithStepsFromExistingRemovesPreviousMounts(t *testing.T) {
```
maiamcc: seems like lots of this test case is irrelevant to the thing your testing for. don't need steps or entrypoint, and comment at ln. 512 is now obsolete. also since we're not testing entrypoint you can just use `assertFilesInImage`

nicks:
why does it make things slower in the iterative case? that seems...concerning, because that's the case we care most about?

jazzdan:
not doing this for now

PR #171, opened 2018-08-23
minikube: add a helper function for reading the docker client from minikube

nicks:
Hello @maiamcc,

Please review the following commits I made in branch nicks/minikube2:

d582d9c1e874ccc95932635cb23beef69294399d (2018-08-23 17:58:52 -0400)
minikube: add a helper function for reading the docker client from minikube

### DIFF
--- a/internal/minikube/docker.go
+++ b/internal/minikube/docker.go
@@ -1,0 +1,43 @@
+package minikube
+
+import (
+	"bufio"
+	"bytes"
+	"context"
+	"fmt"
+	"os/exec"
+	"regexp"
+)
+
+// This isn't perfect (because it won't unquote the value right) but
+// it's good enough for 99% of cases.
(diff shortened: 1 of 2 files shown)


internal/minikube/docker.go
```diff
@@ -0,0 +1,42 @@
...
+	"fmt"
+	"os/exec"
+	"regexp"
+)
+
+// This isn't perfect (because it won't unquote the value right) but
+// it's good enough for 99% of cases.
+var envMatcher = regexp.MustCompile(`export (\w+)="([^"]+)"`)
+
+func DockerEnv() (map[string]string, error) {
```
maiamcc: do we care about running this with a ctx/timeout? this call doesn't seem particularly risky, but idk, good practice? 🤷‍♀️

nicks: good point, added!

PR #173, opened 2018-08-24
build: fix bug where tar treated all top level files as directories

jazzdan:
Huge caveat here: I spent most of the morning trying to write a test for this and I was unable to. Here's how I tested it manually to verify that it is fixed.

1. `make install`
2. `cd ../blorg-tilt`
3.  `tilt up blorg_frontend --watch`
4. Wait for it to come up
5. `cd ../blorg-frontend`
6. Modify main.go

At this point in the old world if you looked at the container on the pod you would see `/go/src/github.com/windmilleng/blorg-frontend/main.go/main.go`. Note the incorrect path. In the new world: you'd see `/go/src/github.com/windmilleng/blorg-frontend/main.go` with the correct, updated contents.

### DIFF
--- a/internal/build/tar.go
+++ b/internal/build/tar.go
@@ -54,6 +54,7 @@ func tarPath(ctx context.Context, tarWriter *tar.Writer, source, dest string) er
 	span.SetTag("dest", dest)
 	defer span.Finish()
 	sourceInfo, err := os.Stat(source)
+
 	if err != nil {
 		if os.IsNotExist(err) {
 			return nil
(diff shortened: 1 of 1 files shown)


internal/build/tar.go
```diff
@@ -84,11 +85,12 @@ func tarPath(ctx context.Context, tarWriter *tar.Writer, source, dest string) er
...
 			// Name of file in tar should be relative to source directory...
 			header.Name = strings.TrimPrefix(path, source)
-		}
-
-		if dest != "" {
-			// ...and live inside `dest` (if given)
-			header.Name = filepath.Join(dest, header.Name)
+			if dest != "" {
+				// ...and live inside `dest` (if given)
+				header.Name = filepath.Join(dest, header.Name)
```
maiamcc: hmm pretty sure it's safe to `filepath.Join("", somePath)` -- so we don't actually need to check if `dest == ""`, can just call join. see: https://play.golang.com/p/XqkOgWmQ9k0

nicks:
lg

PR #175, opened 2018-08-24
build: s/MountsToPath/MountsToPathMappings

jazzdan:


.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
```
nicks: :eyes:

jazzdan: argh good catch thank you

PR #178, opened 2018-08-24
engine: pass through only changed files when building from existing

jazzdan:


jazzdan:
Ah I missed that we have tests for this part of the code now. Will do!

jazzdan:
actually I'm going to punt on that for now, there's a lot of state that needs to be set up to test the watch case.

PR #179, opened 2018-08-24
tilt: make live output match demo

landism:
i.e., https://github.com/windmilleng/tilt/pull/153#issuecomment-415577944

image

caveats:

1. The output assumes we'll always have two pipeline steps. It would be easy for someone to add/remove pipeline step and not notice the const is now wrong.
2. Anything writing directly to the logger looks kinda out of place now. We should probably expand outputter and make it so nothing writes directly to the logger.
3. This is kinda tiptoeing into doing all logging as traces, and it might have been better to just implement this as traces to begin with, but that seemed like too big a task to get done, and this was pretty quick.


internal/output/outputter.go
```diff
@@ -0,0 +1,147 @@
...
+	"io"
+	"time"
+
+	"github.com/windmilleng/tilt/internal/logger"
+)
+
+type Color string
+
+const (
+	cGreen                = Color("\033[32m")
```
nicks: could you use a color library like https://github.com/fatih/color?

most good color libraries will try to detect if you have a terminal that supports color, so that we don't end up printing color codes to terminals that don't understand them

internal/output/outputter.go
```diff
@@ -0,0 +1,147 @@
...
+		cGreen,
+		elapsed.Seconds(),
+		cBlue,
+		cReset)
+	o.curPipelineStep = 0
+}
+
+// how many steps there are per pipeline
+// we might need to change or remove this if it turns out we can't predict how many steps there will be
+const numSteps = 2
```
nicks: maybe this should be a parameter to StartPipeline?

internal/output/outputter.go
```diff
@@ -0,0 +1,147 @@
...
+}
+
+var _ io.Writer = &prefixedWriter{}
+
+func newPrefixedWriter(prefix string, underlying io.Writer) *prefixedWriter {
+	return &prefixedWriter{prefix, underlying, true}
+}
+
+func (i *prefixedWriter) Write(buf []byte) (n int, err error) {
+	for _, c := range buf {
```
nicks: hmmm...writing one character at a time isn't really how the Writer api is supposed to be used, and i'm a little worried this will lead to weird behavior (e.g., defeating buffering on the underlying Writer, breaking multi-byte chars like emoji)

have you considered rewriting the bytes up-front, with something like
s := bytes.Replace(buf, '\n', '\n' + prefix, 0)
i.underlying.Write(s)
?

(obviously, with some extra logic to handle the edge cases)

landism: I suspect it would defeat buffering, but I wouldn't expect it to break multi-byte chars (at least, my experience with unix files is that the way in which you break up the bytes over write calls isn't supposed to affect what comes out).

I sillily figured the single byte implementation would be simpler, but that only really makes sense in terms of allocations, which is pretty unlikely to matter for us, so yeah, might as well go with string replacement

PR #180, opened 2018-08-24
build: don't return nil if we couldn't decode any output from Docker

jazzdan:
I was writing a test and noticed that if we can't read any JSON output from docker we return a nil pointer, which then causes a nil pointer exception. This error is less catastrophic.


nicks:
i don't think this is a good idea. there are a bunch of cases where we don't expect output?

nicks:
i.e., it is expected and well-documented that many docker commands don't all have an 'aux' message

jazzdan:
Yeah ... that's a good point.

PR #177, opened 2018-08-24
mill: enforce repo is absolute path

maiamcc:
Hello @landism, @dmiller,

Please review the following commits I made in branch maiamcc/enforce-abs-repo:

82f8ab53ff1890c62dd166ee29efc0d270458d85 (2018-08-24 12:11:28 -0400)
mill: enforce repo is absolute path

e62010ab1889ca34c5b74a9275540fe2f331b93d (2018-08-24 11:51:23 -0400)
wip

Code review reminders, by giving a LGTM you attest that:

* Commits are adequately tested
* Code is easy to understand and conforms to style guides
* Incomplete code is marked with TODOs
* Code is suitably instrumented with logging and metrics

PR #181, opened 2018-08-24
tilt: make gitignore tests prettier

landism:
1. ignoring .git was failing because we were comparing against relative paths, so use absolute paths when setting up gitignore testers, add a unit test for that case
2. refactor gitignore tests so that they're less tedious

### DIFF
--- a/internal/git/gitignore.go
+++ b/internal/git/gitignore.go
@@ -94,17 +94,11 @@ func (r repoIgnoreTester) IsIgnored(f string, isDir bool) (bool, error) {
 }
 
 func NewRepoIgnoreTester(ctx context.Context, repoRoot string) (IgnoreTester, error) {
-	absRepoRoot, err := filepath.Abs(repoRoot)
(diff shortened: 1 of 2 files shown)


internal/git/gitignore.go
```diff
@@ -94,11 +94,17 @@ func (r repoIgnoreTester) IsIgnored(f string, isDir bool) (bool, error) {
 }
 
 func NewRepoIgnoreTester(ctx context.Context, repoRoot string) (IgnoreTester, error) {
-	g, err := NewGitIgnoreTester(ctx, repoRoot)
+	absRepoRoot, err := filepath.Abs(repoRoot)
```
nicks: still a bit concerned that these need to be absolutified further upstream...

filepath.Abs means 'resolved relative to the process working directory', which seems like a weird thing to do in a low-level library like this

maiamcc:
Matt will #177 solve any of your problems?

PR #183, opened 2018-08-24
build: make spans for relevant steps in Docker output

jazzdan:
Looks like this:

screen shot 2018-08-24 at 3 28 55 pm

### DIFF
--- a/internal/build/docker.go
+++ b/internal/build/docker.go
@@ -288,7 +288,11 @@ func readDockerOutput(ctx context.Context, reader io.Reader) (*json.RawMessage, 
 	defer span.Finish()
 	var result *json.RawMessage
 	decoder := json.NewDecoder(reader)
+	var innerSpan opentracing.Span
 	for decoder.More() {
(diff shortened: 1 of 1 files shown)


internal/build/docker.go
```diff
@@ -298,6 +302,9 @@ func readDockerOutput(ctx context.Context, reader io.Reader) (*json.RawMessage,
 		if len(message.Stream) > 0 && message.Stream != "\n" {
 			msg := strings.TrimSuffix(message.Stream, "\n")
 			output.Get(ctx).Print("%+v", msg)
+			if strings.HasPrefix(msg, "Step") || strings.HasPrefix(msg, "Running") {
+				innerSpan, ctx = opentracing.StartSpanFromContext(ctx, fmt.Sprintf("%s", msg))
```
nicks: msg is already a string? you shouldn't need to sprintf it

jazzdan: oh yeah this was cruft from before, thanks

PR #191, opened 2018-08-27
timing: add span to missingLocalPaths

maiamcc:
also fix bug when fs event contains too many files

### DIFF
--- a/internal/build/path_mapping.go
+++ b/internal/build/path_mapping.go
@@ -1,10 +1,12 @@
 package build
 
 import (
+	"context"
 	"fmt"
 	"os"
 	"path/filepath"
 
+	"github.com/opentracing/opentracing-go"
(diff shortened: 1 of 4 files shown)


internal/build/path_mapping.go
```diff
@@ -63,7 +65,9 @@ func MountsToPathMappings(mounts []model.Mount) []pathMapping {
 }
 
 // Return all the path mappings for local paths that do not exist.
-func missingLocalPaths(mappings []pathMapping) ([]pathMapping, error) {
+func missingLocalPaths(ctx context.Context, mappings []pathMapping) ([]pathMapping, error) {
+	span, ctx := opentracing.StartSpanFromContext(ctx, "daemon-missingLocalPaths")
```
nicks: this seems like a weird thing to want spans for? at this point, why not just use a normal prifiler?

PR #192, opened 2018-08-27
build: improve base dockerfile validation

nicks:
Hello @maiamcc,

Please review the following commits I made in branch nicks/dockerfile:

9aa2f87e596688407ca1b3b8e0e3a12fa3d03256 (2018-08-27 17:57:14 -0400)
build: improve base dockerfile validation

### DIFF
--- a/internal/build/dockerfile.go
+++ b/internal/build/dockerfile.go
@@ -1,12 +1,20 @@
 package build
 
 import (
+	"bytes"
+	"errors"
 	"fmt"
 	"strings"
 
 	"github.com/docker/distribution/reference"
+	"github.com/moby/buildkit/frontend/dockerfile/parser"
 	"github.com/windmilleng/tilt/internal/model"
 )
+
+var ErrEntrypointInDockerfile = errors.New("base Dockerfile contains an ENTRYPOINT/CMD, " +
+	"which is not currently supported -- provide an entrypoint in your Tiltfile")
+var ErrAddInDockerfile = errors.New("base Dockerfile contains an ADD/COPY, " +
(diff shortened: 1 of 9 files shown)


internal/build/dockerfile.go
```diff
@@ -47,15 +55,36 @@ func (d Dockerfile) RmPaths(pathsToRm []pathMapping) Dockerfile {
...
-func (d Dockerfile) ForbidEntrypoint() error {
-	for _, line := range strings.Split(string(d), "\n") {
-		if strings.HasPrefix(line, "ENTRYPOINT") {
+func (d Dockerfile) ValidateBaseDockerfile() error {
+	result, err := parser.Parse(bytes.NewBufferString(string(d)))
+	if err != nil {
+		return fmt.Errorf("ValidateBaseDockerfile: %v", err)
+	}
+
+	err = traverse(result.AST, func(node *parser.Node) error {
```
maiamcc: is this markedly more efficient than doing it the way we did it at left? why bother with this kind of traversal?

so the risk here is that we flag any line that contains the word "add", "copy", "cmd", etc even if it's not Dockerfile syntax, even if it's a comment, etc. -- which is why at left i checked the beginnings of lines. idk if it's worth worrying about now, but something to note.

maiamcc: (or does `parser.Parse` split stuff out in such a way that checking `result.AST` is equivalent to checking the docker verb?)

nicks: i added some tests for cases that you're worried about, both cases where a hand-rolled regexp-based parser would fall down, and cases where you're worried about 'add' in a command.

yes, the AST is structured in a way so that each command is an AST node.

PR #193, opened 2018-08-27
build: use buildkit

jazzdan:
I apologize for the size of this diff, I tried to break it out in to sensible commits below.

This diff replaces all instances of `digest.Digest` with `reference.NamedTagged`. This is necessary for BuildKit to work since it doesn't allow you to use digests in `FROM` statements in Dockerfiles anymore. Doing this touched a lot of files and involved changing not just building from existing docker files, but also from scratch, pushing, and more.

I pass the Buildkit flag in `ImageBuildOptions`.

Instead of returning a Digest from build, I tag the digest immediately and return the `reference.NamedTagged`.

Things that are broken:
1. History. I'm going to fix this in a later diff.
2. Streaming output.

### DIFF
--- a/internal/build/docker.go
+++ b/internal/build/docker.go
@@ -6,7 +6,6 @@ import (
 	"context"
 	"encoding/json"
 	"errors"
-	"flag"
 	"fmt"
 	"io"
 	"log"
@@ -208,12 +207,8 @@ func (l *localDockerBuilder) buildFromDf(ctx context.Context, df Dockerfile, pat
 	imageBuildResponse, err := l.dcli.ImageBuild(
 		ctx,
 		archive,
-		types.ImageBuildOptions{
-			Context:    archive,
-			Dockerfile: "Dockerfile",
-			Remove:     shouldRemoveImage(),
-			Version:    types.BuilderBuildKit,
-		})
+		Options(archive),
+	)
 	spanBuild.Finish()
 	if err != nil {
 		return nil, err
@@ -377,13 +372,6 @@ func getDigestFromAux(aux json.RawMessage) (digest.Digest, error) {
 		return "", fmt.Errorf("getDigestFromAux: ID not found")
 	}
 	return digest.Digest(id), nil
-}
-
-func shouldRemoveImage() bool {
-	if flag.Lookup("test.v") == nil {
-		return false
-	}
-	return true
 }
 
 func digestAsTag(d digest.Digest) (string, error) {
--- a/internal/build/docker.go
+++ b/internal/build/docker.go
@@ -34,8 +34,8 @@ type localDockerBuilder struct {
 }
 
 type Builder interface {
-	BuildDockerFromScratch(ctx context.Context, baseDockerfile Dockerfile, mounts []model.Mount, steps []model.Cmd, entrypoint model.Cmd) (digest.Digest, error)
-	BuildDockerFromExisting(ctx context.Context, existing digest.Digest, paths []pathMapping, steps []model.Cmd) (digest.Digest, error)
+	BuildDockerFromScratch(ctx context.Context, baseDockerfile Dockerfile, mounts []model.Mount, steps []model.Cmd, entrypoint model.Cmd) (reference.NamedTagged, error)
+	BuildDockerFromExisting(ctx context.Context, existing reference.NamedTagged, paths []pathMapping, steps []model.Cmd) (reference.NamedTagged, error)
 	PushDocker(ctx context.Context, name reference.Named, dig digest.Digest) (reference.NamedTagged, error)
 	TagDocker(ctx context.Context, name reference.Named, dig digest.Digest) (reference.NamedTagged, error)
 }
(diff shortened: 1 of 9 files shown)


jazzdan:
In a separate commit I am going to change all of our `digest` variable names to something that makes more sense.

internal/build/docker.go
```diff
@@ -217,10 +212,11 @@ func (l *localDockerBuilder) buildFromDf(ctx context.Context, df Dockerfile, pat
 			Context:    archive,
 			Dockerfile: "Dockerfile",
 			Remove:     shouldRemoveImage(),
+			Version:    types.BuilderBuildKit,
```
jazzdan: Here's where we tell it to use BuildKit.

internal/build/docker.go
```diff
@@ -231,10 +227,20 @@ func (l *localDockerBuilder) buildFromDf(ctx context.Context, df Dockerfile, pat
...
-		return "", fmt.Errorf("ImageBuild: %v", err)
+		return nil, fmt.Errorf("ImageBuild: %v", err)
+	}
+
+	digest, err := getDigestFromAux(*result)
+	if err != nil {
+		return nil, fmt.Errorf("getDigestFromAux: %v", err)
+	}
+
+	nt, err := l.TagDocker(ctx, ref, digest)
```
jazzdan: Here's where we do the tag inside of the build.

internal/engine/build_and_deployer.go
```diff
@@ -89,20 +85,19 @@ func (l localBuildAndDeployer) BuildAndDeploy(ctx context.Context, service model
...
-		name = token.n
+		n = newDigest
 	}
 
 	logger.Get(ctx).Verbosef("(Adding checkpoint to history)")
-	err = l.history.AddAndPersist(ctx, name, d, checkpoint, service)
-	if err != nil {
-		return nil, err
-	}
+	// err = l.history.AddAndPersist(ctx, name, d, checkpoint, service)
```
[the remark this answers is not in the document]
jazzdan: yeah I left a note about that above. Will add a real TODO.

internal/engine/build_and_deployer.go
```diff
@@ -60,27 +58,25 @@ func (l localBuildAndDeployer) BuildAndDeploy(ctx context.Context, service model
...
-		output.Get(ctx).EndPipelineStep()
-		d = newDigest
-
-		name, err = reference.ParseNormalizedNamed(service.DockerfileTag)
+		output.Get(ctx).StartPipelineStep("Building from scratch: [%s]", service.DockerfileTag)
+		newDigest, err := l.b.BuildDockerFromScratch(ctx, name, build.Dockerfile(service.DockerfileText), service.Mounts, service.Steps, service.Entrypoint)
 		if err != nil {
 			return nil, err
 		}
+		output.Get(ctx).EndPipelineStep()
```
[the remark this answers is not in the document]
jazzdan: yes I think so, good call

internal/build/docker.go
jazzdan: I think that's a good idea if it's easy, I'll try it first thing tomorrow morning.

jazzdan:
@yuindustries and I are going to pair on a separate PR to fix the output now. There will be another PR to add history back.

internal/build/options.go
```diff
@@ -0,0 +1,26 @@
...
+package build
+
+import (
+	"bytes"
+	"flag"
+
+	"github.com/docker/docker/api/types"
+)
+
+func Options(archive *bytes.Reader) types.ImageBuildOptions {
```
maiamcc: there are many options in the docker client lib, maybe better name for this func?

internal/build/docker.go
maiamcc: also why does this function now take the ref to the existing docker image??

internal/build/docker.go
```diff
@@ -315,7 +321,7 @@ func readDockerOutput(ctx context.Context, reader io.Reader) (*json.RawMessage,
 			return nil, errors.New(message.Error.Message)
 		}
 
-		if message.Aux != nil {
+		if message.Aux != nil && message.ID != "moby.buildkit.trace" {
```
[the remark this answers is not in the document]
maiamcc: legit. comment?

internal/build/docker_test.go
```diff
@@ -288,10 +296,14 @@ func TestBuildFailingStep(t *testing.T) {
 		model.ToShellCmd("echo hello && exit 1"),
 	}
 
-	_, err := f.b.BuildDockerFromScratch(f.ctx, simpleDockerfile, []model.Mount{}, steps, model.Cmd{})
+	_, err := f.b.BuildDockerFromScratch(f.ctx, f.getNameFromTest(), simpleDockerfile, []model.Mount{}, steps, model.Cmd{})
 	if assert.NotNil(t, err) {
 		assert.Contains(t, err.Error(), "hello")
-		assert.Contains(t, err.Error(), "returned a non-zero code: 1")
+		if runtime.GOOS == "darwin" {
```
maiamcc: awwwwkward :-/ good catch tho

internal/build/options_buildkit.go
```diff
@@ -0,0 +1,26 @@
+// +build buildkit
+
+package build
```
maiamcc: waiiiiit how do we differentiate these files?!

internal/build/docker.go
```diff
@@ -56,37 +55,37 @@ func NewLocalDockerBuilder(dcli DockerClient) *localDockerBuilder {
...
 	defer span.Finish()
 
 	dfForExisting := DockerfileFromExisting(existing)
-	return l.buildDocker(ctx, dfForExisting, paths, steps, model.Cmd{})
+	return l.buildDocker(ctx, dfForExisting, paths, steps, model.Cmd{}, existing)
 }
 
 func (l *localDockerBuilder) buildDocker(ctx context.Context, baseDockerfile Dockerfile,
-	paths []pathMapping, steps []model.Cmd, entrypoint model.Cmd) (digest.Digest, error) {
+	paths []pathMapping, steps []model.Cmd, entrypoint model.Cmd, ref reference.Named) (reference.NamedTagged, error) {
```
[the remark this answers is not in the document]
jazzdan: The main gist of this PR is to make the build API operate in terms of references, not digests. References are a superset of digest, and contain the digest, the name and the tag. The name of the image is something specified by the Tiltfile so, in order to create our reference correctly deep in the bowels of the builder we need to pass the name that the user supplied all the way through.

internal/build/options_buildkit.go
jazzdan: Check out the build tag in the comment at the top! If the buildkit flag is set, this file will be compiled. The other file has a comment that looks like this:

```golang
// +build !buildkit
```

This means that it will only be compiled if the buildkit flag is _not_ set. They are mutually exclusive.

PR #196, opened 2018-08-28
build: parse and display buildkit output

jazzdan:


internal/build/local_docker.go
```diff
@@ -284,6 +289,26 @@ func tarContextAndUpdateDf(ctx context.Context, df Dockerfile, paths []pathMappi
...
+
+	displayStatus := func(displayCh chan *client.SolveStatus) {
+		out := os.Stdout
+		c, err := console.ConsoleFromFile(out)
+		if err != nil {
+			output.Get(ctx).Print("Error making console: %s", err)
+			return
+		}
+		go func() {
+			err := progressui.DisplaySolveStatus(context.TODO(), "", c, out, displayCh)
```
nicks: use the real context?

internal/build/local_docker.go
```diff
@@ -313,10 +338,58 @@ func readDockerOutput(ctx context.Context, reader io.Reader) (*json.RawMessage,
 			return nil, errors.New(message.Error.Message)
 		}
 
+		if message.ID == "moby.buildkit.trace" {
```
nicks: move this out into a helper method?

internal/build/local_docker.go
```diff
@@ -313,10 +338,58 @@ func readDockerOutput(ctx context.Context, reader io.Reader) (*json.RawMessage,
...
+			}
+
+			displayCh <- &s
+		}
+
 		if message.Aux != nil && message.ID != "moby.buildkit.trace" {
 			result = message.Aux
 		}
 	}
+	close(displayCh)
```
nicks: should we put this in a `defer`?

internal/build/local_docker.go
```diff
@@ -284,6 +289,26 @@ func tarContextAndUpdateDf(ctx context.Context, df Dockerfile, paths []pathMappi
 func readDockerOutput(ctx context.Context, reader io.Reader) (*json.RawMessage, error) {
 	span, ctx := opentracing.StartSpanFromContext(ctx, "daemon-readDockerOutput")
 	defer span.Finish()
+
+	displayCh := make(chan *client.SolveStatus)
+
+	displayStatus := func(displayCh chan *client.SolveStatus) {
+		out := os.Stdout
+		c, err := console.ConsoleFromFile(out)
```
nicks: it might be worthwhile to pair on how to use dependency injection to get a console, rather than hard-coding os.Stdout here

jazzdan: I'd love that! maybe tomorrow morning?

PR #197, opened 2018-08-28
k8s: add some code to extract loadbalancers

nicks:
Hello @maiamcc, @jazzdan,

Please review the following commits I made in branch nicks/forward2:

fac11ca66e7d42508e347c3baa3b25841d9365df (2018-08-28 14:23:18 -0400)
k8s: add some code to extract loadbalancers

### DIFF
--- a/internal/k8s/entity.go
+++ b/internal/k8s/entity.go
@@ -1,6 +1,7 @@
 package k8s
 
 import (
+	"k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 )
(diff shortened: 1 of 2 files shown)


internal/k8s/entity.go
```diff
@@ -28,3 +29,47 @@ func ImmutableEntities(entities []K8sEntity) []K8sEntity {
...
 	return result
 }
+
+type LoadBalancer struct {
+	Name  string
+	Ports []int32
+}
+
+func ToLoadBalancers(entities []K8sEntity) []LoadBalancer {
+	result := make([]LoadBalancer, 0)
```
jazzdan: what's the point of making a slice of size 0, and then put something in it? Shouldn't it be a slice of size 1?

internal/k8s/entity_test.go
```diff
@@ -21,3 +21,20 @@ func TestImmutableFilter(t *testing.T) {
...
+
+func TestLoadBalancers(t *testing.T) {
+	entities, err := ParseYAMLFromString(BlorgBackendYAML)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	lbs := ToLoadBalancers(entities)
+	if len(lbs) != 1 {
+		t.Fatalf("Expected 1 loadbalancer, actual: %d", len(lbs))
```
maiamcc: maybe `"Expected 1 loadbalancer, got %d: %v", len(lbs), lbs` for debugging purposes?

internal/k8s/entity.go
nicks: no, because that would create a slice with 1 element, the default element, which is not what we want.

internal/k8s/entity_test.go
nicks: done

internal/k8s/entity.go
jazzdan: I see, so this could also be rephrased as:

```golang
result := []LoadBalancer{}
```

PR #199, opened 2018-08-28
build: turn on buildkit for everyone

jazzdan:


nicks:
seems like there are still problems?

jazzdan:
@nicks just a weird test compatibility problem that I was confused by: I thought the output of a failed docker build was different between old engine and buildkit. What's actually happening is that in buildkit the output seems to differ between darwin and linux, whereas before it didn't. Should be good now.

PR #201, opened 2018-08-28
build, engine: rename/shuffle builders, buildAndDeployers

maiamcc:
Hello tbd,

Please review the following commits I made in branch maiamcc/renames-and-moves:

b6a5d2e5731a9944855edfa0e73193a30a928f44 (2018-08-28 17:47:40 -0400)
build, engine: rename/shuffle builders, buildAndDeployers


maiamcc:
no functional changes, just moving stuff around and renaming stuff.

@nicks proposed that instead of local vs. remote we think about builders/deployers in terms of image-based and container-based, which i like, so i rejiggered this to be in terms of that.

also, it seems silly to have the former `remoteDockerUpdater` implement `Builder` when it didn't use 3/4 of the functionality, so i went ahead and gave it its own interface -- the point of swapping out the two will be the `BuildAndDeployer` interface rather than the `Builder` interface.

PR #202, opened 2018-08-28
engine: automatically port-forward on minikube/gke like docker-for-desktop does

nicks:
Hello @jazzdan, @maiamcc,

Please review the following commits I made in branch nicks/forward3:

71055c2e58ada9264bb62e7548f1e5db32b5a169 (2018-08-28 18:16:14 -0400)
engine: automatically port-forward on minikube/gke like docker-for-desktop does

### DIFF
--- a/internal/cli/wire.go
+++ b/internal/cli/wire.go
@@ -6,7 +6,6 @@ package cli
 import (
 	"context"
 
-	"github.com/docker/docker/client"
 	"github.com/google/go-cloud/wire"
 	"github.com/windmilleng/tilt/internal/build"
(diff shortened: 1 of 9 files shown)


nicks:
the more i play with this, the more i think this is...not a great experience.

1) the kubectl port-forwarding semantics are bad for this use-case, and it will take a bit of wrestling to get them to behave correctly
2) you still have to memorize the port to visit in your web browser

i wonder if we should have a UX more like Glitch, where we automatically open a browser window with your app, and auto-refresh it when your app changes. will think on this a bit.

nicks:
i no longer think this is a good idea! closing

PR #203, opened 2018-08-29
build: if build output is nil then error

jazzdan:
note that ... I don't understand why this fixes the problem. =\

What I expected to happen after making this change is that I'd see the error printed instead of getting the segfault. Instead, afaict no error is printed.


jazzdan:
to be clear this is a fix for https://app.clubhouse.io/windmill/story/75/5-mb-timing-case-fails-with-buildkit

PR #208, opened 2018-08-29
outputter: fix a broken call to Infof, change Print to Printf so that go vet checks it

nicks:
Hello @yuindustries,

Please review the following commits I made in branch nicks/printf:

576e4932c86e9e8bf9810cd2dd76352e51d9f421 (2018-08-29 12:07:05 -0400)
outputter: fix a broken call to Infof, change Print to Printf so that go vet checks it

### DIFF
--- a/internal/output/outputter.go
+++ b/internal/output/outputter.go
@@ -88,9 +88,9 @@ func (o *Outputter) StartBuildStep(format string, a ...interface{}) {
 	o.curBuildStep++
 }
 
-func (o *Outputter) Print(format string, a ...interface{}) {
+func (o *Outputter) Printf(format string, a ...interface{}) {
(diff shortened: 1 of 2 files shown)


internal/output/outputter.go
```diff
@@ -88,9 +88,9 @@ func (o *Outputter) StartBuildStep(format string, a ...interface{}) {
 	o.curBuildStep++
 }
 
-func (o *Outputter) Print(format string, a ...interface{}) {
+func (o *Outputter) Printf(format string, a ...interface{}) {
```
nicks: go vet has magic where it will check that the format string has the right number of arguments if you call the function "Printf"

yuindustries: Neat! Good to know.

PR #211, opened 2018-08-29
build: inject console and stdout

jazzdan:


internal/build/image_builder.go
```diff
@@ -45,6 +47,17 @@ func DefaultBuilder(b *dockerImageBuilder) ImageBuilder {
 	return b
 }
 
+func DefaultConsole() console.Console {
```
nicks: you should take the io.Writer as a parameter here and let wire take care of it

jazzdan: so, `ConsoleFromFile` wants an `*os.File` so I don't think that works

internal/build/image_builder.go
```diff
@@ -402,8 +410,8 @@ func messageIsFromBuildkit(msg jsonmessage.JSONMessage) bool {
 	return msg.ID == "moby.buildkit.trace"
 }
 
-func getDigestFromBuildOutput(ctx context.Context, reader io.Reader) (digest.Digest, error) {
-	aux, err := readDockerOutput(ctx, reader)
+func (l *dockerImageBuilder) getDigestFromBuildOutput(ctx context.Context, reader io.Reader) (digest.Digest, error) {
```
maiamcc: we should probs eventually change the name of this receiver, given the struct name change.

jazzdan: isn't it still a docker image builder? It's an image builder that builds docker images?

maiamcc: ah i mean the letter `l` no longer stands for anything :P

jazzdan: ohhh got it, yes will change

PR #212, opened 2018-08-29
Open the browser if your service has a load balancer

nicks:
Hello @maiamcc, @yuindustries,

Please review the following commits I made in branch nicks/lb:

7568aff34aae951042fce8ce0f144fb5db55b5a9 (2018-08-29 14:49:44 -0400)
Open the browser if your service has a load balancer


nicks:
i think the long-term behavior should be something like

"by default, Tilt runs in auto-browser mode, where it tries to detect the best endpoint to open a browser"

and that you can override the auto behavior in the tiltfile, like

composite_service(my_services, browser="auto")
or
composite_service(my_services, browser="name-of-service")

internal/engine/upper.go
```diff
@@ -34,16 +35,22 @@ type Upper struct {
...
 	watcherMaker watcherMaker
 	timerMaker   timerMaker
+	k8s          k8s.Client
 }
 
 type watcherMaker func() (watch.Notify, error)
 type timerMaker func(d time.Duration) <-chan time.Time
 
-func NewUpper(ctx context.Context, b BuildAndDeployer) (Upper, error) {
+func NewUpper(ctx context.Context, b BuildAndDeployer, k8s k8s.Client) (Upper, error) {
```
maiamcc: while you're in this code, is it easy to remove the spurious `error` return value?

yuindustries:
I see that @maiamcc has reviewed the code, but I'll chime in that in my testing this works well and feels like a definite improvement.

internal/engine/upper.go
nicks: done!

PR #214, opened 2018-08-29
engine: fix bugs in how we pass around build state

nicks:
Hello @jazzdan, @maiamcc,

Please review the following commits I made in branch nicks/token:

218a7675b3b1bb395f36673b3426537f995bb45c (2018-08-29 15:53:49 -0400)
engine: fix bugs in how we pass around build state

### DIFF
--- a/internal/engine/upper.go
+++ b/internal/engine/upper.go
@@ -57,7 +57,7 @@ func (u Upper) CreateServices(ctx context.Context, services []model.Service, wat
 	span, ctx := opentracing.StartSpanFromContext(ctx, "daemon-Up")
 	defer span.Finish()
 
-	buildTokens := make(map[model.ServiceName]*buildToken)
+	buildStates := make(map[model.ServiceName]BuildState)
 
 	var sw *serviceWatcher
 	var err error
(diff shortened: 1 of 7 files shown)


internal/engine/upper.go
```diff
@@ -76,27 +76,30 @@ func (u Upper) CreateServices(ctx context.Context, services []model.Service, wat
...
-				var err error
-				token, err := u.b.BuildAndDeploy(
+				result, err := u.b.BuildAndDeploy(
 					ctx,
 					event.service,
-					buildTokens[event.service.Name],
-					event.files)
+					buildState)
 				if err != nil {
 					logger.Get(ctx).Infof("build failed: %v", err.Error())
```
maiamcc: aaaah i see, because we only set a new build state from the result if err == nil, this way we retain the changed files from the last time through this loop if the build fails and we have to try again, yeah?

internal/engine/upper_test.go
```diff
@@ -132,17 +145,17 @@ func TestUpper_UpWatchFileChangeThenError(t *testing.T) {
 		f.timerMaker.maxTimerLock.Lock()
 		call := <-f.b.calls
 		assert.Equal(t, service, call.service)
-		assert.Equal(t, []string(nil), call.files)
+		assert.Equal(t, []string{}, call.state.FilesChanged())
```
maiamcc: ...this eems weird. why not assert len == 0?

nicks: we want assert.Equal to print the full list of files changed on failure

maiamcc: aaaah sure.

internal/engine/upper.go
nicks: yup! also added a test for this case, which is the main motivation for this pr

PR #215, opened 2018-08-29
build: refactor tar code [ch86]

jazzdan:


internal/build/container_updater.go
```diff
@@ -70,21 +69,19 @@ func (r *containerUpdater) UpdateInContainer(ctx context.Context, cID containerI
...
 func ArchivePathsIfExist(ctx context.Context, paths []pathMapping) (*bytes.Buffer, error) {
-	buf := new(bytes.Buffer)
-	tw := tar.NewWriter(buf)
+	ab := newArchiveBuilder()
 	defer func() {
-		err := tw.Close()
+		err := ab.close()
 		if err != nil {
 			log.Printf("Error closing tar writer: %s", err.Error())
 		}
```
maiamcc: why don't we put the error printing in `ab.close()` so we only need to remember to defer one thing?

internal/build/tar.go
```diff
@@ -2,28 +2,46 @@ package build
...
+func (a *archiveBuilder) close() error {
+	return a.tw.Close()
+}
+
+func (a *archiveBuilder) archiveDf(ctx context.Context, df Dockerfile) error {
 	span, ctx := opentracing.StartSpanFromContext(ctx, "daemon-archiveDf")
 	defer span.Finish()
 	tarHeader := &tar.Header{
 		Name: "Dockerfile",
 		Size: int64(len(df)),
```
maiamcc: i bet putting a recent `ModTime` in here will fix that problem with the unrealistic time stamp or whatever

internal/build/container_updater.go
maiamcc: (tho this is maybe irrelevant if we stop deferring `close()` calls like i'm suggesting in my comment on the PR)

jazzdan:
@maiamcc I agree that it could be better, I'll tackle that in a separate PR now that we have tests!

PR #216, opened 2018-08-29
build: tar archive has a length [ch97] [ch86]

jazzdan:


internal/build/tar_test.go
```diff
@@ -56,6 +56,24 @@ func TestArchivePathsIfExists(t *testing.T) {
...
+	ab := newArchiveBuilder()
+	dfText := "FROM alpine"
+	df := Dockerfile(dfText)
+	err := ab.archiveDf(context.Background(), df)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	ab.close()
+	expected := 2048
```
jazzdan: this is just sort of a golden master test. I couldn't think of a better way to do this.

PR #218, opened 2018-08-29
build: tie the closing of the tar to reading [ch86]

jazzdan:


internal/build/tar.go
```diff
@@ -162,5 +163,5 @@ func tarContextAndUpdateDf(ctx context.Context, df Dockerfile, paths []pathMappi
 		return nil, fmt.Errorf("archiveDf: %v", err)
 	}
 
-	return ab.buf, nil
+	return ab.bytesBuffer()
```
maiamcc: yessss, and this also means that we can get rid of the intermediate `TarContextAndUpdateDf` method and just call this private one directly.

jazzdan: yes!

PR #219, opened 2018-08-29
build: toggle buildkit on env variable [ch98]

jazzdan:
a future commit will make a cli flag that will turn buildkit on


jazzdan:
@nicks what do you think about this env variable approach?

PR #220, opened 2018-08-29
build: get rid of TarContextAndUpdateDf [ch86]

jazzdan:


internal/build/image_builder.go
```diff
@@ -210,10 +210,11 @@ func (d *dockerImageBuilder) buildFromDf(ctx context.Context, df Dockerfile, pat
 
 	output.Get(ctx).StartBuildStep("tarring context")
 
-	archive, err := TarContextAndUpdateDf(ctx, df, paths)
+	tar, err := tarContextAndUpdateDf(ctx, df, paths)
 	if err != nil {
 		return nil, err
 	}
+	archive := bytes.NewReader(tar.Bytes())
```
nicks: out of curiosity, is there a reason you need to write this in a new Reader? i thought the original `tar` implemented the Reader interface just fines

jazzdan: good call, I was just being too specific in the ImageBuildOptions function!

PR #221, opened 2018-08-29
engine: re-wire to make a containerUpdater build-and-deployer (no-op)

maiamcc:
should be a no-op -- containerBuildAndDeployer has an imageBuildAndDeployer on it, currently just calls that one. this PR is the set-up for actually hooking up container-based incremental builds

### DIFF
--- a/internal/build/container_updater.go
+++ b/internal/build/container_updater.go
@@ -15,17 +15,15 @@ import (
 
 const pauseCmd = "/pause"
 
-type ContainerUpdater interface {
-	UpdateInContainer(ctx context.Context, cID containerID, paths []pathMapping, steps []model.Cmd) error
-}
-
-var _ ContainerUpdater = &containerUpdater{}
-
-type containerUpdater struct {
+type ContainerUpdater struct {
 	dcli DockerClient
 }
 
-func (r *containerUpdater) UpdateInContainer(ctx context.Context, cID containerID, paths []pathMapping, steps []model.Cmd) error {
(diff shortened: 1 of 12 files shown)


scripts/timing.py
```diff
@@ -65,11 +65,11 @@ def main():
 
     cases = [
         Case('tilt up 1x', test_tilt_up_once),
-        Case('tilt up again, no change', test_tilt_up_again_no_change),
-        Case('tilt up again, new file', test_tilt_up_again_new_file),
-        Case('watch build from changed file', test_watch_build_from_changed_file),
-        Case('watch build from many changed files', test_watch_build_from_many_changed_files),
-        Case('tilt up, big file (5MB)', test_tilt_up_big_file),
+        # Case('tilt up again, no change', test_tilt_up_again_no_change),
```
jazzdan: 👀

maiamcc: asdfg i KEEP DOING THIS

maiamcc:
added some comments about the chunking of dependencies in the wire code -- is the chunking logical/helpful?

internal/build/container_updater.go
```diff
@@ -15,17 +15,15 @@ import (
 
 const pauseCmd = "/pause"
 
-type ContainerUpdater interface {
```
maiamcc: i guess theoretically one day we might have different kinds of container updaters... this one uses the docker cli, but we might have one that does other stuff?? but idk, i guess we can leave it as as type rather than an interface for now...

PR #223, opened 2018-08-30
engine: skip builds that only contain spurious changed files

nicks:
Hello @jazzdan, @maiamcc,

Please review the following commits I made in branch nicks/ink:

078eda52aff2e2bc9bca9de8f084c2a482da4df7 (2018-08-30 12:36:00 -0400)
engine: skip builds that only contain spurious changed files


nicks:
not super-happy with this solution, but it fixes the problem without requiring the user to make any changes to their Tiltfile, which is nice

PR #225, opened 2018-08-30
build: fix broken output when buildkit is off

nicks:
Hello @jazzdan,

Please review the following commits I made in branch nicks/output:

5d068fe1542aa9e2168d0cc52191648af3dbf810 (2018-08-30 14:16:59 -0400)
build: fix broken output when buildkit is off


nicks:
i think the main reason we're breaking output so much is that we don't have good tests for it; gonna spend some time working on an output-testing harness

PR #229, opened 2018-08-30
engine: wire up container build and deployer

jazzdan:
Mostly getting this up to get some eyes on it. I wouldn't merge this in yet.

Involved writing some code to wrangle which pod we need to be talking to, which @maiamcc wrote.

Currently updating the container takes more than 5 seconds. =\

Three outstanding issues:
1) The output currently looks terrible
```
files changed. rebuilding blorg_frontend. observed 1 changes: [/Users/dan/go/src/github.com/windmilleng/blorg-frontend/main.go]
Deploying to pod: blorg-fe-6f9cddf8cf-ggq9j
Got container ID for pod: 5393f4a83f7d25124249fcfc132ca6f3b01e617d554d04f441f8382ec0dfaebe
Updating container...
Container updated
```
2) ~~We only succesfully update one time, after which it throws this error:~~ Fixed!
```
Updating container...
build failed: Error: No such container:path: :/
```

3) Do we want to have a flag for this behavior?

I think it might be a bug with how we're handling the pathMappings?

### DIFF
--- a/internal/k8s/pod.go
+++ b/internal/k8s/pod.go
@@ -1,0 +1,60 @@
+package k8s
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"os/exec"
+	"strings"
+
+	"github.com/docker/distribution/reference"
+)
+
+// PodWithImage returns the ID of the pod running the given image. We expect exactly one
+// matching pod: if too many or too few matches, throw an error.
+func (k KubectlClient) PodWithImage(ctx context.Context, image reference.NamedTagged) (PodID, error) {
+	_, err := imagesToPods(ctx)
+	if err != nil {
+		return PodID(""), err
+	}
+	// pods := imgPodMap[image.String()]
+	return PodID(""), nil
+}
+
+func imagesToPods(ctx context.Context) (map[reference.NamedTagged][]PodID, error) {
+	c := exec.CommandContext(ctx, "kubectl", "get", "pods", "-o=jsonpath={range .items[*]}{\"\\n\"}{.metadata.name}{\"\\t\"}{range .spec.containers[*]}{.image}")
+	stdoutBuf, stderrBuf := &bytes.Buffer{}, &bytes.Buffer{}
+	c.Stdout = stdoutBuf
+	c.Stderr = stderrBuf
+
+	err := c.Run()
+	if err != nil {
+		return nil, fmt.Errorf("imagesToPods: %v (stderr: %s)", err, stderrBuf.String())
+	}
+
+	fmt.Printf(stdoutBuf.String())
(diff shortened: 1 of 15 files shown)


internal/k8s/pod.go
```diff
@@ -0,0 +1,65 @@
...
+		return PodID(""), fmt.Errorf("Unable to find pods for %s. Found: %+v", image, ip)
+	}
+	if len(pods) > 1 {
+		return PodID(""), fmt.Errorf("Too many pods found for %s: %d", image, len(pods))
+	}
+	return pods[0], nil
+}
+
+func imagesToPods(ctx context.Context) (map[reference.NamedTagged][]PodID, error) {
+	c := exec.CommandContext(ctx, "kubectl", "get", "pods", "-o=jsonpath={range .items[*]}{\"\\n\"}{.metadata.name}{\"\\t\"}{range .spec.containers[*]}{.image}")
```
nicks: maybe use backtick strings, to avoid all the escaping?

internal/k8s/pod.go
```diff
@@ -0,0 +1,65 @@
...
+		if strings.TrimSpace(ln) == "" {
+			continue
+		}
+
+		pair := strings.Split(ln, "\t")
+		if len(pair) != 2 {
+			return nil, fmt.Errorf("could not split line in two on tab: %s", ln)
+		}
+
+		nt, err := ParseNamedTagged(pair[1])
```
nicks: can you also account for the case where a pod has multiple containers?

internal/k8s/pod.go
```diff
@@ -0,0 +1,65 @@
...
+	err := c.Run()
+	if err != nil {
+		return nil, fmt.Errorf("imagesToPods: %v (stderr: %s)", err, stderrBuf.String())
+	}
+
+	return imgPodMapFromOutput(stdoutBuf.String())
+}
+
+func imgPodMapFromOutput(output string) (map[reference.NamedTagged][]PodID, error) {
+	imgsToPods := make(map[reference.NamedTagged][]PodID)
```
nicks: i'm nervous about using an interface as a map key. Go doesn't make any real guarantees that this won't segfault. can you just use the string representation as the key?

internal/k8s/pod.go
jazzdan: That's accounted for [here](https://github.com/windmilleng/tilt/blob/22ec793c4d593a966605429314c8221984b65661/internal/build/container_updater.go#L103).

jazzdan:
Will re-organize, write more tests and submit as smaller PRs. Thanks for the feedback @nicks!

jazzdan:
closing in favor of #237

PR #230, opened 2018-08-31
cli: add a --browser flag that we can turn off in the timing script

nicks:
Hello @jazzdan,

Please review the following commits I made in branch nicks/browser:

dfdf49cb6b9efecfd4c10ebe29fb64788ca1ddc0 (2018-08-31 11:32:58 -0400)
cli: add a --browser flag that we can turn off in the timing script

### DIFF
--- a/internal/cli/up.go
+++ b/internal/cli/up.go
@@ -12,6 +12,7 @@ import (
 	"github.com/fatih/color"
 	"github.com/opentracing/opentracing-go"
 	"github.com/spf13/cobra"
+	"github.com/windmilleng/tilt/internal/engine"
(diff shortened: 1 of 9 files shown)


internal/cli/up.go
```diff
@@ -12,6 +12,7 @@ import (
 	"github.com/fatih/color"
 	"github.com/opentracing/opentracing-go"
 	"github.com/spf13/cobra"
+	engine "github.com/windmilleng/tilt/internal/engine"
```
jazzdan: is this package aliasing unnecessary?

nicks: i dunno, not sure why goimports did that

PR #234, opened 2018-08-31
k8s: can get pod from NamedTagged [ch84]

jazzdan:


internal/k8s/pod_test.go
```diff
@@ -0,0 +1,25 @@
...
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+var podsToImagesOut = `blorg-fe-6b4477ffcd-xf98f	blorg.io/blorgdev/blorg-frontend:tilt-361d98a2d335373f
+cockroachdb-0	cockroachdb/cockroach:v2.0.5
+cockroachdb-1	cockroachdb/cockroach:v2.0.5
+cockroachdb-2	cockroachdb/cockroach:v2.0.5
```
nicks: can you add a test for a pod with more than one container?

jazzdan: yes, will do!

jazzdan:
good call @nicks, that definitely would have broken things.

PR #236, opened 2018-08-31
engine: hook up container update

jazzdan:


jazzdan:
whoops, hang on this is all messed up

PR #237, opened 2018-08-31
engine: wire up container build and deployer

jazzdan:


internal/engine/container_build_and_deployer.go
```diff
@@ -6,33 +6,83 @@ import (
...
+		pID, err := cbd.k8sClient.PodWithImage(ctx, state.LastResult.Image)
+		if err != nil {
+			logger.Get(ctx).Infof("Unable to find pod, falling back to image deploy")
+			return cbd.ibd.BuildAndDeploy(ctx, service, state)
+		}
+		logger.Get(ctx).Infof("Deploying to pod: %s", pID)
+		// get containerID from pID (see container_updater.go --> containerIdForPod)
+		cID, err = cbd.cu.ContainerIDForPod(ctx, pID)
+		if err != nil {
+			logger.Get(ctx).Infof("Unable to find container, falling back to image deploy")
```
nicks: also print the error?

internal/engine/container_build_and_deployer.go
```diff
@@ -6,33 +6,83 @@ import (
...
+		return cbd.ibd.BuildAndDeploy(ctx, service, state)
+	}
+	// Service has already been deployed; try to update in the running container
+
+	// If we don't know the pod that we just deployed to/container we just deployed, get it
+	var cID k8s.ContainerID
+	if state.LastResult.Container.String() == "" {
+		pID, err := cbd.k8sClient.PodWithImage(ctx, state.LastResult.Image)
+		if err != nil {
+			logger.Get(ctx).Infof("Unable to find pod, falling back to image deploy")
```
nicks: also print the error?

internal/engine/build_and_deployer_test.go
```diff
@@ -51,6 +55,88 @@ func TestDockerForMacDeploy(t *testing.T) {
...
+	br := BuildResult{
+		Image: nt,
+	}
+
+	newBR, err := f.bd.BuildAndDeploy(f.Ctx(), SanchoService, NewBuildState(br))
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if f.docker.PushCount != 0 {
```
maiamcc: maybe worth doing this with GKE so we DO expect a push to docker and extra obvious if we don't get one? cuz we use the same assertion (`pushCount != 0`) below to assert that we used container-based build...)

internal/engine/container_build_and_deployer.go
```diff
@@ -6,33 +6,93 @@ import (
...
-	env k8s.Env
+	cu        *build.ContainerUpdater
+	env       k8s.Env
+	k8sClient k8s.Client
 
 	// containerBD can't do initial build, so will call out to ibd for that.
 	// May also fall back to ibd for certain error cases.
 	ibd ImageBuildAndDeployer
+
+	skipContainer bool
```
maiamcc: comment here?

internal/engine/container_build_and_deployer.go
```diff
@@ -6,33 +6,93 @@ import (
...
+		k8sClient:     kCli,
+		ibd:           ibd,
+		skipContainer: skipContainer,
 	}
 }
 
 func (cbd *ContainerBuildAndDeployer) BuildAndDeploy(ctx context.Context, service model.Service, state BuildState) (BuildResult, error) {
 	span, ctx := opentracing.StartSpanFromContext(ctx, "daemon-ContainerBuildAndDeployer-BuildAndDeploy")
 	defer span.Finish()
+	if cbd.skipContainer {
```
maiamcc: and/or comment here?

internal/engine/container_build_and_deployer.go
```diff
@@ -6,33 +6,93 @@ import (
...
+	var cID k8s.ContainerID
+	if state.LastResult.Container.String() == "" {
+		pID, err := cbd.k8sClient.PodWithImage(ctx, state.LastResult.Image)
+		if err != nil {
+			logger.Get(ctx).Infof("Unable to find pod, falling back to image deploy: %s", err.Error())
+			return cbd.ibd.BuildAndDeploy(ctx, service, state)
+		}
+		logger.Get(ctx).Infof("Deploying to pod: %s", pID)
+		// get containerID from pID (see container_updater.go --> containerIdForPod)
+		cID, err = cbd.cu.ContainerIDForPod(ctx, pID)
```
maiamcc: imo `ContainerIDForPod` shouldn't actually live on the container updater (it just lives there now vestigially) -- either move it or leave a TODO

internal/engine/container_build_and_deployer.go
```diff
@@ -6,33 +6,93 @@ import (
...
+		if err != nil {
+			logger.Get(ctx).Infof("Unable to find container, falling back to image deploy: %s", err.Error())
```
maiamcc: nit: `"...%v...", err` saves you chars

internal/engine/container_build_and_deployer.go
```diff
@@ -6,33 +6,93 @@ import (
...
+			return cbd.ibd.BuildAndDeploy(ctx, service, state)
+		}
+		logger.Get(ctx).Infof("Got container ID for pod: %s", cID)
```
maiamcc: nbd but `cID.ShortStr()` is maybe more readable

internal/engine/container_build_and_deployer.go
```diff
@@ -6,33 +6,93 @@ import (
...
+	} else {
+		cID = state.LastResult.Container
+	}
+	// once have cID -- can call cbd.cu.UpdateContainer(...)
+	cf, err := build.FilesToPathMappings(state.FilesChanged(), service.Mounts)
+	if err != nil {
+		return BuildResult{}, err
```
maiamcc: TODO: in some cases we may want to fall back to image build if the container build errors...?

internal/engine/build_and_deployer_test.go
jazzdan: I think we're already doing this?

https://github.com/windmilleng/tilt/blob/71233b012dcc87b443023e32989b6bf09936a22c/internal/engine/build_and_deployer_test.go#L29

PR #241, opened 2018-08-31
buildkit: sensible output (WIP)

yuindustries:
Not done yet. 

- Need to dedupe
- Figure out why errors aren't printing

docker output


internal/build/buildkit_printer.go
```diff
@@ -0,0 +1,78 @@
...
+	for _, v := range vertexes {
+		vMap[v.digest] = vertexAndLogs{
+			vertex: v,
+		}
+	}
+
+	for _, l := range logs {
+		if val, ok := vMap[l.vertex]; ok {
+			if len(l.msg) > 0 {
+				vMap[l.vertex] = vertexAndLogs{
```
jazzdan: It might be better to make `vertexAndLog` a pointer so we don't have to recreate it every time we modify the map.

internal/build/buildkit_printer.go
```diff
@@ -0,0 +1,78 @@
...
+					vertex: val.vertex,
+					logs:   append(val.logs, l),
+				}
+			}
+		}
+	}
+
+	for _, v := range vMap {
+		cmdPrefix := "/bin/sh -c "
```
jazzdan: should we leave a note that this won't work in all cases and we will need to be smarter here?

internal/build/image_builder.go
```diff
@@ -270,19 +267,6 @@ func (d *dockerImageBuilder) readDockerOutput(ctx context.Context, reader io.Rea
 	span, ctx := opentracing.StartSpanFromContext(ctx, "daemon-readDockerOutput")
 	defer span.Finish()
 
-	displayCh := make(chan *client.SolveStatus)
-	defer close(displayCh)
-
-	initDisplayStatusOnce := sync.Once{}
-	initDisplayStatus := func() {
-		go func() {
-			err := progressui.DisplaySolveStatus(ctx, "", d.console, d.out, displayCh)
```
jazzdan: 👋

PR #242, opened 2018-08-31
Add timing scripts that wait for the pod to receive traffic

nicks:
Hello @maiamcc,

Please review the following commits I made in branch nicks/timing2:

03bd50b3175d3c45d9daabec53f5e72248c63192 (2018-08-31 18:57:52 -0400)
Add timing scripts that wait for the pod to receive traffic


jazzdan:
On macOS you'll need to install GNU sed with:

```
brew install gnu-sed --with-default-names
```

scripts/timing.py
```diff
@@ -190,6 +205,34 @@ def write_file(n: int):
...
+    return now
+
+def wait_for_fe_token(token: str):
+    url = fe_url()
+    print('Waiting for token to appear: {}'.format(url))
+
+    found = False
+    while not found:
+        out = curl(url)
+        found = out.find(token) != -1
```
maiamcc: or `if token not in out` -- saves you a line

nicks: good point!

nicks: oh wait, this doesn't work, b/c you need the `found` variable elsewhere in the function

maiamcc: Ha you're right nvm

PR #247, opened 2018-09-04
minikube: set k8s polling interval to 1s

nicks:
Hello @maiamcc,

Please review the following commits I made in branch nicks/minikube:

75c6728e4d244e2d48ebcb53f980c8b678a5c27e (2018-09-04 17:39:34 -0400)
minikube: set k8s polling interval to 1s

### DIFF
--- a/internal/k8s/client.go
+++ b/internal/k8s/client.go
@@ -62,7 +62,10 @@ func (k KubectlClient) OpenService(ctx context.Context, lb LoadBalancer) error {
 
 	if k.env == EnvMinikube {
 		logger.Get(ctx).Infof("Waiting on minikube to load service: %s", lb.Name)
-		cmd := exec.CommandContext(ctx, "minikube", "service", lb.Name, "--url")
+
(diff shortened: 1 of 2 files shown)


internal/k8s/client.go
```diff
@@ -62,7 +62,10 @@ func (k KubectlClient) OpenService(ctx context.Context, lb LoadBalancer) error {
 
 	if k.env == EnvMinikube {
 		logger.Get(ctx).Infof("Waiting on minikube to load service: %s", lb.Name)
-		cmd := exec.CommandContext(ctx, "minikube", "service", lb.Name, "--url")
+
+		intervalSec := "1" // 1s is the smallest polling interval we can set :raised_eyebrow:
```
maiamcc: i'm sad that this isn't the actual emoji but that's probably for the best

nicks: the official name is :face_with_one_eyebrow_raised:, but github doesn't support unicode 10 yet :sob:

PR #248, opened 2018-09-04
engine: respond to @maiamcc feedback

jazzdan:


internal/engine/container_build_and_deployer.go
```diff
@@ -65,29 +66,31 @@ func (cbd *ContainerBuildAndDeployer) BuildAndDeploy(ctx context.Context, servic
...
-		logger.Get(ctx).Infof("Got container ID for pod: %s", cID)
+		logger.Get(ctx).Infof("Got container ID for pod: %s", cID.ShortStr())
 	} else {
 		cID = state.LastResult.Container
 	}
 	// once have cID -- can call cbd.cu.UpdateContainer(...)
 	cf, err := build.FilesToPathMappings(state.FilesChanged(), service.Mounts)
 	if err != nil {
-		return BuildResult{}, err
+		logger.Get(ctx).Infof("Unable to translate changed files to path mappings, falling back to image deploy: %s", err)
```
maiamcc: don't we need to do this same step in an image deploy? i don't think this is a good case to fall back in

internal/engine/container_build_and_deployer.go
```diff
@@ -21,6 +21,7 @@ type ContainerBuildAndDeployer struct {
 	// May also fall back to ibd for certain error cases.
 	ibd ImageBuildAndDeployer
 
+	// skipContainer if true, we don't do a container build, and instead do an image build.
```
maiamcc: `// if true...`?

internal/engine/container_build_and_deployer.go
jazzdan: yes good call

internal/engine/container_build_and_deployer.go
jazzdan: yeah, `skipContainer` is a boolean. If it's set to true, we don't do a container build.

maiamcc: ah sorry i mean idk why you have the name of the variable at the start of the comment!

maiamcc: ha yes but also those are supposed to be complete sentences: `// skipContainer indicates whether to skip container build...` etc. nbd but, 🤷‍♀️

PR #250, opened 2018-09-05
timing: refactor to use servantes, Service object; to be smarter about container builds; trigger go builds"

maiamcc:
pending refactor of the blorgly frontend tests (left them till later cuz it'll take me a bit longer).

among other things, this should make it easier to run a single test against multiple services.

### DIFF
--- a/scripts/timing.py
+++ b/scripts/timing.py
@@ -371,7 +371,7 @@ def test_watch_build_from_new_file(serv: Service) -> float:
     # run `tilt up --watch` and wait for it to finish the initial build
     tilt_proc = run_and_wait_for_stdout(serv.tilt_up_watch_cmd(), '[timing.py] finished initial build')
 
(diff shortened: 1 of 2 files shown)


scripts/timing.py
```diff
@@ -25,15 +26,60 @@
...
+        with open(self.main_go_path, 'a') as gofile:
+            gofile.write('\n// timing.py edit\n')
+        self.main_go_changed = True
+
+
+servantes_path = os.path.join(GOPATH, 'src/github.com/windmilleng/servantes')
+SERVANTES_FE = Service("fe", servantes_path, os.path.join(servantes_path, 'servantes'),
+                       os.path.join(servantes_path, 'servantes/main.go'))
+
+DOGGOS = Service("doggos", servantes_path, os.path.join(servantes_path, 'doggos'),
```
nicks: :dog2:

PR #252, opened 2018-09-05
build: buildkit printer streams output [ch100]

jazzdan:


internal/build/buildkit_printer.go
```diff
@@ -10,84 +10,112 @@ import (
 
 type buildkitPrinter struct {
 	output io.Writer
-	vMap   map[digest.Digest]*vertexAndLogs
+	vData  map[digest.Digest]*vertexAndLogs
+	vOrder []digest.Digest
```
jazzdan: This is a space-for-time tradeoff where we keep a separate list of the digests, in order so that we can print them in order.

internal/build/buildkit_printer.go
```diff
@@ -10,84 +10,112 @@ import (
...
 		output: output,
-		vMap:   map[digest.Digest]*vertexAndLogs{},
+		vData:  map[digest.Digest]*vertexAndLogs{},
+		vOrder: []digest.Digest{},
 	}
 }
 
-func (b *buildkitPrinter) parse(vertexes []*vertex, logs []*vertexLog) {
-
+func (b *buildkitPrinter) parseAndPrint(vertexes []*vertex, logs []*vertexLog) error {
```
jazzdan: New strategy:

Collect all output for this "chunk". Then at the end figure out what, if anything, we should print given what we now know from the Docker daemon.

internal/build/buildkit_printer_test.go
```diff
@@ -1,60 +1,104 @@
 package build
 
 import (
+	"fmt"
+	"io/ioutil"
 	"strings"
 	"testing"
 )
 
+const writeGoldenMaster = false
```
jazzdan: This is kinda a hacky golden master test. I'm gonna change this const to be a runtime flag that we can pass to change what the golden master is.

nicks: have you considered doing this with ldflags?

```
go test -ldflags "-X github.com/windmilleng/tilt/internal/build.WriteGoldenMaster=1"
```

should be ok for this use case if we just stick it in a Makefile

jazzdan: ooh that's a good idea! I'll set that up tomorrow morning, and maybe abstract this out to be generally usable.

internal/build/buildkit_printer.go
```diff
@@ -10,84 +10,112 @@ import (
...
+	cmdPrinted bool
 }
 
-type vertexLog struct {
-	vertex digest.Digest
-	msg    []byte
+// HACK: The prefix we assume here isn't valid for RUNS in exec format
+// Ex: RUN ["echo", "hello world"]
+const cmdPrefix = "/bin/sh -c "
+const buildPrefix = "    ╎ "
```
nicks: this might be the wrong abstraction level for buildPrefix. Take a look at how we did this in Outputter. It was kind of hard to get right. And will only get more complicated if you want to manage color correctly. would be worthwhile to hide it behind an abstract io.Writer

internal/build/buildkit_printer.go
```diff
@@ -10,84 +10,112 @@ import (
...
 					return err
 				}
+
+				for _, l := range vl.logs {
+					sl := strings.TrimSpace(string(l.msg))
+					if len(sl) == 0 {
+						continue
+					}
+					msg := fmt.Sprintf("%s  → %s\n", buildPrefix, sl)
+					_, err := b.output.Write([]byte(msg))
```
nicks: maybe abstract out a b.printf helper method that writes to b.output?

internal/build/buildkit_printer.go
```diff
@@ -10,84 +10,112 @@ import (
...
-	for _, v := range b.vMap {
-		buildPrefix := "    ╎ "
-		cmdPrefix := "/bin/sh -c "
-		name := strings.TrimPrefix(v.vertex.name, cmdPrefix)
-
-		if !strings.HasPrefix(v.vertex.name, cmdPrefix) {
-			continue
-		}
+	for _, d := range b.vOrder {
+		if vl, ok := b.vData[d]; ok {
```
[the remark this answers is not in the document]
jazzdan: Yes that's a good point, we decided to be extra "careful" here but yeah it's probably worth errroring if we don't find it.

internal/build/buildkit_printer.go
jazzdan: attn @yuindustries

PR #254, opened 2018-09-05
build: track the order of keys so that iteration order is consistent

nicks:
Hello @jazzdan,

Please review the following commits I made in branch nicks/test:

1a3b9343405cd71c8a67d608b53902feac647389 (2018-09-05 18:25:45 -0400)
build: track the order of keys so that iteration order is consistent


nicks:
i realize this duplicates some of the logic from the other pr, just trying to get the tests to pass right now

jazzdan:
Thanks for taking care of this @nicks!

PR #259, opened 2018-09-06
engine: get containerID earlier in pipeline

maiamcc:
re: https://app.clubhouse.io/windmill/story/126/get-container-id-as-soon-as-initial-deploy-finishes

behavior works, interface-ing is sorta janky, feedback welcome. lmk if this is too ugly to land.

### DIFF
--- a/internal/engine/build_and_deployer.go
+++ b/internal/engine/build_and_deployer.go
@@ -3,6 +3,7 @@ package engine
 import (
 	"context"
 
+	"github.com/windmilleng/tilt/internal/k8s"
 	"github.com/windmilleng/tilt/internal/model"
 )
 
@@ -14,4 +15,10 @@ type BuildAndDeployer interface {
 	// BuildResult can be used to construct a BuildState, which contains
 	// the last successful build and the files changed since that build.
 	BuildAndDeploy(ctx context.Context, service model.Service, currentState BuildState) (BuildResult, error)
+
+	// BaD needs to be able to get the container that a given build affected, so that
+	// it can do incremental builds on that container if needed.
+	// NOTE(maia): this isn't quite relevant to ImageBuildAndDeployer,
+	// consider putting elsewhere or renaming (`Warm`?).
+	GetContainerForBuild(ctx context.Context, build BuildResult) (k8s.ContainerID, error)
 }
--- a/internal/engine/upper.go
+++ b/internal/engine/upper.go
@@ -149,7 +149,7 @@ func (u Upper) CreateServices(ctx context.Context, services []model.Service, wat
 // populates each BuildState with its corresponding containerID
 func (u Upper) populateContainersForBuildStates(ctx context.Context, buildStates map[model.ServiceName]BuildState) {
 	for serv, state := range buildStates {
-		if state.LastResult.NoContainer() {
+		if !state.LastResult.HasContainer() {
 			cID, err := u.b.GetContainerForBuild(ctx, state.LastResult)
 			if err != nil {
 				logger.Get(ctx).Infof(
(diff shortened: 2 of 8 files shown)


internal/engine/build_and_deployer.go
```diff
@@ -14,4 +15,10 @@ type BuildAndDeployer interface {
 	// BuildResult can be used to construct a BuildState, which contains
 	// the last successful build and the files changed since that build.
 	BuildAndDeploy(ctx context.Context, service model.Service, currentState BuildState) (BuildResult, error)
+
+	// BaD needs to be able to get the container that a given build affected, so that
+	// it can do incremental builds on that container if needed.
+	// NOTE(maia): this isn't quite relevant to ImageBuildAndDeployer,
+	// consider putting elsewhere or renaming (`Warm`?).
+	GetContainerForBuild(ctx context.Context, build BuildResult) (k8s.ContainerID, error)
```
nicks: i thought the original idea was to have this return a new BuildResult, rather than make the caller attach the ContainerID to the BuildResult, tho i guess this is ok too

internal/engine/upper.go
```diff
@@ -139,6 +147,22 @@ func (u Upper) CreateServices(ctx context.Context, services []model.Service, wat
 	return nil
 }
 
+func (u Upper) getContainersForBuildStates(ctx context.Context,
```
nicks: the api signature of this method sounds like you went back and forth on a couple different directions

it uses lots of pointers (which seems to indicate that you're modifying things in place)
it's also called getContainers (which seems to indicate that you're not modifying things in place)
and the actual implementation does both (mutates things in-place and returns things)

let's change this data structure back to a `map[model.ServiceName]BuildState`, and rename the method to `populateContainersForBuildStates`. happy to pair on it if you're confused about mutability and pointers.

scripts/timing.py
```diff
@@ -394,7 +394,7 @@ def test_watch_build_from_new_file(serv: Service, case: str) -> float:
     tilt_proc = run_and_wait_for_stdout(serv.tilt_up_watch_cmd(case), '[timing.py] finished initial build')
 
     # wait a sec for the pod to come up so we can do a container update
-    time.sleep(1.5)
+    time.sleep(2)
```
nicks: :grimacing: seems like this is a sign that you've got some bugs

nicks:
i'm a little bit concerned that that the 2s sleep thing is a strong indicator that concurrency is the right approach here, but concurrency is scaring you more than it should :ghost: but we can tackle that later

internal/engine/upper.go
```diff
@@ -139,6 +145,23 @@ func (u Upper) CreateServices(ctx context.Context, services []model.Service, wat
 	return nil
 }
 
+// populateContainersForBuildStates updates the given map of service --> build state in place;
+// populates each BuildState with its corresponding containerID
+func (u Upper) populateContainersForBuildStates(ctx context.Context, buildStates map[model.ServiceName]BuildState) {
```
maiamcc: @nicks is this what you had in mind?

maiamcc:
@nicks fair enough re: concurrency. i'll probably land this as-is and prettify container output, then come back to making it smart.

Seems like I can run `kubectl rollout status deployment XXXXX` somewhere and wait for it to return, and at that point get the containerID. unless you had a particular other idea? (because at the point at which the imageBuildAndDeploy returns, the pod is still not necessarily up, alas.)

internal/engine/upper.go
nicks: :+1:

nicks:
re: concurrency, i would probably poll until it appears :grimacing:

PR #260, opened 2018-09-06
output: summary section

yuindustries:
Add a placeholder for a section that appears after all the services have been built. It should give some useful information, like the state of various services.

- This PR prioritizes the appearance of the output, and thus is written in a way that doesn't really respect / extend existing systems (ex: logger)

- This kinda overlaps with the `logBuildEvent` function (which I believe @maiamcc wrote). My summary appears the first time `tilt` waits. Output from `logBuildEvent` appears on subsequent file updates. This should probably be handled more elegantly in a future PR.

- I started writing summary.go as its own thing (with a test), thinking the Summary state would contain lots of other information beyond service names. (ex: Error summary, cached status, etc?) But now I'm not sure if maybe it just is a mild extension on the functionality of outputter. In the interest of moving along, I left this in a half-assed state in this PR.

summary


internal/engine/upper.go
```diff
@@ -99,7 +101,9 @@ func (u Upper) CreateServices(ctx context.Context, services []model.Service, wat
 	logger.Get(ctx).Debugf("[timing.py] finished initial build") // hook for timing.py
 
 	if watchMounts {
-		// Give the pod(s) we just deployed a bit to come up
+		output.Get(ctx).PrintSummary(true)
+
+    // Give the pod(s) we just deployed a bit to come up
```
jazzdan: hmm, weird indentation?

internal/output/outputter.go
```diff
@@ -104,6 +105,20 @@ func (o *Outputter) StartBuildStep(format string, a ...interface{}) {
 	o.curBuildStep++
 }
 
+// TODO(Han) - do something better than this watching bool
+func (o *Outputter) PrintSummary(watching bool) {
+	if watching {
```
jazzdan: Would it be sufficient to just say "SUMMARY" every time?

internal/engine/container_build_and_deployer.go
```diff
@@ -76,13 +76,13 @@ func (cbd *ContainerBuildAndDeployer) BuildAndDeploy(ctx context.Context, servic
 	if err != nil {
 		return BuildResult{}, err
 	}
-	logger.Get(ctx).Infof("Updating container...")
+	logger.Get(ctx).Infof("  → Updating container…")
```
jazzdan: I know you said this was left purposefully half-assed. Just a comment that when we make this full-assed we should probably have some programmatic way of indenting/formatting these logs instead of doing it manually.

internal/output/outputter.go
```diff
@@ -124,6 +124,17 @@ func (o *Outputter) StartBuildStep(format string, a ...interface{}) {
 	o.curBuildStep++
 }
 
+func (o *Outputter) PrintSummary(watchMounts bool, summary *summary) {
+	o.logger.Infof("%s", o.blue().Sprint("──┤ Services Built … ├────────────────────────────────────────"))
+	for _, svc := range summary.services {
+		o.logger.Infof("  • %s", svc)
```
jazzdan: Should we put this printing logic on the summary struct as a method?

PR #261, opened 2018-09-06
dockerignore: implement dockerignore ignore tester

jazzdan:
Also reorganized our IgnoreTester interface to separate it from the
gitignore implementation.


internal/ignore/tester.go
```diff
@@ -0,0 +1,5 @@
+package ignore
```
maiamcc: niiice.

internal/dockerignore/ignore_test.go
```diff
@@ -0,0 +1,76 @@
...
+
+func TestOneCharacterExtension(t *testing.T) {
+	tf := newTestFixture(t, "temp?")
+	tf.AssertResult(tf.JoinPath("tempa"), true, false)
+	tf.AssertResult(tf.JoinPath("tempeh"), false, false)
+	tf.AssertResult(tf.JoinPath("temp"), false, false)
+}
+
+type testFixture struct {
+	repoRoot *testutils.TempDirFixture
```
maiamcc: fwiw instead of having this be a named field, if you just do
```
type testFixture struct {
    *testutils.TempDirFixture
    t *testing.T
    ...
}
```
it does something something struct embedding magic that means that a tf has access to all of `TempDirFixture`'s methods on the top level, so you don't need to define `JoinPath` that just calls out to `JoinPath`.

internal/dockerignore/ignore_test.go
```diff
@@ -0,0 +1,76 @@
...
+func TestGlob(t *testing.T) {
+	tf := newTestFixture(t, "*/temp*")
+	tf.AssertResult(tf.JoinPath("somedir", "temporary.txt"), true, false)
+	tf.AssertResult(tf.JoinPath("somedir", "temp"), true, false)
+}
+
+func TestOneCharacterExtension(t *testing.T) {
+	tf := newTestFixture(t, "temp?")
+	tf.AssertResult(tf.JoinPath("tempa"), true, false)
+	tf.AssertResult(tf.JoinPath("tempeh"), false, false)
```
maiamcc: lolz

internal/dockerignore/ignore_test.go
```diff
@@ -0,0 +1,76 @@
...
+	return &tf
+}
+
+func (tf *testFixture) JoinPath(path ...string) string {
+	return tf.repoRoot.JoinPath(path...)
+}
+
+func (tf *testFixture) AssertResult(path string, expectedIsIgnored bool, expectError bool) {
+	isIgnored, err := tf.tester.IsIgnored(path, false)
+	if expectError {
```
maiamcc: ARE you testing any cases where you expect an error if you can think of some, might be worth it?

jazzdan: No, I think I should just remove it. Looking at the code coverage of this test all the errors are IO related, even the ones inside of the library.

PR #262, opened 2018-09-07
output: indicate when pipeline errored

maiamcc:

Please review the following commits I made in branch maiamcc/ch136/good-err-output:

2fbd65bf3b7de7277a884b2650e8da23d504b8fa (2018-09-07 11:35:36 -0400)
output: indicate when pipeline errored

### DIFF
--- a/internal/engine/image_build_and_deployer.go
+++ b/internal/engine/image_build_and_deployer.go
@@ -33,15 +33,15 @@ func NewImageBuildAndDeployer(b build.ImageBuilder, k8sClient k8s.Client, histor
 	}, nil
 }
 
-func (ibd ImageBuildAndDeployer) BuildAndDeploy(ctx context.Context, service model.Service, state BuildState) (BuildResult, error) {
+func (ibd ImageBuildAndDeployer) BuildAndDeploy(ctx context.Context, service model.Service, state BuildState) (br BuildResult, err error) {
 	span, ctx := opentracing.StartSpanFromContext(ctx, "daemon-ImageBuildAndDeployer-BuildAndDeploy")
 	defer span.Finish()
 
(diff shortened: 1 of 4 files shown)


jazzdan:
Can you leave a little sample output here?

internal/engine/image_build_and_deployer.go
```diff
@@ -33,15 +33,15 @@ func NewImageBuildAndDeployer(b build.ImageBuilder, k8sClient k8s.Client, histor
 	}, nil
 }
 
-func (ibd ImageBuildAndDeployer) BuildAndDeploy(ctx context.Context, service model.Service, state BuildState) (BuildResult, error) {
+func (ibd ImageBuildAndDeployer) BuildAndDeploy(ctx context.Context, service model.Service, state BuildState) (br BuildResult, err error) {
```
jazzdan: Why are we using named returns here now?

maiamcc:
screen shot 2018-09-06 at 6 30 38 pm

^ sample output!

internal/engine/image_build_and_deployer.go
maiamcc: cuz we want the deferred call to `EndPipeline` to report on the state of the error that the function is returning. without named returns, we have no way to latch onto the value we're returning, and rely on it having been assigned to a variable beforehand (and not just any variable, but a specific one -- so users can't name multiple errors different things).
```
func Stuff() error {
    var err error
    defer EndPipeline(err)
    ...
    return fmt.Errorf("this error isn't bound to 'err'"!)
```
^ if we don't use named returns, we have no way of grabbing that error we return here.

maiamcc:
gonna land but @dbentley could you also take a look and lmk if i did anything silly?

PR #264, opened 2018-09-07
dockerignore:use official dockerignore library

jazzdan:


internal/dockerignore/ignore.go
```diff
@@ -35,13 +34,36 @@ func NewDockerfileIgnoreTester(repoRoot string) (ignore.Tester, error) {
 	}
 
 	p := path.Join(absRoot, ".dockerignore")
-	patterns, err := dockerignore.ReadIgnoreFile(p)
+	var patterns []string
+
+	f, err := os.Open(p)
+	defer func() { _ = f.Close() }()
```
nicks: this will segfault if the Open failed

internal/dockerignore/ignore.go
```diff
@@ -35,13 +34,36 @@ func NewDockerfileIgnoreTester(repoRoot string) (ignore.Tester, error) {
...
+	switch {
+	case os.IsNotExist(err):
+		pm, err := fileutils.NewPatternMatcher(patterns)
```
nicks: this duplicated code seems like a code smell. maybe break out a helper function? or use the existing ReadDockerignore function to get the patterns?

internal/dockerignore/ignore.go
jazzdan: Whoops, good catch!

PR #271, opened 2018-09-07
engine: reap old images when starting a watch

nicks:
Hello @maiamcc, @jazzdan,

Please review the following commits I made in branch nicks/reaper3:

a6a635487a69c109d4e31167685b4b335bcc79fa (2018-09-07 16:43:45 -0400)
engine: reap old images when starting a watch

### DIFF
--- a/internal/build/fake_client.go
+++ b/internal/build/fake_client.go
@@ -5,6 +5,8 @@ import (
 	"context"
 	"fmt"
 	"io"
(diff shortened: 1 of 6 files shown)


nicks:
fwiw, it's not clear to me how necessary this is -- i still go back and forth on whether there will be cases where we'll want to do incremental images vs container updates

PR #272, opened 2018-09-07
tiltfile: swap order arguments to add()

jazzdan:


README.md
```diff
@@ -41,12 +41,12 @@ Builds a docker image.
   * `entrypoint`: **str**
 * Returns: **Image**
 
-#### Image.add(path, repo)
+#### Image.add(repo, path)
 Adds the content from `repo` into the image at `path`.
```
jazzdan: Should there be better copy here now?

jazzdan:
also will open PRs to fix the syntax in all of our example apps.

internal/tiltfile/skylarkTypes.go
```diff
@@ -97,12 +97,12 @@ func runDockerImageCmd(thread *skylark.Thread, fn *skylark.Builtin, args skylark
...
 
 func addMount(thread *skylark.Thread, fn *skylark.Builtin, args skylark.Tuple, kwargs []skylark.Tuple) (skylark.Value, error) {
-	var mountPoint string
 	var gitRepo gitRepo
+	var mountPoint string
 	if len(fn.Receiver().(*dockerImage).cmds) > 0 {
 		return nil, errors.New("add mount before run command")
 	}
-	err := skylark.UnpackArgs(fn.Name(), args, kwargs, "mount_point", &mountPoint, "local_git_repo", &gitRepo)
+	err := skylark.UnpackArgs(fn.Name(), args, kwargs, "local_git_repo", &gitRepo, "mount_point", &mountPoint)
```
nicks: can we name these arguments `src` and `dest`? i think it would be more idiomatic to support
add(src=repo, dest=path)

internal/tiltfile/tiltfile_test.go
```diff
@@ -29,7 +29,7 @@ func TestGetServiceConfig(t *testing.T) {
 	file := tempFile(
 		fmt.Sprintf(`def blorgly():
   image = build_docker_image("%v", "docker tag", "the entrypoint")
-  image.add('/mount_points/1', local_git_repo('.'))
```
nicks: can you add a test case for passing them in the wrong order, so we can see what the error looks like?

internal/tiltfile/skylarkTypes.go
jazzdan: Yeah sounds good to me.

jazzdan:
Going to hold off on merging this until Monday.

PR #273, opened 2018-09-07
tiltd: remove daemon and proto

nicks:
Hello @maiamcc, @jazzdan,

Please review the following commits I made in branch nicks/proto:

c1f074a01aaa02fdadcd934f809993973ca76a91 (2018-09-07 17:36:57 -0400)
tiltd: remove daemon and proto


nicks:
afaik, we don't plan on bringing these back anytime soon, and we can always get them from git history if we need them

PR #277, opened 2018-09-10
engine: only call GetContainerForBuild if we have an image

jazzdan:
If there's no image, we can't find the pod.

https://app.clubhouse.io/windmill/story/147/segfault-in-incremental-build

### DIFF
--- a/internal/engine/upper_test.go
+++ b/internal/engine/upper_test.go
@@ -29,8 +29,7 @@ type fakeBuildAndDeployer struct {
 	t     *testing.T
 	calls chan buildAndDeployCall
 
-	buildCount                int
-	getContainerForBuildCount int
+	buildCount int
 
 	// Set this to simulate the build failing
 	nextBuildFailure error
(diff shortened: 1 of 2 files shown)


nicks:
unit test?

jazzdan:
@nicks yes will do 👍

jazzdan:
@nicks added, but let me know if this is the right approach. I'm not very familiar with this style of testing. I confirmed that without the first commit this test fails.

internal/engine/upper_test.go
```diff
@@ -61,6 +62,7 @@ func (b *fakeBuildAndDeployer) BuildAndDeploy(ctx context.Context, service model
 }
 
 func (b *fakeBuildAndDeployer) GetContainerForBuild(ctx context.Context, build BuildResult) (k8s.ContainerID, error) {
+	b.getContainerForBuildCount++
 	if build.Image == nil {
```
nicks: do you understand why this test doesn't fail?

nicks: i think the right fix is to make this call b.t.Error

jazzdan: Yeah because this is a test that the build fails. The build itself should still fail. It just shouldn't try to get the container for a build without an image.

 I don't understand the interaction between `nextBuildFailure` and `endToken` though. I can't 100% explain what this test is doing.

jazzdan: ahhhh I understand. Neat!

PR #279, opened 2018-09-10
test: use t.Helper in assert functions

jazzdan:
`t.Helper` is a function in the testing package that omits functions that contain calls to it from stack traces. This is useful for assert functions since we don't care that the assert failed per se, but in which test case the assert failed.


nicks:
i don't think we should do this. this hides information that's useful for diagnosing issues.

i would feel differently if we were writing a general test framework (which is what this method is for) and had evidence that these messages were confusing to users of our test framework, but that's not the case here.

jazzdan:
fair enough!

PR #284, opened 2018-09-10
tiltfile: populate Step triggers

jazzdan:
This is only 80% done, but I need to head out soon and wanted to get this in front of poeple.


internal/model/matcher.go
```diff
@@ -22,3 +26,29 @@ func (c CompositePathMatcher) Matches(f string, isDir bool) (bool, error) {
 }
 
 var _ PathMatcher = CompositePathMatcher{}
+
+type singlePathMatcher struct {
```
jazzdan: there's probably a better place to put this but I put it here for now to avoid circular dependency issues.

internal/model/service.go
```diff
@@ -67,7 +67,7 @@ type Step struct {
 
 	// Optional. If not specified, this step runs on every change.
 	// If specified, we only run the Cmd if the trigger matches the changed file.
-	Trigger PathMatcher
+	Trigger []PathMatcher
```
jazzdan: per discussion on previous PR

nicks: wait, what? you can have a PathMatcher that matches multiple files, so why would you need a slice of PathMatchers here?

internal/model/matcher.go
```diff
@@ -22,3 +26,29 @@ func (c CompositePathMatcher) Matches(f string, isDir bool) (bool, error) {
...
+	repoRoot string
+	path     string
+}
+
+func (p singlePathMatcher) Matches(f string, isDir bool) (bool, error) {
```
nicks: oh, i thought we were going to use the existing docker path matcher that uses golang path patterns?

jazzdan: Yep good idea, I'll just use that which will simplify this PR quite a bit.

internal/model/service.go
jazzdan: yeahhh good point. Okay I'll change it back.

internal/dockerignore/ignore.go
```diff
@@ -84,3 +84,12 @@ func NewMultiRepoDockerIgnoreTester(repoRoots []string) (model.PathMatcher, erro
 
 	return model.CompositePathMatcher{Matchers: testers}, nil
 }
+
+func TriggersToPathMatcher(root string, triggers []string) (model.PathMatcher, error) {
```
jazzdan: this function seems superfluous. I'm going to get rid of it.

internal/tiltfile/tiltfile_test.go
```diff
@@ -351,10 +364,47 @@ func TestRunTrigger(t *testing.T) {
...
+		model.Cmd{
+			Argv: []string{"sh", "-c", "yarn install"},
+		},
+	)
+	packagePath := td.JoinPath("package.json")
+	matches, err := services[0].Steps[0].Trigger.Matches(packagePath, false)
+	if err != nil {
+		t.Fatal(err)
+	}
+	assert.True(
```
nicks: the formatting here seems weird, is there a reason it can't fit on one line?

jazzdan: it totally can be, it was a side effect from when it was much longer. Will fix!

PR #285, opened 2018-09-10
tilt: add synclet proto

landism:


internal/synclet/synclet.proto
```diff
@@ -0,0 +1,22 @@
+syntax = "proto3";
+
+package synclet;
+
+option go_package = "github.com/windmilleng/tilt/internal/synclet/proto;proto";
+
+service Synclet {
+    rpc Update (UpdateRequest) returns (UpdateReply) {}
```
maiamcc: i'm not persuaded that this is the right name for this func? but i'm also not persuaded it's NOT, so 🤷‍♀️

PR #286, opened 2018-09-10
model: Add AsMatchPatterns(), so i can include these in docker ADD statements

nicks:
Hello @jazzdan,

Please review the following commits I made in branch nicks/ch141/match:

fe7d63377fc01837e6ce56d369751935c2b2112e (2018-09-10 18:37:27 -0400)
model: Add AsMatchPatterns(), so i can include these in docker ADD statements

### DIFF
--- a/internal/model/matcher.go
+++ b/internal/model/matcher.go
@@ -4,8 +4,34 @@ type PathMatcher interface {
 	Matches(f string, isDir bool) (bool, error)
 }
 
+type PatternMatcher interface {
+	PathMatcher
+
+	// Express this PathMatcher as a sequence of filepath.Match
+	// patterns. These patterns are widely useful in Docker-land because
+	// they're suitable in .dockerignore or Dockerfile ADD statements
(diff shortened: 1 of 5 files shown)


internal/model/matcher.go
```diff
@@ -2,6 +2,14 @@ package model
...
 type PathMatcher interface {
 	Matches(f string, isDir bool) (bool, error)
+
+	// Express this PathMatcher as a sequence of filepath.Match
+	// patterns. These patterns are widely useful in Docker-land because
+	// they're suitable in .dockerignore or Dockerfile ADD statements
+	// https://docs.docker.com/engine/reference/builder/#add
+	//
+	// Returns false if this PathMatcher is not expressible this way.
+	AsMatchPatterns() ([]string, bool)
```
jazzdan: Should this be a separate interface, since it doesn't apply in all cases?

nicks: i like that idea, done!

PR #287, opened 2018-09-10
synclet: script to deploy synclet, dummy app

maiamcc:
fyi the symlinking scheme is so that you can call ./deploy_synclet from the project root and the dockerfile can pull in the right stuff but you don't have synclet code cluttering the project root too. there's probably a better way but this works :)

### DIFF
--- a/synclet/deploy_utils.py
+++ b/synclet/deploy_utils.py
@@ -6,8 +6,8 @@ ENV_DEVEL = 'devel'
 ENV_PROD = 'prod'
 
 ENV_TO_PROJ = {
-    ENV_DEVEL: 'blorg-dev',
-    ENV_PROD: 'blorg-prod'  # probably? idk!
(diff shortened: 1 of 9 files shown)


synclet/deploy_utils.py
```diff
@@ -0,0 +1,36 @@
+# CONSTANTS
+DEFAULT_TEMPLATE = 'synclet/synclet-conf.template.yaml'
+DEFAULT_DOCKERFILE = 'Dockerfile.synclet'
+
+ENV_DEVEL = 'devel'
+ENV_PROD = 'prod'
+
+ENV_TO_PROJ = {
+    ENV_DEVEL: 'blorg-dev',
```
landism: are these worth changing?

maiamcc: ehhh, that's the cluster name. i suppose we could make a new cluster for all this testing stuff, but 🤷‍♀️

PR #291, opened 2018-09-11
model: implement boil steps

jazzdan:
One open question:

If there are no changed files, should we take that as a signal that this is an initial build? Or have a more explicit signal? Right now if there are no changed files we don't run any cmds. I could instead make it such that if there are no changed files we run _all _ cmds.

I also need to resolve a circular dependency that this code introduces when combined with @yuindustries's PR.


internal/model/service.go
```diff
@@ -143,12 +143,28 @@ func ToShellSteps(cmds []string) []Step {
...
+	res := []Cmd{}
+	for _, step := range steps {
+		if step.Trigger == nil {
+			res = append(res, step.Cmd)
+			continue
+		}
+		for _, f := range filesChanged {
+			matches, err := step.Trigger.Matches(f, false)
+			if err != nil {
+				// TODO(dmiller)
```
jazzdan: going to make this return an error.

nicks:
seems like there are a bunch of build errors?

jazzdan:
yeah @yuindustries' PR is going to fix those, and then I will rebase.

internal/engine/image_build_and_deployer.go
```diff
@@ -66,9 +66,10 @@ func (ibd *ImageBuildAndDeployer) build(ctx context.Context, service model.Servi
 		output.Get(ctx).StartPipelineStep("Building from scratch: [%s]", service.DockerfileTag)
 		defer output.Get(ctx).EndPipelineStep()
 
+		// TODO(dmiller): don't boil steps if this is an initial build
 		df := build.Dockerfile(service.DockerfileText)
-		steps := service.Steps
-		ref, err := ibd.b.BuildImageFromScratch(ctx, name, df, service.Mounts, steps, service.Entrypoint)
+		steps := model.BoilSteps(service.Steps, state.FilesChanged())
```
nicks: I'm pretty sure that FilesChanged() is expected to be an absolute path on the local fileysystem, so i don't see how this can work?

nicks: (it would be helpful if there was some sort of an end-to-end test, because filepath mismatch is something that we've screwed up a lot in the past)

jazzdan: The PathMatchers are created in the Tiltfile with their root set to the absolute path of the cwd. Therefore the fact that FilesChanged() are absolute paths is expected and OK.

jazzdan: aha! we need to be able to associate triggers both with their local paths, and with their container paths. Pausing on this while we figure out how to best support this.

jazzdan:
I'm going to close this for now.

PR #299, opened 2018-09-11
tiltfile should have a read file built in

jazzdan:


jazzdan:
ah yes good call @landism!

jazzdan:
I think I'm going to try to bundle a bunch of these changes up and release them close to simultaneously to reduce the churn for Replicated, just FYI.

PR #309, opened 2018-09-12
build: pass through triggers

jazzdan:
~not ready for review yet~

Good to go!

### DIFF
--- a/internal/build/boil_steps.go
+++ b/internal/build/boil_steps.go
@@ -4,25 +4,20 @@ import (
 	"github.com/windmilleng/tilt/internal/model"
 )
 
-type BoiledStep struct {
-	cmd         model.Cmd
-	pathMapping pathMapping
-}
-
-func BoilSteps(steps []model.Step, pathMappings []pathMapping) ([]BoiledStep, error) {
-	res := []BoiledStep{}
+func BoilSteps(steps []model.Step, pathMappings []pathMapping) ([]model.Cmd, error) {
+	res := []model.Cmd{}
 	for _, step := range steps {
 		if step.Trigger == nil {
-			res = append(res, BoiledStep{cmd: step.Cmd})
+			res = append(res, step.Cmd)
 			continue
 		}
 		for _, pm := range pathMappings {
 			matches, err := step.Trigger.Matches(pm.LocalPath, false)
 			if err != nil {
-				return []BoiledStep{}, err
+				return []model.Cmd{}, err
 			}
 			if matches {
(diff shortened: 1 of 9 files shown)


internal/build/boil_steps.go
```diff
@@ -0,0 +1,31 @@
+package build
+
+import (
+	"github.com/windmilleng/tilt/internal/model"
+)
+
+type BoiledStep struct {
+	cmd         model.Cmd
+	pathMapping pathMapping
```
nicks: i don't understand what the goal of this pathMapping property is? afaict, it's never used

internal/build/boil_steps.go
```diff
@@ -0,0 +1,31 @@
...
+
+func BoilSteps(steps []model.Step, pathMappings []pathMapping) ([]BoiledStep, error) {
+	res := []BoiledStep{}
+	for _, step := range steps {
+		if step.Trigger == nil {
+			res = append(res, BoiledStep{cmd: step.Cmd})
+			continue
+		}
+		for _, pm := range pathMappings {
+			matches, err := step.Trigger.Matches(pm.LocalPath, false)
```
nicks: it seems like you could pass in the LocalPath as FilesChanged(), like you had before? you don't use the rest of the pathMapping

internal/build/boil_steps.go
jazzdan: Yeah looks like we don't need this anymore, I'll remove it.

internal/build/container_updater.go
```diff
@@ -60,6 +60,7 @@ func (r *ContainerUpdater) UpdateInContainer(ctx context.Context, cID k8s.Contai
 
 	// Exec steps on container
 	for _, s := range steps {
+		fmt.Printf("Executing step %v in container\n", s.Argv)
```
nicks: :eyes:

internal/engine/local_container_build_and_deployer.go
```diff
@@ -59,7 +59,13 @@ func (cbd *LocalContainerBuildAndDeployer) BuildAndDeploy(ctx context.Context, s
 		return BuildResult{}, err
 	}
 	logger.Get(ctx).Infof("  → Updating container…")
-	err = cbd.cu.UpdateInContainer(ctx, cID, cf, model.BoilStepsTODO(service.Steps))
+	boiledSteps, err := build.BoilSteps(service.Steps, cf)
+	if err != nil {
+		return BuildResult{}, err
+	}
+
+	fmt.Printf("Boiled steps: %+v \n", boiledSteps)
```
nicks: :eyes:

PR #310, opened 2018-09-12
tilt: implement synclet

landism:


internal/synclet/synclet.go
```diff
@@ -19,8 +23,70 @@ func NewSynclet(dcli build.DockerClient) *Synclet {
 	return &Synclet{dcli: dcli}
 }
 
-func (s Synclet) GetContainerIdForPod(podId string) (string, error) {
-	return "", errors.New("GetContainerIdForPod not implemented")
+const pauseCmd = "/pause"
+
+func (s Synclet) GetContainerIdForPod(ctx context.Context, podId string) (string, error) {
```
maiamcc: should be using k8s.PodID wherever possible -- it IS just a string, but (according to nick) a good way to make args transparent and enforce that we're passing not only a string but a VALID string...?

internal/synclet/synclet.go
```diff
@@ -19,8 +23,70 @@ func NewSynclet(dcli build.DockerClient) *Synclet {
...
+	// What?? No actual matches??!
+	return "", fmt.Errorf("no real containers -- all were '/pause' containers")
+}
+
+func (s Synclet) writeFiles(ctx context.Context, containerId string, tarArchive []byte) error {
+	if tarArchive == nil {
+		return nil
+	}
+
+	err := s.dcli.CopyToContainerRoot(ctx, containerId, bytes.NewBuffer(tarArchive))
```
maiamcc: why not just `return s.dcli.CopyTo...`?

internal/synclet/synclet.go
```diff
@@ -19,8 +23,70 @@ func NewSynclet(dcli build.DockerClient) *Synclet {
...
+	return err
+}
+
+func (s Synclet) rmFiles(ctx context.Context, containerId string, filesToDelete []string) error {
```
maiamcc: again, should take k8s.ContainerId probs?

internal/synclet/synclet.go
```diff
@@ -19,8 +23,70 @@ func NewSynclet(dcli build.DockerClient) *Synclet {
...
+	if len(filesToDelete) == 0 {
+		return nil
+	}
+
+	id := k8s.ContainerID(containerId)
+	cmd := model.Cmd{Argv: append([]string{"rm"}, filesToDelete...)}
+
+	err := s.dcli.ExecInContainer(ctx, id, cmd)
```
maiamcc: `return s.dcli.Exec...`

internal/synclet/synclet.go
```diff
@@ -19,8 +23,70 @@ func NewSynclet(dcli build.DockerClient) *Synclet {
...
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (s Synclet) restartContainer(ctx context.Context, containerId string) error {
+	err := s.dcli.ContainerRestartNoWait(ctx, containerId)
```
maiamcc: `return s.dcli...`

internal/synclet/synclet.go
```diff
@@ -30,10 +96,24 @@ func (s Synclet) UpdateContainer(
 	filesToDelete []string,
 	commands []model.Cmd) error {
 
-	fmt.Println("container:", containerId)
-	fmt.Println("bytes in tar:", len(tarArchive))
-	fmt.Println("files to delete:", filesToDelete)
-	fmt.Println("cmds:", commands)
+	err := s.rmFiles(ctx, containerId, filesToDelete)
+	if err != nil {
+		return err
```
maiamcc: maybe worth attaching metadata to these errs (i.e. the func that threw them)? but 🤷‍♀️

internal/synclet/synclet.go
maiamcc: add a todo to dedupe this function

internal/synclet/synclet.go
landism: Mostly to minimize edits/diffs when adding more calls or reordering calls (in that one wouldn't have to change `return s.dcli.CopyTo` to `err = s.dcli`...`if err != nil`), for the same reason some languages allow trailing commas after the last item in a list, but I don't know what's idiomatic go and don't really feel strongly either way.

internal/synclet/synclet.go
landism: whoops, thanks

internal/synclet/synclet.go
maiamcc: eh. fair. doesn't really matter i guess.

PR #313, opened 2018-09-12
output: Summary section has K8s Info

yuindustries:
Merrily dumping more info here in the Summary section!

summary - k8s

### DIFF
--- a/internal/summary/summary.go
+++ b/internal/summary/summary.go
@@ -13,9 +13,16 @@ type Summary struct {
 }
 
 type Service struct {
-	Name     string
-	Path     string
-	K8sTypes []string
+	Name    string
+	Path    string
+	k8sData []k8sDatum
+}
+
+type k8sDatum struct {
+	LoadBalancer []k8s.LoadBalancer
+	Group        string
+	Kinds        []string
+	Version      string
 }
 
 // NewSummary returns summary state
@@ -36,19 +43,31 @@ func (s *Summary) Gather(services []model.Service) error {
 		}
 
 		svcSummary := &Service{
-			Name: string(svc.Name),
-			Path: path,
+			Name:    string(svc.Name),
+			Path:    path,
+			k8sData: []k8sDatum{},
 		}
+
+		kubeDatum := &k8sDatum{}
 
 		entities, err := k8s.ParseYAMLFromString(svc.K8sYaml)
 		if err != nil {
 			return err
 		}
 
-		for _, e := range entities {
-			svcSummary.K8sTypes = append(svcSummary.K8sTypes, e.Kind.Kind)
+		lbs := k8s.ToLoadBalancers(entities)
+
+		for _, lb := range lbs {
+			kubeDatum.LoadBalancer = append(kubeDatum.LoadBalancer, lb)
 		}
 
+		for _, e := range entities {
+			kubeDatum.Group = e.Kind.Group
+			kubeDatum.Version = e.Kind.Version
+			kubeDatum.Kinds = append(kubeDatum.Kinds, e.Kind.Kind)
+		}
+
+		svcSummary.k8sData = append(svcSummary.k8sData, *kubeDatum)
 		s.Services = append(s.Services, svcSummary)
 	}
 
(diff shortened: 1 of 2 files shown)


internal/summary/summary.go
```diff
@@ -24,29 +33,87 @@ func NewSummary() *Summary {
...
+			}
+
+			if len(k.LoadBalancer) > 0 {
+				ret += fmt.Sprintf("    LOAD BALANCER:")
+				for _, lb := range k.LoadBalancer {
+					ret += fmt.Sprintf(" %s", lb.Name)
+
+					if len(lb.Ports) > 0 {
+						ret += fmt.Sprintf(" | PORT: ")
+						for _, p := range lb.Ports {
```
yuindustries: This seems like an egregious amount of code nesting. I'm open to ways of writing this that is easier to reason about!

yuindustries:
Hm, not totally sure how to fix this test failure in CI...

```
go vet -all -printfuncs=Verbosef,Infof,Debugf ./...
! grep --include=\*.go -rn . -e '^[^/].*defer [^ ]*EndPipeline(' # linting for improperly deferred EndPipeline calls; should be in closure, i.e. `defer func() { ...EndPipeline(err) }()`
errcheck -ignoretests -ignoregenerated ./...
internal/engine/upper.go:81:10:	s.Gather(services)
Makefile:32: recipe for target 'errcheck' failed
make: *** [errcheck] Error 1
Exited with code 2
```

Don't seem to see this error when I run this locally
```
go vet -all -printfuncs=Verbosef,Infof,Debugf ./...
```

Did I do something to affect `EndPipeline`? 

I don't see any error in my IDE for `s.Gather(services)`...

internal/summary/summary.go
```diff
@@ -12,8 +13,16 @@ type Summary struct {
...
 
 type Service struct {
-	Name string
-	Path string
+	Name    string
+	Path    string
+	k8sData []k8sDatum
+}
+
+type k8sDatum struct {
```
maiamcc: ha this is pedantic but also grammatically correct... except that imo this struct holds multiple pieces of data? 🤷‍♀️

internal/summary/summary.go
```diff
@@ -24,29 +33,87 @@ func NewSummary() *Summary {
...
+		kubeDatum := &k8sDatum{}
+
+		entities, err := k8s.ParseYAMLFromString(svc.K8sYaml)
+		if err != nil {
+			return err
+		}
+
+		lbs := k8s.ToLoadBalancers(entities)
+
+		for _, lb := range lbs {
```
maiamcc: any reason not to just do `kubeDatum.lbs = k8s.ToLoadBalancers(entities)`?

or better yet, save yourself a line of code with:
```
kubeDatum := &k8sDatum{
    lbs: k8s.ToLoadBalancers(entities)
}
```

internal/summary/summary.go
```diff
@@ -24,29 +33,87 @@ func NewSummary() *Summary {
...
-			Path: path,
-		})
+
+		svcSummary := &Service{
+			Name:    string(svc.Name),
+			Path:    path,
+			k8sData: []k8sDatum{},
+		}
+
+		kubeDatum := &k8sDatum{}
```
maiamcc: this shouldn't be a pointer if you're ultimately storing `[]k8sDatum` (i.e. a slice of NOT pointers, but the structs themselves) -- see below where you call `append(svcSummary.k8sData, *kubeDatum)`. 

if you initialize here with `kubeDatum := &k8sDatum{}` then you don't have to use the star in the `append` call below. lmk if you wanna chat more about pointers.

internal/summary/summary.go
```diff
@@ -24,29 +33,87 @@ func NewSummary() *Summary {
...
 		}
-		s.Services = append(s.Services, &Service{
-			Name: string(svc.Name),
-			Path: path,
-		})
+
+		svcSummary := &Service{
+			Name:    string(svc.Name),
+			Path:    path,
+			k8sData: []k8sDatum{},
```
maiamcc: don't need this line -- leave it out and the field will auto-initialize to the empty value ([ex here](https://play.golang.com/p/-tdSejvykOi)).

internal/summary/summary.go
```diff
@@ -24,29 +33,87 @@ func NewSummary() *Summary {
...
+
+		lbs := k8s.ToLoadBalancers(entities)
+
+		for _, lb := range lbs {
+			kubeDatum.LoadBalancer = append(kubeDatum.LoadBalancer, lb)
+		}
+
+		for _, e := range entities {
+			kubeDatum.Group = e.Kind.Group
+			kubeDatum.Version = e.Kind.Version
```
maiamcc: ...will the entries ever have a different Group or Version from each other? in this code, your datum will only hang onto the group/version of the very last entry you come upon

internal/summary/summary.go
```diff
@@ -24,29 +33,87 @@ func NewSummary() *Summary {
...
+			kubeDatum.Kinds = append(kubeDatum.Kinds, e.Kind.Kind)
+		}
+
+		svcSummary.k8sData = append(svcSummary.k8sData, *kubeDatum)
```
maiamcc: ...wait, so a given service summary... will only ever have a single datum attached?

internal/summary/summary.go
yuindustries: Oh, this is a very good point! I see the logic error here now. I think I will hastily merge in this commit just to keep barreling forth, but I made a note of this issue to investigate further. It may also be the case that I don't really need this info for my summary, anyhow.

internal/summary/summary.go
yuindustries: Clearly all of this was an elaborate scheme to demonstrate "data vs datum" pedantry. I'm glad you flagged it because in the end I really don't need `[]k8sDatum` at all. Not sure why I thought I did.

PR #322, opened 2018-09-13
tilt: add functions enabling us to find a synclet's pod

landism:
1. pull the actual exec out of KubectlClient so that we can fake it for tests
2. add GetNodeForPod and FindAppByNode so that given a service's PodID, we can find the PodID of the synclet running on the same node

### DIFF
--- a/internal/k8s/client.go
+++ b/internal/k8s/client.go
@@ -100,15 +100,15 @@ func (k KubectlClient) Apply(ctx context.Context, entities []K8sEntity) error {
 	defer span.Finish()
 	// TODO(dmiller) validate that the string is YAML and give a good error
 	logger.Get(ctx).Infof("%sApplying via kubectl", logger.Tab)
-	stderrBuf, err := k.kubectlRunner.cli(ctx, "apply", entities...)
+	_, stderr, err := k.kubectlRunner.cli(ctx, "apply", entities...)
 	if err != nil {
-		return fmt.Errorf("kubectl apply: %v\nstderr: %s", err, stderrBuf.String())
+		return fmt.Errorf("kubectl apply: %v\nstderr: %s", err, stderr)
 	}
 	return nil
 }
 
 func (k KubectlClient) Delete(ctx context.Context, entities []K8sEntity) error {
-	_, err := k.kubectlRunner.cli(ctx, "delete", entities...)
+	_, _, err := k.kubectlRunner.cli(ctx, "delete", entities...)
 	_, isExitErr := err.(*exec.ExitError)
 	if isExitErr {
 		// In general, an exit error is ok for our purposes.
--- a/internal/k8s/client.go
+++ b/internal/k8s/client.go
@@ -1,10 +1,8 @@
 package k8s
 
 import (
-	"bytes"
 	"context"
 	"fmt"
-	"io"
 	"net/url"
 	"os/exec"
 	"strings"
@@ -13,11 +11,11 @@ import (
 	"github.com/opentracing/opentracing-go"
 	"github.com/pkg/browser"
 	"github.com/windmilleng/tilt/internal/logger"
-	"github.com/windmilleng/tilt/internal/output"
 )
 
 type PodID string
 type ContainerID string
+type NodeID string
 
 func (pID PodID) String() string { return string(pID) }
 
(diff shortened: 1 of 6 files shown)


internal/k8s/kubectl_runner.go
```diff
@@ -0,0 +1,48 @@
...
+	"bytes"
+	"context"
+	"fmt"
+	"io"
+	"os/exec"
+
+	"github.com/windmilleng/tilt/internal/output"
+)
+
+type kubectlRunner interface {
```
nicks: this is fine for now, but i'm concerned that we're starting to implement a poor, half-implemented clone of kubernetes/client-go.

I wonder how long we should go down this path before using the official go client,  so that we can re-use connections, get well-typed responses and errors instead of parsing nebulous blobs, etc

internal/k8s/client.go
```diff
@@ -13,11 +11,11 @@ import (
 	"github.com/opentracing/opentracing-go"
 	"github.com/pkg/browser"
 	"github.com/windmilleng/tilt/internal/logger"
-	"github.com/windmilleng/tilt/internal/output"
 )
 
 type PodID string
 type ContainerID string
```
maiamcc: ...it occurs to me that this is actually a Docker thing rather than a k8s thing... ugh we should have a Docker package where this can live...

internal/k8s/pod.go
```diff
@@ -64,3 +64,66 @@ func imgPodMapFromOutput(output string) (map[string][]PodID, error) {
...
 	return imgsToPods, nil
 }
+
+func (k KubectlClient) GetNodeForPod(ctx context.Context, podID PodID) (NodeID, error) {
+	jsonPath := "-o=jsonpath={.spec.nodeName}"
+	b, err := k.kubectlRunner.cli(ctx, fmt.Sprintf("get pods %s '%s'", podID.String(), jsonPath))
+
+	out := b.String()
+
+	if err != nil {
```
nicks: should this be handled by kubectlRunner?

internal/k8s/pod.go
```diff
@@ -64,3 +64,66 @@ func imgPodMapFromOutput(output string) (map[string][]PodID, error) {
...
+		return PodID(lines[0]), nil
+	}
+}
+
+func nonEmptyLines(s string) []string {
+	lines := strings.Split(s, "\n")
+
+	var ret []string
+	for _, line := range lines {
+		if len(line) > 0 {
```
maiamcc: idk what the likelyhood is of a line containing e.g. just spaces, but might want to check `len(strings.TrimSpace(line))` in case.

internal/k8s/pod.go
```diff
@@ -64,3 +64,66 @@ func imgPodMapFromOutput(output string) (map[string][]PodID, error) {
...
+
+	if err != nil {
+		if exitError, ok := err.(*exec.ExitError); ok {
+			return NodeID(""), fmt.Errorf("imagesToPods: stderr: %s)", exitError.Stderr)
+		} else {
+			return NodeID(""), fmt.Errorf("imagesToPods: %v", err.Error())
+		}
+	}
+
+	lines := nonEmptyLines(out)
```
nicks: it might be more idiomatic to use a bufio.Scanner

internal/k8s/client.go
```diff
@@ -29,21 +27,32 @@ func (cID ContainerID) ShortStr() string {
...
+
 	// Waits for the LoadBalancer to get a publicly available URL,
 	// then opens that URL in a web browser.
 	OpenService(ctx context.Context, lb LoadBalancer) error
 }
 
 type KubectlClient struct {
-	env Env
+	env           Env
+	kubectlRunner kubectlRunner
```
maiamcc: since we're mocking at this level, is it now redundant to have a `fakeK8sClient` type? Or are they both useful?

nicks:
also, i've said this a few times now, but i feel like it's going to be way more robust to run the synclet as a sidecar container in the same pod, which is a way more well-supported pattern

internal/k8s/pod.go
landism: yep, thanks

internal/k8s/kubectl_runner.go
landism: Yeah, all this jsonpathing and output parsing makes me uneasy, and using the go client sounds a lot better. I wasn't sure if we had a good reason for execing kubectl.

internal/k8s/client.go
landism: I think they're both useful. faking things out via the runner would require tests using a k8s client to know about its implementation (like what output the kubectl process should generate to get the desired result).

PR #324, opened 2018-09-13
summary: only print the services that have changed

nicks:
Hello @yuindustries,

Please review the following commits I made in branch nicks/ch242/all:

018d99d8e334a90bc6d6be9904ca6d4040b74ff8 (2018-09-13 14:32:13 -0400)
summary: only print the services that have changed

### DIFF
--- a/internal/engine/upper.go
+++ b/internal/engine/upper.go
@@ -158,7 +158,7 @@ func (u Upper) CreateServices(ctx context.Context, services []model.Service, wat
 				}
 				logger.Get(ctx).Debugf("[timing.py] finished build from file change") // hook for timing.py
 
-				output.Get(ctx).Printf("%s", s.Output())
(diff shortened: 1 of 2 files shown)


yuindustries:
This PR conflicts with the idea for the output I got from the conversation I just had with @dbentley. His idea was to always show the full list of services, but in a much more compact format:

```
fe: ./servantes → `kubectl get deploy servantes` [http://localhost:8080]
vigoda: ./vigoda → `kubectl get deploy vigoda`
fortune: ./fortune → `kubectl get deploy fortune`
etc
```

There might also be a section right before this, which summarizes the services which _actually_ updated. But the summary section is meant to be a little bit more comprehensive so you can spot changes from run to run. Let me know if you have more questions about this, @nicks or @maiamcc.

nicks:
hmmm...ok. i'm still worried that it will be confusing to print services that haven't changed, but happy to withdraw this for now

PR #330, opened 2018-09-13
tiltfile: make entrypoint an optional parameter

jazzdan:
Still waiting to be able to test this on master before I call it "done done".

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,13 +32,13 @@ Creates a `repo` with the content at `path`.
     * `path`: **str**
 * Returns: **Repo**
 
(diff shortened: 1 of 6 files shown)


jazzdan:
A consequence of this PR is that we no longer try to validate that things are set up correctly for k8s to execute the manifest. The philosophy than @dbentley imparted on me is that we should be transparent enough about what k8s is doing that it will be obvious why it has failed, rather than try to predict every failure case.

jazzdan:
I'm going to merge this now because, technically, it is a backwards compatible change.

PR #331, opened 2018-09-13
k8s: more generic kubectlRunner so we can use it for more things, test better

maiamcc:
Hello @nicks, @landism,

Please review the following commits I made in branch maiamcc/more-generic-kubectlRunner:

930c223ccb35c2223f06dc330fc8e09abdf6f84d (2018-09-13 16:09:46 -0400)
k8s: more generic kubectlRunner so we can use it for more things, test better

### DIFF
--- a/internal/k8s/client.go
+++ b/internal/k8s/client.go
@@ -1,6 +1,7 @@
 package k8s
 
 import (
+	"bytes"
 	"context"
 	"fmt"
 	"net/url"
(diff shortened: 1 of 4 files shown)


internal/k8s/client.go
```diff
@@ -122,3 +123,15 @@ func (k KubectlClient) Delete(ctx context.Context, entities []K8sEntity) error {
...
 }
+
+func (k KubectlClient) applyOrDeleteFromEntities(ctx context.Context, cmd string, entities []K8sEntity) (stdout string, stderr string, err error) {
+	args := []string{cmd, "-f", "-"}
+
+	rawYAML, err := SerializeYAML(entities)
+	if err != nil {
+		return "", "", fmt.Errorf("serializeYaml for kubectl %s: %v", cmd, err)
+	}
+	stdin := bytes.NewReader([]byte(rawYAML))
```
landism: it might be a bit cleaner to change `execWithStdin` to take `stdin` as a `string` directly (or if we're concerned about not having all the input at once, an `io.Reader`)

PR #332, opened 2018-09-13
output: Summary is concise

yuindustries:
Also include some other minor spacing fixes for the overall output

screen shot 2018-09-13 at 4 10 07 pm


internal/k8s/entity.go
```diff
@@ -83,3 +83,14 @@ func ToLoadBalancer(entity K8sEntity) (LoadBalancer, bool) {
 
 	return result, true
 }
+
+// Get Deployment name
+func GetDeploymentName(entity K8sEntity) string {
```
yuindustries: It didn't seem like the deployment name was available elsewhere so I copied and modified the `ToLoadBalancer` function to do my bidding. Happy to do something else if I'm missing somewhere else where this data already exists.

internal/summary/summary.go
```diff
@@ -70,47 +75,38 @@ func (s *Summary) Gather(services []model.Service) error {
...
+		indent := " "
+		ret += fmt.Sprintf("%s%s — ", indent, svc.Name)
 
-		ret += fmt.Sprintln("    KUBERNETES INFO")
-		if len(k.Version) > 0 {
-			ret += fmt.Sprintf("      • Version: %s\n", k.Version)
+		// Relative Path
+		wd, err := os.Getwd()
+		if err != nil {
+			log.Fatalf("Failed to get working directory: %s", err)
```
yuindustries: Not sure if I should toss away these errors instead? Is this too tedious?

internal/output/outputter.go
```diff
@@ -124,6 +116,12 @@ func (o *Outputter) StartBuildStep(format string, a ...interface{}) {
 	o.curBuildStep++
 }
 
+func (o *Outputter) Summary(format string, a ...interface{}) {
+	o.logger.Infof("%s", o.blue().Sprint("──┤ Status ├──────────────────────────────────────────────────────────"))
```
yuindustries: Per our discussion about how it's confusing to show all services when not all of them have updated...I changed the summary section title to "Status." In my mind that helps a little bit, but I'm not sure if this is a distinction that makes sense to others.

internal/k8s/entity.go
nicks: ya, this is a bit confusing. in all our yaml files, the service name and the deployment name are the same, but there's no guarantee the user is doing this

let me write a quick function to get the Name of an entity, which is tricky, and i think will help with this

the 'deploy' part of 'kubectl get deploy' should be in `entity.GroupVersionKind.Kind`

internal/engine/image_build_and_deployer.go
```diff
@@ -36,6 +36,10 @@ func NewImageBuildAndDeployer(b build.ImageBuilder, k8sClient k8s.Client, env k8
 func (ibd *ImageBuildAndDeployer) BuildAndDeploy(ctx context.Context, manifest model.Manifest, state BuildState) (br BuildResult, err error) {
 	span, ctx := opentracing.StartSpanFromContext(ctx, "daemon-ImageBuildAndDeployer-BuildAndDeploy")
 	defer span.Finish()
+	defer func() {
+		span.Finish()
+		output.Get(ctx).Printf("") // space between Pipelines
```
maiamcc: nbd but this is a weird logical grouping of things. i'd put the extra printline below, at ln 56 along with the `EndPipeline` call (or better yet, just put that logic in the `EndPipeline` function itself)

internal/output/outputter.go
```diff
@@ -115,7 +107,7 @@ func (o *Outputter) StartPipelineStep(format string, a ...interface{}) {
 func (o *Outputter) EndPipelineStep() {
 	elapsed := time.Now().Sub(o.curPipelineStepStart)
 	o.logger.Infof("    (Done %.3fs)", elapsed.Seconds())
```
maiamcc: ...or just `"    (Done %.3fs)\n"`?

internal/summary/summary.go
maiamcc: if we can't get the working director something is VERY WRONG, this is fine

internal/engine/upper.go
```diff
@@ -110,9 +110,10 @@ func (u Upper) CreateManifests(ctx context.Context, manifests []model.Manifest,
 
 	logger.Get(ctx).Debugf("[timing.py] finished initial build") // hook for timing.py
 
-	output.Get(ctx).Printf("%s", s.Output())
+	output.Get(ctx).Summary("%s", s.Output())
```
maiamcc: any reason not to do `Summary(s.Output())`?

internal/engine/upper.go
```diff
@@ -156,7 +157,8 @@ func (u Upper) CreateManifests(ctx context.Context, manifests []model.Manifest,
 				}
 				logger.Get(ctx).Debugf("[timing.py] finished build from file change") // hook for timing.py
 
-				output.Get(ctx).Printf("%s", s.Output())
+				output.Get(ctx).Summary("%s", s.Output())
```
maiamcc: same

internal/output/outputter.go
```diff
@@ -63,7 +63,7 @@ func (o *Outputter) green() *color.Color  { return o.color(color.FgGreen) }
 func (o *Outputter) red() *color.Color    { return o.color(color.FgRed) }
 
 func (o *Outputter) StartPipeline(totalStepCount int) {
-	o.logger.Infof("%s", o.blue().Sprint("──┤ Pipeline Starting … ├────────────────────────────────────────"))
+	o.logger.Infof("%s", o.blue().Sprint("──┤ Pipeline Starting… ├──────────────────────────────────────────────"))
```
maiamcc: believe you can just do `o.blue.Println("──┤ Pipeline Starting........."`

internal/output/outputter.go
```diff
@@ -124,6 +116,12 @@ func (o *Outputter) StartBuildStep(format string, a ...interface{}) {
 	o.curBuildStep++
 }
 
+func (o *Outputter) Summary(format string, a ...interface{}) {
+	o.logger.Infof("%s", o.blue().Sprint("──┤ Status ├──────────────────────────────────────────────────────────"))
+	o.logger.Infof(format, a...)
+	o.logger.Infof("%s", o.blue().Sprint("╴╴╴╴╴╴╴╴╴╴╴╴╴╴╴╴╴╴╴╴╴╴╴╴╴╴╴╴╴╴╴╴╴╴╴╴╴╴╴╴╴╴╴╴╴╴╴╴╴╴╴╴╴╴╴╴╴╴╴╴╴╴╴╴╴╴╴╴╴╴"))
```
maiamcc: again, you can `Println` or `Printf` directly from `o.blue()`

internal/summary/summary.go
```diff
@@ -70,47 +69,38 @@ func (s *Summary) Gather(services []model.Manifest) error {
...
+		rel, err := filepath.Rel(wd, svc.Path)
+		if err != nil {
+			log.Fatalf("Failed to get relative path: %s", err)
 		}
+		ret += fmt.Sprintf("./%s ", rel)
 
+		// K8s info
+		k := svc.k8sData
+		if len(k.Name) > 0 {
+			ret += fmt.Sprintf("→ `kubectl get deploy %s` ", k.Name)
```
maiamcc: ...i don't understand what's helpful about this string?

yuindustries: You mean, why do we say the whole "kubectl get deploy..."? Dan's idea is that this is a hint to our users that they can find out more about the deployed K8s service using that one-liner.

maiamcc:
Okay. Fwiw this isn't immediately apparent to me as a user, seems
potentially more confusing than helpful. we should workshop this UI.

PR #335, opened 2018-09-13
tiltfile: error if local_git_repo isn't a git repo

jazzdan:


jazzdan:
I'm going to add a test for the error case.

nicks:
will servantes break if you commit this?

jazzdan:
I think so, yes. A pre-req for merging this will be https://app.clubhouse.io/windmill/story/110/selectively-mounting-files-in-a-container (so obviously I won't until that's in)

nicks:
ya, i think that's a blocker before we can consider this

jazzdan:
I rolled this in to a different PR.

PR #336, opened 2018-09-13
k8s/buildAndDeploy: PollForPodWithImage (and invoke from PostProcessBuilds) [ch134]

maiamcc:
this keeps us from having to `sleep` for an arbitrary amount of time before trying to talk to pods

### DIFF
--- a/internal/engine/synclet_build_and_deployer.go
+++ b/internal/engine/synclet_build_and_deployer.go
@@ -14,6 +14,8 @@ import (
 	"google.golang.org/grpc"
 )
 
+const podPollTimeoutSynclet = time.Second * 30
+
 var _ BuildAndDeployer = &SyncletBuildAndDeployer{}
 
 type SyncletBuildAndDeployer struct {
--- a/internal/engine/synclet_build_and_deployer.go
+++ b/internal/engine/synclet_build_and_deployer.go
@@ -136,7 +136,9 @@ func (sbd *SyncletBuildAndDeployer) getContainerForBuild(ctx context.Context, bu
 	defer span.Finish()
 
 	// get pod running the image we just deployed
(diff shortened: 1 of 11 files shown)


internal/k8s/pod.go
```diff
@@ -4,46 +4,62 @@ import (
...
 )
 
-// PodWithImage returns the ID of the pod running the given image. We expect exactly one
-// matching pod: if too many or too few matches, throw an error.
+func (k KubectlClient) PollForPodWithImage(ctx context.Context, image reference.NamedTagged, timeout time.Duration) (PodID, error) {
+	span, ctx := opentracing.StartSpanFromContext(ctx, "kubectlClient-PollForPodWithImage")
+	span.SetTag("img", image.String())
+	defer span.Finish()
+
+	deadline := time.Now().Add(timeout)
```
nicks: slightly more idiomatic as

```
start := time.Now()
for time.Since(start) < deadline {
}
```

internal/engine/synclet_build_and_deployer.go
```diff
@@ -134,7 +136,9 @@ func (sbd *SyncletBuildAndDeployer) getContainerForBuild(ctx context.Context, bu
 	defer span.Finish()
 
 	// get pod running the image we just deployed
-	pID, err := sbd.kCli.PodWithImage(ctx, build.Image)
+	// TODO(maia): parallelize this polling (inefficient b/c first we deploy all manifests in series,
+	// then we poll for pods for all of them (again in series)
+	pID, err := sbd.kCli.PollForPodWithImage(ctx, build.Image, time.Second*45)
```
nicks: i remain skeptical of any api that blocks the current goroutine...45 seconds isn't great.

there are lots of "normal" user stories where the pod might never come up, like if they made a mistake in the yaml, or if the server exits immediately due to a bug in their code

PR #337, opened 2018-09-14
internal: mounts have multiple paths

jazzdan:
This is step 1 towards being able to selectively mount individual files. I'm implementing the syntax that @nicks  outlined [here](https://app.clubhouse.io/windmill/story/110/selectively-mounting-files-in-a-container#activity-111). I realized that on the backend we'll need to be able to handle multiple local files being mapped in to the same container mount point. I decided to make that change first.

This PR should not affect the functionality of tilt in any way, it's just to lay the groundwork.

### DIFF
--- a/internal/build/container_test.go
+++ b/internal/build/container_test.go
@@ -25,7 +25,7 @@ func TestMount(t *testing.T) {
 	f.WriteFile("sup", "my name is dan")
 
 	m := model.Mount{
-		Repo:          model.LocalGithubRepo{LocalPath: f.Path()},
+		Repo:          model.LocalGithubRepo{LocalPaths: []string{f.Path()}},
 		ContainerPath: "/src",
 	}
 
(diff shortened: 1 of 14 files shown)


jazzdan:
hmm, this might not be necessary. Playing around with some stuff.

internal/build/container_test.go
```diff
@@ -25,7 +25,7 @@ func TestMount(t *testing.T) {
 	f.WriteFile("sup", "my name is dan")
 
 	m := model.Mount{
-		Repo:          model.LocalGithubRepo{LocalPath: f.Path()},
+		Repo:          model.LocalGithubRepo{LocalPaths: []string{f.Path()}},
```
nicks: this implies to me that multiple local paths of the git repo get mounted to the same container path, and i can't think of any reason why that would be true

it sounds like you have some uncertainty about this as well. would it help to write up a short spec? just weigh the different options. don't spend more than an hour or two on it, and write it more for yourself than for us.

jazzdan: That is what this implies, but that's not how mounts work today:

https://github.com/windmilleng/tilt/pull/337/files#diff-3b4d21189aa2f4f5e13c1af0ee7839e6R106

We take the relpath of the local file and join it with the container path to get the fully qualified path. So this:

```python
mount = image.mount('/go/src/github.com/servantes')
mount.add('package.json')
``` 

becomes `/go/src/github.com/servantes/package.json`.

Unless you meant that "multiple local paths of a git repo get mounted in to the same directory", in which case, yes that's true but that seems desirable to me?

jazzdan: But yes I am unsure about it, playing around with things now.

jazzdan:
closign this, I think I have a better way

PR #338, opened 2018-09-14
buildAndDeploy: store deployInfo on BuildAndDeployer (rather than on build state)

maiamcc:
in preparation for rewiring PostProcessBuilds to be concurrent [ch209]

### DIFF
--- a/internal/engine/local_container_build_and_deployer.go
+++ b/internal/engine/local_container_build_and_deployer.go
@@ -22,14 +22,17 @@ type LocalContainerBuildAndDeployer struct {
 	env       k8s.Env
 	k8sClient k8s.Client
 	analytics analytics.Analytics
+
+	deployInfo map[model.ManifestName]k8s.ContainerID
 }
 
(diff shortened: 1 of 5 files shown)


internal/engine/local_container_build_and_deployer.go
```diff
@@ -95,20 +96,23 @@ func (cbd *LocalContainerBuildAndDeployer) PostProcessBuilds(ctx context.Context
...
-			state.LastResult.Container = cID
-			states[serv] = state
+	for name, state := range states {
+		cbd.postProcessBuild(ctx, name, state)
+	}
+}
+
+func (cbd *LocalContainerBuildAndDeployer) postProcessBuild(ctx context.Context, name model.ManifestName, state BuildState) {
+	if !state.LastResult.HasImage() {
+		logger.Get(ctx).Infof("can't get container for for '%s': BuildResult has no image", name)
```
nicks: for for?

also, a good trick is to use %q for safe quoting in fmt strings

maiamcc:
uhhh CI is down and tests pass locally 😬 🤞

PR #342, opened 2018-09-14
tilt: add port forwarding to kube client

landism:
mostly copied from tesseract

1. renamed KubectlClient to KubeClient, since we're now sometimes going through client-go instead of kubectl
2. I spent a fair amount of time using a fake client-go to get a unit test working for this, but it was a lot of code and complexity, and didn't really make me any more confident about whether any of the code was working, so I left that out. This seems more like a thing that needs an integration test, which isn't a problem that I think is worth blocking on.


internal/k8s/client.go
```diff
@@ -48,28 +52,45 @@ type Client interface {
...
+	ForwardPort(ctx context.Context, namespace string, podID PodID, remotePort int) (localPort int, closer func(), err error)
 }
 
-type KubectlClient struct {
+type K8sClient struct {
 	env           Env
 	kubectlRunner kubectlRunner
+	restClient    k8sRestInterface
+	restConfig    *rest.Config
+	portForwarder PortForwarder
```
maiamcc: what's the logic behind making this a func you attach to the struct rather than a method on the interface/struct?

internal/k8s/portforward.go
```diff
@@ -0,0 +1,112 @@
...
+	}
+
+	errChan := make(chan error)
+	go func() {
+		errChan <- pf.ForwardPorts()
+		err := <-errChan
+		// logging isn't really sufficient, since we're in a goroutine and who knows where the caller
+		// has moved on to by this point, but other options are much more expensive (e.g., monitoring the state
+		// of the port forward from the caller and/or automatically reconnecting port forwards)
+		logger.Get(ctx).Infof("error from port forward: %v", err)
```
maiamcc: if we expect this to be a rare failure and worth tanking the tilt process for, we could pass the context into here and cancel it on error?

internal/k8s/client.go
landism: so that tests can use a fake PortForwarder

internal/k8s/portforward.go
landism: Maybe I overshot the mark with "logging isn't really sufficient". I haven't put much thought or investigation into what sort of errors might turn up here. I'd expect the vast majority of errors here would also tank our synclet clients (e.g., the node gets redeployed with a different node id). Maybe https://app.clubhouse.io/windmill/story/273/make-tunnels-sync-clients-less-fragile-to-connection-errors is the right answer?

PR #343, opened 2018-09-14
tiltfile: selectively mounting files in a container 

jazzdan:
Need to add a lot more tests, this is still a WIP.

There's a lot of terminology in here that needs to be clarified. (what is a baseMount/mountBase? Will fix)

### DIFF
--- a/internal/tiltfile/tiltfile_test.go
+++ b/internal/tiltfile/tiltfile_test.go
@@ -483,3 +483,29 @@ ENTRYPOINT echo hi`)
 	// TODO(dmiller) is this right?
 	assert.Equal(t, []string{"sh", "-c", ""}, manifest.Entrypoint.Argv)
 }
+
+func TestNewMountSyntax(t *testing.T) {
+	dockerfile := tempFile(`FROM alpine
+	ENTRYPOINT echo hi`)
+	file := tempFile(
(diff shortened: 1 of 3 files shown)


internal/tiltfile/tiltfile_test.go
```diff
@@ -483,3 +483,28 @@ ENTRYPOINT echo hi`)
...
 }
+
+func TestNewMountSyntax(t *testing.T) {
+	dockerfile := tempFile(`FROM alpine
+	ENTRYPOINT echo hi`)
+	file := tempFile(
+		fmt.Sprintf(`def blorgly():
+		image = build_docker_image(%q, "docker-tag")
+		repo = local_git_repo('.')
+		mount = image.mount("/app/blorgly")
```
nicks: it seems weird that /app/blorgly is repeated multiple times?

nicks:
ya, i'm concerned that this is becoming a quagmire, and you need to write a spec before going further. let's sync up on this on monday.

nicks:
(by "quagmire", i mean that it seems that what's happening here is that a small syntax change to the tiltfile is triggering a larger change to the data model, which is then triggering more changes to the syntax, which is triggering more changes to the model, and so on. we need to scope this down a bit)

PR #351, opened 2018-09-17
tiltfile: add path() function to gitRepo

jazzdan:


internal/tiltfile/skylarkTypes.go
```diff
@@ -229,3 +230,32 @@ func (gitRepo) Hash() (uint32, error) {
...
+		return nil, nil
+	}
+
+}
+
+func (gr gitRepo) AttrNames() []string {
+	return []string{"path"}
+}
+
+func fullRepoPath(thread *skylark.Thread, fn *skylark.Builtin, args skylark.Tuple, kwargs []skylark.Tuple) (skylark.Value, error) {
```
nicks: hmm...if you make this a method of `gitRepo`, you will get the binding for free, rather than having to bind and unbox?

internal/tiltfile/skylarkTypes.go
```diff
@@ -229,3 +230,32 @@ func (gitRepo) Hash() (uint32, error) {
...
+	r, ok := fn.Receiver().(gitRepo)
+	if !ok {
+		return nil, errors.New("internal error: path called on non-gitRepo")
+	}
+
+	var path skylark.String
```
nicks: i think if you make this a `string`, it will Just Work (tm)?

internal/tiltfile/skylarkTypes.go
```diff
@@ -229,3 +230,32 @@ func (gitRepo) Hash() (uint32, error) {
...
+	err := skylark.UnpackArgs(fn.Name(), args, kwargs, "path", &path)
+	if err != nil {
+		return nil, err
+	}
+
+	return skylark.String(filepath.Join(r.path, string(path))), nil
```
nicks: don't you need some way to carry around the gitignore so you can add it to the FileFilter?

internal/tiltfile/tiltfile_test.go
```diff
@@ -527,3 +528,36 @@ func TestReadFile(t *testing.T) {
...
+	dockerfile := tempFile("docker text")
+	fileToRead := tempFile("hello world")
+	program := fmt.Sprintf(`def blorgly():
+	repo = local_git_repo('.')
+	print(repo.path('subpath'))
+	yaml = read_file(%q)
+	image = build_docker_image("%v", "docker-tag", repo.path('subpath'))
+	return k8s_service(yaml, image)
+`, fileToRead, dockerfile)
+	fmt.Println(program)
```
nicks: hmmmm

internal/tiltfile/skylarkTypes.go
jazzdan: oh! Let me try that.

jazzdan:
@nicks my plan for gitignore was the grab that upon invocation of `local_git_repo` but let me check and make sure that makes sense.

jazzdan:
@nicks actually currently gitignore is [handled completely outside of the Tiltfile](https://github.com/windmilleng/tilt/blob/e2800027fd6aea853cedfa7af24bd81b1d62a770/internal/engine/up_watch.go#L145).

jazzdan:
(I think I will change that, but not in this PR)

PR #356, opened 2018-09-18
tiltfile: img.add requires localPath object

jazzdan:
Not gonna merge this yet asit will cause a segfault if you use it on a subdirectory of a git repo since there's no gitignore, but the code contained here is itself reviewable.

### DIFF
--- a/internal/tiltfile/skylarkTypes.go
+++ b/internal/tiltfile/skylarkTypes.go
@@ -13,7 +13,7 @@ import (
 	"github.com/windmilleng/tilt/internal/model"
 )
 
-const oldMountSyntaxError = "The syntax for `add` has changed. Before it was `.add(dest, src)`. Now it is `.add(src, dest)`."
+const oldMountSyntaxError = "The syntax for `add` has changed. Before it was `.add(dest: string, src: string)`. Now it is `.add(src: localPath, dest: string)`."
 
 type compManifest struct {
 	cManifest []k8sManifest
@@ -71,8 +71,8 @@ func (k8sManifest) Hash() (uint32, error) {
 }
 
 type mount struct {
+	src        localPath
 	mountPoint string
-	repo       gitRepo
 }
 
 type dockerImage struct {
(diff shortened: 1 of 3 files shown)


internal/tiltfile/tiltfile_test.go
```diff
@@ -56,12 +56,12 @@ func TestGetManifestConfig(t *testing.T) {
 	dockerfile := tempFile("docker text")
 	file := tempFile(
 		fmt.Sprintf(`def blorgly():
-  image = build_docker_image("%v", "docker-tag", "the entrypoint")
-  image.add(local_git_repo('.'), '/mount_points/1')
-  print(image.file_name)
-  image.run("go install github.com/windmilleng/blorgly-frontend/server/...")
-  image.run("echo hi")
-  return k8s_service("yaaaaaaaaml", image)
+		image = build_docker_image("%v", "docker-tag", "the entrypoint")
```
jazzdan: My editor was being really weird with these indents

nicks: i'm surprised skylark even accepts this as valid code bc it has tabs

internal/tiltfile/tiltfile_test.go
```diff
@@ -56,12 +56,12 @@ func TestGetManifestConfig(t *testing.T) {
...
+		repo = local_git_repo('.')
+		image.add(repo.path('.'), '/mount_points/1')
```
nicks: hmmm...this doesn't seem right. can you fix the arg deserialization so that image.add(repo, ...) works like it did before?

jazzdan:
@nicks should be fixed!

internal/tiltfile/skylarkTypes.go
```diff
@@ -159,7 +159,17 @@ func addMount(thread *skylark.Thread, fn *skylark.Builtin, args skylark.Tuple, k
 		return nil, errors.New("internal error: add_docker_image_cmd called on non-dockerImage")
 	}
 
-	image.mounts = append(image.mounts, mount{mountPoint, gitRepo})
+	gr, ok := src.(gitRepo)
```
nicks: might be slightly cleaner to do a typeswitch

switch gr := gr.(type) {
  case localPath:
  case gitRepo:
  default:
}

jazzdan: ah yes good call, forgot about those!

PR #359, opened 2018-09-18
k8s: make it easier to resolve the real load balancer url with the k8s client

nicks:
Hello @maiamcc, @yuindustries,

Please review the following commits I made in branch nicks/ch263/summary:

1ef60b72298bc519904653d6fb9e225f1c5c2f48 (2018-09-18 13:35:40 -0400)
k8s: make it easier to resolve the real load balancer url with the k8s client

### DIFF
--- a/internal/k8s/entity.go
+++ b/internal/k8s/entity.go
@@ -1,6 +1,7 @@
 package k8s
 
 import (
+	"net/url"
 	"reflect"
 
 	"k8s.io/api/core/v1"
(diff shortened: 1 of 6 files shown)


internal/summary/summary.go
```diff
@@ -29,9 +31,10 @@ type k8sObject struct {
 }
 
 // NewSummary returns summary state
-func NewSummary() *Summary {
+func NewSummary(client k8s.Client) *Summary {
```
nicks: i'm a little bit nervous about giving Summary a k8sClient versus making it the responsibility of the caller to pass this data in. wdyt?

nicks: on the one hand, it would be nice if the function was "pure" and didn't go look things up from external systems. on the other hand, if our long-term plan is to show a HUD of your cluster state, maybe this is inevitable

internal/k8s/entity.go
```diff
@@ -78,45 +79,50 @@ func ImmutableEntities(entities []K8sEntity) []K8sEntity {
...
 		}
 	}
 	return result
 }
 
-// Try to convert the current entity to a LoadBalancer service
-func ToLoadBalancer(entity K8sEntity) (LoadBalancer, bool) {
+// Try to convert the current entity to a LoadBalancerSpec service
+func ToLoadBalancerSpec(entity K8sEntity) (LoadBalancerSpec, bool) {
 	service, ok := entity.Obj.(*v1.Service)
```
maiamcc: wait, do load balances only exist in v1?

internal/summary/summary.go
maiamcc: :-/ i'm also bummed about this... is it super ugly if you expect the caller to pass the data?

nicks: ok, i tried something different where i pass in a resolver function, which feels less icky. wdyt?

internal/k8s/entity.go
nicks: AFAIK, there's two ways to get traffic from outside the cluster to the pods
- v1.Service
- v1beta.Ingress

my vague understanding is that Ingress is still in flux and only really works on GKE, but i certainly haven't been following the discussions beyond the bare minimum

PR #360, opened 2018-09-18
wmdocker: split docker client code into wmdocker package

maiamcc:
speaking of splitting out responsibilities and packages and stuff — do people agree that Docker code should exist outside of the `build` package (maybe in `wmdocker` package)? now that so much of the updater code exists outside of the build package, and since we need docker for things outside of building (like getting container id), makes sense to me that it should be its own package. but if anyone objects, i don’t have to do it


internal/wmdocker/client_benchmark_test.go
```diff
@@ -0,0 +1,31 @@
+//+build !skipcontainertests
+
+package wmdocker
+
+// func BenchmarkExecInContainer(b *testing.B) {
```
nicks: why did this get commented out? i'm pretty sure we still want this to compile

nicks:
could also call it dockerc or smth (with a hat tip to dockerd)

internal/wmdocker/client_benchmark_test.go
maiamcc: lol whoops

maiamcc: oh right, it's cuz i need to make everything on the DockerBuildFixture public and it was a pita. but it's fine.

maiamcc: ugh nvm, it's a giiiant pain and leads to import cycles. imo the `DockerFixture` should maybe live in `docker/testutils.go` but there's a lotttt of stuff to disentangle. however i assume it would be totally gross to put `client_benchmark_test.go` in the `build` package to avoid import cycles/having to split out the docker fixture?

nicks: oh that's fine just put it in the build package for now

PR #361, opened 2018-09-18
tilt: wire up SyncletBaD to call the Synclet

landism:
with this change, incremental builds work in GKE

### DIFF
--- a/internal/engine/synclet_build_and_deployer.go
+++ b/internal/engine/synclet_build_and_deployer.go
@@ -6,13 +6,14 @@ import (
 	"sync"
 	"time"
 
-	opentracing "github.com/opentracing/opentracing-go"
+	"github.com/docker/distribution/reference"
+	"github.com/pkg/errors"
+
+	"github.com/opentracing/opentracing-go"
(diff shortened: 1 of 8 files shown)


internal/engine/synclet_build_and_deployer.go
```diff
@@ -160,31 +156,61 @@ func (sbd *SyncletBuildAndDeployer) PostProcessBuild(ctx context.Context, manife
...
+	if err != nil {
+		return DeployInfo{}, errors.Wrapf(err, "syncletClient.GetContainerIdForPod (pod = %s)", pID)
+	}
+
+	logger.Get(ctx).Verbosef("talking to synclet client for node %s", nodeID.String())
+
+	return DeployInfo{cID, nodeID}, nil
+}
+
+func getNodeID(ctx context.Context, kCli k8s.Client, image reference.NamedTagged) (k8s.NodeID, error) {
```
maiamcc: wait, we already have the pod ID at the point where this is called...?

landism: whoops, good catch!

PR #362, opened 2018-09-18
tiltfile: read gitignore, associate with docker image, use on backend

jazzdan:
This does three things:

1) Read `.gitignore` and `.Dockerignore`, if present, upon invocation of `local_git_repo`. Store the path matcher on the `gitRepo` object
2) Associate the `gitRepo` to the `dockerImage` so that when we return the manifests we can assign a path matcher to the manifest
3) Use that path matcher in the up_watch code, rather than reading + parsing the ignore files there

Plus various cleanup.

### DIFF
--- a/internal/engine/up_watch.go
+++ b/internal/engine/up_watch.go
@@ -5,7 +5,6 @@ import (
 	"errors"
 	"path/filepath"
 
-	"github.com/windmilleng/tilt/internal/dockerignore"
 	"github.com/windmilleng/tilt/internal/git"
 	"github.com/windmilleng/tilt/internal/model"
 	"github.com/windmilleng/tilt/internal/watch"
@@ -142,28 +141,6 @@ type manifestNotifyPair struct {
 	notify   watch.Notify
 }
 
-func makeFilter(ctx context.Context, manifest model.Manifest) (model.PathMatcher, error) {
-	var repoRoots []string
-
-	for _, mount := range manifest.Mounts {
-		repoRoots = append(repoRoots, mount.LocalPath)
-	}
-
-	mrt, err := git.NewMultiRepoIgnoreTester(ctx, repoRoots)
-	if err != nil {
-		return nil, err
-	}
-
-	dit, err := dockerignore.NewMultiRepoDockerIgnoreTester(repoRoots)
-
-	ci := model.NewCompositeMatcher([]model.PathMatcher{
-		mrt,
-		dit,
-	})
-
-	return ci, nil
-}
-
 // turns a list of (manifest, chan fsevent) pairs into a single chan (manifest, fsevent)
 func snsToManifestWatcher(ctx context.Context, timerMaker timerMaker, sns []manifestNotifyPair) (*manifestWatcher, error) {
 	events := make(chan manifestFilesChangedEvent)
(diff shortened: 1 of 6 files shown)


internal/engine/up_watch.go
```diff
@@ -142,38 +141,16 @@ type manifestNotifyPair struct {
...
 	errs := make(chan error)
 
 	for _, sn := range sns {
 		coalescedEvents := coalesceEvents(timerMaker, sn.notify.Events())
-		filter, err := makeFilter(ctx, sn.manifest)
-		if err != nil {
-			return nil, err
+		filter := sn.manifest.FileFilter
+		if filter == nil {
+			filter = git.FalseIgnoreTester{}
```
jazzdan: This is just for tests that don't instantiate all of a manifest properly.

internal/tiltfile/tiltfile.go
```diff
@@ -107,7 +112,20 @@ func makeSkylarkGitRepo(thread *skylark.Thread, fn *skylark.Builtin, args skylar
 		return nil, fmt.Errorf("Reading path %s: %v", path, err)
 	}
 
-	return gitRepo{absPath}, nil
+	ctx, cancel := context.WithTimeout(context.Background(), 1*time.Second)
+	defer cancel()
+	t1, err := git.NewGitIgnoreTester(ctx, absPath)
```
jazzdan: this will not error if the files do not exist, and instead return a matcher that matches nothing. (Like what happens if there's no `.gitignore` file in regular git.

internal/tiltfile/skylarkTypes.go
```diff
@@ -164,12 +165,13 @@ func addMount(thread *skylark.Thread, fn *skylark.Builtin, args skylark.Tuple, k
...
 		lp = p
 	case gitRepo:
-		lp = localPath{path: p.basePath}
+		lp = localPath{p.basePath, p}
 	default:
 		return nil, fmt.Errorf(oldMountSyntaxError)
 	}
 
 	image.mounts = append(image.mounts, mount{lp, mountPoint})
+	image.repo = lp.repo
```
nicks: what about the case where there's more than one repo?

jazzdan: good call, fixed

internal/tiltfile/tiltfile.go
```diff
@@ -42,13 +46,14 @@ func makeSkylarkDockerImage(thread *skylark.Thread, fn *skylark.Builtin, args sk
...
 
-	return &dockerImage{dockerfileName, tag, []mount{}, []model.Step{}, entrypoint}, nil
+	return &dockerImage{dockerfileName, tag, []mount{}, []model.Step{}, entrypoint, []model.PathMatcher{git.FalseIgnoreTester{}}}, nil
 }
 
 func makeSkylarkK8Manifest(thread *skylark.Thread, fn *skylark.Builtin, args skylark.Tuple, kwargs []skylark.Tuple) (skylark.Value, error) {
 	var yaml skylark.String
 	var dockerImage *dockerImage
-	err := skylark.UnpackArgs(fn.Name(), args, kwargs, "yaml", &yaml, "dockerImage", &dockerImage)
+	var repo gitRepo
```
nicks: this is obsolete now?

internal/tiltfile/skylarkTypes.go
```diff
@@ -76,11 +76,12 @@ type mount struct {
...
-	fileTag    reference.Named
-	mounts     []mount
-	steps      []model.Step
-	entrypoint string
+	fileName     string
+	fileTag      reference.Named
+	mounts       []mount
+	steps        []model.Step
+	entrypoint   string
+	pathMatchers []model.PathMatcher
```
nicks: might be work calling these filters to indicate that they're the files we're ignoring

internal/tiltfile/tiltfile.go
jazzdan: ah yes, thank you

PR #363, opened 2018-09-18
build: factor out ContainerResolver, bc this is about to get more complex

nicks:
Hello @maiamcc, @landism,

Please review the following commits I made in branch nicks/resolver:

dbf6d7b8dbe71615303f3a1b952a6521ebfaf45c (2018-09-18 15:48:33 -0400)
build: factor out ContainerResolver, bc this is about to get more complex

### DIFF
--- a/internal/build/container_resolver.go
+++ b/internal/build/container_resolver.go
@@ -1,0 +1,60 @@
+package build
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/docker/docker/api/types"
+	"github.com/docker/docker/api/types/filters"
+	opentracing "github.com/opentracing/opentracing-go"
+	"github.com/windmilleng/tilt/internal/k8s"
+)
+
+type ContainerResolver struct {
+	dcli DockerClient
+}
+
+func NewContainerResolver(dcli DockerClient) *ContainerResolver {
+	return &ContainerResolver{dcli: dcli}
+}
+
+// containerIdForPod looks for the container ID associated with the pod.
+// Expects to find exactly one matching container -- if not, return error.
+// TODO: support multiple matching container IDs, i.e. restarting multiple containers per pod
+func (r *ContainerResolver) ContainerIDForPod(ctx context.Context, podName k8s.PodID) (k8s.ContainerID, error) {
+	span, ctx := opentracing.StartSpanFromContext(ctx, "ContainerResolver-containerIdForPod")
+	defer span.Finish()
+
+	a := filters.NewArgs()
+	a.Add("name", string(podName))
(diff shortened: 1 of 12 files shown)


nicks:
(the main motivation for this is because i was going back and forth between implementing this with a docker client vs a kubectl client; there are advantages and disadvantages to each, and really this seems like a separable thing)

internal/build/container_resolver.go
```diff
@@ -0,0 +1,60 @@
...
+
+func NewContainerResolver(dcli DockerClient) *ContainerResolver {
+	return &ContainerResolver{dcli: dcli}
+}
+
+// containerIdForPod looks for the container ID associated with the pod.
+// Expects to find exactly one matching container -- if not, return error.
+// TODO: support multiple matching container IDs, i.e. restarting multiple containers per pod
+func (r *ContainerResolver) ContainerIDForPod(ctx context.Context, podName k8s.PodID) (k8s.ContainerID, error) {
+	span, ctx := opentracing.StartSpanFromContext(ctx, "daemon-containerIdForPod")
```
landism: if I understand our span naming correctly, "ContainerResolver-containerIDForPod"?

internal/synclet/synclet.go
```diff
@@ -19,47 +17,18 @@ const Port = 23551
...
 
-func NewSynclet(dcli build.DockerClient) *Synclet {
-	return &Synclet{dcli: dcli}
+func NewSynclet(dcli build.DockerClient, cr *build.ContainerResolver) *Synclet {
+	return &Synclet{dcli: dcli, cr: cr}
 }
 
 const pauseCmd = "/pause"
 
 // TODO(matt) dedupe with ContainerUpdater - https://app.clubhouse.io/windmill/story/227/de-dupe-containeridforpod
```
landism: can nuke the TODO, too

nicks: done

internal/build/container_resolver.go
nicks: done

internal/build/container_resolver.go
```diff
@@ -0,0 +1,60 @@
+package build
```
maiamcc: this would be a great thing to NOT lump in the `build` package, imo -- maybe a separate `containers` package? or the `docker` package? but it's not reallllly the domain of `build` imo.

nicks:
ya, it's an RPC in synclet build and deployer.

per the slack discussion, i think a totally reasonable abstraction here would be for SyncletClient to implement a ContainerResolver interface

PR #365, opened 2018-09-18
tilt: deploy_synclet: log output on failure

landism:
FWIW all the junk between try: and except: is just indentation. I'm kind of surprised github doesn't render that as a whitespace change.


landism:
> FWIW all the junk between try: and except: is just indentation. I'm kind of surprised github doesn't render that as a whitespace change.

craaaazy: https://github.com/windmilleng/tilt/pull/365/files?utf8=%E2%9C%93&diff=split&w=1

PR #368, opened 2018-09-19
tilt: tiltify synclet

landism:


synclet/Dockerfile
```diff
@@ -3,22 +3,6 @@ FROM iron/go:dev as builder
```
nicks: fwiw, i've liked the convention of renaming these to Dockerfile.base to distinguish them from "real" dockerfiles.

I think we'll want to keep the original dockerfile around so that we can publish to a registry?

PR #373, opened 2018-09-19
tilt: stream synclet logs back to client

landism:


internal/synclet/synclet.proto
```diff
@@ -30,6 +34,24 @@ message UpdateContainerRequest {
...
+    DEBUG = 2;
+}
+
+message LogMessage {
+    LogLevel level = 1;
+    bytes message = 2;
+}
+
+message UpdateContainerReply {
+    LogMessage logMessage = 1;
```
nicks: protobuf style is slug_case for field names
https://developers.google.com/protocol-buffers/docs/style

internal/synclet/synclet.proto
```diff
@@ -30,6 +34,24 @@ message UpdateContainerRequest {
...
 }
 
-message UpdateContainerReply {}
+message Context {
+    bool logColorsEnabled = 1;
```
nicks: ditto

internal/synclet/synclet.proto
```diff
@@ -30,6 +34,24 @@ message UpdateContainerRequest {
...
+    LogLevel level = 1;
+    bytes message = 2;
+}
+
+message UpdateContainerReply {
+    LogMessage logMessage = 1;
 }
 
-message UpdateContainerReply {}
+message Context {
```
nicks: Context as a name seems very likely to collide with things. LogStyle?

internal/synclet/synclet.proto
```diff
@@ -5,20 +5,24 @@ package synclet;
...
+    rpc GetContainerIdForPod (GetContainerIdForPodRequest) returns (stream GetContainerIdForPodReply) {}
 
     // updates the specified container and then restarts it
     // (much functionality packed into one rpc to minimize latency)
-    rpc UpdateContainer (UpdateContainerRequest) returns (UpdateContainerReply) {}
+    rpc UpdateContainer (UpdateContainerRequest) returns (stream UpdateContainerReply) {}
 }
 
 message GetContainerIdForPodRequest {
     string pod_id = 1;
```
nicks: not _really_ related to this pr, but is there a reason you split this up into two RPCs, rather than putting the pod_id in the UpdateContainer RPC?

internal/synclet/server_adapter.go
```diff
@@ -21,30 +23,84 @@ func NewGRPCServer(del *Synclet) *GRPCServer {
...
-func (s *GRPCServer) UpdateContainer(ctx context.Context, req *proto.UpdateContainerRequest) (*proto.UpdateContainerReply, error) {
+func (s *GRPCServer) UpdateContainer(req *proto.UpdateContainerRequest, server proto.Synclet_UpdateContainerServer) error {
 	var commands []model.Cmd
 	for _, cmd := range req.Commands {
 		commands = append(commands, model.Cmd{Argv: cmd.Argv})
 	}
-	return &proto.UpdateContainerReply{}, s.del.UpdateContainer(ctx, k8s.ContainerID(req.ContainerId), req.TarArchive, req.FilesToDelete, commands)
+
+	send := func(m *proto.LogMessage) error {
+		return server.Send(&proto.UpdateContainerReply{LogMessage: m})
```
nicks: i think you need some synchronization here? loggers are usually concurrency-safe, but server.Send is not
https://groups.google.com/forum/#!topic/grpc-io/aI6L6M4fzQ0

ditto for the other server.Send

internal/synclet/client_adapter.go
```diff
@@ -41,26 +66,69 @@ func (s *SyncletCli) UpdateContainer(
...
+	stream, err := s.del.UpdateContainer(ctx, &proto.UpdateContainerRequest{
+		Context:       newProtoContext(ctx),
 		ContainerId:   containerId.String(),
 		TarArchive:    tarArchive,
 		FilesToDelete: filesToDelete,
 		Commands:      protoCmds,
 	})
 
+	if err != nil {
+		return err
```
nicks: can you add context to this error with fmt.Errorf?

internal/synclet/client_adapter.go
```diff
@@ -41,26 +66,69 @@ func (s *SyncletCli) UpdateContainer(
...
+	}
+
+	for {
+		reply, err := stream.Recv()
+
+		if err == io.EOF {
+			return nil
+		} else if err != nil {
+			return err
```
nicks: ditto

internal/synclet/client_adapter.go
```diff
@@ -41,26 +66,69 @@ func (s *SyncletCli) UpdateContainer(
...
+		}
+
+		level, err := protoLogLevelToLevel(reply.LogMessage.Level)
+		if err != nil {
+			return err
```
nicks: ditto

internal/synclet/client_adapter.go
```diff
@@ -28,6 +33,26 @@ func NewGRPCClient(conn *grpc.ClientConn) *SyncletCli {
...
+	var level logger.Level
+	switch protoLevel {
+	case proto.LogLevel_INFO:
+		level = logger.InfoLvl
+	case proto.LogLevel_VERBOSE:
+		level = logger.VerboseLvl
+	case proto.LogLevel_DEBUG:
+		level = logger.DebugLvl
+	default:
+		return logger.InfoLvl, fmt.Errorf("unknown log level '%v'", protoLevel)
```
nicks: hmmm...not sure this is a good idea, because it makes the protocol non-backwards compatible

internal/synclet/client_adapter.go
```diff
@@ -28,6 +33,26 @@ func NewGRPCClient(conn *grpc.ClientConn) *SyncletCli {
 	return &SyncletCli{del: proto.NewSyncletClient(conn), conn: conn}
 }
 
+func protoLogLevelToLevel(protoLevel proto.LogLevel) (logger.Level, error) {
+	var level logger.Level
+	switch protoLevel {
+	case proto.LogLevel_INFO:
+		level = logger.InfoLvl
```
nicks: simpler to return directly?

internal/synclet/synclet.proto
landism: I don't recall a reason, and looking at the code now, I don't see a reason we need the container ID in the client. Maybe the code has changed, or maybe we were just paralleling the way the local container builds worked? AFAICT, the only thing we use the container ID for is to pass it back into UpdateContainer. I don't want to block this PR on that. I'll double-check with @maiamcc when she's back.

landism: ah, this lets us start paying the cost of finding the containerID in the background as soon as the first build finishes, rather than synchronously before the second build starts

PR #375, opened 2018-09-20
k8s: capture namespaces and show them in summary

nicks:
Hello @maiamcc,

Please review the following commits I made in branch nicks/ch298/namespaces:

1ac632ffd405f6d408da38e952ee8d0fcdbc587b (2018-09-20 13:58:42 -0400)
k8s: capture namespaces and show them in summary

### DIFF
--- a/internal/summary/summary.go
+++ b/internal/summary/summary.go
@@ -26,8 +26,9 @@ type Service struct {
 }
 
 type k8sObject struct {
-	Kind string
-	Name string
+	Kind      string
+	Name      string
+	Namespace string
 }
 
 // NewSummary returns summary state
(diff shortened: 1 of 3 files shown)


internal/summary/summary.go
```diff
@@ -91,14 +93,25 @@ func (s *Summary) Output(ctx context.Context, resolver LBResolver) string {
...
-			ret += fmt.Sprintf("→ `kubectl get svc %s` ", svc.K8sLbs[0].Name)
+			lb := svc.K8sLbs[0]
+			namespaceFlag := ""
+			if lb.Namespace != "" {
+				namespaceFlag = fmt.Sprintf(" -n %s", lb.Namespace)
+			}
+
+			ret += fmt.Sprintf("→ `kubectl get svc %s%s` ", lb.Name, namespaceFlag)
 
 			url := resolver(ctx, svc.K8sLbs[0])
```
maiamcc: ...seems like the majority of this code can live in the same conditional, with this big being tacked on the end only if there exists a load balancer, no?

nicks: not really, because it's parsing the data from different struct types

PR #376, opened 2018-09-20
tiltfile: disallow ./ and / in repo localPaths

jazzdan:


landism:
This feels kind of weird to me. What's the problem this is trying to solve?
Aside from anything else, I'd have expected the `filepath.Join` to remove the `./`.

jazzdan:
This was a @dbentley feature request. I think `./` are confusing, and `/` should be disallowed? I'll let him chime in though.

dbentley:
The problem is confusion. Why use '.' here? It's never needed to use "./" in this context because the '.' is different from the '.' where it's written. And trying to use '..' is very confusing (as discussed with @jazzdan ; do you want to update this PR to do the clean thing)

landism:
I don't want to block merging this; I just don't understand how this is different than arguing `cd` should ban paths starting with `./`, since it's redundant there, as well.

jazzdan:
This also seems to be the case for reading files:

```
[/tmp]> touch '-a file'
touch: invalid option -- ' '
Try 'touch --help' for more information.
[/tmp]> touch ./'-a file'
[/tmp]> cat '-a file'
cat: invalid option -- 'a'
Try 'cat --help' for more information.
[/tmp]> cat ./'-a file'
```

landism:
> I did find one use case for "./" after doing some research:

IIUC, that's unrelated to this PR, since this PR is just about specifying directory names, where "./" doesn't really enable new behavior.

FWIW, if you want some annoying but sometimes useful trivia: the, uh, proper? way to make a file named `-a` is `touch -- -a`. `--` generally communicates something like "ok the rest of the args are just positional args, so don't treat them as flags". Also useful in, e.g., `kubectl exec $POD -- ls -la`, to ensure `-la` is an arg to `ls` and not `kubectl`.

(it's an old prank to make a file named `-rf` so that someone tries to delete it)

jazzdan:
I agree that this seems unnecessary, and in the spirit of writing the least amount of code possible I won't merge it. Happy to revisit this though!

PR #378, opened 2018-09-20
vendor: fixing bad state

maiamcc:
if folks are concerned about this i can dig in deeper, but my and matt's explorations with zipkin put my deps in a weird state, so here's me fixing them


maiamcc:
nvm i'm a goof.

PR #379, opened 2018-09-21
build: edits to Tiltfile shouldn't cause docker cache miss

jazzdan:
Story details: https://app.clubhouse.io/windmill/story/222

### DIFF
--- a/internal/build/tar.go
+++ b/internal/build/tar.go
@@ -103,6 +103,7 @@ func (a *ArchiveBuilder) tarPath(ctx context.Context, source, dest string) error
 	dest = strings.TrimPrefix(dest, "/")
 
 	err = filepath.Walk(source, func(path string, info os.FileInfo, err error) error {
+		// TODO(dmiller): filter out Tiltfile here
 		if err != nil {
 			return fmt.Errorf("error walking to %s: %v", path, err)
 		}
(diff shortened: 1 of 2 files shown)


internal/build/tar_test.go
```diff
@@ -77,6 +77,35 @@ func TestLen(t *testing.T) {
...
+		},
+	}
+
+	err := ab.ArchivePathsIfExist(f.ctx, paths)
+	if err != nil {
+		f.t.Fatal(err)
+	}
+	actual := tar.NewReader(ab.buf)
+	f.assertFileInTar(actual, expectedFile{Path: "a", Contents: "a"})
+	f.assertFileInTar(actual, expectedFile{Path: "Tiltfile", Missing: true})
```
nicks: this won't do what you expect, because assertFileInTar has already consumed the tarball, you need to use the multi-file version

internal/build/tar.go
```diff
@@ -103,10 +111,15 @@ func (a *ArchiveBuilder) tarPath(ctx context.Context, source, dest string) error
 	dest = strings.TrimPrefix(dest, "/")
 
 	err = filepath.Walk(source, func(path string, info os.FileInfo, err error) error {
+		// TODO(dmiller): filter out Tiltfile here
```
maiamcc: isn't this... done?

PR #380, opened 2018-09-21
engine: postProcessBuild gets new info if image has chnaged

maiamcc:
Having some trouble writing tests with our current testing infra :-/ if anyone has suggestions for a simple way to test, I'm all ears.

### DIFF
--- a/internal/engine/local_container_build_and_deployer.go
+++ b/internal/engine/local_container_build_and_deployer.go
@@ -6,6 +6,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/docker/distribution/reference"
 	"github.com/opentracing/opentracing-go"
 	"github.com/windmilleng/tilt/internal/build"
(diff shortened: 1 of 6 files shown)


internal/engine/local_container_build_and_deployer.go
```diff
@@ -25,7 +26,7 @@ type LocalContainerBuildAndDeployer struct {
 	k8sClient k8s.Client
 	analytics analytics.Analytics
 
-	deployInfo   map[model.ManifestName]k8s.ContainerID
+	deployInfo   map[reference.NamedTagged]k8s.ContainerID
```
nicks: i don't think we want to use a NamedTagged as a map key. 

there are multiple internal representations that satisfy this interface. Go will use the internal representation as the key, not the NamedTagged part of it.

we probably want to use a custom struct with a Name and a Tag field

maiamcc: good call, will do

PR #383, opened 2018-09-21
tiltfile: use an implicit build context for the docker image

jazzdan:
Rather than passing around the image explicitly.

### DIFF
--- a/internal/tiltfile/tiltfile.go
+++ b/internal/tiltfile/tiltfile.go
@@ -20,6 +20,7 @@ import (
 )
 
 const FileName = "Tiltfile"
+const buildContextKey = "buildContext"
 
 type Tiltfile struct {
 	globals  skylark.StringDict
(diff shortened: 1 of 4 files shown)


internal/tiltfile/tiltfile.go
```diff
@@ -167,6 +170,13 @@ func readFile(thread *skylark.Thread, fn *skylark.Builtin, args skylark.Tuple, k
 	return skylark.String(dat), nil
 }
 
+func stopBuild(thread *skylark.Thread, fn *skylark.Builtin, args skylark.Tuple, kwargs []skylark.Tuple) (skylark.Value, error) {
+	r := buildContext
+	defer func() { buildContext = nil }()
```
jazzdan: Want to call y'all's attention to this part. Is this a safe thing to do?

jazzdan: (this doesn't need to be a defer)

internal/tiltfile/tiltfile.go
```diff
@@ -27,6 +27,8 @@ type Tiltfile struct {
 	thread   *skylark.Thread
 }
 
+var buildContext *dockerImage
```
jazzdan: make this thread local

internal/tiltfile/tiltfile_test.go
```diff
@@ -694,3 +705,69 @@ func TestReadsIgnoreFiles(t *testing.T) {
...
+	}
+	_, err = tiltconfig.GetManifestConfigs("blorgly")
+	if err == nil {
+		t.Fatal("Expected GetManifestConfigs to error, but it didn't")
+	}
+	expected := "error running 'blorgly': run called without a build context"
+	if err.Error() != expected {
+		t.Errorf("Expected %s to equal %s", err.Error(), expected)
+	}
+}
```
jazzdan: test for starting two build contexts without finishing one

jazzdan:
(got a verbal LGTM from @dbentley, modulo those comments I just added)

jazzdan:
I'm going to wait until Monday to merge this.

PR #389, opened 2018-09-21
tilt: switch to public opentracer endpoint, allow tracing from synclet

landism:
a lot of the diff is reorganizing the startup code so that:
1. we've parsed the args before we initialize the tracer (to allow --trace)
2. the same function is responsible for initializing and cleaning up the tracer (mostly motivated by broken ^C: when we canceled the context, we'd close the trace collector and back up out of the stack, and calls to `span.Finish` would hang)

Sample trace here: [here](https://ui.honeycomb.io/windmill/datasets/traces/result/8NSQj2RMLn3/trace/2e6d2cf534fecadc)


Makefile
```diff
@@ -31,9 +31,6 @@ benchmark:
 errcheck:
 	errcheck -ignoretests -ignoregenerated ./...
 
-start_tracer:
```
maiamcc: is there no longer any setup code for the tracer? don't I need something running locally? are there setup steps if we're running this for the first time in a new cluster?

internal/tracer/tracer.go
```diff
@@ -1,30 +1,50 @@
...
+}
+
+func (zl zipkinLogger) Log(keyvals ...interface{}) error {
+	logger.Get(zl.ctx).Debugf("%v", keyvals)
+	return nil
+}
+
+var _ zipkin.Logger = zipkinLogger{}
+
+func Init(ctx context.Context) (func() error, error) {
```
maiamcc: this is dark sorcery, thanks for doing this!

Makefile
landism: No setup - it all goes to the globally public opentracing.windmill.build, regardless of running locally or in whatever cluster.

I've pushed its k8s config to https://github.com/windmilleng/honeycomb-opentracing-proxy, which might not be the best place for it.

landism: I should say: I wanted to keep it opt-in, so users aren't unwittingly sending us their data, so `--trace` is now required if you want a trace.

PR #396, opened 2018-09-24
buildAndDeploy: block on PostProcessBuild

maiamcc:
container deploys block on PostProcessBuild [ch334] **WIP**

**WIP** still need tests and to do this for LocalContainerBuildAndDeployer but wanted to make sure I'm on the right track.

don't quite know my way around concurrency-land, left some Q's in comments inline (look for `~~~`).

In particular @nicks I now lock `sbd.deployInfoMu` when retrieving deploy info's off the struct, but the map value is a pointer, so i don't really write back to the map as much as just modify the value behind the pointer... (i lock a mutex on the `*DeployInfo` itself before I edit it but not before I read from it.) will this explode anything?

### DIFF
--- a/internal/engine/synclet_build_and_deployer.go
+++ b/internal/engine/synclet_build_and_deployer.go
@@ -26,35 +26,85 @@ type SyncletBuildAndDeployer struct {
 
 	kCli k8s.Client
 
-	deployInfo   map[docker.ImgNameAndTag]DeployInfo
+	deployInfo   map[docker.ImgNameAndTag]*DeployInfo
 	deployInfoMu sync.Mutex
 }
 
 type DeployInfo struct {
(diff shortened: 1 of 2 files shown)


nicks:
ya, this isn't quite right, but there are some standard patterns on how to do this, let me get you a reference impl

nicks:
take a look at how skylark does this here:
https://github.com/google/skylark/blob/master/example_test.go#L185

basically you have a map protected by a mutex
each entry in the map has a ready channel
the first time you create an entry in the map, you kick off a process that does the fetching, then closes the ready channel when it's done
if the entry is already in the map, you wait on the ready channel

i can walk you thru it in more detail if that would help

maiamcc:
Resolved in #404

PR #397, opened 2018-09-24
k8s: can set PollForPod delay on fake client

maiamcc:
we'll want this for #396

f30b2a80aaec1de21e9a42dc9996aadc9cb28c91 (2018-09-24 17:46:52 -0400)
k8s: can set PollForPod delay on fake client

### DIFF
--- a/internal/k8s/fake_client.go
+++ b/internal/k8s/fake_client.go
@@ -64,7 +64,7 @@ func (c *FakeK8sClient) SetPollForPodWithImageDelay(dur time.Duration) {
 }
 
 func (c *FakeK8sClient) PollForPodWithImage(ctx context.Context, image reference.NamedTagged, timeout time.Duration) (*v1.Pod, error) {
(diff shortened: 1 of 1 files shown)


internal/k8s/fake_client.go
```diff
@@ -58,7 +59,19 @@ func (c *FakeK8sClient) PodWithImage(ctx context.Context, image reference.NamedT
 	}, nil
 }
 
+func (c *FakeK8sClient) SetPollForPodWithImageDelay(dur time.Duration) {
+	c.PollForPodWithImageDelay = dur
+}
+
 func (c *FakeK8sClient) PollForPodWithImage(ctx context.Context, image reference.NamedTagged, timeout time.Duration) (*v1.Pod, error) {
+	defer func() { c.PollForPodWithImageDelay = 0 }()
```
nicks: defer SetPollForPodWithImageDelay?

maiamcc: yesss that's better!

PR #399, opened 2018-09-24
k8s: add a function for injecting sidecars

nicks:
Hello @landism, @maiamcc,

Please review the following commits I made in branch nicks/sidecar:

5be67f5b820efbbad6c837701d49aba9b198eacd (2018-09-24 18:25:26 -0400)
k8s: add a function for injecting sidecars

### DIFF
--- a/internal/synclet/sidecar/sidecar.go
+++ b/internal/synclet/sidecar/sidecar.go
@@ -1,0 +1,41 @@
+package sidecar
+
+import (
+	"fmt"
+
+	"k8s.io/api/core/v1"
+)
+
+func syncletPrivileged() *bool {
+	val := true
+	return &val
+}
+
+// When we deploy Tilt, we override this with LDFLAGS
+const SyncletTag = "latest"
+
+const SyncletImageName = "gcr.io/windmill-public-containers/tilt-synclet"
+
+var SyncletContainer = v1.Container{
+	Name:            "tilt-synclet",
+	Image:           fmt.Sprintf("%s:%s", SyncletImageName, SyncletTag),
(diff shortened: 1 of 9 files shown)


internal/synclet/sidecar/sidecar.go
```diff
@@ -0,0 +1,38 @@
+package sidecar
+
+import (
+	"fmt"
+
+	"k8s.io/api/core/v1"
+)
+
+var syncletPrivileged = true
```
landism: We don't seem to set this anywhere yet, and I don't know how we're going to use this, so I don't know what the motivations behind making this var are, but can this be a const or a func param?

internal/k8s/synclet.go
```diff
@@ -0,0 +1,35 @@
+package k8s
+
+import (
+	"github.com/docker/distribution/reference"
+	"github.com/windmilleng/tilt/internal/synclet/sidecar"
+)
+
+func InjectSyncletSidecar(entity K8sEntity, matchRef reference.Named) (K8sEntity, bool, error) {
```
landism: FFTI, but I'd have a slight preference for putting this in the sidecar package, since it feels kinda weird for k8s to know about synclets, and the sidecar package already has a k8s dep anyway (and is kind of explicitly about mucking with k8s?). (Or, if we're keeping this here, it still seems like maybe the synclet entity defs should be next to this function?)

nicks: done!

internal/synclet/sidecar/sidecar.go
nicks: done!

nicks:
yup, that's the plan

PR #400, opened 2018-09-25
synclet: deploy as sidecar

nicks:
Hello @maiamcc, @landism,

Please review the following commits I made in branch nicks/deploy:

7e58168ae48fee3adee5eed50b76c54d5724dfc9 (2018-09-25 11:56:16 -0400)
synclet: deploy as sidecar

### DIFF
--- a/internal/engine/image_build_and_deployer.go
+++ b/internal/engine/image_build_and_deployer.go
@@ -11,6 +11,7 @@ import (
 	"github.com/windmilleng/tilt/internal/k8s"
 	"github.com/windmilleng/tilt/internal/model"
 	"github.com/windmilleng/tilt/internal/output"
+	"github.com/windmilleng/tilt/internal/synclet/sidecar"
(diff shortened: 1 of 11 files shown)


internal/engine/image_build_and_deployer.go
```diff
@@ -155,7 +162,17 @@ func (ibd *ImageBuildAndDeployer) deploy(ctx context.Context, manifest model.Man
 		}
 		if replaced {
 			didReplace = true
+
+			if ibd.injectSynclet {
```
maiamcc: maybe a comment on why we run this code here? i'm confused.

internal/engine/synclet_build_and_deployer.go
```diff
@@ -31,15 +31,16 @@ type SyncletBuildAndDeployer struct {
 }
 
 type DeployInfo struct {
+	podID       k8s.PodID
```
maiamcc: pretty sure if we're storing synclet clients by podID we no longer need NodeID

PR #401, opened 2018-09-25
readme: Make README more legible

jazzdan:
Two outstanding TODOs that I'd like feedback on:

1. Should we explain how `start_fast_build` is fast, and what image and container builds are?
2. We should probably explain the concept of the "active" build (if you call `start_fast_build` before calling `stop_build` there will be an error) but I'm not sure the best way to do that.

In the future we should probably use Sphinx, like we did in windmilleng/windmill.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,12 +4,14 @@
 [![GoDoc](https://godoc.org/github.com/windmilleng/tilt?status.svg)](https://godoc.org/github.com/windmilleng/tilt)
 
 ## Installing
+
 Run `go get -u github.com/windmilleng/tilt`
 
 ## Using Tilt
 `tilt up <service_name>` starts a service once; `tilt up --watch <service_name>` starts it and watches for changes.
 
 Tilt reads from a Tiltfile. A simple Tiltfile is below:
+
 ```python
 def backend():
   img = static_build('Dockerfile', 'gcr.io/companyname/backend')
@@ -39,9 +41,11 @@ def backend():
 ```
 
 ## Mill
+
 written in a Mill, a dialect of python. It's based on [starlark](https://github.com/bazelbuild/starlark), using the implementation in [go](https://github.com/google/skylark).
 
 ### Mill Builtins
+
 Mill comes with built-in functions.
 
 #### static_build(dockerfile, ref, context?)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -22,88 +25,125 @@ def backend():
...
 
-* Args:
-  * `dockerfile_path`: **str**
-  * `img_name`: **str**, e.g. blorgdev/backend or gcr.io/project-name/bucket-name
-  * `entrypoint?`: **str**
-* Returns: **Image**
+```python
+class Repo:
+  def path(path: str) -> localPath:
+    """Returns the absolute path to the file specified at path in the repo.
```
maiamcc: imo `path` should be in backticks in this docstring

README.md
```diff
@@ -22,88 +25,125 @@ def backend():
...
+    Returns:
+      A localPath resource, representing a local path on disk.
+    """
+```
+
+#### start_fast_build
 
-#### add(src, dest)
-Adds the content from `src` into the image at path `dest`. Paths must be relative.
+Initiates a docker image build that uses a cache and supports `add`s and `run`s.
```
maiamcc: Maybe:
```
Initiates a docker image build that supports `add`s and `run`s, and that uses a cache for subsequent builds.
```

README.md
```diff
@@ -22,88 +25,125 @@ def backend():
...
+```python
+def add(src: Union[localPath, Repo], dest: str) -> None
+```
+
+#### run
 
-#### run(cmd, trigger?)
 Runs `cmd` as a build step in the image.
-If the `trigger` file is specified, the build step is only run if the file is changed. Path must be relative.
+If the `trigger` file is specified, the build step is only run if the file is changed.
```
maiamcc: > If the `trigger` arg is specified (a file or list of files), the build step is only on on changes to the given file(s).

README.md
```diff
@@ -22,88 +25,125 @@ def backend():
...
 Creates a composite service; tilt will deploy (and watch) all services returned by the functions in `service_fns`.
 
-* Args:
-  * `service_fns`: array of functions that each return **Service**
-* Returns: **Service**
+```python
+def composite_service(List[Callable[[], Service]]) -> Service
+```
+
+#### Service
```
maiamcc: maybe put this def above `composite_service`?

maiamcc: also aren't we no longe calling these "services"?

README.md
```diff
@@ -22,88 +25,125 @@ def backend():
...
+#### read_file
 
-#### read_file(file_path)
 Reads file and returns its contents.
 
-* Args:
-  * `file_path`: **str**
-* Returns: **str**
+```python
+def read_file(file_path: st) -> str
```
maiamcc: s/st/str

README.md
jazzdan: Yeah it wasn't clear to me whether that was an internal rebranding our an external one as well. Happy to go through and make that change in a subsequent PR.

PR #403, opened 2018-09-25
engine: propagate cancellation thru the watch loop

nicks:
Hello @maiamcc,

Please review the following commits I made in branch nicks/ch332/errors:

c9952782f8d835b6b13b8243e415da09c16522ef (2018-09-25 14:18:25 -0400)
engine: propagate cancellation thru the watch loop

### DIFF
--- a/internal/engine/build_and_deployer.go
+++ b/internal/engine/build_and_deployer.go
@@ -3,6 +3,7 @@ package engine
 import (
 	"context"
 
+	"github.com/pkg/errors"
 	"github.com/windmilleng/tilt/internal/build"
 	"github.com/windmilleng/tilt/internal/k8s"
 	"github.com/windmilleng/tilt/internal/logger"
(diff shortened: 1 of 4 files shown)


internal/engine/build_and_deployer.go
```diff
@@ -66,15 +67,28 @@ func (composite *CompositeBuildAndDeployer) BuildAndDeploy(ctx context.Context,
 	return BuildResult{}, lastErr
 }
 
+func isPermanentError(err error) bool {
+	if _, ok := err.(*model.ValidateErr); ok {
+		return true
+	}
+
+	cause := errors.Cause(err)
```
maiamcc: is this robust against fmt.errorf-ing on top of a context.Canceled error?

internal/engine/build_and_deployer.go
```diff
@@ -66,15 +67,28 @@ func (composite *CompositeBuildAndDeployer) BuildAndDeploy(ctx context.Context,
 	return BuildResult{}, lastErr
 }
 
+func isPermanentError(err error) bool {
```
maiamcc: how are you defining "permanent error" here?

internal/engine/upper.go
```diff
@@ -92,6 +92,8 @@ func (u Upper) CreateManifests(ctx context.Context, manifests []model.Manifest,
 		if err == nil {
 			buildStates[manifest.Name] = NewBuildState(buildResult)
 			lbs = append(lbs, k8s.ToLoadBalancerSpecs(buildResult.Entities)...)
+		} else if isPermanentError(err) {
```
maiamcc: does this error get surfaced to the user in a useful way?

internal/engine/upper.go
```diff
@@ -150,6 +152,9 @@ func (u Upper) CreateManifests(ctx context.Context, manifests []model.Manifest,
 					event.manifest,
 					buildState)
 				if err != nil {
+					if isPermanentError(err) {
```
maiamcc: does this error get surfaced to the user in a useful way?

internal/engine/build_and_deployer.go
nicks: no, that's why we use errors.Wrap! :raised_hands:

internal/engine/build_and_deployer.go
nicks: added a comment!

internal/engine/upper.go
nicks: yes

internal/engine/upper.go
nicks: yes

nicks:
screenshot from 2018-09-25 15-38-10

PR #404, opened 2018-09-25
container deploys block on PostProcessBuild [ch334] WIP

maiamcc:
ok let's try this again

(wip, tests/implementation for d4m still outstanding)

### DIFF
--- a/internal/engine/build_and_deployer_test.go
+++ b/internal/engine/build_and_deployer_test.go
@@ -10,6 +10,7 @@ import (
 	"path/filepath"
 	"strings"
 	"testing"
+	"time"
 
 	"github.com/docker/docker/api/types"
 	"github.com/stretchr/testify/assert"
@@ -118,7 +119,6 @@ func TestIncrementalBuild(t *testing.T) {
 	if f.docker.BuildCount != 0 {
 		t.Errorf("Expected no docker build, actual: %d", f.docker.BuildCount)
 	}
-
 	if f.docker.PushCount != 0 {
 		t.Errorf("Expected no push to docker, actual: %d", f.docker.PushCount)
 	}
@@ -130,6 +130,32 @@ func TestIncrementalBuild(t *testing.T) {
 	}
 	if len(f.docker.RestartsByContainer) != 1 {
 		t.Errorf("Expected 1 container to be restarted, actual: %d", len(f.docker.RestartsByContainer))
+	}
+}
+
+func TestIncrementalBuildWaitsForPostProcess(t *testing.T) {
+	f := newBDFixture(t, k8s.EnvGKE)
+	defer f.TearDown()
+
+	f.k8s.SetPollForPodWithImageDelay(time.Second)
+	go f.bd.PostProcessBuild(f.ctx, alreadyBuilt) // will take 1s
+	time.Sleep(time.Millisecond * 100)            // let the PostProcessBuild call actually start
+
+	_, err := f.bd.BuildAndDeploy(f.ctx, SanchoManifest, NewBuildState(alreadyBuilt))
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Rather than falling back to image build b/c of lack of deploy info, we should
+	// wait for the concurrent call to PostProcessBuild to finish and do a container build.
+	if f.docker.BuildCount != 0 {
+		t.Errorf("Expected no docker build, actual: %d", f.docker.BuildCount)
+	}
+	if f.docker.PushCount != 0 {
+		t.Errorf("Expected no push to docker, actual: %d", f.docker.PushCount)
+	}
+	if f.sCli.UpdateContainerCount != 1 {
+		t.Errorf("Expected 1 UpdateContaine count via synclet, actual: %d", f.sCli.UpdateContainerCount)
 	}
 }
 
@@ -345,6 +371,7 @@ type bdFixture struct {
 	ctx    context.Context
 	docker *docker.FakeDockerClient
 	k8s    *k8s.FakeK8sClient
+	sCli   *synclet.FakeSyncletClient
 	bd     BuildAndDeployer
 }
 
@@ -376,7 +403,8 @@ func newBDFixtureHelper(t *testing.T, env k8s.Env, fallbackFn FallbackTester) *b
 	}
 	ctx := output.CtxForTest()
 	k8s := k8s.NewFakeK8sClient()
-	bd, err := provideBuildAndDeployer(output.CtxForTest(), docker, k8s, dir, env, synclet.NewFakeSyncletClient(), fallbackFn)
+	sCli := synclet.NewFakeSyncletClient()
+	bd, err := provideBuildAndDeployer(output.CtxForTest(), docker, k8s, dir, env, sCli, fallbackFn)
 	if err != nil {
 		t.Fatal(err)
 	}
@@ -386,6 +414,7 @@ func newBDFixtureHelper(t *testing.T, env k8s.Env, fallbackFn FallbackTester) *b
 		ctx:            ctx,
 		docker:         docker,
 		k8s:            k8s,
+		sCli:           sCli,
 		bd:             bd,
 	}
 }
--- a/internal/engine/synclet_build_and_deployer.go
+++ b/internal/engine/synclet_build_and_deployer.go
@@ -26,35 +26,58 @@ type SyncletBuildAndDeployer struct {
 
 	kCli k8s.Client
 
-	deployInfo   map[docker.ImgNameAndTag]DeployInfo
+	deployInfo   map[docker.ImgNameAndTag]*DeployInfo
 	deployInfoMu sync.Mutex
 }
 
 type DeployInfo struct {
 	containerID k8s.ContainerID
 	nodeID      k8s.NodeID
+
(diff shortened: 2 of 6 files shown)


internal/engine/synclet_build_and_deployer.go
```diff
@@ -26,36 +26,59 @@ type SyncletBuildAndDeployer struct {
...
 	return deployInfo, ok
 }
 
-func (sbd *SyncletBuildAndDeployer) setDeployInfoForImage(img reference.NamedTagged, deployInfo DeployInfo) {
-	sbd.deployInfoMu.Lock()
-	key := docker.ToImgNameAndTag(img)
-	sbd.deployInfo[key] = deployInfo
-	sbd.deployInfoMu.Unlock()
+func (sbd *SyncletBuildAndDeployer) deployInfoForImageOrNew(img reference.NamedTagged) (*DeployInfo, bool) {
+	deployInfo, ok := sbd.deployInfoForImage(img)
```
nicks: let's inline the map access, and make a single Lock()/Unlock() surround the entire function. right now this has race conditions

internal/engine/synclet_build_and_deployer.go
```diff
@@ -26,36 +26,59 @@ type SyncletBuildAndDeployer struct {
...
 type DeployInfo struct {
 	podID       k8s.PodID
 	containerID k8s.ContainerID
 	nodeID      k8s.NodeID
+
+	ready chan struct{} // Close this channel when the DeployInfo is populated
+}
+
+// ~~~ should these verify that the channel is non-nil?
+func (di *DeployInfo) waitUntilReady() { <-di.ready }
```
nicks: in practice, you'll want to do something like
select {
  case <-di.ready:
  case <-ctx.Done():
}

internal/engine/synclet_build_and_deployer.go
```diff
@@ -161,28 +186,37 @@ func (sbd *SyncletBuildAndDeployer) PostProcessBuild(ctx context.Context, result
...
 	}
 }
 
-func (sbd *SyncletBuildAndDeployer) getDeployInfo(ctx context.Context, image reference.NamedTagged) (DeployInfo, error) {
-	span, ctx := opentracing.StartSpanFromContext(ctx, "SyncletBuildAndDeployer-getDeployInfo")
+func (sbd *SyncletBuildAndDeployer) populateDeployInfo(ctx context.Context, image reference.NamedTagged, info *DeployInfo) error {
+	span, ctx := opentracing.StartSpanFromContext(ctx, "SyncletBuildAndDeployer-populateDeployInfo")
 	defer span.Finish()
 
+	if info == nil {
```
nicks: i don't understand what this is for? when would you pass in a nil?

internal/engine/synclet_build_and_deployer.go
```diff
@@ -161,28 +186,37 @@ func (sbd *SyncletBuildAndDeployer) PostProcessBuild(ctx context.Context, result
...
+		info = newEmptyDeployInfo()
+	}
+	defer info.markReady()
```
landism: If we hit an error here, are we marking an unpopulated `deployInfo` as ready? (and then will that stay in the map forever?)

internal/engine/synclet_build_and_deployer.go
maiamcc: will do.

do i need to verify that this channel is non-nil before acting on it? or can i assume that this method is only getting called on a properly-formed `di`?

internal/engine/synclet_build_and_deployer.go
maiamcc: thiiis is a habit i have from the last time i wrote go of always checking whether pointers were nil before doing anything to them because one day someone was going to invoke this function from somewhere new and pass bad data and break stuff. seems like you think it's not worth it, and i can just assume the func is called in a reasonable way?

internal/engine/synclet_build_and_deployer.go
maiamcc: yyyyes we do. good catch.

internal/engine/synclet_build_and_deployer.go
```diff
@@ -161,28 +196,37 @@ func (sbd *SyncletBuildAndDeployer) PostProcessBuild(ctx context.Context, result
...
 	}
 }
 
-func (sbd *SyncletBuildAndDeployer) getDeployInfo(ctx context.Context, image reference.NamedTagged) (DeployInfo, error) {
-	span, ctx := opentracing.StartSpanFromContext(ctx, "SyncletBuildAndDeployer-getDeployInfo")
+func (sbd *SyncletBuildAndDeployer) populateDeployInfo(ctx context.Context, image reference.NamedTagged, info *DeployInfo) (err error) {
+	span, ctx := opentracing.StartSpanFromContext(ctx, "SyncletBuildAndDeployer-populateDeployInfo")
 	defer span.Finish()
 
+	defer func() {
```
nicks: ya, this is fine

internal/engine/synclet_build_and_deployer.go
nicks: this is what `panic` is for~

internal/engine/synclet_build_and_deployer.go
nicks: nope! don't worry about that

internal/engine/synclet_build_and_deployer.go
```diff
@@ -89,8 +118,13 @@ func (sbd *SyncletBuildAndDeployer) canSyncletBuild(ctx context.Context,
 	}
 
 	// Can't do container update if we don't know what container manifest is running in.
-	if _, ok := sbd.getDeployInfoForImage(state.LastResult.Image); !ok {
-		return fmt.Errorf("no container info for this manifest")
+	info, ok := sbd.deployInfoForImageBlocking(ctx, state.LastResult.Image)
+	if !ok || info == nil {
+		return fmt.Errorf("have not yet fetched deploy for this manifest")
```
landism: This seems like a weird error to return. What do you think about populating and blocking instead of erroring here (or probably inside `deployInfoForImageBlocking`)?

internal/engine/synclet_build_and_deployer.go
```diff
@@ -126,10 +160,11 @@ func (sbd *SyncletBuildAndDeployer) updateViaSynclet(ctx context.Context,
 	// TODO(maia): can refactor MissingLocalPaths to just return ContainerPaths?
 	containerPathsToRm := build.PathMappingsToContainerPaths(toRemove)
 
-	deployInfo, ok := sbd.getDeployInfoForImage(state.LastResult.Image)
-	if !ok {
+	deployInfo, ok := sbd.deployInfoForImageBlocking(ctx, state.LastResult.Image)
+	if !ok || deployInfo == nil {
```
landism: Is it possible for ok to be true and deployInfo to be nil, or does that mean we have a bug somewhere? If the former, that seems weird. If the latter, I might prefer to leave the `|| deployInfo == nil` off, so that we get a panic instead of conflating one bug (we're putting nils in there somehow) with another (we tried to get the container id and failed)

internal/engine/wire.go
```diff
@@ -10,12 +10,12 @@ import (
 	"github.com/windmilleng/tilt/internal/build"
 	"github.com/windmilleng/tilt/internal/docker"
 	"github.com/windmilleng/tilt/internal/k8s"
-	"github.com/windmilleng/tilt/internal/synclet"
+	synclet "github.com/windmilleng/tilt/internal/synclet"
```
landism: Am I missing something, or is the addition of "synclet" here a noop?

internal/engine/synclet_build_and_deployer.go
maiamcc: thaaaat seems reasonable -- will leave a todo

internal/engine/synclet_build_and_deployer.go
maiamcc: i think the latter, good call

internal/engine/synclet_build_and_deployer.go
maiamcc: legit -- don't wanna gum up this PR too much but i left a todo / [made a ticket](https://app.clubhouse.io/windmill/story/376/if-we-try-to-build-via-synclet-and-don-t-have-deployinfo-fetch-it)

internal/engine/build_and_deployer_test.go
```diff
@@ -133,6 +133,32 @@ func TestIncrementalBuild(t *testing.T) {
 	}
 }
 
+func TestIncrementalBuildWaitsForPostProcess(t *testing.T) {
+	f := newBDFixture(t, k8s.EnvGKE)
+	defer f.TearDown()
+
+	f.k8s.SetPollForPodWithImageDelay(time.Second)
+	go f.bd.PostProcessBuild(f.ctx, alreadyBuilt) // will take 1s
+	time.Sleep(time.Millisecond * 100)            // let the PostProcessBuild call actually start
```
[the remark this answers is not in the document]
nicks: add a channel somewhere so that you get notified when PostProcessBuild has started querying the pod?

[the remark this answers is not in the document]
maiamcc: niiice! will do in a separate PR

PR #405, opened 2018-09-25
up: wait and os.Exit on sigint/sigterm in case we don't handle context cancel

maiamcc:
in my last PR i was waiting on something w/o checking for context.Cancelled and so i wanted to cover my ass

### DIFF
--- a/internal/cli/cli.go
+++ b/internal/cli/cli.go
@@ -7,6 +7,7 @@ import (
 	"os"
 	"os/signal"
 	"syscall"
+	"time"
 
 	"github.com/windmilleng/tilt/internal/output"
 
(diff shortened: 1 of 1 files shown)


internal/cli/cli.go
```diff
@@ -83,9 +84,13 @@ func preCommand(ctx context.Context) (context.Context, func() error) {
 	go func() {
 		_ = <-sigs
 
-		// We rely on context cancellation being handled elsewhere --
-		// otherwise there's no way to SIGINT/SIGTERM this app o_0
 		cancel()
+
+		// Ideally, cancelling the context should stop the app, but in case we failed
+		// to handle it somewhere, wait a bit, then forcibly exit.
+		time.Sleep(time.Second * 2)
```
nicks: we shouldn't do this, this will make the app hang. can you be more specific about what problem you're trying to solve? e.g., what's the repro case?

maiamcc: wait really? tested just fine for me -- this is happening off in a goroutine somewhere and shouldn't block the app from exiting, no?

i'm trying to guard against someone writing long-running/blocking code that doesn't properly check context and therefore can't be ctrl-c'd.

nicks: how about doing something like:
select {
  case <-sigs:
   log.Println("force quitting...")
   os.Exit(1)
  case <-time.After(2 * time.Second):
   os.Exit(1)
}

maiamcc:
@nicks @landism how about this?

PR #409, opened 2018-09-26
build: always print stdout when --verbose is on

nicks:

Please review the following commits I made in branch nicks/ch348/output:

b9985d08ac0090e24006c86b644248605b4eb00a (2018-09-26 12:31:59 -0400)
build: always print stdout when --verbose is on


nicks:
actually, this isn't quite right, lemme jam on this a bit more

PR #410, opened 2018-09-26
synclet: add polling/retries to synclet client creation, get container id [ch373]

maiamcc:
i'm not positive that the client_adaptor is the right place to put this polling code... but sometimes we are able to connect to the synclet but it's not yet up and running in such a way that it can actually accept calls to `ContainerIdForPod`.

### DIFF
--- a/internal/engine/sidecar_synclet_manager.go
+++ b/internal/engine/sidecar_synclet_manager.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"sync"
+	"time"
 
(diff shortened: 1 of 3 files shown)


maiamcc:
just went and checked, and if we have polling in place for connecting to the client but NOT for `synclet.ContainerIdForPod`, we fail with:

> failed to get deployInfo: syncletClient.GetContainerIdForPod (pod = fortune-857c595d7d-f9qdf): error returned from synclet.GetContainerIdForPod: rpc error: code = Unavailable desc = transport is closing

PR #412, opened 2018-09-26
build: always print stdout when --verbose is on

nicks:

Please review the following commits I made in branch nicks/ch348/output:

b0de2510843020bead209b97d647f031edbe64fa (2018-09-26 13:19:14 -0400)
build: always print stdout when --verbose is on

4179f6aa6d3205d809a17e7f5858e322f7eb5c89 (2018-09-26 13:07:42 -0400)
logger: Add a Level() method b/c we sometimes need different behavior depending on level


nicks:
note that this builds on #411

PR #413, opened 2018-09-26
engine: PostProcessBuild tells callers when they can proceed with exec [381]

maiamcc:
idk about naming of `canResume`, happy for suggestions.

### DIFF
--- a/internal/engine/build_and_deployer.go
+++ b/internal/engine/build_and_deployer.go
@@ -21,7 +21,8 @@ type BuildAndDeployer interface {
 
 	// PostProcessBuild gets any info about the build that we'll need for subsequent builds.
 	// In general, we'll store this info ON the BuildAndDeployer that needs it.
-	PostProcessBuild(ctx context.Context, result BuildResult)
+	// When channel `canResume` is closed, we know that concurrent code can continue to execute.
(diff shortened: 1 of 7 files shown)


nicks:
can you elaborate a bit on what problem you're trying to solve? i couldn't figure it out

maiamcc:
So partly in response to [this test here](https://github.com/windmilleng/tilt/pull/404#discussion_r220612283) but in general it seemed nice to have the ImageBuild not return until we were sufficiently into fetching DeployInfo that any necessary locks were in place.

In particular, in the old world, if an edit event came in while the image build was in progress, it would be processed _immediately_ after the image build finished, before we got far enough into `PostProcessBuild` to even indicate that we were starting to look for information for this deploy, and as a result when processing that edit, we said "whoops no deploy info, fall back to image build."

I suppose that this issue could as easily be solved by: [if we try to build via synclet and don't have deployInfo, fetch it](https://app.clubhouse.io/windmill/story/376/if-we-try-to-build-via-synclet-and-don-t-have-deployinfo-fetch-it). If we do that solution, i don't think we need the change in this PR (except maybe for that test)? If I merge this PR, it's unclear if we'll want to implement that ticket as well...?

maiamcc:
talked with nick offline, tackled way better in #422

PR #416, opened 2018-09-26
engine: fix some bogus tests

nicks:
Hello @maiamcc, @landism,

Please review the following commits I made in branch nicks/skip:

bff7bb485b3031b5bbf11521b313c15b1309f8bd (2018-09-26 15:17:33 -0400)
engine: fix some bogus tests

### DIFF
--- a/internal/engine/build_and_deployer_test.go
+++ b/internal/engine/build_and_deployer_test.go
@@ -5,8 +5,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"io/ioutil"
-	"os"
 	"path/filepath"
 	"strings"
 	"testing"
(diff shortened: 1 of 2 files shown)


internal/engine/build_and_deployer_test.go
```diff
@@ -424,3 +411,11 @@ func (f *bdFixture) withContainerForBuild(build BuildResult) *bdFixture {
 	f.bd.PostProcessBuild(f.ctx, build)
 	return f
 }
+
+func (f *bdFixture) assertContainerRestarts(count int) {
+	expected := map[string]int{}
```
maiamcc: ...i'm confused, why not just `assert.Equal(f.T(), count, f.docker.RestartsByContainer[build.MagicTestContainerID]`?

nicks: i'm trying to avoid the bug where we assert 0 restarts, but some other container id got restarted

nicks:
well, the test wasn't getting run because of the Skip(), and then when i removed the Skip() it was failing...

internal/engine/build_and_deployer_test.go
maiamcc: aah. good catch! (Also, if i came upon this code, i would probably just change it... maybe leave a comment?)

another possibility:
```
if assert.Equal(f.T(), 1, len(f.docker.RestartsByContainer, "these tests should only ever effect a single container") {
    assert.Equal(f.T(), count, f.docker.RestartsByContainer[build.MagicTestContainerID])
}
```
?

PR #420, opened 2018-09-26
tilt: log traceid on startup

landism:
Ideally we'd be dropping the trace id into a URL that'd visualize the trace for you, but I haven't yet found any way to do that with honeycomb.

### DIFF
--- a/internal/cli/up.go
+++ b/internal/cli/up.go
@@ -8,6 +8,8 @@ import (
 	"os"
 	"time"
 
+	"github.com/windmilleng/tilt/internal/logger"
+
 	"github.com/fatih/color"
 	"github.com/opentracing/opentracing-go"
 	"github.com/spf13/cobra"
@@ -52,6 +54,14 @@ func (c *upCmd) run(ctx context.Context, args []string) error {
 	}
 
 	logOutput(fmt.Sprintf("Starting Tilt (built %s)…\n", buildDateStamp()))
+
+	if trace {
+		traceID, err := tracer.TraceID(ctx)
+		if err != nil {
+			return err
+		}
+		logger.Get(ctx).Infof("TraceID: %s", traceID)
+	}
 
 	tf, err := tiltfile.Load(tiltfile.FileName, os.Stdout)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


internal/cli/up.go
```diff
@@ -53,6 +55,14 @@ func (c *upCmd) run(ctx context.Context, args []string) error {
 
 	logOutput(fmt.Sprintf("Starting Tilt (built %s)…\n", buildDateStamp()))
 
+	if trace {
+		traceID, err := tracer.TraceID(ctx)
+		if err != nil {
+			return err
```
maiamcc: is this really cause for us to bail out of the whole program?

internal/cli/up.go
```diff
@@ -53,6 +55,14 @@ func (c *upCmd) run(ctx context.Context, args []string) error {
 
 	logOutput(fmt.Sprintf("Starting Tilt (built %s)…\n", buildDateStamp()))
 
+	if trace {
```
maiamcc: maybe put this in cli.go since it's not `up` specific and if there are ever other tilt commands we'll probs want this logic too

internal/cli/up.go
landism: My logic was: If the user is running with --trace, they (presumably? I guess someone might just run with --trace always) want a trace. If we log the error and continue running, it'd be easy to miss the error and then wonder why the trace isn't showing up in honeycomb. If we bail, the error is impossible to miss and if they didn't actually want the trace that badly, they can rerun without --trace.

I don't feel particularly strongly, though.

maiamcc: aaaah right, i forgot that tracing was opt-in. this makes sense, leave it as-is!

internal/cli/up.go
landism: Yeah, that was where I first tried to put it, but we can't get the trace id there since we haven't made a span yet, and I wasn't sure I wanted to make a span just for that purpose. I figured that's a thing we can worry about if/when we have other commands (there's probably already other stuff there we'll want to generalize anyway).

PR #421, opened 2018-09-26
tiltfile: get the absolute filename

nicks:
Hello @landism,

feb0b52d088037334826d41402a46093a59e3fc9 (2018-09-26 17:15:09 -0400)
tiltfile: get the absolute filename


nicks:
not totally convinced this is the right interface, but it fixes the immediate issue

PR #422, opened 2018-09-26
engine: PostProcessDeploy should only pass control [ch381]

maiamcc:
Hello @landism, @nicks,

Please review the following commits I made in branch maiamcc/better-async:

77efe492438f114c144c86ec071ceb889fb4f9cb (2018-09-26 17:27:03 -0400)
engine: PostProcessDeploy should only pass control back to main thread after it has taken out locks

### DIFF
--- a/internal/engine/build_and_deployer.go
+++ b/internal/engine/build_and_deployer.go
@@ -21,6 +21,7 @@ type BuildAndDeployer interface {
 
 	// PostProcessBuild gets any info about the build that we'll need for subsequent builds.
 	// In general, we'll store this info ON the BuildAndDeployer that needs it.
+	// Each implementation of PostProcessBuild is responsible for executing long-running steps async.
 	PostProcessBuild(ctx context.Context, result BuildResult)
 }
 
@@ -35,6 +36,8 @@ type CompositeBuildAndDeployer struct {
 	builders       BuildOrder
 	shouldFallBack FallbackTester
 }
+
+var _ BuildAndDeployer = &CompositeBuildAndDeployer{}
 
 func DefaultShouldFallBack() FallbackTester {
 	return FallbackTester(shouldImageBuild)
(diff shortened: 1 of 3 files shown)


internal/engine/build_and_deployer_test.go
```diff
@@ -133,25 +133,33 @@ func TestIncrementalBuildWaitsForPostProcess(t *testing.T) {
...
+	res, err := f.bd.BuildAndDeploy(f.ctx, SanchoManifest, BuildStateClean)
 	if err != nil {
 		t.Fatal(err)
 	}
 
-	// Rather than falling back to image build b/c of lack of deploy info, we should
-	// wait for the concurrent call to PostProcessBuild to finish and do a container build.
-	if f.docker.BuildCount != 0 {
-		t.Errorf("Expected no docker build, actual: %d", f.docker.BuildCount)
+	// HACK: if we fail to container build here, the test has already failed. However,
```
nicks: ???

i'm confused by the statement "this test setup doesn't support incremental image build"

TestIncrementalBuildTwiceDeadPod should test this?

internal/engine/build_and_deployer.go
```diff
@@ -21,6 +21,7 @@ type BuildAndDeployer interface {
 
 	// PostProcessBuild gets any info about the build that we'll need for subsequent builds.
 	// In general, we'll store this info ON the BuildAndDeployer that needs it.
+	// Each implementation of PostProcessBuild is responsible for executing long-running steps async.
```
nicks: did you intend to add async in this pr?

nicks: it looks like you tried to add it in one commit, then reverted it in the second commit

maiamcc: ...oh, awkward, yes. i just failed to commit it oops.

internal/engine/build_and_deployer_test.go
maiamcc: AAAAHHHH i figured it out. what it doesn’t support is two image builds in a row (Test...DeadPod is some container builds and then a single image build) since the return val for `fake_client.ImageBuild` is a reader. so we read til the end in the first image build, and then there’s nothing to read in the second. Will fix in a separate PR

internal/engine/build_and_deployer.go
maiamcc: 😳

maiamcc:
alright should be good

PR #424, opened 2018-09-26
tilt: don't call ContainerIDForPod in a nonstop loop

landism:
We were ending up with 1000s of ContainerIDForPod spans, leading to thousands of `[msg backlog too long, disposing spans. count 1]`s getting printed.

And as long as we're here, let's honor `ctx.Done()`.

(also, we probably don't want to try to call the synclet 1000s of times anyway)


internal/synclet/client_adapter.go
```diff
@@ -82,17 +83,26 @@ func (s *SyncletCli) UpdateContainer(
 }
 
 func (s *SyncletCli) ContainerIDForPod(ctx context.Context, podID k8s.PodID, imageID reference.NamedTagged) (cID k8s.ContainerID, err error) {
-	start := time.Now()
-	for time.Since(start) < containerIdTimeout {
+	timeout := time.NewTimer(containerIdTimeout)
```
maiamcc: this uh. seems more complicated than it needs to be? all you need is `time.Sleep(retryDelay)` before calling the func again, and `if ctx.Err() != nil { return "", ctx.Err() }`

landism: My understanding is that the idiomatic way to do things is to never actually Sleep, with pauses in loops always on some form of `select`, so that we're always ready to abort in event of timeout or cancelled context. @nicks does this look right to you?

nicks: ya, tho i think i'd usually be more idiomatic to use time.After()

PR #426, opened 2018-09-27
engine: support static rebuilds

nicks:
Hello @landism, @maiamcc,

Please review the following commits I made in branch nicks/ch390/watch:

4c9968773b98a6ad7d093147c35cd0b2c613338b (2018-09-27 12:00:13 -0400)
engine: support static rebuilds

### DIFF
--- a/internal/engine/upper_test.go
+++ b/internal/engine/upper_test.go
@@ -391,6 +391,35 @@ func TestRebuildWithSpuriousChangedFiles(t *testing.T) {
 	assert.Equal(t, endToken, err)
 }
 
+func TestStaticRebuildWithChangedFiles(t *testing.T) {
+	f := newTestFixture(t)
+	defer f.TearDown()
+	manifest := f.newManifest("foobar", nil)
+	manifest.StaticDockerfile = `FROM golang
+ADD ./ ./
+go build ./...
+`
+	manifest.StaticBuildPath = f.Path()
+
+	endToken := errors.New("my-err-token")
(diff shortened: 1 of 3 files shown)


internal/engine/upper_test.go
```diff
@@ -391,6 +391,36 @@ func TestRebuildWithSpuriousChangedFiles(t *testing.T) {
...
+	go func() {
+		call := <-f.b.calls
+		assert.True(t, call.state.IsEmpty())
+
+		// Simulate a change to main.go
+		mainPath := filepath.Join(f.Path(), "main.go")
+		f.watcher.events <- watch.FileEvent{Path: mainPath}
+
+		// The next build should treat main.go as changed, and build
+		// on the last successful result, from before a.go changed.
```
maiamcc: this comment isn't relevant to this test

maiamcc: ...also am i wrong, or are changed files not really relevant to a static build (except that the watcher tells us that something has changed)?

nicks: updated!

no, static builds still trigger a rebuild when the file changes. i don't think the name "static" build is a good one.

maiamcc: ah sorry, i mean: it's true that a file change triggers a rebuild, but that that information (i.e. the file(s) changed) isn't used anywhere in the build itself -- i.e. `call.state.FilesChanged()` is irrelevant to the building. right?

PR #427, opened 2018-09-27
tilt: clean up synclet clients and tunnels

landism:


internal/engine/build_and_deployer_test.go
```diff
@@ -353,6 +353,23 @@ RUN ["go", "install", "github.com/windmilleng/sancho"]`,
...
+	_, err := f.bd.BuildAndDeploy(f.ctx, SanchoManifest, NewBuildState(alreadyBuilt))
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if f.sCli.ClosedCount != 1 {
+		t.Errorf("Expected 1 synclet client close, actual: %d", f.sCli.ClosedCount)
+	}
+
+=======
```
nicks: ?

internal/engine/build_and_deployer.go
```diff
@@ -50,13 +54,27 @@ func NewCompositeBuildAndDeployer(builders BuildOrder, shouldFallBack FallbackTe
...
+
 func (composite *CompositeBuildAndDeployer) BuildAndDeploy(ctx context.Context, manifest model.Manifest, currentState BuildState) (BuildResult, error) {
 	var lastErr error
 	for _, builder := range composite.builders {
 		br, err := builder.BuildAndDeploy(ctx, manifest, currentState)
 		if err == nil {
 			// TODO(maia): maybe this only needs to be called after certain builds?
 			// I.e. should be called after image build but not after a successful container build?
+			if currentState.LastImage() != nil && br.Image != currentState.LastImage() {
+				composite.forgetImage(ctx, currentState.LastImage())
```
nicks: it seems weird that this is a separate api...? could we make this part of PostProcessBuild?

internal/engine/synclet_build_and_deployer.go
```diff
@@ -87,6 +87,22 @@ func (sbd *SyncletBuildAndDeployer) deployInfoForImageBlocking(ctx context.Conte
 	return deployInfo, ok
 }
 
+func (sbd *SyncletBuildAndDeployer) forgetImage(ctx context.Context, img reference.NamedTagged) error {
+	sbd.deployInfoMu.Lock()
+	defer sbd.deployInfoMu.Unlock()
+
+	imgNameAndTag := docker.ToImgNameAndTag(img)
+
+	deployInfo, ok := sbd.deployInfo[imgNameAndTag]
```
nicks: are we concerned about the cases where an image is deployed more than once, to different pods?

internal/engine/build_and_deployer_test.go
landism: sorry, sloppy merge

internal/engine/build_and_deployer.go
landism: good call

internal/engine/synclet_build_and_deployer.go
landism: I think we should be, but that's also broken independently of this change (i.e., we're already keying podid + synclet client by image)

internal/engine/build_and_deployer.go
```diff
@@ -23,6 +26,8 @@ type BuildAndDeployer interface {
 	// In general, we'll store this info ON the BuildAndDeployer that needs it.
 	// Each implementation of PostProcessBuild is responsible for executing long-running steps async.
 	PostProcessBuild(ctx context.Context, result BuildResult)
+
+	forgetImage(ctx context.Context, image reference.NamedTagged) error
```
maiamcc: any reason this one is private and the rest are public?

internal/engine/build_and_deployer.go
```diff
@@ -50,13 +55,29 @@ func NewCompositeBuildAndDeployer(builders BuildOrder, shouldFallBack FallbackTe
 	}
 }
 
+func (composite *CompositeBuildAndDeployer) forgetImage(ctx context.Context, image reference.NamedTagged) error {
+	for _, builder := range composite.builders {
+		fmt.Printf("forgetting on %T\n", builder)
```
maiamcc: stray print? (/should use logger?)

internal/engine/build_and_deployer.go
maiamcc: ++

internal/engine/sidecar_synclet_manager.go
```diff
@@ -66,6 +84,22 @@ func (ssm SidecarSyncletManager) ClientForPod(ctx context.Context, podID k8s.Pod
 	return client, nil
 }
 
+func (ssm SidecarSyncletManager) ForgetPod(ctx context.Context, podID k8s.PodID) error {
+	ssm.mutex.Lock()
+	defer ssm.mutex.Unlock()
+
+	client, ok := ssm.clients[podID]
+	if !ok {
+		fmt.Printf("no client for pod '%s', so nothing to close!\n", podID)
```
maiamcc: use logger

internal/engine/synclet_build_and_deployer.go
maiamcc: 👀 
uhhh.
well the answer to your question is "no". but maybe that's not the RIGHT answer...

maiamcc:
a lot of `fmt.Printf`'s around that should probably either be deleted or be `logger` calls...

landism:
> a lot of `fmt.Printf`'s around that should probably either be deleted or be `logger` calls...

I think I got all of them in the commit last night. Sorry the sloppy PR.

PR #431, opened 2018-09-27
synclet: wait for container up

nicks:
Hello @landism, @maiamcc,

Please review the following commits I made in branch nicks/ch393/wait:

b5b16534fab7a2ea2ceb6081b2375c95594183d1 (2018-09-27 18:25:07 -0400)
synclet: wait for container up
in theory, this should failover more gracefully if the synclet or pod is
crashlooping or not coming up for some reason

### DIFF
--- a/internal/k8s/container.go
+++ b/internal/k8s/container.go
@@ -3,6 +3,7 @@ package k8s
 import (
 	"context"
 	"fmt"
+	"strings"
 
 	"github.com/docker/distribution/reference"
 	"github.com/pkg/errors"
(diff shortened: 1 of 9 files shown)


internal/k8s/container.go
```diff
@@ -40,32 +41,32 @@ func WaitForContainerReady(ctx context.Context, client Client, pod *v1.Pod, ref
...
-			} else if ready {
-				return nil
+				return "", err
+			} else if cID != "" {
+				return cID, nil
 			}
 		}
 	}
-	return nil
+	return "", nil
```
maiamcc: hmm it feels weird me to return a nil error here? even tho it's a case we'll theoretically never hit...? 🤷‍♀️

internal/k8s/container.go
```diff
@@ -94,3 +95,12 @@ func ContainerMatching(pod *v1.Pod, ref reference.Named) (v1.ContainerStatus, er
 	}
 	return v1.ContainerStatus{}, nil
 }
+
+func ContainerIDFromContainerStatus(status v1.ContainerStatus) (ContainerID, error) {
+	id := status.ContainerID
+	prefix := "docker://"
```
maiamcc: prefix should maybe be a const?

internal/k8s/container.go
```diff
@@ -3,34 +3,35 @@ package k8s
...
 	}
 
 	watch, err := client.WatchPod(ctx, pod)
 	if err != nil {
-		return errors.Wrap(err, "WaitForContainerReady")
+		return "", errors.Wrap(err, "WaitForContainerReady")
 	}
 	defer watch.Stop()
 
 	for true {
```
maiamcc: seems like this func renders [`pollForNewClient`](https://github.com/windmilleng/tilt/blob/1b2f93f2f64d24ca19524c725b30089a7a7b5823/internal/engine/sidecar_synclet_manager.go#L69) obsolete, since we only need a single call to this func and it'll do all the waiting and smart decision making for us?

internal/k8s/container.go
nicks: what do you think it should return?

internal/k8s/container.go
```diff
@@ -40,32 +41,32 @@ func WaitForContainerReady(ctx context.Context, client Client, pod *v1.Pod, ref
...
-	return nil
+	return "", nil
 }
 
-func waitForContainerReadyHelper(pod *v1.Pod, ref reference.Named) (bool, error) {
+func waitForContainerReadyHelper(pod *v1.Pod, ref reference.Named) (ContainerID, error) {
 	cStatus, err := ContainerMatching(pod, ref)
 	if err != nil {
-		return false, errors.Wrap(err, "WaitForContainerReady")
+		return "", errors.Wrap(err, "WaitForContainerReady")
```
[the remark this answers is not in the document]
nicks: done

internal/k8s/container.go
nicks: done

internal/k8s/container.go
nicks: it probably doesn't hurt? there might be some other race conditions around opening connections, etc

PR #437, opened 2018-09-28
k8s: fail if the pod is unschedulable

nicks:
Hello @maiamcc,

Please review the following commits I made in branch nicks/unschedulable:

ccdcd7d189a1af9b985f705912b4a809ef55175e (2018-09-28 14:44:05 -0400)
k8s: fail if the pod is unschedulable


nicks:
yup! i tested it, we get the event

PR #439, opened 2018-09-28
grpc: inc max msg size (and consolidate options)

maiamcc:
hat tip to @jazzdan for the original code in Tesseract, which I just stole ^_^''

[ch425]

### DIFF
--- a/internal/options/grpc.go
+++ b/internal/options/grpc.go
@@ -1,0 +1,40 @@
+package options
+
+import (
+	"github.com/grpc-ecosystem/grpc-opentracing/go/otgrpc"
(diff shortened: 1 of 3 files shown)


internal/options/grpc.go
```diff
@@ -0,0 +1,40 @@
+package options
+
+import (
+	"github.com/grpc-ecosystem/grpc-opentracing/go/otgrpc"
+	"github.com/opentracing/opentracing-go"
+	"google.golang.org/grpc"
+)
+
+// 1024 MB
```
nicks: :world_map:

PR #446, opened 2018-10-01
tilt: add --verbose, --debug flags to synclet

landism:
This only affects logging that happens outside of request handling.

This is mostly just to get logging from the zipkin tracer.


Makefile
```diff
@@ -19,7 +19,7 @@ install:
 	./hide_tbd_warning go install ./...
 
 install-dev:
-	if ! [[ -e "$(SYNCLET_DEV_IMAGE_TAG_FILE)" ]]; then echo "No dev synclet found. Run make synclet-dev."; exit 1; fi
+	@if ! [[ -e "$(SYNCLET_DEV_IMAGE_TAG_FILE)" ]]; then echo "No dev synclet found. Run make synclet-dev."; exit 1; fi
```
landism: `@` so that the command isn't printed so that people don't spend time wondering about the error message on the screen when there's not actually an error

maiamcc: Ah, thanks!

maiamcc: (This uh. Seems like poor design on the part of Make...)

PR #449, opened 2018-10-02
tiltfile: record reads of Dockerfile and Tiltfile

jazzdan:


jazzdan:
@nicks I added a unit test but I don't understand how we were recording a read to the Dockerfile in `TestConfigMatcherWithFastBuild` prior to my change.

PR #450, opened 2018-10-02
engine: if a config file changes, re-execute the tiltfile

jazzdan:
…and do an image build with new manifest

requires #449

### DIFF
--- a/internal/engine/build_and_deployer.go
+++ b/internal/engine/build_and_deployer.go
@@ -2,12 +2,15 @@ package engine
 
 import (
 	"context"
+	"fmt"
+	"os"
 
 	"github.com/pkg/errors"
 	"github.com/windmilleng/tilt/internal/build"
 	"github.com/windmilleng/tilt/internal/k8s"
 	"github.com/windmilleng/tilt/internal/logger"
 	"github.com/windmilleng/tilt/internal/model"
+	"github.com/windmilleng/tilt/internal/tiltfile"
 )
 
 type BuildAndDeployer interface {
@@ -52,6 +55,26 @@ func NewCompositeBuildAndDeployer(builders BuildOrder, shouldFallBack FallbackTe
 
 func (composite *CompositeBuildAndDeployer) BuildAndDeploy(ctx context.Context, manifest model.Manifest, currentState BuildState) (BuildResult, error) {
 	var lastErr error
+	changedConfigFiles := getChangedConfigFiles(currentState.FilesChanged(), manifest)
+	if len(changedConfigFiles) > 0 {
+		logger.Get(ctx).Verbosef("Detected a config file change (%v), re-executing tiltfile and doing an image build", changedConfigFiles)
+		lastBuilder := composite.builders[len(composite.builders)-1]
+		configChangeState := currentState.NewStateWithConfigFilesRemoved(manifest.ConfigMatcher)
+		tf, err := tiltfile.Load(tiltfile.FileName, os.Stdout)
+		if err != nil {
+			return BuildResult{}, err
+		}
+		newManifests, err := tf.GetManifestConfigs(manifest.Name.String())
+		if err != nil {
+			return BuildResult{}, err
+		}
+		// NOTE(dmiller) assume there's only one service that we're rebuilding?
+		if len(newManifests) != 1 {
+			return BuildResult{}, fmt.Errorf("Expected there to be one manifest for name %s, got %d", manifest.Name.String(), len(newManifests))
+		}
+
+		return lastBuilder.BuildAndDeploy(ctx, newManifests[0], configChangeState)
+	}
 	for _, builder := range composite.builders {
 		br, err := builder.BuildAndDeploy(ctx, manifest, currentState)
 		if err == nil {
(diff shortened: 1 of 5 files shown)


internal/engine/build_and_deployer.go
```diff
@@ -52,6 +55,26 @@ func NewCompositeBuildAndDeployer(builders BuildOrder, shouldFallBack FallbackTe
...
+		configChangeState := currentState.NewStateWithConfigFilesRemoved(manifest.ConfigMatcher)
+		tf, err := tiltfile.Load(tiltfile.FileName, os.Stdout)
+		if err != nil {
+			return BuildResult{}, err
+		}
+		newManifests, err := tf.GetManifestConfigs(manifest.Name.String())
+		if err != nil {
+			return BuildResult{}, err
+		}
+		// NOTE(dmiller) assume there's only one service that we're rebuilding?
```
jazzdan: Is this safe to do?

internal/engine/build_and_deployer.go
```diff
@@ -52,6 +55,26 @@ func NewCompositeBuildAndDeployer(builders BuildOrder, shouldFallBack FallbackTe
 
 func (composite *CompositeBuildAndDeployer) BuildAndDeploy(ctx context.Context, manifest model.Manifest, currentState BuildState) (BuildResult, error) {
 	var lastErr error
+	changedConfigFiles := getChangedConfigFiles(currentState.FilesChanged(), manifest)
+	if len(changedConfigFiles) > 0 {
+		logger.Get(ctx).Verbosef("Detected a config file change (%v), re-executing tiltfile and doing an image build", changedConfigFiles)
+		lastBuilder := composite.builders[len(composite.builders)-1]
```
jazzdan: This is intentionally hacky. In the future we'll be smarter about what kind of build we do depending on what has changed.

internal/engine/build_and_deployer.go
```diff
@@ -70,6 +93,21 @@ func (composite *CompositeBuildAndDeployer) BuildAndDeploy(ctx context.Context,
 	return BuildResult{}, lastErr
 }
 
+func getChangedConfigFiles(changedFiles []string, m model.Manifest) []string {
+	crf := []string{}
+	for _, f := range changedFiles {
+		matches, err := m.ConfigMatcher.Matches(f, false)
+		if err != nil {
+			// TODO(dmiller) log
```
jazzdan: This might not be necessary?

jazzdan:
A big TODO is to add tests but I wanted to get eyes on this now to make sure that this is a worthwhile naive thing to merge.

internal/model/manifest.go
```diff
@@ -20,6 +20,7 @@ type Manifest struct {
 	DockerRef  reference.Named
 
 	// Local files read while reading the Tilt configuration.
+	ConfigFiles []string
```
nicks: this seems redundant with ConfigMatcher? it seems like we should only have one

internal/engine/build_and_deployer.go
```diff
@@ -52,6 +55,26 @@ func NewCompositeBuildAndDeployer(builders BuildOrder, shouldFallBack FallbackTe
 
 func (composite *CompositeBuildAndDeployer) BuildAndDeploy(ctx context.Context, manifest model.Manifest, currentState BuildState) (BuildResult, error) {
 	var lastErr error
+	changedConfigFiles := getChangedConfigFiles(currentState.FilesChanged(), manifest)
+	if len(changedConfigFiles) > 0 {
+		logger.Get(ctx).Verbosef("Detected a config file change (%v), re-executing tiltfile and doing an image build", changedConfigFiles)
+		lastBuilder := composite.builders[len(composite.builders)-1]
+		configChangeState := currentState.NewStateWithConfigFilesRemoved(manifest.ConfigMatcher)
```
nicks: this doesn't seem right. it seems like you have to fall back to a BuildStateClean, because you don't know what's changed?

internal/engine/build_and_deployer.go
nicks: i think returning an error here is ok

internal/engine/build_and_deployer.go
```diff
@@ -52,6 +55,26 @@ func NewCompositeBuildAndDeployer(builders BuildOrder, shouldFallBack FallbackTe
...
+		tf, err := tiltfile.Load(tiltfile.FileName, os.Stdout)
+		if err != nil {
+			return BuildResult{}, err
+		}
+		newManifests, err := tf.GetManifestConfigs(manifest.Name.String())
```
nicks: what if we need to watch new files?

jazzdan:
yeah after talking with @nicks I think we need to move this up a level of abstraction in to the upper so that we can change the files that we are watching.

internal/model/manifest.go
jazzdan: We need to know the names of the config files in order to watch them. They could be anywhere on disk, and aren't guaranteed to be in the working directory with the Tiltfile (as is the case with Servantes).

nicks: ya, tho we should probably get rid of the matcher as a separate field then

PR #452, opened 2018-10-02
k8s: break out methods for applying labels

nicks:
Hello @landism, @maiamcc,

Please review the following commits I made in branch nicks/labels:

b466fd3598197ad9b981d991393d1b8250c57807 (2018-10-02 18:16:26 -0400)
k8s: break out methods for applying labels
this is part of a larger change to do a better job selecting the right pods, but
labelling everything tilt deploys seems generally useful for auditing purposes

### DIFF
--- a/internal/k8s/extract.go
+++ b/internal/k8s/extract.go
@@ -5,6 +5,7 @@ import (
 	"reflect"
 
 	"k8s.io/api/core/v1"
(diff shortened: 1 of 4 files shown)


nicks:
(i liked this bit enough in another branch matt sent me that i want to use it in other places)

PR #454, opened 2018-10-03
engine: inject and select labels to reduce the risk of collisions, and recognize that their might be more than one pod

nicks:
Hello @landism, @maiamcc,

Please review the following commits I made in branch nicks/ch436/labels:

45886270bb1a0b3c0f05861f3cc954e93d6fa68a (2018-10-02 20:01:23 -0400)
engine: inject and select labels to reduce the risk of collisions, and recognize that their might be more than one pod

57e0110c610c00d3b2305d52a63feb4204cef0ba (2018-10-02 19:50:59 -0400)
k8s: break out methods for applying labels
this is part of a larger change to do a better job selecting the right pods, but
labelling everything tilt deploys seems generally useful for auditing purposes

### DIFF
--- a/internal/engine/build_and_deployer_test.go
+++ b/internal/engine/build_and_deployer_test.go
@@ -134,7 +134,7 @@ func TestIncrementalBuildWaitsForPostProcess(t *testing.T) {
 	f := newBDFixture(t, k8s.EnvGKE)
 	defer f.TearDown()
 
-	f.k8s.SetPollForPodWithImageDelay(time.Second * 1)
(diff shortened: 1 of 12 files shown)


internal/k8s/testyaml/testyaml.go
```diff
@@ -243,3 +243,19 @@ spec:
           hostPath:
             path: /var/run/docker.sock
 `
+
+// We deliberately create a pod without any labels, to
+// ensure code works without them.
+const LonelyPodYAML = `
```
landism: d'aww

PR #457, opened 2018-10-03
engine: re-execute Tiltfile when config files for manifest change

jazzdan:
Same deal as last time: there are a couple outstanding TODOs and no tests but I wanted to validate the approach.

One question that's not in the code: It seems to me that we don't want to filter out any file events based off of information in the Tiltfile. Does that seem right? We certainly don't want to filter out file events for the dockerfile and tiltfile anymore and I couldn't think of any other situations.

### DIFF
--- a/internal/engine/upper.go
+++ b/internal/engine/upper.go
@@ -136,6 +136,7 @@ func (u Upper) CreateManifests(ctx context.Context, manifests []model.Manifest, 
 				return ctx.Err()
 			case event := <-sw.events:
 				if eventContainsConfigFiles(event) {
+					fmt.Println("Event contains config files")
 					newManifest, err := getNewManifestFromTiltfile(ctx, event.manifest.Name)
 					if err != nil {
 						logger.Get(ctx).Infof("getting new manifest error: %v", err)
--- a/internal/engine/upper.go
+++ b/internal/engine/upper.go
@@ -121,7 +121,6 @@ func (u Upper) CreateManifests(ctx context.Context, manifests []model.Manifest, 
 	output.Get(ctx).Summary(s.Output(ctx, u.resolveLB))
 
 	if watchMounts {
-		haveValidTiltfile := true
 		go func() {
 			err := u.reapOldWatchBuilds(ctx, manifests, time.Now())
 			if err != nil {
@@ -140,17 +139,14 @@ func (u Upper) CreateManifests(ctx context.Context, manifests []model.Manifest, 
 					newManifest, err := getNewManifestFromTiltfile(ctx, event.manifest.Name)
 					if err != nil {
 						logger.Get(ctx).Infof("getting new manifest error: %v", err)
-						haveValidTiltfile = false
 						continue
 					}
 					buildState := BuildStateClean
 					err = u.buildManifestFromBuildState(ctx, newManifest, buildState)
 					if err != nil {
 						logger.Get(ctx).Infof("config watch error: %v", err)
-						haveValidTiltfile = false
 						continue
 					}
-					haveValidTiltfile = true
 					u.manifestOverlay[newManifest.Name] = newManifest
 				} else {
 					oldState := u.buildStates[event.manifest.Name]
@@ -168,20 +164,10 @@ func (u Upper) CreateManifests(ctx context.Context, manifests []model.Manifest, 
 					}
 
 					manifest := u.getMostRecentManifest(event)
-					if !haveValidTiltfile {
-						newManifest, err := getNewManifestFromTiltfile(ctx, event.manifest.Name)
-						if err != nil {
-							logger.Get(ctx).Infof("getting new manifest error: %v", err)
-							haveValidTiltfile = false
-							continue
-						}
-						manifest = newManifest
-					}
 					err = u.buildManifestFromBuildState(ctx, manifest, buildState)
 					if err != nil {
 						return err
 					}
-					haveValidTiltfile = true
 					u.manifestOverlay[manifest.Name] = manifest
 
 					output.Get(ctx).Summary(s.Output(ctx, u.resolveLB))
--- a/internal/engine/upper.go
+++ b/internal/engine/upper.go
@@ -40,12 +40,14 @@ const maxChangedFilesToPrint = 5
 // TODO(nick): maybe this should be called 'BuildEngine' or something?
 // Upper seems like a poor and undescriptive name.
 type Upper struct {
-	b            BuildAndDeployer
-	watcherMaker watcherMaker
-	timerMaker   timerMaker
-	k8s          k8s.Client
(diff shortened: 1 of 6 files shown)


internal/engine/upper.go
```diff
@@ -173,6 +183,38 @@ func (u Upper) CreateManifests(ctx context.Context, manifests []model.Manifest,
...
+
+	result, err := u.b.BuildAndDeploy(ctx, m, b)
+	if err != nil {
+		if isPermanentError(err) {
+			return err
+		}
+		o := output.Get(ctx)
+		o.PrintColorf(o.Red(), "build failed: %v", err)
+	} else {
+		buildStates[m.Name] = NewBuildState(result)
```
nicks: i wonder if buildStates should be a field of the upper?

internal/engine/upper.go
```diff
@@ -131,40 +133,48 @@ func (u Upper) CreateManifests(ctx context.Context, manifests []model.Manifest,
...
-
-				spurious, err := buildState.OnlySpuriousChanges()
-				if err != nil {
-					logger.Get(ctx).Infof("build watch error: %v", err)
-				}
+				if eventContainsConfigFiles(event) {
+					t, err := tiltfile.Load(tiltfile.FileName, os.Stdout)
+					if err != nil {
+						// TODO(dmiller) should we fail here, or is this OK?
+						return err
```
nicks: i don't think we should fail here. we kind of expect that people will make mistakes writing a tiltfile

we probably should keep some state variable that the Tiltfile needs to be reloaded, and only reset that state variable when the tiltfile successfully loads. if it doesn't load, that just counts as a failed build

internal/engine/upper.go
```diff
@@ -131,40 +133,48 @@ func (u Upper) CreateManifests(ctx context.Context, manifests []model.Manifest,
...
+					}
+					newManifests, err := t.GetManifestConfigs(string(event.manifest.Name))
+					if err != nil {
+						return err
```
nicks: ditto here

internal/engine/upper.go
```diff
@@ -131,40 +133,48 @@ func (u Upper) CreateManifests(ctx context.Context, manifests []model.Manifest,
...
+						continue
+					}
 
-				result, err := u.b.BuildAndDeploy(
-					ctx,
-					event.manifest,
-					buildState)
-				if err != nil {
-					if isPermanentError(err) {
+					err = u.buildManifestFromBuildState(ctx, event.manifest, buildState, buildStates)
```
nicks: i think this manifest will be stale, right?

internal/engine/upper.go
```diff
@@ -131,40 +133,48 @@ func (u Upper) CreateManifests(ctx context.Context, manifests []model.Manifest,
...
 			case event := <-sw.events:
-				oldState := buildStates[event.manifest.Name]
-				buildState := oldState.NewStateWithFilesChanged(event.files)
-				buildStates[event.manifest.Name] = buildState
-
-				spurious, err := buildState.OnlySpuriousChanges()
-				if err != nil {
-					logger.Get(ctx).Infof("build watch error: %v", err)
-				}
+				if eventContainsConfigFiles(event) {
```
nicks: question: when you edit the Tiltfile manually with Servantes, what happens? does it do reload the Tiltfile 7 times, one for each service?

nicks: (this might be an ok failure case for now, i'm just curious what happens because i'm honestly not sure)

jazzdan: Yes it redeploys all of the services.

internal/engine/upper.go
jazzdan: Agreed, good idea on that extra piece of state.

nicks: what about the next loop iteration?

internal/engine/upper.go
nicks: what about the next loop iteration?

jazzdan: Aha, yeah it could be the case that the events have an old manifest.

jazzdan:
@nicks let me know what you think of this test. This doesn't work for kubernetes yaml yet, and I'll investigate why but I'm thinking that + additional refactors should wait until after this is merged to prevent merge conflicts.

internal/tiltfile/tiltfile_test.go
```diff
@@ -536,56 +536,6 @@ func TestReadFile(t *testing.T) {
 	assert.Equal(t, manifest.K8sYaml, "hello world")
 }
 
-func TestConfigMatcherWithFastBuild(t *testing.T) {
```
nicks: let's keep this test, but modify it to create the Matcher from ConfigFiles. this should be trivial because there's already a method for it, come talk to me if it's not trivial

internal/engine/upper.go
```diff
@@ -131,40 +136,57 @@ func (u Upper) CreateManifests(ctx context.Context, manifests []model.Manifest,
...
+					}
 
-				result, err := u.b.BuildAndDeploy(
-					ctx,
-					event.manifest,
-					buildState)
-				if err != nil {
-					if isPermanentError(err) {
+					manifest := u.getMostRecentManifest(event)
+					if !haveValidTiltfile {
```
nicks: I'm pretty confident that this `if` statement is unnecessary? if the Tiltfile is broken, and no config files have changed, then it will still be broken. if you remove this block, do any tests break?

internal/engine/upper.go
```diff
@@ -131,40 +136,57 @@ func (u Upper) CreateManifests(ctx context.Context, manifests []model.Manifest,
...
-					logger.Get(ctx).Infof("build watch error: %v", err)
-				}
+				if eventContainsConfigFiles(event) {
+					newManifest, err := getNewManifestFromTiltfile(ctx, event.manifest.Name)
+					if err != nil {
+						logger.Get(ctx).Infof("getting new manifest error: %v", err)
+						haveValidTiltfile = false
+						continue
+					}
+					buildState := BuildStateClean
```
nicks: you need to store BuildStateClean in u.buildStates. otherwise, if the build fails, then previous builds will poison things

internal/engine/upper.go
```diff
@@ -131,40 +136,57 @@ func (u Upper) CreateManifests(ctx context.Context, manifests []model.Manifest,
...
+					err = u.buildManifestFromBuildState(ctx, newManifest, buildState)
```
nicks: if you store BuildStateClean in u.buildStates, then you can remove a bunch of code. remove this, and remove the 'else', and let this fall into the code below

internal/engine/upper.go
jazzdan: Yep, good call.

PR #460, opened 2018-10-03
integration: add a basic integration test fixture

nicks:
Hello @jazzdan,

Please review the following commits I made in branch nicks/ch455/simpleup:

3b4635165319b81fb6f1e11486985afb27879cec (2018-10-03 15:09:18 -0400)
integration: add a basic integration test fixture
Note that this doesn't run on circleci yet

### DIFF
--- a/integration/fixture_test.go
+++ b/integration/fixture_test.go
@@ -1,0 +1,177 @@
+//+build integration
+
+package integration
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"go/build"
+	"io"
+	"io/ioutil"
+	"net/http"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/pkg/errors"
+)
+
+var packageDir string
+
+const namespaceFlag = "-n=tilt-integration"
+
+func init() {
+	pkg, err := build.Default.Import("github.com/windmilleng/tilt/integration", ".", build.FindOnly)
+	if err != nil {
+		panic(fmt.Sprintf("Could not find integration test source code: %v", err))
+	}
+
+	packageDir = pkg.Dir
+}
+
+type fixture struct {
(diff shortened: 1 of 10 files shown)


integration/fixture_test.go
```diff
@@ -0,0 +1,177 @@
...
+	body, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return "", errors.Wrap(err, "Curl")
+	}
+	return string(body), nil
+}
+
+func (f *fixture) CurlUntil(ctx context.Context, url string, expectedContents string) {
+	actualContents := ""
+	for true {
```
jazzdan: Could this be shortened to just `for {`?

integration/fixture_test.go
```diff
@@ -0,0 +1,177 @@
...
+				url, expectedContents, actualContents)
+		}
+
+		actualContents, err := f.Curl(url)
+		if err != nil {
+			continue
+		} else if strings.Contains(actualContents, expectedContents) {
+			return
+		}
+		time.Sleep(200 * time.Millisecond)
```
jazzdan: The way I'd expect to see this written in Go is a for loop that listens on a couple channels, `time.Tick`, `context.Cancel`, etc. But this is fine too.

integration/oneup/main.go
```diff
@@ -0,0 +1,14 @@
+package main
+
+import (
+	"net/http"
+)
+
+func main() {
+	http.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
+		msg := "🍄 One-Up! 🍄"
```
jazzdan: 😆

integration/fixture_test.go
nicks: good points! i cleaned up the loop a bit

integration/fixture_test.go
nicks: done

PR #462, opened 2018-10-03
tilt: add hud package

landism:
first stab at some types to communicate between tilt and the hud


internal/hud/model/model.go
```diff
@@ -0,0 +1,42 @@
...
+func NewResource(directoryWatched string) Resource {
+	return Resource{
+		DirectoryWatched:   directoryWatched,
+		LatestFileChanges:  []string{},
+		LastFileChangeTime: time.Now(),
+		Status:             ResourceStatusUnbuilt,
+	}
+}
+
+type Model struct {
```
maiamcc: this is a pretty un-descriptive name. if you want to call it `model` i'd put in the `hud` package so it'd descriptively/usefully imported as `hud.Model`; if you want to leave it here, maybe call it something more descriptive?

maiamcc: actually, wait, why does this need its own package at all? why not `internal/hud/model.go`?

internal/hud/view/model_to_view.go
```diff
@@ -0,0 +1,43 @@
...
+
+import (
+	"time"
+
+	hudmodel "github.com/windmilleng/tilt/internal/hud/model"
+	tiltmodel "github.com/windmilleng/tilt/internal/model"
+)
+
+func NewView(m hudmodel.Model) View {
+	var resources []Resource = nil
```
maiamcc: or just `var resources []Resource`

internal/hud/view/model_to_view.go
```diff
@@ -0,0 +1,43 @@
...
+	for name := range m.Resources {
```
maiamcc: can do `for name, r := range...` and then don't need to do `m.Resources[name]` later

internal/hud/view/model_to_view.go
```diff
@@ -0,0 +1,43 @@
...
+	ret := Resource{
+		Name:                    name.String(),
+		DirectoryWatched:        r.DirectoryWatched,
+		LatestFileChanges:       r.LatestFileChanges,
+		TimeSinceLastFileChange: time.Now().Sub(r.LastFileChangeTime),
+		Status:                  ResourceStatusStale,
+		StatusDesc:              "No pod found",
+	}
+
+	pod, ok := m.Pods[name]
```
maiamcc: nbd but can do as a oneliner:
`pod, ok := m.Pods[name]; if ok { ...`

internal/hud/view/model_to_view.go
```diff
@@ -0,0 +1,43 @@
+package view
+
+import (
+	"time"
+
+	hudmodel "github.com/windmilleng/tilt/internal/hud/model"
+	tiltmodel "github.com/windmilleng/tilt/internal/model"
+)
+
+func NewView(m hudmodel.Model) View {
```
maiamcc: spidey sense says that this should be a method ON a model (`m.ToView` or `m.NewView` or whatever), but i don't care too much 🤷‍♀️

internal/hud/view/model_to_view.go
```diff
@@ -0,0 +1,43 @@
...
+	var resources []Resource = nil
+	for name := range m.Resources {
+		resources = append(resources, NewResourceView(m, name, *m.Resources[name]))
+	}
+
+	return View{resources}
+}
+
+func NewResourceView(m hudmodel.Model, name tiltmodel.ManifestName, r hudmodel.Resource) Resource {
```
maiamcc: again, could make this a method on the model?

internal/hud/model/model.go
```diff
@@ -0,0 +1,42 @@
...
+	Resources map[model.ManifestName]*Resource
```
maiamcc: doesn't need to happen in this PR, but this is gonna need some concurrency protection o_0 (or we can do [what DB did in his example code](https://github.com/windmilleng/tilt/blob/dbentley/watch_model/internal/state/states/store.go) and give it a channel that it listens on for updates...)

internal/hud/model/model.go
landism: I had model and view in the same package, but then `model.ResourceStatus` conflicts with `view.ResourceStatus`, and having `model.ResourceStatus` as a peer to `view.ResourceStatus` seems nicer? 

`model` is super annoying since it conflicts with `internal/model`, and I had to go through and do renaming imports everywhere, which is probably a good reason to do something different.

Moving this into `engine` seems like a good solution.

internal/hud/view/model_to_view.go
landism: That's what I had originally, but then I figured maybe it's better to not give `Model` a dependency on `View`. If, e.g., at some point we have some other kind of view for a different UI, then it's weird to just have the one "NewView" instance method. That's so hypothetical and cheap to change anyway that it doesn't really matter, but I didn't have any good arguments in the other direction, either.

internal/hud/model/model.go
landism: yeah, hoping to end up with the latter

PR #463, opened 2018-10-03
proto: upd synclet proto

maiamcc:
uhh I guess there was a change to how we gen protofiles or something? updating the synclet proto in its own PR so my next proto-relevant PR doesn't have random synclet stuff in it


nicks:
lg

PR #468, opened 2018-10-04
tilt: have output.StartPipeline return a Context with a new Outputter

landism:
This means that once we have async builds, different threads writing to Outputter won't depend on the current pipeline state.


nicks:
hmmm...can you talk a little bit more about what the contract is? like, lots of other methods modify the outputter too...

landism:
I think the right solution here is to pull the state out of Outputter and explicitly pass it around the various pipeline-related methods, but that's a lot of work that's not on the critical path to getting the hud working.

landism:
no longer relevant after #581

PR #469, opened 2018-10-04
integration: fix a race condition: port-forwarding has to wait until all pods are up

nicks:
Hello @jazzdan,

Please review the following commits I made in branch nicks/race:

71e5785a79d27d4225e3f4056b2e79ac7cffa615 (2018-10-04 15:14:33 -0400)
integration: fix a race condition: port-forwarding has to wait until all pods are up

### DIFF
--- a/integration/fixture_test.go
+++ b/integration/fixture_test.go
@@ -1,5 +1,3 @@
-//+build integration
-
 package integration
 
 import (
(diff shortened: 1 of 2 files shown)


integration/fixture_test.go
```diff
@@ -1,5 +1,3 @@
-//+build integration
```
jazzdan: Did you mean to remove this?

nicks: yup! it makes the tooling check this code a bit better when the tags aren't on

PR #476, opened 2018-10-04
integration: push to windmill-test-containers and delete images between runs

nicks:
Hello @jazzdan,

Please review the following commits I made in branch nicks/images:

def394e4e66e22f1b2ddc7dc4675ffbda304b864 (2018-10-04 18:31:50 -0400)
integration: push to windmill-test-containers and delete images between runs

### DIFF
--- a/integration/purge-image.sh
+++ b/integration/purge-image.sh
@@ -1,0 +1,8 @@
+#!/bin/sh
+
+set -ex
+IMAGE="$1"
+TAGS=$(gcloud container images list-tags $IMAGE --limit=999999 --format='get(digest)')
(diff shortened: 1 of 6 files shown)


integration/purge-image.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+
+IMAGE="$1"
```
jazzdan: `set -ex` here as well?

nicks:
probably going to sleep on this before merging, not totally convinced it's a good idea. it would make it weird to run two integration test suites at the same time, for example

PR #477, opened 2018-10-05
model: manifests can be compared

jazzdan:


jazzdan:
Should I add another test that fails if the structure of Manifest has changed so that we know to update the equality function?

internal/model/manifest.go
```diff
@@ -102,6 +103,14 @@ func (m Manifest) validate() *ValidateErr {
 	return nil
 }
 
+func (m1 Manifest) Equal(m2 Manifest) bool {
+	primitivesMatch := m1.Name == m2.Name && m1.K8sYaml == m2.K8sYaml && m1.DockerRef == m2.DockerRef && m1.BaseDockerfile == m2.BaseDockerfile && m1.StaticDockerfile == m2.StaticDockerfile && m1.StaticBuildPath == m2.StaticBuildPath
+	cmdMatch := m1.Entrypoint.Equal(m2.Entrypoint)
+	pmMatch := m1.FileFilter == m2.FileFilter
```
nicks: i don't think this will work for any non-trivial file filter?

nicks: also, did you forget to compare ConfigFiles?

nicks: and Mounts

jazzdan: These are in now.

jazzdan: (after #493 is landed, this'll have to be changed a bit and then it should be good to merge)

PR #483, opened 2018-10-05
tilt: connect hud to upper

landism:
1. `upperState` has been replaced by `engineState`, so delete `engineState`
2. Actually update the hud with the current view on each iteration of the hud's loop


internal/engine/engine_state.go
```diff
@@ -38,3 +49,42 @@ func newManifestState(manifest model.Manifest) *manifestState {
 		pendingFileChanges: make(map[string]bool),
 	}
 }
+
+func stateToView(s engineState) view.View {
+	fmt.Printf("engine state is '%v'\n", s)
```
nicks: :eyes:

PR #484, opened 2018-10-05
tilt: track pod status in engine state

landism:


internal/engine/engine_state.go
```diff
@@ -45,9 +47,19 @@ func newManifestState(manifest model.Manifest) *manifestState {
...
 }
 
+type Pod struct {
+	Name      string
+	StartedAt time.Time
+	Status    string
+}
+
+// manifestState.Pod will be set to this if we don't know anything about its pod
+var unknownPod = Pod{Name: "no pod yet found"}
```
nicks: i'm surprised you didn't use the zero value for this?

internal/engine/engine_state.go
```diff
@@ -45,9 +47,19 @@ func newManifestState(manifest model.Manifest) *manifestState {
...
 		manifest:           manifest,
 		pendingFileChanges: make(map[string]bool),
+		pod:                unknownPod,
 	}
 }
 
+type Pod struct {
+	Name      string
+	StartedAt time.Time
+	Status    string
```
nicks: maybe use the PodPhase type and constants?
https://godoc.org/k8s.io/api/core/v1#PodPhase

internal/engine/pod_watch.go
```diff
@@ -0,0 +1,78 @@
...
+	"github.com/windmilleng/tilt/internal/k8s"
+	"k8s.io/api/core/v1"
+)
+
+type podWatcher struct {
+	events <-chan *v1.Pod
+}
+
+func makePodWatcher(ctx context.Context, kCli k8s.Client) (*podWatcher, error) {
+	ch, err := kCli.WatchPods(ctx, "default", []k8s.LabelPair{TiltRunLabel()})
```
nicks: i think we want to be able to support things in the non-default namespace?

internal/engine/upper.go
```diff
@@ -41,35 +43,38 @@ const maxChangedFilesToPrint = 5
...
-		b:            b,
-		watcherMaker: watcherMaker,
-		timerMaker:   time.After,
-		k8s:          k8s,
-		browserMode:  browserMode,
-		reaper:       reaper,
-		hud:          hud,
+		b:               b,
+		fsWatcherMaker:  fsWatcherMaker,
+		podWatcherMaker: makePodWatcher,
```
nicks: can we use wire.go to inject this?

in theory, this shouldn't need to take a k8s.Client, because wire.go is supposed to take care of that.

internal/k8s/watch_test.go
```diff
@@ -0,0 +1,125 @@
...
+	}
+
+	ch, err := tf.kCli.WatchPods(tf.ctx, Namespace("default"), []LabelPair{})
+	if !assert.NoError(tf.t, err) {
+		return
+	}
+
+	var observedPods []runtime.Object
+
+	timeout := time.After(time.Millisecond * 500)
```
nicks: this seems like it will always make the test take 500ms? can we do better?

internal/k8s/watch.go
```diff
@@ -0,0 +1,47 @@
...
+
+	watcher, err := kCli.core.Pods(ns.String()).Watch(metav1.ListOptions{LabelSelector: ls.String()})
+	if err != nil {
+		return nil, err
+	}
+
+	go func() {
+		for {
+			select {
+			case event := <-watcher.ResultChan():
```
nicks: you probably want to use the two argument pull off the channel, so you can tell if the channel is closed by the server, and clean up appropriately.

internal/engine/engine_state.go
landism: Yeah, I need to get more comfortable with zero values. I'm still getting used to not having `null` or `Option`

internal/engine/engine_state.go
landism: I feel like the UI should be able to represent statuses that show up in `kubectl get pods`, and afaik there's no const with `CrashLoopBackOff`.

internal/engine/upper.go
landism: yep, thanks

PR #485, opened 2018-10-08
tilt: move initial builds into event loop

landism:
This makes it so that we update the engine state and HUD while the initial round of builds is executing.

It also slightly reduces duplicated code (we now only handle BaD in one place).

I don't have a good mental model about what the LB stuff is doing and am not sure I've made the right change. It's working on my machine, but I don't know if that's just because there are already LBs deployed or something.

### DIFF
--- a/internal/engine/engine_state.go
+++ b/internal/engine/engine_state.go
@@ -1,6 +1,7 @@
 package engine
 
 import (
+	"github.com/windmilleng/tilt/internal/k8s"
 	"github.com/windmilleng/tilt/internal/ospath"
 
 	"github.com/windmilleng/tilt/internal/hud/view"
@@ -15,6 +16,11 @@ type engineState struct {
 	manifestsToBuild  []model.ManifestName
 	currentlyBuilding model.ManifestName
 	completedBuilds   chan completedBuild
+
+	initialBuildCount   int
+	completedBuildCount int
+
+	openBrowserOnNextLB bool
 }
 
 type manifestState struct {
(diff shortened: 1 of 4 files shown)


internal/engine/engine_state.go
```diff
@@ -15,13 +16,20 @@ type engineState struct {
 	manifestsToBuild  []model.ManifestName
 	currentlyBuilding model.ManifestName
 	completedBuilds   chan completedBuild
+
+	initialBuildCount   int
```
jazzdan: Hmm, could there be a better name for this? I think a lot of people are going to be looking at `engineState`. Maybe a comment? The others are clear to me by name though. 👍

internal/engine/upper_test.go
```diff
@@ -151,18 +151,10 @@ func TestUpper_UpWatchError(t *testing.T) {
 		f.watcher.errors <- errors.New("bazquu")
 	}()
 	err := f.upper.CreateManifests(output.CtxForTest(), []model.Manifest{manifest}, true)
-	close(f.b.calls)
 
 	if assert.NotNil(t, err) {
 		assert.Equal(t, "bazquu", err.Error())
 	}
-
-	var startedManifests []model.Manifest
```
nicks: why did this test change?

landism: before, we would always run a set of builds before pulling from the file watching chans

now it's all in one select loop, so it's possible we get the watch error before BaD is invoked (`dispatch` is always invoked before we pull from the chan, but then it calls BaD in a separate goroutine).

PR #488, opened 2018-10-08
tiltfile: don't parse ignore files at runtime

jazzdan:
Instead just pass the contents of the ignore files along so they can be
parsed later. This is important so that we can end up with a manifest
that consists entirely of comparable fields.

### DIFF
--- a/internal/git/gitignore.go
+++ b/internal/git/gitignore.go
@@ -2,15 +2,10 @@ package git
 
 import (
 	"context"
-	"os"
-	"path"
 	"path/filepath"
 	"strings"
-	"syscall"
 
 	"github.com/monochromegane/go-gitignore"
-	"github.com/windmilleng/tilt/internal/logger"
-	"github.com/windmilleng/tilt/internal/model"
 	"github.com/windmilleng/tilt/internal/ospath"
 )
 
@@ -26,9 +21,7 @@ type gitIgnoreTester struct {
 	ignoreMatcher gitignore.IgnoreMatcher
 }
 
-var _ model.PathMatcher = gitIgnoreTester{}
-
-func (i gitIgnoreTester) Matches(f string, isDir bool) (bool, error) {
+func (i *gitIgnoreTester) Matches(f string, isDir bool) (bool, error) {
 	_, isChild := ospath.Child(i.repoRoot, f)
 	if !isChild {
 		return false, nil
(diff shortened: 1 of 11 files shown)


jazzdan:
I think there's some more refactoring/code deleting that can be done in the git and dockerignore packages now that they aren't responsible for reading the files.

jazzdan:
I also want to test this more manually before I merge it in.

internal/git/gitignore.go
```diff
@@ -81,10 +61,16 @@ func (r repoIgnoreTester) Matches(f string, isDir bool) (bool, error) {
...
-	g, err := NewGitIgnoreTester(ctx, repoRoot)
+func NewRepoIgnoreTester(ctx context.Context, repoRoot string, gitignoreContents string) (*repoIgnoreTester, error) {
+	g, err := NewGitIgnoreTesterFromContents(ctx, repoRoot, gitignoreContents)
 	if err != nil {
 		return nil, err
 	}
 	return &repoIgnoreTester{repoRoot, g}, nil
 }
+
+type nothingMatcher struct{}
```
nicks: there should already be something like this in the main model package?

internal/model/matcher.go
```diff
@@ -73,6 +73,9 @@ func NewCompositeMatcher(matchers []PathMatcher) PathMatcher {
 
 func (c CompositePathMatcher) Matches(f string, isDir bool) (bool, error) {
 	for _, t := range c.Matchers {
+		if t == nil {
```
nicks: this seems wrong. we should disallow nils in the matcher repr

internal/tiltfile/tiltfile.go
```diff
@@ -200,20 +196,26 @@ func makeSkylarkGitRepo(thread *skylark.Thread, fn *skylark.Builtin, args skylar
 		return nil, fmt.Errorf("%s isn't a valid git repo: it doesn't have a .git/ directory", absPath)
 	}
 
-	ctx, cancel := context.WithTimeout(context.Background(), 1*time.Second)
-	defer cancel()
-	t1, err := git.NewRepoIgnoreTester(ctx, absPath)
+	gitignoreContents, err := ioutil.ReadFile(filepath.Join(absPath, ".gitignore"))
 	if err != nil {
-		return nil, err
+		if !os.IsNotExist(err) {
```
nicks: if err != nil && !os.IsNotExist(err)
?

internal/git/gitignore.go
jazzdan: Yes but I can't include the `model` package because then I have a circular dependency. :(

I could try moving it out in to another package though.

internal/model/matcher.go
jazzdan: Yeah, I need to track down where this is coming from.

internal/git/gitignore.go
jazzdan: After talking with @nicks I'm going to reorganize the model package so that it doesn't depend on the gitignore or dockerignore packages in a separate PR.

PR #490, opened 2018-10-08
model: stop referring to external ignore packages

jazzdan:


jazzdan:
Open to other ideas for how to do this :)

internal/model/manifest.go
```diff
@@ -52,21 +50,21 @@ func (m Manifest) IsStaticBuild() bool {
 	return m.StaticDockerfile != ""
 }
 
-func (m Manifest) Filter() PathMatcher {
+// IgnoreContentsToPathMatcher is a function that takes the GitignoreContents of a
+// manifest and the dockerignoreContents of a manifest and returns a PathMatcher
+// that represents both.
+type IgnoreContentsToPathMatcher func(repoPath string, gitignoreContents string, dockerignoreContents string) (PathMatcher, error)
```
nicks: this api seems convoluted...

can you just have a single function

ignore.CreateFilter(manifest)

rather than bouncing the flow of control between these two functions?

PR #492, opened 2018-10-09
tilt: make renderer actually write to hud

landism:


landism:
there may be some ugliness in the pen/tcell code, but that's mostly well-swept under a rug, and getting this into master lets us make things actually work

internal/engine/upper.go
```diff
@@ -150,6 +162,8 @@ func (u Upper) CreateManifests(ctx context.Context, manifests []model.Manifest,
 			handlePodEvent(ctx, engineState, pod)
 		case err := <-sw.errs:
 			return err
+		case <-time.After(refreshInterval):
```
nicks: i don't understand what this code is doing. why would we need to do a hud update if nothing has changed?

internal/hud/renderer.go
```diff
@@ -1,25 +1,60 @@
...
-	ctx     context.Context
+	screen tcell.Screen
 }
 
 func (r *Renderer) Render(v view.View) error {
-	// TODO: draw v
-	// if r.ttyPath != "" {
-	//     log.Infof("drawing to tty: %s\n", r.ttyPath)
-	// }
+	if r.screen != nil {
```
nicks: i worry about the race conditions here, it might be better to have a readyCh that closes when the screen is populated, and wait on readyCh here

landism: I was mostly relying on the `screen` assignment at the end of `SetUp` being atomic, but:
1. Apparently it's not fair to assume pointer assignments are atomic in go
2. You're right that that's asking for bugs

I've gone with a mutex instead of a channel because we don't want this method to block when no hud is attached.

internal/engine/upper.go
nicks: ok

i have a long-term concern that the design pattern here (have a central state object that's updated in a loop and sent to listeners) is basically isomorphic to the redux design pattern, and that we're going to end up reinventing a lot of the concepts there. So I'd rather borrow liberally rather than try inventing approaches that don't work.

The "show the age of a thing" is pretty much in every React 101 tutorial, and tends to be handled by component-local state, and tends to be discouraged in global state.

internal/hud/renderer.go
nicks: :+1:

PR #498, opened 2018-10-09
tiltfile: keep track of repos by manifest name

jazzdan:
In the composite manifest case we are loading repos for different
manifests.

### DIFF
--- a/internal/tiltfile/tiltfile.go
+++ b/internal/tiltfile/tiltfile.go
@@ -199,7 +199,10 @@ func makeSkylarkGitRepo(thread *skylark.Thread, fn *skylark.Builtin, args skylar
 	}
 
 	repo := gitRepo{absPath, string(gitignoreContents), string(dockerignoreContents)}
-	err = addRepo(thread, repo)
+	if thread.Caller() == nil {
+		return nil, fmt.Errorf("internal error: no Caller for local_git_repo")
+	}
+	err = addRepo(thread, repo, thread.Caller().Callable().Name())
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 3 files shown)


jazzdan:
Still want to add tests for this, just checking something in CI.

internal/tiltfile/tiltfile.go
```diff
@@ -363,7 +363,7 @@ func (tiltfile Tiltfile) GetManifestConfigs(manifestName string) ([]model.Manife
 	}
 }
 
-func skylarkManifestToDomain(manifest k8sManifest, repos []gitRepo) (model.Manifest, error) {
+func skylarkManifestToDomain(manifest k8sManifest, repos map[string][]gitRepo, name model.ManifestName) (model.Manifest, error) {
```
nicks: this seems...weird

would it be easier if the git repo was attached to the k8sManifest object? i don't understand why it's not.

jazzdan:
(added a simple test)

internal/tiltfile/tiltfile.go
jazzdan: yeah, let's see what refactors I can do here.

nicks: on second thought, this may be totally redundant? i'm pretty sure you can pull the git repo off of manifest.mount.localPath.repo.

jazzdan: This doesn't work out of the box but maybe that's a simpler way to associate repos with mounts in the tiltfile. Let me play around with that.

jazzdan:
closing in favor of #499

PR #499, opened 2018-10-10
tiltfile: associate repos to mounts iff they are used

jazzdan:
Note that with this approach we only record a repo if it is used in a mount. This might be OK, but if you used a `repo.path` and interpolated in to like a `local` command and _never_ used it in a mount we wouldn't know that you used it

### DIFF
--- a/internal/tiltfile/skylarkTypes.go
+++ b/internal/tiltfile/skylarkTypes.go
@@ -105,6 +105,7 @@ func (k *k8sManifest) createPortForward(thread *skylark.Thread, fn *skylark.Buil
 type mount struct {
 	src        localPath
 	mountPoint string
+	repo       gitRepo
 }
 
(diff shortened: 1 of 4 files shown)


internal/tiltfile/skylarkTypes.go
```diff
@@ -204,17 +205,18 @@ func addMount(thread *skylark.Thread, fn *skylark.Builtin, args skylark.Tuple, k
 		return nil, err
 	}
 
-	var lp localPath
+	m := mount{}
 	switch p := src.(type) {
 	case localPath:
-		lp = p
+		m.src = p
```
nicks: i thought localPaths had git repos attached?

jazzdan: A mount can either be from a gitRepo, or from a plain old path on disk. localPAths are those plain old paths, and they won't have a gitRepo.

PR #502, opened 2018-10-10
engine: fix gitignore/dockerignore regression

nicks:
Hello @jazzdan, @landism,

Please review the following commits I made in branch nicks/ch543/upper:

6387d9b7896060f582a25a98f27bd938d5eb3436 (2018-10-10 12:51:44 -0400)
engine: fix gitignore/dockerignore regression


nicks:
for what it's worth, i think this API doesn't make sense at all anymore

for example, you could have gitignored generated k8s yaml. i.e., you want a change to the k8s yaml to mark the manifest dirty, but might not require a rebuild

i think we probably want to move to an architecture where the file changes mark "dirty" bits on the state machine, and the various file filters affect which bits are marked dirty, rather than filtering out the file events entirely.

PR #503, opened 2018-10-10
engine: organize watch loop to use reactive programming conventions

nicks:
Hello @landism,

Please review the following commits I made in branch nicks/actions:

6fe0c1d87cae5248332f883f0261050cda8f0bde (2018-10-10 15:47:01 -0400)
engine: organize watch loop to use reactive programming conventions

### DIFF
--- a/internal/engine/actions.go
+++ b/internal/engine/actions.go
@@ -1,0 +1,23 @@
+package engine
+
+import "k8s.io/api/core/v1"
+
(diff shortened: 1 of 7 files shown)


nicks:
(basically, i want to add new action types, but was worried about the proliferation of channels being passed around...this solves my immediate problem and moves us closer to where i think we should be going)|

PR #504, opened 2018-10-10
model: make trigger a primitive

jazzdan:


internal/model/manifest.go
```diff
@@ -201,10 +201,10 @@ func (r1 LocalGithubRepo) Equal(r2 LocalGithubRepo) bool {
 type Step struct {
 	// Required. The command to run in this step.
 	Cmd Cmd
-
 	// Optional. If not specified, this step runs on every change.
 	// If specified, we only run the Cmd if the trigger matches the changed file.
-	Trigger PathMatcher
+	Triggers         []string
+	WorkingDirectory string
```
nicks: hmmmm....i would expect `WorkingDirectory` to be the working directory where the command is run in the container, not the root directory of the trigger patterns

jazzdan: yeah... `BasePath`? `TriggerDirectory`?

PR #506, opened 2018-10-10
model: compare steps and cmds in equality function

jazzdan:


internal/model/manifest.go
```diff
@@ -177,6 +178,24 @@ func (m1 Manifest) portForwardsEqual(m2 Manifest) bool {
 	return true
 }
 
+func (m1 Manifest) stepsEqual(s2 []Step) bool {
+	if (m1.Steps == nil) != (s2 == nil) {
```
nicks: this should be unnecessary, because len(nil) == 0

internal/model/manifest.go
```diff
@@ -208,6 +227,32 @@ type Step struct {
...
+func (s1 Step) Equal(s2 Step) bool {
+	if s1.BaseDirectory != s2.BaseDirectory {
+		return false
+	}
+
+	if !s1.Cmd.Equal(s2.Cmd) {
+		return false
+	}
+
+	if (s1.Triggers == nil) != (s2.Triggers == nil) {
```
nicks: ditto

PR #507, opened 2018-10-10
tilt: save latest build log in engine state

landism:


internal/logger/logger.go
```diff
@@ -140,3 +140,28 @@ func (l logger) Writer(level Level) io.Writer {
 func (l logger) SupportsColor() bool {
 	return l.supportsColor
 }
+
+func CtxWithForkedOutput(ctx context.Context, writer io.Writer) context.Context {
```
nicks: can you write some comments on what this does? it took me a few minutes to figure it out, it's not obvious.

PR #508, opened 2018-10-10
engine: stream the logs from the deployed container

nicks:

Please review the following commits I made in branch nicks/ch520/podlogs:

ba10bf1aea98bcc3555cdbc9b827a7634ff05b77 (2018-10-10 17:34:54 -0400)
engine: stream the logs from the deployed container

### DIFF
--- a/internal/cli/wire.go
+++ b/internal/cli/wire.go
@@ -35,6 +35,7 @@ func wireManifestCreator(ctx context.Context, browser engine.BrowserMode) (model
 		engine.DeployerWireSet,
 		engine.DefaultShouldFallBack,
 		engine.ProvidePodWatcherMaker,
+		engine.NewPodLogManager,
 
 		hud.NewDefaultHeadsUpDisplay,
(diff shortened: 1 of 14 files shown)


nicks:
i still need to write some unit tests, but i wanted to post this to get you a feel for what it looks like

here's some screenshots of it in action with a small 2-server setup, where one server is running webpack and the other is running a go frontend

the webpack server logs are prefixed with [webpack], the go frontend logs are prefixed with [server]

screenshot from 2018-10-10 17-36-15

screenshot from 2018-10-10 17-36-37

nicks:
it's particularly interesting that because we're updating the container in-place, the logs start streaming before the build gets a chance to report completion

nicks:
ok, added some tests too

nicks:
updated based on offline discussion, with some tests for error-handling conditions

landism:
I think it's a step forward as-is, and we can worry about the interleaved output later.

PR #514, opened 2018-10-11
tilt: support build log replay

landism:
The 1-9 keys will now replay the last build log for that resource (in display order).

This is not the interface we intend to ship with the alpha, but is a cheap way to get us basic usability while we mature the UI code.

### DIFF
--- a/internal/engine/upper.go
+++ b/internal/engine/upper.go
@@ -75,15 +75,6 @@ func NewUpper(ctx context.Context, b BuildAndDeployer, k8s k8s.Client, browserMo
 		return watch.NewWatcher()
 	}
 
-	// Run the HUD in the background
-	go func() {
-		err := hud.Run(ctx)
-		if err != nil {
-			//TODO(matt) this might not be the best thing to do with an error - seems easy to miss
-			logger.Get(ctx).Infof("error in hud: %v", err)
-		}
-	}()
-
 	return Upper{
 		b:               b,
 		fsWatcherMaker:  fsWatcherMaker,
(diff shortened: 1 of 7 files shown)


internal/engine/upper_test.go
```diff
@@ -878,19 +937,42 @@ func newTestFixture(t *testing.T) *testFixture {
...
+		fsWatcher:      watcher,
+		timerMaker:     &timerMaker,
+		docker:         docker,
+		hud:            hud,
+		podEvents:      podEvents,
+		log:            log,
 	}
 }
 
+func (f *testFixture) Start(manifests []model.Manifest, watchMounts bool) {
```
landism: I feel like this Start/Stop is a bit cleaner than the way we've been writing upper tests:
1. no `go` in the tests themselves
2. canceling the context instead of feeding sentinels through watch errors
3. removes a bit of redundancy

I haven't yet converted existing tests to this.

internal/engine/upper.go
```diff
@@ -497,4 +498,22 @@ func (u Upper) reapOldWatchBuilds(ctx context.Context, manifests []model.Manifes
 	return nil
 }
 
+func replayBuildLog(ctx context.Context, state store.EngineState, resourceNumber int) {
```
nicks: it seems weird for this to be an action, because it doesn't really modify state? i think this is fine for now, but this probably should be a subscriber with some way to get a read lock on the state (or mutable state)

internal/engine/upper_test.go
```diff
@@ -836,15 +891,17 @@ func makeFakePodWatcherMaker(ch chan *v1.Pod) func(context.Context, *store.Store
...
+	cancel                func()
+	upper                 Upper
+	b                     *fakeBuildAndDeployer
+	fsWatcher             *fakeNotify
+	timerMaker            *fakeTimerMaker
+	docker                *docker.FakeDockerClient
+	hud                   *hud.FakeHud
+	podEvents             chan *v1.Pod
+	createManifestsResult chan error
+	log                   *bytes.Buffer
```
nicks: fwiw, i've been using a ThreadSafeBuffer to make this a bit easier to check the buffer without stopping

internal/engine/upper_test.go
nicks: ya, i like this pattern!

PR #515, opened 2018-10-11
tilt: hud: allow more than one mount

landism:


internal/store/engine_state.go
```diff
@@ -77,12 +77,25 @@ type Pod struct {
 // manifestState.Pod will be set to this if we don't know anything about its pod
 var UnknownPod = Pod{}
 
-func shortenFileList(baseDir string, files []string) []string {
+func shortenFile(baseDirs []string, f string) string {
+	ret := f
+	for _, baseDir := range baseDirs {
+		trimmed := strings.TrimPrefix(strings.TrimPrefix(f, baseDir), "/")
```
nicks: fwiw, ospath.Child might be helpful here

PR #517, opened 2018-10-11
engine: deploy based on equality

jazzdan:


internal/engine/upper.go
```diff
@@ -195,6 +195,13 @@ func (u Upper) dispatch(ctx context.Context, state *store.EngineState) {
 			ms.LastBuildDuration = 0
 			return
 		}
+
+		if newManifest.Equal(ms.Manifest) {
+			logger.Get(ctx).Verbosef("Manifest %s hasn't changed, not rebuilding", ms.Manifest.Name)
```
nicks: maybe Debugf?

internal/engine/upper.go
```diff
@@ -195,6 +195,13 @@ func (u Upper) dispatch(ctx context.Context, state *store.EngineState) {
...
+			state.CurrentlyBuilding = ""
+			return
```
nicks: i don't think returning is right here? there could be other pending changes that require a rebuild

nicks: now that we have a Store, it should be easy to test this condition by doing
store.Dispatch(manifestFilesChangedAction{
  files: []string{ /* multiple files */ },
  manifest: manifest,
})

jazzdan: huh? test by dispatching an action?

nicks: ya, as opposed to pushing a single file event

jazzdan: @nicks I think the `return` is necessary. What we're effectively saying here is if there's a config file change re-execute the Tiltfile. If there's no resultant change in this manifest then we don't need to do anything.

If I remove the `return` every manifest is re-deployed when the Tiltfile changes, since that's a dependency of every manifest.

jazzdan: But you're right about the other pending changes.

nicks: ya, to be clear, i'm not saying "just remove the return and everything will be fixed", i'm saying "some cases need the return and other cases it's wrong, so the logic will need to be more complex"

jazzdan:
@nicks okay I think this is ready for another review.

PR #518, opened 2018-10-11
engine: use rwlocks to control access to state

nicks:
Hello @landism,

Please review the following commits I made in branch nicks/rlock:

e14b135ef3126a9ff86028e235d116ea94b67cfc (2018-10-11 16:29:16 -0400)
engine: use rwlocks to control access to state


nicks:
this is mostly for the transition...i'm worried about the emerging pattern of "anything that needs access to state needs to be in `Upper`...which is not a good pattern

PR #521, opened 2018-10-11
engine: send pod logs to EngineState

nicks:
Hello @landism,

Please review the following commits I made in branch nicks/ch531/podlogs:

248515369f4ff4f6a5a6813d4ea6681032d0993b (2018-10-11 17:50:38 -0400)
engine: send pod logs to EngineState

### DIFF
--- a/internal/engine/upper.go
+++ b/internal/engine/upper.go
@@ -158,6 +158,8 @@ func (u Upper) reduceAction(ctx context.Context, state *store.EngineState, actio
 		handlePodEvent(ctx, state, action.Pod)
 	case ServiceChangeAction:
 		handleServiceEvent(ctx, state, action.Service)
+	case PodLogAction:
+		handlePodLogAction(state, action)
 	case BuildCompleteAction:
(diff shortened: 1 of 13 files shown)


internal/engine/upper.go
```diff
@@ -345,23 +347,54 @@ func handlePodEvent(ctx context.Context, st *store.Store, pod *v1.Pod) {
...
-		Status:    podStatusToString(*pod),
-	}
+	podID := k8s.PodIDFromPod(pod)
+	startedAt := pod.CreationTimestamp.Time
+	status := podStatusToString(*pod)
 
 	ms, ok := state.ManifestStates[manifestName]
 	if !ok {
-		logger.Get(ctx).Infof("error: got notified of pod for unknown manifest '%s'", manifestName)
+		// This is OK. The user could have edited the manifest recently.
```
landism: IIUC, this should only happen if the list of known manifests changes, which we don't currently support, so hitting this would be a bug. Have I missed a legit case that would result in this?

PR #526, opened 2018-10-12
tilt: change replayBuildLog to showError

landism:
Now, when pressing a resource #, we will:
if the last build failed, replay the build log
if the last build succeeded, replay the pod log since the last build started

### DIFF
--- a/internal/engine/upper_test.go
+++ b/internal/engine/upper_test.go
@@ -1000,7 +1000,6 @@ func TestUpper_ShowErrorPodLog(t *testing.T) {
 	}, "\n")
 	assert.Contains(t, f.log.String(), expectedOutput)
 
-	f.assertAllActionsConsumed()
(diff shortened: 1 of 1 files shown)


internal/engine/upper_test.go
```diff
@@ -1268,3 +1347,11 @@ func (f *testFixture) assertAllHUDUpdatesConsumed() {
 		f.T().Fatalf("Update not consumed: %+v", update)
 	}
 }
+
+func (f *testFixture) assertAllActionsConsumed() {
```
nicks: hmmmm...i'm not sure this is a good idea? in theory, the test has no control over what actions are consumed? can you add some comments on what this is trying to solve?

landism: I added this while trying to debug my failing test, and thought it might be good to leave in, but you're right, it's worse than nothing.

PR #527, opened 2018-10-12
store: add a subscriber api, and update the HUD async

nicks:
Hello @landism,

Please review the following commits I made in branch nicks/sub:

43a83c75d84883cc0f346ea94ccf89f8edf6abb1 (2018-10-12 15:51:48 -0400)
store: add a subscriber api, and update the HUD async


nicks:
(the goal of this is to move towards an API where we can do lots of things in parallel, like builds and port-forwarding and HUD updates and not have to worry so much about blocking the main loop)

nicks:
gah i need to fix up a bunch of tests

PR #529, opened 2018-10-12
engine: config files aren't ignored at the watch level

jazzdan:
If config files are ignored then we won't learn about changes to them,
which breaks users expectations that, if they change yaml or the
Tiltfile that their manifests are changed.

### DIFF
--- a/internal/engine/up_watch.go
+++ b/internal/engine/up_watch.go
@@ -130,7 +130,12 @@ func snsToManifestWatcher(ctx context.Context, st *store.Store, timerMaker timer
 			//defer close(events)
 			//defer close(errs)
 
-			filter := ignore.CreateFileChangeFilter(manifest)
+			filter, err := ignore.CreateFileChangeFilter(manifest)
+			if err != nil {
+				st.Dispatch(NewErrorAction(err))
+				return
+			}
+
 			for {
 				select {
 				case err, ok := <-watcher.Errors():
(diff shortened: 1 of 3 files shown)


internal/engine/up_watch.go
```diff
@@ -131,6 +132,10 @@ func snsToManifestWatcher(ctx context.Context, st *store.Store, timerMaker timer
 			//defer close(errs)
 
 			filter := ignore.CreateFileChangeFilter(manifest)
+			configMatcher, err := manifest.ConfigMatcher()
```
nicks: i wonder if we should bake this logic into CreateFileChangeFilter? so FileChangeFilter simply wouldn't filter out anything that matched a ConfigFile

jazzdan: Yeah I thought `CreateFileChangeFilter` was used somewhere else, but I see now that it isn't. Will do.

internal/ignore/path_matcher.go
```diff
@@ -35,6 +49,11 @@ func CreateBuildContextFilter(m model.Manifest) model.PathMatcher {
 // Filter out files that should not trigger new builds.
 func CreateFileChangeFilter(m model.Manifest) model.PathMatcher {
 	matchers := []model.PathMatcher{}
+	configMatcher, err := m.ConfigMatcher()
+	if err == nil {
+		matchers = append(matchers, configMatcher)
```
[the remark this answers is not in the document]
jazzdan: whoops, yes good call

PR #535, opened 2018-10-15
tilt: use rty to render hud

landism:
this took 10 minutes this morning and worked on the first try!


landism:
> And looks like we can close https://app.clubhouse.io/windmill/story/501/make-tcell-using-code-less-wip-y right?

Whoops - I guess I should actually delete all that pen code that we're no longer using.

Yeah, I think we're good on that.

PR #536, opened 2018-10-15
engine: fix a flaky test

nicks:
Hello @nicks,

Please review the following commits I made in branch nicks/segfault:

3af804eef7c6336d3c9870bbb20fb47fa137ff26 (2018-10-15 16:38:41 -0400)
engine: fix a flaky test


nicks:
on second thought, i think this might be indicative of a real bug, lemme think about this more

PR #545, opened 2018-10-16
kustomize: get dependencies from a kustomize setup

jazzdan:


internal/kustomize/dependencies.go
```diff
@@ -0,0 +1,97 @@
...
+
+import (
+	"io/ioutil"
+	"path/filepath"
+
+	yaml "gopkg.in/yaml.v2"
+)
+
+// kustomization is the content of a kustomization.yaml file.
+type kustomization struct {
```
nicks: is this copied from the kustomize repo? would it ever make sense to import their code directly (i haven't looked into how well their code is organized...)

internal/kustomize/dependencies_test.go
```diff
@@ -0,0 +1,178 @@
...
+		"staging/kustomization.yaml",
+		"production/kustomization.yaml",
+		"kustomization.yaml",
+	}
+	f.assertDeps(expected)
+}
+
+type kustomizeFixture struct {
+	t       *testing.T
+	tempdir *tempdir.TempDirFixture
```
nicks: we usually embed TempDirFixture directly, so you can just do f.T() and f.WriteDir() out of the box

https://golang.org/doc/effective_go.html#embedding

internal/kustomize/dependencies.go
jazzdan: [It is](https://github.com/GoogleContainerTools/skaffold/blob/master/pkg/skaffold/deploy/kustomize.go), but it doesn't look very importable unless I'm missing something.

PR #549, opened 2018-10-16
engine: correctly handle the case when the container id hasn't materialized

nicks:
Hello @landism, @maiamcc,

Please review the following commits I made in branch nicks/cid:

9dbd65f3303d9617f753b53579f523e043bc8f8d (2018-10-16 15:19:17 -0400)
engine: correctly handle the case when the container id hasn't materialized

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,3 +22,4 @@
 *.generated.yaml
 
 .synclet-dev-image-tag
+.#*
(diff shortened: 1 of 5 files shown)


.gitignore
```diff
@@ -22,3 +22,4 @@
 *.generated.yaml
 
 .synclet-dev-image-tag
+.#*
```
maiamcc: ...huh?

PR #550, opened 2018-10-16
tiltfile: kustomize built-in

jazzdan:


internal/tiltfile/tiltfile_test.go
```diff
@@ -317,7 +317,7 @@ func TestGetManifestConfigNonFunction(t *testing.T) {
 func TestGetManifestConfigTakesArgs(t *testing.T) {
 	file := tempFile(
 		`def blorgly2(x):
-			return "foo"
+      return "foo"
```
jazzdan: My editor had some weird interaction with indentation, tabs vs spaces and `gofmt`.

jazzdan:
Whoops, gonna add docs to README also.

internal/tiltfile/tiltfile.go
```diff
@@ -248,6 +259,29 @@ func stopBuild(thread *skylark.Thread, fn *skylark.Builtin, args skylark.Tuple,
 	return buildContext, nil
 }
 
+func callKustomize(thread *skylark.Thread, fn *skylark.Builtin, args skylark.Tuple, kwargs []skylark.Tuple) (skylark.Value, error) {
+	var path string
```
jazzdan: This should probably take a localPath instead of just a string.

jazzdan: I dunno, what do y'all think? I'm not convinced.

nicks: can you use the localPathFromSkylarkValue i recently added?

jazzdan: Yeah I'll do that. I think we should do an audit of the Tiltfile built-ins that we currently have and which ones should take strings and which ones should take localPaths. Maybe only `read_file` should allow a string as a parameter?

PR #552, opened 2018-10-16
tilt: don't set black background in hud

landism:


landism:
I had been tempted to do that, but didn't. Thanks for convincing me - this feels a lot better.

PR #553, opened 2018-10-17
tilt: hud waits for up to start instead of erroring

landism:
theoretically this could have been done by making a custom dialer that retries connections to dead sockets, but that seems way more complex

### DIFF
--- a/internal/cli/hud.go
+++ b/internal/cli/hud.go
@@ -51,18 +51,6 @@ func (c *hudCmd) run(ctx context.Context, args []string) error {
 
 	logOutput(fmt.Sprintf("Starting the HUD (built %s)…\n", buildDateStamp()))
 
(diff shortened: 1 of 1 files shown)


internal/cli/hud.go
```diff
@@ -93,8 +95,17 @@ func connectHud(ctx context.Context) error {
...
 	if err != nil {
-		return err
+		log.Printf("Waiting for `tilt up` to start.")
+
+		for {
+			stream, err = cli.ConnectHud(ctx)
+			if err == nil {
+				break
+			}
+			time.Sleep(time.Second)
```
maiamcc: I might do a smaller sleep? right now ctrl+C on the hud can be unresponsive for up to a second, which feels lousy as a user. (either that or trap SIGINT/SIGTERM? but that feels like overkill...)

PR #556, opened 2018-10-17
engine: hud reflects restart count of pod [ch564]

maiamcc:

Please review the following commits I made in branch maiamcc/restart-count:

9b3e192cbbe522ff77c55aa90357dfcf838417cd (2018-10-17 12:28:38 -0400)
engine: hud reflects restart count of pod [ch564]

### DIFF
--- a/internal/engine/upper.go
+++ b/internal/engine/upper.go
@@ -312,7 +312,7 @@ func (u Upper) handleCompletedBuild(ctx context.Context, engineState *store.Engi
 		ms.LastSuccessfulDeployEdits = ms.CurrentlyBuildingFileChanges
 		ms.CurrentlyBuildingFileChanges = nil
 
-		ms.Pod.OldRestarts = ms.Pod.TotalRestarts // # of pod restarts from old code (shouldn't be reflected in HUD)
+		ms.Pod.OldRestarts = ms.Pod.ContainerRestarts // # of pod restarts from old code (shouldn't be reflected in HUD)
 	}
 
 	if engineState.WatchMounts {
@@ -457,7 +457,6 @@ func handlePodEvent(ctx context.Context, state *store.EngineState, pod *v1.Pod) 
 	// Update the status
 	ms.Pod.Phase = pod.Status.Phase
 	ms.Pod.Status = podStatusToString(*pod)
-	ms.Pod.TotalRestarts = RestartsForPod(pod)
 
 	// Check if the container is ready.
 	cStatus, err := k8s.ContainerMatching(pod, ms.Manifest.DockerRef)
@@ -468,6 +467,7 @@ func handlePodEvent(ctx context.Context, state *store.EngineState, pod *v1.Pod) 
 		return
 	}
 	populateContainerStatus(ctx, ms, pod, cStatus)
+	ms.Pod.ContainerRestarts = int(cStatus.RestartCount)
 }
 
 func handlePodLogAction(state *store.EngineState, action PodLogAction) {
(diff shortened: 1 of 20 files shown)


internal/engine/upper.go
```diff
@@ -313,6 +313,8 @@ func (u Upper) handleCompletedBuild(ctx context.Context, engineState *store.Engi
 		ms.LastBuild = store.NewBuildState(cb.Result)
 		ms.LastSuccessfulDeployEdits = ms.CurrentlyBuildingFileChanges
 		ms.CurrentlyBuildingFileChanges = nil
+
+		ms.Pod.OldRestarts = ms.Pod.TotalRestarts // # of pod restarts from old code (shouldn't be reflected in HUD)
```
maiamcc: so due to a bug, currently we can't container update a previously crashed pod, so this isn't relevant, but theoretically if Pod X had bad code and restarted 5 times, then we update the pod in place with new code, the pod will still think it has 5 restarts, but current code caused none of them, so we want to show the user 0 restarts.

internal/k8s/pod.go
```diff
@@ -105,6 +105,20 @@ func NodeIDFromPod(pod *v1.Pod) NodeID {
 	return NodeID(pod.Spec.NodeName)
 }
 
+func RestartsFromPod(pod *v1.Pod) int {
+	var restarts int32 = 0
+	for _, cStatus := range pod.Status.ContainerStatuses {
+		if cStatus.Name == "tilt-synclet" { // this should be a constant but there's nowhere good to put it :(
```
maiamcc: any thoughts on where to put this? ought to go in `sidecar.go` but then i get circular dependencies with `sidecar` and `k8s` packages :(

landism: for better or worse, I've been trying to avoid adding ways in which `k8s` is tilt-specific, which would lead me to either:
1. make this more generic, so that tilt can do what it needs (e.g., `func RestartsFromPod(pod *v1.Pod, []string containersToIgnore)`, `func RestartsFromContainer(pod *v1.Pod, ContainerName)`, `func RestartsFromPod(pod *v1.Pod) (restartCountByContainerName map[ContainerName]int)`
2. just move the whole function out of `k8s`

I think I'd lean towards #2, since all the signatures in #1 seem kinda ugly. There's not really so much code here that it'd be awful to put it in the engine package.

internal/k8s/pod.go
```diff
@@ -105,6 +105,20 @@ func NodeIDFromPod(pod *v1.Pod) NodeID {
...
+			continue
+		}
+		// If pod has multiple containers, take the max # of restarts
```
landism: what do you think about summing instead of maxing?

internal/engine/upper.go
```diff
@@ -408,6 +411,12 @@ func ensureManifestStateWithPod(state *store.EngineState, pod *v1.Pod) *store.Ma
 			Namespace: ns,
 		}
 	}
+
+	if podID == ms.Pod.PodID {
+		// Always write restart info, it may have changed even if pod has not changed.
+		ms.Pod.TotalRestarts = restarts
```
landism: this should probably be in `handlePodEvent`

internal/hud/renderer.go
```diff
@@ -220,6 +220,9 @@ func renderResource(r view.Resource) rty.Component {
 		l.Add(rty.ColoredString("    K8S: ", cLightText))
 		l.Add(rty.TextString(fmt.Sprintf("Pod [%s] • %s ago — ", r.PodName, formatDuration(time.Since(r.PodCreationTime)))))
 		l.Add(rty.ColoredString(r.PodStatus, podStatusColor))
+		if r.PodRestarts > 0 {
```
landism: it might be good to also show # restarts even when it's 0, but make it gray or green instead of red

this might be bikesheddy and don't bother if you don't feel compelled

internal/k8s/pod.go
maiamcc: ah yyyyep that's totally the behavior of `kubectl get pods` isn't it? will do.

internal/k8s/pod.go
maiamcc: kk. put in engine/pod.go -- not suuuure that's the right place for it?

PR #557, opened 2018-10-17
watch: use `sinceWhen` and `HistoryDone` to avoid spurious events

jazzdan:
TODO: I need to write a test for this new behavior, wanted to get this solution in front of y'all first.

Here's our new watch strategy on Darwin in a nutshell:

1. Create an fsevents stream for events "since" the last event ID that
we saw globally.
2. Add a path that we want to watch
3. Add that path to a map of paths that we're watching _directly_.
4. Restart the event stream to pick up the new path.
5. Ignore all events for all watches until we've seen a `HistoryDone`
event.
6. Ignore the first `ItemCreated` event for paths we're watching that
are also directories
7. Otherwise, forward along all events.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -48,12 +48,12 @@ jobs:
       - checkout
       - restore_cache:
           keys:
-            - homebrew_cache_v3
+            - homebrew_cache_v5
       - run: brew install go@1.10 kustomize # bump cache version when changing this
       - save_cache:
           paths:
             - /usr/local/Homebrew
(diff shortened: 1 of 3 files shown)


jazzdan:
yeah agreed, hence my TODO at the top of this PR. :)

jazzdan:
Hmm I don't think this strategy of using vim to test this directly is going to work. I'm going to rewrite this to do the renames manually.

jazzdan:
Okay I think this is ready for review @nicks @landism.

PR #559, opened 2018-10-17
tilt: shut down hud gracefully

landism:
I'm not excited about storing ctx on `ServerAdapter`, but I was unable to find any way to propagate our context into the grpc server, and the only contexts I could find that get propagated to `stream.Context()` were starting from `context.Background()`. I'd be happy to be mistaken, but I'm not sure it's worth more time.


landism:
This isn't worth looking at yet. I think this is more complicated than it needs to be and instead of sending DoneReply we can just return from the server function.

nicks:
fwiw, when i started setting up the demo, i was surprised at how complex the server/client API is

landism:
ok, this should be good to review now

internal/hud/hud.go
```diff
@@ -24,21 +24,22 @@ type Hud struct {
...
 		r: NewRenderer(),
 	}, nil
 }
 
 func (h *Hud) Run(ctx context.Context, st *store.Store) error {
+	a, err := NewServer(ctx)
+	if err != nil {
+		return err
+	}
+	h.a = a
```
nicks: i don't think this needs to be a field anymore?

internal/engine/upper.go
```diff
@@ -109,20 +109,22 @@ func (u Upper) CreateManifests(ctx context.Context, manifests []model.Manifest,
...
+		close(hudErrorCh)
 	}()
 
 	for {
 		timer := u.timerMaker(refreshInterval)
 
 		select {
 		case <-ctx.Done():
+			// make sure the hud has had a chance to clean up
+			<-hudErrorCh
```
maiamcc: do we need to allow the HUD time to clean up if we return from other place sin this func as well (eg ln 153)?

landism: yep, good catch!

PR #561, opened 2018-10-17
engine: don't image build if config file changes but doesn't result in manifest change

maiamcc:
Hello @jazzdan, @landism,

Please review the following commits I made in branch maiamcc/break-unbreak:

0e4e0c70a5e510dec426d99ca34a0194afed8c81 (2018-10-17 16:52:51 -0400)
engine: don't image build if config file changes but doesn't result in manifest change

### DIFF
--- a/internal/engine/upper.go
+++ b/internal/engine/upper.go
@@ -218,29 +218,30 @@ func (u Upper) maybeStartBuild(ctx context.Context, st *store.Store) {
 		}
 
 		if newManifest.Equal(ms.Manifest) {
-			logger.Get(ctx).Debugf("Manifest %s hasn't changed, not rebuilding", ms.Manifest.Name)
-			matcher, err := ms.Manifest.ConfigMatcher()
+			logger.Get(ctx).Debugf("Detected config change, but manifest %s hasn't changed",
+				ms.Manifest.Name)
+
+			// TODO(maia): if ms.LastError is a bad-config error, clear it [ch603]
+
+			changedFilesWithoutConfigFiles, err := ms.PendingFileChangesWithoutConfigFiles(ctx)
 			if err != nil {
-				logger.Get(ctx).Infof("Error getting config matcher: %v", err)
+				logger.Get(ctx).Infof(err.Error())
 				return
 			}
-			remainingPendingFileChanges := make(map[string]bool)
-			for f, _ := range ms.PendingFileChanges {
-				matches, err := matcher.Matches(f, false)
-				if err != nil {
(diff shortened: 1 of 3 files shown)


internal/engine/upper_test.go
```diff
@@ -547,10 +548,7 @@ func TestNoOpChangeToDockerfile(t *testing.T) {
 	// incremental build.
 	call = <-f.b.calls
 	assert.Equal(t, "foobar", string(call.manifest.Name))
-	assert.ElementsMatch(t, []string{
-		f.JoinPath("Dockerfile"),
-		f.JoinPath("random_file.go"),
-	}, call.state.FilesChanged())
+	assert.ElementsMatch(t, []string{f.JoinPath("random_file.go")}, call.state.FilesChanged())
```
maiamcc: so the result of my magic in upper.go is that if config files change but don't result in a manifest change, i remove config files from `state.FilesChanged()` (so that it just represents files we want to copy/rm from the container). idk if this is shifty -- thoughts?

maiamcc: ...tbh i don't understand why in the old code we don't freak out trying to process `Tiltfile` as a changed file -- but before i modified `ms.PendingFileChanges` builds were failing with:
```
build failed: file /Users/.../servantes/Tiltfile matches no mounts
```
maybe a difference in how we process changed files in image build vs container build? idk...

internal/engine/upper.go
```diff
@@ -218,29 +218,30 @@ func (u Upper) maybeStartBuild(ctx context.Context, st *store.Store) {
...
 
 		if newManifest.Equal(ms.Manifest) {
-			logger.Get(ctx).Debugf("Manifest %s hasn't changed, not rebuilding", ms.Manifest.Name)
-			matcher, err := ms.Manifest.ConfigMatcher()
+			logger.Get(ctx).Debugf("Detected config change, but manifest %s hasn't changed",
+				ms.Manifest.Name)
+
+			// TODO(maia): if ms.LastError is a bad-config error, clear it [ch603]
+
+			changedFilesWithoutConfigFiles, err := ms.PendingFileChangesWithoutConfigFiles(ctx)
```
nicks: how do you know the config files aren't needed in the container?

maiamcc: Legit. I'll make a ticket for fixing this down the line.

maiamcc: [Tracked here](https://app.clubhouse.io/windmill/story/621/configfiles-not-copied-to-container)

PR #562, opened 2018-10-17
tiltfile: read relative to tiltfile dir instead of working dir

nicks:
Hello @jazzdan,

Please review the following commits I made in branch nicks/rel:

9c7473d76e94154269be992404ef1fece87885e9 (2018-10-17 17:35:32 -0400)
tiltfile: read relative to tiltfile dir instead of working dir
We expect that the Tilt demo will have an arbitrary workdir


nicks:
good point! fixed

nicks:
eventually ended up defining my own absPath() for use in Tiltfile, which made things clearer too

PR #564, opened 2018-10-17
tilt: fix esc key, make resources scrollable

landism:
The scrolling has a couple flaws at the moment:
1. The selected resource is always at the top, so if you have 3 resources and you hit down to select the second resource, now you can only see 2 resources.
2. There is no indication of which resource is selected (mostly because while (1) is the case, I'd kinda prefer to hide the fact that scrolling exists).

### DIFF
--- a/internal/hud/hud.go
+++ b/internal/hud/hud.go
@@ -4,6 +4,8 @@ import (
 	"context"
 	"sync"
 
+	"github.com/windmilleng/tcell"
+
 	"github.com/pkg/errors"
 
 	"github.com/windmilleng/tilt/internal/hud/view"
@@ -54,6 +56,7 @@ func (h *Hud) Run(ctx context.Context, st *store.Store) error {
 
 	h.a = a
 
+	var screenEvents chan tcell.Event
 	for {
 		select {
 		case <-ctx.Done():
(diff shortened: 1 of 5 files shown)


internal/hud/view/view.go
```diff
@@ -34,6 +34,7 @@ type Resource struct {
 type ViewState struct {
 	ShowNarration    bool
 	NarrationMessage string
+	SelectedResource int
```
nicks: is this used anywhere?

internal/hud/hud.go
```diff
@@ -83,6 +88,32 @@ func (h *Hud) Close() {
...
+			switch r := ev.Rune(); {
+			case r >= '1' && r <= '9':
+				st.Dispatch(NewShowErrorAction(int(r - '0')))
+			}
+		case tcell.KeyUp:
+			h.r.rty.ElementScroller("resources").UpElement()
+		case tcell.KeyDown:
+			h.r.rty.ElementScroller("resources").DownElement()
+		case tcell.KeyEnter:
+			activeItem := h.r.rty.ElementScroller("resources").GetSelectedIndex()
```
nicks: i guess this is fine for now, but i'm a little bit nervous about how state is flowing thru this system, and how we're digging thru the component tree to read/write state

PR #566, opened 2018-10-18
demo: add a script

nicks:

Please review the following commits I made in branch nicks/ch586/checkout:

554877407d5f6c60ec101d57508e86e5436da17e (2018-10-17 20:04:12 -0400)
demo: add a script

ee3ddef9e82ac192a07dd686d72bdd424fc1e3f8 (2018-10-17 20:04:00 -0400)
demo: make the demo connect to the hud directly

6ab36f26c6af4a2c657f8c5353f2f870055681e1 (2018-10-17 20:04:00 -0400)
tiltfile: read relative to tiltfile dir instead of working dir
We expect that the Tilt demo will have an arbitrary workdir

7b5b823c90383e74e1f7a5e371b6db95c4688468 (2018-10-17 20:00:32 -0400)
build: fmt.Printf interrupts TTY / HUD. everything needs to go thru logger

### DIFF
--- a/internal/demo/script.go
+++ b/internal/demo/script.go
@@ -3,9 +3,14 @@ package demo
 import (
 	"context"
 	"fmt"
+	"io"
 	"io/ioutil"
 	"os"
+	"os/exec"
+	"path/filepath"
+	"time"
 
+	"github.com/pkg/errors"
 	"github.com/windmilleng/tilt/internal/engine"
 	"github.com/windmilleng/tilt/internal/hud"
 	"github.com/windmilleng/tilt/internal/hud/client"
@@ -13,7 +18,9 @@ import (
 	"github.com/windmilleng/tilt/internal/logger"
 	"github.com/windmilleng/tilt/internal/output"
 	"github.com/windmilleng/tilt/internal/store"
+	"github.com/windmilleng/tilt/internal/tiltfile"
 	"golang.org/x/sync/errgroup"
+	"k8s.io/api/core/v1"
 )
 
 // Runs the demo script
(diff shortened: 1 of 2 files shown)


nicks:
this builds on a bunch of other PRs, but you can see the screencast here
https://asciinema.org/a/MDuD4Uufo4oNaXpZgK2smR5d1

internal/demo/script.go
```diff
@@ -3,28 +3,183 @@ package demo
...
+	return s
+}
+
+type podMonitor struct {
+	podsReady   bool
+	podsReadyCh chan bool
+}
+
+func (m *podMonitor) OnChange(ctx context.Context, store *store.Store) {
+	state := store.RLockState()
```
jazzdan: Why isn't the defer for the unlock right after this call?

internal/demo/script.go
```diff
@@ -3,28 +3,183 @@ package demo
...
+				return nil
+			}
+		}
 	}
 }
 
 func (s Script) Run(ctx context.Context) error {
-	fmt.Println("TODO(nick): implement demo")
+	if !s.env.IsLocalCluster() {
+		_, _ = fmt.Fprintf(os.Stderr, "tilt demo mode only supports docker-for-desktop or Minikube\n")
```
jazzdan: Maybe some additional verbiage here for docker-for-desktop? It seems less obvious and google-eable than Minikube

nicks: good point! i changed it to Docker For Mac

internal/demo/script.go
nicks: ah, i meant to move this into a helper, good catch!

PR #568, opened 2018-10-18
upper: if tiltfile breaks and then unbreaks, clear error [ch603]

maiamcc:
Hello @landism, @jazzdan,

Please review the following commits I made in branch maiamcc/break-unbreak-tiltfile:

da45038a902b0600599552d4b2650defd2f7b142 (2018-10-18 11:31:44 -0400)
upper: if tiltfile breaks and then unbreaks, clear error


internal/engine/upper.go
```diff
@@ -612,17 +617,17 @@ func eventContainsConfigFiles(manifest model.Manifest, e manifestFilesChangedAct
 	return false
 }
 
-func getNewManifestFromTiltfile(ctx context.Context, name model.ManifestName) (model.Manifest, error) {
+func getNewManifestFromTiltfile(name model.ManifestName) (model.Manifest, *newManifestErr) {
 	t, err := tiltfile.Load(tiltfile.FileName, os.Stdout)
 	if err != nil {
-		return model.Manifest{}, err
+		return model.Manifest{}, newManifestErrf(err.Error())
```
maiamcc: teeeechnically we're losing some error history here by not wrapping this... do we care?

internal/engine/upper.go
```diff
@@ -700,3 +705,15 @@ func showError(ctx context.Context, state *store.EngineState, resourceNumber int
...
+
+type newManifestErr struct {
+	s string
+}
+
+func (e *newManifestErr) Error() string { return e.s }
+
+var _ error = &newManifestErr{}
+
+func newManifestErrf(format string, a ...interface{}) *newManifestErr {
```
jazzdan: I think we could remove the assertion above by changing this line to:

```suggestion
func newManifestErrf(format string, a ...interface{}) error {
```

internal/engine/upper.go
```diff
@@ -612,17 +617,17 @@ func eventContainsConfigFiles(manifest model.Manifest, e manifestFilesChangedAct
 	return false
 }
 
-func getNewManifestFromTiltfile(ctx context.Context, name model.ManifestName) (model.Manifest, error) {
+func getNewManifestFromTiltfile(name model.ManifestName) (model.Manifest, *newManifestErr) {
```
jazzdan: Which would mean changing this to an `error` too.

maiamcc: i sorta like having `newManifestError` in the signature so ppl don't ignore it and add a clause later that returns some other `error` -- but also i don't have strong feelings.

PR #575, opened 2018-10-18
tilt: handle winch

landism:
This got super annoying when testing #573, and my fix for #573 probably made window resizing even uglier, so I figured I'd do this real quick.


internal/hud/hud.go
```diff
@@ -81,6 +81,8 @@ func (h *Hud) Run(ctx context.Context, st *store.Store, refreshRate time.Duratio
 			if err != nil {
 				return err
 			}
+		case <-a.winchCh:
```
dbentley: As discussed aloud, tcell.NewScreenFromTty lets you pass in a chan for winch so you get it as an event.

PR #578, opened 2018-10-18
hud: always reset when closing

nicks:
Hello @landism, @maiamcc,

Please review the following commits I made in branch nicks/reset:

8ec1200b7f898879e5c486c020b5ec5832072907 (2018-10-18 16:42:49 -0400)
hud: always reset when closing


nicks:
this fixes a bug where the demo would finish, and call cancel(), but the terminal would end in a weird state

PR #580, opened 2018-10-18
model (manifest.Validate): check that all mount.LocalPaths are absPaths

maiamcc:
Hello tbd,

Please review the following commits I made in branch maiamcc/move-mount-validate:

f01ef77e4015e225e53bf5e515e02d0dd6a2c44f (2018-10-18 17:45:15 -0400)
model (manifest.Validate): check that all mount.LocalPaths are absPaths


maiamcc:
So according to how we process the Tiltfile currently, i... don't think this case can ever happen? is it even worth having this check?

maiamcc:
(this PR is prep for [ch621](https://app.clubhouse.io/windmill/story/621/configfiles-not-copied-to-container) -- want to split it up into multiple PRs to reduce size)

internal/build/path_mapping.go
```diff
@@ -81,35 +84,24 @@ func FilterMappings(mappings []pathMapping, matcher model.PathMatcher) ([]pathMa
...
-	pms, err := filesToPathMappings(files, mounts)
-	if err != nil {
-		return pms, err
-	}
-	return pms, nil
-}
-
-func filesToPathMappings(files []string, mounts []model.Mount) ([]pathMapping, *PathMappingErr) {
+// associates local filepaths with their mounts and destination paths). If a file does
+// not belong to a mapping, ignore it.
```
nicks: this behavior change seems wrong? this has caught a lot of legitimate problems. can you give an example of a case where we wouldn't want this to be an error?

internal/build/path_mapping.go
```diff
@@ -118,10 +110,12 @@ func fileToPathMapping(file string, mounts []model.Mount) (pathMapping, *PathMap
...
 				LocalPath:     file,
 				ContainerPath: filepath.Join(m.ContainerPath, relPath),
-			}, nil
+			}
 		}
 	}
-	return pathMapping{}, pathMappingErrf("file %s matches no mounts", file)
+
+	// File didn't match any mounts
+	return pathMapping{}
```
nicks: ditto here

internal/build/path_mapping.go
maiamcc: aah I was imagining passing a Tiltfile along as a changed file -- doesn't belong to a mount, but that's an expected case. I didn't realize this had caught legitimate problems tho, I thought this case never got tripped b/c of the way we set up watches. will rework.

maiamcc:
@nicks reverted -- does this look okay now? (just moving the check for mount paths being absolute)

PR #586, opened 2018-10-19
engine: rewrite watch to be reactive

jazzdan:
Question: does my internal state in watchmanager need a mutex?

TODO: write some more tests


internal/engine/upper.go
```diff
@@ -44,7 +44,7 @@ const maxChangedFilesToPrint = 5
 // Upper seems like a poor and undescriptive name.
 type Upper struct {
 	b                   BuildAndDeployer
-	fsWatcherMaker      fsWatcherMaker
+	fsWatcherMaker      FsWatcherMaker
```
nicks: i think you can remove this field?

internal/engine/watchmanager.go
```diff
@@ -0,0 +1,154 @@
...
+
+func NewWatchManager(watcherMaker FsWatcherMaker) *WatchManager {
+	return &WatchManager{
+		watches:        make(map[model.ManifestName]manifestNotifyCancel),
+		fsWatcherMaker: watcherMaker,
+	}
+}
+
+func (w *WatchManager) diff(ctx context.Context, state store.EngineState) (setup []model.Manifest, teardown []model.Manifest) {
+	setup = []model.Manifest{}
```
nicks: can you move the RLockState inside here? you don't really need to read the state after this function completes (we can assume the Manifests in setup and teardown are immutable)

internal/engine/watchmanager.go
```diff
@@ -0,0 +1,154 @@
...
+	}
+
+	for _, m := range manifestsToProcess {
+		teardown = append(teardown, m)
+	}
+
+	return setup, teardown
+}
+
+func (w *WatchManager) OnChange(ctx context.Context, st *store.Store) {
```
nicks: to answer your question at the top of the PR: you don't need to add a lock here because the subscriber framework won't call OnChange concurrently

internal/engine/watchmanager.go
```diff
@@ -0,0 +1,154 @@
...
+	state := st.RLockState()
+	defer st.RUnlockState()
+
+	setup, teardown := w.diff(ctx, state)
+
+	for _, m := range teardown {
```
nicks: maybe teardown should be an array of watches, rather than an array of Manifests? it seems weird that you have to look them up again.

internal/engine/watchmanager.go
```diff
@@ -0,0 +1,154 @@
...
+		for _, cf := range manifest.ConfigFiles {
+			err = watcher.Add(cf)
+			if err != nil {
+				st.Dispatch(NewErrorAction(err))
+			}
+		}
+
+		ctx, cancel := context.WithCancel(ctx)
+
+		go func(ctx context.Context, manifest model.Manifest, watcher watch.Notify) {
```
nicks: maybe move this goroutine into a separate function?

internal/engine/watchmanager.go
jazzdan: ah okay, cool

internal/engine/watchmanager.go
jazzdan: Yes sounds good!

internal/engine/watchmanager.go
jazzdan: True but we need the manifest name to subsequently delete the manifestWatchCancel pair from the map.

PR #589, opened 2018-10-19
cli: pull out a 'down' command

nicks:
Hello @jazzdan,

Please review the following commits I made in branch nicks/ch585/cleancmd:

f2262a64de1f498329dae68e4eaa988ec5d01cd3 (2018-10-19 18:45:11 -0400)
cli: pull out a 'down' command


nicks:
i think that's currently impossible to implement, because there's no easy way to distinguish between functions that product manifests and functions that do not

PR #590, opened 2018-10-19
engine: refactor into a reactive builder, so we can build in response to things like pod restarts, etc

nicks:
Hello @landism, @maiamcc,

Please review the following commits I made in branch nicks/reactive:

796185c3234d118064b9ae07667627215373f99c (2018-10-19 19:27:51 -0400)
engine: refactor into a reactive builder, so we can build in response to things like pod restarts, etc


nicks:
this isn't quite where i want it to be yet, but moves us in the right direction

nicks:
@maiamcc of course!

PR #592, opened 2018-10-22
engine: refactor upper tests to use NewUpper

jazzdan:
In support of additional changes to #586.

I also deleted `assertAllHUDUpdatesConsumed` on the advice of @nicks. It doesn't make sense in the future when there are lots of ambient changes state that can cause the HUD to update.


jazzdan:
You'll just have to trust me that `make all` passes. :)

nicks:
:handshake:

PR #594, opened 2018-10-22
upper: tests for config file changes

maiamcc:
Hello @landism, @nicks,

Please review the following commits I made in branch maiamcc/changed-config-file-tests:

fa745a75e5a07b2a4e9d2c551a04392358e2c511 (2018-10-22 12:39:33 -0400)
upper: tests for config file changes

### DIFF
--- a/internal/engine/upper.go
+++ b/internal/engine/upper.go
@@ -228,7 +228,6 @@ func (u Upper) maybeStartBuild(ctx context.Context, st *store.Store) {
 				return
 			}
 			ms.PendingFileChanges = mountedChangedFiles
-			fmt.Println("changed files:", mountedChangedFiles)
 
 			if len(ms.PendingFileChanges) == 0 {
 				// No mounted files changed, no need to build.
--- a/internal/engine/upper.go
+++ b/internal/engine/upper.go
@@ -228,6 +228,7 @@ func (u Upper) maybeStartBuild(ctx context.Context, st *store.Store) {
 				return
 			}
 			ms.PendingFileChanges = mountedChangedFiles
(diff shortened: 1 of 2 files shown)


maiamcc:
oops sorry meant to request review from @jazzdan not @nicks

internal/engine/upper.go
```diff
@@ -228,6 +228,7 @@ func (u Upper) maybeStartBuild(ctx context.Context, st *store.Store) {
 				return
 			}
 			ms.PendingFileChanges = mountedChangedFiles
+			fmt.Println("changed files:", mountedChangedFiles)
```
nicks: :eyes:

internal/engine/upper_test.go
```diff
@@ -33,6 +33,13 @@ import (
 	"github.com/windmilleng/tilt/internal/watch"
 )
 
+const (
+	testTiltfile = `def foobar():
```
jazzdan: maybe a slightly more descriptive name?
```suggestion
	testSimpleTiltfile = `def foobar():
```

jazzdan: I think it'd be cool to have a 'lil library of these: multiple manifests, mounts, runs, etc.

internal/engine/upper.go
maiamcc: 😅

internal/engine/upper_test.go
maiamcc: i agree! but also the way we load manifests is inconsistent in this file -- some we `load` from the tiltfile, and some we assemble piecemeal, and urgh.

PR #598, opened 2018-10-22
engine: move more logic into BuildController, take tiltfile reloading out of the write lock

nicks:
Hello @maiamcc, @jazzdan,

Please review the following commits I made in branch nicks/reactive:

bb3a36120ed02c9b25feeacca72f966e60785a14 (2018-10-22 17:33:40 -0400)
engine: move more logic into BuildController, take tiltfile reloading out of the write lock

### DIFF
--- a/internal/store/engine_state.go
+++ b/internal/store/engine_state.go
@@ -32,16 +32,26 @@ type EngineState struct {
 	// How many builds have been completed (pass or fail) since starting tilt
 	CompletedBuildCount int
 
+	// For synchronizing BuildController so that it's only
+	// doing one action at a time. In the future, we might
+	// want to allow it to parallelize builds better, but that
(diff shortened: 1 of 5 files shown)


internal/store/engine_state.go
```diff
@@ -32,16 +32,26 @@ type EngineState struct {
 	// How many builds have been completed (pass or fail) since starting tilt
 	CompletedBuildCount int
 
+	// For synchronizing BuildController so that it's only
+	// doing one action at a time. In the future, we might
+	// want to allow it to parallelize builds better, but that
+	// would better tools for triaging output to different streams.
```
jazzdan:
```suggestion
	// would require better tools for triaging output to different streams.
```

nicks: done

PR #602, opened 2018-10-23
engine: note expected container/detect when code is running in unexpected container [ch660] [ch661]

jazzdan:
https://app.clubhouse.io/windmill/story/661/detect-when-code-is-running-in-unexpected-container

Currently doesn't do anything. I left a TODO to initiate an image build when we detect this condition.


internal/engine/upper.go
```diff
@@ -448,7 +448,14 @@ func handlePodEvent(ctx context.Context, state *store.EngineState, pod *v1.Pod)
 	} else if cStatus.Name == "" {
 		return
 	}
+
 	populateContainerStatus(ctx, ms, pod, cStatus)
+	if ms.ExpectedContainerId == "" {
+		ms.ExpectedContainerId = ms.Pod.ContainerID
```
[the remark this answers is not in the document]
jazzdan: Cool, that sounds like what @maiamcc proposed originally.

maiamcc:
okay @nicks this is ready for review (🤞). We:
1. capture containerId on image build, and
2. check expectedContainerId against actualContainerId when receiving pod events.

Would particularly love thoughts on race conditions that we're missing 😱

internal/engine/deploy.go
```diff
@@ -91,64 +90,71 @@ func (d *DeployDiscovery) populateDeployInfo(ctx context.Context, image referenc
...
+
+	info.nodeID = nID
+	info.podID = pID
+	info.containerID = cID
+	info.containerName = cName
+
+	return nil
+}
+
+func podInfoForImage(ctx context.Context, kCli k8s.Client, image reference.NamedTagged, ns k8s.Namespace) (k8s.NodeID, k8s.PodID, k8s.ContainerID, k8s.ContainerName, error) {
```
jazzdan: just refactored this to be usable with a deployer.

internal/engine/upper.go
```diff
@@ -541,6 +552,7 @@ func (u Upper) handleInitAction(ctx context.Context, engineState *store.EngineSt
 	watchMounts := action.WatchMounts
 	manifests := action.Manifests
 
+	fmt.Println("we init'd!")
```
nicks: :eyes:

internal/engine/upper.go
```diff
@@ -290,7 +290,7 @@ func handleBuildStarted(ctx context.Context, state *store.EngineState, action Bu
 	state.CurrentlyBuilding = mn
 }
 
-func handleCompletedBuild(ctx context.Context, engineState *store.EngineState, cb BuildCompleteAction) error {
+func (u *Upper) handleCompletedBuild(ctx context.Context, engineState *store.EngineState, cb BuildCompleteAction) error {
```
nicks: hmmmm...i don't think this needs to be a method of Upper anymore?

i'm going to imminently change the API so that the reducers aren't allowed to have access to any state, so making this a method will be impossible soon

PR #608, opened 2018-10-24
tilt: display logs inline in HUD

landism:
image

### DIFF
--- a/internal/store/engine_state.go
+++ b/internal/store/engine_state.go
@@ -98,7 +98,7 @@ type Pod struct {
 	// The log for the previously active pod, if any
 	PreRestartLog []byte
 	// The log for the currently active pod, if any
-	Log []byte
+	CurrentLog []byte
 
 	// Corresponds to the deployed container.
 	ContainerName  k8s.ContainerName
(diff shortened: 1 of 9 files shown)


internal/store/engine_state.go
```diff
@@ -112,6 +112,23 @@ type Pod struct {
...
+// if the current pod has crashed, then just print the current pod
+// if the current pod is live, print the current pod plus the last pod
+func (p Pod) Log() string {
+	var podLog string
+	// if the most recent pod is up, then we want the log from the last run (if any), since it crashed
+	if p.ContainerReady {
+		podLog = string(p.PreRestartLog) + string(p.CurrentLog)
+	} else {
+		// otherwise, the most recent pod has the crash itself, so just return itself
+		podLog = string(append([]byte{}, p.CurrentLog...))
```
nicks: is there a reason you copy this twice? note that the `string()` call already makes a copy

internal/store/engine_state.go
```diff
@@ -206,6 +223,11 @@ func StateToView(s EngineState) view.View {
 			}
 		}
 
+		lastBuildLog := ""
+		if ms.LastBuildLog != nil {
+			lastBuildLog = string(append([]byte{}, ms.LastBuildLog.Bytes()...))
```
nicks: similarly, this would be shorter as ms.LastBuildLog.String()

internal/store/engine_state.go
landism: I'd been thinking of `string()` as a cast, and apparently had not yet felt the need to look up whether go strings are immutable. Thanks!

PR #609, opened 2018-10-24
engine: unexpected container triggers image build [ch663]

maiamcc:


internal/engine/buildcontroller.go
```diff
@@ -76,7 +77,14 @@ func (c *BuildController) needsBuild(ctx context.Context, st *store.Store) (buil
 	}, true
 }
 
+func (c *BuildController) disableForTesting() {
```
jazzdan: I guess this isn't necessary if we're just doing this in testing from the engine package.

PR #621, opened 2018-10-25
tilt: add tilt stream command

landism:


internal/cli/stream.go
```diff
@@ -0,0 +1,102 @@
...
+	return cmd
+}
+
+// blocks until we have a successful connection established
+func connect(ctx context.Context) (<-chan stream.MessageOrError, func() error, error) {
+	cli := stream.NewStreamClient()
+	ch, err := cli.Connect(ctx)
+
+	if err != nil {
+		log.Printf("Waiting for `tilt up` to start.")
```
nicks: should this use the context logger?

internal/cli/stream.go
```diff
@@ -0,0 +1,102 @@
...
+
+		err = processConnection(ctx, ch)
+		if err != nil {
+			if err == context.Canceled {
+				return nil
+			}
+			return errors.Wrap(err, "error handling tilt stream")
+		}
+
+		log.Printf("Connection to tilt completed.\n")
```
nicks: ditto

internal/cli/stream_server_logger.go
```diff
@@ -0,0 +1,28 @@
...
+type StreamWriter struct {
+	s stream.StreamServer
+}
+
+func (sw StreamWriter) Write(b []byte) (int, error) {
+	sw.s.Send(string(b))
+	return len(b), nil
+}
+
+func ContextWithStreamServerLogger(ctx context.Context) (ctx2 context.Context, close func(), err error) {
```
nicks: maybe just delete the named params? they're not adding anything

internal/stream/stream.proto
```diff
@@ -0,0 +1,14 @@
+syntax = "proto3";
```
nicks: should this go in the proto directory?

internal/stream/stream.proto
```diff
@@ -0,0 +1,14 @@
...
+
+option go_package = "github.com/windmilleng/tilt/internal/stream/proto;proto";
+
+service Stream {
+    rpc Connect (ConnectRequest) returns (stream StreamMessage) {}
+}
+
+message ConnectRequest {}
+message StreamMessage {
+    string s = 1;
```
nicks: do we need log levels like we did last time we did this?

internal/stream/socket.go
```diff
@@ -0,0 +1,17 @@
...
+	"path/filepath"
+)
+
+func locateSocket() (string, error) {
+	home := os.Getenv("HOME")
+	if home == "" {
+		return "", errors.New("$HOME environment variable is undefined. unable to find socket for stream")
+	}
+
+	tiltDir := filepath.Join(home, ".tilt")
```
nicks: can we use WindmillDirs to get this, instead of hard-coding it?

internal/cli/stream_server_logger.go
landism: You mean drop the names of the return values and keep the return values themselves?
I added them because I figured the intent of `func()` isn't very clear without a name. I guess I can go the `context.CancelFunc` route.

internal/stream/stream.proto
landism: I'd been figuring we'd just mirror whatever is emitted by the main tilt process, and hadn't considered honoring the -debug/etc on the stream process, but that makes more sense.

internal/cli/stream.go
```diff
@@ -0,0 +1,102 @@
...
+	analyticsService.Incr("cmd.stream", nil)
+	defer analyticsService.Flush(time.Second)
+
+	span, ctx := opentracing.StartSpanFromContext(ctx, "stream")
+	defer span.Finish()
+
+	for {
+		ch, closer, err := connect(ctx)
+		if err != nil {
+			if err == context.Canceled {
```
maiamcc: we've got this check in a few places, shouldn't we just check [here instead](http://bit.ly/2PkTwNW)? (Doesn't need to happen now)

landism:
discussed offline, decided to just build this into the hud rather than muck about with a separate process

PR #623, opened 2018-10-25
model: new type of manifest that is only YAML

jazzdan:
Unified in the appropriate places with the new `WatchableManifest`
interface, that both `model.Manifest` and `model.GlobalYAMLManifest`
implement.

### DIFF
--- a/internal/model/globalyaml.go
+++ b/internal/model/globalyaml.go
@@ -1,0 +1,29 @@
+package model
+
+type YAMLManifest struct {
+	name    ManifestName
+	k8sYaml string
+
+	configFiles []string
+}
+
+func (y YAMLManifest) Dependencies() []string {
+	return y.configFiles
+}
+
+func (y YAMLManifest) ManifestName() ManifestName {
+	return y.name
+}
+
+func (y YAMLManifest) ConfigMatcher() (PathMatcher, error) {
+	configMatcher, err := NewSimpleFileMatcher(y.configFiles...)
+	if err != nil {
+		return nil, err
+	}
+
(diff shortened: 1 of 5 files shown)


jazzdan:
One cleanup task I want to do is make private all the fields on `model.Manifest` that are now exposed via functions.

internal/model/globalyaml.go
```diff
@@ -0,0 +1,29 @@
+package model
+
+type YAMLManifest struct {
```
maiamcc: maybe assert it fulfills the interface so the code is easier to scan?

internal/store/engine_state.go
```diff
@@ -39,6 +39,11 @@ type EngineState struct {
 	BuildControllerActionCount int
 
 	PermanentError error
+
+	// GlobalYAML is a special manifest that has no images, but has dependencies
+	// and a bunch of YAML that is deployed when those dependencies change.
```
maiamcc: nit: s/deployed/applied

internal/model/globalyaml.go
maiamcc: ah I just mean `var _ WatchableManifest = YAMLManifest{}`

jazzdan: ah, that seems superfluous to me when the interface is defined right next to the implementation. That's not explicitly against this principle, but you run the risk of violating it unnecessarily when you try to do that because you're tempted to export interfaces to add compile time checks throughout your codebase.

I don't feel strongly either way though, this method is just a thought technology I'm playing with.

PR #628, opened 2018-10-26
tilt: make hud the default view

landism:
1. `tilt up` now opens the hud instead of the stream
2. `l` now opens a modal with the stream
3. `esc` now closes the active modal instead of the hud, and `q` now quits tilt


internal/cli/up.go
```diff
@@ -96,9 +96,15 @@ func (c *upCmd) run(ctx context.Context, args []string) error {
 		return err
 	}
 
+	l := upper.LogActionLogger(ctx)
+	origCtx := ctx
+	ctx = logger.WithLogger(ctx, l)
```
maiamcc: or... call this `ctxWithLogger` and only use it in the one place? won't this potentially do weird things to the code below if we're mucking with the value of `ctx` in a goroutine?

internal/engine/upper.go
```diff
@@ -92,6 +92,14 @@ func NewUpper(ctx context.Context, b BuildAndDeployer, k8s k8s.Client,
 	}
 }
 
+func (u Upper) LogActionLogger(ctx context.Context) logger.Logger {
```
maiamcc: maybe call this `NewLogActionLogger` or `MakeLogActionLogger`, or else "log" sounds like the verb here (like you are logging an action logger)

internal/hud/renderer.go
```diff
@@ -338,13 +348,13 @@ func (r *Renderer) renderResource(res view.Resource, selected bool) rty.Componen
 	return layout
 }
 
-func (r *Renderer) SetUp(event ReadyEvent, sigwinch chan os.Signal) (chan tcell.Event, error) {
+func (r *Renderer) SetUp() (chan tcell.Event, error) {
 	r.mu.Lock()
 	defer r.mu.Unlock()
 
 	// TODO(maia): pass term name along with ttyPath via RPC. Temporary hack:
 	// get termName from current terminal, assume it's the same 🙈
```
maiamcc: this todo can do away

internal/hud/hud.go
```diff
@@ -83,42 +86,32 @@ func (h *Hud) Run(ctx context.Context, st *store.Store, refreshRate time.Duratio
...
 				return nil
 			}
-		case ready := <-a.readyCh:
-			screenEvents, err = h.r.SetUp(ready, a.winchCh)
-			if err != nil {
-				return err
-			}
-			h.Refresh(ctx)
-		case <-a.winchCh:
+		case <-sigwinchCh:
```
maiamcc: pretty sure that if you're using the regular `tcell.NewScreen` (rather than the hacky `NewScreenFromTty`, which we built) it takes care of refreshing on sigwinch for you so you don't need this channel/case anymore (see `tcell.defaultSigwinch` and any call to `t.sigwinch.Notify`).

internal/hud/view/view.go
```diff
@@ -38,11 +38,23 @@ type ViewState struct {
...
 }
 
 type View struct {
+	Log       string
 	Resources []Resource
 	ViewState ViewState
 }
+
+type LogModal struct {
+	// if non-0, which resource's log is currently shown in a modal (1-based index)
```
maiamcc: 1-based index 😭

landism: yeahhhhhhhh
I originally had 0-based, and then -1 for "no log", but hit a few bugs where the struct hadn't been initialized so it was opening the log for the first resource, and decided to embrace the notion that the zero state of a struct should give reasonable default behavior

internal/hud/hud.go
landism: Ah, I figured tcell could only take care of refreshing tcell itself on sigwinch, since it doesn't have a hook to call back into rty code so that rty can re-render with the new dimensions. It turns out tcell emits resizes as a `tcell.Event`. Thanks!

internal/cli/up.go
landism: yeah, fair

This comment also made me realize that it was a bit dangerous to just slurp all of our logs into the hud logger - if there's a problem with the hud, we won't see those logs.

landism:
> does the place that we capture sigint/sigterm clean up the hud well?

Yes, buuuuut, apparently tcell intercepts sigint/sigterm and just makes them do nothing? I've added ^C to the key switch.

internal/engine/upper.go
```diff
@@ -92,9 +92,21 @@ func NewUpper(ctx context.Context, b BuildAndDeployer, k8s k8s.Client,
...
+func (u Upper) NewLogActionLogger(ctx context.Context) logger.Logger {
+	l := logger.Get(ctx)
+	return logger.NewFuncLogger(l.SupportsColor(), l.Level(), func(level logger.Level, b []byte) error {
+		u.store.Dispatch(LogAction{b})
+		return nil
+	})
+}
+
 func (u Upper) RunHud(ctx context.Context) error {
+	state := u.store.LockMutableState()
```
nicks: this seems...weird? can we have Hud dispatch an action when it starts running and move this into the reducers?

internal/engine/upper_test.go
```diff
@@ -1548,6 +1549,11 @@ func (f *testFixture) WaitUntil(msg string, isDone func(store.EngineState) bool)
 
 		select {
 		case <-ctx.Done():
+			// dump the stacks of all goroutines
+			pprof.Lookup("goroutine").WriteTo(os.Stdout, 1)
```
nicks: did you mean to leave this here?

internal/store/engine_state.go
```diff
@@ -7,6 +7,8 @@ import (
 	"sort"
 	"time"
 
+	"github.com/windmilleng/tilt/internal/testutils/bufsync"
```
nicks: i don't think we want to use testutils in real code?

internal/store/engine_state.go
```diff
@@ -41,10 +43,18 @@ type EngineState struct {
 
 	PermanentError error
 
+	// The user has indicated they want to exit
+	Exit bool
+
+	// The full log stream for tilt. This might deserve gc or file storage at some point.
+	Log *bufsync.ThreadSafeBuffer
```
nicks: shouldn't we just Rlock the state and copy this when we want to read from it? we don't want individual pieces of enginestate to have their own mutexes; it will get confusing and deadlock-prone fast

internal/engine/upper_test.go
landism: Yeah, I found it useful when debugging deadlocked tests. Do you think it's likely to be more noisy than useful?

PR #631, opened 2018-10-26
tilt: change last build duration rendering

landism:
Fixing two problems from the feedback doc
1. "done in 0s" can leave one wondering if it was broken, so let's show fractional seconds for short durations.
2. At a glance, one might think "done in 3s" means it will be done in 3s, instead of it was done in 3s, so let's use some clear past tense.

before:
image
after:
image


internal/hud/renderer.go
```diff
@@ -50,7 +50,13 @@ func formatPreciseDuration(d time.Duration) string {
 		return fmt.Sprintf("%dm", minutes)
 	}
 
-	return fmt.Sprintf("%ds", int(d.Seconds()))
+	seconds := int(d.Seconds())
+	if seconds > 10 {
+		return fmt.Sprintf("%ds", seconds)
+	}
+
+	fractionalSeconds := float64(d.Nanoseconds()) / (1000 * 1000 * 1000)
```
maiamcc: maybe cleaner to do:
```suggestion
	fractionalSeconds := float64(d) / float64(time.Second)
```
? 🤷‍♀️

PR #632, opened 2018-10-26
tilt: suggest manifest names when user specifies invalid name

landism:


internal/tiltfile/tiltfile.go
```diff
@@ -367,7 +368,15 @@ func (tiltfile Tiltfile) getManifestConfigsHelper(manifestName string) ([]model.
...
-		return nil, fmt.Errorf("%v does not define a global named '%v'", tiltfile.filename, manifestName)
+		var globalNames []string
+		for name := range tiltfile.globals {
+			globalNames = append(globalNames, name)
+		}
+
+		return nil, fmt.Errorf(
+			"Tiltfile does not define a global named '%v'. perhaps you want one of: %s",
+			manifestName,
+			strings.Join(globalNames, " "))
```
maiamcc: nit, i'd prefer a newline between these.

PR #634, opened 2018-10-26
tilt: add status bar with error count

landism:
footer at bottom, e.g.:
image


internal/hud/renderer.go
```diff
@@ -115,6 +116,28 @@ func (r *Renderer) layout(v view.View) rty.Component {
...
+	} else {
+		sb.Fg(cBad).Text("✖").Fg(tcell.ColorBlack).Textf(" [%d] errors", errorCount)
+	}
+	line := rty.NewLine()
+	line.Add(sb.Build())
+	return rty.NewFixedSize(rty.Bg(line, tcell.ColorWhiteSmoke), rty.GROW, 1)
+}
+
+func isInError(res view.Resource) bool {
+	return res.LastBuildError != "" || res.PodRestarts > 0 || podStatusColors[res.PodStatus] == cBad
```
maiamcc: having PodRestarts > 0 doesn't necessarily mean something is currently wrong...?

internal/hud/renderer.go
```diff
@@ -115,6 +116,28 @@ func (r *Renderer) layout(v view.View) rty.Component {
...
+	for _, res := range v.Resources {
+		if isInError(res) {
+			errorCount++
+		}
+	}
+	sb := rty.NewStringBuilder()
+	if errorCount == 0 {
+		sb.Fg(cGood).Text("✓").Fg(tcell.ColorBlack).Text(" OK")
+	} else {
+		sb.Fg(cBad).Text("✖").Fg(tcell.ColorBlack).Textf(" [%d] errors", errorCount)
```
maiamcc: nit: "error(s)"

internal/hud/renderer.go
landism: Hmm, I was going for whatever we color red. Yeah, maybe it doesn't make sense to count pod restarts until we have some notion of marking them "read" (this conversation would be *so* much more fun out loud).

internal/hud/renderer.go
landism: pfft, "(s)"s are for lazybones

PR #640, opened 2018-10-29
engine: image build and deployer substeps take an interface

jazzdan:
Mostly renaming and hiding struct fields behind functions. Next PR will pass the global YAML manifest, which now implements, albeit questionably, `buildableImageManifest`.

### DIFF
--- a/internal/engine/image_build_and_deployer.go
+++ b/internal/engine/image_build_and_deployer.go
@@ -20,20 +20,6 @@ import (
 	"k8s.io/api/core/v1"
 )
 
-type imageManifest interface {
-	Validate() error
-}
-
-type buildableImageManifest interface {
-	DockerRef() reference.Named
-	IsStaticBuild() bool
-	StaticDockerfile() string
-	StaticBuildPath() string
-	Mounts() []model.Mount
-	Steps() []model.Cmd
-	Entrypoint() model.Cmd
-}
-
 type deployableImageManifest interface {
 	K8sYAML() string
 	ManifestName() model.ManifestName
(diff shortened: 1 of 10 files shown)


internal/engine/image_build_and_deployer.go
```diff
@@ -20,6 +20,26 @@ import (
 	"k8s.io/api/core/v1"
 )
 
+type imageManifest interface {
```
nicks: hmmmm...this seems like an interface that all manifests should implement, not just imageManifests

jazzdan: oh I meant to remove this, I didn't end up using it.

internal/model/manifest.go
```diff
@@ -77,11 +77,11 @@ func (m Manifest) validate() *ValidateErr {
 		return validateErrf("[validate] manifest missing name: %+v", m)
 	}
 
-	if m.DockerRef == nil {
+	if m.dockerRef == nil {
```
maiamcc: seems like for symmetry, this should be a method call

internal/model/globalyaml.go
```diff
@@ -34,3 +36,17 @@ func (y YAMLManifest) ConfigMatcher() (PathMatcher, error) {
 func (YAMLManifest) LocalRepos() []LocalGithubRepo {
 	return []LocalGithubRepo{}
 }
+
+func (y YAMLManifest) K8sYAML() string {
+	return y.k8sYAML
+}
+
+// TODO(dmiller): not sure if this is right
+func (YAMLManifest) DockerRef() reference.Named {
```
maiamcc: 😢

PR #642, opened 2018-10-29
tiltfile: Tiltfile parse returns global yaml manifest

maiamcc:
Hello @jazzdan, @nicks,

Please review the following commits I made in branch maiamcc/return-global-yaml-manifest:

50dc531489bcbf1339948561774b3d5707ee17eb (2018-10-29 18:18:53 -0400)
tiltfile: Tiltfile parse returns global yaml manifest


internal/model/manifest.go
```diff
@@ -214,7 +215,12 @@ func (m Manifest) Dependencies() []string {
 		deps = append(deps, f)
 	}
 
-	return deps
+	return sliceutils.DedupeStringSlice(deps)
+}
+
+func (m Manifest) WithDependencies(deps []string) Manifest {
```
jazzdan: hmm, I think this should be `WithConfigFiles`. Dependencies aren't an explicit field on the struct, they're calculated from fields. While on globalyaml these are one-to-one, on regular manifests they are composed of many fields.

Just a nit, though.

internal/tiltfile/tiltfile.go
```diff
@@ -382,17 +382,27 @@ func Load(ctx context.Context, filename string) (*Tiltfile, error) {
 // a manifest representing the global yaml
 func (t Tiltfile) GetManifestConfigsAndGlobalYAML(names ...string) ([]model.Manifest, model.YAMLManifest, error) {
 	var manifests []model.Manifest
+
+	globalYAML := model.NewYAMLManifest(model.GlobalYAMLManifestName, t.globalYAMLStr, t.globalYAMLDeps)
```
jazzdan: Actually do we need `WithDependencies` if we move this line down below the for loop and pass in the fully realized list of config files to the `NewYAMLManifest` constructor?

Since this is a new struct my instinct is to make all the fields private, which I haven't done, and try to have as many fields initialized at creation time as possible rather than have builders for all of the fields.

internal/utils/sliceutils/sliceutils.go
```diff
@@ -0,0 +1,13 @@
+package sliceutils
```
jazzdan: I'm not wild about the depth of the package tree here: `internal/utils/sliceutils/sliceutils.go`

Can we put this `sliceutils.go` file right in the `utils` package?

maiamcc: I did that based on feedback I got a while ago from uhhh... Dan or Nick, idr, about worries that a single `utils` package would just balloon -- but i don't feel strongly.

jazzdan: I agree that's a good point, could we make `sliceutils` package at the top level?

internal/model/manifest.go
maiamcc: Nah, good call, will change

internal/tiltfile/tiltfile.go
maiamcc: nah we need `WithDependencies` (or `WithConfigFiles` as the case may be) to modify the computed _resource_ manifests so all treat globalYAML dependencies as their own dependencies. Unless I'm missing something...?

internal/utils/sliceutils/sliceutils.go
maiamcc: sure, will do that

internal/tiltfile/tiltfile_test.go
```diff
@@ -63,7 +63,7 @@ func newGitRepoFixture(t *testing.T) *gitRepoFixture {
 	}
 }
 
-func (f *gitRepoFixture) LoadManifests(names ...string) []model.Manifest {
+func (f *gitRepoFixture) LoadManifests(names ...string) ([]model.Manifest, model.YAMLManifest) {
```
nicks: since this is only used for testing, i'm surprised you didn't keep the interface the same?

it seems weird to have a test function that has 2 return values, but all callers only care about 1.

internal/tiltfile/tiltfile.go
jazzdan: Oh- I misunderstood, for some reason I thought you were modifying the global YAML manifest. Carry on!

internal/tiltfile/tiltfile_test.go
maiamcc: So far callers only care about one return value, but there are gonna be some that care about both... this is fair tho, i might split it out.

PR #647, opened 2018-10-30
engine: global YAML can be updated

jazzdan:
Adds a new subscriber and some actions to detect when the global YAML manifest has changed correctly, and trigger a global YAML build (to be defined) in that case.

### DIFF
--- a/internal/engine/buildcontroller.go
+++ b/internal/engine/buildcontroller.go
@@ -92,7 +92,10 @@ func (c *BuildController) OnChange(ctx context.Context, st *store.Store) {
 
 	go func() {
 		if entry.needsConfigReload {
-			newManifest, err := getNewManifestFromTiltfile(entry.ctx, entry.manifest.Name)
+			newManifest, globalYAML, err := getNewManifestFromTiltfile(entry.ctx, entry.manifest.Name)
+			st.Dispatch(GlobalYAMLManifestReloadedAction{
+				GlobalYAML: globalYAML,
+			})
 			st.Dispatch(ManifestReloadedAction{
 				OldManifest: entry.manifest,
 				NewManifest: newManifest,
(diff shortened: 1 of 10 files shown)


internal/model/globalyaml.go
```diff
@@ -52,3 +52,14 @@ func (YAMLManifest) DockerRef() reference.Named {
 
 	return n
 }
+
+func (y YAMLManifest) Empty() bool {
+	return y.k8sYAML == ""
+}
+
+func (y1 YAMLManifest) Equal(y2 YAMLManifest) bool {
```
nicks: it would probably be better for GlobalYAMLBuildController to do its own equality check with the state that it's copying

internal/engine/upper.go
```diff
@@ -80,6 +80,8 @@ func NewUpper(ctx context.Context, b BuildAndDeployer, k8s k8s.Client,
 	st.AddSubscriber(fwm)
 	st.AddSubscriber(pw)
 	st.AddSubscriber(sw)
+	gybc := NewGlobalYAMLBuildController()
```
nicks: let's dependency inject this with `wire` instead of creating it here

internal/model/globalyaml.go
jazzdan: Like [this](https://github.com/windmilleng/tilt/pull/647/commits/199292617d04f9f070d6ffeaf551da9486d839bd)?

nicks: no, i mean it compares the individual fields it cares about

nicks: the upshot is that this Equality function is not good. 

you even acknowledge that it's not good in the TODO.

the reason it's not good is because it treats this equality function as a general equality function. But in fact it's an equality function that's very narrowly tailored to GlobalYAMLBuildController's needs

jazzdan: yeah, good point :)

internal/engine/upper_test.go
```diff
@@ -1378,6 +1379,31 @@ func TestCompletingUpperClosesHud(t *testing.T) {
 	assert.True(t, f.hud.Closed)
 }
 
+func TestInitWithGlobalYAML(t *testing.T) {
```
maiamcc: this is testing more than just init, either break into multiple tests or rename

internal/model/globalyaml.go
```diff
@@ -52,3 +52,7 @@ func (YAMLManifest) DockerRef() reference.Named {
 
 	return n
 }
+
+func (y YAMLManifest) Empty() bool {
```
maiamcc: is this used?

internal/engine/buildcontroller.go
```diff
@@ -92,7 +92,10 @@ func (c *BuildController) OnChange(ctx context.Context, st *store.Store) {
 
 	go func() {
 		if entry.needsConfigReload {
-			newManifest, err := getNewManifestFromTiltfile(entry.ctx, entry.manifest.Name)
+			newManifest, globalYAML, err := getNewManifestFromTiltfile(entry.ctx, entry.manifest.Name)
+			st.Dispatch(GlobalYAMLManifestReloadedAction{
+				GlobalYAML: globalYAML,
```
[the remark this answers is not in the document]
jazzdan: That's correct: applying the same yaml twice in a row doesn't do anything. However, we will only apply the YAML if it has changed.

internal/engine/global_yaml_buildcontroller.go
```diff
@@ -0,0 +1,70 @@
...
+		return
+	}
+
+	state := st.RLockState()
+	m := state.GlobalYAML
+	st.RUnlockState()
+
+	newK8sEntities := []k8s.K8sEntity{}
+	if m.K8sYAML() != c.lastGlobalYAMLManifest.K8sYAML() {
+		logger.Get(ctx).Debugf("Gotta rebuild/apply the global YAML!")
```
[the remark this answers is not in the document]
jazzdan: Yeah I accidentally left this in.

internal/engine/global_yaml_buildcontroller.go
```diff
@@ -0,0 +1,70 @@
...
+		entities, err := k8s.ParseYAMLFromString(m.K8sYAML())
+		if err != nil {
+			st.Dispatch(NewErrorAction(err))
```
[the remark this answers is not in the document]
jazzdan: Ah, good call!

PR #648, opened 2018-10-30
store: move the main action loop into the store

nicks:
Hello @maiamcc, @jazzdan,

Please review the following commits I made in branch nicks/ch713/store:

6912f36a75326baa0d94d1136d9b496f7bf5f7bb (2018-10-30 15:04:56 -0400)
store: move the main action loop into the store

### DIFF
--- a/internal/store/store.go
+++ b/internal/store/store.go
@@ -16,18 +16,24 @@ type Store struct {
 	actionCh    chan Action
 	mu          sync.Mutex
 	stateMu     sync.RWMutex
+	reduce      Reducer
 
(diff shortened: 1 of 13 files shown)


internal/store/store.go
```diff
@@ -74,6 +75,44 @@ func (s *Store) Close() {
...
+		}
+		s.NotifySubscribers(ctx)
+	}
+}
+
+func (s *Store) maybeFinished() (bool, error) {
+	state := s.RLockState()
+	defer s.RUnlockState()
+
+	if len(state.ManifestStates) == 0 {
```
maiamcc: what does this state indicate?

PR #650, opened 2018-10-30
engine: fix error handling, rename test, remove dead code

jazzdan:


internal/engine/global_yaml_buildcontroller.go
```diff
@@ -34,17 +33,16 @@ func (c *GlobalYAMLBuildController) OnChange(ctx context.Context, st *store.Stor
 
 	newK8sEntities := []k8s.K8sEntity{}
 	if m.K8sYAML() != c.lastGlobalYAMLManifest.K8sYAML() {
-		logger.Get(ctx).Debugf("Gotta rebuild/apply the global YAML!")
 		entities, err := k8s.ParseYAMLFromString(m.K8sYAML())
 		if err != nil {
-			st.Dispatch(NewErrorAction(err))
```
nicks: shouldn't we surface this in the HUD ui somehow? we should have some action that sends the error to enginestate

internal/engine/global_yaml_buildcontroller.go
```diff
@@ -62,7 +60,7 @@ func (c *GlobalYAMLBuildController) OnChange(ctx context.Context, st *store.Stor
 
 		err = c.k8sClient.Upsert(ctx, newK8sEntities)
 		if err != nil {
-			st.Dispatch(NewErrorAction(fmt.Errorf("Unable to upsert global YAML: %v", err)))
+			logger.Get(ctx).Infof("Error upserting global_yaml: %v", err)
```
nicks: ditto

internal/engine/global_yaml_buildcontroller.go
jazzdan: Yeah I'm working on HUD integration now, and I'll do that as part of that work.

PR #652, opened 2018-10-30
engine: detect minikube ip

nicks:
Hello @maiamcc, @landism,

Please review the following commits I made in branch nicks/ch709/minikube:

778d9feb60854fb5b412ce773c9f334d4e69f5d9 (2018-10-30 18:17:08 -0400)
engine: detect minikube ip

a115596700b9d1ff2d85f693106f7775036fe45e (2018-10-30 17:41:40 -0400)
k8s: delete some dead code

### DIFF
--- a/internal/engine/upper.go
+++ b/internal/engine/upper.go
@@ -124,7 +124,7 @@ var UpperReducer = store.Reducer(func(ctx context.Context, state *store.EngineSt
 	case PodChangeAction:
 		handlePodEvent(ctx, state, action.Pod)
 	case ServiceChangeAction:
-		handleServiceEvent(ctx, state, action.Service)
+		handleServiceEvent(ctx, state, action)
 	case PodLogAction:
 		handlePodLogAction(state, action)
 	case BuildCompleteAction:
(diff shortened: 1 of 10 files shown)


nicks:
(this is the k8s API call that minikube is making internally, and since we already are doing most of the k8s API work i went ahead and switched it to that)

internal/engine/upper.go
```diff
@@ -502,13 +503,7 @@ func handleServiceEvent(ctx context.Context, state *store.EngineState, service *
...
 	}
 
-	url, err := k8s.ServiceURL(service)
-	if err != nil {
-		logger.Get(ctx).Infof("error resolving service %s: %v", manifestName, err)
-		return
-	}
-
-	ms.LBs[k8s.ServiceName(service.Name)] = url
+	ms.LBs[k8s.ServiceName(service.Name)] = action.URL
```
maiamcc: maybe only add this if url is non-nil?

nicks: nope! this is necessary for correctness. for example, if you edit the service yaml

nicks: added a unit test for this case

maiamcc: ah gotcha! Sure. (and i realize elsewhere that we check if the elem in here is nil before adding it to []endpoints so 👌 )

PR #654, opened 2018-10-31
tilt: move hud instance out of upper

landism:


internal/hud/hud.go
```diff
@@ -23,7 +23,7 @@ const DefaultRefreshInterval = 1 * time.Second
 type HeadsUpDisplay interface {
 	store.Subscriber
 
-	Run(ctx context.Context, st *store.Store, refreshRate time.Duration) error
+	Run(ctx context.Context, dispatch func(action store.Action), refreshRate time.Duration) error
```
nicks: this seems like a weird pattern. is there a reason why you don't want to pass in the store?

landism: I generally prefer using narrower interfaces when it's simple so that the extent of coupling is evident.

If this takes a store, then the reader of this signature is left to wonder whether it might be mutating state or adding subscribers or anything.

Really, the motivation for this change was that we're now calling this from outside of `Upper`, which means `Upper` would have to expose its store, which means readers of `Upper` would be left with the possibility that any random code outside of `Upper` could be getting the store and mutating state.

nicks: fwiw, the store no longer allows you to mutate the state. 

:man_shrugging: passing the store around is a pretty common pattern, and i expect we will eventually want the other methods

PR #664, opened 2018-11-01
engine: separate tiltfile watcher

jazzdan:
This PR does 2 things:
1. Sets the path of the Tiltfile in the engineState, passed through from the CLI.
2. Sets up a separate watcher to watch that file. In the future it will dispatch a `tiltfileChanged` action or something that will reload the tiltfile and set things on state. As a result this part isn't tested yet, but will be in the next PR.

Future PRs will include:
1. The aforementioned action, and test.
2. A change to the BuildController to not re-execute the tiltfile.


internal/cli/up.go
```diff
@@ -104,8 +110,7 @@ func (c *upCmd) run(ctx context.Context, args []string) error {
 		}
 	}()
 
-	// TODO(maia): send along globalYamlManifest (returned by GetManifest...Yaml above)
-	err = upper.CreateManifests(ctx, manifests, globalYAML, c.watch)
+	err = upper.CreateManifests(ctx, manifests, globalYAML, c.watch, absTfPath)
```
nicks: i'm a little bit worried about how this API is evolving, where you have different sets of arguments that are optional in different cases.

maybe we need to break this up into two separate methods?
upper.Start(tfPath, watchMounts)
upper.StartForTesting(manifests, yaml, watchMounts)
?

internal/engine/watchmanager.go
```diff
@@ -70,11 +71,33 @@ func (w *WatchManager) diff(ctx context.Context, st *store.Store) (setup []Watch
...
 }
 
+func (w *WatchManager) maybeSetupTFWatch(st *store.Store) error {
+	state := st.RLockState()
+	defer st.RUnlockState()
+	if w.tiltfileWatch != nil {
+		return nil
+	}
+
+	watcher, err := w.fsWatcherMaker()
```
nicks: it seems weird to set up the watcher while you have an RLock on the state?

maybe the diff() method could just return something to indicate that this needs to be setup?

jazzdan: Yeah I thought about that but it felt gross. Maybe I can refactor that function a bit. I agree overall.

internal/cli/up.go
jazzdan: Yeah I like moving the parsing of the tiltfile to only one place. I'll make a story for this.

PR #665, opened 2018-11-01
tiltfile: k8s_service takes yaml as an OPTIONAL arg [ch715]

maiamcc:
Hello @nicks, @jazzdan,

Please review the following commits I made in branch maiamcc/yaml-opt-arg:

671c8cda19dbf0dc3d8cb718a863e2b9771edf89 (2018-11-01 18:00:01 -0400)
tiltfile: k8s_service takes yaml as an OPTIONAL arg

### DIFF
--- a/README.md
+++ b/README.md
@@ -147,7 +147,7 @@ class Service
 ```
 
 #### global_yaml
-Call this _on the top level of your Tiltfile_ on a string of YAML.
+Call this _on the top level of your Tiltfile_ with a string of YAML.
 
 We will infer what (if any) of the k8s resources defined in your YAML correspond to `Services` defined elsewhere in your Tiltfile (matching based on the DockerImage ref and on pod selectors). Any remaining YAML is _global YAML_, i.e. YAML that Tilt applies to your k8s cluster independently of any `Service` you define. 
 ```python
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ Tilt reads from a Tiltfile. A simple Tiltfile is below:
 def backend():
   img = static_build('Dockerfile', 'gcr.io/companyname/backend')
   yaml = read_file('backend.yaml')
-  return k8s_service(yaml, img)
+  return k8s_service(img, yaml=yaml)
 ```
 
 ## Optimizing Tilt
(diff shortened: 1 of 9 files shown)


README.md
```diff
@@ -146,12 +146,20 @@ class Service
     """
 ```
 
+#### global_yaml
+Call this _on the top level of your Tiltfile_ on a string of YAML.
```
jazzdan:
```suggestion
Call this _on the top level of your Tiltfile_ with a string of YAML.
```

README.md
```diff
@@ -146,12 +146,20 @@ class Service
...
+
+We will infer what (if any) of the k8s resources defined in your YAML correspond to `Services` defined elsewhere in your Tiltfile (matching based on the DockerImage ref and on pod selectors). Any remaining YAML is _global YAML_, i.e. YAML that Tilt applies to your k8s cluster independently of any `Service` you define. 
+```python
+def global_yaml(yaml: string) -> None
+```
+
 #### k8s_service
 
-Creates a kubernetes service that tilt can deploy using the yaml text and the image passed in.
+Creates a kubernetes service that tilt can deploy using the the image passed in. Optionally, you may also pass the Kubernetes resource YAML. If the YAML is not passed, we expect to be able to extract it from `global_yaml` (see above).
```
jazzdan: Will we error here if we can't find it?

internal/tiltfile/tiltfile.go
```diff
@@ -24,6 +24,9 @@ import (
 
 const FileName = "Tiltfile"
 
+const oldK8sServiceSyntaxError = "he syntax for `k8s_service` has changed. Before it was `k8s_service(yaml: string, dockerImage: Image)`. " +
```
jazzdan:
```suggestion
const oldK8sServiceSyntaxError = "The syntax for `k8s_service` has changed. Before it was `k8s_service(yaml: string, dockerImage: Image)`. " +
```

README.md
maiamcc: good Q. i believe we won't error right here, but later on, when we call manifest.Validate()

nicks:
do you know what the error looks like if you build an image, but it's not in the local or global yaml? i forget if we have good tests for that case.

nicks:
also, if this gets committed today, i'd love to do the 0.1 release!

maiamcc:
@nicks if we get a manifest with no yaml, that error gets surfaced when we call `manifest.Validate()` in whatever BuildAndDeployer we're using -- which isn't covered in tests anywhere, i don't think, cuz pretty sure we don't have tests for the individual BADs?

PR #675, opened 2018-11-06
hud layout

yuindustries:
new layout


internal/hud/renderer.go
```diff
@@ -239,38 +239,15 @@ func abbreviateLog(s string) []string {
 
 func (r *Renderer) renderResource(res view.Resource, rv view.ResourceViewState, selected bool) rty.Component {
 	layout := rty.NewConcatLayout(rty.DirVert)
+	renderResourceSummary(selected, rv, res, layout)
+	renderResourcesK8s(res, layout, rv)
+	renderResourceBuild(res, r, rv, layout)
```
landism: not a big deal either way, but I might prefer to write it like this:
```
layout.Add(renderResourceSummary(selected, rv, res))
layout.Add(renderResourcesK8s(res, rv))
layout.Add(renderResourceBuild(res, r, rv))
```

so that:
1. the methods are just constructing things rather than mutating
2. if you want to unit test one of these methods, you can just pass in the resource and resource view (or whatever) and check the resulting layout, without having to give it a layout

internal/hud/renderer.go
```diff
@@ -239,38 +239,15 @@ func abbreviateLog(s string) []string {
...
-	if len(res.DirectoriesWatched) > 0 {
-		var dirs []string
-		for _, s := range res.DirectoriesWatched {
-			dirs = append(dirs, fmt.Sprintf("%s/", s))
-		}
-		sb := rty.NewStringBuilder()
-		sb.Fg(cLightText).Textf("  (Watching %s)", strings.Join(dirs, " ")).Fg(tcell.ColorDefault)
-		layout.Add(sb.Build())
-	}
+func renderResourceBuild(res view.Resource, r *Renderer, rv view.ResourceViewState, layout *rty.ConcatLayout) {
```
landism: maybe make `r *Renderer` a receiver rather than an argument, for consistency with the other funcs in this file
(or instead of either, make this method take a clock or a spinner so that it doesn't need `r`)

internal/hud/renderer.go
```diff
@@ -340,24 +314,49 @@ func (r *Renderer) renderResource(res view.Resource, rv view.ResourceViewState,
...
+	if selected && rv.IsCollapsed {
+		p = "▶ "
+	}
 
-	// Kubernetes Info ---------------------------------------
+	sb.Text(p)
+	sb.Textf("%s ", res.Name)
+
+	l.Add(sb.Build())
+	l.Add(rty.NewFillerString('╌'))
```
landism: this is bikesheddy and I'm not sure how to have a constructive conversation about this, but curious if you have more concrete thoughts than I do on broken lines vs continuous lines

internal/rty/layouts.go
```diff
@@ -181,6 +181,28 @@ func (l *Line) Render(w Writer, width int, height int) error {
...
+}
+
+func (f *FillerString) Render(w Writer, width int, height int) error {
+	for i := 0; i < width; i++ {
+		w.SetContent(i, 0, f.ch, nil)
+	}
+	return nil
+}
+
+var _ Component = &FillerString{}
```
landism: fwiw, I've been trying to keep these immediately under the struct definitions, just so they're predictably locatable

PR #677, opened 2018-11-06
engine: when tiltfile changes re execute all the manifests

jazzdan:
Note there are a couple tests that fail, I think I'm just going to rewrite them? For the record they are:

* `TestRebuildDockerfileFailed`
* `TestBreakManifest`
* `TestBreakAndUnbreakManifestWithNoChange`
* `TestBreakAndUnbreakManifestWithChange`

Wanted to get folk's eyes on this while I work on that solicit ideas for what could be wrong with those tests in the meantime. :)

### DIFF
--- a/internal/engine/tiltfile_watcher.go
+++ b/internal/engine/tiltfile_watcher.go
@@ -21,8 +21,8 @@ func NewTiltfileWatcher(watcherMaker FsWatcherMaker) *TiltfileWatcher {
 	}
 }
 
-func (t *TiltfileWatcher) EnableForTesting(enabled bool) {
-	t.disabledForTesting = !enabled
+func (t *TiltfileWatcher) DisableForTesting(disabled bool) {
+	t.disabledForTesting = disabled
 }
 
 func (t *TiltfileWatcher) OnChange(ctx context.Context, st *store.Store) {
--- a/internal/engine/tiltfile_watcher.go
+++ b/internal/engine/tiltfile_watcher.go
@@ -3,6 +3,7 @@ package engine
 import (
 	"context"
 
+	"github.com/windmilleng/tilt/internal/model"
 	"github.com/windmilleng/tilt/internal/store"
 	"github.com/windmilleng/tilt/internal/watch"
 )
@@ -59,7 +60,7 @@ func (t *TiltfileWatcher) setupWatch(path string) error {
 	return nil
 }
 
-func (t *TiltfileWatcher) watchLoop(ctx context.Context, st *store.Store, initManifests []string) {
+func (t *TiltfileWatcher) watchLoop(ctx context.Context, st *store.Store, initManifests []model.ManifestName) {
 	watcher := t.tiltfileWatcher
 	for {
 		select {
--- a/internal/engine/tiltfile_watcher.go
+++ b/internal/engine/tiltfile_watcher.go
@@ -1,0 +1,86 @@
+package engine
+
+import (
+	"context"
+
+	"github.com/windmilleng/tilt/internal/store"
+	"github.com/windmilleng/tilt/internal/watch"
+)
+
+type TiltfileWatcher struct {
+	tiltfilePath       string
+	fsWatcherMaker     FsWatcherMaker
+	disabledForTesting bool
+	tiltfileWatcher    watch.Notify
+}
+
+func NewTiltfileWatcher(watcherMaker FsWatcherMaker) *TiltfileWatcher {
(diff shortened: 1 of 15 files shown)


internal/engine/upper.go
```diff
@@ -405,6 +416,32 @@ func handleGlobalYAMLApplyError(
...
+func handleTiltfileReloaded(
+	ctx context.Context,
+	state *store.EngineState,
+	event TiltfileReloadedAction,
+) {
+	manifests := event.Manifests
+	globalYAML := event.GlobalYAML
+	err := event.Err
+	if err != nil {
+		logger.Get(ctx).Infof("Unable to parse Tiltfile: %v", err)
```
nicks: This probably needs to return after the error?

Can you add a test for this case?

internal/store/engine_state.go
```diff
@@ -54,6 +54,9 @@ type EngineState struct {
 	GlobalYAMLState *YAMLManifestState
 
 	TiltfilePath string
+
+	// InitManifests is the list of manifest names that we were told to init from the CLI.
+	InitManifests []string
```
nicks: Shouldn't this be a list of ManifestNames? This just creates extra work to convert to strings

internal/engine/upper.go
nicks: Also: where does this error appear? We probably want to print the backtrace

internal/store/engine_state.go
jazzdan: Yeah agreed

internal/engine/upper.go
jazzdan: Right now this error just appears in the logs (`L` in the UI).

I don't think we want to return here. I think if you save the Tiltfile with a syntax error `tilt` itself shouldn't stop or crash. What do you think?

nicks: i don't think we're disagreeing?

if you introduce a syntax error in the Tiltfile, we should keep displaying the services that are there. 

right now, you've changed it so that it will delete all the services by setting the list of manifests to empty

i'm suggesting that you return here, skipping the codepath below, so that it doesn't reset the manifests.

i'm happy to write a unit test for the expected behavior if it's unclear.

jazzdan: Ohhh I thought if you `return`ed here it would stop everything. Great, will do that with a test!

jazzdan: Actually I've changed the behavior so that if a rebuild of a manifest fails we set the `LastManifestLoadError` to the Tiltfile execution error on each manifest state. This matches the current behavior, and displays alongside each manifest in the HUD.

internal/engine/upper.go
```diff
@@ -405,6 +422,39 @@ func handleGlobalYAMLApplyError(
...
+	}
+	newDefOrder := make([]model.ManifestName, len(manifests))
+	for i, m := range manifests {
+		state.ManifestStates[m.ManifestName()].LastManifestLoadError = nil
+		oldManifest := state.ManifestStates[m.ManifestName()].Manifest
+
+		newDefOrder[i] = m.ManifestName()
+		if !oldManifest.Equal(m) {
+			newManifestState := store.NewManifestState(m)
+			state.ManifestStates[m.ManifestName()] = newManifestState
```
landism: Replacing the whole `ManifestState` makes me a bit nervous - is there stuff there we explicitly want to throw away, or could we get away with only replacing the existing state's `Manifest` field?

e.g., IIUC, if this event comes in while the manifest is queued with pending changes, we're now discarding that state (along with last build result and pod info etc)

internal/engine/tiltfile_watcher.go
```diff
@@ -0,0 +1,87 @@
...
+			}
+			st.Dispatch(NewErrorAction(err))
+		case <-ctx.Done():
+			return
+		case _, ok := <-watcher.Events():
+			if !ok {
+				return
+			}
+
+			manifests, globalYAML, err := getNewManifestsFromTiltfile(ctx, initManifests)
```
landism: just to be sure I understand - loading new manifests after startup is out of scope for this PR, right?

jazzdan: That's right, but this should enable that work.

internal/engine/upper.go
jazzdan: Good question, I can play around with it! I'm not sure.

jazzdan: Yeah good call this works!

PR #679, opened 2018-11-07
hud: key legend

yuindustries:
I also updated the keys in a way that I consider sensible.
When the modal appears, keys update accordingly!

screen shot 2018-11-07 at 1 36 35 pm
screen shot 2018-11-07 at 1 37 57 pm


internal/hud/hud.go
```diff
@@ -137,35 +132,39 @@ func (h *Hud) handleScreenEvent(ctx context.Context, dispatch func(action store.
 				} else {
 					logger.Get(ctx).Infof("no urls for resource '%s' ¯\\_(ツ)_/¯", selected.Name)
 				}
-			case r == 'q': // [Q]uit
-				h.Close()
-				dispatch(ExitAction{})
-				return true
-			case r == 'l': // [L]og
+			case r == 'p': // [P]od log
```
landism: the current implementation is this shows the build log if the last build failed, or the pod log if it didn't

internal/hud/renderer.go
```diff
@@ -163,19 +176,17 @@ func (r *Renderer) renderResourceLogModal(res view.Resource, background rty.Comp
...
+	return r.renderLogModal(fmt.Sprintf(" POD LOG: %s ", res.Name), s, background)
 }
 
 func (r *Renderer) renderLogModal(title string, s string, background rty.Component) rty.Component {
 	sl := rty.NewTextScrollLayout(logScrollerName)
 	sl.Add(rty.TextString(s))
 	box := rty.NewBox()
 	box.SetInner(sl)
-	box.SetTitle(title)
+	box.SetTitle(fmt.Sprintf(" %s ", title))
```
landism: oh cool, I wanted to do this, but I think it'd be better to do in the box implementation itself

internal/hud/renderer.go
```diff
@@ -111,46 +113,57 @@ func (r *Renderer) layout(v view.View, vs view.ViewState) rty.Component {
 	split := rty.NewFlexLayout(rty.DirVert)
 
 	split.Add(r.renderResources(v, vs))
-	split.Add(r.renderStatusBar(v))
+	split.Add(r.renderFooter(v))
 	l.Add(split)
 
 	if vs.LogModal.TiltLog {
+		keys = "(esc) to exit view "
```
landism: 1. should we instead be setting `keys` before we call `renderFooter`?
2. global mutable variables (`keys`) are generally not great, and leave us open to all sorts of issues. If we fix (1), then we can just make `keys` a local variable here and pass it to `renderFooter` as a parameter.

internal/hud/renderer.go
yuindustries: I had the same thought! I spent a minute digging into how to make the change in RTY but it wasn't obvious, so let's look at this together.

internal/hud/hud.go
yuindustries: Good point. I'm not entirely happy with the key legend design. But I settled on this for now:

L key — Tilt logs
V key — view log (build or pod)

I don't *love* it but it's no longer actively misleading?

PR #680, opened 2018-11-07
tilt: handle render panics and errors

landism:
Problem:
When RTY panics, we don't reset the terminal and we end up with a stack trace and an error printed in black on white text with no carriage returns and it's really annoying to read.
Also, it turns out, when render returns an error, we just print it to the log, where it's easy to miss

Solution:
in RTY, convert panics into errors
in the hud, errors now cause us to exit with an error message instead of just being hidden in the tilt log


internal/rty/render.go
```diff
@@ -20,7 +20,17 @@ type rty struct {
...
 
-func (r *rty) Render(c Component) error {
+func (r *rty) Render(c Component) (err error) {
+	defer func() {
+		e := recover()
+		if e != nil {
+			var ok bool
+			err, ok = e.(error)
+			if !ok {
+				err = fmt.Errorf("render panicked with non-error: %v", e)
```
nicks: fwiw, i thought we expect most panics to be non-errors?

landism: I have no expectation or baseline for this kind of thing. I thought panics were dirty and returning `error`s was the idiomatic option.
Dan's explanation of the idiom he used in rty is that it's cool to use panics for error propagation within a library, as long as the library owns the call stack and converts the panics back to returning `error`s at the boundaries.
From inspection, all of the existing panics in rty are errors, so this is what I went with. I have no idea how idiomatic that is.

nicks: here's a good playground for playing around with panics.
https://play.golang.org/p/kS4x-wdN_l8

i rarely expect them to be invoked in a library. Rather, I expect panics to be thrown by the runtime (e.g., out of array bounds errors).

https://blog.golang.org/defer-panic-and-recover

nicks: ok

landism: When writing unit tests, I hit a segment violation, which obviously this code turns into an error.

I then tried just "re-panicking" on runtime errors, e.g.:
```
defer func() {
		e := recover()
		if e != nil {
			var ok bool
			runtimeErr, ok := e.(runtime.Error)
			if ok {
				panic(runtimeErr)
			}
	
			err, ok = e.(error)
			if !ok {
				err = fmt.Errorf("render panicked with non-error: %v", e)
			}
		}
	}()
```

Except that's unfortunate because now the stack trace is telling me where my error handling code ran instead of where the error actually occured (like `throw e;` vs `throw;` in Java), and AFAIK there's no way to do this reasonably in Go. I'm inclined to just get rid of the panics and swear off of this pattern Dan chose.

PR #687, opened 2018-11-08
hud: GlobalYAML doesn't have logs

jazzdan:
YAML manifests won't have logs for the foreseeable future so don't present the user with the option.

@yuindustries this might require a change in your keyboard shortcut UI to not indicate to the user that they can press enter to get logs for the global YAML manifest?


internal/hud/hud.go
```diff
@@ -161,8 +161,10 @@ func (h *Hud) handleScreenEvent(ctx context.Context, dispatch func(action store.
 			h.selectedScroller(h.r.rty).Bottom()
 		case tcell.KeyEnter:
 			if !h.currentViewState.LogModal.IsActive() {
-				selectedIdx, _ := h.selectedResource()
-				h.currentViewState.LogModal = view.LogModal{ResourceLogNumber: selectedIdx + 1}
+				selectedIdx, r := h.selectedResource()
+				if !r.IsYAMLManifest {
+					h.currentViewState.LogModal = view.LogModal{ResourceLogNumber: selectedIdx + 1}
+				}
```
landism: I think we want to log an error here, so the user doesn't just think it's a bug.

(and then we'll turn it into a proper error with [ch702](https://app.clubhouse.io/windmill/story/702/hud-error-messages-should-appear-in-the-hud))

landism: yep!

PR #688, opened 2018-11-08
build: add a way to build a cache of directories

nicks:
Hello @jazzdan, @maiamcc,

Please review the following commits I made in branch nicks/ch728/cachebuilder:

f5ec0d2dc4272f44b3712dc71466fc3044f20d04 (2018-11-08 15:45:50 -0500)
build: add a way to build a cache of directories

### DIFF
--- a/internal/build/cache_builder.go
+++ b/internal/build/cache_builder.go
@@ -1,0 +1,125 @@
+package build
+
+import (
+	"context"
+	"crypto/md5"
+	"fmt"
+	"strings"
+
+	"github.com/docker/distribution/reference"
+	"github.com/docker/docker/client"
+	"github.com/pkg/errors"
+	"github.com/windmilleng/tilt/internal/docker"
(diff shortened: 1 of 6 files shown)


internal/build/cache_builder.go
```diff
@@ -0,0 +1,125 @@
...
+	"github.com/docker/docker/client"
+	"github.com/pkg/errors"
+	"github.com/windmilleng/tilt/internal/docker"
+	"github.com/windmilleng/tilt/internal/dockerfile"
+	"github.com/windmilleng/tilt/internal/logger"
+)
+
+// Cache directories are stored at the same image name, but with just the cachedir
+// contents and at a particular tag.
+const CacheTagPrefix = "tilt-cache-"
```
jazzdan: Should this have a version number in case we change the format in the future?

nicks: sure! tho in that case we could just add a version number (and/or bake the version number into the md5 hash)

PR #689, opened 2018-11-08
state/engine: subscribers take read-only store interface (refactor)

maiamcc:
To make it easier to test OnChange methods in isolation (i.e. without needing to call Store.Loop() and verify state changes).


internal/store/store.go
```diff
@@ -5,6 +5,20 @@ import (
...
+type Dispatcher interface {
+	Dispatch(action Action)
+}
+
+type StateReader interface {
+	RLockState() EngineState
+	RUnlockState()
+}
+
+type DispatchingStateReader interface {
```
nicks: i think my main concern with this name is that "Store" is a very common name. i didn't invent it. 
https://redux.js.org/api/store

and that inventing the new name DispatchingStateReader makes this api harder to grok and harder to lookup information on, i.e., hostile to readers

maiamcc: ...fair. Is that a reason to not land this PR? Any other ideas? I'm coming up blank.

jazzdan: `DispatchingStoreReader`?

nicks: no, not a reason not to land this PR, tho i feel like we can come up with a better name. Maybe `Store` is the read-only interface with Dispatch/RLockState, and `SubscribeStore` is the struct with AddSubscriber methods?

nicks: also, for future reference re: "Is that a reason to not land this PR?", i usually approve the PR if all comments are obviously addressable and/or non-blocking, which i remembered to do in this case!

PR #691, opened 2018-11-08
tiltfile: add build_args param to static_build

jazzdan:
In a subsequent PR I will use the new `StaticBuildArgs` field on `model.Manifest` to apply the build args in the image builder.

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,16 +72,17 @@ written in a Mill, a dialect of python. It's based on [starlark](https://github.
 
 Mill comes with built-in functions.
 
-#### static_build(dockerfile, ref, context?)
+#### static_build(dockerfile, ref, build_args?, context?)
 Builds a docker image.
 
 ```python
-def static_build(dockerfile: str, ref: str, context: str = "") -> Image:
+def static_build(dockerfile: str, ref: str, build_args: Dict[str, str] = {}, context: str = "") -> Image:
       """Builds a docker image.
 
     Args:
       dockerfile: The path to a Dockerfile
       ref: e.g. a blorgdev/backend or gcr.io/project-name/bucket-name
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -72,16 +72,17 @@ written in a Mill, a dialect of python. It's based on [starlark](https://github.
...
 
 ```python
-def static_build(dockerfile: str, ref: str, context: str = "") -> Image:
+def static_build(dockerfile: str, ref: str, build_args: Dict[str, str] = {}, context: str = "") -> Image:
       """Builds a docker image.
 
     Args:
       dockerfile: The path to a Dockerfile
       ref: e.g. a blorgdev/backend or gcr.io/project-name/bucket-name
+      build_args?: the build-time variables that are accessed like regular environment variables in the `RUN` instruction of the Dockerfile.
```
maiamcc: fwiw this isn't totally clear to me and could maybe do with an example?

internal/tiltfile/tiltfile.go
```diff
@@ -104,6 +128,19 @@ func (t *Tiltfile) makeStaticBuild(thread *skylark.Thread, fn *skylark.Builtin,
 		return nil, fmt.Errorf("Argument 0 (dockerfile): %v", err)
 	}
 
+	var sba map[string]string
+	if buildArgs != nil {
+		d, ok := buildArgs.(*skylark.Dict)
+		if !ok {
+			return nil, fmt.Errorf("Argument 2 (build_args): expected dict, got %T", buildArgs)
```
maiamcc: oh i see, these are 0-indexed.

...but also that's kinda dumb for human readability and i might 1-index them all? eh.

PR #703, opened 2018-11-09
tiltfile: add syntax for cache paths

nicks:
Hello @jazzdan, @maiamcc,

Please review the following commits I made in branch nicks/ch728/cache:

d4cd1f0bdbb2f6c6c0beb28d50f05d64c5319a7f (2018-11-09 16:50:05 -0500)
tiltfile: add syntax for cache paths

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,6 +57,7 @@ def backend():
       trigger=['package.json'])
   run('go install github.com/companyname/backend/server')
   img = stop_build()
+  img.cache('/root/.cache/go-cache')
 
   yaml = read_file('backend.yaml')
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -178,6 +179,16 @@ Represents a built Docker image
 
 ```python
 class Image
+  def cache(path: string):
```
jazzdan: I've been trying to specify when functions don't return anything. If we don't want to do that we should remove the other references to `None` in the README.

```suggestion
  def cache(path: string)  -> None:
```

nicks: done!

PR #705, opened 2018-11-09
tiltfile/engine: enforce correct parsing of global yaml when loading manifests [ch776]

maiamcc:
Tests coming soon but wanted to push the bulk of this up before I headed out.

### DIFF
--- a/internal/tiltfile/tiltfile.go
+++ b/internal/tiltfile/tiltfile.go
@@ -454,9 +454,10 @@ func handleSkylarkErr(thread *skylark.Thread, err error) error {
 	return err
 }
 
-// GetManifestConfigsAndGlobalYAML executes the Tiltfile to create manifests for all resources and
-// a manifest representing the global yaml.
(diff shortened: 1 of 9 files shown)


internal/tiltfile/tiltfile.go
```diff
@@ -454,17 +454,18 @@ func handleSkylarkErr(thread *skylark.Thread, err error) error {
 	return err
 }
 
-// GetManifestConfigsAndGlobalYAML executes the Tiltfile to create manifests for all resources and
-// a manifest representing the global yaml.
-func (t Tiltfile) GetManifestConfigsAndGlobalYAML(ctx context.Context, names ...model.ManifestName) ([]model.Manifest, model.YAMLManifest, error) {
+// GetAllManifests executes the Tiltfile to create manifests for all resources and
+// a manifest representing the global yaml. NOTE: If you try to call this on only
```
nicks: hmmm...this seems like a fundamental design flaw in the API.

Can we change LoadTiltfile to take all the manifest names? 

fundamentally, it seems like global_yaml depends on the manifest names passed in, which implies that the manifest names passed in should be a global property of the Tiltfile.

jazzdan:
@maiamcc should we close this?

maiamcc:
Ah yep, with new tiltfile this is redundant.

PR #711, opened 2018-11-12
build: wire up staticBuildArgs through image builder

jazzdan:


internal/build/cache_builder.go
```diff
@@ -84,7 +84,7 @@ func (b CacheBuilder) FetchCache(ctx context.Context, ref reference.Named, cache
 }
 
 // Creates a cache image.
-func (b CacheBuilder) CreateCacheFrom(ctx context.Context, baseDf dockerfile.Dockerfile, sourceRef reference.NamedTagged, cachePaths []string) error {
+func (b CacheBuilder) CreateCacheFrom(ctx context.Context, baseDf dockerfile.Dockerfile, sourceRef reference.NamedTagged, cachePaths []string, buildArgs map[string]string) error {
```
nicks: maybe we should create a named type for this `map`

internal/engine/image_build_and_deployer.go
```diff
@@ -137,7 +137,7 @@ func (ibd *ImageBuildAndDeployer) build(ctx context.Context, manifest model.Mani
 
 		df := ibd.baseDockerfile(manifest, cacheRef)
 		steps := manifest.Steps
-		ref, err := ibd.b.BuildImageFromScratch(ctx, ps, name, df, manifest.Mounts, ignore.CreateBuildContextFilter(manifest), steps, manifest.Entrypoint)
+		ref, err := ibd.b.BuildImageFromScratch(ctx, ps, name, df, manifest.Mounts, ignore.CreateBuildContextFilter(manifest), steps, manifest.Entrypoint, manifest.StaticBuildArgs)
```
nicks: i'm not sure what the intended semantics are here. like, is this if you want to have build args in Dockerfile.base?

maybe add a test for this case?

jazzdan: Yeah good point, I did name this _static_ build args. I'll remove this param.

internal/build/image_builder.go
```diff
@@ -42,8 +42,7 @@ type dockerImageBuilder struct {
 }
 
 type ImageBuilder interface {
-	// TODO(dmiller): this needs to take and handle buildArgs
-	BuildDockerfile(ctx context.Context, ps *PipelineState, ref reference.Named, df dockerfile.Dockerfile, buildPath string, filter model.PathMatcher) (reference.NamedTagged, error)
+	BuildDockerfile(ctx context.Context, ps *PipelineState, ref reference.Named, df dockerfile.Dockerfile, buildPath string, filter model.PathMatcher, buildArgs map[string]string) (reference.NamedTagged, error)
```
maiamcc: type `buildArgs` not `map[string]string`?

internal/build/image_builder.go
```diff
@@ -82,7 +81,7 @@ func NewDockerImageBuilder(dcli docker.DockerClient, console console.Console, ou
 	}
 }
 
-func (d *dockerImageBuilder) BuildDockerfile(ctx context.Context, ps *PipelineState, ref reference.Named, df dockerfile.Dockerfile, buildPath string, filter model.PathMatcher) (reference.NamedTagged, error) {
+func (d *dockerImageBuilder) BuildDockerfile(ctx context.Context, ps *PipelineState, ref reference.Named, df dockerfile.Dockerfile, buildPath string, filter model.PathMatcher, buildArgs map[string]string) (reference.NamedTagged, error) {
```
maiamcc: same

internal/build/options.go
```diff
@@ -22,3 +24,12 @@ func shouldRemoveImage() bool {
 	}
 	return true
 }
+
+func manifestBuildArgsToDockerBuildArgs(args model.DockerBuildArgs) map[string]*string {
```
maiamcc: this is... confusing, since the type of `manifestBuildArgs` is `DockerBuildArgs`...  i'd rename this func? or maybe since it's only used in the one place, just do it inside the `Options` func?

PR #712, opened 2018-11-12
tiltfile: store all manifest names on tiltfile [ch776]

maiamcc:
Hello @nicks, @jazzdan,

Please review the following commits I made in branch maiamcc/tiltfile-knows-all-manifests:

168497bcd16635bc97b9cc7a752f267b56f8a76f (2018-11-12 18:02:13 -0500)
tiltfile: store all manifest names on tiltfile

480f7e3745a97bced0f34b8ef64d25dcba8edd34 (2018-11-12 14:20:56 -0500)
wip

### DIFF
--- a/internal/engine/buildcontroller.go
+++ b/internal/engine/buildcontroller.go
@@ -86,6 +86,11 @@ func (c *BuildController) OnChange(ctx context.Context, st store.RStore) {
 	if c.disabledForTesting {
 		return
 	}
+
+	state := st.RLockState()
+	allManifestNames := model.MNamesToStrings(state.InitManifests)
+	st.RUnlockState()
+
 	entry, ok := c.needsBuild(ctx, st)
 	if !ok {
 		return
@@ -93,7 +98,7 @@ func (c *BuildController) OnChange(ctx context.Context, st store.RStore) {
 
 	go func() {
 		if entry.needsConfigReload {
-			newManifest, newGlobalYAML, err := getNewManifestFromTiltfile(entry.ctx, entry.manifest.Name)
+			newManifest, newGlobalYAML, err := getNewManifestFromTiltfile(entry.ctx, allManifestNames, entry.manifest.Name)
 			st.Dispatch(GlobalYAMLManifestReloadedAction{
 				GlobalYAML: newGlobalYAML,
 			})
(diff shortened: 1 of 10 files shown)


maiamcc:
This is the first part of a better solution to #705 -- I'd still love it if the Tiltfile could figure out on load what manifests it defined, but it seems like it'd be even hackier to implement that? @nicks in particular lmk what you think.

internal/tiltfile/tiltfile.go
```diff
@@ -403,7 +407,7 @@ func (t *Tiltfile) globalYaml(thread *skylark.Thread, fn *skylark.Builtin, args
 	return skylark.None, nil
 }
 
-func Load(ctx context.Context, filename string) (*Tiltfile, error) {
+func Load(ctx context.Context, allManifestNames []string, filename string) (*Tiltfile, error) {
```
nicks: shouldn't this be a []model.ManifestName?

internal/engine/buildcontroller.go
```diff
@@ -147,9 +152,8 @@ func (c *BuildController) logBuildEntry(ctx context.Context, entry buildEntry) {
 	}
 }
 
-func getNewManifestFromTiltfile(ctx context.Context, name model.ManifestName) (model.Manifest, model.YAMLManifest, error) {
-	// Sends any output to the CurrentBuildLog
-	t, err := tiltfile.Load(ctx, tiltfile.FileName)
+func getNewManifestFromTiltfile(ctx context.Context, allManifestNames []string, name model.ManifestName) (model.Manifest, model.YAMLManifest, error) {
```
nicks: it seems wrong for the first arg to be strings and the second arg be a ManifestName. shouldn't they both be ManifestNames?

jazzdan: why not have `allManifestNames` take a `[]model.ManifestName` and just convert once on the way in when we're parsing the Tiltfile?

nicks:
re: " I'd still love it if the Tiltfile could figure out on load what manifests it defined", are you sure those are the semantics you want? 

for example, if you "tilt up vigoda", i don't think you want to see all the other services.

nicks:
ya, i think that's the right semantics, i don't think we're in disagreement.

jazzdan:
@maiamcc should we close this?

maiamcc:
yep, with new tiltfile, no longer necessary.

PR #716, opened 2018-11-14
rty: fix a rendering error when you try to resize the window to the size of the help box

nicks:
Hello @landism,

Please review the following commits I made in branch nicks/segfaults:

0ce77c63e5443b55137d4c5e084fad3e076b2333 (2018-11-14 14:27:57 -0500)
rty: fix a rendering error when you try to resize the window to the size of the help box
fixes https://app.clubhouse.io/windmill/story/809/if-i-resize-the-window-while-using-tilt-it-quits-with-an-error


nicks:
The test that's failing currently has this comment:

```
	// this is maybe not the behavior we want long-term, but at least this test will tell us if we accidentally fix it?
	// (and test error propagation in the meantime)
```

so i assume that means it's ok to change this? :grimacing:

PR #718, opened 2018-11-15
engine: close previous tiltfile watch goroutine before starting a new one

jazzdan:


jazzdan:
hmm nevermind this doesn't work

jazzdan:
okay this works!

PR #727, opened 2018-11-20
tiltfile: read_file can take a localPath

jazzdan:


internal/tiltfile/tiltfile.go
```diff
@@ -317,19 +317,30 @@ func (t *Tiltfile) absWorkingDir() string {
...
 	err := skylark.UnpackArgs(fn.Name(), args, kwargs, "path", &path)
 	if err != nil {
 		return nil, err
 	}
 
-	path = t.absPath(path)
-	dat, err := ioutil.ReadFile(path)
+	var pathString string
+
+	switch p := path.(type) {
```
nicks: maybe break this into a helper method like we've done elsewhere?

PR #730, opened 2018-11-20
store: keep track of whether a watched path is a dir or a file

jazzdan:
@yuindustries this should let you differentiate files and directories in the view code and render them differently, with or without a trailing slash.

### DIFF
--- a/internal/engine/upper_test.go
+++ b/internal/engine/upper_test.go
@@ -859,7 +859,14 @@ func TestReapOldBuilds(t *testing.T) {
 func TestHudUpdated(t *testing.T) {
 	f := newTestFixture(t)
 	defer f.TearDown()
-	mount := model.Mount{LocalPath: "/go", ContainerPath: "/go"}
+	oldPWD, err := os.Getwd()
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer os.Chdir(oldPWD)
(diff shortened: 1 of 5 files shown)


internal/engine/upper_test.go
```diff
@@ -1436,7 +1443,7 @@ func TestInitSetsTiltfilePath(t *testing.T) {
 		Manifests:    []model.Manifest{},
 		TiltfilePath: "/Tiltfile",
 	})
-	f.WaitUntil("global YAML manifest gets set on init", func(st store.EngineState) bool {
+	f.WaitUntil("tiltfile path gets set on init", func(st store.EngineState) bool {
```
maiamcc: lulz good catch

jazzdan:
@maiamcc

> you're not actually putting these in the hud, correct, just laying the framework?

That's right

> fwiw I find dirsWatched/pathsWatched to be misleading -- it's not the list of WATCHED dirs/paths, but of CHANGED dirs/paths, right?

Those are separate concepts, see `view.Resource.PendingBuildEdits` and friends. `dirsWatched` is the path of directories that we're watching, and they could be `PendingBuildEdits` that are children of those directories.

maiamcc:
Ohhh! Gotcha.

PR #731, opened 2018-11-20
engine: don't bail out if tiltfile isn't present, wait for it to appear

jazzdan:
This _kind of_ works. @maiamcc I think that this ends up encountering the nil pointer reference that we saw with @dbentley when we were experimenting adding/removing resources.

### DIFF
--- a/internal/engine/upper.go
+++ b/internal/engine/upper.go
@@ -104,20 +104,9 @@ func (u Upper) Start(ctx context.Context, args []string, watchMounts bool) error
 		manifestNames[i] = model.ManifestName(a)
 	}
 
-	var manifests []model.Manifest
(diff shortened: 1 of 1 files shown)


internal/engine/upper.go
```diff
@@ -108,9 +104,20 @@ func (u Upper) Start(ctx context.Context, args []string, watchMounts bool) error
 		manifestNames[i] = model.ManifestName(a)
 	}
 
-	manifests, globalYAML, err := tf.GetManifestConfigsAndGlobalYAML(ctx, manifestNames...)
-	if err != nil {
+	var manifests []model.Manifest
```
nicks: maybe break this into a helper function?

PR #735, opened 2018-11-21
engine: slightly better error message if we can't watch k8s resources

jazzdan:


nicks:
(i think it already does this when using minikube because it runs `minikube ip` somewhere in there

nicks:
oh it's here:
https://github.com/windmilleng/tilt/blob/master/internal/minikube/docker.go#L23

jazzdan:
Ah thanks for the heads up @nicks. I created a separate story for that.

PR #737, opened 2018-11-21
docs: add comparison page

jazzdan:
Just a first pass, all feedback welcome!


nicks:
oh, hmmm, i thought that what you were working on was tutorials on how to migrate from Skaffold -> Tilt or Docker-Compose -> Tilt

I'm a little bit nervous about doing a direct product comparison page like "Tilt is better than X on dimension Y because Z". It's really hard to pull off right without the tone coming off as empty puffery.

A better way to phrase this might be something like "Compony A says they moved from B to Tilt because Z"

@dbentley , what do you think?

nicks:
or maybe, telling the story with a personal vibe, like "we were developing with X, but finding it too slow to wait on a container build after every change"

jazzdan:
Either way I think something like this doesn't belong in the documentation so I'm going to close it for now.

PR #738, opened 2018-11-21
Streamline config change flow [ch867]

maiamcc:
1. Watch config files as their own manifest (instead of each manifest watching for its own config files)
2. on any config file change, re-parse tiltfile/regen ALL configs and dispatch the appropriate events.

tests currently pass, but watchManager tests don't give us any useful coverage. we believe the behavior in this PR is covered incidentally but will write more specific tests.

also note that we don't currently support updating the set of watched config files (e.g. if someone adds a `read_file` call to their Tiltfile) -- [this is tracked here](https://app.clubhouse.io/windmill/story/868/update-watcher-when-configfiles-change-i-e-watcher-can-update-existing-watches)


internal/engine/actions.go
```diff
@@ -126,10 +121,17 @@ func NewHudStoppedAction(err error) HudStoppedAction {
 	return HudStoppedAction{err}
 }
 
-type TiltfileReloadedAction struct {
-	Manifests  []model.Manifest
-	GlobalYAML model.YAMLManifest
-	Err        error
+type ConfigsReloadStartedAction struct {
+	FilesChanged map[string]bool
```
jazzdan: It's slightly more memory efficient and idiomatic to represent maps as `map[X]struct{}`.

internal/engine/watchmanager.go
```diff
@@ -11,13 +11,32 @@ import (
 	"github.com/windmilleng/tilt/internal/watch"
 )
 
+// TODO(maia): throw an error if you try to name a manifest this in your Tiltfile?
+const ConfigsManifestName = "_ConfigsManifest"
```
jazzdan: Maybe we could make it an invalid python method name so that it's impossible to do? Not sure how ugly that'd be though.

internal/engine/actions.go
jazzdan: More on this [here](https://dave.cheney.net/2014/03/25/the-empty-struct).

dbentley: FWIW, I was on this side at my last gig, and lost: https://golang.org/doc/effective_go.html#maps

"A set can be implemented as a map with value type bool."

jazzdan: Oh yeah that makes sense. I take it back!

maiamcc: huh okay! but fwiw this is mirroring `EngineState.PendingConfigFileChanges` which is a `map[string]bool`. I'll fix mon if i have time.

PR #741, opened 2018-11-21
store: remove the build queue

nicks:
Hello @maiamcc, @jazzdan,

Please review the following commits I made in branch nicks/ch719/panic:

2574e17666112a24f7178595988f7eec4d8566bb (2018-11-21 18:35:38 -0500)
store: remove the build queue
Builds are kicked off only based on causation state.
Makes it easy to determine why a build is happening.

### DIFF
--- a/internal/engine/buildcontroller.go
+++ b/internal/engine/buildcontroller.go
@@ -21,6 +21,7 @@ type buildEntry struct {
 	ctx          context.Context
 	manifest     model.Manifest
 	buildState   store.BuildState
+	buildReason  store.BuildReason
 	filesChanged []string
 	firstBuild   bool
 }
@@ -32,23 +33,75 @@ func NewBuildController(b BuildAndDeployer) *BuildController {
 	}
 }
 
+// Algorithm to choose a manifest to build next.
+func nextManifestToBuild(state store.EngineState) model.ManifestName {
+	// First, go through all the manifests in order.
+	// If any of them haven't started yet, build them now.
+	for _, mn := range state.ManifestDefinitionOrder {
+		ms, ok := state.ManifestStates[mn]
+		if ok && !ms.StartedFirstBuild {
+			return mn
+		}
+	}
+
+	// Next go through all the manifests, and check:
+	// 1) all pending file changes, and
+	// 2) all pending manifest changes
+	// The earliest one is the one we want.
+	choiceName := model.ManifestName("")
+	earliest := time.Now()
+
+	// always use a stable iteration order
+	for _, mn := range state.ManifestDefinitionOrder {
+		ms, ok := state.ManifestStates[mn]
+		if !ok {
+			continue
+		}
+
+		// Always prioritize builds that crashes and have an out-of-sync.
+		if ms.NeedsRebuildFromCrash {
+			return mn
+		}
+
+		t := ms.PendingManifestChange
+		if t.Before(earliest) && ms.IsPendingTime(t) {
+			choiceName = mn
+			earliest = t
+		}
+
+		spurious, _ := onlySpuriousChanges(ms.PendingFileChanges)
+		if !spurious {
+			for _, t := range ms.PendingFileChanges {
+				if t.Before(earliest) && ms.IsPendingTime(t) {
+					choiceName = mn
+					earliest = t
+				}
+			}
+		}
+	}
+
+	return choiceName
+}
+
 func (c *BuildController) needsBuild(ctx context.Context, st store.RStore) (buildEntry, bool) {
 	state := st.RLockState()
 	defer st.RUnlockState()
 
-	if len(state.ManifestsToBuild) == 0 {
-		return buildEntry{}, false
-	}
-
+	// Don't start the next build until the previous action has been recorded,
+	// so that we don't accidentally repeat the same build.
 	if c.lastActionCount == state.BuildControllerActionCount {
 		return buildEntry{}, false
 	}
 
-	mn := state.ManifestsToBuild[0]
+	mn := nextManifestToBuild(state)
+	if mn == "" {
+		return buildEntry{}, false
+	}
+
 	c.lastActionCount = state.BuildControllerActionCount
 	ms := state.ManifestStates[mn]
 	manifest := ms.Manifest
-	firstBuild := !ms.HasBeenBuilt
+	firstBuild := !ms.StartedFirstBuild
 
 	filesChanged := make([]string, 0, len(ms.PendingFileChanges))
 	for file, _ := range ms.PendingFileChanges {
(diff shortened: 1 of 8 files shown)


internal/store/build_reason.go
```diff
@@ -0,0 +1,24 @@
+package store
+
+type BuildReason int
+
+const BuildReasonNone = BuildReason(0)
+
+const (
+	BuildReasonFlagFiles BuildReason = 1 << iota
```
jazzdan: Maybe this could be slightly more self explanatory. Like: `BuildReasonFlagMountFiles`?

internal/store/build_reason.go
```diff
@@ -0,0 +1,24 @@
...
+	BuildReasonFlagConfig
+	BuildReasonFlagCrash
+	BuildReasonFlagInit
+)
+
+func (r BuildReason) With(flag BuildReason) BuildReason {
+	return r | flag
+}
+
+func (r BuildReason) Has(flag BuildReason) bool {
```
jazzdan: I love this API

internal/engine/buildcontroller.go
```diff
@@ -32,23 +33,73 @@ func NewBuildController(b BuildAndDeployer) *BuildController {
...
+
+	// always use a stable iteration order
+	for _, mn := range state.ManifestDefinitionOrder {
+		ms, ok := state.ManifestStates[mn]
+		if !ok {
+			continue
+		}
+
+		// Always prioritize builds that are crash-rebuilding.
+		if ms.CrashRebuildInProg {
```
maiamcc: hmm, with this slightly different flow i might rename this flag -- `NeedsRebuildFromCrash` or something? this makes it seem like a pod state (in the old world, this was the state of the MANIFEST)

internal/engine/buildcontroller.go
```diff
@@ -32,23 +33,73 @@ func NewBuildController(b BuildAndDeployer) *BuildController {
...
+
+	return choiceName
+}
+
 func (c *BuildController) needsBuild(ctx context.Context, st store.RStore) (buildEntry, bool) {
 	state := st.RLockState()
 	defer st.RUnlockState()
 
-	if len(state.ManifestsToBuild) == 0 {
+	if c.lastActionCount == state.BuildControllerActionCount {
```
maiamcc: Can you explain this line? i found it confusing when I was last in this code

internal/engine/upper.go
```diff
@@ -329,29 +318,9 @@ func handleFSEvent(
 	}
 
 	ms := state.ManifestStates[event.manifestName]
-
 	for _, f := range event.files {
-		ms.PendingFileChanges[f] = true
+		ms.PendingFileChanges[f] = time.Now()
```
maiamcc: nice

internal/store/build_reason.go
```diff
@@ -0,0 +1,24 @@
+package store
+
+type BuildReason int
+
+const BuildReasonNone = BuildReason(0)
+
+const (
+	BuildReasonFlagFiles BuildReason = 1 << iota
+	BuildReasonFlagConfig
+	BuildReasonFlagCrash
```
maiamcc: not necessary, but i'd like this to be a lil more nuanced -- we're not rebuilding b/c the pod crashed, we're rebuilding bc a pod crash may have caused us to lose state...?

there should at least be a comment.

internal/store/build_reason.go
```diff
@@ -0,0 +1,24 @@
...
+	BuildReasonFlagInit
+)
+
+func (r BuildReason) With(flag BuildReason) BuildReason {
+	return r | flag
+}
+
+func (r BuildReason) Has(flag BuildReason) bool {
+	return r&flag == flag
```
maiamcc: what... what is this syntax? o_0

internal/store/engine_state.go
```diff
@@ -111,6 +114,51 @@ func (ms *ManifestState) MostRecentPod() Pod {
...
+		reason = reason.With(BuildReasonFlagInit)
+	}
+	if !ms.CrashRebuildInProg {
+		reason = reason.With(BuildReasonFlagCrash)
+	}
+	return reason
+}
+
+// Whether a change at the current time would trigger a build.
+func (ms *ManifestState) IsPendingTime(t time.Time) bool {
```
maiamcc: these function names aren't immediately apparent to me, fwiw

internal/store/engine_state.go
```diff
@@ -111,6 +114,51 @@ func (ms *ManifestState) MostRecentPod() Pod {
...
+	return !t.IsZero() && t.After(ms.LastBuildStartTime)
+}
+
+func (ms *ManifestState) PendingBuildSince() time.Time {
```
maiamcc: same

internal/store/build_reason.go
nicks: done

internal/store/build_reason.go
nicks: :heart:

internal/engine/buildcontroller.go
nicks: changed it to NeedsRebuildFromCrash

i'm a little bit confused by your comment "in the old world, this was the state of the MANIFEST"...it's always been on the ManifestState and still is afaict?

internal/engine/buildcontroller.go
nicks: added

```
	// Don't start the next build until the previous action has been recorded,
	// so that we don't accidentally repeat the same build.
```

internal/store/build_reason.go
nicks: added a pointer to the docs on engine state about it

internal/store/build_reason.go
nicks: bitwise &
https://golang.org/ref/spec#Arithmetic_operators

it means "is this bit set"

internal/store/engine_state.go
nicks: fair, but i'm not sure i have an idea for a better one. i added comments.

internal/store/engine_state.go
nicks: ditto

internal/engine/buildcontroller.go
maiamcc: i... am no longer sure what I meant by that comment, ignore 😅

internal/store/engine_state.go
maiamcc: nit: a change at the GIVEN time ("current time" implies time.Now)

nicks: great minds!
https://github.com/windmilleng/tilt/blob/master/internal/store/engine_state.go#L134

PR #743, opened 2018-11-26
engine test: get next buildAndDeployer call with a timeout so tests don't hang if broken

maiamcc:
If you break Engine tests, they hang in annoying and hard-to-debug ways.

This should help.

### DIFF
--- a/internal/engine/buildcontroller_test.go
+++ b/internal/engine/buildcontroller_test.go
@@ -18,14 +18,14 @@ func TestBuildControllerOnePod(t *testing.T) {
 	manifest := f.newManifest("fe", []model.Mount{mount})
 	f.Start([]model.Manifest{manifest}, true)
 
-	call := <-f.b.calls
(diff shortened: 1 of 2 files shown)


internal/engine/buildcontroller_test.go
```diff
@@ -54,10 +54,11 @@ func TestBuildControllerTwoPods(t *testing.T) {
...
 
-	call = <-f.b.calls
+	call = f.nextCall()
 	assert.Equal(t, "", call.state.DeployInfo.PodID.String())
 
 	err := f.Stop()
 	assert.NoError(t, err)
 	f.assertAllBuildsConsumed()
 }
+g
```
nicks: O.o

maiamcc: 🤐

maiamcc: well uh, at least tests don't hang!

PR #744, opened 2018-11-27
docs: fix unformatted code bullet, remove TODO

jazzdan:


docs/api/api.py
```diff
@@ -49,9 +49,7 @@ def local_git_repo(path: str) -> Repo:
 def start_fast_build(dockerfile_path: str, img_name: str, entrypoint: str = "") -> None:
   """Initiates a docker image build that supports ``add`` s and ``run`` s, and that uses a cache for subsequent builds.
 
-  TODO(dmiller): explain how this is fast, and image vs container builds?
-
-  TODO(dmiller): explain the concept of the active build
+    See https://docs.tilt.build/fast_build.html
```
jazzdan: Is there a better way to do this from here @nicks?

nicks: `link text <https://docs.tilt.build/fast_build.html>`_

(note the ending underscore) should usually work out of the box

PR #747, opened 2018-11-27
engine: fix regression in crash rebuild and add better testing

nicks:
Hello @maiamcc, @jazzdan,

Please review the following commits I made in branch nicks/crashrebuild:

982bb3ec70110fb0847da3cdc437f8c242d78ee0 (2018-11-27 14:48:37 -0500)
engine: fix regression in crash rebuild and add better testing

### DIFF
--- a/internal/engine/buildcontroller.go
+++ b/internal/engine/buildcontroller.go
@@ -109,8 +109,12 @@ func (c *BuildController) needsBuild(ctx context.Context, st store.RStore) (buil
 	}
 	sort.Strings(filesChanged)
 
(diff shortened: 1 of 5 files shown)


maiamcc:
btw the comment on `ManifestState.NeedsRebuildFromCrash` is no longer accurate -- back when we immediately enqueued a build on noting a new container ID, this field indicated, "rebuild due to new container currently in progress, don't start another one", but now it acts as a flag saying "rebuild needed".

as i'm thinking about it, maybe a more accurate name is `NeedsRebuildFromChangedContainer` or something? idk. this is fine.

nicks:
ok

PR #748, opened 2018-11-27
docs: upgrade docs from Docker Compose

jazzdan:
Skaffold is a TODO but I'm thinking it will follow this format.


jazzdan:
Also I'll run `make docs` later, didn't want to muck up the diff with that for now.

docs/upgrade.md
```diff
@@ -0,0 +1,134 @@
...
+  spoonerisms:
+    build:
+      context: ./spoonerisms
+      dockerfile: ./spoonerisms/Dockerfile
+```
+
+In Tilt you tell us where your Dockerfile is and we infer your build context from that.
+
+```python
+  img = static_build("spoonerisms/Dockerfile", "gcr.io/myproject/spoonerisms")
```
nicks: should we talk about the `context` argument here?

docs/upgrade.md
```diff
@@ -0,0 +1,134 @@
...
+```python
+def spoonerisms():
+  img = static_build("spoonerisms/Dockerfile", "gcr.io/myproject/spoonerisms")
+  yaml = read_file('spoonerisms.yaml')
+  service = k8s_service(img, yaml=yaml)
+  service.port_forward(9006)
+  return service
+```
+
+# From Skaffold
```
nicks: maybe we should make this a separate doc?

one of the big goals of these docs is to have metrics on what people are clicking on, so a separate page for Skaffold would help us know how often people are reading that doc

nicks:
also, you'll want to add this the the TOC so that it shows up in the sidebar

docs/upgrade.md
jazzdan: Since my docker compose config has `context` I think that's a good idea yeah.

PR #749, opened 2018-11-27
tiltfile: GetManifestConfigsAndGlobalYAML doesn't take args, loads `main` automatically [ch776]

maiamcc:


nicks:
some high-level questions:

what do you do if you have multiple services in a Tiltfile but only want to start one? Or have multiple configurations, but only want to pick one? AFAICT we don't support this anymore? are we certain that no existing users need this functionality?

is there a way to make this feature optional, and measure if people are using it? i.e., if there's a 'main', use that?

maiamcc:
i think i can add the ability to specify specific services/combinations pretty easily in another PR -- but since introducing global yaml, we need the a single function that is guaranteed to know about all of the manifests even if the user is only starting certain ones.

if global.yaml contains yaml for manifests foo, bar, and baz, and user wants to start only foo, if tilt has no way of knowing the set of all manifests that exist, it only extracts SOME stuff from global.yaml and ends up applying stuff relevant to bar and baz thinking it's left over / actually global.

nicks:
> if tilt has no way of knowing the set of all manifests that exist, it only extracts SOME stuff from global.yaml and ends up applying stuff relevant to bar and baz thinking it's left over / actually global.

this seems like desirable behavior? like if i'm not working on those services, then it's ok to use the existing image builds.

> i think i can add the ability to specify specific services/combinations pretty easily in another PR

i'm concerned that this is a major backwards-incompatible change, and that we will spend the next two week finding major bugs & usability holes in it. it seems like there are simple ways to make this backwards-compatible so that we can measure the effect, we should do that.

maiamcc:
with new tiltfile, this PR is no longer relevant

PR #750, opened 2018-11-27
hud: don't render spinner for k8s endlessly

dbentley:
Discovered this independently as I was about to demo Tilt.

Proposed fix for ch906


dbentley:
Updated to skip the line for GlobalYAML because we don't watch anything.

PR #755, opened 2018-11-28
hud: refine quit and alert interaction

nicks:

Please review the following commits I made in branch nicks/ch887/keypress:

4c10c433fca19e9f72e0473e2677529931f352a9 (2018-11-28 14:58:45 -0500)
hud: refine quit and alert interaction


nicks:
screenshot from 2018-11-28 14-57-33

yuindustries:
Hmmmm, now that I'm actually testing this I'm concerned that it's now too easy to accidentally exit Tilt if you hit escape/q twice by accident (when over-enthusiastically closing logs?)

This can, however, also be solved by adding an "are you sure you want to quit" step. But more than that, I'm not yet sure how much "accidentally quitting" is a problem.

So, let's merge this in with the caveat that if we ourselves find that we're often accidentally closing Tilt, that we will fix that issue in a separate Task.

nicks:
ya! this is a common annoyance with terminal UIs that i don't have a good answer to. 

this is why (for example) emacs -nw makes you quit with Ctrl-c / Ctrl-q, and vim makes you use `:q`
 
so one possible solution is that Ctrl-q or Ctrl-c is the only way to quit

PR #757, opened 2018-11-28
engine: Tiltfile manifest doesn't cause errors/crashes

jazzdan:
This PR adds a special Tiltfile manifest to the list of resources in the HUD as well as in the engine state. Right now this manifest doesn't do anything but in the future it will be used to represent errors/feedback from the Tiltfile.

Looks like this:
screen shot 2018-11-28 at 5 07 16 pm

Worked on this with @dbentley


internal/engine/fake_notify_test.go
```diff
@@ -0,0 +1,103 @@
+package engine
+
+import (
+	"sync"
+
+	"github.com/windmilleng/tilt/internal/watch"
+)
+
+type fakeMetaWatcher struct {
```
jazzdan: Prior to using this meta watcher we were only sending events to one watch, not all the watches that the watch manager had set up.

internal/engine/upper_test.go
```diff
@@ -1435,10 +1378,7 @@ func TestInitWithGlobalYAML(t *testing.T) {
 func TestInitSetsTiltfilePath(t *testing.T) {
 	f := newTestFixture(t)
 	f.Start([]model.Manifest{}, true)
-	f.store.Dispatch(InitAction{
```
jazzdan: `f.Start` already dispatches the `InitAction`, don't do it twice.

internal/engine/upper.go
```diff
@@ -133,28 +133,36 @@ func loadAndGetManifests(ctx context.Context, manifestNames []model.ManifestName
...
 		manifests, globalYAML, configFiles, err = tf.GetManifestConfigsAndGlobalYAML(ctx, manifestNames...)
 		if err != nil {
-			manifests = []model.Manifest{}
 			globalYAML = model.YAMLManifest{}
 		}
 	}
 
+	manifests = append(
+		manifests,
+		model.Manifest{
```
nicks: i'm very skeptical of this data model. i think it's going to lead to way too many bugs down the road. @landism, what do you think?

internal/engine/upper_test.go
```diff
@@ -161,21 +130,6 @@ func TestUpper_Up(t *testing.T) {
 	assert.Equal(t, gYaml, state.GlobalYAML)
 }
 
-func TestUpper_UpWatchError(t *testing.T) {
```
nicks: why did this go away?

internal/engine/fake_notify_test.go
nicks: would it be better to break this change out into a separate pr?

internal/engine/upper_test.go
jazzdan: ahh thank you, I need to add error support to the new fakeMetaWatcher.

internal/engine/upper.go
jazzdan: Skeptical about what in particular? That we append the TiltfileManifest to the list of regular manifests?

internal/engine/fake_notify_test.go
jazzdan: Yeah that's a good idea.

internal/engine/upper.go
nicks: yes, and just the idea that we model Tiltfile execution this way

jazzdan: Maybe the four of us (you, me, @landism and @dbentley) should get together after standup to talk trough the plan. My understanding from @dbentley is that there will be more such "special manifests in the future.

PR #758, opened 2018-11-28
tilt: fix build status for pending/in progress

landism:


internal/hud/renderer.go
```diff
@@ -311,51 +311,68 @@ func (r *Renderer) resourceTilt(res view.Resource, rv view.ResourceViewState) rt
...
 	indent := strings.Repeat(" ", 8)
-	statusColor := cPending
-	status := r.spinner()
 	duration := formatBuildDuration(res.LastBuildDuration)
-	editsPrefix := ""
-	edits := ""
+
+	// this is a struct instead of individual vars to reduce the chances someone unintentionally mixes state
+	// (e.g., a pending build's duration with a finished build's edits)
+	type buildStatus struct {
```
nicks: maybe break this out into a function that returns buildStatus?

internal/hud/renderer_test.go
```diff
@@ -133,25 +134,47 @@ ERROR: ImageBuild: executor failed running [/bin/sh -c go install github.com/win
...
+				IsYAMLManifest:      true,
 			},
 		},
 	}
-	rtf.run("global yaml manifest", 70, 20, v, vs)
+	rtf.run("global yaml manifest", 70, 20, v, plainVs)
 
-	vs.AlertMessage = "this is only a test"
-	rtf.run("alert message", 70, 20, v, vs)
+	alertVs := plainVs
```
nicks: fwiw, i've been playing around with a convention where each Test* method uses a common viewstate, which is kind of nice

PR #760, opened 2018-11-28
hud: handle inline log wrapping properly

nicks:

Please review the following commits I made in branch nicks/ch912/logs:

cfad62447116e9a25505ececafaf1c17f2068ba9 (2018-11-28 17:44:20 -0500)
hud: handle inline log wrapping properly


nicks:
screenshot from 2018-11-28 17-46-02
screenshot from 2018-11-28 17-45-22

PR #761, opened 2018-11-29
tilt: clean up global yaml error handling

landism:
this addresses [ch920](https://app.clubhouse.io/windmill/story/920/errors-in-global-yaml-are-handled-poorly)

also fixes bad `LastApplyDuration` on parse errors (due to not emitting a `GlobalYAMLApplyStartedAction`)

### DIFF
--- a/internal/engine/global_yaml_buildcontroller.go
+++ b/internal/engine/global_yaml_buildcontroller.go
@@ -32,50 +32,49 @@ func (c *GlobalYAMLBuildController) OnChange(ctx context.Context, st store.RStor
 	m := state.GlobalYAML
 	st.RUnlockState()
 
-	newK8sEntities := []k8s.K8sEntity{}
 	if m.K8sYAML() != c.lastGlobalYAMLManifest.K8sYAML() {
 		c.lastGlobalYAMLManifest = m
 		st.Dispatch(GlobalYAMLApplyStartedAction{})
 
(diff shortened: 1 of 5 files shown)


internal/engine/global_yaml_buildcontroller.go
```diff
@@ -34,23 +34,23 @@ func (c *GlobalYAMLBuildController) OnChange(ctx context.Context, st store.RStor
...
+		c.lastGlobalYAMLManifest = m
+		st.Dispatch(GlobalYAMLApplyStartedAction{})
+
 		entities, err := k8s.ParseYAMLFromString(m.K8sYAML())
 		if err != nil {
 			err = errors.Wrap(err, "Error parsing global_yaml")
 			logger.Get(ctx).Infof(err.Error())
-			c.lastGlobalYAMLManifest = m
-			st.Dispatch(GlobalYAMLApplyError{Error: err})
+			st.Dispatch(GlobalYAMLApplyCompleteAction{Error: err})
```
maiamcc: I'm not wild about the fact that a dev will have to remember to dispatch a `ApplyComplete` action every time they put a new return statement in this code... maybe do this in a defer?

landism: It's not a defer, but I think this looks a whole lot better now. Thanks!

PR #762, opened 2018-11-29
engine: fake watcher sends events to all watches

jazzdan:
Pulled out of #757


jazzdan:
hmm, I thought this was passing but it looks like it's flaky.

jazzdan:
All good now, review away!

internal/engine/fake_notify_test.go
```diff
@@ -0,0 +1,124 @@
...
+	defer w.mu.Unlock()
+	w.subs = append(w.subs, subCh)
+	w.subsErrors = append(w.subsErrors, errorCh)
+	return newFakeWatcher(subCh, errorCh), nil
+}
+
+func (w *fakeMetaWatcher) getSubs() []chan watch.FileEvent {
+	w.mu.Lock()
+	defer w.mu.Unlock()
+	return w.subs
```
nicks: this probably should copy the slice?

internal/engine/fake_notify_test.go
```diff
@@ -0,0 +1,124 @@
...
+}
+
+func (w *fakeMetaWatcher) getSubErrors() []chan error {
+	w.mu.Lock()
+	defer w.mu.Unlock()
+	return w.subsErrors
```
nicks: ditto

internal/engine/fake_notify_test.go
```diff
@@ -0,0 +1,124 @@
...
+}
+
+func (w *fakeMetaWatcher) loop() {
+	for {
+		select {
+		case e, ok := <-w.events:
+			if !ok {
+				for _, sub := range w.getSubs() {
```
nicks: probably more correct to put this in a defer at the top of the function? bc you want to close this in the other `return` case as well

internal/engine/fake_notify_test.go
```diff
@@ -0,0 +1,124 @@
+package engine
+
+import (
+	"sync"
+
+	"github.com/windmilleng/tilt/internal/watch"
+)
+
+type fakeMetaWatcher struct {
```
nicks: maybe MultiWatcher instead of MetaWatcher?

internal/engine/upper_test.go
```diff
@@ -230,6 +199,7 @@ func TestUpper_UpWatchCoalescedFileChanges(t *testing.T) {
 	for _, fileRelPath := range fileRelPaths {
 		f.fsWatcher.events <- watch.FileEvent{Path: fileRelPath}
 	}
+	time.Sleep(time.Millisecond)
```
nicks: ?

this seems like it means you introduced a bug

internal/engine/upper_test.go
```diff
@@ -1600,6 +1561,7 @@ func (f *testFixture) Start(manifests []model.Manifest, watchMounts bool) {
 	f.WaitUntil("manifests appear", func(st store.EngineState) bool {
 		return len(st.ManifestStates) == len(manifests) && st.WatchMounts == watchMounts
 	})
+	time.Sleep(5 * time.Millisecond)
```
nicks: ditto here, this seems like an obvious bug

internal/engine/fake_notify_test.go
```diff
@@ -0,0 +1,124 @@
...
+	return w.outboundCh
+}
+
+func (w *fakeWatcher) loop() {
+	var q []watch.FileEvent
+	for {
+		var outboundCh chan watch.FileEvent
+		var outboundE watch.FileEvent
+		if len(q) > 0 {
+			outboundCh, outboundE = w.outboundCh, q[0]
```
nicks: this loop is really hard to follow.

it might be a lot simpler if you just have two entirely different branches, one if q is empty and one if it's non-empty. then a lot of the weird outbound variables go away

internal/engine/upper_test.go
jazzdan: Yeah what we really need is a bit on the state that says "we have successfully watched the watches." Right now there's a slight delay between when we tell Tilt to start the watches and when they actually start.

I'll make a story for this.

internal/engine/fake_notify_test.go
jazzdan: I took another pass, let me know what you think!

internal/engine/fake_notify_test.go
nicks: i think you may have misunderstood my comment? i'll just send you a pr

jazzdan: Sounds good!

PR #768, opened 2018-11-29
hud: update key legend

yuindustries:
Truncate this a bit more.
- Allow people to infer that (←) closes
- Tilt (l)og can go away once we do the Log Pane


internal/hud/renderer.go
```diff
@@ -95,11 +95,11 @@ func (r *Renderer) maybeAddAlertModal(vs view.ViewState, layout rty.Component) r
 }
 
 func keyLegend(vs view.ViewState) string {
-	defaultKeys := "(↓) next, (↑) prev ┊ (→) expand, (←) collapse, (enter) log, (b)rowser ┊ Tilt (l)og ┊ (q)uit  "
+	defaultKeys := "Browse (↓) (↑), Expand (→) ┊ (enter) log, (b)rowser ┊ Tilt (l)og ┊ (q)uit  "
```
nicks: maybe (↓↑)?

PR #770, opened 2018-11-29
engine: more robust logic about when BaD falls back (and cleaner logging)

maiamcc:
This seemed easier than putting things in a halfway state as I worked
on DB's commit for docker compose stuff.

Instead of checking "ShouldFallBack" via "is it <this specific error type>?",
we explicitly mark errs that should not cause a fallback at the time at which
they are thrown.

Also, this PR impelemnts `ExpectedError` -- if an `error` is thrown but isn't
of concern to the user (e.g. we can't container build this b/c it hasn't been
built yet), we wrap it as an `ExpectedError` and it doesn't get logged (except in
debug mode).

### DIFF
--- a/internal/engine/build_errors.go
+++ b/internal/engine/build_errors.go
@@ -1,0 +1,68 @@
+package engine
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/pkg/errors"
+	"github.com/windmilleng/tilt/internal/model"
+)
+
+// Nothing is on fire, this is an expected case like a container builder being
+// passed a build with no attached container. Don't need to show this to a user.
+type ExpectedError struct {
+	error
+}
+
+func WrapExpectedError(err error) ExpectedError {
+	return ExpectedError{err}
+}
+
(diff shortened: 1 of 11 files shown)


internal/engine/build_errors.go
```diff
@@ -0,0 +1,68 @@
...
+	"context"
+	"fmt"
+
+	"github.com/pkg/errors"
+	"github.com/windmilleng/tilt/internal/model"
+)
+
+// Nothing is on fire, this is an expected case like a container builder being
+// passed a build with no attached container. Don't need to show this to a user.
+type ExpectedError struct {
```
nicks: i feel like ExpectedError is way too generic a name for this

maybe call this something like `ExpectationFailed` or `RedirectToNextBuilder`?

fwiw, i often find https://httpstatusdogs.com/ helpful for thinking thru taxonomies of errors. in this case, this is more like a 3xx (redirect) or 4xx (request can't be satisfied for some reason)

PR #773, opened 2018-11-29
engine: add last tiltfile error to state and HUD

jazzdan:
Looks like this in the HUD now. Definitely room for improvement:

screen shot 2018-11-29 at 5 41 00 pm


internal/store/engine_state.go
```diff
@@ -458,5 +453,9 @@ func StateToView(s EngineState) view.View {
 
 	ret.Log = string(s.Log)
 
+	if s.LastTiltfileError != nil {
+		ret.TiltfileErrorMessage = s.LastTiltfileError.Error()
```
nicks: does this get the right error message if there's a stack trace?

jazzdan: Oooh good call it doesn't!

PR #777, opened 2018-11-30
tilt: properly truncate the EDITS list

landism:
This improves the behavior of this line:
        ● TILT: Pending • EDITS main.go                                                           DURATION 3.80s

Currently, when we run out of horizontal space, we just start dropping stuff with no apparent logic. Sometimes we don't show the edits list at all. Sometimes we show "DURA" and then it gets cut off at the edge of the screen.

Changes:
1. only show the base name instead of the relative path of each file (e.g., Foo.java instead of src/main/java/engineering/windmill/tilt/Foo.java)
2. make the edits list the only truncatable element in this line
3. implement (2) by modifying ConcatLayout to allow adding elements that are filled from the right rather than the left, which means we don't have to use `FillerString`, which has a weird interaction with dynamic ConcatLayout elements


internal/rty/layouts.go
```diff
@@ -124,8 +124,9 @@ type concatLayoutComponent struct {
 }
 
 type ConcatLayout struct {
```
nicks: i'm concerned that this component is already very complex. The semantics are poorly defined (is it a grid? is it a flexbox?). It has historically proven to be very buggy.

This change increases the complexity a lot. I'm certain there are configurations that have bugs.

i think you can revert this whole file, and replace it with a very simple change: if there are no edits, add a dynamic filler string. I think that would solve your problem without having to introduce AddRight. If that doesn't work, let's talk about this in person.

landism: just to capture a bit of offline discussion:
the problem is that what we want is, e.g.:
"TILT: OK .  EDITS: main.go, blah.go (+3 more) .                  DURATION: 3s"

We want "DURATION: 3s" to be right-aligned, and everything else to be left-aligned.
We want the file list ("main.go, blah.go") to be the first thing truncated when we're short on space.
That is, we want:
left: component component truncatable-component component
right: component

It's not clear what general layout concept would support this without getting overly complex, so for now I've just added an EditStatusLine that does what we need.

PR #783, opened 2018-12-03
[docker-compose] listen for dc events, update hud with dc status [ch977]

maiamcc:
include a hacky map from dc event --> probable status.

no real UI thought given to this yet. Looks like this:
dc hud

### DIFF
--- a/internal/dockercompose/event.go
+++ b/internal/dockercompose/event.go
@@ -1,0 +1,97 @@
+package dockercompose
+
+import (
+	"encoding/json"
+	"fmt"
+	"strconv"
+)
+
+type Event struct {
+	Time       string     `json:"time"` // todo: time
+	Type       Type       `json:"type"`
+	Action     Action     `json:"action"`
+	ID         string     `json:"id"` // todo: type?
+	Service    string     `json:"service"`
+	Attributes Attributes `json:"attributes"`
+}
+
+type Attributes struct {
+	Name  string `json:"name"`
+	Image string `json:"image"`
+}
+
+func EventFromJsonStr(j string) (Event, error) {
+	var evt Event
+
+	b := []byte(j)
+	err := json.Unmarshal(b, &evt)
+
+	return evt, err
+}
+
+// https://docs.docker.com/engine/reference/commandline/events/
+type Type string
+
+const (
+	// Add 'types' here (and to `UnmarshalJSON` below`) as we support them
+	TypeContainer Type = "container"
+)
+
+func (t *Type) UnmarshalJSON(b []byte) error {
+	s := string(b)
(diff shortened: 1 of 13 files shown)


internal/dockercompose/event.go
```diff
@@ -0,0 +1,97 @@
...
+	err := json.Unmarshal(b, &evt)
+
+	return evt, err
+}
+
+// https://docs.docker.com/engine/reference/commandline/events/
+type Type string
+
+const (
+	// Add 'types' here (and to `UnmarshalJSON` below`) as we support them
```
jazzdan: I think you accidentally a ` here :P

internal/engine/docker_compose_watch.go
```diff
@@ -0,0 +1,106 @@
...
+	"github.com/windmilleng/tilt/internal/logger"
+	"github.com/windmilleng/tilt/internal/store"
+)
+
+type DockerComposeWatcher struct {
+	watching bool
+}
+
+func NewDockerComposeWatcher() *DockerComposeWatcher {
+	fmt.Println("making a docker-compose watcher hoorayyyy")
```
jazzdan: 👀

internal/engine/upper.go
```diff
@@ -649,6 +653,24 @@ func handleExitAction(state *store.EngineState, action hud.ExitAction) {
 	}
 }
 
+func handleDockerComposeEvent(ctx context.Context, engineState *store.EngineState, action DockerComposeEventAction) {
+	evt := action.Event
+	mn := evt.Service
+	ms, ok := engineState.ManifestStates[model.ManifestName(mn)]
+	if !ok {
+		// No corresponding manifest, nothing to do
+		logger.Get(ctx).Infof("event for unrecognized manifest %s", mn)
```
jazzdan: Should you `return` in this case?

internal/hud/view/view.go
```diff
@@ -31,16 +31,22 @@ type Resource struct {
...
+	// Relevant to k8s resources (maybe should accomplish via interface?)
 	PodName         string
 	PodCreationTime time.Time
 	PodStatus       string
 	PodRestarts     int
 	Endpoints       []string
-	PodLog          string
+
+	// Relevant to docker-compose resources (maybe should accomplish via interface?)
+	IsDCManifest bool
```
jazzdan: Again I would advocate for making this a separate type, but we previously discussed this and you're foregoing that for now.

internal/engine/docker_compose_watch.go
maiamcc: 👀

internal/engine/upper.go
maiamcc: lol yes 😳

PR #785, opened 2018-12-03
hud: collapse by default, auto-expand on error

nicks:

Please review the following commits I made in branch nicks/ch940/collapse:

082a91987719a3a30ff0878f28ce0bd62c2661f1 (2018-12-03 17:26:36 -0500)
hud: collapse by default, auto-expand on error


landism:
> one minor detail is that once you've manually collapsed or expanded something, it's impossible to get it back into the "auto" state. Which might be the right behavior!

yeah, this probably isn't worth worrying about until someone asks

PR #788, opened 2018-12-04
build: support mounting files to directories

jazzdan:
My one big question: is there a better way to indicate that the destination of a mount is a directory other than requiring that a `/` be appended?

Also I'm pretty sure this approach won't work in Windows.

### DIFF
--- a/internal/build/path_mapping.go
+++ b/internal/build/path_mapping.go
@@ -133,9 +133,8 @@ func fileToPathMapping(file string, mounts []model.Mount) (pathMapping, *PathMap
 	return pathMapping{}, pathMappingErrf("file %s matches no mounts", file)
 }
 
-// TODO(dmiller): this won't work in Windows
 func endsWithSlash(path string) bool {
-	return strings.HasSuffix(path, "/")
+	return strings.HasSuffix(path, string(filepath.Separator))
 }
 
 func isFile(path string) (bool, error) {
@@ -144,7 +143,7 @@ func isFile(path string) (bool, error) {
 		return false, err
 	}
 	mode := fi.Mode()
-	return mode.IsRegular(), nil
+	return !mode.IsDir(), nil
 }
 
 func FileBelongsToMount(file string, mounts []model.Mount) bool {
--- a/internal/build/path_mapping.go
+++ b/internal/build/path_mapping.go
@@ -4,7 +4,9 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"path"
 	"path/filepath"
+	"strings"
 
 	"github.com/opentracing/opentracing-go"
 	"github.com/pkg/errors"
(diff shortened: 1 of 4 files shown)


internal/build/path_mapping.go
```diff
@@ -111,15 +113,40 @@ func fileToPathMapping(file string, mounts []model.Mount) (pathMapping, *PathMap
...
+				ContainerPath: containerPath,
 			}, nil
 		}
 	}
 	return pathMapping{}, pathMappingErrf("file %s matches no mounts", file)
 }
 
+// TODO(dmiller): this won't work in Windows
+func endsWithSlash(path string) bool {
+	return strings.HasSuffix(path, "/")
```
nicks: Use filepath.Separator?

internal/build/path_mapping.go
```diff
@@ -111,15 +113,40 @@ func fileToPathMapping(file string, mounts []model.Mount) (pathMapping, *PathMap
...
+}
+
+func isFile(path string) (bool, error) {
```
nicks: This isn't quite what you want. First, it's buggy (this treats symlinks as non-files). I'm also not sure it's what you want semantically. What does docker do if you COPY dir1 dir2/? I thought it created dir2/dir1

internal/build/tar.go
```diff
@@ -132,10 +132,14 @@ func (a *ArchiveBuilder) tarPath(ctx context.Context, source, dest string) error
 			header.Name = strings.TrimPrefix(path, source)
 			// ...and live inside `dest`
 			header.Name = filepath.Join(dest, header.Name)
+		} else if strings.HasSuffix(dest, "/") { // TODO(dmiller): this won't work in Windows
```
nicks: Ditto here re: separator

internal/build/path_mapping.go
jazzdan: Good point about symlinks! Will fix.

I didn't change the handling of copying directories to directories. This should only take effect if the source is a file but the destination is a directory.

> What does docker do if you COPY dir1 dir2/? I thought it created dir2/dir1
Looks like it doesn't nest directories in that case.

```dockerfile
FROM iron/go:dev

RUN mkdir /remotedir2
COPY dir1 /remotedir2/
```

```
> tree
.
├── Dockerfile
├── dir1
│   └── hi
```

```
> docker run -it 303a3d0f2157 /bin/sh
/go # cd /tmp/
/tmp # ls
/tmp # cd /
/ # ls
bin         etc         home        media       proc        root        sbin        sys         usr
dev         go          lib         mnt         remotedir2  run         srv         tmp         var
/ # cd remotedir2/
/remotedir2 # ls
hi
```

nicks: i would probably change this to be !isDir instead of isFile, which is probably closer to what we want?

PR #795, opened 2018-12-04
[docker-compose] prelim log support [ch978]

maiamcc:
Known bug: doesn't support getting logs since a certain time. usually
not a problem since we rebuild with --force-recreate so have a clean
container every time. however, it's annoying if there are already dc
services running when the user calls `tilt up` -- then we periodically
get old error logs for the old versions of the services as their deps
all die, until the services themselves reboot. not great.

### DIFF
--- a/internal/engine/docker_compose_log_manager.go
+++ b/internal/engine/docker_compose_log_manager.go
@@ -13,18 +13,18 @@ import (
 
 // Collects logs from running docker-compose services.
 type DockerComposeLogManager struct {
-	watches map[model.ManifestName]DockerComposeLogWatch
+	watches map[model.ManifestName]dockerComposeLogWatch
 }
 
 func NewDockerComposeLogManager() *DockerComposeLogManager {
 	return &DockerComposeLogManager{
-		watches: make(map[model.ManifestName]DockerComposeLogWatch),
+		watches: make(map[model.ManifestName]dockerComposeLogWatch),
 	}
 }
 
 // Diff the current watches against set of current docker-compose services, i.e.
 // what we SHOULD be watching, returning the changes we need to make.
-func (m *DockerComposeLogManager) diff(ctx context.Context, st store.RStore) (setup []DockerComposeLogWatch, teardown []DockerComposeLogWatch) {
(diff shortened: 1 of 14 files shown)


internal/engine/docker_compose_log_manager.go
```diff
@@ -0,0 +1,148 @@
...
+	if !state.WatchMounts {
+		return nil, nil
+	}
+
+	for _, ms := range state.ManifestStates {
+		if !ms.Manifest.IsDockerCompose() {
+			continue
+		}
+
+		existing, isActive := m.watches[ms.Manifest.Name]
```
jazzdan: does `isActive` here mean that a watch is active if the `DockerComposeLogWatch` in that map is not null?

jazzdan: If so, how can it be null?

internal/engine/docker_compose_log_manager.go
```diff
@@ -0,0 +1,148 @@
...
+	multiWriter := io.MultiWriter(logWriter, actionWriter)
+
+	_, err = io.Copy(multiWriter, NewHardCancelReader(watch.ctx, readCloser))
+	if err != nil && watch.ctx.Err() == nil {
+		logger.Get(watch.ctx).Infof("Error streaming %s logs: %v", name, err)
+		return
+	}
+}
+
+type DockerComposeLogWatch struct {
```
jazzdan: Does this type need to be exported?

internal/engine/docker_compose_log_manager.go
```diff
@@ -0,0 +1,148 @@
...
+	// cID             container.ID
+	// cName           container.Name
+}
+
+type DockerComposeLogActionWriter struct {
+	store        store.RStore
+	manifestName model.ManifestName
+}
+
+func (w DockerComposeLogActionWriter) Write(p []byte) (n int, err error) {
```
jazzdan: Funky, I love it! Is this used for anything though?

maiamcc: it's part of the multiWriter at ln 113 -- so when we copy to the multiWriter at ln 115 we're writing to both the output logger (i.e. allll the tilt logs) and the `DockerComposeLogActionWriter` (which, when we call `Write()`, dispatches a `DockerComposeLogAction` event)

PR #806, opened 2018-12-05
tilt: make (l) cycle through log views

landism:
image

not super thrilled with the `(l) maximize log ↑`, but we can fix that in the next couple days


internal/hud/renderer.go
```diff
@@ -130,16 +151,16 @@ func (r *Renderer) renderPaneHeader(v view.View) rty.Component {
 				// This space intentionally left blank
 			}
 		}
-		sbLeft.Fg(cBad).Text("✖").Fg(tcell.ColorDefault).Fg(cText).Textf("%s%s", errorCountMessage, tiltfileError.String()).Fg(tcell.ColorDefault)
+		sb.Fg(cBad).Text("✖").Fg(tcell.ColorDefault).Fg(cText).Textf("%s%s", errorCountMessage, tiltfileError.String()).Fg(tcell.ColorDefault)
 	}
-	header.Add(sbLeft.Build())
-	header.Add(rty.NewFillerString(' '))
-	return rty.NewFixedSize(rty.Bg(header, tcell.ColorWhiteSmoke), rty.GROW, 1)
+	l := rty.NewLine()
```
nicks: fwiw, i added a `rty.OneLine` helper for this idiom

landism: Oh nice. I've been putting that off for like 5 PRs now.

PR #811, opened 2018-12-06
tiltfile2: add test for loading only one manifest

jazzdan:


internal/tiltfile2/tiltfile_test.go
```diff
@@ -475,6 +505,10 @@ func (f *fixture) assertManifest(name string, opts ...interface{}) model.Manifes
 	return m
 }
 
+func (f *fixture) assertLenManifests(expected int) {
```
dbentley: Is this used anywhere? Also, I'd call it assertNumManifests or assertNManifests

PR #813, opened 2018-12-06
tiltfile2: remove unused makeBlob function

jazzdan:


dbentley:
Or we could expose it (and test it) for people who have a string that they want to turn into a blob... Which could be useful. Anyway, your choice, and good for us to have shared context on.

jazzdan:
Currently, and I think this should always be the case, anything that _takes_ a Blob actually takes a union of Blob and str. I don't see any reason why a function would _only_ take a Blob, do you?

PR #819, opened 2018-12-07
docs: add a feature roadmap

nicks:
Hello @dbentley,

Please review the following commits I made in branch nicks/roadmap:

6aa77658d94a7057796c89d41b0d63be7b7717e7 (2018-12-07 13:21:26 -0500)
docs: add a feature roadmap


ROADMAP.md
```diff
@@ -0,0 +1,144 @@
+# Roadmap
+
+Welcome to the future!
+
+This page is for features that the Tilt team has concrete
```
dbentley: "concrete plans" makes it sound like we have time frames; is there a way to say "we know about this and have some idea how"?

nicks: "plaster plans"?

dbentley: Hahahahaha.

"There's a lot we could do; we'd like feedback in how to prioritize these options based on what would be most useful"

dbentley: (anyway, yeah, looks good)

PR #821, opened 2018-12-07
hud: if showing logs inline, try crashLogs first [ch1090]

maiamcc:
Hello @nicks, @landism,

Please review the following commits I made in branch maiamcc/show-crash-log-first:

14508345a9d31328df387b80f629b67955034789 (2018-12-07 16:00:11 -0500)
hud: if showing logs inline, try crashLogs first [ch1090]


nicks:
unit test?

maiamcc:
yep on it now

maiamcc:
To be clear, this doesn't differentiate between showing crash log vs. pod log according to whether we're crash rebuilding or not -- it's just an emergent property of this code that, since we don't set crash log UNLESS we're doing a crash rebuild, that we'll only show the crash log for cases where we would otherwise show stale logs.

something that programmatically does this would look like:
```
if isCrashing(v.res) {
		podLog := v.res.CrashLog
		// Show the current podLog inline if a. we don't have a crash log to show, or
		// b. if we're NOT CrashRebuilding -- i.e. we didn't replace the pod, so can
		// just show the most recent pod logs.
		if podLog == "" || !isCrashRebuilding(v.res) {
			podLog = v.res.PodLog
		}
    ...
}
```

would you rather this code be in there?

nicks:
i'm not sure!

PR #826, opened 2018-12-10
tiltfile: use new Tiltfile syntax

jazzdan:


Tiltfile
```diff
@@ -1,27 +1,27 @@
...
-  start_fast_build('Dockerfile.base', 'gcr.io/blorg-dev/tilt-test', entrypoint)
-  add(local_git_repo('.'), '/go/src/github.com/windmilleng/tilt')
-  run('cd src/github.com/windmilleng/tilt && go build ./...')
-  image = stop_build()
+def get_username():
+  return str(local('whoami')).rstrip('\n')
 
-  return k8s_service(image, yaml=read_file('tilt-test.yaml'))
+# docserver
+k8s_resource('docserver', 'docserver.yaml', port_forwards=10000)
```
nicks: have you tested this change? i'm pretty sure this won't work, because the resource name is 'tilt-docserver'

nicks: (i sent out a separate pr that did something different...ended up deleting the other parts, since it seems unlikely you'd ever want to start these 3 services together... https://github.com/windmilleng/tilt/pull/828)

jazzdan: SGTM!

PR #827, opened 2018-12-10
engine: fix crash-rebuilds when the pod updates before the build finishes

nicks:
Hello @maiamcc,

Please review the following commits I made in branch nicks/crash:

97336d23e7bc06a1e0d74924aa17b36f861a9d6b (2018-12-10 15:50:40 -0500)
engine: fix crash-rebuilds when the pod updates before the build finishes

### DIFF
--- a/internal/engine/upper_test.go
+++ b/internal/engine/upper_test.go
@@ -1023,6 +1023,108 @@ func TestPodUnexpectedContainerStartsImageBuild(t *testing.T) {
 	f.waitForCompletedBuildCount(1)
 }
 
+func TestPodUnexpectedContainerStartsImageBuildOutOfOrderEvents(t *testing.T) {
+	f := newTestFixture(t)
+	defer f.TearDown()
+	f.bc.DisableForTesting()
+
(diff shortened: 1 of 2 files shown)


internal/engine/upper_test.go
```diff
@@ -1023,6 +1023,44 @@ func TestPodUnexpectedContainerStartsImageBuild(t *testing.T) {
...
+	defer f.TearDown()
+	f.bc.DisableForTesting()
+
+	mount := model.Mount{LocalPath: "/go", ContainerPath: "/go"}
+	name := model.ManifestName("foobar")
+	manifest := f.newManifest(name.String(), []model.Mount{mount})
+
+	f.Start([]model.Manifest{manifest}, true)
+
+	// Start and end a fake build to set manifestState.ExpectedContainerId
```
maiamcc: put some paragraph breaks in this code so i can tell where the setup code ends and the actual testing starts?

nicks: done

PR #834, opened 2018-12-11
rty: fix crashes when the terminal window is too small

nicks:
Hello @maiamcc,

Please review the following commits I made in branch nicks/ch1052/crash:

8c4e8d154169d7349963202253e3c5680b8348b9 (2018-12-11 15:20:59 -0500)
rty: fix crashes when the terminal window is too small


maiamcc:
what does it look like when we try to render in a too-small window?

nicks:
screenshot from 2018-12-11 16-01-21

PR #835, opened 2018-12-11
hud: enter toggles selected resource log modal

maiamcc:
before `enter` only opened the log modal -- now it toggles

### DIFF
--- a/internal/hud/hud.go
+++ b/internal/hud/hud.go
@@ -110,7 +110,7 @@ func (h *Hud) handleScreenEvent(ctx context.Context, dispatch func(action store.
 	h.mu.Lock()
 	defer h.mu.Unlock()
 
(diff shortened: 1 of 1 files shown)


internal/hud/hud.go
```diff
@@ -190,6 +191,13 @@ func (h *Hud) handleScreenEvent(ctx context.Context, dispatch func(action store.
 
 				h.currentViewState.LogModal = view.LogModal{ResourceLogNumber: selectedIdx + 1}
 				h.activeModal().Bottom()
+			} else {
+				_, isLogModal := am.(logModal)
+				if isLogModal {
+					if am != nil {
```
nicks: i think this check is redundant?

maiamcc: ha you're right

PR #836, opened 2018-12-12
docker: some versions of docker reject requests for buildkit with error if it's not supported :sob:

nicks:
Hello @maiamcc,

Please review the following commits I made in branch nicks/ch1104/minikube:

e19ed6c02a570b45d3af177bb7cbc9f8c6a189ed (2018-12-12 11:09:58 -0500)
docker: some versions of docker reject requests for buildkit with error if it's not supported :sob:


nicks:
going to merge this and cut a release, so we don't have a version that doesn't work with the latest minikube

PR #839, opened 2018-12-13
engine: clear podset in crashrebuild so we don't try to build on old container

maiamcc:
When a pod crashes, k8s spins up a new pod from the last image it knows about,
which may be running stale code. this pod gets added to the manifest's podset --
but we don't ever want to try and build on it, cuz it has the wrong code!
If we leave it in the podset and try to build this manfest again before our crash rebuild
has finished and told us about the new, accurate pod, we could attempt to build on
this transient pod with bad code, and weird state ensues.

testing this is a royal PITA -- is this safe enough to land w/o explicit tests?


nicks:
what codepath do you think this changes? 

AFAICT, buildcontroller already ignores PodSet when NeedsRebuildFromCrash is true

notice that PodSet can get populated by k8s events at any time, so clearing it here will just introduce new race conditions.

maiamcc:
> AFAICT, buildcontroller already ignores PodSet when NeedsRebuildFromCrash is true

yes, this is trying to get at the case where it's NOT a crash rebuild, but a build triggered by code changes, only the "best pod" in the PodSet is the one that we're trying to get rid of with the crash rebuild.

We change some code that causes a crash, our in-memory-updated container crashes and k8s replaces it with OldCodeContainer (which gets added to the podset). We notice the unexpected container and trigger a crash rebuild. If, however, we start trying to do another build before stuff from the crash rebuild has come up and added itself to the PodSet, our next build will try to operate on OldCodeContainer (because it is the most recent pod in the PodSet) and lead to more weird state.

(In particular, we try to build on OldCodeContainer, but then CrashRebuiltContainer comes up and we treat it as ANOTHER unexpected container event and trigger ANOTHER crash rebuild.)

does that make sense?

I know there are race conditions and i'm not thrilled about them :-/ but this is an attempt to fix ANOTHER race condition...?

nicks:
In that case, can we synchronize on a check on the image tag? Presumably the newly built pod will have a different tag.

Alternatively, we could tag the build with some other sort of build ID to help synchronize

jazzdan:
What's the status of this PR?

maiamcc:
UHHH good Q. i think this problem only tends to come up when you break and fix something real quickly, i.e. for demos, and if people haven't been feeling this pain in demos lately, then i'm fine closing this PR -- it's a weird edge case that's more trouble than it's worth to fix if it's not bothering anyone.

maiamcc:
Oh whoops, I never actually closed this PR after my comment last month!

PR #840, opened 2018-12-13
tiltfile2: auto-infer git repos

nicks:
Hello @maiamcc,

cc @jazzdan

Please review the following commits I made in branch nicks/dockerignore:

d894bfdb4ece7a10e74f94f4ee56a83820f001dd (2018-12-13 11:59:02 -0800)
tiltfile2: auto-infer git repos
this is the first part of a bigger change in how we
do dockerignore/gitignore to accomodate the new tiltfile changes

### DIFF
--- a/internal/tiltfile2/tiltfile_test.go
+++ b/internal/tiltfile2/tiltfile_test.go
@@ -12,23 +12,25 @@ import (
 
 	"github.com/stretchr/testify/assert"
 
+	"github.com/windmilleng/tilt/internal/ignore"
 	"github.com/windmilleng/tilt/internal/k8s"
 	"github.com/windmilleng/tilt/internal/k8s/testyaml"
 	"github.com/windmilleng/tilt/internal/model"
+	"github.com/windmilleng/tilt/internal/ospath"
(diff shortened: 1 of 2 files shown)


internal/tiltfile2/tiltfile_test.go
```diff
@@ -703,6 +726,42 @@ func (f *fixture) assertManifest(name string, opts ...interface{}) model.Manifes
...
+			filterName := "BuildContextFilter"
+			if opt.fileChange {
+				var err error
+				filter, err = ignore.CreateFileChangeFilter(m)
+				assert.NoError(f.t, err)
+				filterName = "FileChangeFilter"
+			}
+
+			actualFilter, err := filter.Matches(path, false)
+			assert.NoError(f.t, err)
```
maiamcc: this won't stop exec if there IS an error, so either nest the stuff below in
```if assert.NoError(...) {
    ...
}```
or `f.t.Error` if there IS an error.

nicks: i changed it to a Fatal. if either if these fail, we probably want to stop everything, since they're really not supposed to fail.

PR #842, opened 2018-12-13
tiltfile: docker compose support to new tiltfile syntax [ch1108]

maiamcc:
haaa forgot about this part

### DIFF
--- a/internal/tiltfile2/docker_compose.go
+++ b/internal/tiltfile2/docker_compose.go
@@ -1,0 +1,41 @@
+package tiltfile2
+
+import (
+	"fmt"
+
+	"github.com/google/skylark"
+	"github.com/windmilleng/tilt/internal/dockercompose"
+)
+
+// dcResource represents a single docker-compose config file and all its associated services
+type dcResource struct {
+	yamlPath string
+
+	services []dockercompose.Service
+}
+
+func (s *tiltfileState) dockerCompose(thread *skylark.Thread, fn *skylark.Builtin, args skylark.Tuple, kwargs []skylark.Tuple) (skylark.Value, error) {
+	var path skylark.String
+	err := skylark.UnpackArgs(fn.Name(), args, kwargs, "path", &path)
+	if err != nil {
+		return nil, err
+	}
+
+	// ~~ i still don't understand why path.String() returns it quoted / if there's a more idiomatic way
+	yamlPath, ok := skylark.AsString(path)
+	if !ok {
+		return nil, fmt.Errorf("couldn't AsString skylark.String: %v", path)
+	}
+
+	// TODO: support more than one docker-compose.yaml file
+	services, _, err := dockercompose.ParseConfig(s.ctx, []string{yamlPath})
+	if err != nil {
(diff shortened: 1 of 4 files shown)


internal/tiltfile2/tiltfile_test.go
```diff
@@ -502,6 +502,18 @@ k8s_resource('bar', 'bar.yaml')
 	}
 }
 
+func TestDockerCompose(t *testing.T) {
```
nicks: O.o

internal/tiltfile2/docker_compose.go
```diff
@@ -0,0 +1,41 @@
...
+
+// dcResource represents a single docker-compose config file and all its associated services
+type dcResource struct {
+	yamlPath string
+
+	services []dockercompose.Service
+}
+
+func (s *tiltfileState) dockerCompose(thread *skylark.Thread, fn *skylark.Builtin, args skylark.Tuple, kwargs []skylark.Tuple) (skylark.Value, error) {
+	var path skylark.String
```
nicks: var path string?

it seems like you'd want skylark to do the conversion for you, instead of doing it yourself

internal/tiltfile2/tiltfile_state.go
```diff
@@ -91,10 +95,10 @@ func (s *tiltfileState) builtins() skylark.StringDict {
 	return r
 }
 
-func (s *tiltfileState) assemble() ([]*k8sResource, []k8s.K8sEntity, error) {
+func (s *tiltfileState) assemble() ([]dcResource, []*k8sResource, []k8s.K8sEntity, error) {
```
nicks: maybe we should have a ResourceSet struct?

internal/tiltfile2/tiltfile_state.go
```diff
@@ -309,6 +317,32 @@ func (s *tiltfileState) translate(resources []*k8sResource) ([]model.Manifest, e
 	return result, nil
 }
 
+func (s *tiltfileState) translateDC(resources []dcResource) ([]model.Manifest, error) {
+	if len(resources) == 0 {
+		return nil, nil
+	}
+
+	// TODO(maia): support more than one docker-compose file.
```
nicks: i'm not sure this is worth implementing until we know someone needs it? it seems like it'd be easier to emit a good error message at the dockerCompose call site, so that the error message gets appropriate line numbers, etc

internal/tiltfile2/tiltfile_test.go
maiamcc: I mean, just the tiltfile part of it! ...point taken tho.

internal/tiltfile2/docker_compose.go
maiamcc: fair! i think I blindly copypasta'd this from elsewhere in the code.

internal/tiltfile2/tiltfile_state.go
maiamcc: aah yeah, good call.

PR #844, opened 2018-12-14
create dc manifests with mounts derived from df so we can watch files [ch974]

maiamcc:
DC manifests get created with the correct mounts, afaict. The watching
is still a little buggy, will sort that out in a subsequent PR.


maiamcc:
Aforementioned watching bugs worked around (though perhaps not FIXED) in #845

internal/dockercompose/compose.go
```diff
@@ -2,55 +2,150 @@ package dockercompose
...
+
+	svcNames, err := svcNames(ctx, configPath)
+	if err != nil {
+		return nil, err
+	}
+
+	var services []Service
+
+	for _, name := range svcNames {
+		if name == "" {
```
nicks: i think you already filtered these out in the svcNames function?

internal/dockercompose/compose.go
```diff
@@ -2,55 +2,150 @@ package dockercompose
...
+func dcOutput(ctx context.Context, configPath string, args ...string) (string, error) {
+	args = append([]string{"-f", configPath}, args...)
 	output, err := exec.CommandContext(ctx, "docker-compose", args...).Output()
 	if err != nil {
 		errorMessage := fmt.Sprintf("command 'docker-compose %q' failed.\nerror: '%v'\nstdout: '%v'", args, err, string(output))
 		if err, ok := err.(*exec.ExitError); ok {
 			errorMessage += fmt.Sprintf("\nstderr: '%v'", string(err.Stderr))
 		}
-		err = errors.New(errorMessage)
+		err = fmt.Errorf(errorMessage)
```
nicks: are you sure you want to make this change? note that this will crash if errorMessage contains any format specifiers. errors.New seemed like the right function?

internal/dockerfile/dockerfile.go
```diff
@@ -122,6 +123,55 @@ func (d Dockerfile) ValidateBaseDockerfile() error {
...
+	if !(cmd == "ADD" || cmd == "COPY") {
+		return model.Mount{}, fmt.Errorf("nodeToMounts works on ADD/COPY nodes; got '%s'", cmd)
+	}
+
+	srcNode := node.Next
+	dstNode := srcNode.Next
+
+	src := srcNode.Value
+	if !path.IsAbs(src) {
+		src = path.Join(context, src)
```
nicks: this should use `filepath`. `path` is for urls

internal/dockerfile/dockerfile.go
```diff
@@ -122,6 +123,55 @@ func (d Dockerfile) ValidateBaseDockerfile() error {
...
+	}
+
+	return model.Mount{
+		LocalPath:     src,
+		ContainerPath: dstNode.Value,
```
nicks: should this error out if the container path is relative? I'm not sure what will happen with a model.Mount with a relative ContainerPath

internal/dockercompose/compose.go
```diff
@@ -2,55 +2,150 @@ package dockercompose
...
 	}
 	return string(output), err
```
nicks: it seems like it'd be simpler to return a []byte here, rather than converting it back to []byte later

internal/dockerfile/dockerfile.go
maiamcc: 😳

internal/dockercompose/compose.go
maiamcc: ah yes, copy 🍝

internal/dockerfile/dockerfile.go
maiamcc: I... am also not sure.

internal/dockercompose/compose.go
maiamcc: i elsewhere need a string, so it's kind of six of one at this point 🤷‍♀️

PR #849, opened 2018-12-18
hud: show build history

nicks:
Hello @yuindustries, @jazzdan,

Please review the following commits I made in branch nicks/ch1047/historyux:

9762b916b6847989a32eac2582c2ce02d1d4798d (2018-12-17 19:23:33 -0500)
hud: show build history


nicks:
@yuindustries and i reviewed the design, and decided that the `(see above)` text was too distracting. i changed it so that it repeats the line above, but grayed out, which is still not great, but a little bit better.

screenshot from 2018-12-18 17-28-45

PR #850, opened 2018-12-18
tiltfile2: basic test for loading a docker compose manifest

jazzdan:
@maiamcc any other assertions I should add for this basic test?

I also changed the docker compose yaml path to be absolute to match how we treat other files that we read and store in memory.

### DIFF
--- a/internal/tiltfile2/docker_compose.go
+++ b/internal/tiltfile2/docker_compose.go
@@ -2,6 +2,7 @@ package tiltfile2
 
 import (
 	"fmt"
+	"path/filepath"
 	"reflect"
 
 	"github.com/google/skylark"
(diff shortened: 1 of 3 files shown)


internal/tiltfile2/docker_compose.go
```diff
@@ -23,17 +24,21 @@ func (s *tiltfileState) dockerCompose(thread *skylark.Thread, fn *skylark.Builti
 	if err != nil {
 		return nil, err
 	}
+	absConfigPath, err := filepath.Abs(configPath)
```
nicks: maybe ospath.RealAbs, like we did elsewhere?

jazzdan: yes, thank you!

internal/tiltfile2/tiltfile_test.go
```diff
@@ -614,7 +614,24 @@ k8s_resource('foo', 'foo.yaml')
 }
 
 func TestDockerComposeResourceCreation(t *testing.T) {
-	// TODO(maia)
+	f := newFixture(t)
+	defer f.TearDown()
+
+	f.setupFoo()
```
maiamcc: this is a buncha k8s-specific setup, i'm concerned it muddies the test. maybe just specifically set up the dc-specific bits you need?

internal/tiltfile2/tiltfile_test.go
```diff
@@ -614,7 +614,24 @@ k8s_resource('foo', 'foo.yaml')
...
+    build: ./foo
+    command: sleep 100
+    ports:
+      - "12312:12312"`)
+	YAMLPath := f.TempDirFixture.JoinPath("docker-compose.yml")
+	tf := fmt.Sprintf("docker_compose('%s')", YAMLPath)
+	f.file("Tiltfile", tf)
+
+	f.load("foo")
+	f.assertManifest("foo", dcYAMLPath(YAMLPath))
```
maiamcc: nice. could you add a todo for me? -- assert Dockerfile and docker-compose.yml are config files

maiamcc:
hmm do we want a docker compose client that we can mock out for tests so we don't actually have to call dc?

jazzdan:
@maiamcc in my opinion not right now- we only use `docker-compose` here to read the YAML, not do any execution. It's similar to how this test suite depends on `kustomize`. But probably in the Upper we'll want a mock, yeah.

maiamcc:
> @maiamcc in my opinion not right now- we only use docker-compose here to read the YAML, not do any execution. It's similar to how this test suite depends on kustomize. But probably in the Upper we'll want a mock, yeah.

`docker-compose config` VALIDATES the config as well as reading it so it does rely on our test fixture having a valid DC setup, which is maybe not ideal / muddies the test / makes set up more of a pain? but also eh, i agree that we can wait til we need it.

PR #851, opened 2018-12-18
tiltfile2: docker-compose error test cases

jazzdan:


internal/tiltfile2/tiltfile_test.go
```diff
@@ -636,11 +636,56 @@ services:
 }
 
 func TestMultipleDockerComposeNotSupported(t *testing.T) {
-	// TODO(maia)
+	f := newFixture(t)
+	defer f.TearDown()
+
+	f.setupFoo()
+	dc1 := f.TempDirFixture.JoinPath("docker-compose1.yml")
```
nicks: why does this need to be an absolute path? i would have expected a relative path to work? this seems like a bug that would be a major blocker to using this.

jazzdan: yes I know what the problem is. Will address this in the next PR!

PR #852, opened 2018-12-18
tiltfile2: resolve docker-compose path relative to tiltfile root

jazzdan:
Also introduce the concept of a tiltfileRoot which could, in the future,
allow the Tiltfile and the directory we operate in to be disjoint paths.

### DIFF
--- a/internal/tiltfile2/docker_compose.go
+++ b/internal/tiltfile2/docker_compose.go
@@ -2,12 +2,10 @@ package tiltfile2
 
 import (
 	"fmt"
-	"path/filepath"
 	"reflect"
 
 	"github.com/google/skylark"
 	"github.com/windmilleng/tilt/internal/dockercompose"
-	"github.com/windmilleng/tilt/internal/ospath"
 )
 
 // dcResource represents a single docker-compose config file and all its associated services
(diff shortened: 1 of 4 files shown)


internal/tiltfile2/tiltfile_state.go
```diff
@@ -23,6 +23,7 @@ type tiltfileState struct {
 	// set at creation
 	ctx      context.Context
 	filename string
+	root     string
```
nicks: hmmm...doesn't TiltfileState already have a function for this? it should be absPath or smth?

jazzdan: We have functions for dealing with the paths for static (image) builds, but not one for the Tiltfile afaict. Got a link?

jazzdan: Aha! it was in the `files` file.

internal/tiltfile2/docker_compose.go
```diff
@@ -24,7 +25,10 @@ func (s *tiltfileState) dockerCompose(thread *skylark.Thread, fn *skylark.Builti
 	if err != nil {
 		return nil, err
 	}
-	absConfigPath, err := filepath.Abs(configPath)
+	if !filepath.IsAbs(configPath) {
```
nicks: this is equivalent to configPath = s.absPath(configDir)

PR #855, opened 2018-12-18
dockcompose: docker compose client interface

jazzdan:


internal/dockercompose/wrapper.go
```diff
@@ -0,0 +1,135 @@
...
+	"io"
+	"os/exec"
+	"strings"
+
+	"github.com/pkg/errors"
+	"github.com/windmilleng/tilt/internal/logger"
+)
+
+type DockerComposeClient interface {
+	Up(ctx context.Context, pathToConfig, serviceName string, stdout, stderr io.Writer) error
```
maiamcc: nit: elsewhere we say `configPath` rather than `pathToConfig`

internal/dockercompose/wrapper.go
```diff
@@ -0,0 +1,135 @@
...
+	Services(ctx context.Context, pathToConfig string) (string, error)
+}
+
+type cmdDCClient struct{}
+
+func NewDockerComposeClient() DockerComposeClient {
+	return &cmdDCClient{}
+}
+
+func (c *cmdDCClient) Up(ctx context.Context, pathToConfig, serviceName string, stdout, stderr io.Writer) error {
```
maiamcc: seems like you could split out the fact that it always takes args `docker-compose -f <configPath>` if you wanted

internal/dockercompose/wrapper.go
```diff
@@ -0,0 +1,135 @@
...
+	"fmt"
+	"io"
+	"os/exec"
+	"strings"
+
+	"github.com/pkg/errors"
+	"github.com/windmilleng/tilt/internal/logger"
+)
+
+type DockerComposeClient interface {
```
nicks: i wonder if DockerComposeClient should be per docker-compose file, and then pathToConfig would just be a property of the client. (similar to how a db client would have the db address be a property of the client)

internal/dockercompose/wrapper.go
```diff
@@ -0,0 +1,135 @@
...
+	return &cmdDCClient{}
+}
+
+func (c *cmdDCClient) Up(ctx context.Context, pathToConfig, serviceName string, stdout, stderr io.Writer) error {
+	cmd := exec.CommandContext(ctx, "docker-compose", "-f", pathToConfig, "up", "--no-deps", "--build", "--force-recreate", "-d", serviceName)
+	cmd.Stdout = stdout
+	cmd.Stderr = stderr
+
+	err := cmd.Run()
+	return FormatError(cmd, nil, err)
```
maiamcc: why not pass stdout to `FormatError`?

internal/dockercompose/wrapper.go
```diff
@@ -0,0 +1,135 @@
...
+}
+
+func (c *cmdDCClient) Down(ctx context.Context, pathToConfig string, stdout, stderr io.Writer) error {
+	cmd := exec.CommandContext(ctx, "docker-compose", "-f", pathToConfig, "down")
+	cmd.Stdout = logger.Get(ctx).Writer(logger.InfoLvl)
```
maiamcc: should be using `stdout`/`stderr` that you passed, no?

internal/dockercompose/wrapper.go
```diff
@@ -0,0 +1,135 @@
...
+
+	err := cmd.Run()
+	if err != nil {
+		return FormatError(cmd, nil, err)
+	}
+
+	return nil
+}
+
+func (c *cmdDCClient) Logs(ctx context.Context, pathToConfig, serviceName string) (io.ReadCloser, error) {
```
maiamcc: hmm maybe worth a comment that this keeps running/tailing logs? (or a rename to indicate?)

internal/dockercompose/wrapper.go
```diff
@@ -0,0 +1,135 @@
...
+		err = cmd.Wait()
+		if err != nil {
+			logger.Get(ctx).Debugf("cmd `docker-compose %s` exited with error: \"%v\" (stderr: %s)",
+				strings.Join(args, " "), err, errBuf.String())
+		}
+	}()
+	return stdout, nil
+}
+
+func (c *cmdDCClient) Events(ctx context.Context, pathToConfig string) (<-chan string, error) {
```
maiamcc: same comment as above re: logs

internal/dockercompose/wrapper.go
jazzdan: yes I think that would be a good refactor. I'd rather leave this as is as it causes the fewest changes in the code but enables testing. Then we can make those changes much more confidently.

internal/dockercompose/wrapper.go
maiamcc: oh yo wait, just use `dcOutput`

internal/dockercompose/wrapper.go
```diff
@@ -0,0 +1,135 @@
...
+
+func (c *cmdDCClient) Services(ctx context.Context, pathToConfig string) (string, error) {
+	return dcOutput(ctx, pathToConfig, "config", "--services")
+}
+
+func dcOutput(ctx context.Context, configPath string, args ...string) (string, error) {
+	args = append([]string{"-f", configPath}, args...)
+	output, err := exec.CommandContext(ctx, "docker-compose", args...).Output()
+	if err != nil {
+		errorMessage := fmt.Sprintf("command 'docker-compose %q' failed.\nerror: '%v'\nstdout: '%v'", args, err, string(output))
```
maiamcc: this should be `FormatError` i think

internal/engine/docker_compose_build_and_deployer.go
```diff
@@ -29,12 +28,11 @@ func (bd *DockerComposeBuildAndDeployer) BuildAndDeploy(ctx context.Context, man
 		return store.BuildResult{}, RedirectToNextBuilderf("not a docker compose manifest")
 	}
 
-	cmd := exec.CommandContext(ctx, "docker-compose", "-f", manifest.DCConfigPath, "up", "--no-deps", "--build", "--force-recreate", "-d", manifest.Name.String())
-	cmd.Stdout = logger.Get(ctx).Writer(logger.InfoLvl)
-	cmd.Stderr = logger.Get(ctx).Writer(logger.InfoLvl)
+	dcc := dockercompose.NewDockerComposeClient()
+	stdout := logger.Get(ctx).Writer(logger.InfoLvl)
```
maiamcc: consider having the `cmdDCClient` just always write to the logger, and so when you call `NewD...Client()` it makes the logger and stores it on itself -- then you don't need to pass stdout/stderr in with every function call.

maiamcc: ...tbh we could even make a dockercompose client specific to the dc-yaml and then we wouldnt have to pass that in on every call, either, just at construction?

internal/dockercompose/wrapper.go
maiamcc: ha yes i agree. leave a todo/ticket?

internal/dockercompose/wrapper.go
jazzdan: oh yes good call

internal/dockercompose/wrapper.go
jazzdan: yeah I'm going to avoid making any changes in functionality if I can until we have tests, then I'll go nuts.

internal/engine/docker_compose_build_and_deployer.go
jazzdan: yup agreed, I left a TODO for that

PR #859, opened 2018-12-19
Add docker-compose to development dependencies

jazzdan:


DEVELOPING.md
```diff
@@ -20,6 +20,7 @@ features that we have in mind and might already be in-progress.
   so that you don't need to install extra toolchains locally (e.g., the protobuf compiler).
 - **[kubectl](https://kubernetes.io/docs/tasks/tools/install-kubectl/)** (for tests)
 - **[kustomize](https://github.com/kubernetes-sigs/kustomize)**: `go get -u sigs.k8s.io/kustomize` (for tests)
+- **[docker compose](https://docs.docker.com/compose/install/)**: (for tests)
```
nicks: might be worth mentioning that MacOS users don't need to install this separately

it's worth pointing out that this is a production dependency if you're using docker_compose in your Tiltfile, but i'm not sure how to communicate that effectively

PR #865, opened 2018-12-20
model: nest resource-specific info on manifest/manifestState [ch1136]

maiamcc:
To avoid shoving everything into the root level of a single struct, we want to group e.g. docker-compose info separately from k8s info.

This PR splits out docker-compose info into the new world. Following PRs will do the same with k8s info.


internal/cli/down.go
```diff
@@ -52,19 +51,22 @@ func (c downCmd) run(ctx context.Context, args []string) error {
 		logger.Get(ctx).Infof("error deleting k8s entities: %v", err)
 	}
 
-	var dcManifests []model.Manifest
+	var dcConfigPath string
 	for _, m := range manifests {
-		if m.IsDockerCompose() {
-			dcManifests = append(dcManifests, m)
+		if dcInfo, ok := m.DCInfo(); ok {
```
nicks: a common Go idiom is to have a Empty() or IsZero() method for the "zero" value of a struct...would that make more sense here?

internal/dockercompose/state.go
```diff
@@ -0,0 +1,48 @@
...
+	ActionDie:     StatusDown,
+	ActionKill:    StatusDown,
+	ActionRename:  StatusInProg,
+	ActionRestart: StatusUp, // ??
+	ActionStart:   StatusUp,
+	ActionStop:    StatusDown,
+	ActionUpdate:  StatusUp, // ??
+}
+
+func (evt Event) GuessStatus() (string, bool) {
```
nicks: ditto here re: using zero values rather than a separate bool

internal/cli/down.go
maiamcc: OH HEY WAIT @nicks @jazzdan i remember why i wanted to do it this way -- b/c if if you get back an empty dcInfo from `m.DCInfo` there's no way of telling whether
a. there's nothing in `m.deployInfo` (i.e. it's safe to overwrite with docker stuff), or
b. there's something ELSE in `m.deployInfo` (like k8s stuff) that failed the type check

I imagine this `dcInfo, ok` syntax mirroring that of a type cast. I can comment with this explanation, but -- does that make more sense / sound more acceptable?

nicks: Ok

PR #866, opened 2018-12-20
engine: docker compose logs test

jazzdan:


internal/engine/local_container_build_and_deployer_test.go
```diff
@@ -1,44 +1,13 @@
...
-	"github.com/windmilleng/wmclient/pkg/analytics"
 )
 
 const pod1 = k8s.PodID("pod1")
 
 var image1 = container.MustParseNamedTagged("re.po/project/myapp:tilt-936a185caaa266bb")
 
 const digest1 = "sha256:936a185caaa266bb9cbe981e9e05cb78cd732b0b3280eb944412bb6f8f8f07af"
-
-type containerBaDFixture struct {
```
jazzdan: this is unused afaict

internal/engine/upper_test.go
```diff
@@ -96,6 +104,9 @@ func (b *fakeBuildAndDeployer) BuildAndDeploy(ctx context.Context, manifest mode
 		return store.BuildResult{}, err
 	}
 
+	if manifest.IsDockerCompose() {
+		return b.nextBuildResult(imageID), nil
```
jazzdan: This is kind of hacky. Any better ideas?

internal/engine/local_container_build_and_deployer_test.go
maiamcc: ¯\_(ツ)_/¯ cool! then yes delete it!

internal/engine/upper_test.go
```diff
@@ -2030,6 +2069,31 @@ func (f *testFixture) WriteConfigFiles(args ...string) {
 	f.store.Dispatch(manifestFilesChangedAction{manifestName: ConfigsManifestName, files: filenames})
 }
 
+func (f *testFixture) setupHAProxy() model.Manifest {
```
maiamcc: consider renaming this method to be obvious that we're setting up "some docker-compose test junk" instead of like, an actual proxy doing actual things.

also, instead of tracking origWD and reading then writing files, you could just store the file contents strings -- see [testyaml.go](https://github.com/windmilleng/tilt/blob/6dc5d88efeac03ec0cc2da3ed3c26ccb30fb8c41/internal/k8s/testyaml/testyaml.go)

internal/engine/upper_test.go
```diff
@@ -1620,6 +1631,20 @@ func TestNewMountsAreWatched(t *testing.T) {
 	})
 }
 
+func TestDockerComposeRecordsLogs(t *testing.T) {
+	f := newTestFixture(t)
+	m := f.setupHAProxy()
+	expected := "spoonerisms_1  | 2018-12-20T16:11:04.070480042Z yarn install v1.10."
```
maiamcc: ...spoonerisms? 🤔

internal/engine/upper_test.go
maiamcc: yeahhh kind of hacky. fine for now tho -- except that docker compose builds don't return any info in the BuildResult (cuz there's not an image id) -- so maybe change `b.nextBuildResult` to work okay given a nil `imageID`.

internal/engine/upper_test.go
maiamcc: also wait i'm confused, is this actually running `docker-compose up`??!

maiamcc: nvm i see! if you wanna be extra clear, make this bit its own paragraph with comment (like "call to `docker-compose logs` will return this line -- make sure log watcher picks it up")

also -- test that it ends up in manifest...logs? (Or todo for that test.)

internal/engine/upper_test.go
jazzdan: I have had a TON of problems with indentations when storing YAML that way so I am trying to sidestep it for now

PR #876, opened 2018-12-20
hud: print docker compose logs to log modal

jazzdan:
Currently looks something like this:

screen shot 2018-12-20 at 5 36 29 pm


internal/engine/docker_compose_log_manager.go
```diff
@@ -142,7 +142,7 @@ type DockerComposeLogActionWriter struct {
 func (w DockerComposeLogActionWriter) Write(p []byte) (n int, err error) {
 	w.store.Dispatch(DockerComposeLogAction{
 		ManifestName: w.manifestName,
-		Log:          append([]byte{}, p...),
```
jazzdan: This seems unnecessary to me?

internal/engine/upper.go
```diff
@@ -197,7 +197,7 @@ func handleBuildStarted(ctx context.Context, state *store.EngineState, action Bu
 	}
 
 	if dcState := ms.DCResourceState(); !dcState.Empty() {
-		ms.ResourceState = dcState.WithCurrentLog([]byte{}) // TODO(maia): when reset(/not) CrashLog for DC service?
+		ms.ResourceState = dcState.WithCurrentLog(dcState.CurrentLog) // TODO(maia): when reset(/not) CrashLog for DC service?
```
jazzdan: @maiamcc can you elaborate on this comment?

internal/engine/docker_compose_log_manager.go
nicks: from the interface docs:
https://golang.org/pkg/io/#Writer
"Implementations must not retain p"

slices are mutable, so you don't have any real guarantee that this byte slice will be the same later, when the action is processed

jazzdan: aha! Thank you, I learned something new. :)

internal/hud/renderer.go
```diff
@@ -196,14 +196,9 @@ func isCrashing(res view.Resource) bool {
...
 func bestLogs(res view.Resource) string {
-	// TODO(maia): get the BEST logs (currently getting any old logs)
-	if res.IsDC() {
-		lastBuildLog := strings.TrimSpace(string(res.LastBuild().Log))
-		if res.LastBuild().StartTime.Before(res.PodCreationTime) &&
-			len(lastBuildLog) > 0 {
-			return lastBuildLog + "\n\n" + res.PodLog
-		}
-		return res.PodLog
+	// TODO(maia): get the BEST logs (currently getting all logs)
```
maiamcc: again, b/c we're not wrestling with pods that come and go, I think the "best" log for a dc service is just... it's log. I'll remove this comment.

PR #888, opened 2018-12-21
model: fast build fields to dockerInfo.buildDetails [ch1136]

maiamcc:
Hello @jazzdan, @nicks,

Please review the following commits I made in branch maiamcc/fast-build:

8d94a3f1414dcdf67fe240ef07d67b96cf270ff5 (2018-12-21 16:10:23 -0500)
model: fast build fields to dockerInfo.buildDetails


internal/model/docker_info.go
```diff
@@ -50,3 +50,17 @@ type FastBuild struct {
 
 func (FastBuild) buildDetails()  {}
 func (fb FastBuild) Empty() bool { return reflect.DeepEqual(fb, FastBuild{}) }
+
+// func (m1 Manifest) stepsEqual(s2 []Step) bool {
```
jazzdan: 👀

internal/model/manifest.go
```diff
@@ -104,24 +97,33 @@ func (m Manifest) Dockerignores() []Dockerignore {
 func (m Manifest) LocalPaths() []string {
 	if sbInfo := m.StaticBuildInfo(); !sbInfo.Empty() {
```
jazzdan: Could we transform this in to a case statement?

internal/model/manifest.go
```diff
@@ -104,24 +97,33 @@ func (m Manifest) Dockerignores() []Dockerignore {
...
 func (m Manifest) Validate() error {
 	if m.Name == "" {
 		return fmt.Errorf("[validate] manifest missing name: %+v", m)
 	}
-	for _, m := range m.Mounts {
-		if !filepath.IsAbs(m.LocalPath) {
-			return fmt.Errorf(
-				"[validate] mount.LocalPath must be an absolute path (got: %s)", m.LocalPath)
+
+	if fbInfo := m.FastBuildInfo(); !fbInfo.Empty() {
```
jazzdan: This is a lot of indentation, could we move the check about?

```go
if fbInfo := m.FastBuildInfo(); fbInfo.Empty() {
    return nil
}
// ... 
```

internal/model/manifest.go
maiamcc: not really, cuz we're looking at the type of two different fields :-/

PR #897, opened 2019-01-02
tiltfile2: make it more clear when k8s_yaml accepts a string

jazzdan:


nicks:
I'm confused...the PR description is "k8s yaml accepts string", but the PR is "k8s yaml does not accept string"?

jazzdan:
Whoops, sorry @nicks that was an old title. Re-titling.

internal/tiltfile2/tiltfile_test.go
```diff
@@ -746,6 +746,20 @@ services:
...
 }
 
+func TestK8sYAMLInput(t *testing.T) {
+	f := newFixture(t)
+	defer f.TearDown()
+
+	f.setupFoo()
+	f.WriteFile("bar.yaml", "im not yaml")
+	f.file("Tiltfile", `
+k8s_yaml('bar.yaml')
```
nicks: can you also add a test case for something like

```
k8s_yaml(str(read_file('foo.yaml')))
```

internal/tiltfile2/k8s.go
```diff
@@ -132,10 +133,22 @@ func (s *tiltfileState) yamlEntitiesFromSkylarkValue(v skylark.Value) ([]k8s.K8s
 		if err != nil {
 			return nil, err
 		}
-		return k8s.ParseYAMLFromString(string(bs))
+		entities, err := k8s.ParseYAMLFromString(string(bs))
```
nicks: would be slightly more idiomatic for the err != nil check to live here instead of the handle function

PR #903, opened 2019-01-02
store: batch actions together so that a flood of LogActions doesn't overwhelm the subscribers

nicks:
Hello @landism, @maiamcc,

Please review the following commits I made in branch nicks/actions:

333607fc775e39925305fcc1837741d6cf199613 (2019-01-02 16:09:15 -0500)
store: batch actions together so that a flood of LogActions doesn't overwhelm the subscribers


nicks:
also added a bunch of tests because this code was sadly under-tested!

PR #904, opened 2019-01-02
vendor: update to new starlark

jazzdan:
This will allow us to enable if statements and other control structures outside of function bodies.


Gopkg.toml
```diff
@@ -57,6 +57,10 @@
   name = "k8s.io/client-go"
   version = "kubernetes-1.12.1"
 
+[[constraint]]
+  name = "go.starlark.net/starlark.git"
+  branch = "options"
```
nicks: hahaha ok

PR #906, opened 2019-01-02
Jazzdan/starlark enable global reassign

jazzdan:


internal/tiltfile2/tiltfile.go
```diff
@@ -20,6 +20,7 @@ const unresourcedName = "k8s_yaml"
 func init() {
 	resolve.AllowLambda = true
 	resolve.AllowNestedDef = true
+	resolve.AllowGlobalReassign = true
```
jazzdan: Here's where we enable top-level if statements.

jazzdan:
@nicks I feel like that might cross in to the territory of just testing things that skylark's tests cover, but it's easy enough so I'll do it.

jazzdan:
Actually good thing I did! I think I ran in to some more dep weirdness.

PR #910, opened 2019-01-03
cli: down works on gyaml

jazzdan:
From Kubernetes slack:

>Quick question: should `tilt down` delete all resources created from `k8s_yaml`? I’m not seeing an easy way to clean up configmaps etc created by `tilt up`.
>
> I’m seeing this with any resource that doesn’t reference an image that tilt knows about… so if my Tiltfile just contains
>
>```k8s_yaml("pod.yaml")```
>
>`tilt up` creates the pod, but it remains after `tilt down`

We simply weren't also deleting the global YAML resources that we record in the Tiltfile when down is called.

While I was here I made a change to the k8s client to print out which resources we're deleting, similarly to how we do it in `Upset`. Looks like this:

```
[~/go/src/github.com/windmilleng/servantes]> tilt down
Deleting via kubectl: Deployment/dan-fe
Deleting via kubectl: Service/dan-fe
Deleting via kubectl: Deployment/dan-vigoda
Deleting via kubectl: Deployment/dan-snack
Deleting via kubectl: Deployment/dan-doggos
Deleting via kubectl: Deployment/dan-fortune
Deleting via kubectl: Deployment/dan-hypothesizer
Deleting via kubectl: Service/dan-hypothesizer
Deleting via kubectl: Deployment/dan-spoonerisms
Deleting via kubectl: Deployment/dan-emoji
Deleting via kubectl: Deployment/dan-words
Deleting via kubectl: Deployment/dan-secrets
Deleting via kubectl: Service/dan-secrets
Deleting via kubectl: Role/pod-reader
Deleting via kubectl: RoleBinding/read-pods
Deleting via kubectl: Secret/dan-servantes-stuff
```

### DIFF
--- a/internal/k8s/client.go
+++ b/internal/k8s/client.go
@@ -11,15 +11,14 @@ import (
 
 	"github.com/pkg/errors"
 	"github.com/windmilleng/tilt/internal/container"
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/validation"
 	"k8s.io/apimachinery/pkg/watch"
 	apiv1 "k8s.io/client-go/kubernetes/typed/core/v1"
 	"k8s.io/client-go/rest"
 
 	"github.com/docker/distribution/reference"
-	"github.com/opentracing/opentracing-go"
-	"github.com/pkg/browser"
+	opentracing "github.com/opentracing/opentracing-go"
 	"github.com/windmilleng/tilt/internal/logger"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/tools/clientcmd"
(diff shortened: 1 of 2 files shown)


jazzdan:
I didn't write a test because it seemed hard to test and isn't testing a whole lot of logic. Open to ideas though!

internal/k8s/client.go
```diff
@@ -108,11 +107,6 @@ func NewK8sClient(
 	restConfig *rest.Config,
 	pf PortForwarder) K8sClient {
 
-	// TODO(nick): I'm not happy about the way that pkg/browser uses global writers.
-	writer := logger.Get(ctx).Writer(logger.DebugLvl)
```
nicks: why did you remove this? this is pretty important

internal/k8s/client.go
```diff
@@ -235,6 +229,13 @@ func maybeImmutableFieldStderr(stderr string) bool {
 // Currently ignores any "not found" errors, because that seems like the correct
 // behavior for our use cases.
 func (k K8sClient) Delete(ctx context.Context, entities []K8sEntity) error {
+	l := logger.Get(ctx)
+	var b strings.Builder
+	for _, e := range entities {
+		b.WriteString(fmt.Sprintf("Deleting via kubectl: %s/%s\n", e.Kind.Kind, e.Name()))
+	}
+	l.Infof(b.String())
```
nicks: This will crash if the string contains any % signs. 

You should either use Infof inside the loop (rather than have a separate Printf) or use something like Infof("%s", string)

internal/k8s/client.go
jazzdan: Yeah I realized that and added it back.

PR #911, opened 2019-01-03
web: live reload HUD updates

nicks:
Hello @yuindustries,

Please review the following commits I made in branch nicks/ch1176/prototype:

1954a7f8224c4b350d47781f8c2ff37e8d995dd8 (2019-01-03 13:54:52 -0500)
web: live reload HUD updates

### DIFF
--- a/web/src/App.js
+++ b/web/src/App.js
@@ -5,15 +5,20 @@ import './App.css';
 class App extends Component {
   constructor(props) {
     super(props)
-    this.state = {Resources: []}
+    this.state = {
+      Message: '',
+      View: {Resources: []},
+    }
   }
 
   render() {
     let el = null
-    if (!this.state.Resources || !this.state.Resources.length) {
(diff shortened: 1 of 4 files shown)


web/src/App.js
```diff
@@ -5,15 +5,20 @@ import './App.css';
...
+      Message: '',
+      View: {Resources: []},
+    }
   }
 
   render() {
     let el = null
-    if (!this.state.Resources || !this.state.Resources.length) {
-      el = <LoadingScreen />
+    let view = this.state.View
```
nicks: i struggled a bit with whether to make the capitalization rules consistent with Go or with JS and decided to try to use the Go conventions for now...tho could easily be convinced otherwise...

yuindustries: I'm fine with this decision! I'll let you know if I develop a stronger opinion later on.

PR #913, opened 2019-01-03
engine: paramaterize Tiltfile

jazzdan:
This allows us to specify a different Tiltfile to load with a command line flag.

### DIFF
--- a/internal/cli/up.go
+++ b/internal/cli/up.go
@@ -8,7 +8,7 @@ import (
 	"time"
 
 	"github.com/fatih/color"
-	"github.com/opentracing/opentracing-go"
+	opentracing "github.com/opentracing/opentracing-go"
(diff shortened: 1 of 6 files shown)


nicks:
AFAIR, this code is super-sensitive to absolute paths vs relative paths. you need both and to use them consistently -- absolute for certain cases (like equality checking) and relative for other cases (like error messages)

That's why the old code used tiltfile2.FileName to Load(). It's a relative path. You've changed it so that one call uses a relative path and the other uses an absolute path, which is going to lead to weird issues

i think what we probably need to do is store the relative path on the engine state, and add a method for the absolute path. Does that make sense? I can explain more in person; it's pretty finicky.

jazzdan:
Yup, that makes sense. Will do.

PR #917, opened 2019-01-03
tilt: pull k8s fields out to K8SResourceInfo

landism:


internal/hud/renderer.go
```diff
@@ -217,22 +223,27 @@ func bestLogs(res view.Resource) string {
...
-	// Two cases:
-	// 1) The last build finished before this pod started
-	// 2) This log is from an in-place container update.
-	// in either case, prepend them to pod logs.
-	if (res.LastBuild().StartTime.Equal(res.PodUpdateStartTime) ||
-		res.LastBuild().StartTime.Before(res.PodCreationTime)) &&
-		len(res.LastBuild().Log) > 0 {
-		return string(res.LastBuild().Log) + "\n" + res.PodLog
-	}
+	// TODO(matt) figure out what to do with DC logs once we have DC timestamps
```
maiamcc: we've already got a type switch for DC further up in this function, put this todo up there

if it's NOT either DC or k8s, do we want to freak out?

internal/hud/renderer.go
```diff
@@ -217,22 +223,27 @@ func bestLogs(res view.Resource) string {
...
+
+	if res.IsK8S() {
```
maiamcc: i'd do `if kri := m.K8sInfo(); !kri.Empty() {...}` to save a line, but 🤷‍♀️ 
also that stands for... kubernetes resource info? i don't care much but it's inconsistent with naming scheme i used for . `dcInfo`.

internal/hud/renderer_test.go
```diff
@@ -24,6 +24,7 @@ func TestRender(t *testing.T) {
 			{
 				Name:               "foo",
 				DirectoriesWatched: []string{"bar"},
+				ResourceInfo:       view.K8SResourceInfo{},
```
maiamcc: what purpose does this serve?

internal/hud/resourceview.go
```diff
@@ -386,16 +373,15 @@ func (v *ResourceView) resourceExpandedError() rty.Component {
...
-// TODO(maia): rename this method to be generic (thiiink it already works with k8s AND dc?)
-func (v *ResourceView) resourceExpandedK8sError() (rty.Component, bool) {
+func (v *ResourceView) resourceExpandedRuntimeError() (rty.Component, bool) {
 	pane := rty.NewConcatLayout(rty.DirVert)
 	ok := false
 	if isCrashing(v.res) {
-		podLog := v.res.CrashLog
-		if podLog == "" {
-			podLog = v.res.PodLog
+		runtimeLog := v.res.CrashLog
```
maiamcc: niiiiice rename 👌

internal/hud/view/view.go
```diff
@@ -9,22 +9,55 @@ import (
...
 	resourceInfoView()
+	RuntimeLog() string
+	Status() string
 }
 
 type DCResourceInfo struct {
 	ConfigPath string
-	Status     string
+	status     string
 	Log        string
```
maiamcc: do you want to make `log` private as well now that it has an accessor method? (i don't actually have an opinion here.)

internal/hud/view/view.go
```diff
@@ -9,22 +9,55 @@ import (
 
 type ResourceInfoView interface {
```
maiamcc: hmm should the resourceinfo stuff be its own file?

internal/hud/renderer_test.go
landism: It causes the resource to be rendered as a k8s resource, which seems like it makes for a more interesting UI test, but something's smelly about adding this everywhere.

internal/hud/view/view.go
landism: oh, sure...

(fwiw, making status private wasn't about encapsulation as much as allowing it to compile by removing the name collision)

internal/hud/renderer_test.go
maiamcc: ...oh ugh i see what you mean. uhhh add a note/todo maybe? 🤷‍♀️ 

also, in your code, `isK8s()` returns `true` if resource info is `K8SResourceInfo{}` (i.e. the right type, just empty); the `isDC()` method behaves slightly differently, I think -- requires that it be both `DockerComposeResourceInfo` AND non-empty. but i think your way makes more sense. maybe change the dc method to be consistent?

(if you wanna be really thorough, could you also change `manifest.IsDC`/`IsK8s` and `DockerInfo.IsFastBuild`/`IsStaticBuild`?)

internal/hud/renderer.go
landism: I didn't implement `Empty` for k8sInfo, mostly because I've got an engrained bias against reflection that hasn't proven necessary, but maybe I'm just being squeamish and that's premature optimization.

Anyway, per your other [comment](https://github.com/windmilleng/tilt/pull/917#discussion_r245344865), I'll change other things to match this.

internal/hud/renderer.go
landism: nope, it might be yaml

internal/hud/renderer_test.go
landism: > maybe change the dc method to be consistent?
done

> (if you wanna be really thorough, could you also change manifest.IsDC/IsK8s and DockerInfo.IsFastBuild/IsStaticBuild?)
as a separate PR

internal/hud/view/view.go
landism: maybe, but that's easy to do later, and we're about to have a few commits moving fields between these types, so might as well leave them together for now?

internal/hud/renderer_test.go
landism: > as a separate PR

https://github.com/windmilleng/tilt/pull/922

PR #922, opened 2019-01-04
tilt: replace some Empty()s with typechecks

landism:


internal/model/manifest.go
```diff
@@ -36,43 +36,43 @@ type Manifest struct {
 type DockerBuildArgs map[string]string
 
 func (m Manifest) StaticBuildInfo() StaticBuild {
-	if info, ok := m.DockerInfo.buildDetails.(StaticBuild); ok {
-		return info
-	}
-	return StaticBuild{}
+	ret, _ := m.DockerInfo.BuildDetails.(StaticBuild)
```
maiamcc: as discussed, maybe change these to return `ret, bool`

internal/model/manifest.go
```diff
@@ -95,34 +95,40 @@ func (m Manifest) Dockerignores() []Dockerignore {
...
+		for i, mount := range bd.Mounts {
 			result[i] = mount.LocalPath
 		}
 		return result
-	} else if dcInfo := m.DCInfo(); !dcInfo.Empty() {
-		result := make([]string, len(dcInfo.Mounts))
-		for i, mount := range dcInfo.Mounts {
-			result[i] = mount.LocalPath
+	default:
+		// TODO(matt?) DC mounts should probably be on BuildDetails, not DeployInfo
```
maiamcc: ...imo we should put them somewhere else entirely, but they're not relevant to the build for DC manifests cuz there is no build (that's in our control)

internal/tiltfile/tiltfile_test.go
```diff
@@ -977,18 +977,18 @@ func (f *fixture) assertManifest(name string, opts ...interface{}) model.Manifes
 		case []model.PortForward:
 			assert.Equal(f.t, opt, m.K8sInfo().PortForwards)
 		case dcConfigPathHelper:
-			dcInfo := m.DCInfo()
-			if assert.False(f.t, dcInfo.Empty(), "expected a docker-compose manifest") {
+			if assert.True(f.t, m.IsDC(), "expected a docker-compose manifest") {
```
maiamcc: 🤔 is there weird merge stuff going on? i think this lives in `tiltfile_docker_compose_test.go` on master...?

PR #924, opened 2019-01-04
docker-compose: integration tests

nicks:
Hello @maiamcc,

Please review the following commits I made in branch nicks/ch1128/integration:

bb849691a2070057522d7e3303e95c10013f1288 (2019-01-04 16:23:52 -0500)
docker-compose: integration tests

### DIFF
--- a/integration/fixture_test.go
+++ b/integration/fixture_test.go
@@ -88,19 +88,74 @@ func (f *fixture) Curl(url string) (string, error) {
 }
 
 func (f *fixture) CurlUntil(ctx context.Context, url string, expectedContents string) {
+	f.WaitUntil(ctx, fmt.Sprintf("curl(%s)", url), func() (string, error) {
+		return f.Curl(url)
+	}, expectedContents)
+}
+
+func (f *fixture) WaitUntil(ctx context.Context, msg string, fun func() (string, error), expectedContents string) {
 	for {
-		actualContents, err := f.Curl(url)
+		actualContents, err := fun()
 		if err == nil && strings.Contains(actualContents, expectedContents) {
 			return
 		}
 
 		select {
 		case <-ctx.Done():
-			f.t.Fatalf("Timed out waiting for server %q to return contents %s. "+
-				"Current error:\n%v\nCurrent contents:\n%s\n",
-				url, expectedContents, err, actualContents)
+			f.t.Fatalf("Timed out waiting for expected result (%s)\n"+
+				"Expected: %s\n"+
+				"Actual: %s\n"+
+				"Current error: %v\n",
+				msg, expectedContents, actualContents, err)
 		case <-time.After(200 * time.Millisecond):
 		}
+	}
+}
+
+func (f *fixture) dockerCmd(args []string, outWriter io.Writer) *exec.Cmd {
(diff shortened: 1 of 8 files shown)


integration/onedc/Dockerfile
```diff
@@ -0,0 +1,5 @@
+FROM golang:1.11
+WORKDIR /go/src/github.com/windmilleng/integration/onedc
+ADD . .
```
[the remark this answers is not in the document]
nicks: that's why we have an integration test ! :)

nicks:
it's just red bc i screwed up installing docker-compose inside the container blargh

maiamcc:
bump -- @nicks what's the status of this?

nicks:
currently tabled for higher priority things until i can get back to it

nicks:
basically the problem is with the way that docker works inside circleci. our test process runs in container A, but the docker server is running in container B, so when docker-compose exposes a port it's not available to our test process

nicks:
It works!!! Needed to do some docker-fu. please take another look?

integration/onedc_test.go
```diff
@@ -0,0 +1,49 @@
...
+
+	f.dockerKillAll("onedc_web")
+	f.TiltWatch("web")
+
+	ctx, cancel := context.WithTimeout(f.ctx, time.Minute)
+	defer cancel()
+
+	f.WaitUntil(ctx, "onedc_web up", func() (string, error) {
+		out := &bytes.Buffer{}
+		cmd := f.dockerCmd([]string{
```
maiamcc: oooh good idea!

integration/onedc_test.go
```diff
@@ -0,0 +1,49 @@
...
+			"exec", cID, "curl", "-s", "localhost:8000",
+		}, out)
+		err := cmd.Run()
+		return out.String(), err
+	}, "🍄 One-Up! 🍄")
+
+	// TODO(nick): uncomment when file-watching works
+	// f.ReplaceContents("main.go", "One-Up", "Two-Up")
```
maiamcc: does this work now?

integration/fixture_test.go
```diff
@@ -88,22 +88,75 @@ func (f *fixture) Curl(url string) (string, error) {
 }
 
 func (f *fixture) CurlUntil(ctx context.Context, url string, expectedContents string) {
+	f.WaitUntil(ctx, fmt.Sprintf("curl(%s)", url), func() (string, error) {
+		return f.Curl(url)
+	}, expectedContents)
+}
+
+func (f *fixture) WaitUntil(ctx context.Context, fnName string, fun func() (string, error), expectedContents string) {
```
maiamcc: nit: instead of `fnName`, consider `msg`, because that's more like what it is--and the fatal is something like:
```
Timed out waiting for expected result (<msg>)
expected: <exp>
actual: <act>
```

integration/onedc_test.go
nicks: going to iterate on this in a separate pr

integration/fixture_test.go
nicks: i like that! done

PR #930, opened 2019-01-06
Update README.md

dosire:
https://news.ycombinator.com/item?id=18834398


nicks:
Thanks! This is a good step in the right direction, going to merge it and re-work a bit into the other copy.

PR #935, opened 2019-01-07
DC: no red on container rebuild

landism:
this makes the error count in the status bar match the number of red dots, which is what we want anyway

luckily for the purposes of ch1212, `statusColor` is already correct for the "restarting container after rebuild" case


internal/hud/renderer.go
```diff
@@ -188,10 +188,8 @@ func keyLegend(v view.View, vs view.ViewState) string {
 	return defaultKeys
 }
 
-func isInError(res view.Resource) bool {
-	return res.LastBuild().Error != nil ||
-		(res.ResourceInfo != nil && statusColors[res.ResourceInfo.Status()] == cBad) ||
-		isCrashing(res)
+func isInError(res view.Resource, triggerMode model.TriggerMode) bool {
```
maiamcc: it uh. seems somewhat questionable to made this decision by status color and not the other way around? but also eh, it's no worse than it was before 😅

landism:
> hunhhh is statusColor really correct in this case? doesn't look like it will be from the code...? but i trust you...

I haven't dug into the DC code, but I've verified there's no flicker. I think it's taken care of by the first if:
```
	if !res.CurrentBuild.Empty() && !res.CurrentBuild.Reason.IsCrashOnly() {
		return cPending
```

As long as there is currently a build running, we make it yellow. Apparently DC doesn't consider the build done until the container is back up? (at least, the build status timer keeps ticking until the container status is up)

internal/hud/renderer.go
landism: yeah, really we should have an enum for bad/pending/good, make `statusColor` return that enum, and delete `isInError`

maybe next time

PR #939, opened 2019-01-07
engine: docker compose build logs show up in manifest specific log

jazzdan:
Take 2!


internal/dockercompose/fake_client.go
```diff
@@ -32,7 +36,9 @@ func (c *FakeDCClient) Down(ctx context.Context, pathToConfig string, stdout, st
 }
 
 func (c *FakeDCClient) StreamLogs(ctx context.Context, pathToConfig, serviceName string) (io.ReadCloser, error) {
-	return ioutil.NopCloser(bytes.NewReader([]byte(c.logOutput))), nil
+	output := c.RunLogOutput[model.ManifestName(serviceName)]
+	fmt.Printf("Got output: %s for manifest %s\n", output, serviceName)
```
nicks: ?

internal/engine/upper_test.go
```diff
@@ -90,6 +92,10 @@ func (b *fakeBuildAndDeployer) nextBuildResult(ref reference.Named) store.BuildR
 }
 
 func (b *fakeBuildAndDeployer) BuildAndDeploy(ctx context.Context, manifest model.Manifest, state store.BuildState) (store.BuildResult, error) {
+	if manifest.IsDC() {
```
nicks: why would this be specific to docker-compose?

internal/engine/upper_test.go
```diff
@@ -2233,6 +2255,14 @@ func (f *testFixture) setupDCFixture() (redis, server model.Manifest) {
 	return manifests[0], manifests[1]
 }
 
+func (f *testFixture) setDCBuildLogOutput(mn model.ManifestName, output string) {
+	f.b.buildOutput[mn] = output
```
nicks: again, this doesn't look docker-compose specific

jazzdan: That's fair, I'll remove that restriction.

jazzdan: (It's now removed)

internal/engine/upper_test.go
```diff
@@ -74,6 +74,8 @@ type fakeBuildAndDeployer struct {
 
 	// Set this to simulate the build failing. Do not set this directly, use fixture.SetNextBuildFailure
 	nextBuildFailure error
+
+	buildOutput map[model.ManifestName]string
```
maiamcc: consider `buildLogOutput` to match naming in fakeDcCli

internal/engine/upper_test.go
```diff
@@ -1729,37 +1737,52 @@ func TestDockerComposeEventSetsStatus(t *testing.T) {
...
 	m, _ := f.setupDCFixture()
-	expected := "spoonerisms_1  | 2018-12-20T16:11:04.070480042Z yarn install v1.10."
-	f.dcc.SetLogOutput(expected + "\n")
+	expected := "yarn install"
+	f.setBuildLogOutput(m.ManifestName(), expected)
 
-	f.Start([]model.Manifest{m}, true)
-	f.waitForCompletedBuildCount(1)
+	f.loadAndStart()
+	f.waitForCompletedBuildCount(2)
```
maiamcc: wait why are there two builds now?

internal/hud/renderer_test.go
```diff
@@ -355,6 +355,11 @@ func TestRenderLogModal(t *testing.T) {
 					"Hi hello I'm a docker compose log",
 					time.Now().Add(time.Second*-12),
 				),
+				BuildHistory: []model.BuildStatus{
+					model.BuildStatus{
+						Log: []byte("Hi hello I'm a docker compose build log"),
```
maiamcc: v good dummy text 👍

internal/engine/upper_test.go
jazzdan: The docker compose fixture now has two services ("redis" and "server), before it only had one.

maiamcc: oh. just call `f.Start([]model.Manifest{m}, true)`, we've already loaded above. there are two for the cases where they're needed/useful, but should probs just use one unless it's a specific test for >1 manifest.

PR #948, opened 2019-01-08
docs: add faq on docker build options

nicks:
Hello @jazzdan, @maiamcc,

Please review the following commits I made in branch nicks/ch1235/faq:

cdfdf71cb674717effdbff667d5fd9f21c9cbc78 (2019-01-08 14:09:20 -0500)
docs: add faq on docker build options


docs/faq.md
```diff
@@ -0,0 +1,47 @@
+Frequently Asked Questions
+==========================
+
+Building Container Images
+-------------------------
+
+*Q: All the Tilt examples store the image at `gcr.io/windmill-public-containers`. Isn't it really slow to push images up to Google's remote repository for local development?
```
jazzdan: Could we make this an optional argument for local development cases to simplify things further?

nicks: we chatted about this offline. There's a bunch of open questions here about:
- image URL matching from Tiltfile to k8s yaml
- local vs remote clusters and when you need an explicit repository
- how much we can infer vs auto-generate for you
but that's beyond the scope of this pr!

PR #949, opened 2019-01-08
build: clear UID and GID before adding to tar

jazzdan:
I couldn't think of a good way to test this since `chown` requires root. Open to any ideas!

### DIFF
--- a/internal/testutils/tar.go
+++ b/internal/testutils/tar.go
@@ -7,12 +7,17 @@ import (
 	"testing"
 )
 
+const expectedUidAndGid = 0
+
 type ExpectedFile struct {
 	Path     string
 	Contents string
 
 	// If true, we will assert that the file is not in the tarball.
 	Missing bool
+
+	// If true, we will assert that UID and GIF are 0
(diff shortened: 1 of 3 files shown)


jazzdan:
@nicks ah good point, I was thinking of testing it from the file system all the way through the image build code. But that definitely works. :)

jazzdan:
(updated with test)

internal/testutils/tar.go
```diff
@@ -7,12 +7,17 @@ import (
...
+
 type ExpectedFile struct {
 	Path     string
 	Contents string
 
 	// If true, we will assert that the file is not in the tarball.
 	Missing bool
+
+	// If true, we will assert that UID and GIF are 0
+	AssertUidAndGidAreZero bool
```
nicks: should we just assert this all the time?

jazzdan: No because we use this in other tests where we assert that files exist in docker images.

I found that out the hard way because that was my first approach, heh.

PR #950, opened 2019-01-08
tilt: p is for profile

landism:
'p' now starts and stops a pprof profile

### DIFF
--- a/internal/hud/hud.go
+++ b/internal/hud/hud.go
@@ -3,6 +3,7 @@ package hud
 import (
 	"context"
 	"fmt"
+	"os"
 	"sync"
 	"time"
 
@@ -37,6 +38,8 @@ type Hud struct {
 	currentViewState view.ViewState
 	mu               sync.RWMutex
 	isRunning        bool
+
+	currentProfileFile *os.File
 }
 
 var _ HeadsUpDisplay = (*Hud)(nil)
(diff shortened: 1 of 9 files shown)


internal/store/engine_state.go
```diff
@@ -65,6 +65,9 @@ type EngineState struct {
 
 	TriggerMode  model.TriggerMode
 	TriggerQueue []model.ManifestName
+
+	Profiling   bool
+	ProfileFile *os.File
```
nicks: it seems like a bad idea to have a pointer to an external OS thing in the engineState, especially as we start cloning/serializing/copying this state to the web ui, protobuf, etc

internal/hud/hud.go
```diff
@@ -37,6 +38,8 @@ type Hud struct {
 	currentViewState view.ViewState
 	mu               sync.RWMutex
 	isRunning        bool
+
+	currentProfileFile *os.File
```
nicks: is this used anywhere?

maiamcc:
https://www.youtube.com/watch?v=_dvPhtNZCj0

internal/store/engine_state.go
landism: Oh, I didn't realize that serializability was a requirement for EngineState, and I didn't really see anywhere else Upper stores state. I'll inject something into Upper to hold the state.

internal/hud/hud.go
landism: cruft from an earlier iteration. thanks!

PR #952, opened 2019-01-08
hud: separate "unresourced" from other K8s entities

yuindustries:
- Unique list of "unresourced" items
- Separate out non-Deployment / Service items

2019-01-08 - k8s_yaml unresourced and other


internal/hud/resourceview.go
```diff
@@ -186,15 +186,54 @@ func (v *ResourceView) resourceExpandedYAML() rty.Component {
...
+					kinds = append(kinds, res.Kind)
+				}
+			}
+			unresourced = appendIfMissing(unresourced, fmt.Sprintf("- %s (%s)", r.Name, strings.Join(kinds, ", ")))
+		} else {
+			otherK8s = append(otherK8s, fmt.Sprintf("- %s (%s)", r.Name, r.Kind))
+		}
+	}
+
+	return strings.Join(unresourced, "\n"), strings.Join(otherK8s, "\n")
```
yuindustries: If there's a more elegant or succinct way to write this code — please, I'm all 👂🏼!

internal/hud/resourceview.go
```diff
@@ -186,15 +186,54 @@ func (v *ResourceView) resourceExpandedYAML() rty.Component {
...
 	l.AddDynamic(rhs)
 	return l
 }
 
+func unresourcedK8sItems(resources []model.YAMLManifestResource) (string, string) {
+	var unresourced []string
+	var otherK8s []string
+
+	for _, r := range resources {
+		if r.Kind == "Deployment" || r.Kind == "Service" {
```
landism: ugh, maybe we really should have done this further up the chain.

It turns out the code that joins deployments and services to a manifest actually doesn't refer to deployment or service at all!

It looks like it first pulls out entities with images (which could also be, e.g., jobs or daemonsets) and then pulls out entities with labels that match those entities ([here](https://github.com/windmilleng/tilt/blob/b1a699887da864bb598f1ebb0d20dd251ea56a8c/internal/tiltfile/tiltfile_state.go#L238)).

We should probably have the tiltfile code also group unresourced things if we want to do this right.

internal/tiltfile/tiltfile.go
```diff
@@ -74,12 +74,18 @@ func Load(ctx context.Context, filename string, matching map[string]bool) (manif
 			return nil, model.YAMLManifest{}, nil, err
 		}
 
-		var resourceNames []string
+		var resources []model.YAMLManifestResource
+
 		for _, e := range unresourced {
-			resourceNames = append(resourceNames, fmt.Sprintf("%s (%s)", e.Name(), e.Kind.Kind))
+			// if e.Kind.Kind == "Deployment" || e.Kind.Kind == "Service" {
```
landism: crufty comments

yuindustries:
I'm closing this PR because Maia's work on Unbuilt Resources (CH Epic ["Visualize servers without Dockerfiles"](https://app.clubhouse.io/windmill/epic/1466/visualize-servers-without-dockerfiles)) will pull out anything with pods and display them as resource items above k8s_yaml. 

This PR was intended to reduce the number of items under k8s_yaml and make them less redundant. That's less important with the change above.

PR #953, opened 2019-01-08
k8s: handle CRDs properly

nicks:
Hello @landism, @maiamcc,

Please review the following commits I made in branch nicks/ch1236/crd:

9dd876eb84ee94e52399a0b061c61d363035a269 (2019-01-08 17:27:42 -0500)
k8s: handle CRDs properly

### DIFF
--- a/internal/k8s/serialize_test.go
+++ b/internal/k8s/serialize_test.go
@@ -8,12 +8,16 @@ import (
 	"github.com/windmilleng/tilt/internal/k8s/testyaml"
 )
 
-func TestTracerYAML(t *testing.T) {
-	entities, err := ParseYAMLFromString(testyaml.TracerYAML)
+func MustParseYAMLFromString(t *testing.T, s string) []K8sEntity {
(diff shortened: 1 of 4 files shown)


internal/k8s/serialize_test.go
```diff
@@ -73,3 +73,32 @@ func TestHelmGeneratedRedisYAML(t *testing.T) {
 		"StatefulSet",
 	}, kinds)
 }
+
+func TestCRDYAML(t *testing.T) {
+	entities, err := ParseYAMLFromString(testyaml.CRDYAML)
+	assert.NoError(t, err)
```
maiamcc: you should `if err != nil {t.Fatal(err)}` here so we don't continue to execute even if this step is borked and then get weird and confusing errors further down the test.

nicks:
did you see the test case? there is a good example of a YAML config that didn't work before, and works now

internal/k8s/serialize_test.go
nicks: i added a helper for it!

PR #958, opened 2019-01-09
DC: don't unnecessarily talk to k8s

landism:
result: no longer crashes when running only DC


internal/engine/upper.go
```diff
@@ -97,11 +95,6 @@ func (u Upper) Start(ctx context.Context, args []string, watchMounts bool, trigg
 
 	startTime := time.Now()
 
-	err := u.kcli.ConnectedToCluster(ctx)
```
nicks: did you intend to run the ConnectedToCluster call somewhere else?

landism: I didn't. I started looking for somewhere to move it to, didn't see anywhere great, and then didn't see a compelling reason to keep it in. Do you?

nicks: ya, there were a couple bugs filed about it. It was a common onboarding complaint that Tilt would just hang indefinitely and/or spit out cryptic errors if it couldn't connect to K8s.

maybe move this into a separate subscriber that tries to connect on startup if necessary, and emits a permanent error if it can't?

nicks: maybe there should be some sort of K8sConnected field in EngineState, and the pod watcher/service watcher only tries to connect if we've verified the connection?

landism: > It was a common onboarding complaint that Tilt would just hang indefinitely and/or spit out cryptic errors if it couldn't connect to K8s.

Oh, OK. I actually tested this case before pushing the PR and felt the error message was vastly improved by the removal of `CheckConnected`:

w/ CheckConnected:
```
$ tilt up
Error: Unable to connect to cluster via `kubectl cluster-info`:
stdout: Kubernetes master is running at https://localhost:6443

To further debug and diagnose cluster problems, use 'kubectl cluster-info dump'.

stderr: The connection to the server localhost:6443 was refused - did you specify the right host or port?
: exit status 1
```

w/o:
```
$ tilt up
Error: Error watching pods. Are you connected to kubernetes?
: Get https://localhost:6443/api/v1/pods?labelSelector=tilt-runid%3D9cebd343-c1f4-4bb5-bcd6-a39de933cbf1&watch=true: dial tcp [::1]:6443: connect: connection refused
```

before:
1. tells you there's a cluster running even though there isn't
2. gives you a next step that's going to uselessly fail because there's no cluster running
3. asks you to specify host/port when in most cases that's not relevant

after:
says it has connection problems and doesn't give you misleading suggestions

I only found one bug related to this, and it predated the "Are you connected to Kubernetes?" text.

One might argue that we want a central check for k8s connectivity so that we don't have to rely on every single k8s callsite providing a good error message, or one might argue that we want to have good error messages at every k8s callsite regardless, because k8s might go down after startup.

I feel like this PR is a step forward as-is.

nicks: ok

PR #960, opened 2019-01-09
better type support for Events from docker compose [ch1277]

maiamcc:
Hello @jazzdan, @nicks,

Please review the following commits I made in branch maiamcc/dc-action-enum:

3a44420f5d96d862506408b56a6cfb138fe31ca8 (2019-01-09 14:01:04 -0500)
fix types while i'm here

2b5e07897242c019a6edc3c89ff4cbade3dfc706 (2019-01-09 13:53:55 -0500)
compose: change container actions to be enums, don't barf on unknown actions

### DIFF
--- a/internal/dockercompose/event.go
+++ b/internal/dockercompose/event.go
@@ -51,25 +51,46 @@ func (t *Type) UnmarshalJSON(b []byte) error {
 	return nil
 }
 
-type Action string
+type Action int
 
 const (
-	// Add 'actions' here (and to `UnmarshalJSON` below`) as we support them
+	// Add 'actions' here (and to `stringToAction` below`) as we support them
 
 	// CONTAINER actions
-	ActionAttach     Action = "attach"
-	ActionCreate     Action = "create"
-	ActionDie        Action = "die"
-	ActionExecDie    Action = "exec_die"
-	ActionExecAttach Action = "exec_attach"
-	ActionExecCreate Action = "exec_create"
-	ActionKill       Action = "kill"
-	ActionRename     Action = "rename"
-	ActionRestart    Action = "restart"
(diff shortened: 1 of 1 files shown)


jazzdan:
I'm confused: how does `ActionUnknown` ever get set?

maiamcc:
> I'm confused: how does `ActionUnknown` ever get set?

```action := stringToAction[s]
*a = action
return nil```
if there's no match in the map, we return the zero value of `Action` -- which, as an int, is `0`, i.e. `ActionUnknown`. Will add a comment.

File: internal/dockercompose/event.go [hunk 1]
```diff
@@ -30,79 +30,122 @@ func EventFromJsonStr(j string) (Event, error) {
...
+	"restart":       ActionRestart,
+	"start":         ActionStart,
+	"stop":          ActionStop,
+	"top":           ActionTop,
+	"unpause":       ActionUnpause,
+	"update":        ActionUpdate,
+}
+
+func (a Action) String() string {
+	for str, act := range stringToAction {
```
jazzdan: Why not make it an action to string map, and then do a lookup in to it rather than iterating through the entire map?

internal/dockercompose/event.go
```diff
@@ -30,79 +30,122 @@ func EventFromJsonStr(j string) (Event, error) {
...
+	TypeUnknown Type = iota
+	TypeContainer
 )
 
+var stringToType = map[string]Type{
+	"container": TypeContainer,
+}
+
+func (t Type) String() string {
+	for str, typ := range stringToType {
```
jazzdan: Why not make it a type to string map, and then do a lookup in to it rather than iterating through the entire map?

PR #961, opened 2019-01-09
.: fix a few bugs

dbentley:
In trying to repro cases, I came across a few weirdnesses. Some crashes, some UI oddities.

What's the right way to get these fixed? Should I go add tests for each, or hand this off to someone?

### DIFF
--- a/internal/hud/renderer.go
+++ b/internal/hud/renderer.go
@@ -74,7 +74,7 @@ func (r *Renderer) layout(v view.View, vs view.ViewState) rty.Component {
 	var ret rty.Component = l
 	if vs.LogModal.TiltLog == view.TiltLogFullScreen {
 		ret = r.renderTiltLog(v, vs, keyLegend(v, vs), ret)
-	} else if vs.LogModal.ResourceLogNumber != 0 {
+	} else if vs.LogModal.ResourceLogNumber != 0 && len(v.Resources) > 0 {
 		ret = r.renderResourceLogModal(v.Resources[vs.LogModal.ResourceLogNumber-1], ret)
 	}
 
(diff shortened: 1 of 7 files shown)


nicks:
i'm not sure i completely understand the rty / renderer changes or how to test them, maybe @landism would be a better reviewer?

internal/hud/renderer.go
```diff
@@ -74,7 +74,7 @@ func (r *Renderer) layout(v view.View, vs view.ViewState) rty.Component {
 	var ret rty.Component = l
 	if vs.LogModal.TiltLog == view.TiltLogFullScreen {
 		ret = r.renderTiltLog(v, vs, keyLegend(v, vs), ret)
-	} else if vs.LogModal.ResourceLogNumber != 0 {
+	} else if vs.LogModal.ResourceLogNumber != 0 && len(v.Resources) > 0 {
```
landism: This doesn't seem like it's correctly solving a problem. What behavior led to this change?

ResourceLogNumber is 1-indexed, so that the default value of 0 means no log.

There's a bug here if you have a log modal open while you delete that log, but that's not a 0/1 issue - if you have resource 4's log open and delete resource 2, you'll now be looking at resource 5 (or get an error if there is no resource 5).

dbentley: Open Tilt with no Tiltfile and press enter; Tilt dies. Would love to prevent this.

internal/rty/scroll.go
```diff
@@ -408,13 +408,21 @@ func adjustElementScroll(prevInt interface{}, newChildren []string) (*ElementScr
...
+		return next, ""
+	}
 	if len(prev.children) == 0 {
 		sel := ""
 		if len(next.children) > 0 {
 			sel = next.children[0]
 		}
 		return next, sel
 	}
+	if prev.elementIdx >= len(prev.children) {
```
landism: I'd probably add a comment along the lines of "this should be impossible, but it happened, so here we are"

PR #963, opened 2019-01-09
hud: make headings and statuses generic [ch1221] [ch1222]

maiamcc:

this commit
* replaces hud header "k8s" with "deploy status" (to go along with "build status")
* takes out "dc" from dc status; i.e. "dc up" --> "up"

looks like this:
screen shot 2019-01-09 at 4 27 34 pm

@yuindustries do the headers look ok to you? also should we consider flipping the order of "build status" and "deploy status" so that it reads left to right "build --> deploy"?


maiamcc:
ready for review!

PR #964, opened 2019-01-09
model: change all the names to move towards abstract build graph

nicks:
Hello @maiamcc, @jazzdan,

Please review the following commits I made in branch nicks/targets:

74ae9cb7e9559e39ce87758da9cbd2761d46318a (2019-01-09 16:38:26 -0500)
model: change all the names to move towards abstract build graph

### DIFF
--- a/internal/model/target.go
+++ b/internal/model/target.go
@@ -1,0 +1,29 @@
+package model
+
+import "fmt"
+
+// An abstract build graph of targets and their dependencies.
+// Each target should have a unique ID.
+
+type TargetType string
+
+const (
+	TargetTypeK8s           TargetType = "k8s"            // Deployed k8s entities
+	TargetTypeImage         TargetType = "image"          // Image builds
+	TargetTypeDockerCompose TargetType = "docker-compose" // Docker-compose service build+deploy
+)
+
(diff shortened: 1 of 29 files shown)


internal/model/target.go
```diff
@@ -0,0 +1,29 @@
+package model
```
nicks: the big change is this file. everything else is just renaming

nicks:
going to merge this now to avoid merge conflicts but will follow-up on any other revisions on names!

internal/model/target.go
```diff
@@ -0,0 +1,29 @@
...
+
+// An abstract build graph of targets and their dependencies.
+// Each target should have a unique ID.
+
+type TargetType string
+
+const (
+	TargetTypeK8s           TargetType = "k8s"            // Deployed k8s entities
+	TargetTypeImage         TargetType = "image"          // Image builds
+	TargetTypeDockerCompose TargetType = "docker-compose" // Docker-compose service build+deploy
```
maiamcc: worth including in the note that we treat it as a deploy target right now but that that's sort of a hack

nicks: done in the next pr

PR #973, opened 2019-01-10
dockercompose: support dockerignore and gitignore

landism:


internal/tiltfile/docker_compose.go
```diff
@@ -191,32 +192,42 @@ func parseDCConfig(ctx context.Context, configPath string) ([]dcService, error)
...
 	configFiles []string, err error) {
 	dcInfo := model.DockerComposeTarget{
 		ConfigPath: dcConfigPath,
-		YAMLRaw:    s.ServiceConfig,
-		DfRaw:      s.DfContents,
+		YAMLRaw:    service.ServiceConfig,
+		DfRaw:      service.DfContents,
 	}
+
+	dockerIgnores := pathsToDockerIgnores([]string{path.Dir(service.DfPath), path.Dir(dcConfigPath)})
```
maiamcc: hmm, you should probs do all this AFTER we get mounts -- e.g. what if we're mounting `./foo/bar/baz`, which is its own git repo, but the dockerfile lives in like `./docker`? (idk how likely that case is... but trying to get gitignore/dockerignore for every mounted path as well as for dependencies is closest to how we do it for k8s)

internal/tiltfile/tiltfile_docker_compose_test.go
```diff
@@ -190,6 +190,58 @@ RUN echo hi`
...
 
+func TestDockerComposeHonorsDockerIgnore(t *testing.T) {
+	f := newFixture(t)
+	defer f.TearDown()
+
+	df := `FROM alpine
+
+ADD . /app
+COPY ./thing.go /stuff
+RUN echo hi`
```
maiamcc: does the fact that this DF has stuff in it have anything to do with the behavior you're testing? nbd but it distracted me a lil 🤷‍♀️

internal/tiltfile/docker_compose.go
```diff
@@ -191,32 +192,42 @@ func parseDCConfig(ctx context.Context, configPath string) ([]dcService, error)
 	return services, nil
 }
 
-func (s dcService) ToManifest(dcConfigPath string) (manifest model.Manifest,
+func (s *tiltfileState) ToManifest(service dcService, dcConfigPath string) (manifest model.Manifest,
```
maiamcc: now that it's not being called ON a `dcService`, the name is less descriptive. maybe `DcServiceToManifest`? (or maybe the types indicate that for you, idk.)

internal/tiltfile/docker.go
```diff
@@ -361,15 +367,21 @@ func (s *tiltfileState) dockerignoresToDomain(image *dockerImage) []model.Docker
 		})
 	}
 
+	return result
+}
+
+func (s *tiltfileState) dockerignoresToDomain(image *dockerImage) []model.Dockerignore {
```
maiamcc: see above: `dockerignoresForImage` maybe?

internal/tiltfile/docker.go
```diff
@@ -325,34 +325,40 @@ func (s *tiltfileState) reposToDomain(image *dockerImage) []model.LocalGitRepo {
 		})
 	}
 
+	return result
+}
+
+func (s *tiltfileState) reposToDomain(image *dockerImage) []model.LocalGitRepo {
```
maiamcc: this function name/signature doesn't make much sense to begin with, less so now that we may be handling multiple types of manifests. maybe `reposForImage`?

internal/tiltfile/docker.go
```diff
@@ -308,14 +308,14 @@ func (s *tiltfileState) mountsToDomain(image *dockerImage) []model.Mount {
 	return result
 }
 
-func (s *tiltfileState) reposToDomain(image *dockerImage) []model.LocalGitRepo {
+func pathsToRepos(paths []localPath) []model.LocalGitRepo {
```
maiamcc: fwiw this func name confused me -- seems like we're converting the paths to repos? consider something like `reposForPaths` (ugh, or something better, don't love that either) and/or a docstring.

internal/tiltfile/docker.go
```diff
@@ -325,34 +325,40 @@ func (s *tiltfileState) reposToDomain(image *dockerImage) []model.LocalGitRepo {
...
+		image.staticDockerfilePath,
+		image.staticBuildPath,
+		image.tiltfilePath)
 
-	return result
+	return pathsToRepos(paths)
 }
 
-func (s *tiltfileState) dockerignoresToDomain(image *dockerImage) []model.Dockerignore {
+func pathsToDockerIgnores(paths []string) []model.Dockerignore {
```
maiamcc: same comment as `pathsToRepos`

also, nit, your capitalization of `dockerIgnore` is inconsistent.

internal/tiltfile/docker_compose.go
landism: whoops, yeah, meant to change this. thanks

internal/tiltfile/tiltfile_docker_compose_test.go
landism: nope! removed

internal/tiltfile/docker.go
landism: yeah, don't love that either, but don't see a better option (and I wasn't a big fan of the existing name before this change, either)

landism:
> more importantly, you should search a dc manifest's mounts for gitignore/dockerignore as well.

I figured that'd be a separate PR as part of https://app.clubhouse.io/windmill/story/1284/don-t-watch-mounted-files, or am I misunderstanding?

internal/tiltfile/tiltfile_docker_compose_test.go
landism: whoops, it turns out it does:

if you try to ignore a file that's not mounted, the ignore test helper will fail:
```
    tiltfile_test.go:927: Path /private/var/folders/0t/c8t9m67j7qs4y021nl5_3_yr0000gp/T/TestDockerComposeHonorsGitIgnore075810220/foo/tmp is not a dependency of manifest foo
```

PR #981, opened 2019-01-10
docs: commit rebuilt docs

landism:
just ran `make docs` on master and getting the built docs in sync with the source, so that diffs in other PRs reflect the actual source changes in those PRs (at least, until we make push changes without `make docs`)


nicks:
uhhh...seems fine but it sounds like your docker just has a different version of sphinx cached, and it will go back the other way when someone runs it with a fresh container.

landism:
I was surprised that we even build release docs and binaries on devs' laptops....another option would be to have CI do this, and then we can't forget to do it, and this kind of thing stays out of PRs...anyway, just gonna pull this diff into #979 and close this

PR #990, opened 2019-01-11
tilt: programmatize release process a bit more

landism:
Having gotten this far, I'm pretty sure it wasn't worth the time, but here we are.

**Problem**
When doing a release yesterday, I noticed a few things:
1. the instructions missed one of the docs that needed to be updated
2. we have a "-dev" suffix even on official releases
3. the instructions aren't clear that the dev version in build.go should be higher than the release version
4. there were more manual, error-prone steps than I liked

(maybe 2 and 3 are intentional and I'm just seeing things through my own biases?)

**Solution**
1. generally programmatize things and reduce room for manual error.
2. drop "-dev" suffix from versions built by goreleaser
3. more clearly communicate that dev version should be higher than release version


.goreleaser.yml
```diff
@@ -4,6 +4,8 @@ builds:
   goos:
   - linux
   - darwin
+  ldflags:
+  - -X=github.com/windmilleng/tilt/internal/cli.version={{ .Version }}
```
nicks: you're aware that goreleaser already does this automatically, right?

it sets this variable
https://github.com/windmilleng/tilt/blob/master/cmd/tilt/main.go#L11

the whole point of the internal/cli variable you're setting here is because we see lots of people build from source with `go get -u`.

This change defeats the whole point of that variable, because now it's just equivalent to the other variable.

Makefile
```diff
@@ -17,7 +17,7 @@ proto: scripts/protocc/protocc.py
 # TODO(nick): We should have a release build that bakes in a particular
 # SYNCLET_IMAGE tag.
 install:
-	./hide_tbd_warning go install ./...
+	./hide_tbd_warning go install -ldflags "-X 'github.com/windmilleng/tilt/internal/cli.version=$$(<dev_version)-dev'" ./...
```
nicks: again, i think this line misunderstands the problem that this variable is trying to solve. it's for people that don't use our Makefile

Makefile
```diff
@@ -114,7 +114,7 @@ docs:
 	docker rm tiltdocs || exit 0
 	rm -fR docs/_build
 	docker build -t tilt/docs -f Dockerfile.docs .
-	docker run --name tiltdocs tilt/docs
+	docker run -e "TILT_VERSION=$$(<release_version)" -e "TILT_RELEASE_DATE=$$(<release_date)" --name tiltdocs tilt/docs
```
nicks: this is cool, but it would be nice if it interoperated with the DATE/VERSION management that goreleaser is already doing. i'm not a big fan of having a duplicate, parallel system for managing release DATE + VERSION.

landism:
docs aren't even in this repo any more

PR #994, opened 2019-01-11
tiltfile: helm builtin

jazzdan:


.circleci/Dockerfile
```diff
@@ -5,3 +5,10 @@ RUN go get -u github.com/kisielk/errcheck
 RUN go get -u github.com/google/wire/cmd/wire
 
 RUN go get -u sigs.k8s.io/kustomize
+
+USER root
```
jazzdan: is this a cool thing to do?

internal/tiltfile/tiltfile_state.go
```diff
@@ -15,6 +15,12 @@ import (
 	"github.com/windmilleng/tilt/internal/model"
 )
 
+// TODO(dmiller) this needs to add all of the files in the directory as dependencies
```
jazzdan: Should we have a `add_dir_as_dependencies` function?  Or is this a sign that we shouldn't be trying to implement this in Starlark, and should implement it in Go so that we can walk that chart directory and record dependencies?

.circleci/Dockerfile
nicks: ya

jazzdan:
@dbentley will do after lunch. In the meantime [here](https://github.com/windmilleng/tilt/blob/70b7155617692fc8c5c3370deef4d6978aef69db/internal/tiltfile/tiltfile_test.go#L712-L715) is a test if it helps.

internal/tiltfile/tiltfile_state.go
dbentley: I think we need to have lots of thoughts about dependencies, and shouldn't block this change on that.

Are you sure that's the right way to get deps from helm? I don't know enough about it to know how else it could have deps. What does skaffold do?

internal/tiltfile/tiltfile_state.go
```diff
@@ -15,6 +15,12 @@ import (
...
+const helmFunc = `
+def helm(path):
+	return local("helm template " + path)
```
nicks: do we also need a way to mark the path as read, so it registers as a config file?

internal/tiltfile/tiltfile_state.go
```diff
@@ -82,25 +89,36 @@ const (
...
-	add := func(name string, fn func(thread *starlark.Thread, fn *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error)) {
-		r[name] = starlark.NewBuiltin(name, fn)
-	}
 
-	add(dockerComposeN, s.dockerCompose)
+	r = addBuiltin(r, localN, s.local)
+	r = addBuiltin(r, readFileN, s.skylarkReadFile)
+
+	// TODO(dmiller) this really only needs to be done once, on startup
+	r, err := starlark.ExecFile(thread, "helm.builtin", helmFunc, r)
```
nicks: can we use a dummy thread here? i don't see a good reason to pass in an external thread, do you?

nicks: also probably don't need to re-exec this every time we exec a new tiltfile

internal/tiltfile/tiltfile_test.go
```diff
@@ -702,6 +702,23 @@ if True:
...
 
+// TODO(dmiller): test dependencies
+func TestHelm(t *testing.T) {
+	f := newFixture(t)
+	defer f.TearDown()
+
+	f.setupHelm()
+
+	f.file("Tiltfile", `
+yml = helm('helm')
```
nicks: can you also add a test that this works with git_repo('.').path('helm')?

internal/tiltfile/tiltfile_state.go
jazzdan: Yeah see TODO, and @dbentley's comment.

jazzdan:
@dbentley added docs

docs/write_your_tiltfile.md
```diff
@@ -18,6 +18,9 @@ k8s_yaml(['foo.yaml', 'bar.yaml'])
 # run kustomize to generate yaml
 k8s_yaml(kustomize('config_dir'))
 
+# run helm to generate yaml
+k8s_yaml(helm('chart_dir'))
+
 # run a custom command to generate yaml
```
dbentley: Can we move local (the general case) above the specific cases (helm and kustomize), and then maybe a comment like "we also handle common tools to generate yaml"?

internal/tiltfile/tiltfile.go
```diff
@@ -34,9 +34,6 @@ func Load(ctx context.Context, filename string, matching map[string]bool) (manif
 	tfRoot, _ := filepath.Split(absFilename)
 
 	s := newTiltfileState(ctx, absFilename, tfRoot)
-	defer func() {
```
jazzdan: this doesn't seem to do anything, but maybe it has something to do with named returns that I'm not understanding?

jazzdan:
fyi @nicks I replaced the python implementation with a Go implementation to make tracking dependencies easier. It was a good experiment and we'll need to support things like this in the future for when people write Tilt libraries but I decided it wasn't worth the effort for now.

internal/tiltfile/tiltfile.go
nicks: seems to work ok for me
https://play.golang.org/p/N2e7s_wsABh

can you give a specific example?

.circleci/config.yml
```diff
@@ -64,8 +64,11 @@ jobs:
       - save_cache:
           paths:
             - /usr/local/Homebrew
-          key: homebrew_cache_v7
-      - run: echo 'export PATH="/usr/local/opt/go@1.10/bin:$PATH"' >> $BASH_ENV
+          key: homebrew_cache_v8
+      - run: echo 'export PATH="/usr/local/opt/go@1.11/bin:$PATH"' >> $BASH_ENV
+      - run: curl --silent --show-error --location --fail --retry 3 --output /tmp/helm.tar.gz  https://storage.googleapis.com/kubernetes-helm/helm-v2.12.1-darwin-amd64.tar.gz &&
```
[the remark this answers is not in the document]
jazzdan: Yes, can do. Basically it looks like homebrew isn't being updated any more, it has a much older verison of helm that actually causes the tests to fail in a minor way (just a capitalization difference).

internal/tiltfile/tiltfile.go
nicks: if there's an error, i think this function needs to return non-nil? i can write a test for it to double-check the behavior if you want.

internal/tiltfile/files.go
```diff
@@ -285,3 +285,43 @@ func (s *tiltfileState) kustomize(thread *starlark.Thread, fn *starlark.Builtin,
...
+
+	cmd := fmt.Sprintf("helm template %s", path)
+	yaml, err := s.execLocalCmd(cmd)
+	if err != nil {
+		return nil, err
+	}
+
+	deps := []string{}
+
+	err = filepath.Walk(localPath.path, func(path string, info os.FileInfo, err error) error {
```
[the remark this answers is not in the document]
jazzdan: Yep this is a good point, let's just add the whole directory as a "configFile". Seems to work just fine.

internal/tiltfile/tiltfile.go
jazzdan: Ohhh I think I get it now. Okay I'll add it back.

PR #996, opened 2019-01-11
DC: ignore (most?) changes in mounted directories

landism:
It's perhaps worth drawing particular attention to [this](https://github.com/windmilleng/tilt/compare/mlandis/ch1284/don-t-watch-mounted-files?expand=1#diff-4366126544582e719708ea21ba2216f3R82). The docs seemed a bit unclear for those cases and it seemed hairy enough that I erred on the side of "wait until someone asks for it". Most of the docker-compose.yml files I looked at are using the simple case that's now supported.


internal/tiltfile/docker_compose.go
```diff
@@ -60,6 +60,38 @@ type dcConfig struct {
...
+
+type Volumes struct {
+	Volumes []Volume
+}
+
+type Volume struct {
+	Source string
+}
+
+func (v *Volumes) UnmarshalYAML(unmarshal func(interface{}) error) error {
```
nicks: not excited about implementing our own marshalling logic here. i think there should be a library for this? but i haven't spent much time looking into how to leverage this

https://github.com/docker/cli/tree/master/cli/compose

internal/engine/watchmanager.go
```diff
@@ -19,6 +19,7 @@ type WatchableManifest interface {
 	ManifestName() model.ManifestName
 	LocalRepos() []model.LocalGitRepo
 	Dockerignores() []model.Dockerignore
+	IgnoredLocalDirectories() []string
```
maiamcc: comment?

internal/ignore/path_matcher.go
```diff
@@ -53,6 +53,7 @@ func CreateBuildContextFilter(m repoManifest) model.PathMatcher {
 type IgnorableManifest interface {
```
maiamcc: hmmm... this seems redundant with `WatchableManifest`. Probably WM should implement IM

internal/ignore/path_matcher.go
```diff
@@ -119,3 +123,12 @@ func (m tempBrokenSymlinkMatcher) Matches(path string, isDir bool) (bool, error)
 
 	return ospath.IsBrokenSymlink(path)
 }
+
+type directoryMatcher struct {
+	dir string
+}
+
+func (d directoryMatcher) Matches(p string, isDir bool) (bool, error) {
```
maiamcc: dumb Q: will this match itself?

internal/model/deploy_info.go
```diff
@@ -17,9 +17,10 @@ type DockerComposeTarget struct {
 	// TODO(nick): It might eventually make sense to represent
 	// Tiltfile as a separate nodes in the build graph, rather
 	// than duplicating it in each DockerComposeTarget.
-	tiltFilename  string
-	dockerignores []Dockerignore
-	repos         []LocalGitRepo
+	tiltFilename            string
+	dockerignores           []Dockerignore
+	repos                   []LocalGitRepo
+	ignoredLocalDirectories []string
```
maiamcc: comment

internal/tiltfile/docker_compose.go
```diff
@@ -102,9 +134,10 @@ type dcBuildConfig struct {
 
 // A docker-compose service, according to Tilt.
 type dcService struct {
-	Name    string
-	Context string
-	DfPath  string
+	Name             string
+	Context          string
+	DfPath           string
+	MountedLocalDirs []string
```
maiamcc: comment to differentiate from tilt `mount`?

internal/tiltfile/tiltfile_docker_compose_test.go
```diff
@@ -242,6 +260,28 @@ RUN echo hi`
...
+RUN echo hi`
+	f.file("foo/Dockerfile", df)
+
+	f.file("docker-compose.yml", configWithMounts)
+	f.file("Tiltfile", "docker_compose('docker-compose.yml')")
+
+	f.load("foo")
+
+	f.assertManifest("foo",
+		buildMatches("foo/Dockerfile"),
```
maiamcc: worth a note about why we want this to NOT be ignored?

internal/tiltfile/docker_compose.go
landism: going to tackle this in a separate PR

internal/ignore/path_matcher.go
landism: Maybe? I kinda assumed they were added by the same person with some intent in mind, but I didn't spend time thinking about it since that seems orthogonal to this PR (though this PR slightly exacerbates any potential redundancy).

internal/ignore/path_matcher.go
landism: It turns out that yes it will, but I consider that a bug in `ospath.Child` since a directory is pretty definitely not a child of itself.

Either way, I don't think whether it will or not is particularly important for this feature.

PR #1000, opened 2019-01-14
build: fix flaky tests that rely on time.Time

jazzdan:


jazzdan:
@nicks the pipeline printer tests. We subtract two `time.Now()` and sometimes the result is 0.00 seconds, sometimes it 0.01 seconds

PR #1006, opened 2019-01-14
tiltfile: add dockerfile_contents parameter

jazzdan:


jazzdan:
whoops, I need to update the docs.

internal/tiltfile/docker.go
```diff
@@ -54,12 +55,35 @@ func (s *tiltfileState) dockerBuild(thread *starlark.Thread, fn *starlark.Builti
...
+	shouldReadDF := false
+	switch v := dockerfileVal.(type) {
+	case *blob:
+		dockerfileContents = v.text
+	case localPath:
+		dockerfilePath = v
+		shouldReadDF = true
+	case starlark.String:
+		absPath := s.absPath(v.GoString())
+		if _, err := os.Stat(absPath); os.IsNotExist(err) {
```
nicks: i'm not convinced this is a good idea

this reminds me of that old jquery function `load()` that either fetched a url or added a 'load' event handler based on what the arguments looked like

people hated it and they eventually removed it

i'm worried that this has the same problems. If you typo a filename, or accidentally remove a file, suddenly your Tiltfile is trying to do something very different

possible alternative options:
- emit an error if you try to pass in a string with newlines, with instructions on how to fix the issue. like: "This doesn't look like a filename. Did you intend to use make_blob()?"
- different named arguments like dockerfile='Dockerfile' vs build='FROM golang'

jazzdan: I'm happy to make it a separate parameter, and in fact already have such an implementation ready to go. :) Let me add that back to the PR.

jazzdan:
updated @nicks

PR #1014, opened 2019-01-15
docs: add simple Helm docs under 'configs from other tools'

jazzdan:


docs/helm.md
```diff
@@ -0,0 +1,17 @@
+# Helm Support in Tilt
```
nicks: "in Tilt" seems redundant to me

maybe "For Helm Users"?

docs/helm.md
```diff
@@ -0,0 +1,17 @@
...
+k8s_yaml(yml)
+```
+
+This could also be implemented as a `local` command if you require further customization:
+
+```python
+local("helm template path/to/chart")
+```
+
+For complete documentation of both `tilt` and `local` visit the [Tilt API reference](api.html).
```
nicks: is this supposed to read "For complete documentation of both `helm` and `local`"?

docs/helm.md
```diff
@@ -0,0 +1,17 @@
+# Helm Support in Tilt
+
+Tilt has native support for Helm! To run `helm template` on a chart directory from Tilt simply pass the chart directory to the `helm` built-in like so:
+
+```python
+yml = helm('path/to/chart')
+# The result is just YAML, which you can use throughout Tilt anywhere you would normally use YAML
+k8s_yaml(yml)
```
nicks: maybe add a section on how to add a Dockerfile to substitute in your own build process?

this could largely be copy-and-pasted from another part of the guide

PR #1022, opened 2019-01-16
model: add setter for image target

nicks:
Hello @nicks,

Please review the following commits I made in branch nicks/imagetarget:

50c053f0249429a7bcc78de9b8fd1afb4abeb655 (2019-01-15 21:13:24 -0500)
model: add setter/getter for image target


nicks:
just modified this to be a setter-only. this should be non-controversial

PR #1024, opened 2019-01-16
store: track pending files per-image instead of per-manifest

nicks:
Hello @landism, @jazzdan,

Please review the following commits I made in branch nicks/pending:

fe4b3290ebb93ef151418793b77d16b3efe93804 (2019-01-16 10:40:17 -0500)
store: track pending files per-image instead of per-manifest
in preparation for multiple images per manifest

### DIFF
--- a/internal/store/build_result.go
+++ b/internal/store/build_result.go
@@ -53,6 +53,15 @@ func (b BuildResult) ShallowCloneForContainerUpdate(filesReplacedSet map[string]
 
 type BuildResultSet map[model.TargetID]BuildResult
 
+func (set BuildResultSet) AsOneResult() BuildResult {
+	if len(set) == 1 {
+		for _, result := range set {
+			return result
+		}
+	}
+	return BuildResult{}
+}
+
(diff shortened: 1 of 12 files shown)


internal/store/build_result.go
```diff
@@ -53,6 +53,15 @@ func (b BuildResult) ShallowCloneForContainerUpdate(filesReplacedSet map[string]
 
 type BuildResultSet map[model.TargetID]BuildResult
 
+func (set BuildResultSet) AsOneResult() BuildResult {
+	if len(set) == 1 {
+		for _, result := range set {
```
jazzdan: Can't this for loop be statically unrolled since it's guaranteed to be iterating over a collection with only one thing in it?

jazzdan: I'm confused about what this code is doing.

nicks: ya, this just means "get the first item from the map"

jazzdan: ohhhhh okay, got it

PR #1027, opened 2019-01-16
dockercompose: use a channel to more accurately simulate streaming output

nicks:
Hello @maiamcc,

Please review the following commits I made in branch nicks/flakey:

becc6dde8f702853d2de7f7f5c3dfed3e1a3aac8 (2019-01-16 13:21:41 -0500)
dockercompose: use a channel to more accurately simulate streaming output

### DIFF
--- a/internal/engine/upper_test.go
+++ b/internal/engine/upper_test.go
@@ -1815,7 +1815,10 @@ func TestDockerComposeRecordsRunLogs(t *testing.T) {
 	f := newTestFixture(t)
 	m, _ := f.setupDCFixture()
 	expected := "hello world"
-	f.setDCRunLogOutput(m.DockerComposeTarget(), expected)
+	output := make(chan string, 1)
+	output <- expected
+	defer close(output)
+	f.setDCRunLogOutput(m.DockerComposeTarget(), output)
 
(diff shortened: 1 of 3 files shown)


internal/engine/upper_test.go
```diff
@@ -1815,7 +1815,10 @@ func TestDockerComposeRecordsRunLogs(t *testing.T) {
 	f := newTestFixture(t)
 	m, _ := f.setupDCFixture()
 	expected := "hello world"
-	f.setDCRunLogOutput(m.DockerComposeTarget(), expected)
+	output := make(chan string, 1)
+	output <- expected
+	defer close(output)
+	f.setDCRunLogOutput(m.DockerComposeTarget(), output)
```
maiamcc: yay so much better thank you! could all this happen in `setDCRunLogOutput` tho? (i.e. initialize the channel if it doesn't exist, and send the output to the channel)

nicks: ya, i went down that path initially, but it making it thread-safe and closing the channel properly got complex

nicks: (creating the channel outside the method lets the caller use `defer`)

PR #1035, opened 2019-01-17
docs: fix confusing docstring for docker_build

maiamcc:
Hello @jazzdan, @nicks,

Please review the following commits I made in branch maiamcc/ref-examples:

0dfca6dc618c62758c9130bd0cda9f46ff86247d (2019-01-17 11:48:30 -0500)
docs: fix confusing docstring for docker_build

screen shot 2019-01-17 at 11 48 18 am


docs/api/api.py
```diff
@@ -25,14 +25,16 @@ def local_git_repo(path: str) -> Repo:
...
   Args:
-    ref: e.g. a blorgdev/backend or gcr.io/project-name/bucket-name
-    context: The path to use as the Docker build context.
-    build_args: the build-time variables that are accessed like regular environment variables in the ``RUN`` instruction of the Dockerfile. See `the Docker Build Arg documentation <https://docs.docker.com/engine/reference/commandline/build/#set-build-time-variables---build-arg>`_
-    dockerfile: The path to a Dockerfile
-    dockerfile_contents: the raw contents of the Dockerfile to use for this build
+    ref: name for this image (e.g. 'myproj/backend' or 'myregistry/myproj/backend'). If this image will be used in a k8s resource(s), this ref must match the ``spec.container.image`` param for that resource(s).
+    context: path to use as the Docker build context.
+    build_args: build-time variables that are accessed like regular environment variables in the ``RUN`` instruction of the Dockerfile. See `the Docker Build Arg documentation <https://docs.docker.com/engine/reference/commandline/build/#set-build-time-variables---build-arg>`_
+    dockerfile: path to the Dockerfile to build (may be absolute, or relative to the build context)
```
nicks: i thought this was relative to the Tiltfile, not relative to the build context

maiamcc: Ugh you're right, good catch.

maiamcc: Hmmm maybe we want to rethink this? I'm think of both the docker cli and Docker Compose, which assume the Dockerfile is relative to the build context...

jazzdan: I don't think the docker cli is at least:

```
> docker build go/src/github.com/windmilleng/servantes/snack/ -f Dockerfile
unable to prepare context: unable to evaluate symlinks in Dockerfile path: lstat /Users/dan/Dockerfile: no such file or directory
```

maiamcc: 🤔 huh okay! (what if you don't pass -f tho?)

nicks: this isn't true? I just tried:

```
docker build -f Dockerfile.docserver docs
```

definitely uses the Dockerfile in the current working directory, not relative to docs.

jazzdan: then yes, it will look for `Dockerfile` in the root of the build context

maiamcc: yeah if you pass `-f` it assumes the arg is relative to cwd but if you don't it looks for Dockerfile in the context... hmmm...

jazzdan: Yep! And that's by design and documented behavior

PR #1038, opened 2019-01-17
engine: get status of dc resources on startup

jazzdan:


internal/engine/upper.go
```diff
@@ -286,6 +287,19 @@ func handleBuildCompleted(ctx context.Context, engineState *store.EngineState, c
...
 	}
 
+	if mt.Manifest.IsDC() {
+		state, _ := ms.ResourceState.(dockercompose.State)
+
+		cid := cb.Result.AsOneResult().ContainerID
+		spew.Dump(cb.Result)
+		if cid != "" {
+			state = state.WithContainerID(cid)
+			state = state.WithStatus(dockercompose.StatusUp)
```
maiamcc: 🤔 but how do we know that the container is up?

internal/engine/upper.go
```diff
@@ -286,6 +287,19 @@ func handleBuildCompleted(ctx context.Context, engineState *store.EngineState, c
 		}
 	}
 
+	if mt.Manifest.IsDC() {
+		state, _ := ms.ResourceState.(dockercompose.State)
+
+		cid := cb.Result.AsOneResult().ContainerID
+		spew.Dump(cb.Result)
```
maiamcc: 👀

internal/engine/docker_compose_build_and_deployer.go
```diff
@@ -23,15 +28,15 @@ func NewDockerComposeBuildAndDeployer(dcc dockercompose.DockerComposeClient) *Do
...
+func (bd *DockerComposeBuildAndDeployer) extract(specs []model.TargetSpec) []targetTuple {
+	result := []targetTuple{}
 	for _, s := range specs {
 		dc, isDC := s.(model.DockerComposeTarget)
 		if !isDC {
 
 			return nil
 		}
-		result = append(result, dc)
+		result = append(result, targetTuple{s.ID(), dc})
```
maiamcc: ...doesn't a DockerComposeTarget (necessarily) implement `.ID()`? If so, this tuple situation is unnecessary.

internal/engine/upper.go
maiamcc: good comment! but whether or not we have a cid doesn't actually tell us whether it's the first build, so the logic is misleading. what do you think of:
> If we have a container ID and no status yet, set status to Up (this is an expected case when we run `docker-compose up` and the service is already running, and we won't get an event to tell us so). If the container is crashing we will get an event subsequently.

maiamcc: should also attach the containerId to the state.

internal/engine/upper_test.go
```diff
@@ -1900,6 +1900,26 @@ func TestDockerComposeDetectsCrashes(t *testing.T) {
 	})
 }
 
+func TestDockerComposeBuildCompletedSetsStatusToUp(t *testing.T) {
+	f := newTestFixture(t)
+	m1, _ := f.setupDCFixture()
+
+	expected := container.ID("aaaaaa")
```
maiamcc: ...do we care that it DOESN'T set status if it doesn't get a containerID back? idk if it's worth a test, but maybe...?

internal/engine/upper.go
```diff
@@ -286,6 +286,25 @@ func handleBuildCompleted(ctx context.Context, engineState *store.EngineState, c
 		}
 	}
 
+	if mt.Manifest.IsDC() {
+		state, _ := ms.ResourceState.(dockercompose.State)
+		logger.Get(ctx).Infof("dc manifest state: %+v", state)
```
maiamcc: 👀

PR #1043, opened 2019-01-17
filewatcher: pick up changes to manifest ignores

landism:


internal/engine/upper_test.go
```diff
@@ -1918,7 +1918,7 @@ func (f fakeTimerMaker) maker() timerMaker {
 		default:
 			// if you hit this, someone (you!?) might have added a new timer with a new duration, and you probably
 			// want to add a case above
-			f.t.Error("makeTimer called on unsupported duration")
+			panic("makeTimer called on unsupported duration")
```
maiamcc: why?

internal/engine/watchmanager.go
```diff
@@ -128,19 +130,23 @@ func (w *WatchManager) diff(ctx context.Context, st store.RStore) (setup []Watch
...
+			gitignores = append(gitignores, r.GitignoreContents)
 		}
+		return gitPaths, gitignores
 	}
 
-	return true
+	gitPaths1, gitignores1 := watchableTargetReposAsArrays(w1)
+	gitPaths2, gitignores2 := watchableTargetReposAsArrays(w2)
+
+	return sliceutils.StringSliceEquals(w1.Dependencies(), w2.Dependencies()) &&
```
maiamcc: what about `func ignoresMatch(i1, i2 IgnorableTarget) bool {...}` and then
```
func watchesMatch(w1, w2 WatchableTarget) bool {
    return sliceutils.StringSliceEquals(w1.Dependencies(), w2.Dependencies()) &&
        ignoresMatch(i1, i2)
}
```
? then you can implement ignore-specific methods right on `IgnorableTarget` and don't need that awkward note above `WatchableTarget`

internal/sliceutils/sliceutils.go
```diff
@@ -15,3 +15,18 @@ func DedupedAndSorted(slice []string) (result []string) {
 	sort.Strings(result)
 	return result
 }
+
+func StringSliceEquals(a, b []string) bool {
```
maiamcc: nit but this should probably compare elements regardless of order?

internal/engine/watchmanager_test.go
```diff
@@ -0,0 +1,148 @@
...
+}
+
+func newWMFixture(t *testing.T) *wmFixture {
+	f := tempdir.NewTempDirFixture(t)
+	actions := make(chan store.Action, 20)
+	reducer := store.Reducer(func(ctx context.Context, s *store.EngineState, action store.Action) {
+		select {
+		case actions <- action:
+		case <-time.After(100 * time.Millisecond):
+			panic("would block when writing to action chan. perhaps too many actions or too small a buffer")
```
maiamcc: again, why panic rather than `t.Fatal`?

internal/engine/watchmanager_test.go
```diff
@@ -0,0 +1,148 @@
...
+		}
+	})
+	st := store.NewStore(reducer, false)
+	timerMaker := makeFakeTimerMaker(t)
+	fakeMultiWatcher := newFakeMultiWatcher()
+	wm := NewWatchManager(fakeMultiWatcher.newSub, timerMaker.maker())
+
+	ctx, cancel := context.WithCancel(context.Background())
```
maiamcc: can cut the lil bit of logger setup below if instead of `context.Background()` you use `output.CtxForTest()`

internal/engine/watchmanager_test.go
```diff
@@ -0,0 +1,148 @@
...
+func (f *wmFixture) ReadActionsUntil(lastFile string) ([]targetFilesChangedAction, error) {
+	var actions []targetFilesChangedAction
+	var done = false
+	for !done {
+		var tfca targetFilesChangedAction
+		select {
+		case a := <-f.actions:
+			tfca = a.(targetFilesChangedAction)
+
+			// 1. unpack to one file per action, for deterministic inspection
```
maiamcc: comment above that `lastFile` should be relative to cwd (or also call  `ospath.Child` on it)

internal/engine/watchmanager_test.go
```diff
@@ -0,0 +1,148 @@
...
+			// 2. make paths relative to cwd
+			for _, absPath := range tfca.files {
+				p := absPath
+				if relPath, ok := ospath.Child(f.Path(), absPath); ok {
+					p = relPath
+				}
+				actions = append(actions, targetFilesChangedAction{tfca.targetID, []string{p}})
+				if p == lastFile {
+					done = true
```
maiamcc: why not just `return actions, nil` from here?

internal/engine/watchmanager_test.go
```diff
@@ -0,0 +1,148 @@
...
+		paths = append(paths, a.files[0])
+	}
+
+	// not asserting exact contents because we can end up with duplicates since the old watch loop isn't stopped
+	// until after the new watch loop is started
+	assert.Contains(t, paths, "bar")
+}
+
+type wmFixture struct {
+	*tempdir.TempDirFixture
```
maiamcc: any particular reason for this to embed TempDirFixture? Seems like we could get rid of all the `f.Join` calls and just have all the path matching be absolute?

internal/engine/watchmanager_test.go
```diff
@@ -0,0 +1,148 @@
...
+		state := f.store.LockMutableStateForTesting()
+		state.UpsertManifestTarget(&mt)
+		state.WatchMounts = true
+		f.store.UnlockMutableState()
+		f.wm.OnChange(f.ctx, f.store)
+	}
+
+	target := model.DockerComposeTarget{Name: "foo"}
+	target.Mounts = []model.Mount{{LocalPath: f.Path()}}
+	setManifestTarget(target.WithIgnoredLocalDirectories([]string{f.JoinPath("bar")}))
```
maiamcc: to be really thorough you'd probably want to test here that "bar" DOES get ignored?

internal/engine/watchmanager_test.go
```diff
@@ -0,0 +1,148 @@
...
+
+	setManifestTarget(target)
+
+	f.ChangeFile(t, "bar")
+	f.ChangeFile(t, "stop")
+
+	actions, err := f.ReadActionsUntil("stop")
+	if !assert.NoError(t, err) {
```
maiamcc: nit: takes up the same number of lines as `if err != nil { f.t.Fatal(err) }` and imo the latter is easier to read, but 🤷‍♀️

internal/engine/watchmanager_test.go
```diff
@@ -0,0 +1,148 @@
...
+		return
+	}
+
+	var paths []string
+	for _, a := range actions {
+		if !assert.Equal(t, 1, len(a.files)) {
```
maiamcc: same nit as above

maiamcc: also this means something is weird with the test fixture, doesn't it? maybe note that in the error message.

internal/engine/upper_test.go
maiamcc: uhh what i mean is: any particular reason to change this?

landism: oops, that didn't belong in this PR

(I made this change while debugging, and remembered having issues with calls on `testing.T` not working from goroutines, but I think that's only calls to `Fatal` that are problematic)

internal/engine/watchmanager.go
landism: I don't think that affects whether that awkward note is needed - if we add a field to `WatchableTarget`, we still need to update `watchesMatch` (and, likewise, if we add a field to `IgnorableTarget`, we need to update `ignoresMatch`).

I don't see a compelling argument one way or the other for whether ignoresMatch should be broken out. It feels kinda premature if it's not being called by anything else.

internal/engine/watchmanager_test.go
landism: I like the "Received unexpected error" text that `NoError` adds to the output

internal/engine/watchmanager.go
maiamcc: fair nough

internal/engine/watchmanager_test.go
maiamcc: then this should maybe also be a `t.Fatal`?

internal/engine/watchmanager_test.go
```diff
@@ -0,0 +1,148 @@
...
+	fakeMultiWatcher := newFakeMultiWatcher()
+	wm := NewWatchManager(fakeMultiWatcher.newSub, timerMaker.maker())
+
+	ctx, cancel := context.WithCancel(context.Background())
+	l := logger.NewLogger(logger.DebugLvl, os.Stdout)
+	ctx = logger.WithLogger(ctx, l)
+	go func() {
+		err := st.Loop(ctx)
+		if err != nil && err != context.Canceled {
+			panic(fmt.Sprintf("st.Loop exited with error: %+v", err))
```
maiamcc: `t.Fatal`?

internal/engine/watchmanager_test.go
landism: to be really thorough, that test should have been on the last PR and this file should also test all the other stuff in watchmanager :)

internal/engine/watchmanager_test.go
landism: not sure what precedes the "then" in this comment, but this runs in a goroutine, so `t.Fatal` won't work

internal/engine/watchmanager_test.go
landism: ditto

internal/sliceutils/sliceutils.go
maiamcc: yeah good point, nit retracted!

internal/engine/watchmanager.go
```diff
@@ -128,14 +129,43 @@ func (w *WatchManager) diff(ctx context.Context, st store.RStore) (setup []Watch
 	return setup, teardown
 }
 
-func dependenciesMatch(d1 []string, d2 []string) bool {
-	if len(d1) != len(d2) {
+func watchRulesMatch(w1, w2 WatchableTarget) bool {
```
nicks: have you considered using go-cmp like we do in model?
https://github.com/google/go-cmp

internal/ignore/path_matcher.go
```diff
@@ -130,6 +138,19 @@ type directoryMatcher struct {
 	dir string
 }
 
+var _ model.PathMatcher = directoryMatcher{}
+
+func newDirectoryMatcher(dir string) (directoryMatcher, error) {
+	if !path.IsAbs(dir) {
+		wd, err := os.Getwd()
```
nicks: maybe use filepath.Abs? i think it does the same thing you're doing here

internal/ignore/path_matcher.go
```diff
@@ -2,9 +2,13 @@ package ignore
 
 import (
 	"context"
+	"os"
+	"path"
```
nicks: `path` is for URLs, you usually want `filepath` when dealing with files

PR #1047, opened 2019-01-18
Track images added by docker_build and fast_build by ref

dougsparling:
Hey, was having some trouble with Tilt 0.5.1 and decided to dig into this one myself. Basically Tilt was complaining that an image added by `docker_build` wasn't being used by any resource. In my case this wasn't right as I had a `k8s_resource` whose YAML file declared a container using that very same image.

So it turns out `docker_build` and `fast_build` were parsing the image attribute to get the _name_, while `k8s_resource` and friends were using the _ref_. Meanwhile, I was specifying a name plus tag in my `docker_build`, which caused Tilt's confusion.

This PR aims to reconcile the two by using refs everywhere. I could also see this going the other way for usability purposes by simply comparing names everywhere with no regard to tags. But then this can create an unpleasant situation where your manifests and published images might work fine in Tilt but not "in production" when you push the image, apply the manifest, etc.

Anyway, I don't think the PR is entirely done (I have some failing tests that I haven't dug into yet), but I was hoping it could be a platform for discussion so I know if this is going in the right direction, etc.

### DIFF
--- a/internal/tiltfile/docker.go
+++ b/internal/tiltfile/docker.go
@@ -101,7 +101,7 @@ func (s *tiltfileState) dockerBuild(thread *starlark.Thread, fn *starlark.Builti
 		dockerfileContents = string(bs)
 	}
 
-	if s.imagesByName[ref.Name()] != nil {
+	if s.imagesByRef[ref.String()] != nil {
 		return nil, fmt.Errorf("Image for ref %q has already been defined", ref.Name())
 	}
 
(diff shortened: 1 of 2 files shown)


nicks:
Thanks for the report! This is a good point. But I think we probably want to go in the other direction, and only match by name?

i.e., if you have serverA:tagX in the K8s yaml and server:tagY in the docker_build() call, they should match. What do you think?

For what it's worth, Tilt ignores all the tags anyway and injects it's own content-based tags based on the digest. So either matching on imageName or matching on imageName:tag is reasonable to implement; it's just about developer ergonomics.

dougsparling:
Sure, I'll try and get it working in the next few days!

nicks:
hi Doug! is this still something you wanted to contribute? Filing the issue is appreciated, and we can have one of the regular contributors take the PR the rest of the way if you no longer have time for it.

dougsparling:
Ah, sorry for the delay. Yup, this turned out to be a bit trickier than I had time to dive into this week. Please feel free to hand it off to somebody faster :)

PR #1048, opened 2019-01-18
engine: dcbad can build images [ch1371]

maiamcc:
tests to follow but want a check first that i'm on a reasonable track


internal/build/container_test.go
```diff
@@ -33,7 +33,7 @@ ADD dir/c.txt .
 	f.WriteFile("dir/c.txt", "c")
 	f.WriteFile("missing.txt", "missing")
 
-	ref, err := f.b.BuildDockerfile(f.ctx, f.ps, f.getNameFromTest(), df, f.Path(), model.EmptyMatcher, model.DockerBuildArgs{})
+	ref, err := f.b.BuildDockerfile(f.ctx, f.ps, f.getNameFromTest(), df, f.Path(), model.EmptyMatcher, model.DockerBuildArgs{}, false)
```
nicks: have you considered having a separate Tag() call?

i.e., docker tag fortune:tilt-cabcab fortune:latest

(i also don't think TagImage should be part of the normal BuildDockerfile call, but we can make that decision independently)

internal/build/image_builder.go
```diff
@@ -215,10 +217,17 @@ func (d *dockerImageBuilder) addRemainingSteps(df dockerfile.Dockerfile, remaini
 }
 
 // Tag the digest with the given name and wm-tilt tag.
-func (d *dockerImageBuilder) TagImage(ctx context.Context, ref reference.Named, dig digest.Digest) (reference.NamedTagged, error) {
-	tag, err := digestAsTag(dig)
-	if err != nil {
-		return nil, errors.Wrap(err, "TagImage")
+func (d *dockerImageBuilder) TagImage(ctx context.Context, ref reference.Named, dig digest.Digest, tagLatest bool) (reference.NamedTagged, error) {
```
nicks: maybe there should be two separate methods?
TagImage(reference.Named, tag sdig Digest)
TagImageFromDigest(reference.Named, dig Digest)

presumably, the second one could use the first one

internal/engine/docker_compose_build_and_deployer.go
```diff
@@ -12,58 +14,92 @@ import (
...
+
 	for _, s := range specs {
-		dc, isDC := s.(model.DockerComposeTarget)
-		if !isDC {
-			return nil
+		switch s := s.(type) {
+		case model.ImageTarget:
+			iTargets = append(iTargets, s)
+		case model.DockerComposeTarget:
+			dcTargets = append(dcTargets, s)
```
nicks: should probably return nil if there's a spec it doesn't recognize, rather than quietly dropping it on the floor. see the other extract() method

internal/engine/docker_compose_build_and_deployer.go
```diff
@@ -12,58 +14,92 @@ import (
...
-	brs := store.BuildResultSet{}
-	stdout := logger.Get(ctx).Writer(logger.InfoLvl)
-	stderr := logger.Get(ctx).Writer(logger.InfoLvl)
-	for _, dc := range dcs {
-		err := bd.dcc.Up(ctx, dc.ConfigPath, dc.Name, stdout, stderr)
-		if err != nil {
-			return store.BuildResultSet{}, err
-		}
+	if haveImage {
+		logger.Get(ctx).Infof("~~ i'm building ur image for: %s", dcTarget.Name)
```
nicks: O.o

internal/engine/docker_compose_build_and_deployer.go
```diff
@@ -12,58 +14,92 @@ import (
...
+		ps := build.NewPipelineState(ctx, 1, bd.ibd.clock)
+		defer func() { ps.End(ctx, err) }()
 
-		// NOTE(dmiller): right now we only need this the first time. In the future
-		// it might be worth it to move this somewhere else
-		cid, err := bd.dcc.ContainerID(ctx, dc.ConfigPath, dc.Name)
+		// NOTE(maia): we assume that this func takes one DC target and up to one image target
+		// corresponding to that service. If this func ever supports specs for more than one
+		// service at once, we'll have to match up image build results to DC target by ref.
+		ref, err = bd.ibd.build(ctx, iTarget, currentState[iTarget.ID()], ps, true)
```
nicks: maybe this is a good sign that this needs to be broken out into a common utility class? it seems very weird, architecturally, for one ImageBuildAndDeployer to container another.

internal/engine/docker_compose_build_and_deployer.go
maiamcc: 😅

internal/build/image_builder.go
maiamcc: no longer relevant after refactor

internal/engine/docker_compose_build_and_deployer.go
maiamcc: will do

maiamcc:
@nicks @jazzdan this is ready for y'all

internal/build/image_and_cache_builder.go
```diff
@@ -0,0 +1,175 @@
...
+
+func NewImageAndCacheBuilder(ib ImageBuilder, cb CacheBuilder, updateMode mode.UpdateMode) *ImageAndCacheBuilder {
+	return &ImageAndCacheBuilder{
+		ib:         ib,
+		cb:         cb,
+		updateMode: updateMode,
+	}
+}
+
+func (icb *ImageAndCacheBuilder) Build(ctx context.Context, iTarget model.ImageTarget, state store.BuildState, ps *PipelineState, canSkipPush bool) (reference.NamedTagged, error) {
```
nicks: should canSkipPush be a method of this struct? it seems like a weird thing to be a parameter

internal/dockercompose/client.go
```diff
@@ -33,8 +33,20 @@ func NewDockerComposeClient() DockerComposeClient {
...
-func (c *cmdDCClient) Up(ctx context.Context, configPath string, serviceName model.TargetName, stdout, stderr io.Writer) error {
-	cmd := exec.CommandContext(ctx, "docker-compose", "-f", configPath, "up", "--no-deps", "--build", "-d", serviceName.String())
+func (c *cmdDCClient) Up(ctx context.Context, configPath string, serviceName model.TargetName, shouldBuild bool, stdout, stderr io.Writer) error {
+	args := []string{"-f", configPath, "up", "--no-deps", "-d"}
+	if shouldBuild {
+		args = append(args, "--build")
+	} else {
+		// !shouldBuild implies that Tilt will take care of building, which implies that
+		// we should recreate container so that we pull the new image
+		// NOTE(maia): this is maybe the WRONG thing to do if we're deploying a service
```
nicks: for what it's worth, this is another reason why our k8s-based builder uses content-based tags: it's a way to make sure kubernetes creates a new pod iff the contents changed.

internal/build/image_and_cache_builder.go
```diff
@@ -0,0 +1,175 @@
+package build
```
nicks: fwiw, i probably would have left this in the same package...i feel like moving it to a different package forced you to make a bunch of odd architecture decisions around `mode`, which distracts from the main focus of this PR

internal/build/image_and_cache_builder.go
maiamcc: yeahhh we'd need to mess with the way we k8s env in and how it interacts with Docker Compose -- because we still populate this value even when we're doing Docker Compose stuff instead of worrying about k8s :-/ i'll leave a todo

internal/build/image_and_cache_builder.go
maiamcc: 😅 legit!

PR #1052, opened 2019-01-18
model: remove YAMLManifest

nicks:
Hello @jazzdan,

Please review the following commits I made in branch nicks/delete:

1bc13b190fa351cafacf55a7bd2b00acecd0f25a (2019-01-18 18:27:23 -0500)
model: remove YAMLManifest


internal/store/engine_state.go
```diff
@@ -172,7 +172,7 @@ func (e EngineState) RelativeTiltfilePath() (string, error) {
 }
 
 func (e EngineState) IsEmpty() bool {
-	return len(e.ManifestTargets) == 0 && e.GlobalYAML.Empty()
+	return len(e.ManifestTargets) == 0 && e.GlobalYAML.Name == ""
```
jazzdan: This feels like a weird isEmpty check, but I'm OK with it.

nicks: i agree! tho i think this will eventually go away as a separate field, so hopefully it will stop feeling weird soon!

PR #1066, opened 2019-01-25
tiltfile: store normalized image name on dc_resource for matching against imagesByName

maiamcc:
whoops try this one

### DIFF
--- a/internal/docker/image.go
+++ b/internal/docker/image.go
@@ -19,3 +19,19 @@ func ToImgNameAndTag(nt reference.NamedTagged) ImgNameAndTag {
 		Tag:  nt.Tag(),
 	}
 }
+
+func NormalizeRefName(ref string) (string, error) {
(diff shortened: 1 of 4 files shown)


internal/docker/image.go
```diff
@@ -19,3 +19,19 @@ func ToImgNameAndTag(nt reference.NamedTagged) ImgNameAndTag {
...
+
+func NormalizeRefName(ref string) (string, error) {
+	named, err := reference.ParseNormalizedNamed(ref)
+	if err != nil {
+		return "", err
+	}
+	return named.String(), nil
+}
+
+func MustNormalizeRefName(ref string) string {
```
jazzdan: This could live in the test package as it's only used there

maiamcc: that makes sense, though i'm inclined to leave it near the function it `Must`'s? 🤷‍♀️

PR #1067, opened 2019-01-25
tiltfile: speed up load for DC manifests [ch1201]

maiamcc:
loading dc manifests can take c. 0.8s b/c we were making two calls to
`docker-compose config` (one with `--services` to get services in
topologically sorted order). this pr removes the second call (with
`--services`) in favor of making sure to preserve service order when
we unmarshal the results of `docker-compose config`.


jazzdan:
What's the new run time?

maiamcc:
NEVER MIND, THIS PR IS WRONG 😢

PR #1078, opened 2019-01-30
k8s: if can't get kubecontext (e.g. no kubectl), return EnvNone instead of throwing error [ch1443]

maiamcc:
running w/o k8s/kubectl is now a possible use-case that we shouldn't crash
for (e.g. someone is running tilt x DC and doesn't need k8s at all).

### DIFF
--- a/internal/k8s/client.go
+++ b/internal/k8s/client.go
@@ -32,6 +32,7 @@ type ServiceName string
 type KubeContext string
 
 const DefaultNamespace = Namespace("default")
+const KubeContextNone = KubeContext("none") // stand-in for when k8s not running
 
 func (pID PodID) Empty() bool    { return pID.String() == "" }
 func (pID PodID) String() string { return string(pID) }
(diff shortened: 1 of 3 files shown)


internal/k8s/client.go
```diff
@@ -32,6 +32,7 @@ type ServiceName string
 type KubeContext string
 
 const DefaultNamespace = Namespace("default")
+const KubeContextNone = KubeContext("none") // stand-in for when k8s not running
```
nicks: KubeContext(EnvNone)?

internal/k8s/env.go
```diff
@@ -49,6 +52,8 @@ func EnvFromString(s string) Env {
 		return EnvDockerDesktop
 	} else if s == EnvMicroK8s {
 		return EnvMicroK8s
+	} else if s == string(KubeContextNone) {
```
nicks: EnvNone?

maiamcc: 😳 👍

PR #1085, opened 2019-01-31
k8s: don't try to get k8s rest api etc if no k8s configured [ch1433]

maiamcc:
Friend w/o kubectl was trying to run k8s and got error:

> Error: could not get config for context ({"" "" "" "" map[]}): invalid configuration: no configuration has been provided

This PR checks whether user has kubectl present at all before trying to get REST API/other k8s-relevant configs.

### DIFF
--- a/internal/k8s/client.go
+++ b/internal/k8s/client.go
@@ -35,7 +35,6 @@ type ServiceName string
 type KubeContext string
 
 const DefaultNamespace = Namespace("default")
-const KubeContextNone = KubeContext(EnvNone) // stand-in for when k8s not running
 
 func (pID PodID) Empty() bool    { return pID.String() == "" }
 func (pID PodID) String() string { return string(pID) }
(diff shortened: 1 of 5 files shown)


internal/k8s/client.go
```diff
@@ -107,13 +106,31 @@ var _ Client = K8sClient{}
 
 type PortForwarder func(ctx context.Context, restConfig *rest.Config, core apiv1.CoreV1Interface, namespace string, podID PodID, localPort int, remotePort int) (closer func(), err error)
 
+func ProvideK8sClient(ctx context.Context, env Env) (K8sClient, error) {
+	if env == EnvNone {
+		// No k8s, so no need to get any further configs
+		return K8sClient{env: env}, nil
```
nicks: should we have some sort of ExplodingK8sClient that emits proper errors? i'm concerned that this will just emit panics from null pointer dereferences

nicks: i.e., please make sure that you test the behavior where the user doesn't have `kubectl` but tries to use `k8s_yaml`, it seems like this will explode in that case

maiamcc:
ready for review. If you don't have `kubectl` error looks like this:
```$ tilt up
Error: Error watching pods. Are you connected to kubernetes?
: could not set up k8s client: DetectKubeContext failed: exec: "kubectl": executable file not found in $PATH```

PR #1086, opened 2019-01-31
k8s_resource: add extra_pod_labels param

landism:
a la https://github.com/windmilleng/tilt/issues/1069#issuecomment-459405658

or, hopefully the docs do a good job of describing the intent:
```
extra_pod_labels: In addition to relying on Tilt's heuristics to automatically find K8S resources associated with this resource, a user may specify extra labelsets to force entities to be associated with this resource. An entity will be associated with this resource if it has all of the labels in at least one of the entries in this list (but still also if it meets any of Tilt's usual mechanisms).
```


internal/model/label_pair.go
```diff
@@ -0,0 +1,6 @@
+package model
+
+type LabelPair struct {
```
nicks: this is fine for now, but i feel like we're eventually going to need to split up the k8s package into k8s data structures and k8s client code.

internal/engine/pod_watch.go
```diff
@@ -22,33 +24,74 @@ func NewPodWatcher(kCli k8s.Client) *PodWatcher {
...
 		}
 	}
-	return atLeastOneK8S && state.WatchMounts && !w.watching
+	if atLeastOneK8S {
+		trl := TiltRunLabel()
+
+		neededWatches = append(neededWatches, PodWatch{labels: labels.Set{trl.Key: trl.Value}})
+	}
+
+	return subtract(neededWatches, w.watches), subtract(w.watches, neededWatches)
```
nicks: neat!

internal/tiltfile/k8s.go
```diff
@@ -92,12 +97,14 @@ func (s *tiltfileState) k8sResource(thread *starlark.Thread, fn *starlark.Builti
 	var yamlValue starlark.Value
 	var imageVal starlark.Value
 	var portForwardsVal starlark.Value
+	var extraPodLabelsVal starlark.Value
```
nicks: if you make this a skylark.Sequence, does unpack args do the right thing by default?

internal/engine/pod_watch.go
```diff
@@ -12,8 +14,8 @@ import (
 )
 
 type PodWatcher struct {
```
nicks: a test for the PodWatcher changes would be nice

internal/tiltfile/tiltfile_test.go
```diff
@@ -928,6 +930,49 @@ docker_build('gcr.io/some-project-162817/sancho', '.')
 		m.ImageTargetAt(0).Ref.String())
 }
 
+func TestExtraPodLabels(t *testing.T) {
+	f := newFixture(t)
+	defer f.TearDown()
+
+	f.setupExtraPodLabels("[{'baz': 'quu'}]")
```
nicks: can you add tests for:

```
{"labelA": "valueA", "labelB": "valueB"}
[{"labelA": "valueA", "labelB": "valueB"}]
[{"labelA": "valueA"}. {"labelB": "valueB"}]
```

and assert the differences between them? i don't think the first one currently works, but think it probably should work?

internal/engine/upper.go
```diff
@@ -487,7 +489,17 @@ func handleConfigsReloaded(
 func ensureManifestTargetWithPod(state *store.EngineState, pod *v1.Pod) (*store.ManifestTarget, *store.Pod) {
 	manifestName := model.ManifestName(pod.ObjectMeta.Labels[ManifestNameLabel])
 	if manifestName == "" {
-		return nil, nil
+		// if there's no ManifestNameLabel, then maybe it matches some manifest's ExtraPodLabels
+		for _, m := range state.Manifests() {
+			if m.IsK8s() {
+				for _, lps := range m.K8sTarget().ExtraPodLabels {
+					if lps.AsSelector().Matches(labels.Set(pod.ObjectMeta.GetLabels())) {
```
nicks: i think the AsSelector() call here points to a bug in the API: ExtraPodLabels should be ExtraPodSelectors? i think?

the k8s API does recognize a distinction between LabelSelectors (https://godoc.org/k8s.io/apimachinery/pkg/apis/meta/v1#LabelSelector) and the LabelSets that they match.

internal/tiltfile/k8s.go
landism: if I make it a `starlark.Sequence`, I get a panic:
```
panic: interface conversion: interface is nil, not starlark.Value [recovered]
	panic: interface conversion: interface is nil, not starlark.Value
```
I guess `starlark.Sequence` isn't a `starlark.Value`?

If I make it a `starlark.List`, I get a `log.Fatalf`, and the program dies!:
```
2019/01/31 14:07:40 internal error: invalid pointer type: *starlark.List
```

If I make it a `starlark.Iterable`, it works, but I'm not sure that's what we want since it also accepts non-list iterables, like dicts, which then gets you this confusing error if you give it a dict instead of a dict in a 1-element list:
```
pod labels elements must be dicts; got starlark.String
```

internal/engine/upper.go
landism: The type here is `labels.Selector` rather than `meta.LabelSelector` ([code](https://godoc.org/k8s.io/apimachinery/pkg/labels#Selector)).

Either way, it seems better to make it a selector than a labelset throughout, however:

The `labels.Set` and `labels.Selector` are relatively closely related, and the package has functions to convert [back](https://github.com/windmilleng/tilt/blob/3c4fb73a971b3ea741e1c113865ff5de5130dc90/vendor/k8s.io/apimachinery/pkg/labels/labels.go#L157) and [forth](https://github.com/windmilleng/tilt/blob/3c4fb73a971b3ea741e1c113865ff5de5130dc90/vendor/k8s.io/apimachinery/pkg/labels/labels.go#L61)

k8s exposes an `Equals` methods for `Set`s, but not for `Selector`s, which we need for the diff in the subscriber.

I switched all the `label.Set`s to `label.Selector`s, but then to replace `label.Equals`, I had to implement equality methods, which risk breaking if they add any relevant fields to selectors (or we could use reflection, which might be expensive (or that might not be relevant), and also might break if they add any irrelevant fields to selectors:
```
func setEqual(a, b sets.String) bool {
	if len(a) != len(b) {
		return false
	}

	for k, v := range a {
		if b[k] != v {
			return false
		}
	}

	return true
}

func requirementEqual(a, b labels.Requirement) bool {
	return a.Key() == b.Key() &&
		a.Operator() == b.Operator() &&
		setEqual(a.Values(), b.Values())
}

func requirementsEqual(a, b labels.Requirements) bool {
	if len(a) != len(b) {
		return false
	}

	for i, e1 := range a {
		if !requirementEqual(e1, b[i]) {
			return false
		}
	}

	return true
}
```

We could convert back from selectors to sets for the purpose of equality testing, but that feels kinda weird in the same way that the `AsSelector` here does.

I don't think I have a strong opinion here, and I think I'd lean towards keeping the code as it is in this PR, but I'm fine either way. I just wanted to share the context I had to see if you see a compelling argument one way or another.

internal/tiltfile/k8s.go
landism: allowing either dict or list makes this irrelevant, but that was still kinda surprising

PR #1089, opened 2019-01-31
engine: no manifests is an error condition

jazzdan:


nicks:
surprised there's no unit test? :grimacing:

jazzdan:
@nicks on it!

PR #1090, opened 2019-01-31
tiltfile: reorganize how we match image targets to deploy targets

nicks:
Hello @landism, @jazzdan,

Please review the following commits I made in branch nicks/ch1447/match:

d16bbce5baf729648598e891d0c862cfdc2cd46a (2019-01-31 15:22:10 -0500)
tiltfile: reorganize how we match image targets to deploy targets
fixes https://github.com/windmilleng/tilt/issues/1059

### DIFF
--- a/internal/tiltfile/build_index.go
+++ b/internal/tiltfile/build_index.go
@@ -1,0 +1,78 @@
+package tiltfile
+
+import (
+	"fmt"
+
+	"github.com/docker/distribution/reference"
+)
+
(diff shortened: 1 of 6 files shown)


nicks:
sorry this ended up being a bigger change than i expected...there were lots of places where we were using "reference.Named" and "just the name part of the reference" interchangeably. Most of this PR is just being much more clear about the difference with better typing, etc

PR #1098, opened 2019-02-01
tiltfile: add fail function

jazzdan:


thrawny:
Tested, works great. Thx a lot for the improvement!

Issue #894, opened 2018-12-23
k8s_yaml documented as expecting a string, but does not accept a string


nicks:
https://docs.tilt.build/api.html#api.k8s_yaml

this doc says that you should be able to pass in a string of yaml, but this does not actually work. it only accepts a file path or a blob.

not sure if the API should be changed to accept a string, or if there should be some other function that parses a string into a form that k8s_yaml recognizes

jazzdan:
I addressed part of this in https://github.com/windmilleng/tilt/commit/211eb0c4a6a607f9a172726834d90b593fd997da

jazzdan:
And the other part is now merged.

Now you can pass arbitrary YAML in to a `k8s_yaml` using the new [`yaml`](https://docs.tilt.build/api.html#api.yaml) call like so:

```python
deployment = '''
apiVersion: apps/v1
kind: Deployment
metadata:
  name: sancho
  namespace: sancho-ns
  labels:
    app: sancho
spec:
  replicas: 1
  selector:
    matchLabels:
      app: sancho
  template:
    metadata:
      labels:
        app: sancho
    spec:
      containers:
      - name: sancho
        image: gcr.io/some-project-162817/sancho
        env:
          - name: token
            valueFrom:
              secretKeyRef:
                name: slacktoken
                key: token
      - name: sancho-sidecar
        image: gcr.io/some-project-162817/sancho-sidecar'''
k8s_yaml(yaml(deployment)
```

BladeMcCool:
idk the history but currently there is no "yaml" function but there is a "blob" function that fixed my problem.

# Step 1: Define the namespace manifest for 'cool-namespace'
namespace_yaml = """
apiVersion: v1
kind: Namespace
metadata:
  name: cool-namespace
"""

# Step 2: Load the namespace YAML into Tilt
k8s_yaml(blob(namespace_yaml)) #works (didnt work without wrapping the namespace_yaml in some func, tried 'yaml' like in this thread, tilt said it didnt know what it was, additional research found some post mention 'blob')

PR #1111, opened 2019-02-04
tiltfile: add yaml function

jazzdan:
So that you can take a Starlark string and pass it in to `k8s_yaml`, as demonstrated in the unit test.

This is the last part of #894.


docs/api/api.py
```diff
@@ -113,3 +113,6 @@ def helm(pathToChartDir: Union[str, LocalPath]) -> Blob:
 
 def fail(msg: str) -> None:
   """Raises an error that cannot be intercepted. Can be used anywhere in a Tiltfile."""
+
+def blob(input: str) -> Blob:
```
nicks: i'm still concerned that `blob` is a bad name for this function, but not sure i can come up with a better one

internal/tiltfile/files.go
```diff
@@ -327,3 +327,13 @@ func (s *tiltfileState) helm(thread *starlark.Thread, fn *starlark.Builtin, args
 
 	return newBlob(string(yaml)), nil
 }
+
+func (s *tiltfileState) blob(thread *starlark.Thread, fn *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
+	var input starlark.String
+	err := starlark.UnpackArgs(fn.Name(), args, kwargs, "path", &input)
```
nicks: "path" -> "input" ?

internal/tiltfile/files.go
```diff
@@ -327,3 +327,13 @@ func (s *tiltfileState) helm(thread *starlark.Thread, fn *starlark.Builtin, args
...
+	if err != nil {
+		return nil, fmt.Errorf("Argument 0 (path): %v", err)
```
nicks: i don't think you need this fmt.Error? UnpackArgs should already give you a far better and more descriptive message.

nicks:
i guess the Go way to do this would be to have k8s_yaml accept a YAML type, and make the conversion from blob -> yaml implicit. So the function would be `yaml('string')`? but if you want to go with `blob`, that's ok too

jazzdan:
@nicks I like that approach better, I'm going to do that.

PR #1114, opened 2019-02-05
docs: mention docker compose in readme/install so DC users don't install k8s stuff

maiamcc:
idk if this is strictly necessary but my friend who was trying DC blindly
followed the install guide and ended up with kubectl and was a little annoyed.

I assume this is on the level with how public we want to be about DC support?


README.md
```diff
@@ -28,7 +28,7 @@ If you don't know where to start, start here:
 Download the Tilt binary on
 [the github releases page](https://github.com/windmilleng/tilt/releases).
 
-Tilt expects that you already have Docker and `kubectl` installed.
+Tilt expects that you already have Docker and `kubectl` installed. (Alternately, you can skip Kubernetes altogether and run Tilt with your pre-existing `docker-compose.yml`.)
```
jazzdan: I would remove the parenthetical.

docs/install.md
```diff
@@ -9,6 +9,9 @@ You'll also need:
 - Kubectl, to cuddle your cluster
 - A local Kubernetes cluster (on MacOS, Docker For Mac works for this!)
 
+
+(Not a fan of Kubernetes? You can also use Tilt to [run your existing Docker Compose setup](docker_compose.html), in which case all you need besides Tilt is Docker Compose, and you can ignore Kubernetes-specific instructions on this page.)
```
jazzdan: Same here, the parens feels weird.

jazzdan: Also maybe "Don't use Kubernetes?" instead of "Not a fan of Kubernetes?"

nicks:
should we also remove kubernetes-cli from goreleaser.yml?

maiamcc:
@nicks hmm maybe? idk how expected it is to have everything bundled into your brew install, but i feel like it's unlikely for someone to install tilt if they don't already HAVE docker/k8s... and Joe Beda thought it was maybe weird that we basically re-installed docker for him (since he had it but not via brew)...?

jazzdan:
My 2c is that we should remove kubernetes-cli as a strict dependency.

PR #1117, opened 2019-02-05
k8s: inject built image digests into env vars and CRDs

landism:


internal/k8s/image.go
```diff
@@ -49,6 +50,29 @@ func InjectImageDigest(entity K8sEntity, injectRef reference.Named, policy v1.Pu
 		return K8sEntity{}, false, fmt.Errorf("INTERNAL TILT ERROR: Cannot set PullNever with digest")
 	}
 
+	replaced := false
+	for _, f := range []func(K8sEntity, reference.Named) (K8sEntity, bool, error){
+		func(entity K8sEntity, named reference.Named) (K8sEntity, bool, error) {
+			return injectImageDigestInContainers(entity, named, policy)
+		},
+		injectImageDigestInEnvVars,
```
nicks: this...seems like a lot of indirection to avoid an extra "err != nil" block

landism: well, two extra ```		if err != nil {
			return K8sEntity{}, false, err
		}
		if r {
			replaced = true
		}``` blocks, but fair enough

PR #1127, opened 2019-02-06
fast_build: fix add error message

landism:
This was generating errors like
```
Error: fast_build("add").add(): invalid type for src. Got string want gitRepo OR localPath
```
the `"add"` there is super weird

now we get:
```
Error: fast_build("docker.io/library/my_picnic_image").add(): invalid type for src. Got string, want gitRepo
        OR localPath
```


maiamcc:
oh AWKWARD, good call.

PR #1131, opened 2019-02-07
hud: make modals look more like windows

landism:
effectively, just use a different set of chars when drawing our boxes

Almost all of the code in this PR is to allow us to preserve the existing box behavior, which might not have been worth it until we actually want to use Plain Ol' Boxes again.

image


internal/hud/renderer.go
```diff
@@ -93,9 +93,9 @@ func (r *Renderer) maybeAddAlertModal(vs view.ViewState, layout rty.Component) r
 		l.Add(rty.Fg(rty.TextString(msg), tcell.ColorDefault))
 		l.Add(rty.TextString(""))
 
-		b := rty.NewBox(l)
-		b.SetTitle("! Alert !")
-		layout = r.renderModal(rty.Fg(b, tcell.ColorRed), layout, false)
+		w := rty.NewWindow(l)
+		w.SetTitle("! Alert !")
```
yuindustries: Throwaway comment: This always bugged me. Is this vaguely Spanish? Also, we are way too alarmist for something as innocuous as — say — YAML resources not having logs.

landism: consider it thrown away!

PR #1134, opened 2019-02-07
docs: reorganize table of contents, small fixes

jazzdan:
This PR only contains organizational changes and small edits, no real content change.

I'd like to move the docs to be split in to 4 parts, that you could obviously follow sequentially.

1. Getting Started.
- What we have now.
2. Guides
- Step by step how-to's for things like using Tilt with helm, using Tilt with a react frontend application, using kustomize to templatize your YAML, etc
3. Concepts/Learn More
- Explain the fundamental concepts of tilt in detail: repos, local, images, fast build, resources, etc.
4. API reference
5. About

Prior to this change the table of contents looked like this:
screen shot 2019-02-07 at 2 27 39 pm

Now it looks like this:
screen shot 2019-02-07 at 2 28 00 pm


jazzdan:
I'm going to move skaffold in to Getting Started after talking with @dbentley in person, because it's not really a guide, it's more of a potential onboarding ramp.

README.md
```diff
@@ -40,7 +40,7 @@ Down with YAML!
 Configure Tilt with a `Tiltfile`, written in a small subset of Python called
 [Starlark](https://github.com/bazelbuild/starlark#tour).
 
-To get started, check out some [examples](https://docs.tilt.build/first_example.html) or dive into the
+To get started, check out some [examples](https://docs.tilt.build/example_projects.html) or dive into the
```
dbentley: I'd push them towards the tutorial, or just the docs page in general.

Also, let's remove "Down with YAML!", because that makes it sound like want them to throw out their yaml.

docs/index.md
```diff
@@ -73,13 +73,18 @@ We expect everyone in our community (users, contributors, and employees alike) t
    install
    tutorial
    skaffold
+
+.. toctree::
+   :maxdepth: 1
+   :caption: Guides
```
dbentley: I love the idea of Guides; I'd keep these in Getting Started because I think they're about onboarding, but it's certainly not blocking.

I want more Guides that aren't just for onboarding.

dbentley: My fear: a user does the Tutorial but it takes a while because they use X, then sees the article "X", clicks on it and goes "wow, I wish I'd known this when I was doing the Tutorial."

Therefore, articles that users wish they'd see before the Tutorial should be in "Getting started".

Which makes me think Helm and Docker Compose should be in there. (I also think that Example Projects is kinda in there, but if you don't have a project, you'll figure it out some way).

WDYT?

jazzdan: SGTM! I'll move them back.

jazzdan:
For now I just removed the guide until sections until/if we write more. It felt silly to have one article in a section.

PR #1137, opened 2019-02-07
analytics: periodically report some tilt usage information

landism:
1. Report an "up.running" event every 4 hours while tilt is running (theoretically we have a compelling use case that involves just leaving Tilt running all the time, and we want to count these users appropriately).
2. Count builds, resources, resources w/ DC, resources w/ k8s, resources w/ fast build.

### DIFF
--- a/internal/engine/analytics_reporter.go
+++ b/internal/engine/analytics_reporter.go
@@ -11,7 +11,7 @@ import (
 )
 
 // How often to perodically report data for analytics while Tilt is running
-const analyticsReportingInterval = time.Hour * 4
+const analyticsReportingInterval = time.Hour * 1
 
 type AnalyticsReporter struct {
 	a       analytics.Analytics
@@ -28,11 +28,13 @@ func (ar *AnalyticsReporter) OnChange(ctx context.Context, st store.RStore) {
 	defer st.RUnlockState()
 
 	// wait until state has been kinda initialized
-	if !state.TiltStartTime.IsZero() {
+	if !state.TiltStartTime.IsZero() && state.LastTiltfileError() == nil {
 		ar.started = true
 		go func() {
-			ar.report()
-			time.Sleep(analyticsReportingInterval)
+			for {
+				ar.report()
+				time.Sleep(analyticsReportingInterval)
+			}
 		}()
 	}
 }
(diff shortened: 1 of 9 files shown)


internal/engine/analytics_reporter.go
```diff
@@ -0,0 +1,73 @@
...
+	"strconv"
+	"time"
+
+	"github.com/windmilleng/wmclient/pkg/analytics"
+
+	"github.com/windmilleng/tilt/internal/store"
+)
+
+// How often to perodically report data for analytics while Tilt is running
+const analyticsReportingInterval = time.Hour * 4
```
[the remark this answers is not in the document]
landism: I've lowered it to 1 hour, which I think should be good enough for measuring what we're looking to measure in the next couple months. If we want to get into how often people are editing files and/or reporting data more often, we can tackle that in a later PR.

internal/engine/analytics_reporter.go
```diff
@@ -0,0 +1,81 @@
...
+	state := st.RLockState()
+	defer st.RUnlockState()
+
+	// wait until state has been kinda initialized
+	if !state.TiltStartTime.IsZero() && state.LastTiltfileError() == nil {
+		ar.started = true
+		go func() {
+			for {
+				ar.report()
+				time.Sleep(analyticsReportingInterval)
```
nicks: we usually do this as

select {
  case <-time.After(interval):
  case <-ctx.Done():
    return
}

so that the goroutine stops when the context is canceled (helpful on tests)

internal/engine/analytics_reporter.go
```diff
@@ -0,0 +1,81 @@
...
+
+	tiltfileIsInError := "false"
+	if st.LastTiltfileError() != nil {
+		tiltfileIsInError = "true"
+	}
+
+	ar.a.Incr("up.running", map[string]string{
+		"up.starttime":                 st.TiltStartTime.Format(time.RFC3339),
+		"tiltfile.error":               tiltfileIsInError,
+		"builds.completed_count":       strconv.Itoa(st.CompletedBuildCount),
```
nicks: i would feel better if the metrics system doesn't report these at all on error, so that they don't mess up averages.

PR #1138, opened 2019-02-08
tilt: only save the last 120,000 bytes of logs

landism:


internal/model/log.go
```diff
@@ -0,0 +1,38 @@
...
+
+func (l Log) Empty() bool {
+	return len(l.content) == 0
+}
+
+func (l Log) Copy() Log {
+	return Log{append([]byte{}, l.content...)}
+}
+
+func (l *Log) Append(b []byte) {
```
maiamcc: maybe i'm misunderstanding this func? what happens if the log is:
```shortline\n
really really really really reaaaaaaallly longline\n
more recent stuff...```

even if appending `more recent stuff` gets you MANY bytes over the limit, I think this func will only chop off the very first line (i.e. `shortline`)?

File: internal/store/engine_state.go [hunk 2]
```diff
@@ -473,9 +475,9 @@ type Pod struct {
 	Deleting bool
 
 	// The log for the previously active pod, if any
-	PreRestartLog []byte `testdiff:"ignore"`
+	PreRestartLog model.Log `testdiff:"ignore"`
 	// The log for the currently active pod, if any
-	CurrentLog []byte `testdiff:"ignore"`
+	CurrentLog model.Log `testdiff:"ignore"`
```
nicks: i'm confused about how this works? doesn't this need to be mutable? so shouldn't it be a pointer?

internal/engine/upper.go
```diff
@@ -663,8 +664,8 @@ func handlePodChangeAction(ctx context.Context, state *store.EngineState, pod *v
 	checkForPodCrash(ctx, ms, *podInfo)
 
 	if int(cStatus.RestartCount) > podInfo.ContainerRestarts {
-		podInfo.PreRestartLog = append([]byte{}, podInfo.CurrentLog...)
-		podInfo.CurrentLog = []byte{}
+		podInfo.PreRestartLog = podInfo.CurrentLog.Copy()
```
nicks: i'm not sure that you need a Copy() method? presumably, it it's append only, normal struct copying will do what we want (retain a pointer to the byte-array and the length of the slice)

internal/model/log.go
```diff
@@ -0,0 +1,38 @@
+package model
+
+// At this limit, with one resource having a 120k log, render time was ~20ms and CPU usage was ~70% on an MBP.
+// 70% still isn't great when tilt doesn't really have any necessary work to do, but at least it's usable.
+// A render time of ~40ms was about when the interface started being noticeably laggy to me.
+const maxLogLengthInBytes = 120 * 1000
+
+type Log struct {
```
nicks: i worry that the use of pointer receiver tricks to control mutability might be a bit too clever. maybe we should have two separate structs, Log and ReadOnlyLog?

internal/model/log.go
landism: I think you're misunderstanding this func?

Does this help?
https://play.golang.org/p/vguuX3pJJYX

I'm kinda not sure how the situation you're describing is unusual enough to be worth commenting, so I worry I don't understand your concern.

internal/model/log.go
landism: Maybe I misunderstand what I did here - how is this a "trick" or "clever" and not just the almost the most basic use of pointers?

The use of pointer receivers here is not to *control* mutability, but simply because we can't use pointer receivers throughout renderer.go because there's a bunch of code like `LastBuild().Log.String()`

Making these two separate structs is non-trivial since we need to be able to call both `String` and `Append` on `BuildRecord.Log` (though maybe we want to do that anyway since the renderer should be operating on an unchanging copy of the log, and it looks like right now it's sharing these bytes with the engine state).

I guess I don't really understand your concern since changing `Append` to be a function instead of a method would mean we're no longer using pointer receiver "tricks", but I don't see what problem it solves.

internal/store/engine_state.go
landism: I don't understand - its internal `[]byte` is just as mutable inside the `Log` struct as outside. I'm pretty sure you know how this works better than I do, though, so I'm not sure what I'm missing.

PR #1141, opened 2019-02-08
engine: inject deployID on k8s deploy, use that to find pods [ch1544]

maiamcc:
This lets us simplify the logic to match pods to image, and will let us more easily match pods that were deployed by tilt but don't have tilt-built images.

### DIFF
--- a/internal/engine/buildcontroller_test.go
+++ b/internal/engine/buildcontroller_test.go
@@ -1,14 +1,13 @@
 package engine
 
 import (
-	"fmt"
 	"testing"
 	"time"
 
 	"github.com/stretchr/testify/assert"
 	"github.com/windmilleng/tilt/internal/hud/view"
 	"github.com/windmilleng/tilt/internal/model"
-	store "github.com/windmilleng/tilt/internal/store"
+	"github.com/windmilleng/tilt/internal/store"
 	"github.com/windmilleng/tilt/internal/watch"
 )
 
@@ -35,7 +34,7 @@ func TestBuildControllerOnePod(t *testing.T) {
 	f.assertAllBuildsConsumed()
 }
 
-func TestBuildControllerTwoPods(t *testing.T) {
+func TestBuildControllerWontContainerBuildWithTwoPods(t *testing.T) {
 	f := newTestFixture(t)
 	defer f.TearDown()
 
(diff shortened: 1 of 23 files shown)


internal/engine/buildcontroller_test.go
```diff
@@ -47,15 +46,17 @@ func TestBuildControllerTwoPods(t *testing.T) {
 	assert.Equal(t, manifest.ImageTargetAt(0), call.image())
 	assert.Equal(t, []string{}, call.oneState().FilesChanged())
 
+	// Associate the pods with the manifest state
 	podA := f.testPod("pod-a", "fe", "Running", testContainer, time.Now())
 	podB := f.testPod("pod-b", "fe", "Running", testContainer, time.Now())
-	image := fmt.Sprintf("%s:%s", f.imageNameForManifest("fe").String(), "now")
```
nicks: this code is critical for the test, because it makes sure both pods have the same image. (otherwise, they will get different image ids)

internal/engine/upper.go
```diff
@@ -261,6 +261,7 @@ func handleBuildCompleted(ctx context.Context, engineState *store.EngineState, c
 	bs := ms.CurrentBuild
 	bs.Error = err
 	bs.FinishTime = time.Now()
+	bs.DeployID = cb.Result.DeployID
```
nicks: i'm pretty sure this introduces a race condition in the case where you get the Pod event before the BuildComplete event

one solution to this would be to pick the deploy id in BuildController and attach it to the BuildStarted event, so that it's guaranteed to get into the model before the deploy happens.

internal/engine/buildcontroller_test.go
maiamcc: But we no longer match on the image the pod is running, so this is moot, right? Is there some other behavior you think we should be testing?

maiamcc: (`testPod` be default labels the pods with TestDeployId, and TestBuildAndDeployer by default deploys with TestDeployId, so BOTH these pods will get attached to `manifest`.)

internal/engine/upper.go
maiamcc: aaahhhh shoot good call.

maiamcc:
ugh okay 😪 re-review pls?

maiamcc:
Bump!

PR #1143, opened 2019-02-08
k8s: more robust watches and error handling

nicks:
Hello @landism, @maiamcc,

Please review the following commits I made in branch nicks/namespaces:

44e8b7306fa92e1a3534a6cc65bc7284901c7ae2 (2019-02-08 17:54:21 -0500)
k8s: more robust watches and error handling
fixes https://github.com/windmilleng/tilt/issues/1139

### DIFF
--- a/internal/k8s/watch.go
+++ b/internal/k8s/watch.go
@@ -2,21 +2,58 @@ package k8s
 
 import (
 	"context"
+	"fmt"
+	"net/http"
 
+	"github.com/pkg/errors"
 	"github.com/windmilleng/tilt/internal/model"
 
 	"k8s.io/api/core/v1"
+	apiErrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/watch"
 )
+
+type watcherFactory func(namespace string) watcher
+type watcher interface {
+	Watch(options metav1.ListOptions) (watch.Interface, error)
+}
+
+func (kCli K8sClient) makeWatcher(f watcherFactory, ls labels.Selector) (watch.Interface, error) {
+	// passing "" gets us all namespaces
+	watcher, err := f("").Watch(metav1.ListOptions{LabelSelector: ls.String()})
+	if err == nil {
+		return watcher, nil
+	}
+
+	// If the request failed, we might be able to recover.
+	statusErr, isStatusErr := err.(*apiErrors.StatusError)
+	if !isStatusErr {
+		return nil, err
+	}
+
+	status := statusErr.ErrStatus
+	if status.Code == http.StatusForbidden {
+		// If this is a forbidden error, maybe the user just isn't allowed to watch this namespace.
+		// Let's narrow our request to just the config namespace, and see if that helps.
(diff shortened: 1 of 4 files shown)


internal/k8s/watch.go
```diff
@@ -2,23 +2,59 @@ package k8s
...
+type watcherFactory func(namespace string) watcher
+type watcher interface {
+	Watch(options metav1.ListOptions) (watch.Interface, error)
+}
 
+func (kCli K8sClient) makeWatcher(f watcherFactory, ls labels.Selector) (watch.Interface, error) {
 	// passing "" gets us all namespaces
-	watcher, err := kCli.core.Pods("").Watch(metav1.ListOptions{LabelSelector: ls.String()})
-	if err != nil {
+	watcher, err := f("").Watch(metav1.ListOptions{LabelSelector: ls.String()})
```
maiamcc: what benefit do we get from watching all namespaces rather than just the one in the config?

nicks: good questions! Two reasons:
1) The config doesn't actually restrict what namespace you deploy to, the RBAC rules do that. The config is just a hint at what your RBAC rules might be.
2) The vast majority of users won't have namespaces in their configs - I don't expect that anyone at Windmill team does. You'd usually have it set up for you by a cluster-admin.

landism: I might describe it as "the config just specifies the default namespace to use when you run kubectl without --namespace"

> I don't expect that anyone at Windmill team does
FWIW, I found this alias last week and have found it to be a *huge* QOL improvement when using non-default namespaces `alias kcd='kubectl config set-context $(kubectl config current-context) --namespace '` (though I expect it to cause me problems at some point when I change namespaces and unintentionally cause tilt to start deploying to the new namespace)

nicks: hmmm...what i tested this last week, AFAICT, this doesn't affect what namespace you deploy to when you use tilt? 

the namespaces are in the yaml. if you don't have a namespace in your yaml, it still uses the default namespace, regardless of what's in your kubectl context.

were you seeing something different?

landism: yep, if I run `tilt up snack`, then in another window run `kcd fission`, then edit snack, then tilt deploys snack to the fission namespace

internal/k8s/watch.go
```diff
@@ -2,23 +2,59 @@ package k8s
...
+	}
+
+	// If the request failed, we might be able to recover.
+	statusErr, isStatusErr := err.(*apiErrors.StatusError)
+	if !isStatusErr {
 		return nil, err
 	}
 
+	status := statusErr.ErrStatus
+	if status.Code == 403 {
```
landism: This is fine as-is, but I wonder if it'd be worth getting the namespace from the k8s entities instead of from the kube context. The current solution leaves us with a bit of a corner case:
* user's kube context is set to `default`
* user deploys service to namespace `foo`
* user lacks access to watch namespace `bar`
* watching `""` fails because of `bar`, so tilt falls back to watching `default`, resulting in tilt never getting pod status and never reporting an error, so the user is kinda left with no information other than "tilt doesn't know what's going on"

I think this PR is still an improvement over status quo, though.

landism: `http.StatusForbidden`, if you want

internal/k8s/watch.go
nicks: hmmmm...that's different than i was seeing...will play with it a bit more...

internal/k8s/watch.go
nicks: done

PR #1146, opened 2019-02-11
k8s: pull context from the k8s api instead of from the cli, and capture namespace in kubectlRunner

nicks:
Hello @landism, @maiamcc,

Please review the following commits I made in branch nicks/kubecontext:

7d4b489451493b89e2a667adafd97ac40d4850ff (2019-02-11 16:27:14 -0500)
k8s: pull context from the k8s api instead of from the cli, and capture namespace in kubectlRunner

### DIFF
--- a/internal/k8s/client.go
+++ b/internal/k8s/client.go
@@ -103,38 +103,27 @@ var _ Client = K8sClient{}
 
 type PortForwarder func(ctx context.Context, restConfig *rest.Config, core apiv1.CoreV1Interface, namespace string, podID PodID, localPort int, remotePort int) (closer func(), err error)
 
-func ProvideK8sClient(ctx context.Context, envOrErr EnvOrError) (Client, error) {
-	env := envOrErr.Env
-	if env == EnvNone {
-		// No k8s, so no need to get any further configs
-		return &explodingClient{err: envOrErr.Err}, nil
-	}
-
-	loader := ProvideClientConfig()
-	config, err := ProvideRESTConfig(loader)
-	if err != nil {
-		return K8sClient{}, err
-	}
-	namespace, err := ProvideConfigNamespace(loader)
-	if err != nil {
-		return K8sClient{}, err
-	}
-	coreV1Interface, err := ProvideRESTClient(config)
-	if err != nil {
-		return K8sClient{}, err
-	}
-	portForwarder := ProvidePortForwarder()
-	k8sClient := NewK8sClient(ctx, env, coreV1Interface, config, portForwarder, namespace)
-	return k8sClient, nil
-}
-
-func NewK8sClient(
+func ProvideK8sClient(
 	ctx context.Context,
 	env Env,
-	core apiv1.CoreV1Interface,
-	restConfig *rest.Config,
 	pf PortForwarder,
-	configNamespace Namespace) K8sClient {
+	configNamespace Namespace,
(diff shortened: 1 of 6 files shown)


nicks:
(not totally convinced this is the right path to go down, tho was playing around with getting the --klog into the kubectl calls, and then started wondering why we're calling out to kubectl at all...)

nicks:
nm, the namespace flag didn't work quite like i thought it did, i rolled that part of the pr back

internal/k8s/env.go
```diff
@@ -25,40 +22,17 @@ func (e Env) IsLocalCluster() bool {
...
-type EnvOrError struct {
-	Env Env
-	Err error
+func ProvideEnv(kubeContext KubeContext) Env {
+	return EnvFromString(string(kubeContext))
 }
 
-func ProvideEnvOrError(ctx context.Context) EnvOrError {
-	kubeContext, err := detectKubeContext()
+func ProvideKubeContext(clientLoader clientcmd.ClientConfig) (KubeContext, error) {
```
maiamcc: if you get rid of the `EnvOrError` struct and just have this return an error if it fails, then the whole wire-up will fail if we can't get kubeContext -- which is NOT what we want if the user is just using Docker Compose. Can you put back the envOrError so things don't blow up if you don't have kubernetes configured? (Or is that a case we don't actually want to support?)

internal/k8s/client.go
```diff
@@ -103,38 +103,27 @@ var _ Client = K8sClient{}
...
-	loader := ProvideClientConfig()
-	config, err := ProvideRESTConfig(loader)
-	if err != nil {
-		return K8sClient{}, err
-	}
-	namespace, err := ProvideConfigNamespace(loader)
+	restConfig, err := ProvideRESTConfig(clientLoader)
 	if err != nil {
-		return K8sClient{}, err
+		return &explodingClient{err: err}
```
maiamcc: good call

internal/k8s/client.go
```diff
@@ -103,38 +103,27 @@ var _ Client = K8sClient{}
...
+	ctx context.Context,
+	env Env,
+	pf PortForwarder,
+	configNamespace Namespace,
+	runner kubectlRunner,
+	clientLoader clientcmd.ClientConfig) Client {
 	if env == EnvNone {
 		// No k8s, so no need to get any further configs
-		return &explodingClient{err: envOrErr.Err}, nil
+		return &explodingClient{err: fmt.Errorf("Kubernetes context not set")}
```
maiamcc: imo it's useful to pass the original error along to this point so that user actually gets some useful info if their k8s stuff fails...?

internal/k8s/env.go
nicks: don't panic! it still works. now you get an error that looks like this:

```
Error: Error watching services. Are you connected to kubernetes?
: could not set up k8s client: invalid configuration: no configuration has been provided
```

internal/k8s/client.go
nicks: see above comment

nicks:
fwiw, i'm increasingly convinced that we shouldn't call out to kubectl at all. We should just init our own unstructured go-client and make any API calls we need. It would be easier to debug and less error-prone. Currently we only use kubectl for apply/delete

PR #1148, opened 2019-02-12
tilt: release v0.7.3

landism:


docs/first_example.md
```diff
@@ -13,7 +13,7 @@ To double-check, run:
 
 ```
 $ tilt version
-v0.5.0-dev, built 2019-01-14
+v0.7.4-dev, built 2019-02-12
```
jazzdan: Won't the production version not have a `-dev`, unless they built Tilt from scratch?

landism: yeah

apparently you just deleted this file anyway though, so that makes things easy

PR #1150, opened 2019-02-12
ci: store macos test results

landism:


internal/cli/analytics_test.go
```diff
@@ -26,6 +26,7 @@ func TestNormalizedGitRemoteUsername(t *testing.T) {
 }
 
 func TestGitOrigin(t *testing.T) {
+	t.Fatal("foo")
```
jazzdan: 👀

landism: forcing a failure to test the behavior before I add reviewers

AFAIK there's no way to test this without creating a PR

jazzdan: ahhh I gotcha

.circleci/config.yml
```diff
@@ -72,9 +72,13 @@ jobs:
       - run: curl --silent --show-error --location --fail --retry 3 --output /tmp/helm.tar.gz  https://storage.googleapis.com/kubernetes-helm/helm-v2.12.1-darwin-amd64.tar.gz &&
              tar -xz -C /tmp -f /tmp/helm.tar.gz &&
              mv /tmp/darwin-amd64/helm /usr/local/bin/helm
+      - run: go get -u gotest.tools/gotestsum
```
landism: this is potentially worth caching, but it was only 20s of a 5m32s run

(also, our existing homebrew caching looks like it might not be working, which makes me hesitant to dig into this)

PR #1151, opened 2019-02-12
tiltfile: parallelize docker-compose config calls for faster manifest load [ch1201]

maiamcc:
this will save us .4s per dc.yml when loading manifests.

in particular, a big gain for tests:
```
$ # on master
$ go test github.com/windmilleng/tilt/internal/engine -count=1 && go test github.com/windmilleng/tilt/internal/tiltfile -count=1
ok  	github.com/windmilleng/tilt/internal/engine	15.682s
ok  	github.com/windmilleng/tilt/internal/tiltfile	14.437s

$ # on this branch
$ go test github.com/windmilleng/tilt/internal/engine -count=1 && go test github.com/windmilleng/tilt/internal/tiltfile -count=1
ok  	github.com/windmilleng/tilt/internal/engine	10.958s
ok  	github.com/windmilleng/tilt/internal/tiltfile	8.815s
```
So that's a savings of c. 10s per test suite run.

### DIFF
--- a/internal/tiltfile/docker_compose.go
+++ b/internal/tiltfile/docker_compose.go
@@ -8,7 +8,6 @@ import (
 	"path/filepath"
 	"reflect"
 	"strings"
-	"time"
 
 	"github.com/docker/distribution/reference"
 	"github.com/pkg/errors"
@@ -16,6 +15,7 @@ import (
 	"github.com/windmilleng/tilt/internal/dockerfile"
 	"github.com/windmilleng/tilt/internal/model"
 	"go.starlark.net/starlark"
+	"golang.org/x/sync/errgroup"
 	"gopkg.in/yaml.v2"
 )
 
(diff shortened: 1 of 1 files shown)


internal/tiltfile/docker_compose.go
```diff
@@ -293,6 +282,56 @@ func parseDCConfig(ctx context.Context, dcc dockercompose.DockerComposeClient, c
 	return services, nil
 }
 
+func getConfigAndServiceNames(ctx context.Context, dcc dockercompose.DockerComposeClient,
+	configPath string) (conf dcConfig, svcNames []string, err error) {
+	// calls to `docker-compose config` take a bit, and we need two,
+	// so do them in parallel to make things faster
+	configGot := make(chan error, 1)
```
jazzdan: Could we make these unbuffered channels? I know it's guaranteed that each channel will only receive one thing but I find the behavior of unbuffered channels easier to reason about.

internal/tiltfile/docker_compose.go
```diff
@@ -293,6 +282,56 @@ func parseDCConfig(ctx context.Context, dcc dockercompose.DockerComposeClient, c
...
+	servicesGot := make(chan error, 1)
+
+	go func() {
```
nicks: the idiomatic way to do this is with an errgroup
https://godoc.org/golang.org/x/sync/errgroup#ex-Group--Parallel

internal/tiltfile/docker_compose.go
```diff
@@ -293,6 +282,56 @@ func parseDCConfig(ctx context.Context, dcc dockercompose.DockerComposeClient, c
...
+		}
+		close(servicesGot)
+	}()
+
+	select {
+	case err := <-configGot:
+		if err != nil {
+			return conf, svcNames, err
+		}
+	case <-time.After(time.Second * 2):
```
nicks: it seems weird to have a timeout here when there wasn't a timeout before? also 2s is shorter than some "normal" times i've seen

internal/tiltfile/docker_compose.go
maiamcc: oh yeah, right! will do

internal/tiltfile/docker_compose.go
maiamcc: yeahhh i just got antsy about leaving a `<- chan` in the middle of the code? but you're right that the previous behavior wasn't any safer in that regard.

PR #1153, opened 2019-02-12
hud: add resource log benchmark

jazzdan:
With a log of just "hi hello" benchmark looks like this on my machine:

```
goos: darwin
goarch: amd64
pkg: github.com/windmilleng/tilt/internal/hud
BenchmarkBigLog-4   	    2000	    913658 ns/op	  225756 B/op	    2496 allocs/op
```

```
goos: darwin
goarch: amd64
pkg: github.com/windmilleng/tilt/internal/hud
BenchmarkBigLog-4   	     200	   8741978 ns/op	 5877859 B/op	   53829 allocs/op
```


jazzdan:
@yuindustries oh, that's already there! 😅

PR #1156, opened 2019-02-12
engine: allow explicit k8s_resource with just yaml, no image, to appear in hud, get logs, etc. [ch1457]

maiamcc:
Hello @nicks, @dbentley,

Please review the following commits I made in branch maiamcc/explicit-resource-without-build:

9f0a81cb93dd44fe0abd9f1afdddc3f78c6acd36 (2019-02-12 18:28:27 -0500)
make sure logs get into log modal for resource even if redeploy is really fast

9336cf7f783305620846159132d417cf243fad44 (2019-02-12 17:44:58 -0500)
test container matching

43a94866c2121f7eab823d326ad896671a074d49 (2019-02-12 16:49:51 -0500)
port forwarding

d54847b886d33d67ae12b35ae357e1d43e87a653 (2019-02-12 14:57:04 -0500)
ImageBuildAndDeployer can handle calls with only deploy targets, no build targets (i.e. only k8s yaml, no image builds)

### DIFF
--- a/internal/engine/upper_test.go
+++ b/internal/engine/upper_test.go
@@ -151,8 +151,6 @@ func (b *fakeBuildAndDeployer) BuildAndDeploy(ctx context.Context, st store.RSto
 		buildID = call.image().ID()
 		buildImageRef = call.image().Ref
 
-	} else {
-		b.t.Fatalf("Invalid call: %+v", call)
 	}
 
 	ids := []model.TargetID{}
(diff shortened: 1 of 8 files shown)


internal/engine/upper_test.go
```diff
@@ -151,8 +151,6 @@ func (b *fakeBuildAndDeployer) BuildAndDeploy(ctx context.Context, st store.RSto
 		buildID = call.image().ID()
 		buildImageRef = call.image().Ref
 
-	} else {
```
nicks: could we still emit an error in the case where there are no k8s deploys?

PR #1159, opened 2019-02-14
tiltfile: smartly pull up/cluster pod-creating yamls to be watched for status/logs [ch1583]

maiamcc:
Hello @nicks, @jazzdan, @dbentley,

Please review the following commits I made in branch maiamcc/unbuilt-resources:

959b692ce0d60676f4f631e80ba6f60e21f81ec4 (2019-02-14 11:21:00 -0500)
tests

b337fdc53d246df01675235458a8e01d69e9bb4f (2019-02-13 17:08:48 -0500)
i think this code works? let's test


internal/tiltfile/tiltfile.go
```diff
@@ -74,15 +75,66 @@ func Load(ctx context.Context, filename string, matching map[string]bool, logs i
...
+		podTemplates, err := k8s.ExtractPodTemplateSpec(e)
+		if err != nil {
+			return nil, model.Manifest{}, errors.Wrap(err, "extracting pod template spec")
+		}
+		if len(podTemplates) == 0 {
+			return nil, model.Manifest{}, errors.Wrap(err, "expected at least one pod template spec, got none")
+		}
+
+		if len(podTemplates) > 1 {
+			logger.Get(ctx).Infof("Found multiple pod templates on %s, "+
```
jazzdan: I think this should be printed to the [Tiltfile log](https://github.com/windmilleng/tilt/blob/b4a3adbffa6396679ccbd26089a8fbc66e27d971/internal/tiltfile/tiltfile_state.go#L44), not the global log.

nicks: i'm surprised we created a separate writer for the Tiltfile log, rather than attaching it to the context?

internal/tiltfile/tiltfile.go
```diff
@@ -74,15 +75,66 @@ func Load(ctx context.Context, filename string, matching map[string]bool, logs i
...
+	if err != nil {
+		return nil, model.Manifest{}, err
+	}
+
+	for _, e := range withPodSpec {
+		podTemplates, err := k8s.ExtractPodTemplateSpec(e)
+		if err != nil {
+			return nil, model.Manifest{}, errors.Wrap(err, "extracting pod template spec")
+		}
+		if len(podTemplates) == 0 {
```
nicks: it seems weird to have both a filter function, and a filter check here? it seems like the FilterByHasPodTemplateSpec function isn't saving you any work. you could just make this a `continue` and you'd be fine.

internal/tiltfile/tiltfile.go
```diff
@@ -74,15 +75,66 @@ func Load(ctx context.Context, filename string, matching map[string]bool, logs i
...
+		if err != nil {
+			return nil, model.Manifest{}, err
+		}
+		podCreators = append(podCreators, m)
+
+		allRest = rest
 	}
 
-	return manifests, yamlManifest, s.configFiles, err
+	remaining, err = k8s.NewK8sOnlyManifest(unresourcedName, allRest)
```
nicks: did you intend to do something with this error?

internal/tiltfile/tiltfile.go
```diff
@@ -74,15 +75,66 @@ func Load(ctx context.Context, filename string, matching map[string]bool, logs i
...
+	return manifests, remainingYamlManifest, s.configFiles, err
+}
+
+// k8sManifestsForPodCreators groups the given k8s entities into pod creator clusters
+// (e.g. DeploymentA is a pod creator, and ServiceA would belong to its pod creator
+// cluster b/c it applies to pods created by DeploymentA); returns a k8s-only manifest
+// per pod creator cluster, and a single manifest containing the rest of the entities
+// (will be deployed/displayed in HUD as a unit).
+func k8sManifestsForPodCreators(ctx context.Context, entities []k8s.K8sEntity) (podCreators []model.Manifest,
+	remaining model.Manifest, err error) {
```
nicks: i'm not sure that you're using the return parameter names here? i don't think we should do this just for documentation purposes, because it changes the semantics of your program and makes errors harder to catch, like in the mistake below.

internal/tiltfile/tiltfile_test.go
```diff
@@ -1127,13 +1170,37 @@ func (f *fixture) yaml(path string, entities ...k8sOpts) {
...
+			}
 			s = strings.Replace(s, testyaml.SnackName, e.name, -1)
 			objs, err := k8s.ParseYAMLFromString(s)
 			if err != nil {
 				f.t.Fatal(err)
 			}
 
 			entityObjs = append(entityObjs, objs...)
+		case serviceHelper:
+			s := testyaml.DoggosServiceYaml
```
nicks: i'm surprised you didn't use the InjectLabels function, which seems a bit more robust

internal/tiltfile/tiltfile.go
jazzdan: The aim was to disambiguate errors/messages that could occur anywhere in Tilt and messages that happen _during_ Tiltfile execution.

jazzdan: Ah I see, I could have instead created a new context with that logger that we use during Tiltfile execution. That's a good idea and we should do that. :)

internal/tiltfile/tiltfile_test.go
maiamcc: Oh whoops I forgot about that -- tho it might be more of a pain to take out the existing labels with that?

internal/tiltfile/tiltfile.go
maiamcc: good call. no longer relevant with refactor.

internal/tiltfile/tiltfile.go
maiamcc: good call. don't think this is relevant w/ the refactor tho

internal/tiltfile/tiltfile.go
maiamcc: 😅

internal/tiltfile/tiltfile_test.go
maiamcc: ah, i'm a goof. this was fine, used the yaml manipulation funcs to do this in my latest commit

maiamcc:
Did some refactoring on @dbentley's advice, would love a re-review!

PR #1160, opened 2019-02-14
hud: image build = N/A for resources w/o image build [ch1594]

maiamcc:
changed the `BUILD STATUS` header to `IMAGE BUILD`, and resources w/o
image targets (including Tiltfile, Global Yaml, DC resources) show
"N/A" instead of "OK"/"building"/the timetake there.

the trade-off here is that we lose the live-ness of having timetake
for these resources. this isn't such a big deal for YAML-only resources,
which usu. take <0.5s, but for docker compose resources, this will
feel really weird :(


maiamcc:
screen shot 2019-02-14 at 11 37 30 am

maiamcc:
screen shot 2019-02-14 at 2 56 36 pm
@yuindustries better?

PR #1162, opened 2019-02-14
k8s: watch pods with informer

jazzdan:


internal/k8s/client.go
```diff
@@ -278,6 +285,15 @@ func (k K8sClient) actOnEntities(ctx context.Context, cmdArgs []string, entities
...
+func ProvideClientSet(cfg *rest.Config) (*kubernetes.Clientset, error) {
+	clientSet, err := kubernetes.NewForConfig(cfg)
+	if err != nil {
+		return nil, err
+	}
+
+	return clientSet, nil
+}
+
 func ProvideCoreInterface(cfg *rest.Config) (apiv1.CoreV1Interface, error) {
```
nicks: let's update this to take a ClientSet, so that we don't have two independent ClientSets running around

internal/k8s/watch.go
```diff
@@ -48,39 +52,61 @@ func (kCli K8sClient) makeWatcher(f watcherFactory, ls labels.Selector) (watch.I
 }
 
 func (kCli K8sClient) WatchPods(ctx context.Context, ls labels.Selector) (<-chan *v1.Pod, error) {
-	ch := make(chan *v1.Pod)
-	watcher, err := kCli.makeWatcher(func(ns string) watcher {
+	// HACK(dmiller): There's no way to get errors out of an informer. See https://github.com/kubernetes/client-go/issues/155
+	// In the meantime, at least to get authorization and some other errors let's try to set up a watcher and then just
+	// throw it away.
+	_, err := kCli.makeWatcher(func(ns string) watcher {
```
nicks: you'll need to call watcher.Stop()

internal/k8s/watch.go
```diff
@@ -48,39 +52,61 @@ func (kCli K8sClient) makeWatcher(f watcherFactory, ls labels.Selector) (watch.I
...
 
-	go func() {
-		for {
-			select {
-			case event, ok := <-watcher.ResultChan():
-				if !ok {
-					close(ch)
-					return
-				}
+	// HACK(dmiller): this can log to stderr. To prevent that we remove the error handler.
```
nicks: are you sure? i'm pretty sure this logs to klog by default, not stderr, which we already have special handling for, so it should be ok

internal/k8s/watch.go
```diff
@@ -48,39 +52,61 @@ func (kCli K8sClient) makeWatcher(f watcherFactory, ls labels.Selector) (watch.I
...
-					continue
-				}
+	informer.AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc: func(obj interface{}) {
+			mObj, ok := obj.(*v1.Pod)
+			if ok {
+				ch <- mObj
+			}
+		},
+		DeleteFunc: func(obj interface{}) {
```
nicks: this is ok for now, but i wonder if we should pass along that this was a Delete event. we have some logic in upper to try detect when the pod is being deleted but tbh it probably misses cases

jazzdan: I agree that it would be great to do that in a future PR. I'll make a story for it.

PR #1170, opened 2019-02-15
model: Add DependencyIDs to TargetSpec, so it looks more like a real graph

nicks:
Hello @landism, @jazzdan,

Please review the following commits I made in branch nicks/deps:

bb51db5a602dfbb0673c9464872a495f5d243deb (2019-02-15 15:18:59 -0500)
model: Add DependencyIDs to TargetSpec, so it looks more like a real graph


nicks:
(note that these aren't consumed anywhere yet...this just clears the way for ImageTargets that depend on other ImageTargets)

PR #1172, opened 2019-02-15
tiltfile: infer image dependencies

nicks:
Hello @jazzdan, @landism,

Please review the following commits I made in branch nicks/findimages:

378f2e0e588320038faf387724d7b21243588de5 (2019-02-15 16:44:51 -0500)
tiltfile: infer image dependencies

### DIFF
--- a/internal/dockerfile/dockerfile_test.go
+++ b/internal/dockerfile/dockerfile_test.go
@@ -117,3 +117,45 @@ func TestNoAddsToNoMounts(t *testing.T) {
 	}
 	assert.Empty(t, mounts)
 }
+
+func TestFindImages(t *testing.T) {
+	df := Dockerfile(`FROM gcr.io/image-a`)
+	images, err := df.FindImages()
+	assert.NoError(t, err)
+	if assert.Equal(t, 1, len(images)) {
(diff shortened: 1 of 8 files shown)


internal/dockerfile/dockerfile_test.go
```diff
@@ -117,3 +117,20 @@ func TestNoAddsToNoMounts(t *testing.T) {
...
+func TestFromImage(t *testing.T) {
+	df := Dockerfile(`FROM gcr.io/image-a`)
+	images, err := df.FindImages()
+	assert.NoError(t, err)
+	if assert.Equal(t, 1, len(images)) {
+		assert.Equal(t, "gcr.io/image-a", images[0].String())
+	}
+}
+
+func TestFromImageMalformed(t *testing.T) {
```
jazzdan: Maybe add a test for a Dockerfile with a malformed FROM?  Like:

```Dockerfile
FROM
```

or

```Dockerfile
FROM foo bar
```

Also do we want to support this?
```Dockerfile
FROM foo AS bar
```

nicks: added some more tests!

jazzdan: Awesome, thanks!

PR #1180, opened 2019-02-19
hud: show warnings for Tiltfile exec

yuindustries:
2019-02-19 - warning


internal/tiltfile/tiltfile_state.go
```diff
@@ -157,7 +158,8 @@ func (s *tiltfileState) assemble() (resourceSet, []k8s.K8sEntity, error) {
 
 	err = s.buildIndex.assertAllMatched()
 	if err != nil {
-		return resourceSet{}, nil, err
+		s.logger.Printf("WARNING: %s", err)
```
jazzdan: Do we need a new line here?

PR #1181, opened 2019-02-19
engine: add timeout for service watch

jazzdan:
This may fix a problem Frank saw with Tilt hanging forever upon start when his Kube cluster wasn't ready

### DIFF
--- a/internal/engine/service_watch.go
+++ b/internal/engine/service_watch.go
@@ -14,6 +14,8 @@ import (
 	"github.com/windmilleng/tilt/internal/k8s"
 )
 
+var watchTimeout = 5 * time.Second
+
 type ServiceWatcher struct {
 	kCli     k8s.Client
 	watching bool
@@ -46,7 +48,7 @@ func (w *ServiceWatcher) OnChange(ctx context.Context, st store.RStore) {
 	}
 	w.watching = true
 
-	ctx2, cancel := context.WithTimeout(ctx, 5*time.Second)
+	ctx2, cancel := context.WithTimeout(ctx, watchTimeout)
 	defer cancel()
 	ch, err := w.kCli.WatchServices(ctx2, []model.LabelPair{k8s.TiltRunLabel()})
 	if err != nil {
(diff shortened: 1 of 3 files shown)


internal/engine/service_watch.go
```diff
@@ -45,7 +46,9 @@ func (w *ServiceWatcher) OnChange(ctx context.Context, st store.RStore) {
 	}
 	w.watching = true
 
-	ch, err := w.kCli.WatchServices(ctx, []model.LabelPair{k8s.TiltRunLabel()})
+	ctx2, cancel := context.WithTimeout(ctx, 5*time.Second)
+	defer cancel()
+	ch, err := w.kCli.WatchServices(ctx2, []model.LabelPair{k8s.TiltRunLabel()})
 	if err != nil {
 		err = errors.Wrap(err, "Error watching services. Are you connected to kubernetes?\n")
```
maiamcc: "Are you connected to Kubernetes" is not a helpful error message if we timed out, I might remove that bit and trust the `err` to contain the relevant context.

nicks: fwiw, i thought we added that precisely because we got user reports that they didn't understand the error message. there are a wide variety of weird error messages you can get from this function

internal/engine/service_watch.go
```diff
@@ -45,7 +46,9 @@ func (w *ServiceWatcher) OnChange(ctx context.Context, st store.RStore) {
 	}
 	w.watching = true
 
-	ch, err := w.kCli.WatchServices(ctx, []model.LabelPair{k8s.TiltRunLabel()})
+	ctx2, cancel := context.WithTimeout(ctx, 5*time.Second)
+	defer cancel()
+	ch, err := w.kCli.WatchServices(ctx2, []model.LabelPair{k8s.TiltRunLabel()})
```
nicks: can you also update WatchPods? there's no guarantee which one will run first

internal/engine/service_watch.go
maiamcc: Right, but if we get this error due to a timeout, it doesn't imply that we're not connected to k8s, right, but rather that something ELSE is wrong? in that case, the hint "are you connected to k8s" sends the user down a rabbit hole.

PR #1189, opened 2019-02-20
tiltfile func to filter yaml by label(s) [ch1626]

maiamcc:
Hello @landism, @jazzdan,

Please review the following commits I made in branch maiamcc/split-out-yaml:

55b701d6041bc2dda17c6e10349aecf5c6a3dc53 (2019-02-20 16:38:02 -0500)
tiltfile func to filter yaml by label

2c69e3cbe075ea0c76669842604ef4a2522d3702 (2019-02-20 16:28:03 -0500)
metadata label match funcs

a0f31ff53c0063a1c51f0ed9a23f4b6001c44616 (2019-02-20 14:51:58 -0500)
benchmark parse/unparse yaml

### DIFF
--- a/internal/tiltfile/tiltfile_test.go
+++ b/internal/tiltfile/tiltfile_test.go
@@ -12,6 +12,7 @@ import (
 
 	"github.com/stretchr/testify/assert"
 	"github.com/windmilleng/tilt/internal/docker"
+	"github.com/windmilleng/tilt/internal/yaml"
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/labels"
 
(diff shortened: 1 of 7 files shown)


maiamcc:
Once this lands, it should be pretty easy to add args for matching by name and/or kind as well.

internal/tiltfile/tiltfile_test.go
```diff
@@ -896,6 +897,21 @@ docker_build("gcr.io/foo", "foo", cache='/path/to/cache')
 	f.loadErrString("no such file or directory")
 }
 
+func TestFilterYaml(t *testing.T) {
+	f := newFixture(t)
+	defer f.TearDown()
+	f.file("k8s.yaml", yaml.ConcatYAML(testyaml.DoggosDeploymentYaml, testyaml.SnackYaml, testyaml.SanchoYAML))
+	f.file("Tiltfile", `labels = {'app': 'doggos'}
```
jazzdan: [Label selectors in Kubernetes](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#label-selectors) offer other options than just strict equality, such as:
* Not equal
* Set contains

If people would like that supported in the future I don't think that could be done in the current syntax. What do you think about changing this

```python
labels = {'app': 'doggos'}
```
in to this:
```python
labels = ["app=doggos"]
```

So that we could do things like this in the future?
```python
labels = ["app=doggos", "environment!=prod"]
```

maiamcc: I think we should support dict's as well, can easily enough make the type fuzzy and support strings later.

PR #1192, opened 2019-02-21
container: normalize all image names

nicks:
Hello @jazzdan,

Please review the following commits I made in branch nicks/normalize:

f5fd5172aebf279d6675375172863bb925c2aae7 (2019-02-21 12:43:07 -0500)
container: normalize all image names


nicks:
(in a follow-up, i think we need to go thru and make all the ref-parsing functions go thru a single place, but this will fix the immediate problem)

PR #1199, opened 2019-02-22
ignore: gitignore shouldn't be respected for build context

jazzdan:


nicks:
(i honestly don't know the answer to that question right now, it might be right to just push this change out and see what the reaction is for having separate build context + file change filters)

jazzdan:
Yeah it's something I'm debating in my head. IIRC one of the primary motivations for parsing + supporting `.gitignore` was to allow users to prevent spurious builds getting kicked off. So I'm inclined to go with this solution, but I'm worried that it's confusing that now the gitignore and file change filters are differently.

nicks:
i think an argument *for* is that people use git pushes to trigger CI builds. Therefore, gitignored files don't trigger CI builds. To match that, gitignored files shouldn't trigger Tilt builds?

this is again a good opportunity for a spec!

jazzdan:
@nicks that's true, but I've seen it done both ways. Hmm. Let me give this a think.

jazzdan:
I think I'm going to hold off on this for now until we have a better story around allowing users to specify Tilt-specific ignores, in the arguments to `docker_build` and `fast_build`.

PR #1202, opened 2019-02-22
engine: add ConfigsController test

landism:
also:
1. Make `FakeStore` a little more useful for testing subscribers (exposing dispatched actions)
2. Add a `FakeClock` class for the obvious
3. Add a `FakeTiltfileLoader` to avoid having to generate actual tilt configs in `Tiltfile.Load`'s callers (I'm not super excited about all the result `NewTiltfileLoader`s - open to suggestions)

### DIFF
--- a/internal/engine/configs_controller_test.go
+++ b/internal/engine/configs_controller_test.go
@@ -41,18 +41,17 @@ func TestConfigsController(t *testing.T) {
 
 func (f *ccFixture) run() ConfigsReloadedAction {
 	// configs_controller uses state.RelativeTiltfilePath, which is relative to wd
-	origDir, err := os.Getwd()
-	if err != nil {
-		f.T().Fatalf("error getting wd: %v", err)
-	}
-	err = os.Chdir(f.Path())
+	// sometimes the original directory was invalid (e.g., it was another test's temp dir, which was deleted,
+	// but not changed out of), and if it was already invalid, then let's not worry about it.
+	origDir, _ := os.Getwd()
+	err := os.Chdir(f.Path())
 	if err != nil {
 		f.T().Fatalf("error changing dir: %v", err)
 	}
 	defer func() {
-		// sometimes the original directory was invalid (e.g., it was another test's temp dir, which was deleted,
-		// but not changed out of), so changing back to the original directory will fail, and we probably don't care.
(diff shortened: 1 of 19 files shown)


internal/store/store.go
```diff
@@ -51,8 +51,16 @@ func NewStore(reducer Reducer, logActions LogActionsFlag) *Store {
 	}
 }
 
-func NewStoreForTesting() *Store {
-	return NewStore(EmptyReducer, false)
+func NewStoreForTesting() (*Store, <-chan Action) {
```
nicks: i'm not sure that an Action channel is the right API here. (the panic() seems like a good sign that the api is wrong)

it might eventually make more sense to use a MockStore with a getActions() method, like this project: https://github.com/dmitry-zaets/redux-mock-store

internal/engine/configs_controller_test.go
```diff
@@ -0,0 +1,100 @@
...
+	// configs_controller uses state.RelativeTiltfilePath, which is relative to wd
+	origDir, err := os.Getwd()
+	if err != nil {
+		f.T().Fatalf("error getting wd: %v", err)
+	}
+	err = os.Chdir(f.Path())
+	if err != nil {
+		f.T().Fatalf("error changing dir: %v", err)
+	}
+	defer func() {
```
nicks: should this be part of the TearDown method?

alternatively, maybe all this code should just be an option on TempDirFixture

internal/store/store.go
landism: kinda feeling https://www.jtolio.com/2016/03/go-channels-are-bad-and-you-should-feel-bad/ right now :) (and how often I've wanted growing buffered channels)

I've changed it to use a `func getActions() []Action`, but I'm not sure it feels better.

internal/engine/configs_controller_test.go
landism: > should this be part of the TearDown method?
I like it where it is, so that the state mutation and unmutation are in the same place. Also, so that the scope of the changed wd is minimized, and no other part of the test has to worry about what this function has done to global state.

I'd kind of prefer to fix our wd dependency rather than encoding it into TempDirFixture. I was tempted to do it for this PR, but I'm already 3 yaks deep.

PR #1203, opened 2019-02-22
tiltfile: filter_yaml also supports name, namespace, kind [ch1626]

maiamcc:
Hello @landism, @nicks,

Please review the following commits I made in branch maiamcc/filter-yaml-name-kind:

0867c1ef7fa83025f3eb41fb6aeb551de9d3532b (2019-02-22 17:41:58 -0500)
implement, tests

c08611729845979378bafaabd70a1bf6dae6f8d9 (2019-02-22 15:49:54 -0500)
stub out func

### DIFF
--- a/internal/k8s/entity.go
+++ b/internal/k8s/entity.go
@@ -203,6 +203,14 @@ func FilterByMetadataLabels(entities []K8sEntity, labels map[string]string) (pas
 	return Filter(entities, func(e K8sEntity) (bool, error) { return e.MatchesMetadataLabels(labels) })
 }
 
+func FilterByName(entities []K8sEntity, name string) (passing, rest []K8sEntity, err error) {
+	return Filter(entities, func(e K8sEntity) (bool, error) { return true, nil })
+}
+
+func FilterByKind(entities []K8sEntity, kind string) (passing, rest []K8sEntity, err error) {
+	return Filter(entities, func(e K8sEntity) (bool, error) { return true, nil })
+}
+
 func FilterByHasPodTemplateSpec(entities []K8sEntity) (passing, rest []K8sEntity, err error) {
 	return Filter(entities, func(e K8sEntity) (bool, error) {
 		templateSpecs, err := ExtractPodTemplateSpec(&e)
--- a/internal/k8s/entity.go
+++ b/internal/k8s/entity.go
@@ -4,10 +4,11 @@ import (
 	"fmt"
 	"net/url"
 	"reflect"
+	"strings"
 
 	"github.com/docker/distribution/reference"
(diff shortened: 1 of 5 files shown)


internal/k8s/entity.go
```diff
@@ -239,3 +252,29 @@ func FilterByMatchesPodTemplateSpec(withPodSpec K8sEntity, entities []K8sEntity)
 func (e K8sEntity) ResourceName() string {
 	return fmt.Sprintf("k8s%s-%s", e.Kind.Kind, e.Name())
 }
+
+func (e K8sEntity) HasName(name string) (bool, error) {
+	metas, err := extractObjectMetas(&e)
```
nicks: i don't think `extractObjectMetas` is the function you want here. This will give you all the object metas...like if you have a deployment, it will give you the PodTemplateSpec metas. and i think you only want the top-level one?

is there a reason you didn't do `e.Name() == name`?

internal/tiltfile/k8s.go
```diff
@@ -139,16 +143,56 @@ func (s *tiltfileState) filterYaml(thread *starlark.Thread, fn *starlark.Builtin
 		return nil, err
 	}
 
-	match, rest, err := k8s.FilterByMetadataLabels(entities, metaLabels)
-	if err != nil {
-		return nil, err
+	var match, allRest []k8s.K8sEntity
+	match = append(match, entities...)
```
nicks: is there a reason you copy entities into a new slice? maybe rename `entities` to `match`?

internal/tiltfile/k8s.go
```diff
@@ -139,16 +143,56 @@ func (s *tiltfileState) filterYaml(thread *starlark.Thread, fn *starlark.Builtin
...
+	}
+
+	if name != "" {
+		var rest []k8s.K8sEntity
+		match, rest, err = k8s.FilterByName(match, name)
+		if err != nil {
+			return nil, err
+		}
+
+		if len(rest) > 0 {
```
nicks: out of curiosity, can you skip the `if` check and just do a plain append?

internal/k8s/entity.go
maiamcc: ...no, no there isn't 😅

internal/tiltfile/k8s.go
maiamcc: ...some vague sense that i shouldn't overwrite the `entities` slice? no real reason, no.

internal/tiltfile/k8s.go
maiamcc: hmm maybe, let me look.

PR #1204, opened 2019-02-22
tiltfile: test for top level for loop

jazzdan:


internal/tiltfile/tiltfile_test.go
```diff
@@ -932,6 +932,20 @@ if True:
...
+	defer f.TearDown()
+
+	f.setupFoo()
+
+	f.file("Tiltfile", `
+for i in range(1, 3):
+	print(i)
+`)
+
+	f.load()
```
maiamcc: this... doesn't test anything? are you just trying to make sure it doesn't crash?

jazzdan: Yep!

maiamcc: ...okay 🤷‍♀️ can you leave a comment or rename the test to `TestTopLevelForLoopDoesntCrash` or something?

jazzdan: @maiamcc for context this isn't allowed normally in Starlark, and is behavior I [specifically enabled](https://github.com/windmilleng/tilt/blob/master/internal/tiltfile/tiltfile.go#L27) to allow top level if statements and I wanted to make sure that it also applied to the for loop control structure.

maiamcc: Oooh i see! (Worth noting that in a comment)

PR #1205, opened 2019-02-25
output: capture all stdout and stderr, redirect to file

jazzdan:
This prevents other libraries that we import (looking at you, Docker)
from printing to stdout or stderr and breaking the HUD.

In the future we might want to read from this file and handle things that are written to it. For now it just gets blackholed (though a user could look at it if they want to)

### DIFF
--- a/internal/build/image_builder.go
+++ b/internal/build/image_builder.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"os"
 	"regexp"
 	"strings"
 
@@ -15,7 +16,6 @@ import (
 	"github.com/windmilleng/tilt/internal/ignore"
 	"github.com/windmilleng/tilt/internal/logger"
 	"github.com/windmilleng/tilt/internal/model"
-	"github.com/windmilleng/tilt/internal/output"
 
 	"github.com/containerd/console"
 	"github.com/docker/cli/cli/command"
@@ -55,14 +55,14 @@ func DefaultImageBuilder(b *dockerImageBuilder) ImageBuilder {
 }
 
 func DefaultConsole() console.Console {
-	out := output.File()
+	out := os.Stdout
 	c, _ := console.ConsoleFromFile(out)
 
 	return c
 }
 
 func DefaultOut() io.Writer {
-	return output.File()
+	return os.Stdout
 }
 
 var _ ImageBuilder = &dockerImageBuilder{}
--- a/internal/build/image_builder.go
+++ b/internal/build/image_builder.go
@@ -5,7 +5,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
-	"os"
 	"regexp"
 	"strings"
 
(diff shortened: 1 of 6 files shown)


internal/build/image_builder.go
```diff
@@ -55,14 +55,14 @@ func DefaultImageBuilder(b *dockerImageBuilder) ImageBuilder {
 }
 
 func DefaultConsole() console.Console {
-	out := os.Stdout
+	out := output.File()
```
nicks: i don't think this will work like you expect? Notice that DefaultConsole is called before `output.CaptureAll`, so the output will not get captured here.

As an aside, this PR is a pretty good example of the order-of-initialization bugs that come up when you use global state, and which dependency injection tries to resolve!

I would probably change this PR by
1) Remove the global state in output.CaptureAll
2) Make output.CaptureAll return the new output writer, so that you can use it as a provider
3) Add output.CaptureAll to your DI graph (instead of DefaultOut), to ensure it gets initialized first

internal/output/output.go
```diff
@@ -0,0 +1,27 @@
...
+	"os"
+)
+
+var outputFile *os.File
+
+func CaptureAll() {
+	if outputFile != nil {
+		return
+	}
+	tmpfile, err := ioutil.TempFile("", "tilt.stdout.*")
```
nicks: shouldn't this write to logger by default, like other things?

internal/build/image_builder.go
jazzdan: That's a good idea, I'll do that!

jazzdan:
I changed it to use wire @nicks :)

internal/output/output.go
```diff
@@ -0,0 +1,20 @@
...
+func CaptureAll() io.Writer {
+	tmpfile, err := ioutil.TempFile("", "tilt.output.*")
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	os.Stdout = tmpfile
+	os.Stderr = tmpfile
+
+	return tmpfile
```
nicks: oh! i remember my comment now. I meant that CaptureAll could return a different writer that didn't write to these temp files

internal/engine/wire.go
```diff
@@ -20,7 +21,7 @@ import (
 var DeployerBaseWireSet = wire.NewSet(
 	// dockerImageBuilder ( = ImageBuilder)
 	build.DefaultConsole,
```
nicks: did you also intend to change DefaultConsole to not use os.Stdout, like you did in the original pr?

internal/output/output.go
jazzdan: Aha I see!

internal/engine/wire.go
jazzdan: It looks like Console also _has_ to take an `*os.File` and doesn't accept an `io.Writer` >.>

internal/output/output.go
jazzdan: I think it's desirable to write to an actual file and keep it around. That way if there's an error that being swallowed we can still find it somewhere, even on a production build on a users machine.

PR #1209, opened 2019-02-25
analytics: report tiltfile builtin invocation counts

landism:
Any time we load the tiltfile, report stats on how many times each builtin was called, as a catchall way of telling which builtins are being used.

(Pretty much summed up by [this](https://github.com/windmilleng/tilt/compare/mlandis/ch1659/report-tiltfile-builtin-usage?expand=1#diff-c9f495b4db5477504f7005e67ff97d75R47) bit of unit test)

### DIFF
--- a/internal/tiltfile/tiltfile.go
+++ b/internal/tiltfile/tiltfile.go
@@ -7,13 +7,15 @@ import (
 	"log"
 	"os"
 	"path/filepath"
+	"strconv"
 
 	"github.com/pkg/errors"
 	"go.starlark.net/resolve"
 	"go.starlark.net/starlark"
 
-	"github.com/windmilleng/tilt/internal/k8s"
+	"github.com/windmilleng/wmclient/pkg/analytics"
 
+	"github.com/windmilleng/tilt/internal/k8s"
 	"github.com/windmilleng/tilt/internal/model"
 	"github.com/windmilleng/tilt/internal/ospath"
 )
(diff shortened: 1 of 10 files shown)


internal/tiltfile/tiltfile.go
```diff
@@ -28,23 +28,26 @@ func init() {
 }
 
 type TiltfileLoader interface {
-	Load(ctx context.Context, filename string, matching map[string]bool, logs io.Writer) (manifests []model.Manifest, global model.Manifest, configFiles []string, warnings []string, err error)
+	Load(ctx context.Context, filename string, matching map[string]bool, logs io.Writer) (manifests []model.Manifest, global model.Manifest, configFiles []string, warnings []string, builtinCallCounts map[string]int, err error)
```
nicks: this api seems increasingly unmanageable.

can we dependency-inject TiltfileLoader, and make it call analytics directly?

landism: I was thinking it'd be nice to not need to worry about analytics just to parse a file. I guess we kind of have to either way, either in the ctor or the return value.

How about making Load return a `ProcessedTiltfile` and pulling all the return values into that? It's currently pretty ugly even without `builtinCallCounts`

nicks: i would be fine with ProcessedTiltfile! 

But the builtinCallCounts would still seem like a weird thing to return from this function, since it's still very coupled to the analytics interface.

PR #1210, opened 2019-02-25
container: centralize ref-parsing

nicks:
Hello @maiamcc,

Please review the following commits I made in branch nicks/parse:

b07103888ea9a70aee09ce0df864f95a04d770f2 (2019-02-25 17:09:22 -0500)
container: centralize ref-parsing


nicks:
yes! there's a long comment about how these things fit together here that you may be interested in reading

https://github.com/containerd/containerd/blob/master/reference/reference.go#L39

it expressed the current state of the world better than i could!

PR #1218, opened 2019-02-27
engine: add `custom_build` and `kubectl exec` sync

jazzdan:
## What's in this PR?

I split this PR in to two parts main commits, excluding a vendor commit:
1. `custom_build`

`custom_build` is a Tiltfile function that adds a new model (`CustomBuild`) that is used by a new builder (`CustomBuilder`). This builder knows how to invoke the command specified in `custom_build` and look for a resulting image in the image repository.

2. "synclet" deploys via `kubectl exec`.

We can't talk to the Docker socket in other container runtimes, like CRI-O. In order to do sync then we need to sync files the old fashioned way: via `kubectl exec`. This is slower, but a good enough solution for now. The `SyncletBuildAndDeployer` currently looks at the container run time and uses the existing synclet setup for that. Otherwise it uses this new `kubectl exec` syncing strategy.

## What's left to do
I'm pretty happy with the test coverage around `custom_build`. I'd like to add an integration test for the `kubectl exec` stuff, or maybe more extensive unit test coverage.


nicks:
could we split this up into separate PRs? 

I'm a little bit concerned that this is going to take so long to review that it will create lots of merge conflicts and frustrate everyone on the team.

A good division of PRs might be:
PR #1) data model additions
PR #2) k8s vendoring
PR #3) engine implementation
PR #4) Tiltfile changes

this should easily be separable, because we've done a good job separating the responsibility of these packages.

nicks:
(also, I suspect Maia will be a better 2nd reviewer than me, since she has the most context on the different Build strategy divisions)

jazzdan:
Yeah if we feel that this is too much I'm happy to split out. # 3 in that plan is going to be by _far_ the largest PR.

I'll get started on splitting it out now.

jazzdan:
Closing in favor of split out PRs.

PR #1220, opened 2019-02-27
tiltfile: include sources in k8s yaml parse errors

landism:
before:
```
       Error: error converting YAML to JSON: yaml: line 3:
       could not find expected ':'
```

after:
```
       Error: Error reading yaml from cmd: 'm4 -Dvarowner="matt"
       -Dvarimgname="gcr.io/windmill-public-containers/servantes/doggos"
       "deploy/doggos.yaml"': error converting YAML to JSON: yaml: line 3:
       could not find expected ':'
```

### DIFF
--- a/internal/tiltfile/files.go
+++ b/internal/tiltfile/files.go
@@ -200,17 +200,18 @@ func (s *tiltfileState) skylarkReadFile(thread *starlark.Thread, fn *starlark.Bu
 		return nil, err
 	}
 
-	return newBlob(string(bs)), nil
+	return newBlob(string(bs), fmt.Sprintf("file: %s", path)), nil
 }
 
 type blob struct {
-	text string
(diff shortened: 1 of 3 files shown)


internal/tiltfile/files.go
```diff
@@ -200,17 +200,18 @@ func (s *tiltfileState) skylarkReadFile(thread *starlark.Thread, fn *starlark.Bu
 		return nil, err
 	}
 
-	return newBlob(string(bs)), nil
+	return newBlob(string(bs), fmt.Sprintf("file: %s", path)), nil
```
nicks: path is probably not right here...i think you want to pull the path off `p`, in case path is a non-string?

landism: good catch, thanks!

PR #1226, opened 2019-02-27
model: add CustomBuild

jazzdan:


internal/model/custom_build_test.go
```diff
@@ -0,0 +1,63 @@
...
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/windmilleng/tilt/internal/container"
+)
+
+func TestMaybeFastBuildInfo(t *testing.T) {
+	fb := FastBuild{
+		BaseDockerfile: "FROM alpine",
```
landism: I thought the whole point of CustomBuild was that we aren't using Docker, so I'm kind of surprised to see this here. What did I misunderstand?

internal/model/docker_info.go
```diff
@@ -188,4 +215,13 @@ type FastBuild struct {
 
 func (FastBuild) buildDetails() {}
 
+type CustomBuild struct {
+	Command string
+	Deps    []string
```
landism: I was confused by this, could we use a comment or rename or a type other than `string` - `DependentTargets`? `FileDeps`? `[]LocalPath`?

internal/model/custom_build_test.go
jazzdan: A `CustomBuild` will be able to have a `FastBuild` attached to it. Let's look at two examples:

```
custom_build(
  'gcr.io/windmill-public-containers/servantes/hypothesizer',
  'docker build -t $TAG hypothesizer',
  ['hypothesizer'])
```

This is a normal custom build, it should function exactly like the equivalent `docker_build` today.

```
hfb = custom_build(
  'gcr.io/windmill-public-containers/servantes/hypothesizer',
  'docker build -t $TAG hypothesizer',
  ['hypothesizer']).add_fast_build()
hfb.add(repo.path('hypothesizer'), '/app')
hfb.run('cd /app && pip install -r requirements.txt', trigger='hypothesizer/requirements.txt')
hfb.hot_reload()
```
This is a custom build with a fast build. The fast build will use as its base the image that `custom_build` expects to be present in the registry after the `Command` has completed (in this case `gcr.io/windmill-public-containers/servantes/hypothesizer`). It is otherwise a normal fast build.

PR #1228, opened 2019-02-27
engine: custom_build and kubectl exec syncer

jazzdan:


internal/engine/build_and_deployer.go
```diff
@@ -54,7 +54,7 @@ func (composite *CompositeBuildAndDeployer) BuildAndDeploy(ctx context.Context,
 			logger.Get(ctx).Debugf("(expected error) falling back to next build and deploy method "+
 				"after error: %v", err)
 		} else {
-			logger.Get(ctx).Verbosef("falling back to next build and deploy method "+
+			logger.Get(ctx).Infof("falling back to next build and deploy method "+
```
landism: Was this intended to be left at Info? I think we had general agreement this was more distracting than useful.

internal/k8s/container.go
```diff
@@ -120,10 +120,11 @@ func ContainerIDFromContainerStatus(status v1.ContainerStatus) (container.ID, er
 		return "", nil
 	}
 
-	if !strings.HasPrefix(id, ContainerIDPrefix) {
+	components := strings.SplitN(id, "://", 2)
+	if len(components) != 2 {
 		return "", fmt.Errorf("Malformed container ID: %s", id)
 	}
-	return container.ID(id[len(ContainerIDPrefix):]), nil
+	return container.ID(components[1]), nil
```
landism: Just out of curiosity - do we really need to drop the prefix or care what this id looks like? does this open us up to conflicts in a cluster that has, e.g., both cri-o and docker? Is it weird to assume that k8s has this prefix there for a reason?

(though this seems pretttty unlikely to bite us, so probably isn't worth spending much time on)

internal/engine/synclet_build_and_deployer.go
```diff
@@ -124,3 +134,63 @@ func (sbd *SyncletBuildAndDeployer) updateViaSynclet(ctx context.Context,
...
+	w := l.Writer(logger.InfoLvl)
+
+	if len(filesToDelete) > 0 {
+		filesToShow := filesToDelete
+		if len(filesToShow) > 5 {
+			filesToShow = append([]string(nil), filesToDelete[0:5]...)
+			filesToShow = append(filesToShow, "...")
+		}
+
+		l.Infof("removing %v files %v", len(filesToDelete), filesToShow)
```
landism: the logging in this file seems like it might be a bit much, and more than what we do for other builders - do you have feelings on this?

internal/build/custom_builder.go
```diff
@@ -0,0 +1,86 @@
...
+	tmpTag := fmt.Sprintf("tilt-build-%d", b.clock.Now().Unix())
+	result, err := reference.WithTag(ref, tmpTag)
+	if err != nil {
+		return nil, err
+	}
+
+	cmd := exec.CommandContext(ctx, "sh", "-c", command)
+
+	l := logger.Get(ctx)
+	l.Infof("Custom Build: Injecting Environment Variables")
```
landism: this looks like potentially leftover printf debugging

internal/docker/client.go
```diff
@@ -93,35 +96,44 @@ type Cli struct {
...
+			func(key string) string {
 				val := os.Getenv(key)
 				if val == "" && key == "DOCKER_HOST" {
 					return microK8sDockerHost
 				}
 				return val
-			}, nil
-		}
+			},
+			[]string{"DOCKER_HOST=" + microK8sDockerHost},
```
landism: it seems kinda weird that the func returns stuff from `os.Getenv` but the `[]string` doesn't. Maybe it'd make sense to just throw all of `os.Getenv` in there?

(also, unrelated to this PR, but do you know if it's intentional that `os.Getenv` isn't used in minikube's func?)

internal/build/custom_builder.go
dbentley: If we don't have this, it's hard to understand what's happening when you're trying to set up your custom build.

E.g., you run a command, but it doesn't leave a docker image on your docker daemon because we set DOCKER_HOST, and you're like WTF.

internal/docker/client.go
dbentley: Build Injections is what needs to be injected in addition to the current environment.

No idea about the os.Getenv use/non-use (and added a TODO above)

internal/engine/build_and_deployer.go
dbentley: There are now two cases:
a) expected error, at DebugF
b) unexpected error, which I think is worth upping to Info

But don't feel super strongly.

internal/engine/synclet_build_and_deployer.go
dbentley: The in-container building is weirder, and so I think worth being more explicit, so that people can debug when things go wrong instead of blaming us.

internal/k8s/container.go
dbentley: I'm not sure if Docker accepts docker://<id> or just <id>. I think you need to split it out before sending it (but could be wrong), but yeah, we could keep it here and split it later, but, *shrug*

internal/engine/synclet_build_and_deployer.go
jazzdan: Yeah considering this is an alpha feature I feel OK leaving this in. We can always remove it later.

PR #1233, opened 2019-02-28
globalyaml: apply namespaces first

landism:
Problem:
If the yaml contains both namespaces and objects in those namespaces, trying to apply the objects in the namespaces before the namespaces themselves will result in failure (because the namespaces don't exist)

Solution:
sort namespaces to the front of global yaml
This doesn't fully solve the problem, since global yaml gets applied in parallel with resourced entities, and those resourced entities could be in a namespace specified in global yaml, but that's a problem to tackle separately.

### DIFF
--- a/internal/engine/global_yaml_buildcontroller_test.go
+++ b/internal/engine/global_yaml_buildcontroller_test.go
@@ -3,6 +3,8 @@ package engine
 import (
 	"testing"
 
+	"github.com/windmilleng/tilt/internal/yaml"
+
 	"github.com/pkg/errors"
 	"github.com/stretchr/testify/assert"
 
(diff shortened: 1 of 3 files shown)


internal/k8s/testyaml/testyaml.go
```diff
@@ -1112,6 +1115,16 @@ spec:
   replicas: 1
 `
 
+const MyNamespaceYAML = `apiVersion: v1
+kind: Namespace
+metadata:
+  name: mynamespace
+`
+
+func ConcatYAML(y1, y2 string) string {
```
maiamcc: this exists in `yaml/utils.go`

internal/engine/global_yaml_buildcontroller_test.go
```diff
@@ -91,6 +92,21 @@ func TestGlobalYamlFailUpsert(t *testing.T) {
...
+	st := newTestingStoreWithGlobalYAML(yaml)
+	bc := newGlobalYamlBuildControllerForTest(testyaml.SecretYaml)
+
+	bc.OnChange(output.CtxForTest(), st)
+
+	entities, err := k8s.ParseYAMLFromString(bc.k8sClient.Yaml)
+	if err != nil {
+		t.Fatal(err)
+	}
+	assert.Equal(t, "mynamespace", entities[0].Name())
```
maiamcc: if `len(entities) != 2`, we'll get hard-to-debug index errors. consider:
```
if assert.Equal(t, 2, len(entities)) {
    assert.Equal(...)
    ...
}
```

maiamcc:
err * noice

PR #1234, opened 2019-02-28
integration: support optional GOPATH

jazzdan:


integration/fixture_test.go
```diff
@@ -162,7 +162,10 @@ func (f *fixture) dockerKillAll(name string) {
 func (f *fixture) tiltCmd(tiltArgs []string, outWriter io.Writer) *exec.Cmd {
 	outWriter = io.MultiWriter(f.logs, outWriter)
 
-	gopath := os.Getenv("GOPATH")
+	gopath, ok := os.LookupEnv("GOPATH")
+	if !ok {
+		gopath = build.Default.GOPATH
```
nicks: Oh I didn't know about this! Should we just be using it always?

jazzdan: Probably not, it wouldn't respect the GOPATH env if set it just supports the correct default for your environment (ie `$HOME/go` on unixes, and probably something else on Windows, etc)

PR #1235, opened 2019-02-28
integration: custom_build integration test

jazzdan:


jazzdan:
Wow this was actually ... pretty easy.

integration/oneup_custom_test.go
```diff
@@ -0,0 +1,10 @@
+//+build integration
+
+package integration
+
+import "testing"
+
+func TestOneUpCustom(t *testing.T) {
+	f := newFixture(t, "oneup_custom")
```
nicks: could you also add a CurlUntil like in the other integration tests to ensure the server comes up?

integration/oneup_custom/Tiltfile
```diff
@@ -0,0 +1,9 @@
+# -*- mode: Python -*-
+
+custom_build(
+  'gcr.io/windmill-test-containers/integration/oneup-custom',
+  'docker build -t $TAG .',
```
nicks: out of curiosity, how does this image get pushed?

nicks: (a better test might be to use a non-docker image builder...)

integration/oneup_custom_test.go
jazzdan: Whoops, I added that to the wrong file locally and didn't commit it. 😬

integration/fixture_test.go
```diff
@@ -220,7 +221,10 @@ func (f *fixture) AllPodsReady(ctx context.Context, selector string) (bool, stri
 	cmd := exec.Command("kubectl", "get", "pods",
 		namespaceFlag, "--selector="+selector, "-o=template",
 		"--template", "{{range .items}}{{.metadata.name}} {{.status.phase}}{{println}}{{end}}")
-	out, err := cmd.Output()
+	log.Printf("CMD: %+v", cmd.Args)
+	out, err := cmd.CombinedOutput()
```
jazzdan: whoops, will remove these

integration/oneup_custom/Tiltfile
jazzdan: So the builder returns a ref that the image_build_and_deployer pushes, just like the regular docker_build. :)

PR #1236, opened 2019-02-28
tiltfile: don't pull dockerignores for repo root

maiamcc:
Hello @jazzdan,

Please review the following commits I made in branch maiamcc/correct-dockerignore-behavior:

1e63990a5e36cec9ae624c5b4bd384fe5a3acd89 (2019-02-28 16:47:48 -0500)
tiltfile: don't pull dockerignores for repo root
from Docker docs:
> Before the docker CLI sends the context to the docker daemon, it looks for a file named .dockerignore **in the root directory of the context**.

we should not check the repo root for .dockerignores, it's different from the behavior of docker

### DIFF
--- a/internal/tiltfile/docker.go
+++ b/internal/tiltfile/docker.go
@@ -513,6 +513,16 @@ func (s *tiltfileState) dockerignoresForImage(image *dockerImage) []model.Docker
 
 	for _, m := range image.mounts {
 		paths = append(paths, m.src.path)
+
+		// NOTE(maia): this doesn't reflect the behavior of Docker, which only
+		// looks in the build context for the .dockerignore file. Leaving it
+		// for now, though, for fastbuild cases where .dockerignore doesn't
+		// live in the user's mount(s) (e.g. user only mounts several specific
+		// files, not a directory containing the .dockerignore).
+		repo := m.src.repo
+		if repo != nil {
+			paths = append(paths, repo.basePath)
+		}
 	}
 	paths = append(paths, image.staticBuildPath.path)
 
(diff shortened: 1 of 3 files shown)


jazzdan:
What's the context for this change?

maiamcc:
ah sorry -- I was trying to fix `TestFastBuildDockerignoreRoot` as part of #1240 by taking out the call to `local_git_repo` and just passing a string to `fb.add` -- and when i did, the test failed. digging in, it's because we check the `GitRepo` for `.dockerignore`. talking to matt about this tho, that's DIFFERENT from the behavior of docker, so instead of changing tilt to be able to find `.dockerignore`'s in the wrong places even without a `GitRepo`, i just took out the wrong behavior and the test.

jazzdan:
Does this mean that fast builds no longer respect dockerignores if you don't use a Repo?

maiamcc:
...fuck. let me check.

maiamcc:
```
func (s *tiltfileState) dockerignoresForImage(image *dockerImage) []model.Dockerignore {
	var paths []string

	for _, m := range image.mounts {
		paths = append(paths, m.src.path)
        ...
}```
soooo we'll respect the dockerignore if it lives at a path that you mounted. which uhhh. i feel like will be MANY cases but not all...

maiamcc:
So, i don't want to miss a .dockerignore b/c say, a user only mounted specific files and not the directory containing the .dockerignore... but using the .dockerignore at the ROOT of the repo isn't the right behavior either (i.e. it doesn't match the docker spec).

how would you feel about being able to specify `fb.dockerignore('path/to/.dockerignore')` if the `.dockerignore` isn't included in a mounted directory?

jazzdan:
That could work, but I think we should consider and scope out an epic of work for reworking the way we do ignores in Tilt rather than rush in to a solution.

maiamcc:
Oh yeah, sorry, wasn't suggesting implementing it now.

On further reflection, I'd rather leave users an out if their dockerignore is in a weird place, even if it teeechnically goes against the behavior of docker. I'll put the code back and leave a note.

PR #1239, opened 2019-02-28
engine: prioritize building no-build resources first

landism:


internal/engine/buildcontroller_test.go
```diff
@@ -137,3 +139,45 @@ func TestBuildControllerManualTrigger(t *testing.T) {
 	assert.Equal(t, []string{f.JoinPath("main.go")}, call.oneState().FilesChanged())
 	f.waitForCompletedBuildCount(2)
 }
+
+// any manifests with no built images should be deployed before any manifests that have built images
```
maiamcc:
```suggestion
// any manifests without image targets should be deployed before any manifests WITH image targets
```

internal/engine/buildcontroller.go
```diff
@@ -78,6 +83,50 @@ func nextTargetToBuild(state store.EngineState) *store.ManifestTarget {
 	return choice
 }
 
+type noBuildManifestsFirst struct {
+	mts             []*store.ManifestTarget
+	origIndexByName map[string]int
```
nicks: I'm surprised you didn't make the key a ManifestName

internal/engine/buildcontroller.go
```diff
@@ -34,9 +34,14 @@ func NewBuildController(b BuildAndDeployer) *BuildController {
 
 // Algorithm to choose a manifest to build next.
 func nextTargetToBuild(state store.EngineState) *store.ManifestTarget {
+	// put no-build manifests first since they're more likely to be
+	// 1. fast and 2. dependencies of other services (e.g., redis)
+	targets := append([]*store.ManifestTarget{}, state.Targets()...)
```
nicks: Does the Targets() method already make a new array so you don't need to clone here?

internal/engine/buildcontroller.go
landism: me too!

internal/engine/buildcontroller.go
landism: Is that a detail that's worth relying on? I haven't seen much in the way of memory ownership semantics in go code, and worry mildly some later refactor might have `state.Targets()` returning the raw slice and now we're mutating `state`. I've often felt unsure whether it's appropriate to make a copy of a slice. Maybe I've been in languages with immutable collections for too long.

This change is already reducing the number of copies from 3 to 2, so I'm not too concerned about the performance implications for this change in particular, but I'm interested in the approach in general.

PR #1242, opened 2019-03-01
engine: skip images that have already been built and don't have updates

nicks:
Hello @maiamcc, @jazzdan,

Please review the following commits I made in branch nicks/fastbuild:

e1961ac645e8381c1359fc81e7e093f7918ea4ed (2019-02-28 19:34:19 -0800)
engine: skip images that have already been built and don't have updates

### DIFF
--- a/internal/engine/build_and_deployer_test.go
+++ b/internal/engine/build_and_deployer_test.go
@@ -236,7 +236,9 @@ func TestIncrementalBuildKilled(t *testing.T) {
 	f := newBDFixture(t, k8s.EnvDockerDesktop)
 	defer f.TearDown()
 
-	bs := resultToStateSet(alreadyBuiltSet, nil, f.deployInfo())
+	changed := f.WriteFile("a.txt", "a")
+
+	bs := resultToStateSet(alreadyBuiltSet, []string{changed}, f.deployInfo())
 	f.docker.ExecErrorToThrow = docker.ExitError{ExitCode: build.TaskKillExitCode}
 
 	manifest := NewSanchoFastBuildManifest(f)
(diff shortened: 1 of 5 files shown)


internal/engine/build_and_deployer_test.go
```diff
@@ -236,7 +236,9 @@ func TestIncrementalBuildKilled(t *testing.T) {
 	f := newBDFixture(t, k8s.EnvDockerDesktop)
 	defer f.TearDown()
 
-	bs := resultToStateSet(alreadyBuiltSet, nil, f.deployInfo())
+	f.WriteFile("a.txt", "a")
```
maiamcc: comment what's going on here?

internal/engine/build_and_deployer_test.go
```diff
@@ -259,7 +261,8 @@ func TestFallBackToImageDeploy(t *testing.T) {
 
 	f.docker.ExecErrorToThrow = errors.New("some random error")
 
-	bs := resultToStateSet(alreadyBuiltSet, nil, f.deployInfo())
+	f.WriteFile("a.txt", "a")
```
maiamcc: comment what's going on here?

internal/store/build_result.go
```diff
@@ -148,6 +148,14 @@ func (b BuildState) HasImage() bool {
 	return b.LastResult.HasImage()
 }
 
+// Whether the image represented by this state needs to be built.
+// If the image has already been built, and no files have been
+// changed since then, then we can re-use the previous result.
+func (b BuildState) NeedsImageBuild() bool {
+	alreadyBuilt := b.LastResult.HasImage()
+	return !alreadyBuilt || len(b.FilesChangedSet) > 0
```
maiamcc: hmm will this respect config changes? I believe "this manifest is different now b/c of config file changes" is tracked somewhere besides `FilesChangedSet`.

internal/engine/build_and_deployer_test.go
nicks: i changed it so that WriteFile returns a path, and then the path passed further down. ergonomically, that should make it easier to connect the two statements.

internal/engine/build_and_deployer_test.go
nicks: ditto

internal/store/build_result.go
nicks: yes.

The BuildState needs to be the BuildState for this manifest. If the config files changed the manifest, then the BuildState is reset. see:

https://github.com/windmilleng/tilt/blob/master/internal/engine/upper.go#L480

(Note that other things use the FilesChangedSet, so if this was broken, this wouldn't be the first place it was breaking)

PR #1243, opened 2019-03-01
engine: tiltfile logs also written to global logs

maiamcc:
I remember it was originally an intitional design choice to not have the
Tiltfile log to the global log pane, but it confuses me every damn time,
and [some users too](https://kubernetes.slack.com/archives/CESBL84MV/p1551395345022900).

I propose that until Tiltfile output gets unweildy, we display it in the global
log with everything else -- how do y'all feel about that?

### DIFF
--- a/internal/engine/upper.go
+++ b/internal/engine/upper.go
@@ -9,6 +9,7 @@ import (
 
 	opentracing "github.com/opentracing/opentracing-go"
 	"github.com/pkg/errors"
+	"github.com/windmilleng/tilt/internal/tiltfile"
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/labels"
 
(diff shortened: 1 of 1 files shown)


internal/engine/upper.go
```diff
@@ -875,6 +875,7 @@ func handleDockerComposeLogAction(state *store.EngineState, action DockerCompose
 	ms.ResourceState = dcState.WithCurrentLog(append(dcState.CurrentLog, action.Log...))
 }
 
-func handleTiltfileLogAction(state *store.EngineState, action TiltfileLogAction) {
+func handleTiltfileLogAction(ctx context.Context, state *store.EngineState, action TiltfileLogAction) {
 	state.CurrentTiltfileBuild.Log = model.AppendLog(state.CurrentTiltfileBuild.Log, action.Log)
+	logger.Get(ctx).Infof("[TILTFILE] " + string(action.Log))
```
jazzdan: Should the name of the resource in square brackets match the casing of the resource as it appears in the resource list in the UI?

maiamcc: ha sure

PR #1245, opened 2019-03-01
roadmap: add multi-container pods

maiamcc:
Hello @nicks, @landism,

Please review the following commits I made in branch maiamcc/multi-container-roadmap:

2eeb7f7aad0362b9a8faf2ed270621528399eb08 (2019-03-01 16:52:36 -0500)
roadmap: add multi-container pods


nicks:
tho this seems weird filing as feature roadmap item vs filing it as a bug that we intend to fix. the advantage of filing it as a bug is that people can follow it, ask questions, get notified when it's closed, etc

maiamcc:
Fair enough -- also expect it to be at least hackily done by EOW, so I'll just not bother with this PR.

PR #1249, opened 2019-03-04
support fast build container updates with image deps

nicks:
Hello @maiamcc, @jazzdan,

Please review the following commits I made in branch nicks/fastbuild2:

48bea7425c8beba8d8e495e441da7ab993618a21 (2019-03-03 19:54:59 -0800)
support fast build container updates with image deps

### DIFF
--- a/internal/engine/build_and_deployer_test.go
+++ b/internal/engine/build_and_deployer_test.go
@@ -116,7 +116,8 @@ func TestNamespaceGKE(t *testing.T) {
 	deployInfo := f.deployInfo()
 	deployInfo.Namespace = "sancho-ns"
 
-	bs := resultToStateSet(result, nil, deployInfo)
+	changed := f.WriteFile("a.txt", "a")
+	bs := resultToStateSet(result, []string{changed}, deployInfo)
 	result, err = f.bd.BuildAndDeploy(f.ctx, f.st, buildTargets(manifest), bs)
 	if err != nil {
 		t.Fatal(err)
@@ -129,9 +130,10 @@ func TestContainerBuildLocal(t *testing.T) {
 	f := newBDFixture(t, k8s.EnvDockerDesktop)
 	defer f.TearDown()
 
+	changed := f.WriteFile("a.txt", "a")
 	manifest := NewSanchoFastBuildManifest(f)
 	targets := buildTargets(manifest)
-	bs := resultToStateSet(alreadyBuiltSet, nil, f.deployInfo())
+	bs := resultToStateSet(alreadyBuiltSet, []string{changed}, f.deployInfo())
 	result, err := f.bd.BuildAndDeploy(f.ctx, f.st, targets, bs)
 	if err != nil {
 		t.Fatal(err)
@@ -152,14 +154,17 @@ func TestContainerBuildLocal(t *testing.T) {
 	f.assertContainerRestarts(1)
 
 	id := manifest.ImageTargetAt(0).ID()
-	assert.Equal(t, k8s.MagicTestContainerID, result[id].ContainerID.String())
+	_, hasResult := result[id]
+	assert.True(t, hasResult)
+	assert.Equal(t, k8s.MagicTestContainerID, result.AsOneResult().ContainerID.String())
 }
 
 func TestContainerBuildSynclet(t *testing.T) {
 	f := newBDFixture(t, k8s.EnvGKE)
 	defer f.TearDown()
 
-	bs := resultToStateSet(alreadyBuiltSet, nil, f.deployInfo())
+	changed := f.WriteFile("a.txt", "a")
+	bs := resultToStateSet(alreadyBuiltSet, []string{changed}, f.deployInfo())
 	manifest := NewSanchoFastBuildManifest(f)
 	targets := buildTargets(manifest)
 	result, err := f.bd.BuildAndDeploy(f.ctx, f.st, targets, bs)
@@ -177,8 +182,7 @@ func TestContainerBuildSynclet(t *testing.T) {
 		t.Errorf("Expected 1 synclet containerUpdate, actual: %d", f.sCli.UpdateContainerCount)
 	}
 
-	id := manifest.ImageTargetAt(0).ID()
-	assert.Equal(t, k8s.MagicTestContainerID, result[id].ContainerID.String())
+	assert.Equal(t, k8s.MagicTestContainerID, result.AsOneResult().ContainerID.String())
 	assert.False(t, f.sCli.UpdateContainerHotReload)
 }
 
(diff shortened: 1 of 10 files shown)


nicks:
(this is another pr i put together on the plane when trying to get multi-stage builds working better...most of this is unifying the two container-based build and deployers a bit better)

internal/engine/local_container_build_and_deployer.go
```diff
@@ -30,22 +30,26 @@ func NewLocalContainerBuildAndDeployer(cu *build.ContainerUpdater,
 }
 
 func (cbd *LocalContainerBuildAndDeployer) BuildAndDeploy(ctx context.Context, st store.RStore, specs []model.TargetSpec, stateSet store.BuildStateSet) (store.BuildResultSet, error) {
-	iTargets, kTargets := extractImageAndK8sTargets(specs)
-	if len(kTargets) != 1 || len(iTargets) != 1 {
-		return store.BuildResultSet{}, RedirectToNextBuilderf(
-			"LocalContainerBuildAndDeployer requires exactly one image spec and one k8s deploy spec")
+	iTargets, err := extractImageTargetsForLiveUpdates(specs, stateSet)
```
nicks: in theory, this build and deployer should work fine with docker-compose, but i haven't verified it all yet manually.

internal/engine/extractors.go
```diff
@@ -16,3 +19,64 @@ func extractImageAndK8sTargets(specs []model.TargetSpec) (iTargets []model.Image
 	}
 	return iTargets, kTargets
 }
+
+// Check to see if we have k8s targets.
+func extractK8sTargets(specs []model.TargetSpec) []model.K8sTarget {
+	kTargets := make([]model.K8sTarget, 0)
```
jazzdan: Why do we bother setting the length to 0 here?

internal/engine/build_and_deployer_test.go
```diff
@@ -486,6 +493,38 @@ func TestCustomBuild(t *testing.T) {
...
+	assert.Equal(t, 0, f.docker.BuildCount)
+	assert.Equal(t, 0, f.docker.PushCount)
+	assert.Equal(t, 1, f.docker.CopyCount)
+	assert.Equal(t, 1, len(f.docker.ExecCalls))
+	f.assertContainerRestarts(1)
+
+	// The BuildComplete action handler expects to get exactly one result
+	_, hasResult0 := result[manifest.ImageTargetAt(0).ID()]
+	assert.False(t, hasResult0)
+	_, hasResult1 := result[manifest.ImageTargetAt(1).ID()]
```
maiamcc: 🙌 

but why not `res, hasResult1 := ...` and use `res` below instead of calling `AsOneResult`?

internal/engine/extractors.go
```diff
@@ -16,3 +19,64 @@ func extractImageAndK8sTargets(specs []model.TargetSpec) (iTargets []model.Image
...
+		if !state.NeedsImageBuild() {
+			continue
+		}
+
+		fbInfo := iTarget.MaybeFastBuildInfo()
+		if fbInfo == nil {
+			return nil, RedirectToNextBuilderf("In-place build only supports FastBuilds")
+		}
+
+		// Otherwise, manifest has already been deployed; try to update in the running container
```
maiamcc: sigh, we have name collision with deployInfo = info needed to deploy vs. deployInfo = info about the existing deployment :/ comment suggestion below to clarify a little
```suggestion
		// Otherwise, manifest has already been deployed; IF we have info about the running container, update in place
```

(you might also clarify the err message at ln 63)

internal/engine/image_build_and_deployer.go
```diff
@@ -225,19 +225,7 @@ func (ibd *ImageBuildAndDeployer) deploy(ctx context.Context, st store.RStore, p
 // We can also skip the push of the image if it isn't used
 // in any k8s resources! (e.g., it's consumed by another image).
```
maiamcc: I still don't think this comment applies to the whole function, and should therefore be either a. more complete or b. live INSIDE the func rather than at the top level

internal/engine/extractors.go
```diff
@@ -16,3 +19,64 @@ func extractImageAndK8sTargets(specs []model.TargetSpec) (iTargets []model.Image
...
+		if !ok {
+			continue
+		}
+		kTargets = append(kTargets, t)
+	}
+	return kTargets
+}
+
+// Extract image targets iff they can be updated in-place in a container.
+func extractImageTargetsForLiveUpdates(specs []model.TargetSpec, stateSet store.BuildStateSet) ([]model.ImageTarget, error) {
```
maiamcc: are we assuming that intermediate images won't have fast-build instructions? that they won't have `state.DeployInfo` populated?

internal/engine/synclet_build_and_deployer.go
```diff
@@ -33,44 +32,26 @@ func NewSyncletBuildAndDeployer(sm SyncletManager, kCli k8s.Client) *SyncletBuil
...
-	if state.IsEmpty() {
-		return fmt.Errorf("prev. build state is empty; synclet build does not support initial deploy")
+	if len(iTargets) != 1 {
+		return store.BuildResultSet{}, RedirectToNextBuilderf("Synclet container builder needs exactly one image target")
 	}
 
-	if image.MaybeFastBuildInfo() == nil {
-		return fmt.Errorf("container build only supports FastBuilds")
+	iTarget := iTargets[0]
+	if !isImageDeployedToK8s(iTarget, extractK8sTargets(specs)) {
```
maiamcc: ah so I see why we check `isImageDeployedToK8s` here -- but we DON'T check in the localContainerBuilder cuz that might be called with DockerCompose, yeah? BUT is there a risk of the localContainerBuilder trying to in-place update an _intermediate_ image?

internal/engine/build_and_deployer_test.go
nicks: mostly because the upper.go call also uses AsOneResult()

internal/engine/extractors.go
nicks: it's a required argument? i could understand tweaking it if this was a performance-sensitive codepath, but i don't think it matters much here. what would you do instead?

internal/engine/extractors.go
nicks: ya, i added some safeguards to make sure that the BuildState will only have DeployInfo populated if the image names match.

internal/engine/extractors.go
nicks: ya, this should really be called ContainerInfo or something. updated the comments + error message

internal/engine/image_build_and_deployer.go
nicks: updated!

internal/engine/synclet_build_and_deployer.go
nicks: ya, the synclet won't work with docker-compose.

the local container builder should be ok as long as the BuildState for the image target is correct

internal/engine/extractors.go
jazzdan: I guess I normally do `ktargets := []model.K8sTarget{}` so that's why it looked weird. This SGMT though!

PR #1250, opened 2019-03-05
tiltfile: add k8s_kind

landism:
Problems:
1. In order to use CRDs, users had to manually join images with k8s entities by passing them all together to `k8s_resource`, which felt kind of clunky.
2. `k8s_resource` only takes one image, which meant this didn't work with CRDs with multiple images.

Solution:
Add `k8s_extra_image_location` method, which says where to find the image definitions for a given k8s object type, so now CRDs can have multiple images and don't need a call to `k8s_resource`, but can just use the normal tiltfile joining logic.

### DIFF
--- a/internal/tiltfile/k8s.go
+++ b/internal/tiltfile/k8s.go
@@ -334,35 +334,32 @@ func podLabelsFromStarlarkValue(v starlark.Value) ([]labels.Selector, error) {
 	}
 }
 
-func (s *tiltfileState) k8sExtraImageLocation(thread *starlark.Thread, fn *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
+func (s *tiltfileState) k8sKind(thread *starlark.Thread, fn *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
 	var kind string
-	var jsonPath starlark.Value
-	if err := starlark.UnpackArgs(fn.Name(), args, kwargs,
+	if err := starlark.UnpackArgs(fn.Name(), args, nil,
 		"kind", &kind,
-		"json_path", &jsonPath,
 	); err != nil {
 		return nil, err
 	}
 
-	switch x := jsonPath.(type) {
-	case starlark.String:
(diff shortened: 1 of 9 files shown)


internal/tiltfile/tiltfile_state.go
```diff
@@ -93,10 +95,11 @@ const (
...
 	// k8s functions
-	k8sYamlN     = "k8s_yaml"
-	filterYamlN  = "filter_yaml"
-	k8sResourceN = "k8s_resource"
-	portForwardN = "port_forward"
+	k8sYamlN              = "k8s_yaml"
+	filterYamlN           = "filter_yaml"
+	k8sResourceN          = "k8s_resource"
+	portForwardN          = "port_forward"
+	k8sExtraImageLocation = "k8s_extra_image_location"
```
maiamcc: not thrillllled about this function name... `k8s_extra_image_location_for_kind`? (That's also not great... feel free to leave it for now if you can't come up w/ anything better.)

internal/tiltfile/k8s.go
```diff
@@ -334,6 +334,39 @@ func podLabelsFromStarlarkValue(v starlark.Value) ([]labels.Selector, error) {
...
+		"kind", &kind,
+		"json_path", &jsonPath,
+	); err != nil {
+		return nil, err
+	}
+
+	switch x := jsonPath.(type) {
+	case starlark.String:
+		s.k8sImageJsonPathsByKind[kind] = []string{jsonPath.String()}
+	case *starlark.List:
```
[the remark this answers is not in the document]
landism: We kiiiinda have this logic elsewhere, but it's doing slightly different things (e.g., k8s_yaml has to handle blobs, or extra_pod_labels accepts dicts), but I made a helper method that takes a Value that might be a List, and turns it into a []Value, which at least gets the starlark iteration code out of a lot of our functions

internal/tiltfile/tiltfile_state.go
landism: Yeah, naming this thing is causing me grief. I'll come up with a few options and bounce them off people on slack.

landism: ok, this seems not as terrible as most options:
`k8s_kind('Environment', image_json_path='{.spec.runtime.image}')`

PR #1251, opened 2019-03-05
engine: hack to allow us to watch all containers of a pod for logs [ch1737]

maiamcc:
This is the quickest way to support multi-container logs.

It doesn't actually improve our model in any useful way -- that's
for a separate PR.

### DIFF
--- a/internal/engine/podlogmanager.go
+++ b/internal/engine/podlogmanager.go
@@ -86,7 +86,7 @@ func (m *PodLogManager) diff(ctx context.Context, st store.RStore) (setup []PodL
 			containerInfos := pod.ContainerInfos
 			if len(containerInfos) == 0 {
 				containerInfos = []store.ContainerInfo{
-					store.ContainerInfo{Name: pod.ContainerName, ID: pod.ContainerID},
+					store.ContainerInfo{ID: pod.ContainerID, Name: pod.ContainerName},
 				}
 			}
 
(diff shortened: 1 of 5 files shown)


maiamcc:
It looks real ugly, but it does the thing:

screen shot 2019-03-05 at 3 43 19 pm
screen shot 2019-03-05 at 3 43 27 pm

landism:
I'd expect this to show synclet logs, which we probably don't want. Have you tested this with a fast build in GKE?

internal/engine/podlogmanager.go
```diff
@@ -50,9 +56,18 @@ func (m *PodLogManager) diff(ctx context.Context, st store.RStore) (setup []PodL
 				continue
 			}
 
-			watchByPod, ok := m.watchesByPod[pod.PodID]
-			if ok && watchByPod.cID != pod.ContainerID {
-				watchByPod.cancel()
+			// ASDFG what?!
+			// maybe check for main container id and shut down all if main cID doesn't match?
```
landism: this comment block feels more like a WIP comment than one we'd want to have to interpret in a few weeks

maiamcc: lol indeed, this is a WIP comment 😅 deleted

maiamcc:
> I'd expect this to show synclet logs, which we probably don't want. Have you tested this with a fast build in GKE?

yep, was just talking to DM about this. fixed!

PR #1255, opened 2019-03-06
development/logging: logger that writes to file (for hud debugging)

maiamcc:
was getting frustrated trying to debug someting in the hud,
threw this together. does it seem useful? if it's not worth
merging into master, I can just hang onto it on local.

### DIFF
--- a/tools/devlog/log_to_file.go
+++ b/tools/devlog/log_to_file.go
@@ -34,7 +34,7 @@ func (lf *logFile) maybeLoad() {
 	lf.mu.Lock()
 	defer lf.mu.Unlock()
 
-	f, err := os.OpenFile(filename, os.O_RDWR|os.O_CREATE, 0644)
+	f, err := os.OpenFile(filename, os.O_APPEND|os.O_WRONLY|os.O_CREATE, 0644)
 	if err != nil {
 		panic(fmt.Sprintf("creating logfile: %v", err))
 	}
@@ -42,6 +42,9 @@ func (lf *logFile) maybeLoad() {
 	lf.f = f
 }
 
+// THIS SHOULD NEVER END UP IN IN PRODUCTION CODE. This func is for debugging
+// when it's awkward to Printf/use the real logger. Calls to this func should
+// never end up on `master.`
 func Logf(msg string, a ...interface{}) {
 	s := fmt.Sprintf(msg, a...)
 	theLogfile.write(s + "\n")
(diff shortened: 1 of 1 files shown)


tools/devlog/log_to_file.go
```diff
@@ -0,0 +1,37 @@
...
+func logFileName() string {
+	now := time.Now()
+	tStr := now.Format("2006-01-02T15.04.05")
+	return fmt.Sprintf(logfileBaseName, tStr)
+}
+
+func LogToFilef(msg string, a ...interface{}) {
+	var err error
+
+	if theLogfile == nil {
```
landism: prob deserves a mutex

tools/devlog/log_to_file.go
```diff
@@ -0,0 +1,37 @@
...
+	"os"
+	"time"
+)
+
+var logfileBaseName = "/tmp/tilt-log-%s"
+var theLogfile *os.File // THERE CAN ONLY BE ONE!
+
+func logFileName() string {
+	now := time.Now()
+	tStr := now.Format("2006-01-02T15.04.05")
```
landism: I'd vote for just using a constant filename (no timestamp) and opening the file for append, so that we get one file and easy tab completion instead of potentially hundreds of files and annoying tab completion conflicts

tools/devlog/log_to_file.go
```diff
@@ -6,32 +6,44 @@ package devlog
...
-	_, err = theLogfile.WriteString(s)
+func (lf *logFile) maybeLoad() {
+	if lf.f != nil {
+		return
+	}
+
+	lf.mu.Lock()
+	defer lf.mu.Unlock()
+
+	f, err := os.OpenFile(filename, os.O_RDWR|os.O_CREATE, 0644)
```
landism: I think this should be `os.O_APPEND | os.O_WRONLY | os.O_CREATE`

tools/devlog/log_to_file.go
```diff
@@ -6,32 +6,44 @@ package devlog
...
 	if err != nil {
-		panic(fmt.Sprintf("writing to logfile %s: %v", theLogfile.Name(), err))
+		panic(fmt.Sprintf("creating logfile: %v", err))
 	}
+
+	lf.f = f
+}
+
+func Logf(msg string, a ...interface{}) {
```
landism: it'd be good to have a comment here along the lines of "this is for printf-debugging in cases when it'd be awkward to write to the hud or use the real tilt logger, and calls to this function should not be merged to master"

maiamcc: yeah will do!

PR #1257, opened 2019-03-06
engine: keep trying to watch for Docker Compose events if we didn't actually start watching the first time

maiamcc:
Hello @jazzdan, @landism,

Please review the following commits I made in branch maiamcc/fix-dc-event-watch:

e902d4be932125238e28e5bce01559cdebb03510 (2019-03-06 17:12:22 -0500)
engine: keep trying to watch for Docker Compose events if we didn't actually start watching the first time

### DIFF
--- a/internal/engine/docker_compose_event_watcher.go
+++ b/internal/engine/docker_compose_event_watcher.go
@@ -30,7 +30,6 @@ func (w *DockerComposeEventWatcher) OnChange(ctx context.Context, st store.RStor
 	if !w.needsWatch(st) {
 		return
 	}
-	w.watching = true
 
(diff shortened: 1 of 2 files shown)


maiamcc:
Ha whoops I didn't give context. because of the way we now `init` before all the manifests have been loaded, the first call to DCEventWatcher.OnChange() would think there were no DC manifests and return -- but not before setting watch = True, which meant that no subsequent call to DCEventWatcher.OnChange() would start a new watch.

nicks:
unit test?

maiamcc:
thx for the merge @nicks !

PR #1264, opened 2019-03-07
engine: make pod logs resilient to network interruptions

jazzdan:


nicks:
unit test?

jazzdan:
@nicks added!

internal/engine/podlogmanager.go
```diff
@@ -146,38 +145,43 @@ func (m *PodLogManager) consumeLogs(watch PodLogWatch, st store.RStore) {
...
+			containerName := watch.cName
+			ns := watch.namespace
+			startTime := watch.startWatchTime
+			readCloser, err := m.kClient.ContainerLogs(watch.ctx, pID, containerName, ns, startTime)
+			if err != nil {
+				logger.Get(watch.ctx).Infof("Error streaming %s logs: %v", name, err)
+				break
+			}
+			defer func() {
+				_ = readCloser.Close()
```
nicks: i don't think this will do what you want? this will execute when the function returns, but you actually want it to execute when the block finishes.

nicks: maybe move this whole block into a separate function?

internal/engine/podlogmanager_test.go
```diff
@@ -87,6 +87,12 @@ func TestLogsFailed(t *testing.T) {
 	f.plm.OnChange(f.ctx, f.store)
 	f.AssertOutputContains("Error streaming server logs")
 	assert.Contains(t, f.out.String(), "my-error")
```
nicks: can you assert the number of times this contains my-error?

internal/engine/podlogmanager.go
```diff
@@ -146,38 +145,43 @@ func (m *PodLogManager) consumeLogs(watch PodLogWatch, st store.RStore) {
...
+		default:
+			name := watch.name
+			pID := watch.podID
+			containerName := watch.cName
+			ns := watch.namespace
+			startTime := watch.startWatchTime
+			readCloser, err := m.kClient.ContainerLogs(watch.ctx, pID, containerName, ns, startTime)
+			if err != nil {
+				logger.Get(watch.ctx).Infof("Error streaming %s logs: %v", name, err)
+				break
```
nicks: i'm concerned about the infinite loop potential here. i'm pretty sure this will spin out of control quickly if the network dies. can you add some sort of exponential backoff to the outer loop?

internal/engine/podlogmanager.go
```diff
@@ -146,38 +145,43 @@ func (m *PodLogManager) consumeLogs(watch PodLogWatch, st store.RStore) {
...
+				podID:        pID,
+			}
+			multiWriter := io.MultiWriter(prefixLogWriter, actionWriter)
+			if watch.shouldPrefix {
+				prefix = fmt.Sprintf("[%s] ", watch.cName)
+				multiWriter = logger.NewPrefixedWriter(prefix, multiWriter)
+			}
+			_, err = io.Copy(multiWriter, NewHardCancelReader(watch.ctx, readCloser))
+			if err != nil && watch.ctx.Err() == nil {
+				logger.Get(watch.ctx).Infof("Error copying %s logs: %v", name, err)
```
nicks: when this goes around the loop again, this will re-stream all the logs from `startTime`, so i think you need to do something to make sure logs aren't duplicated, like set startTime here?

internal/engine/podlogmanager.go
```diff
@@ -146,38 +145,43 @@ func (m *PodLogManager) consumeLogs(watch PodLogWatch, st store.RStore) {
 		watch.cancel()
```
nicks: i'm pretty sure this cancel() and terminationTime is obsolete now, since the only way to exit the loop is to cancel the watch and remove it from the watches array

jazzdan:
I'm going to close this for now: I don't think this was as big of a problem as I thought. It appears that the connection reset messages are just log spam and container logs continue to work afterwords.

PR #1268, opened 2019-03-08
engine: deploy synclet for custom_builds with fast build info

jazzdan:


internal/engine/image_build_and_deployer.go
```diff
@@ -115,7 +115,7 @@ func (ibd *ImageBuildAndDeployer) BuildAndDeploy(ctx context.Context, st store.R
 		}
 
 		q.SetResult(iTarget.ID(), store.BuildResult{Image: ref})
-		anyFastBuild = anyFastBuild || iTarget.IsFastBuild()
+		anyFastBuild = anyFastBuild || iTarget.MaybeFastBuildInfo() != nil
```
landism: It seems bad that `IsFastBuild` isn't accurate here - we should update or remove it.

internal/engine/build_and_deployer_test.go
```diff
@@ -466,7 +466,7 @@ func TestIgnoredFiles(t *testing.T) {
 	})
 }
 
-func TestCustomBuild(t *testing.T) {
+func TestCustomBuildWithFastBuild(t *testing.T) {
```
landism: do we want a non-fast-build test?

(definitely not a blocker for this PR since this PR isn't causing that problem, just revealing it)

jazzdan: Yeah I added to my todo list to add more tests here now that I have a better understanding of how it works.

internal/engine/image_build_and_deployer.go
jazzdan: I think @dbentley was saying that it's desirable for `IsFastBuild()` to mean that it is a fast build and _only_ a fast build and not to overload it to mean "HasFastBuild" or "ContainsFastBuild".

dbentley: Yeah, I think the way we think of "Fast Build" is weird. We use this now in just one place (analytics_reporter.go); maybe make it use MaybeFastBuild and delete IsFastBuild. (We probably don't need to assert, in tiltfile_docker_compose_test.go, that a target is both staticbuild and not fastbuild)

PR #1277, opened 2019-03-08
tiltfile: have an image type enum

jazzdan:


internal/tiltfile/tiltfile_state.go
```diff
@@ -494,22 +494,22 @@ func (s *tiltfileState) imgTargetsForDependencyIDsHelper(ids []model.TargetID, c
 			Ref: image.ref,
 		}.WithCachePaths(image.cachePaths)
 
-		switch {
-		case !image.staticBuildPath.Empty():
+		switch imageType := image.Type(); imageType {
```
landism: nit: I think this can be `switch imageType.(type) {`

internal/tiltfile/docker.go
```diff
@@ -43,6 +43,31 @@ func (d *dockerImage) ID() model.TargetID {
 	return model.ImageID(d.ref)
 }
 
+type dockerImageBuildType int
+
+const (
+	StaticBuild = iota
+	FastBuild
+	CustomBuild
+	UnknownBuild
```
landism: It probably doesn't really matter given how this is currently being used, but I think it's good to put `Unknown` at the top of the iota, so that we're more likely to catch uninitialized values rather than unintentionally plodding along with StaticBuild. I thought I read this in Effective Go, but now I can't find a citation. Anyway, it seems like a good idea.

PR #1285, opened 2019-03-11
tiltfile: support image json paths by entity name

landism:
Problem:
Using image names in env vars requires monkeying with `k8s_resource`, which doesn't work when there are multiple images for an entity. (I think, anyway? Apparently I didn't write any automated tests for env var support, so I'm not exactly sure what setup I used when manually testing this).

Solution:
Rename `k8s_kind` to `k8s_image_json_paths` and add `name` and `namespace` parameters.

This feature is still incomplete since these paths are not used at injection time. Image ID injection currently looks for image names in all fields (for CRDs) or in all env vars and container specs (for non-CRDs). It's currently sufficiently done to unblock env var usage, and we'll just keep it undocumented for now.

### DIFF
--- a/internal/k8s/entity.go
+++ b/internal/k8s/entity.go
@@ -19,8 +19,6 @@ import (
 type K8sEntity struct {
 	Obj  runtime.Object
 	Kind *schema.GroupVersionKind
-	// JSON paths to places that contain images other than Container specs
-	ImageJSONPaths []string
 }
 
 type k8sMeta interface {
@@ -195,8 +193,8 @@ func Filter(entities []K8sEntity, test func(e K8sEntity) (bool, error)) (passing
 	return passing, rest, nil
 }
 
-func FilterByImage(entities []K8sEntity, img container.RefSelector) (passing, rest []K8sEntity, err error) {
-	return Filter(entities, func(e K8sEntity) (bool, error) { return e.HasImage(img) })
+func FilterByImage(entities []K8sEntity, img container.RefSelector, imageJSONPaths func(K8sEntity) []JSONPath) (passing, rest []K8sEntity, err error) {
+	return Filter(entities, func(e K8sEntity) (bool, error) { return e.HasImage(img, imageJSONPaths(e)) })
 }
 
 func FilterBySelectorMatchesLabels(entities []K8sEntity, labels map[string]string) (passing, rest []K8sEntity, err error) {
--- a/internal/k8s/entity.go
+++ b/internal/k8s/entity.go
@@ -7,17 +7,20 @@ import (
 	"strings"
 
 	"github.com/pkg/errors"
-	"github.com/windmilleng/tilt/internal/container"
 	"k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+
+	"github.com/windmilleng/tilt/internal/container"
 )
 
 type K8sEntity struct {
 	Obj  runtime.Object
 	Kind *schema.GroupVersionKind
+	// JSON paths to places that contain images other than Container specs
+	ImageJSONPaths []string
 }
 
 type k8sMeta interface {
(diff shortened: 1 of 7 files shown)


internal/tiltfile/k8s.go
```diff
@@ -324,32 +324,39 @@ func podLabelsFromStarlarkValue(v starlark.Value) ([]labels.Selector, error) {
...
 	}
-	s.k8sImageJsonPathsByKind[kind] = paths
+
+	k := imageJSONPathSelector{
+		kind:      kind,
+		name:      name,
+		namespace: namespace,
+	}
+
+	s.k8sImageJSONPaths[k] = paths
```
maiamcc: not a huge deal, but is it maybe worth it to try and parse the image path here so if it's invalid, we throw an error at the line the user specified it and not from some random place during execution?

internal/tiltfile/k8s.go
```diff
@@ -543,3 +550,19 @@ func (s *tiltfileState) portForwardsToDomain(r *k8sResource) []model.PortForward
 	}
 	return result
 }
+
+func (s *tiltfileState) imageJSONPaths(e k8s.K8sEntity) []string {
```
maiamcc: hmmm this returns the imageJSONPaths that apply to the given entity, yeah?

internal/k8s/entity.go
```diff
@@ -7,17 +7,20 @@ import (
...
 	"k8s.io/apimachinery/pkg/runtime/schema"
+
+	"github.com/windmilleng/tilt/internal/container"
 )
 
 type K8sEntity struct {
 	Obj  runtime.Object
 	Kind *schema.GroupVersionKind
+	// JSON paths to places that contain images other than Container specs
+	ImageJSONPaths []string
```
maiamcc: ...I feel weird putting a Tiltfile-specific construct on a kubernetes object...?

internal/tiltfile/tiltfile_state.go
```diff
@@ -31,13 +31,15 @@ type tiltfileState struct {
...
-	dc                      dcResourceSet // currently only support one d-c.yml
-	k8sImageJsonPathsByKind map[string][]string
+	configFiles    []string
+	buildIndex     *buildIndex
+	k8s            []*k8sResource
+	k8sByName      map[string]*k8sResource
+	k8sUnresourced []k8s.K8sEntity
+	dc             dcResourceSet // currently only support one d-c.yml
+
+	// JSON paths to locations that contain images other than Container specs
```
maiamcc:
```suggestion
	// JSON paths to locations in k8s YAML (other than Container specs) that contain images
```

internal/tiltfile/tiltfile_state.go
```diff
@@ -583,3 +591,10 @@ const (
 	claimPending
 	claimFinished
 )
+
+// A selector matches an entity if all non-empty selector fields match the corresponding entity fields
+type imageJSONPathSelector struct {
```
maiamcc: ooh neat. do we want to generalize this to be a k8s selector rather than just an imageJsonPathSelector? (maybe leave a todo?)

internal/tiltfile/tiltfile_test.go
```diff
@@ -1706,7 +1708,7 @@ func TestExtraImageLocationOneImage(t *testing.T) {
 	f.dockerfile("env/Dockerfile")
 	f.dockerfile("builder/Dockerfile")
 	f.file("Tiltfile", `k8s_yaml('crd.yaml')
-k8s_kind('Environment', image_json_path='{.spec.runtime.image}')
+k8s_image_json_path(kind='Environment', path='{.spec.runtime.image}')
```
maiamcc: nit: you should probably have `path` be the first arg (and be an arg not a kwarg) if that's how we expect users to do it.

internal/tiltfile/tiltfile_state.go
landism: that's pretty much just a rename, unless I misunderstand? so I just did it

internal/tiltfile/k8s.go
landism: yes, good idea!

internal/tiltfile/tiltfile_state.go
maiamcc: yeah! i guess in future we could use this for like, `filter_yaml` and stuff? but no need to implement that now.

PR #1288, opened 2019-03-11
hud: add a tab pane, behind an env variable

nicks:

Please review the following commits I made in branch nicks/ch1708/tui-inspector-pane:

bbb05fb1311c9150b23a8875378ff6131d531150 (2019-03-11 16:43:43 -0400)
hud: add a tab pane, behind an env variable


nicks:
Han and I paired on this. still needs a bit of polish but here's what it looks like right now:
Screenshot from 2019-03-11 16-44-48

internal/hud/hud.go
```diff
@@ -161,19 +163,29 @@ func (h *Hud) handleScreenEvent(ctx context.Context, dispatch func(action store.
 				dispatch(view.AppendToTriggerQueueAction{
 					Name: selected.Name,
 				})
+			case r == '1':
```
yuindustries: We forgot to update these keys to reflect the new order of tabs!

nicks: updated!

PR #1291, opened 2019-03-12
engine: show all files that triggered a Tiltfile build

jazzdan:
Now when you change a a config file that causes a manifest's definition to change, and thus be re-deployed, that file will show up as an edited file in both the Tiltfile _and_ the manifest that changed.

Here's an example: I changed `deploy/snack.yaml` in Servantes and both `snack` and `(Tiltfile)` are marked as edited, but `hypothesizer` is not.

Screen Shot 2019-03-12 at 12 56 33 PM

### DIFF
--- a/internal/engine/upper.go
+++ b/internal/engine/upper.go
@@ -9,7 +9,8 @@ import (
 
 	"github.com/opentracing/opentracing-go"
 	"github.com/pkg/errors"
-	"k8s.io/api/core/v1"
+	"github.com/windmilleng/tilt/internal/tiltfile"
+	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/labels"
 
 	"github.com/windmilleng/tilt/internal/container"
@@ -21,7 +22,6 @@ import (
 	"github.com/windmilleng/tilt/internal/model"
 	"github.com/windmilleng/tilt/internal/store"
 	"github.com/windmilleng/tilt/internal/synclet/sidecar"
-	"github.com/windmilleng/tilt/internal/tiltfile"
 	"github.com/windmilleng/tilt/internal/watch"
 )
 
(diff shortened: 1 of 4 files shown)


jazzdan:
TODO: write a unit test (will do before I merge)

internal/engine/upper.go
```diff
@@ -198,8 +198,11 @@ func handleBuildStarted(ctx context.Context, state *store.EngineState, action Bu
 		return
 	}
 
+	edits := []string{}
+	edits = append(edits, action.FilesChanged...)
+
 	bs := model.BuildRecord{
-		Edits:     append([]string{}, action.FilesChanged...),
+		Edits:     append(edits, ms.ConfigFilesThatCausedChange...),
```
nicks: AFAICT, this introduces a bug where all subsequent builds will contain ContfigFilesThatCausedChange...

internal/engine/upper.go
```diff
@@ -484,13 +487,18 @@ func handleConfigsReloaded(
...
+		mt.State.ConfigFilesThatCausedChange = []string{}
+		configFilesThatChanged := state.LastTiltfileBuild.Edits
 		if !m.Equal(mt.Manifest) {
 			mt.Manifest = m
 
 			// Manifest has changed, ensure we do an image build so that we apply the changes
 			state := mt.State
 			state.BuildStatuses = make(map[model.TargetID]*store.BuildStatus)
 			state.PendingManifestChange = time.Now()
+			state.ConfigFilesThatCausedChange = configFilesThatChanged
```
nicks: it might be better to update state.PendingFiles? And then change BuildController so that it knows this is not really the first build if there are existing PendingFiles?

Part of the problem, AFAICT, is that BuildController thinks this is the first build when the manifest gets cleared.

internal/engine/upper.go
jazzdan: Yep, you're right.

jazzdan:
@nicks Now it's attached to the manifest state and consumed (reset) when those files are appended to the build record in `handleBuildStarted`

PR #1294, opened 2019-03-12
engine: don't kick off a build while tiltfile execution is pending

nicks:
Hello @landism, @maiamcc,

Please review the following commits I made in branch nicks/ch1826/manifest:

b12b1646ad721a79ec3b126eea4df5545511995a (2019-03-12 16:10:45 -0400)
engine: don't kick off a build while tiltfile execution is pending

### DIFF
--- a/internal/engine/configs_controller.go
+++ b/internal/engine/configs_controller.go
@@ -14,6 +14,7 @@ type ConfigsController struct {
 	disabledForTesting bool
 	tfl                tiltfile.TiltfileLoader
 	clock              func() time.Time
+	activeBuild        bool
 }
 
 func NewConfigsController(tfl tiltfile.TiltfileLoader) *ConfigsController {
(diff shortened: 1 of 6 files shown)


internal/engine/configs_controller.go
```diff
@@ -27,23 +28,44 @@ func (cc *ConfigsController) DisableForTesting(disabled bool) {
 	cc.disabledForTesting = disabled
 }
 
+// Modeled after BuildController.nextTargetToBuild. Check to see that:
+// 1) There's currently no Tiltfile build running,
+// 2) There are pending file changes, and
+// 3) Those files have changed since the last Tiltfile build
+//    (so that we don't keep re-running a failed build)
+func (cc *ConfigsController) hasChangesToBuild(state store.EngineState) bool {
+	isRunning := !state.CurrentTiltfileBuild.StartTime.IsZero()
```
landism: naming nit: "has changes to build" isn't quite the same as "should start a build" (i.e., there could be changes to build but we choose not to build because another build is already running)

maybe `shouldBuild`?

nicks: done!

nicks:
re: moving everything into build controller: ya, possibly? for some definition of "build controller" (it obviously would obviously have to evolve a bit). 

i've been thinking a lot about what the internals of other modern graph-based build systems look like (e.g., Bazel, Buck). I would expect BuildController to just pick the next target to build, and dispatch abstract BuildStarted / BuildCompleted events. Then it would have some sort of Handler table that knew how to kick off the builds of each type (image builds vs k8s deploys vs tiltfile runs)

PR #1295, opened 2019-03-12
build: print dockerfile that is being built

jazzdan:


jazzdan:
Open question: does this level of detail belong at the top level? Or should this live at the debug level?

PR #1298, opened 2019-03-12
tiltfile: add json_decode and read_json function

jazzdan:


internal/tiltfile/tiltfile_state.go
```diff
@@ -108,6 +108,8 @@ const (
 	kustomizeN    = "kustomize"
 	helmN         = "helm"
 	listdirN      = "listdir"
+	jsonDecodeN   = "json_decode"
```
landism: Is it worth being consistent about putting "json" at either the start or end of these names, or is there a tradeoff here?

internal/tiltfile/files.go
```diff
@@ -372,3 +373,84 @@ func (s *tiltfileState) listdir(thread *starlark.Thread, fn *starlark.Builtin, a
...
+
+	contents, err := s.readFile(localPath)
+	if err != nil {
+		return nil, err
+	}
+
+	var decodedJSON interface{}
+
+	if err := json.Unmarshal(contents, &decodedJSON); err != nil {
+		return nil, err
```
landism: I think it's worth writing a test for invalid JSON and then wrapping this error to make sure it's useful.

When I try to read an invalid JSON file, it currently looks like the Tiltfile itself has a syntax error:
```
       2019/03/13 11:21:40 Beginning Tiltfile execution
         /Users/matt/go/src/github.com/windmilleng/servantes/Tiltfile:2: in
       <toplevel>
       Error: invalid character '\n' in string literal
```

internal/tiltfile/files.go
```diff
@@ -372,3 +373,84 @@ func (s *tiltfileState) listdir(thread *starlark.Thread, fn *starlark.Builtin, a
...
+	if err := starlark.UnpackArgs(fn.Name(), args, kwargs, "path", &path); err != nil {
+		return nil, err
+	}
+
+	localPath, err := s.localPathFromSkylarkValue(path)
+	if err != nil {
+		return nil, fmt.Errorf("Argument 0 (path): %v", err)
+	}
+
+	contents, err := s.readFile(localPath)
```
landism: I made a two-line Tiltfile:

```
read_json('foo.json')
print(x["foo"])
```
and changes to foo.json didn't trigger a Tiltfile build

but then if I made the same Tiltfile with `read_file`, I hit the same problem, so I guess it's separate from this PR.

landism: filed CH1839

internal/tiltfile/tiltfile_state.go
jazzdan: The consistency in my head was if the function is `read_file` then the thing that reads json from a file should be `read_json` or at least `read_json_file`.

And I'm now realizing I stole `json_decode` [from PHP](http://php.net/manual/en/function.json-decode.php) 😬

PR #1302, opened 2019-03-13
engine: fix up dc deploys when tilt is configured to talk to k8s

nicks:
Hello @maiamcc,

Please review the following commits I made in branch nicks/dcimage4:

10fa555e9e7a75144305b6cc7abee82e56862927 (2019-03-12 21:08:59 -0400)
engine: fix up dc deploys when tilt is configured to talk to k8s

### DIFF
--- a/internal/engine/build_and_deployer.go
+++ b/internal/engine/build_and_deployer.go
@@ -105,5 +105,5 @@ func DefaultBuildOrder(sbad *SyncletBuildAndDeployer, cbad *LocalContainerBuildA
 	if runtime == container.RuntimeDocker {
 		ibad.SetInjectSynclet(true)
 	}
-	return BuildOrder{sbad, dcbad, ibad}
+	return BuildOrder{sbad, cbad, dcbad, ibad}
 }
--- a/internal/engine/build_and_deployer_test.go
+++ b/internal/engine/build_and_deployer_test.go
@@ -553,6 +553,31 @@ func TestContainerBuildMultiStage(t *testing.T) {
 	assert.Equal(t, k8s.MagicTestContainerID, result.OneAndOnlyContainerID().String())
 }
 
+func TestDockerComposeFastBuild(t *testing.T) {
+	f := newBDFixture(t, k8s.EnvGKE)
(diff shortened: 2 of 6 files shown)


internal/engine/build_and_deployer.go
```diff
@@ -105,5 +105,5 @@ func DefaultBuildOrder(sbad *SyncletBuildAndDeployer, cbad *LocalContainerBuildA
 	if runtime == container.RuntimeDocker {
 		ibad.SetInjectSynclet(true)
 	}
-	return BuildOrder{sbad, dcbad, ibad}
+	return BuildOrder{sbad, cbad, dcbad, ibad}
```
maiamcc: 😳 good catch!

internal/engine/extractors.go
```diff
@@ -45,6 +45,18 @@ func extractImageTargets(specs []model.TargetSpec) []model.ImageTarget {
 	return iTargets
 }
 
+func extractDockerComposeTargets(specs []model.TargetSpec) []model.DockerComposeTarget {
+	iTargets := make([]model.DockerComposeTarget, 0)
```
maiamcc: `iTarget` is shorthand for "image targets", call this something else

internal/engine/build_and_deployer_test.go
```diff
@@ -553,6 +553,27 @@ func TestContainerBuildMultiStage(t *testing.T) {
 	assert.Equal(t, k8s.MagicTestContainerID, result.OneAndOnlyContainerID().String())
 }
 
+func TestDockerComposeFastBuild(t *testing.T) {
+	f := newBDFixture(t, k8s.EnvNone)
```
maiamcc: imo this test is extra compelling if you set a k8s Env that would normally do a synclet build, and then asserted that there was no synclet call

internal/engine/extractors.go
nicks: done

internal/engine/build_and_deployer_test.go
nicks: done!

PR #1303, opened 2019-03-13
docker-compose: strip newline off docker-compose output

nicks:
Hello @jazzdan, @maiamcc,

Please review the following commits I made in branch nicks/newline:

de46c77129d27a472e73a4e5aa3918b91e76d04c (2019-03-12 22:33:13 -0400)
docker-compose: strip newline off docker-compose output
this was causing all sorts of weird behavior, depending on whether the api trimmed spaces implicitly


nicks:
(in particular, we had container ids with newlines running around 😬 )

PR #1304, opened 2019-03-13
tiltfile: add group and version to k8s_kind/k8s_image_json_path

landism:
to allow the fully specified group/version/kind instead of just the ambiguous kind

### DIFF
--- a/internal/tiltfile/k8s.go
+++ b/internal/tiltfile/k8s.go
@@ -338,10 +338,12 @@ func starlarkValuesToJSONPaths(values []starlark.Value) ([]k8s.JSONPath, error) 
 }
 
 func (s *tiltfileState) k8sImageJsonPath(thread *starlark.Thread, fn *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
-	var kind, name, namespace string
+	var group, version, kind, name, namespace string
 	var imageJSONPath starlark.Value
 	if err := starlark.UnpackArgs(fn.Name(), args, kwargs,
 		"path", &imageJSONPath,
+		"group?", &group,
+		"version?", &version,
 		"kind?", &kind,
 		"name?", &name,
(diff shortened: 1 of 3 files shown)


internal/tiltfile/tiltfile_test.go
```diff
@@ -1776,23 +1779,41 @@ docker_build('test/mycrd-env', 'env')
...
 
-	f.load("mycrd-env")
-	f.assertNextManifest("mycrd-env",
-		db(
-			image("docker.io/test/mycrd-env"),
-		),
-		k8sObject("mycrd", "Environment"),
-	)
+func TestK8SKind(t *testing.T) {
+	t.Run("match kind", testK8SKind("'Environment', image_json_path='{.spec.runtime.image}'", true))
```
nicks: idiomatically, you'd usually do this with a table?
https://github.com/golang/go/wiki/TableDrivenTests

i don't think i've ever seen this style

internal/tiltfile/k8s.go
```diff
@@ -338,10 +338,12 @@ func starlarkValuesToJSONPaths(values []starlark.Value) ([]k8s.JSONPath, error)
 }
 
 func (s *tiltfileState) k8sImageJsonPath(thread *starlark.Thread, fn *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
-	var kind, name, namespace string
+	var group, version, kind, name, namespace string
 	var imageJSONPath starlark.Value
 	if err := starlark.UnpackArgs(fn.Name(), args, kwargs,
 		"path", &imageJSONPath,
+		"group?", &group,
```
nicks: have we double-checked this syntax with somebody external?

usually public-facing k8s yaml uses apiVersion instead "{group}/{kind}", e.g., "core/v1". 

for users who have never hacked on kubernetes internals, i wouldn't expect them to be familiar with "group"

PR #1306, opened 2019-03-13
engine: watch config files read by broken tiltfiles

landism:


internal/engine/upper.go
```diff
@@ -477,6 +477,19 @@ func handleConfigsReloaded(
 	state.CurrentTiltfileBuild = model.BuildRecord{}
 	if event.Err != nil {
 		// There was an error, so don't update status with the new, nonexistent state
+
+		// EXCEPT for the config file list, because we want to watch new config files even when the tiltfile is broken
+		// append any new config files found in the reload action
+		existingConfigs := make(map[string]bool)
```
nicks: i wonder if we should have some sort of deduplicating append; i feel like we do this a lot

PR #1310, opened 2019-03-13
tiltfile: make staticbuild helper (i.e. make image target assertions specific to the type of build)

maiamcc:
Hello @landism, @nicks,

Please review the following commits I made in branch maiamcc/static-build-helper:

bb2a3ca255fde5e0f8231ded904e736b45768b48 (2019-03-13 16:11:34 -0400)
tiltfile: make staticbuild helper (i.e. make image target assertions specific to the type of build)

### DIFF
--- a/internal/tiltfile/tiltfile_docker_compose_test.go
+++ b/internal/tiltfile/tiltfile_docker_compose_test.go
@@ -313,7 +313,7 @@ dc_resource('foo', 'gcr.io/foo')
 
 	f.load()
 
-	m := f.assertNextManifest("foo", db(image("gcr.io/foo")))
+	m := f.assertNextManifest("foo", sb(image("gcr.io/foo")))
 	assert.True(t, m.ImageTargetAt(0).IsStaticBuild())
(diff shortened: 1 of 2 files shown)


nicks:
i thought we were moving away from the StaticBuild terminology and towards the DockerBuild terminology? and this seems to move us in the wrong direction?

maiamcc:
this PR isn't about the terminology as much as making the helper explicitly assert what kind of build it is (as opposed to just checking the fields common to staticBuild/fastBuild). I'm super down to rename this later, but i'd rather not do a big rename in the codebase before we've even phased out fastBuild in its current form.

nicks:
ah, i'm confused, this pr does a big rename from db -> sb? why not just leave it as sb?

maiamcc:
...oh. because we're asserting on the `staticBuild` struct, not the call to `docker_build` that created it.

maiamcc:
...i'm rolling back most of this rename as part of a PR getting ready for LiveUpdate and i'm full of chagrin 😅

PR #1312, opened 2019-03-13
tiltfile: support regex in k8s_kind and k8s_image_json_path

landism:
a couple of use cases:
1. When templating entity names, the tiltfile will have to duplicate the templating logic in order to pass the entity name to k8s_image_json path. e.g., in the simple Servantes case, the tiltfile would need `username() + "-vigoda"`. This allows the user to pass `name=".*-vigoda"`.
2. When specifying apiVersion, the user might care about the group but not the version. This allows the user to specify, e.g., `fission.io/.*`.

### DIFF
--- a/internal/tiltfile/tiltfile_test.go
+++ b/internal/tiltfile/tiltfile_test.go
@@ -1904,9 +1904,11 @@ func TestK8SImageJSONPathArgs(t *testing.T) {
 	}{
 		{"match name", "name='foo'", true},
 		{"don't match name", "name='bar'", false},
+		{"match name w/ regex", "name='.*o'", true},
 		{"match kind", "name='foo', kind='Deployment'", true},
 		{"don't match kind", "name='bar', kind='asdf'", false},
 		{"match apiVersion", "name='foo', api_version='apps/v1'", true},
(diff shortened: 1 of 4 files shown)


internal/tiltfile/tiltfile_test.go
```diff
@@ -1904,9 +1904,11 @@ func TestK8SImageJSONPathArgs(t *testing.T) {
 	}{
 		{"match name", "name='foo'", true},
 		{"don't match name", "name='bar'", false},
+		{"match name w/ regex", "name='.*o'", true},
 		{"match kind", "name='foo', kind='Deployment'", true},
 		{"don't match kind", "name='bar', kind='asdf'", false},
 		{"match apiVersion", "name='foo', api_version='apps/v1'", true},
+		{"match apiVersion+kind w/ regex", "name='foo', kind='Deployment', api_version='apps/.*'", true},
```
nicks: can you also add a test case for what happens if a regexp is invalid?

landism:
> nice! should FilterYaml support regexes too? (and use k8sObjectSelector under the hood?) if it's easy and you could do that, great

done

PR #1315, opened 2019-03-14
tiltfile: add context to some readfile errors

landism:


internal/tiltfile/docker.go
```diff
@@ -130,13 +130,13 @@ func (s *tiltfileState) dockerBuild(thread *starlark.Thread, fn *starlark.Builti
 
 		bs, err := s.readFile(dockerfilePath)
 		if err != nil {
-			return nil, err
+			return nil, errors.Wrap(err, "error reading docker file")
```
nicks: docker file -> dockerfile
per https://docs.docker.com/engine/reference/builder/

PR #1317, opened 2019-03-14
engine: use staticBuild.FastBuild for LiveUpdate where possible [ch1833]

maiamcc:
Hello @nicks, @landism,

Please review the following commits I made in branch maiamcc/nested-fb-live-update:

4c668b17fcebd2cb035190df4fcd0fbea5daa0d1 (2019-03-14 12:31:16 -0400)
imageTarget.FastBuildInfo consistent with MaybeFastBuild

67bf8e72b78a198514cabb7c588b9f2421e390de (2019-03-14 12:12:12 -0400)
extract nested fastbuild as a live update target

### DIFF
--- a/internal/model/docker_info.go
+++ b/internal/model/docker_info.go
@@ -109,16 +109,15 @@ func (i ImageTarget) MaybeFastBuildInfo() *FastBuild {
 	return nil
 }
 
+// FastBuildInfo returns the TOP LEVEL BUILD DETAILS, if a FastBuild.
+// Does not return nested FastBuild details.
 func (i ImageTarget) FastBuildInfo() FastBuild {
-	fb := i.MaybeFastBuildInfo()
-	if fb == nil {
-		return FastBuild{}
-	}
-	return *fb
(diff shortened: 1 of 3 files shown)


internal/model/docker_info.go
```diff
@@ -101,17 +101,24 @@ func (i ImageTarget) MaybeFastBuildInfo() *FastBuild {
...
 	case CustomBuild:
 		return details.Fast
 	}
 	return nil
 }
 
 func (i ImageTarget) FastBuildInfo() FastBuild {
-	ret, _ := i.BuildDetails.(FastBuild)
-	return ret
+	fb := i.MaybeFastBuildInfo()
```
nicks: i'm confused...why did this change?

after this change, FastBuildInfo() and MaybeFastBuildInfo() do the exact same thing?

as far as i can tell, FastBuildInfo() is only used in one place, and this would be the wrong behavior in that case

maiamcc: ...I changed it b/c matt and I were confused about WHY the behavior was different but then I did the wrong thing, let me fix.

PR #1323, opened 2019-03-15
tilt: add .tiltignore support

landism:
docs: https://github.com/windmilleng/tilt.build/pull/77

1. I'm not super thrilled about doing the loading itself in the Tiltfile loader, but that's where we're already loading configs, and that's what the Configs Controller deals with? Maybe the answer is just to wrap the tiltfile loader with a configs loader that does tiltfile + tiltignore?
2. One unfortunate behavior right now is that changing .tiltignore causes reexecution of the Tiltfile, but that doesn't seem like a blocker.

### DIFF
--- a/internal/engine/watchmanager.go
+++ b/internal/engine/watchmanager.go
@@ -147,7 +147,6 @@ func (w *WatchManager) diff(ctx context.Context, st store.RStore) (setup []Watch
 		if tiltIgnoreChanged || !watchRulesMatch(m, mnc.target) {
 			teardown = append(teardown, name)
 			setup = append(setup, m)
-			continue
 		}
 	}
 
--- a/internal/engine/watchmanager.go
+++ b/internal/engine/watchmanager.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/google/go-cmp/cmp"
 
+	"github.com/windmilleng/tilt/internal/dockerignore"
 	"github.com/windmilleng/tilt/internal/ignore"
 	"github.com/windmilleng/tilt/internal/logger"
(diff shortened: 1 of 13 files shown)


internal/engine/watchmanager.go
```diff
@@ -133,17 +135,19 @@ func (w *WatchManager) diff(ctx context.Context, st store.RStore) (setup []Watch
...
 			teardown = append(teardown, name)
 			continue
 		}
 
-		if !watchRulesMatch(m, mnc.target) {
+		if tiltIgnoreChanged || !watchRulesMatch(m, mnc.target) {
 			teardown = append(teardown, name)
 			setup = append(setup, m)
-			break
+			continue
```
jazzdan: Isn't this continue unnecessary as the last possible statement that will be executed in this for loop?

internal/engine/watchmanager_test.go
```diff
@@ -78,7 +78,6 @@ func TestWatchManager_WatchesReappliedOnDockerComposeMountsChange(t *testing.T)
 	f.SetManifestTarget(target)
 
 	f.ChangeFile(t, "bar")
-	f.ChangeFile(t, "stop")
```
nicks: why did this test change?

landism: When I was writing the new test, I thought it was silly and error-prone that I had to `f.ChangeFile(t, "stop")` myself, and that it'd be much safer to have `f.Stop` do that.

It turned out that `f.Stop` already does that, so I removed the redundant `f.ChangeFile(t, "stop")`s from other tests.

PR #1324, opened 2019-03-15
engine: add kubectl exec update mode flag

jazzdan:


internal/engine/synclet_build_and_deployer.go
```diff
@@ -171,3 +172,7 @@ func (sbd *SyncletBuildAndDeployer) updateViaExec(ctx context.Context,
 
 	return nil
 }
+
+func (sbd *SyncletBuildAndDeployer) ForceUpdateViaExec(b bool) {
```
nicks: could we just inject the UpdateMode via Wire instead of having a setter?

jazzdan: Yeah, let me play around with that.

nicks:
what's the status on this pr?

jazzdan:
Sorry, going to try to get back to it today.

jazzdan:
@nicks updated this to use wire, WDYT?

internal/cli/up.go
```diff
@@ -203,3 +203,7 @@ func provideWebURL() (model.WebURL, error) {
 	}
 	return model.WebURL(*url), nil
 }
+
+func provideSyncletExecFlag() engine.ExecFlag {
```
nicks: hmmm...can you move this provider into Engine? This should use the injected UpdateMode rather than reading directly from the flag

internal/engine/synclet_build_and_deployer.go
```diff
@@ -17,17 +17,21 @@ import (
...
+type ExecFlag bool
+
 var _ BuildAndDeployer = &SyncletBuildAndDeployer{}
 
 type SyncletBuildAndDeployer struct {
-	sm   SyncletManager
-	kCli k8s.Client
+	sm                 SyncletManager
+	kCli               k8s.Client
+	forceUpdateViaExec ExecFlag
```
nicks: this is fine too, but i'm surprised you didn't just inject UpdateMode

PR #1327, opened 2019-03-15
hud: start up the hud server by default on port 10350

nicks:
Hello @landism, @jazzdan,

Please review the following commits I made in branch nicks/ch1836/server:

c2a42361957ce5cb183f732ff6dcd14737efe908 (2019-03-15 14:32:58 -0400)
hud: start up the hud server by default on port 10350


nicks:
(i also like this pr as an example of how to move things out of up.go)

nicks:
nope, right now it's just to get the init stuff out of the cli. it might make sense to give subscribers an Init hook

PR #1328, opened 2019-03-15
engine: if all BaD calls fail, return last UNEXPECTED error [ch1854]

maiamcc:
Hello @nicks, @jazzdan,

Please review the following commits I made in branch maiamcc/weird-dcbad-err-msg:

adc40817017fc3adda8765ce8e94cdb554ab5d88 (2019-03-15 15:27:45 -0400)
engine: if all BaD calls fail, return last UNEXPECTED error [ch1854]

6b72d020648b56b15ada3b94d84dfaf38ae7854a (2019-03-15 15:23:23 -0400)
wip

### DIFF
--- a/internal/engine/build_and_deployer.go
+++ b/internal/engine/build_and_deployer.go
@@ -55,7 +55,7 @@ func NewCompositeBuildAndDeployer(builders BuildOrder) *CompositeBuildAndDeploye
 }
 
 func (composite *CompositeBuildAndDeployer) BuildAndDeploy(ctx context.Context, st store.RStore, specs []model.TargetSpec, currentState store.BuildStateSet) (store.BuildResultSet, error) {
-	var lastErr error
+	var lastErr, lastUnexpectedErr error
 	logger.Get(ctx).Debugf("Building with BuildOrder: %s", composite.builders.String())
 	for i, builder := range composite.builders {
 		logger.Get(ctx).Debugf("Trying to build and deploy with %T", builder)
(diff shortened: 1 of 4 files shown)


internal/engine/docker_compose_build_and_deployer.go
```diff
@@ -93,12 +94,12 @@ func (bd *DockerComposeBuildAndDeployer) BuildAndDeploy(ctx context.Context, st
 			// service at once, we'll have to match up image build results to DC target by ref.
 			ref, err = bd.icb.Build(ctx, iTarget, currentState[iTarget.ID()], ps, true)
 			if err != nil {
-				return store.BuildResultSet{}, err
+				return store.BuildResultSet{}, errors.Wrap(err, "DockerComposeBuildAndDeployer")
```
nicks: this seems like a weird thing to show if the user has a compiler error. i'd expect them to ignore it and blame us, which is not what we want.

internal/engine/build_and_deployer.go
```diff
@@ -71,13 +71,20 @@ func (composite *CompositeBuildAndDeployer) BuildAndDeploy(ctx context.Context,
 		if _, ok := err.(RedirectToNextBuilder); ok {
 			logger.Get(ctx).Debugf("(expected error) falling back to next build and deploy method "+
 				"after error: %v", err)
-		} else if i+1 < len(composite.builders) {
-			logger.Get(ctx).Infof("falling back to next build and deploy method "+
-				"after unexpected error: %v", err)
+		} else {
+			lastUnexpectedErr = err
+			if i+1 < len(composite.builders) {
+				logger.Get(ctx).Infof("falling back to next build and deploy method "+
```
nicks: this still seems like a confusing thing to print in the docker-compose case, where there are more deployers but none of them are applicable

internal/engine/docker_compose_build_and_deployer.go
maiamcc: Ah but it already tags them with "ImageBuild:", which to me is misleading about the specific deployer we're using. though now that i think of it, that's probably way more confusing to us than to a user. i'll take this out.

internal/engine/build_and_deployer.go
maiamcc: cool, i'll change it to something that surfaces the error w/o necessarily indicating that we're falling back.

PR #1329, opened 2019-03-15
web: use sass instead of css so we can have variables

nicks:
Hello @nicks,

Please review the following commits I made in branch nicks/sass:

56969b1e298f31de35218490b1fa21d6e8d37412 (2019-03-15 17:58:17 -0400)
web: use sass instead of css so we can have variables


web/src/constants.scss
```diff
@@ -0,0 +1,21 @@
+$font-family: 'Courier New', Courier, monospace;
```
nicks: just copied these colors from tilt.dev

PR #1332, opened 2019-03-18
hud: rename columns

maiamcc:
right now on master, resources not associated with an image that tilt builds (i.e. resources that are just k8s, or just docker compose) show “build status: N/A”:
Screen Shot 2019-03-12 at 12 32 50 PM

we made this decision on the assumption that `kubectl apply` would be quick enough that showing liveness wouldn’t be a concern, but lol turns out sometimes it isn’t -- see this example here, where `kubectl apply` took about 30s.

In order to show live-ness, I want to put the timer back in this column even when we're not building an image--it's a timer for "all the work that tilt is doing" (as opposed to the column at left, which is "status of the container in the orchestrator, which tilt has no control over).

here's one proposed change -- since we deliberately want this column to be wide enough to fit both a status and a time, I figured `BUILD + DEPLOY` wasn't _too_ long. I'm very open to suggestions tho. thoughts?

Screen Shot 2019-03-18 at 11 33 36 AM


maiamcc:
Screen Shot 2019-03-18 at 11 33 36 AM

dbentley:
WDYT of "Update"?

1) It fits into our "update+understand" theme
2) It's one word, which is shorter and stronger.
3) Being one word means we could give it a one-letter key shortcut that represented it.

maiamcc:
@dbentley yes I like `Update`! however, is there something we can do with the last column so it's not `UPDATE` and `UPDATED` side by side?

dbentley:
s/"Updated"/"As Of/"?

maiamcc:
Screen Shot 2019-03-27 at 12 24 15 PM
Screen Shot 2019-03-27 at 12 23 59 PM
Screen Shot 2019-03-27 at 12 23 54 PM

PR #1333, opened 2019-03-19
engine: support KIND

jazzdan:


internal/k8s/env.go
```diff
@@ -42,6 +43,8 @@ func EnvFromString(s string) Env {
 		return EnvDockerDesktop
 	} else if Env(s) == EnvMicroK8s {
 		return EnvMicroK8s
+	} else if Env(s) == "kubernetes-admin@kind" {
```
nicks: can you add this to the test table?

fwiw, i've also seen this as kubernetes-admin@kind-1, so it might be worth doing a prefix check or smth

internal/engine/image_build_and_deployer.go
```diff
@@ -80,7 +106,7 @@ func (ibd *ImageBuildAndDeployer) BuildAndDeploy(ctx context.Context, st store.R
 		ibd.analytics.Timer("build.image", time.Since(startTime), tags)
 	}()
 
-	numStages := len(iTargets)
+	numStages := len(iTargets) * 2 // each image target has two stages: one for build, and one for push
```
nicks: not really related to this pr (except that you're updating the logic) but this probably needs to take into account the number of NeedsImageBuild(). i wish we had a better way to assert this number is correct.

internal/k8s/env.go
jazzdan: d'oh, yes! Will do

Issue #1336, opened 2019-03-19
can't match containers if image in Tiltfile has tag


maiamcc:
See [@drubin's repro case](https://github.com/drubin/tilt-nodejs-example) -- in-place updates were failing for him with:

> In-place build needs container info.

On further investigation, this was b/c we aren't properly matching the container info [when processing a pod event](https://github.com/windmilleng/tilt/blob/dc0a49b5ad83ef7486d27fcbcd2e4a0ecc8fe1c2/internal/engine/upper.go#L683). in [`k8s.ContainerMatching`](https://github.com/windmilleng/tilt/blob/a28da6f326821f8b596df0d8b31775f9a74e4148/internal/k8s/container.go#L102), we compare the container's image ref with the image target's image ref. If the image target was initialized with a tag, then we're always checking if the running container's image matches `userimage:usertag`, which it NEVER will, because tilt-built images have the form `userimage:tilt-acb123`.

The short-term fix is to not match on tags in that particular part of the code -- it's not the most exact, but it'll only cause problems if someone's running multiple containers on the same pod with different tags. 😬 🤞

drubin:
@maiamcc This solved it for me on a master branch so I think you can consider this fixed. 

Also, I think the approach is sane.

maiamcc:
Amazing! Closing this issue.

PR #1337, opened 2019-03-19
engine: (for now) don't require image TAGS to match when checking container image for match with image target

maiamcc:
Hello @nicks, @landism,

Please review the following commits I made in branch maiamcc/match-tagged-images:

c1bdbb632fa44ad53d2cbcea549d50ab34afff10 (2019-03-19 17:21:16 -0400)
test

e367b7e29b3b6c6ec8133a031c4aa1d18bbb33cd (2019-03-19 17:20:29 -0400)
engine: (for now) don't require image TAGS to match when checking container image for match with image target

### DIFF
--- a/internal/engine/upper_test.go
+++ b/internal/engine/upper_test.go
@@ -1002,6 +1002,11 @@ func (f *testFixture) testPodWithDeployID(podID string, manifestName string, pha
 	}
 }
 
+func setImage(pod *v1.Pod, image string) {
+	pod.Spec.Containers[0].Image = image
+	pod.Status.ContainerStatuses[0].Image = image
+}
+
 func setRestartCount(pod *v1.Pod, restartCount int) {
 	pod.Status.ContainerStatuses[0].RestartCount = int32(restartCount)
 }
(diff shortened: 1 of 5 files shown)


maiamcc:
Addresses #1336

internal/engine/upper_test.go
```diff
@@ -1002,6 +1002,11 @@ func (f *testFixture) testPodWithDeployID(podID string, manifestName string, pha
 	}
 }
 
+func setImage(pod *v1.Pod, image string) {
+	pod.Spec.Containers[0].Image = image
```
nicks: i wonder if we should have an assertion that this is a special tilt tag (like "tilt-abcdef")

maiamcc: ...huh? can you say more? how would this be helpful?

nicks: we want our tests to accurately mock out production behavior

maiamcc: we might want to test pod events that have an image without a tilt tag, so i'm gonna not add that for now

nicks: 👍

PR #1339, opened 2019-03-20
tiltfile: add default_registry

landism:
docs [here](https://github.com/windmilleng/tilt.build/pull/82)

### DIFF
--- a/internal/container/default_registry.go
+++ b/internal/container/default_registry.go
@@ -16,16 +16,16 @@ func escapeName(s string) string {
 // Produces a new image name that is in the specified registry.
 // The name might be ugly, favoring uniqueness and simplicity and assuming that the prettiness of ephemeral dev image
 // names is not that important.
-func ReplaceRegistry(defaultRegistry string, rs RefSelector) (RefSelector, error) {
+func ReplaceRegistry(defaultRegistry string, rs RefSelector) (reference.Named, error) {
 	if defaultRegistry == "" {
-		return rs, nil
+		return rs.AsNamedOnly(), nil
 	}
 
 	newNs := fmt.Sprintf("%s/%s", defaultRegistry, escapeName(rs.RefName()))
 	newN, err := reference.ParseNamed(newNs)
 	if err != nil {
-		return RefSelector{}, err
+		return nil, err
 	}
 
-	return NameSelector(newN), nil
+	return newN, nil
 }
--- a/internal/container/default_registry.go
+++ b/internal/container/default_registry.go
@@ -1,0 +1,50 @@
+package container
+
+import (
+	"fmt"
+	"regexp"
+
+	"github.com/docker/distribution/reference"
+)
+
+// Produces a new image name that is in the specified registry.
+// The name might be ugly, favoring uniqueness and simplicity and assuming that the prettiness of ephemeral dev image
+// names is not that important.
+func replaceNamedTagged(defaultRegistry string, name reference.NamedTagged) (reference.NamedTagged, error) {
+	newN, err := replaceNamed(defaultRegistry, MustParseNamed(name.Name()))
+	if err != nil {
+		return nil, err
+	}
+
+	newNT, err := reference.WithTag(newN, name.Tag())
+	if err != nil {
+		return nil, err
+	}
+
+	return newNT, nil
+}
+
+var escapeRegex = regexp.MustCompile(`[/:@]`)
+
+func escapeName(s string) string {
+	return string(escapeRegex.ReplaceAll([]byte(s), []byte("_")))
+}
+
+// Produces a new image name that is in the specified registry.
+// The name might be ugly, favoring uniqueness and simplicity and assuming that the prettiness of ephemeral dev image
+// names is not that important.
+func replaceNamed(defaultRegistry string, name reference.Named) (reference.Named, error) {
+	if defaultRegistry == "" {
+		return name, nil
+	}
+
+	ns := name.String()
+
+	newNs := fmt.Sprintf("%s/%s", defaultRegistry, escapeName(ns))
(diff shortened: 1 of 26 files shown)


nicks:
to be clear, this just covers the tiltfile changes and not the engine/injection, right?

internal/container/default_registry.go
```diff
@@ -0,0 +1,50 @@
...
+	"regexp"
+
+	"github.com/docker/distribution/reference"
+)
+
+// Produces a new image name that is in the specified registry.
+// The name might be ugly, favoring uniqueness and simplicity and assuming that the prettiness of ephemeral dev image
+// names is not that important.
+func replaceNamedTagged(defaultRegistry string, name reference.NamedTagged) (reference.NamedTagged, error) {
+	newN, err := replaceNamed(defaultRegistry, MustParseNamed(name.Name()))
```
nicks: instead of serializing and de-serializing and panicking on error, you'd usually use `reference.TrimNamed(name)` here

internal/container/selector.go
```diff
@@ -12,36 +12,57 @@ const (
 )
 
 type RefSelector struct {
-	ref       reference.Named
+	// the ref to use when trying to match against a ref (i.e., pre-default-registry)
+	matchRef reference.Named
+	// the ref to use when injecting this ref (i.e., post-default-registry)
+	injectRef reference.Named
```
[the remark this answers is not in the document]
landism: Yeah, that's a lot better. I was leaning too heavily on dockerImage's field being named `ref`.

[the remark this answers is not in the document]
nicks: ya, i wish i had changed that....i feel like there's a good cleanup task for a friday afternoon to make a bunch of big name changes to be more consistent with current thinking

landism:
> to be clear, this just covers the tiltfile changes and not the engine/injection, right?

Nope, this handles injection as well.

internal/container/default_registry.go
```diff
@@ -0,0 +1,31 @@
...
+
+// Produces a new image name that is in the specified registry.
+// The name might be ugly, favoring uniqueness and simplicity and assuming that the prettiness of ephemeral dev image
+// names is not that important.
+func ReplaceRegistry(defaultRegistry string, rs RefSelector) (RefSelector, error) {
+	if defaultRegistry == "" {
+		return rs, nil
+	}
+
+	newNs := fmt.Sprintf("%s/%s", defaultRegistry, escapeName(rs.RefName()))
```
nicks: i think this is going to break the case where we have image builds for the same name but different tags? see
https://github.com/jazzdan/tilt_debug
for an example

internal/store/build_result.go
```diff
@@ -233,7 +234,7 @@ func NewDeployInfo(iTarget model.ImageTarget, podSet PodSet) DeployInfo {
 	}
 
 	// Only return the pod if it matches our image.
-	if pod.ContainerImageRef == nil || !iTarget.Ref.Matches(pod.ContainerImageRef) {
+	if pod.ContainerImageRef == nil || !iTarget.DeploymentRef.Matches(pod.ContainerImageRef) {
```
nicks: i think this is going to interact badly with Maia's changes?
https://github.com/windmilleng/tilt/pull/1337

internal/model/docker_info.go
```diff
@@ -11,8 +11,9 @@ import (
 )
 
 type ImageTarget struct {
-	Ref          container.RefSelector
-	BuildDetails BuildDetails
+	ConfigurationRef container.RefSelector
+	DeploymentRef    container.RefSelector
```
nicks: i don't think you actually want this to be a Selector? i.e., selector matching is never the right semantic behavior on deployed resources. again, see https://github.com/windmilleng/tilt/pull/1337

internal/container/default_registry.go
landism: as discussed offline, image builds for same name but different tags have two potential problems:
1. at tiltfile assembly time, joining image builds with k8s resources
2. when monitoring pods, finding the container for a k8s resource

The return value of this is not used at all for (1), and (2) is the problem kinda addressed by Maia's branch, and which behavior is already buggy, and I don't think this is changing.

I've also run this branch against Dan's tilt debug repo and it worked fine, and added a new unit test (`TestDefaultRegistryTwoImagesOnlyDifferByTag`) to verify the behavior. LMK if I've misunderstood the problem.

internal/store/build_result.go
landism: I think I've merged master in a way that maintains Maia's changes.

internal/engine/docker_compose_build_and_deployer.go
```diff
@@ -84,7 +85,7 @@ func (bd *DockerComposeBuildAndDeployer) BuildAndDeploy(ctx context.Context, st
 			return store.BuildResultSet{}, err
 		}
 
-		expectedRef := iTarget.Ref
+		expectedRef := iTarget.ConfigurationRef
```
nicks: maybe we should forbid subtitute repo in the docker_compose case

internal/engine/image_build_and_deployer.go
```diff
@@ -245,7 +247,7 @@ func (ibd *ImageBuildAndDeployer) deploy(ctx context.Context, st store.RStore, p
 					return fmt.Errorf("Internal error: missing build result for dependency ID: %s", depID)
 				}
 
-				selector := iTargetMap[depID].Ref
+				selector := iTargetMap[depID].ConfigurationRef
```
nicks: shouldn't this be the DeploymentRef?

nicks:
also, it would be nice if there was a unit test in the engine that the substituted repo is deployed

PR #1340, opened 2019-03-20
integration: add tests for multiple images with same tag

nicks:
Hello @jazzdan, @landism,

Please review the following commits I made in branch nicks/ch1941/integration:

921d95eb08dfec4f87502a8abb9b1ba7a1700b03 (2019-03-20 16:35:48 -0400)
integration: add tests for multiple images with same tag


nicks:
took this opportunity to clean up the integration tests a bit. 

- there were a bunch of things that were weird with 'go run' signal handling so i removed the 'go run'. now there are less orphaned processes running around, and the tests run a lot faster
- there's actually a bug in --watch=false :(

PR #1341, opened 2019-03-20
engine: add flag to disable push on custom build

jazzdan:
@rohansingh needs this for his bazel + KIND setup. For the moment bazel is much better at doing the layer negotiation and pushing only what is necessary. `kind load` is less efficient. @rohansingh has a workaround where he tees the image to both docker and KIND, allowing custom_build to verify that the image was created correctly and allowing KIND to get the image in a fast way. What he needs is a way to disable the image push in this case. 

I decided to add this as a flag to `custom_build` because its a pretty special case, only is used from `custom_builds` right now and fits in with `custom_build`'s evolving, experimental API.

### DIFF
--- a/internal/engine/image_build_and_deployer.go
+++ b/internal/engine/image_build_and_deployer.go
@@ -170,9 +170,14 @@ func (ibd *ImageBuildAndDeployer) push(ctx context.Context, ref reference.NamedT
 	ps.StartPipelineStep(ctx, "Pushing %s", ref.String())
 	defer ps.EndPipelineStep(ctx)
 
+	cbSkip := false
+	if iTarget.IsCustomBuild() {
+		cbSkip = iTarget.CustomBuildInfo().DisablePush
+	}
+
 	// We can also skip the push of the image if it isn't used
 	// in any k8s resources! (e.g., it's consumed by another image).
-	if ibd.canAlwaysSkipPush() || !isImageDeployedToK8s(iTarget, kTargets) {
+	if ibd.canAlwaysSkipPush() || !isImageDeployedToK8s(iTarget, kTargets) || cbSkip {
 		ps.Printf(ctx, "Skipping push")
 		return ref, nil
 	}
--- a/internal/engine/image_build_and_deployer_test.go
+++ b/internal/engine/image_build_and_deployer_test.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/docker/distribution/reference"
 	"github.com/docker/docker/api/types"
+	"github.com/opencontainers/go-digest"
 	"github.com/stretchr/testify/assert"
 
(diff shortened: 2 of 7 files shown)


internal/engine/image_build_and_deployer_test.go
```diff
@@ -261,6 +261,23 @@ func TestKINDPush(t *testing.T) {
...
+
+func TestCustomBuildDisablePush(t *testing.T) {
+	f := newIBDFixture(t, k8s.EnvKIND)
+	defer f.TearDown()
+
+	manifest := NewSanchoCustomBuildManifestWithPushDisabled(f)
+	_, err := f.ibd.BuildAndDeploy(f.ctx, f.st, buildTargets(manifest), store.BuildStateSet{})
+	// the image doesn't exist, so the custom build fails
+	// TODO(dmiller): change this test to use the static tag parameter when that's added
+	assert.Error(t, err)
```
nicks: can you add an assertion that this is the error you expect, and not some other error?

jazzdan: Yes, but I probably should have made this more clear: this test doesn't do anything at the moment because the tag is generated by the build and deployer and I can't know what it is. Thus the TODO to change this test to use the static tag parameter that @maiamcc is going to add. So really this is more of a stub, and once that parameter is added I can inject the static image tag in to the fake docker client and the build and deployer won't return an error at all.

internal/engine/image_build_and_deployer.go
```diff
@@ -168,9 +168,14 @@ func (ibd *ImageBuildAndDeployer) push(ctx context.Context, ref reference.NamedT
 	ps.StartPipelineStep(ctx, "Pushing %s", ref.String())
 	defer ps.EndPipelineStep(ctx)
 
+	cbSkip := false
+	if iTarget.IsCustomBuild() {
+		cbSkip = iTarget.CustomBuildInfo().DisablePush
+	}
+
```
maiamcc: tbh you could just to `cbSkip := iTarget.CustomBuildInfo().DisablePush` (or even take all this out and in the conditional below, just put `iTarget.CustomBuildInfo().DisablePush`)

jazzdan: Couldn't it be nil though?

jazzdan:
@maiamcc no worries, I'll wait for you to merge yours in and then resolve the conflicts over here. :D

internal/engine/image_build_and_deployer.go
maiamcc: Nah, if it's not a custom build, this func returns an EMPTY `CustomBuild` (not a nil one). though i guess you're right that that's maybe not functionality we want to rely on...?

maiamcc:
k, it's merged! glhfdd.

PR #1342, opened 2019-03-21
k8s: get KubeContext from config instead of context name

jazzdan:
This is a more reliable way to detect cluster types that are likely to have different names, like KIND.

### DIFF
--- a/internal/k8s/env.go
+++ b/internal/k8s/env.go
@@ -5,6 +5,7 @@ import (
 
 	"github.com/pkg/errors"
 	"k8s.io/client-go/tools/clientcmd"
+	"k8s.io/client-go/tools/clientcmd/api"
 )
 
 type Env string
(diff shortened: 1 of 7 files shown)


internal/k8s/env.go
```diff
@@ -36,6 +37,16 @@ func ProvideKubeContext(clientLoader clientcmd.ClientConfig) (KubeContext, error
 	return KubeContext(config.CurrentContext), nil
 }
 
+func ProvideKubeConfig(clientLoader clientcmd.ClientConfig) (*api.Config, error) {
```
nicks: can you also change ProvideKubeContext to take *api.Config as a parameter?

internal/k8s/env.go
```diff
@@ -54,3 +65,29 @@ func EnvFromString(s string) Env {
...
+
+func EnvFromConfig(config *api.Config) Env {
+	n := config.CurrentContext
+
+	c, ok := config.Contexts[n]
+	if !ok {
+		return EnvUnknown
+	}
+
+	cn := c.Cluster
```
nicks: should we also add Cluster to `tilt doctor`?

jazzdan: Yeah good idea

PR #1343, opened 2019-03-21
build: custom build can support deterministic tags (rather than generating temp tag to check against) [ch1909]

maiamcc:
in Tiltfile, user can pass an optional `tag` arg to `custom_build`. If this arg is set, we exec `CustomBuild.command` and check that it produced `user-image:customBuild.Tag`; otherwise, we generate our own temp tag, exec the command, and check that it produced `user-image:tmpTag`.

### DIFF
--- a/internal/container/selector.go
+++ b/internal/container/selector.go
@@ -12,7 +12,7 @@ const (
 )
 
 type RefSelector struct {
-	ref       reference.Named
+	Ref       reference.Named
 	matchType MatchType
 }
 
@@ -27,7 +27,7 @@ func NewRefSelector(ref reference.Named) RefSelector {
 		matchType = matchExact
 	}
 	return RefSelector{
-		ref:       ref,
+		Ref:       ref,
 		matchType: matchType,
 	}
 }
@@ -41,7 +41,7 @@ func MustParseTaggedSelector(s string) RefSelector {
 }
 
 func (s RefSelector) RefsEqual(other RefSelector) bool {
-	return s.ref.String() == other.ref.String()
+	return s.Ref.String() == other.Ref.String()
 }
 
 func (s RefSelector) WithNameMatch() RefSelector {
@@ -55,31 +55,32 @@ func (s RefSelector) WithExactMatch() RefSelector {
 }
 
 func (s RefSelector) Matches(toMatch reference.Named) bool {
-	if s.ref == nil {
+	if s.Ref == nil {
 		return false
 	}
 
 	if s.matchType == matchName {
-		return toMatch.Name() == s.ref.Name()
+		return toMatch.Name() == s.Ref.Name()
 	}
-	return toMatch.String() == s.ref.String()
+	return toMatch.String() == s.Ref.String()
 }
 
 func (s RefSelector) Empty() bool {
-	return s.ref == nil
+	return s.Ref == nil
 }
 
 func (s RefSelector) RefName() string {
-	return s.ref.Name()
+	return s.Ref.Name()
 }
 
+// AsNamedOnly returns the Ref as a Named, REMOVING THE TAG if one exists
 func (s RefSelector) AsNamedOnly() reference.Named {
-	return reference.TrimNamed(s.ref)
+	return reference.TrimNamed(s.Ref)
 }
 
 func (s RefSelector) String() string {
-	if s.ref == nil {
+	if s.Ref == nil {
 		return ""
 	}
-	return s.ref.String()
+	return s.Ref.String()
 }
--- a/internal/container/selector.go
+++ b/internal/container/selector.go
@@ -12,7 +12,7 @@ const (
 )
 
 type RefSelector struct {
-	Ref       reference.Named
+	ref       reference.Named
 	matchType MatchType
 }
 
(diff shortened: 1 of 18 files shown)


internal/container/selector.go
```diff
@@ -12,7 +12,7 @@ const (
 )
 
 type RefSelector struct {
-	ref       reference.Named
+	Ref       reference.Named
```
nicks: i'm confident that this shouldn't be a public field. it's not something that an API user would want or be able to use correctly. as a proof of point, the only place you did use this field is wrong and a bug.

internal/engine/image_and_cache_builder.go
```diff
@@ -96,7 +96,7 @@ func (icb *imageAndCacheBuilder) Build(ctx context.Context, iTarget model.ImageT
 	case model.CustomBuild:
 		ps.StartPipelineStep(ctx, "Building Dockerfile: [%s]", userFacingRefName)
 		defer ps.EndPipelineStep(ctx)
-		ref, err := icb.custb.Build(ctx, refToBuild, bd.Command)
+		ref, err := icb.custb.Build(ctx, iTarget.ConfigurationRef.Ref, bd.Command)
```
nicks: shouldn't this be the DeploymentRef? see other places where we do builds, e.g. https://github.com/windmilleng/tilt/blob/master/internal/engine/image_and_cache_builder.go#L37

internal/container/selector.go
maiamcc: Not a bug, it was desired behavior -- the idea was to pass in the ref _as specified in Tiltfile_ (rather than as stripped down for deployment) so we knew whether user gave it a tag or not. But I'm redoing this PR on advice from DB, so it's moot.

nicks: hmmm...can you point me to the spec that describes the expected behavior and how it's supposed to interact with substitute_repo?

nicks: (my understanding from previous conversations is we expected custom_build to push to a remote repo in some cases)

maiamcc:
This is how it looks in action: https://github.com/windmilleng/servantes/tree/bazel-tag-param

internal/container/selector.go
maiamcc: that behavior has been removed

internal/engine/build_and_deployer_test.go
```diff
@@ -582,6 +582,34 @@ func TestCustomBuildWithoutFastBuild(t *testing.T) {
...
+	if f.docker.BuildCount != 0 {
+		t.Errorf("Expected 0 docker build, actual: %d", f.docker.BuildCount)
+	}
+
+	if f.docker.PushCount != 1 {
+		t.Errorf("Expected 1 push to docker, actual: %d", f.docker.PushCount)
+	}
+
+	if strings.Contains(f.k8s.Yaml, sidecar.SyncletImageName) {
+		t.Errorf("Should not deploy the synclet for a custom build: %s", f.k8s.Yaml)
```
nicks: hmmm...this behavior is surprising to me? is there something i can read that explains why custom_build would affect synclet behavior?

internal/engine/build_and_deployer_test.go
```diff
@@ -582,6 +582,34 @@ func TestCustomBuildWithoutFastBuild(t *testing.T) {
...
+	_, err := f.bd.BuildAndDeploy(f.ctx, f.st, targets, store.BuildStateSet{})
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if f.docker.BuildCount != 0 {
+		t.Errorf("Expected 0 docker build, actual: %d", f.docker.BuildCount)
+	}
+
+	if f.docker.PushCount != 1 {
```
nicks: hmmm...afaict, this test would pass fine if the functionality in this pr was totally broken...right? please also add an assertion about the image that was pushed

internal/engine/build_and_deployer_test.go
maiamcc: Huhhh idk, I copied this from another test 🤷‍♀️ i think it's relevant b/c it's a custom build W/O fast build? (See [this test](https://github.com/windmilleng/tilt/blob/0c6e7dc834c5cc9ecea60ac9e69118a4d2750f60/internal/engine/build_and_deployer_test.go#L554). It's not relevant to this test case tho, I'll take it out.

internal/engine/build_and_deployer_test.go
maiamcc: nope, it fails without the rest of the functionality in this PR. if `CustomBuild.Tag` isn't set, [we generate our own temporary tag](https://github.com/windmilleng/tilt/blob/0c6e7dc834c5cc9ecea60ac9e69118a4d2750f60/internal/build/custom_builder.go#L36) and set an env variable `$TAG=imagename:temptag`; we then [check that the expected image and expected tag were built](https://github.com/windmilleng/tilt/blob/0c6e7dc834c5cc9ecea60ac9e69118a4d2750f60/internal/build/custom_builder.go#L66).

if i remove the changes to `CustomBuilder.Build`, test fails with:
```
Custom Build: Injecting Environment Variables
TAG=gcr.io/some-project-162817/sancho:tilt-build-1551202573
Running custom build cmd "true"

ERROR: fake docker client error: object not found (fakeClient.Images key: gcr.io/some-project-162817/sancho:tilt-build-1551202573)
```
(i.e. we spoof that the custom build command built `gcr.io/.../sancho:deterministic-tag`; but Build goes looking for `sancho:tilt-build-1551202573`, cant find it, thinks something has gone wrong.)

PR #1345, opened 2019-03-22
tiltfile: default parameter to read_file

jazzdan:


internal/tiltfile/tiltfile_test.go
```diff
@@ -2337,6 +2337,25 @@ default_registry('bar.com')
...
+result = read_file("this_file_does_not_exist", default="foo")
+docker_build('gcr.io/foo', 'foo')
+k8s_resource(str(result), 'foo.yaml')
+`
+
+	f.file("Tiltfile", tiltfile)
+
+	f.load()
+
+	f.assertNextManifest("foo",
```
dbentley: can you test that this_file_does_not_exist is added as a config file, too?

internal/tiltfile/files.go
```diff
@@ -188,7 +188,8 @@ func (s *tiltfileState) readFile(p localPath) ([]byte, error) {
 
 func (s *tiltfileState) skylarkReadFile(thread *starlark.Thread, fn *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
 	var path starlark.Value
-	err := starlark.UnpackArgs(fn.Name(), args, kwargs, "path", &path)
+	defaultReturn := ""
+	err := starlark.UnpackArgs(fn.Name(), args, kwargs, "path", &path, "default?", &defaultReturn)
```
dbentley: While you're here, is it easy to also add it to the json function? (Totally okay to submit without doing that)

jazzdan: Yeah I think I should just do that. Thanks for the reminder!

PR #1347, opened 2019-03-22
integration/oneup: use default_registry

dbentley:
Also changed the behavior to not spew docker.io/library into images.

### DIFF
--- a/integration/oneup/Tiltfile
+++ b/integration/oneup/Tiltfile
@@ -1,12 +1,10 @@
 # -*- mode: Python -*-
 
-fs = listdir('.')
-if 'tilt_personal.json' in fs:
-  # TODO(dbentley): use the default flag to read_json
-  settings = read_json('tilt_personal.json')
-else:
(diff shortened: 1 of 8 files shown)


integration/oneup/Tiltfile
```diff
@@ -1,4 +1,10 @@
 # -*- mode: Python -*-
 
-docker_build('gcr.io/windmill-test-containers/integration/oneup', '.')
+# If you get push errors, you can change the default_registry.
+# Create tilt_option.json with contents: {"default_registry": "gcr.io/my-personal-project"}
+# (with your registry inserted). tilt_option.json is gitignore'd, unlike Tiltfile
+default_registry(read_json('tilt_option.json', {})
+                 .get('default_registry', 'gcr.io/windmill-public-containers/servantes'))
```
nicks: fwiw, we don't give our CI access to windmill-public-containers, only windmill-test-containers

PR #1350, opened 2019-03-22
engine: add injectedRef + custom build test

landism:


internal/engine/image_build_and_deployer_test.go
```diff
@@ -279,14 +282,13 @@ func TestDeployUsesInjectRef(t *testing.T) {
...
-				// this test fails because dockerCLI.Images is never populated, and custom build tries to call ImageInspectRaw
-				t.Skip("custom build IBD tests not yet supported")
-			}
-
 			f := newIBDFixture(t, k8s.EnvGKE)
 			defer f.TearDown()
 
+			// for custom build
+			sha := digest.Digest("sha256:11cd0eb38bc3ceb958ffb2f9bd70be3fb317ce7d255c8a4c3f4af30e298aa1aab")
+			f.docker.Images["foo.com/gcr.io_some-project-162817_sancho:tilt-build-1546304461"] = types.ImageInspect{ID: string(sha)}
```
nicks: should we only do this in the custom build case?

PR #1367, opened 2019-03-26
web: use TypeScript for log app

jazzdan:
Mostly just putting a bunch of `any` types in places unless a better typt was obvious. We can fill them in later, or not at all.

@nicks if you think this will be more trouble than it's worth, that's fine by me and I'll close this PR. As a JavaScript foreigner I like the types safety blanket but I can live without it.

Also create-react-app makes this so easy! 🎉

### DIFF
--- a/web/package.json
+++ b/web/package.json
@@ -3,6 +3,13 @@
   "version": "0.1.0",
   "private": true,
   "dependencies": {
+    "@types/jest": "^24.0.11",
+    "@types/node": "^11.12.0",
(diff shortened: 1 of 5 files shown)


jazzdan:
huh, somehow my package.json changes didn't make it in to this commit. One moment.

nicks:
i think a little bit of tooling work is fine; as long as we're not spending a large % of the eng effort on it this cycle.

PR #1370, opened 2019-03-27
web: add LogPane component to HUD app

jazzdan:


web/src/index.js
```diff
@@ -17,7 +17,7 @@ let Main = () => {
           * New HUD work. HUD will eventually be promoted to the top-level component
           * and we'll delete App & friends
           */}
-        <Route exact path="/hud" component={HUD} />
+        <Route path="/hud" component={HUD} />
```
jazzdan: attn @nicks this is tricky. With exact matching on for this top level any path like `/hud/foo` or `/hud/foo/bar` would be routed to missing route. It seems to work as expected with exact matching off, I think this is due to the way we're doing this nested "switch" "statement"?

_grumble grumble react router_

web/src/LogPane.test.tsx
```diff
@@ -0,0 +1,20 @@
+import React from 'react'
+import ReactDOM from 'react-dom'
+import LogPane from './LogPane'
+import renderer from 'react-test-renderer';
+
+it('renders without crashing', () => {
+  let div = document.createElement('div')
+  Element.prototype.scrollIntoView = jest.fn()
```
nicks: what's this for?

web/src/LogPane.tsx
```diff
@@ -0,0 +1,94 @@
+import React, { Component, createRef } from 'react';
+import Ansi from "ansi-to-react";
+
+type AnsiLineProps = {
+  line: string
+  key: string
```
nicks: i'm surprised you need to specify key? all react child components have key, even if it's implicit

web/src/LogPane.tsx
```diff
@@ -0,0 +1,94 @@
...
+
+
+  constructor(props: LogPaneProps) {
+    super(props)
+
+    this.state = {
+      autoscroll: true,
+    }
+
+    // TODO(dmiller): do I need to bind refreshAutoScroll?
```
nicks: yes

in fact, i don't understand how this code can possibly work if you don't

web/src/index.js
nicks: this is the behavior we want tho? /hud/foo should go to the missing route. what am i missing?

nicks: oh, nm, ya i understand, this is right

web/src/LogPane.test.tsx
jazzdan: `scrollIntoView` doesn't exist on a DOM elements in the testing environment for some reason? Without this the tests fail with a type error from node complaining that that function doesn't exist.

web/src/LogPane.tsx
jazzdan: Yep you're right.

jazzdan:
There's a bug here where it looks like the logs are scrolling to the right instead of top down, will investigate.

jazzdan:
In the spirit of prototype and with the knowledge that this is exactly the problem that I am bad at solving I'm going to merge this with one outstanding problem that I'd like help with.

Screenshot_2019-03-28 Tilt

The logs are showing up and streaming just fine afaict, but they're streaming vertically instead of horizontally. Any idea what might be causing this? My visual diff of the CSS rules that got applied to those divs seemed identical but the problem might lie elsewhere.

PR #1375, opened 2019-03-28
Fix Statusbar crash

jazzdan:


web/src/Statusbar.test.tsx
```diff
@@ -0,0 +1,12 @@
+import React from 'react'
+import ReactDOM from 'react-dom'
+import renderer from 'react-test-renderer';
+import Statusbar from './Statusbar'
+
+it('renders without crashing if theres no last build', () => {
```
nicks: are you sure this exercises the bug? I thought the bug was in the StatusItem constructor, which this doesn't even use

PR #1378, opened 2019-03-28
model: helper function for topologically sorting targets

nicks:
Hello @jazzdan,

Please review the following commits I made in branch nicks/top:

44209ed8245a4f19845fbd4448142dfcf97fa7ee (2019-03-28 13:08:27 -0400)
model: helper function for topologically sorting targets

### DIFF
--- a/internal/model/target.go
+++ b/internal/model/target.go
@@ -1,6 +1,10 @@
 package model
 
-import "fmt"
+import (
+	"fmt"
+
+	"github.com/pkg/errors"
+)
 
 // An abstract build graph of targets and their dependencies.
 // Each target should have a unique ID.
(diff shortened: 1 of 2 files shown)


internal/model/target.go
```diff
@@ -82,3 +86,55 @@ func DedupeTargetIDs(ids []TargetID) []TargetID {
...
+
+// Create a topologically sorted list of targets. Returns an error
+// if the targets can't be topologically sorted. (e.g., there's a cycle).
+func TopologicalSort(targets []TargetSpec) ([]TargetSpec, error) {
+	targetMap := MakeTargetMap(targets)
+	result := make([]TargetSpec, 0, len(targets))
+	inResult := make(map[TargetID]bool, len(targets))
+	searching := make(map[TargetID]bool, len(targets))
+
+	var ensureInResult func(id TargetID) error
```
jazzdan: Why are these on two separate lines?

nicks: this is a common go faq
https://stackoverflow.com/questions/28099441/define-a-recursive-function-within-a-function-in-go

jazzdan: Ah! TIL thank you

PR #1380, opened 2019-03-28
engine: make target queue more like a real build graph. fixes some bugs in dirtyness checking w/ dependent images

nicks:
Hello @jazzdan, @maiamcc,

Please review the following commits I made in branch nicks/ch2033/deps:

ab92cf90cb49e30c567c9c20c9dc016b4eacf457 (2019-03-28 13:20:23 -0400)
engine: make target queue more like a real build graph. fixes some bugs in dirtyness checking w/ dependent images

### DIFF
--- a/internal/engine/target_queue.go
+++ b/internal/engine/target_queue.go
@@ -7,60 +7,120 @@ import (
 	"github.com/windmilleng/tilt/internal/store"
 )
 
-// A little data structure to help iterate through targets in dependency order.
+// Allows the caller to inject its own build strategy for dirty targets.
+type BuildHandler func(
+	target model.TargetSpec,
+	state store.BuildState,
+	depResults []store.BuildResult) (store.BuildResult, error)
+
+// A little data structure to help iterate through dirty targets in dependency order.
 type TargetQueue struct {
-	targets []model.TargetSpec
+	sortedTargets []model.TargetSpec
+
+	// The state from the previous build.
+	// Contains files-changed so we can do incremental builds.
+	state store.BuildStateSet
+
+	// The results of this build.
 	results store.BuildResultSet
+
+	// Whether the target itself needs a rebuilt, either because it has dirty files
+	// or has never been built before.
+	//
+	// A target with dirty files might be able to use the files changed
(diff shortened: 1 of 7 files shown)


internal/engine/target_queue.go
```diff
@@ -7,37 +7,113 @@ import (
 	"github.com/windmilleng/tilt/internal/store"
 )
 
-// A little data structure to help iterate through targets in dependency order.
+// Allows the caller to inject its own build strategy for dirty targets.
```
nicks: fwiw @maiamcc i think an interface that looks like this one will eventually replace BuildAndDeployer...

we'll define a bunch of functions that take
1) a single target spec
2) the current build state of that target
3) the results of the dependencies

we'll have different handlers for building them, and a central build graph (like TargetQueue) that calls the build handlers whenever a target is marked dirty

internal/engine/target_queue.go
```diff
@@ -7,37 +7,113 @@ import (
...
+			result++
 		}
 	}
-	return true
+	return result
 }
 
-func (q *TargetQueue) SetResult(id model.TargetID, result store.BuildResult) {
-	q.results[id] = result
+func (q *TargetQueue) BuildDirty(handler BuildHandler) error {
```
jazzdan: This function name feels weird to me. It shouldn't only be used if the target has any dirty dependencies, it figured that out. Maybe:
`BuildWithState`?

I don't feel strongly though, it's a tough one to name.

nicks: i changed it to RunBuilds

i'm still straightening out in my mind what this interface should look like. you can look at how this is currently handled in Bazel
https://github.com/bazelbuild/bazel/tree/1f684e1b87cd8881a0a4b33e86ba66743e32d674/src/main/java/com/google/devtools/build/lib/actions

where it has an ActionGraph class that returns "here's what you need to do to build this target" and an ActionGraphVisitor that does the visiting/ result accumulation

the way they do it is definitely way too many levels of abstraction for what we need, but it's a good place to look for inspiration

PR #1382, opened 2019-03-28
web: resource specific log routes

jazzdan:
I'd like to:
* Add some more types
* Add a test

But this solves the problem for now.


web/src/HUD.tsx
```diff
@@ -59,11 +71,21 @@ class HUD extends Component {
...
     }
 
+    let LogsRoute = (props: HudProps) => {
+      let name = props.match.params ? props.match.params.name : ""
+      let logs = ""
+      if (name !== "") {
+        let r = view.Resources.find(r => r.Name === name)
+        logs = r ? r.CombinedLog : ""
+      }
+       return <LogPane log={logs} />
```
nicks: weird indentation?

web/src/HUD.tsx
```diff
@@ -73,6 +95,10 @@ class HUD extends Component {
         </Switch>
 
         <Statusbar items={statusItems} toggleSidebar={this.toggleSidebar}  />
+        <Switch>
```
nicks: shouldn't this replace the LogPane down below?

nicks: or just merge this `Switch` into that `Switch`

web/src/HUD.tsx
jazzdan: Good catch

PR #1384, opened 2019-03-28
engine: inject synclet if image target has a LiveUpdate [ch2040]

maiamcc:
Hello @landism, @nicks,

Please review the following commits I made in branch maiamcc/synclet-inject:

5c147e66573148b9c31eb995ebe36e7d377e6346 (2019-03-28 17:29:48 -0400)
engine: inject synclet if image target has a LiveUpdate [ch2040]

### DIFF
--- a/internal/engine/build_and_deployer_live_update_test.go
+++ b/internal/engine/build_and_deployer_live_update_test.go
@@ -60,11 +60,6 @@ func runTestCase(t *testing.T, f *bdFixture, tCase testCase) {
 	f.assertContainerRestarts(tCase.expectDockerRestartCount)
 	assert.Equal(t, tCase.expectSyncletHotReload, f.sCli.UpdateContainerHotReload, "synclet hot reload")
 
-	id := manifest.ImageTargetAt(0).ID()
-	_, hasResult := result[id]
-	assert.True(t, hasResult)
-	assert.Equal(t, k8s.MagicTestContainerID, result.OneAndOnlyContainerID().String())
-
 	if tCase.expectK8sDeploy {
 		expectedYaml := "image: gcr.io/some-project-162817/sancho:tilt-11cd0b38bc3ceb95"
 		if !strings.Contains(f.k8s.Yaml, expectedYaml) {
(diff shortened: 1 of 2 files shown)


internal/engine/build_and_deployer_live_update_test.go
```diff
@@ -75,17 +75,18 @@ func runTestCase(t *testing.T, f *bdFixture, tCase testCase) {
...
-
 	if tCase.expectK8sDeploy {
 		expectedYaml := "image: gcr.io/some-project-162817/sancho:tilt-11cd0b38bc3ceb95"
 		if !strings.Contains(f.k8s.Yaml, expectedYaml) {
 			t.Errorf("Expected yaml to contain %q. Actual:\n%s", expectedYaml, f.k8s.Yaml)
 		}
 		assert.Equal(t, tCase.expectSyncletDeploy, strings.Contains(f.k8s.Yaml, sidecar.SyncletImageName), "expected synclet-deploy = %t (deployed yaml was: %s)", tCase.expectSyncletDeploy, f.k8s.Yaml)
+	} else {
+		// Did a container build, expect result to have the container ID we operated on
+		id := manifest.ImageTargetAt(0).ID()
```
nicks: i don't this move is correct? shouldn't we have a result in both cases?

internal/engine/build_and_deployer_live_update_test.go
```diff
@@ -75,17 +75,18 @@ func runTestCase(t *testing.T, f *bdFixture, tCase testCase) {
...
-	assert.Equal(t, k8s.MagicTestContainerID, result.OneAndOnlyContainerID().String())
-
 	if tCase.expectK8sDeploy {
 		expectedYaml := "image: gcr.io/some-project-162817/sancho:tilt-11cd0b38bc3ceb95"
 		if !strings.Contains(f.k8s.Yaml, expectedYaml) {
 			t.Errorf("Expected yaml to contain %q. Actual:\n%s", expectedYaml, f.k8s.Yaml)
 		}
 		assert.Equal(t, tCase.expectSyncletDeploy, strings.Contains(f.k8s.Yaml, sidecar.SyncletImageName), "expected synclet-deploy = %t (deployed yaml was: %s)", tCase.expectSyncletDeploy, f.k8s.Yaml)
+	} else {
+		// Did a container build, expect result to have the container ID we operated on
```
nicks: fwiw, i think DockerCompose builds also return a container ID

internal/engine/build_and_deployer_live_update_test.go
maiamcc: ah, i see. Yeah we should get a `result` in all cases, moving that check out of the conditional. (we only expect it to have a container ID if we did a docker compose build or a synclet/localContainerBuild tho.)

internal/engine/build_and_deployer_live_update_test.go
maiamcc: yyyeahhh. i'll leave a comment.

maiamcc: hmm actually I'm not sure that's relevant.

PR #1388, opened 2019-03-29
tiltfile: add live_update

landism:


internal/tiltfile/live_update.go
```diff
@@ -0,0 +1,297 @@
...
+func (l liveUpdateRestartContainerStep) declarationPosition() syntax.Position { return l.position }
+func (l liveUpdateRestartContainerStep) liveUpdateStep()                      {}
+
+func (s *tiltfileState) liveUpdateWorkDir(thread *starlark.Thread, fn *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
+	var workDir string
+	if err := starlark.UnpackArgs(fn.Name(), args, kwargs, "dir", &workDir); err != nil {
+		return nil, err
+	}
+
+	if !filepath.IsAbs(workDir) {
```
maiamcc: ehhh i'm not sure we need this check? if we got a relative path i'd just expect us to eval it relative to `/`? but maybe this is safer 🤷‍♀️

internal/tiltfile/live_update.go
```diff
@@ -0,0 +1,297 @@
...
+		return starlark.None, fmt.Errorf("work_dir path '%s' is not absolute", workDir)
+	}
+
+	ret := liveUpdateWorkDirStep{
+		value:    workDir,
+		position: thread.TopFrame().Position(),
+	}
+	s.unconsumedLiveUpdateSteps = append(s.unconsumedLiveUpdateSteps, ret)
```
maiamcc: nice! ...is there some way we can have a common function that you use when implementing new Step types that does this automatically, so ppl writing this code don't forget to track the step here? (otherwise, maybe a big-ass note by where we define the types?)

internal/tiltfile/live_update.go
```diff
@@ -0,0 +1,297 @@
...
+
+func liveUpdateStepToModel(l liveUpdateStep, workDir string) (model.LiveUpdateStep, error) {
+	switch x := l.(type) {
+	case liveUpdateWorkDirStep:
+		return model.LiveUpdateWorkDirStep(x.value), nil
+	case liveUpdateSyncStep:
+		remotePath := x.remotePath
+		if !filepath.IsAbs(remotePath) {
+			if workDir == "" {
+				if !filepath.IsAbs(x.remotePath) {
```
maiamcc: ...uh? didn't you already do this check?

internal/tiltfile/live_update.go
```diff
@@ -0,0 +1,297 @@
...
+					return model.LiveUpdateSyncStep{}, fmt.Errorf("step '%s' must either follow a work_dir or have an absolute remote_path", x.String())
```
maiamcc: again, we could also decide that if there's no workdir, we just treat workdir as "/" -- that seems more intuitive to me

internal/tiltfile/live_update.go
```diff
@@ -0,0 +1,297 @@
...
+	default:
+		return nil, fmt.Errorf("internal error - unknown liveUpdateStep '%v' of type '%T', declared at %s", l, l, l.declarationPosition().String())
+	}
+}
+
+func liveUpdateToModel(l liveUpdateDef) (model.LiveUpdate, error) {
+	return model.NewLiveUpdate(l.steps, l.fullRebuildTriggers)
+}
+
+func (s *tiltfileState) consumeLiveUpdateStep(stepToConsume liveUpdateStep) {
```
maiamcc: fwiw, i found this really hard to reason about. what about:

```
func (s *tiltfileState) consumeLiveUpdateStep(stepToConsume liveUpdateStep) {
	for i, step := s.unconsumedLiveUpdateSteps {
		if step.declarationPosition() == stepToConsume.declarationPosition() {
			n := append([]string{}, a[:i]...)
			n = append(ret, a[i+1:]...)
			s.unconsumedLiveUpdateSteps = n
			break
		}
	}
}
```

internal/tiltfile/live_update.go
```diff
@@ -0,0 +1,297 @@
...
+		"image", &dockerRef,
+		"steps", &steps,
+		"full_rebuild_triggers?", &fullRebuildTriggers,
+	); err != nil {
+		return nil, err
+	}
+
+	var modelSteps []model.LiveUpdateStep
+	stepSlice := starlarkValueOrSequenceToSlice(steps)
+	workDir := ""
```
maiamcc: ...yeah, looking at this again, probs default to "/" here

internal/tiltfile/live_update.go
```diff
@@ -0,0 +1,297 @@
...
+	s.liveUpdates[dockerRef] = &liveUpdateDef{
+		steps:               modelSteps,
+		fullRebuildTriggers: frtStrings,
+	}
+
+	return starlark.None, nil
+}
+
+func (s *tiltfileState) validateLiveUpdates() error {
+	if len(s.unconsumedLiveUpdateSteps) > 0 {
```
maiamcc: niiiice!

internal/tiltfile/live_update.go
```diff
@@ -0,0 +1,297 @@
...
+
+	"github.com/windmilleng/tilt/internal/model"
+)
+
+type liveUpdateDef struct {
+	imageName           string
+	steps               []model.LiveUpdateStep
+	fullRebuildTriggers []string
+
+	// whether this has been matched to a deployed image
```
maiamcc:
```suggestion
	// whether this has been matched to an image we know how to build
```

internal/tiltfile/tiltfile_test.go
```diff
@@ -2764,6 +2764,11 @@ func (f *fixture) assertNextManifest(name string, opts ...interface{}) model.Man
 						}
 						matcher.fb.checkMatchers(f, m, *dbInfo.FastBuild)
 					}
+				case model.LiveUpdate:
+					lu := image.MaybeLiveUpdateInfo()
+					if assert.NotNil(f.t, lu) {
+						assert.Equal(f.t, matcher, *lu)
```
maiamcc: this is fine

internal/tiltfile/live_update.go
landism: nnnnno?

internal/tiltfile/live_update.go
landism: changed to "a deployed image we know how to build"

unfortunately we don't match live updates to builds until we match builds to deploys

maiamcc: oh shoot you're right, it DOES need to be a deployed image. nvm.

internal/tiltfile/live_update.go
landism: The only way I can think of to do this automatically is to wrap `s.builtin` to check the return value of every tiltfile function, and register any `liveUpdateStep`s it finds, but that feels gross. I've added a comment above `liveUpdateStep`.

PR #1389, opened 2019-03-29
web: actually set state in autoscroll timeout

jazzdan:
Story details: https://app.clubhouse.io/windmill/story/2049


jazzdan:
We might want to use [this new API](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API) for detecting if elements are in view instead, but this works for now.

PR #1394, opened 2019-03-29
integration: live_update test

landism:


integration/live_update_test.go
```diff
@@ -0,0 +1,46 @@
...
+	f := newK8sFixture(t, "live_update")
+	defer f.TearDown()
+
+	f.TiltWatch()
+
+	// ForwardPort will fail if all the pods are not ready.
+	//
+	// For the purposes of this integration tests, we want the test to fail if the
+	// Pod is re-created (rather than getting updated in-place).  We deliberately
+	// don't use Tilt-managed port-forwarding because it would auto-connect to the
```
maiamcc: ...wait, are we NOT using tilt-managed port-forwarding?
```
k8s_resource('live-update', 'deploy.yaml', port_forwards=8100)
```

landism: copypasstaaaaaa

It looks like the test sets it up but then doesn't use it. I've removed it from the tiltfile to reduce confusion.

PR #1395, opened 2019-03-29
web: typescript all the things

jazzdan:


nicks:
wait, i thought we weren't committing node_modules to the repo?

package.json
```diff
@@ -0,0 +1,5 @@
+{
+  "dependencies": {
+    "@types/enzyme-adapter-react-16": "^1.0.5"
```
nicks: also, this should be under web, not in the root of the repo

jazzdan:
@nicks whoops, yeah I accidentally `yarn add`'d at the top level which is what caused the node_modules to appear.

PR #1396, opened 2019-04-01
live_update: error if syncing from outside of docker context

landism:
(also added `ospath.IsChild` and updated appropriate calls to `ospath.Child`, because that was annoying)


internal/tiltfile/docker.go
```diff
@@ -190,6 +190,18 @@ func (s *tiltfileState) maybeLiveUpdate(image *dockerImage) (*model.LiveUpdate,
 	if lu, ok := s.liveUpdates[image.configurationRef.String()]; ok {
 		lu.matched = true
 		ret, err := liveUpdateToModel(*lu)
+
+		// if it's a docker build, verify that all sync steps are from within the docker build context
```
maiamcc:
```suggestion
		// if it's a docker build + live update, verify that all sync steps are from within the docker build context
```

PR #1397, opened 2019-04-01
hud: remove logs modals from tui

jazzdan:
Now looks like this:

Screen Shot 2019-04-01 at 11 49 16 AM


jazzdan:
this is on hold for now

PR #1401, opened 2019-04-01
engine: docker compose runtime logs are appended to global log

jazzdan:


internal/engine/upper.go
```diff
@@ -935,6 +935,7 @@ func handleDockerComposeLogAction(state *store.EngineState, action DockerCompose
 
 	dcState, _ := ms.ResourceState.(dockercompose.State)
 	ms.ResourceState = dcState.WithCurrentLog(model.AppendLog(dcState.CurrentLog, action, state.LogTimestamps))
```
nicks: this should also be on the CombinedLog right?

jazzdan: @nicks [`state.Log`](https://github.com/windmilleng/tilt/blob/4f0e74b987d728f1cde3ce10cabb75a1a6f5666f/internal/store/engine_state.go#L55) is the combined log?

internal/engine/upper.go
```diff
@@ -935,6 +935,7 @@ func handleDockerComposeLogAction(state *store.EngineState, action DockerCompose
...
+	state.Log = model.AppendLog(dcState.CurrentLog, action, state.LogTimestamps)
```
landism: IIUC this line is currently replacing the full tilt log with the runtime log of one DC service, and this should be `ms.CombinedLog`, not `state.Log`.

jazzdan: Ohh yes, thank you

internal/engine/upper.go
```diff
@@ -935,6 +935,7 @@ func handleDockerComposeLogAction(state *store.EngineState, action DockerCompose
 
 	dcState, _ := ms.ResourceState.(dockercompose.State)
 	ms.ResourceState = dcState.WithCurrentLog(model.AppendLog(dcState.CurrentLog, action, state.LogTimestamps))
+	state.Log = model.AppendLog(state.Log, action, state.LogTimestamps)
```
landism: 1. This change is modifying the global tilt log, which is already working for this case, and this change would result in the message being appended twice.
2. This should instead be `ms.CombinedLog = model.AppendLog(ms.CombinedLog, action, state.LogTimestamps)`.

internal/engine/upper_test.go
```diff
@@ -1955,6 +1955,9 @@ func TestDockerComposeRecordsRunLogs(t *testing.T) {
 	f.withManifestState(m.ManifestName(), func(st store.ManifestState) {
 		assert.Contains(t, st.DCResourceState().Log(), expected)
```
landism: instead of the addition below, I think we want this here:
```
assert.Contains(t, st.CombinedLog, expected)
```

internal/engine/upper.go
nicks: here's an example of what i mean:
https://github.com/windmilleng/tilt/blob/master/internal/engine/upper.go#L777
why doesn't this code the same as that?

internal/engine/upper.go
nicks: i'm pretty sure this logs to state.Log twice, because it already gets logged to the global log writer here:
https://github.com/windmilleng/tilt/blob/master/internal/engine/docker_compose_log_manager.go#L125

internal/engine/upper_test.go
```diff
@@ -1955,6 +1955,9 @@ func TestDockerComposeRecordsRunLogs(t *testing.T) {
...
 	})
+	f.withState(func(st store.EngineState) {
+		assert.Contains(t, st.Log.String(), expected)
```
nicks: can you modify the test to ensure that this line appears exactly once?

internal/engine/upper.go
jazzdan: Ah I see, I misunderstood what CombinedLog meant. Thanks!

internal/engine/upper_test.go
```diff
@@ -1954,6 +1955,7 @@ func TestDockerComposeRecordsRunLogs(t *testing.T) {
 	// recorded on manifest state
 	f.withManifestState(m.ManifestName(), func(st store.ManifestState) {
 		assert.Contains(t, st.DCResourceState().Log(), expected)
+		assert.Equal(t, 1, len(expectedRegex.FindAllStringIndex(st.CombinedLog.String(), -1)))
```
nicks: i think you're looking for `strings.Count`

PR #1402, opened 2019-04-01
engine: change how we store/check run.triggers (so can better implement full_rebuild_trigger)"

maiamcc:
Hello @landism, @nicks,

Please review the following commits I made in branch maiamcc/run-triggers:

2a4675924f9d73972792dde5c89e37553295fa9b (2019-04-01 17:14:08 -0400)
better live update run.trigger tests

9e91704ff32f69807e254b035499b543e6168c3c (2019-04-01 17:06:47 -0400)
change how we store triggers

b94bd9e83978aea80516956313e3446bd874acee (2019-04-01 13:45:29 -0400)
pull up getChangedFiles

964da22a9ebe1acde6d963ec84651d51a22a1391 (2019-04-01 13:45:29 -0400)
red test for local container build

9a1ffb5839fad455878e8419d0cae914ef581e7b (2019-04-01 13:45:29 -0400)
generalize run.trigger matching

### DIFF
--- a/internal/build/boil_runs.go
+++ b/internal/build/boil_runs.go
@@ -7,24 +7,20 @@ import (
 
 func BoilRuns(runs []model.Run, pathMappings []PathMapping) ([]model.Cmd, error) {
 	res := []model.Cmd{}
+	localPaths := PathMappingsToLocalPaths(pathMappings)
 	for _, run := range runs {
 		if run.Triggers == nil {
 			res = append(res, run.Cmd)
 			continue
 		}
-		matcher, err := ignore.CreateRunMatcher(run)
+
+		anyMatch, err := ignore.MatchesAnyPaths(run.Triggers, localPaths, run.BaseDirectory)
 		if err != nil {
-			return []model.Cmd{}, err
+			return nil, err
 		}
-		for _, pm := range pathMappings {
-			matches, err := matcher.Matches(pm.LocalPath, false)
-			if err != nil {
-				return []model.Cmd{}, err
-			}
-			if matches {
-				res = append(res, run.Cmd)
-				break
-			}
+
+		if anyMatch {
+			res = append(res, run.Cmd)
 		}
 	}
 	return res, nil
--- a/internal/build/boil_runs.go
+++ b/internal/build/boil_runs.go
@@ -9,12 +9,12 @@ func BoilRuns(runs []model.Run, pathMappings []PathMapping) ([]model.Cmd, error)
 	res := []model.Cmd{}
 	localPaths := PathMappingsToLocalPaths(pathMappings)
 	for _, run := range runs {
-		if run.Triggers == nil {
+		if run.Triggers.Empty() {
 			res = append(res, run.Cmd)
 			continue
 		}
 
-		anyMatch, err := ignore.MatchesAnyPaths(run.Triggers, localPaths, run.BaseDirectory)
+		anyMatch, err := ignore.AnyMatchGlobs(localPaths, run.Triggers)
 		if err != nil {
 			return nil, err
 		}
--- a/internal/build/boil_runs.go
+++ b/internal/build/boil_runs.go
@@ -1,7 +1,6 @@
 package build
 
 import (
-	"github.com/windmilleng/tilt/internal/ignore"
 	"github.com/windmilleng/tilt/internal/model"
 )
 
@@ -14,7 +13,7 @@ func BoilRuns(runs []model.Run, pathMappings []PathMapping) ([]model.Cmd, error)
 			continue
 		}
 
-		anyMatch, err := ignore.AnyMatchGlobs(localPaths, run.Triggers)
+		anyMatch, err := model.AnyMatchGlobs(localPaths, run.Triggers)
 		if err != nil {
 			return nil, err
 		}
--- a/internal/build/boil_runs.go
+++ b/internal/build/boil_runs.go
@@ -13,7 +13,7 @@ func BoilRuns(runs []model.Run, pathMappings []PathMapping) ([]model.Cmd, error)
 			continue
 		}
 
-		anyMatch, err := model.AnyMatchGlobs(localPaths, run.Triggers)
+		anyMatch, err := run.Triggers.AnyMatch(localPaths)
 		if err != nil {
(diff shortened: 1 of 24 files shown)


internal/build/boil_runs.go
```diff
@@ -7,24 +7,20 @@ import (
...
+	localPaths := PathMappingsToLocalPaths(pathMappings)
 	for _, run := range runs {
-		if run.Triggers == nil {
+		if run.Triggers.Empty() {
 			res = append(res, run.Cmd)
 			continue
 		}
-		matcher, err := ignore.CreateRunMatcher(run)
+
+		anyMatch, err := ignore.AnyMatchGlobs(localPaths, run.Triggers)
```
nicks: i feel like the existing `ignore` package tends to use the terminology "Match" and "Pattern" a lot. The "glob" terminology seems new? it would be nice if this integrated better with the existing naming conventions

internal/build/image_builder.go
```diff
@@ -130,8 +130,8 @@ func (d *dockerImageBuilder) applyLabels(df dockerfile.Dockerfile, buildMode doc
 func (d *dockerImageBuilder) addConditionalRuns(df dockerfile.Dockerfile, runs []model.Run, paths []PathMapping) (dockerfile.Dockerfile, []model.Run, error) {
 	consumed := 0
 	for _, run := range runs {
-		if run.Triggers == nil {
-			break
+		if run.Triggers.Empty() {
+			continue
```
nicks: i'm pretty sure this change is incorrect? This will re-order runs, which is not what you want.

nicks: can you add a test for this behavior?

internal/engine/local_container_build_and_deployer.go
```diff
@@ -68,39 +68,42 @@ func (cbd *LocalContainerBuildAndDeployer) BuildAndDeploy(ctx context.Context, s
...
 	}
 
-	var syncs []model.Sync
+	var changedFiles []build.PathMapping
 	var runs []model.Run
 	var hotReload bool
 
 	if fbInfo := iTarget.MaybeFastBuildInfo(); fbInfo != nil {
-		syncs = fbInfo.Syncs
+		changedFiles, err = build.FilesToPathMappings(state.FilesChanged(), fbInfo.Syncs)
```
nicks: i don't understand why this moved around? it seems like the old code was better

internal/build/image_builder.go
maiamcc: oh oops, that wasn't supposed to stay in this PR 😅

internal/model/manifest.go
```diff
@@ -188,18 +188,35 @@ type LocalGitRepo struct {
 
 func (LocalGitRepo) IsRepo() {}
 
+type Globset struct {
```
landism: nit: capital S `GlobSet`

internal/tiltfile/live_update.go
```diff
@@ -63,8 +63,13 @@ var _ starlark.Value = liveUpdateRunStep{}
 var _ liveUpdateStep = liveUpdateRunStep{}
 
 func (l liveUpdateRunStep) String() string {
-	return fmt.Sprintf("run step: %s", strconv.Quote(l.command))
+	s := fmt.Sprintf("run step: %s", strconv.Quote(l.command))
+	if len(l.triggers) > 0 {
```
landism: fwiw, I had left these out since I figured the command would probably suffice to identify the step to the user and the triggers might just feel like noise, but I don't really feel strongly either way

(and I should have at least commented that intent)

internal/engine/build_and_deployer_live_update_test.go
```diff
@@ -229,6 +261,28 @@ func TestLiveUpdateDockerBuildDeploysSynclet(t *testing.T) {
 	runTestCase(t, f, tCase)
 }
 
+func _TestLiveUpdateLocalContainerFullBuildTrigger(t *testing.T) {
```
maiamcc: just leaving this in, will be actually implemented in the next commit

internal/model/manifest.go
```diff
@@ -192,14 +192,17 @@ type Run struct {
...
-	// Directory the Triggers are relative to
-	BaseDirectory string
+	// If specified, we only run the Cmd if the changed file matches a trigger.
+	Triggers PathSet
 }
 
-func (r Run) WithTriggers(triggers []string) Run {
-	r.Triggers = triggers
+func (r Run) WithTriggers(paths []string, baseDir string) Run {
+	if len(paths) > 0 {
```
nicks: i don't think this check is right? if len(paths) == 0, it should write an empty PathSet, instead of keeping the previous value.

internal/model/matcher.go
```diff
@@ -43,6 +45,51 @@ func NewSimpleFileMatcher(paths ...string) (fileMatcher, error) {
...
+		Paths:         paths,
+		BaseDirectory: baseDir,
+	}
+}
+
+func (ps PathSet) Empty() bool { return len(ps.Paths) == 0 }
+
+// AnyMatch returns true if any of the given filepaths match any paths contained in the pathset.
+func (ps PathSet) AnyMatch(paths []string) (bool, error) {
+	matcher := NewRelativeFileMatcher(ps.BaseDirectory, ps.Paths...)
```
nicks: i thought the plan was to use the DockerPatternMatcher to preserve the existing behavior? this changes the behavior...right?

maiamcc: Iiiiit _does_, but it still supports all of the behavior we ADVERTISE ourselves as supporting. the fact that we use dockerIgnore under the hood is really weird, considering that we only ever talk about exact path matches.

internal/tiltfile/live_update.go
maiamcc: ah gotcha. I'll leave them in for visibility purposes -- i.e. there's no other way for the user to check that their triggers got correctly populated.

internal/model/manifest.go
maiamcc: fair! (I wanted to avoid a pathset with empty Paths and populated baseDir, cuz that seems weird to me, but 🤷‍♀️ )

PR #1405, opened 2019-04-02
web: fix memory leak warning

jazzdan:
Following the advice here https://www.robinwieruch.de/react-warning-cant-call-setstate-on-an-unmounted-component/

### DIFF
--- a/web/src/HUD.tsx
+++ b/web/src/HUD.tsx
@@ -57,6 +57,9 @@ type HudState = {
 // The Main HUD view, as specified in
 // https://docs.google.com/document/d/1VNIGfpC4fMfkscboW0bjYYFJl07um_1tsFrbN-Fu3FI/edit#heading=h.l8mmnclsuxl1
 class HUD extends Component<HudProps, HudState> {
+  // NOTE(dmiller): this can't be named "isMounted" because that
+  // is a method that already exists on React.Component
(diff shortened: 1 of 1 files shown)


web/src/AppController.ts
```diff
@@ -32,6 +33,14 @@ class AppController {
     this.loadCount = 0
   }
 
+  componentDidMount() {
```
nicks: where do you expect these methods to be called? Note that AppController isn't a react component.

jazzdan: D'oh...

/me drinks more coffee

web/src/HUD.tsx
```diff
@@ -82,10 +85,14 @@ class HUD extends Component<HudProps, HudState> {
 
   componentDidMount() {
     this.controller.createNewSocket()
+    this._isMounted = true
   }
 
   componentWillUnmount() {
-    this.controller.dispose()
+    if (this._isMounted) {
+      this.controller.dispose()
```
nicks: sorry, i still don't understand this pr. why should the component need to be mounted to dispose the controller?

nicks:
can you give more context on what problem you're trying to fix? i think you're very confused about something.

jazzdan:
I'm going to close this because I can't reproduce the error message.

PR #1409, opened 2019-04-02
web: fix sidebar flickering

jazzdan:
Okay, let's dive in to some React and JavaScript truisms:

React components re-render under three conditions:
1. The component's state changed
2. The component's prop changed
3. The component's parent changed (See 1 and 2)

What was causing the flicking was a combination of 1, 2 and 3. While the
Sidebar component's props _weren't_ changing, its grandparents state was which
caused an accidental change to the Sidebar's parent's props, which caused
the sidebar to re-render.

Let's examine this flow in detail. First let's establish the component hierarchy:

[HUD] -> [Route] -> [Sidebar]

1. We get new information from the websocket, which changes HUD's state.
2. That change in state causes the HUD's render function to be re-run
3. This creates a new function, `SidebarRoute` which we pass to Route.
4. In JavaScript the following holds with regards to functions and references:
```
let x, y = () => bar, () => bar
assert(x !== y)
```
5. React uses equality to check if props have changed, and due to the
above property Route's properties have changed.
6. This causes Route to re-render which in turn causes Sidebar to re-render,
even though none of _its_ props or state have changed.

To solve this let's wrap references to functions that are passed to
components in anonymous functions so that they don't change every time
the render function is called.

### DIFF
--- a/web/src/HUD.test.tsx
+++ b/web/src/HUD.test.tsx
@@ -95,6 +95,32 @@ it("opens sidebar on click", async () => {
   expect(sidebar.hasClass("is-open")).toBe(true)
 })
 
+it("doesn't re-render the sidebar when the logs change", async () => {
+  const hud = mount(emptyHUD())
+  let resourceView = oneResourceView()
+  hud.setState({ View: resourceView })
+  let oldDOMNode = hud.find(".Sidebar").getDOMNode()
+  resourceView.Resources[0].PodLog += "hello world\n"
+  hud.setState({ View: resourceView })
+  let newDOMNode = hud.find(".Sidebar").getDOMNode()
+
+  expect(oldDOMNode).toBe(newDOMNode)
+})
+
+it("does re-render the sidebar when the resource list changes", async () => {
+  const hud = mount(emptyHUD())
+
+  let resourceView = oneResourceView()
+  hud.setState({ View: resourceView })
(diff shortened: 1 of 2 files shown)


jazzdan:
@yuindustries @nicks I wrote up this long explanation not because it necessarily deserves it but because I wanted to double check my understanding. Let me know if I got anything wrong?

nicks:
can you add a test?

jazzdan:
@nicks hmm, I'm guessing I can but it might take me a minute to figure out the right way to go about it.

nicks:
try find(selector).getDOMNode() and see if the node changed.

nicks:
(i think a well-written test here is more valuable than all the armchair reasoning in the world)

jazzdan:
Just leaving some notes for myself/others:

```
> let x = () => null
undefined
> let y = () => null
undefined
> x === y
false
> x === x
true
> x === (() => null)
false
> (() => null) === (() => null)
false
> let props1 = {x: x, y: []}
undefined
> let props2 = {x: y, y: ["a"]}
undefined
> props1 === props2
false
> props1 === props1
true
> let props3 = {x: y, y: ["a", "b"]}
undefined
> props2 === props3
false
```

jazzdan:
@yuindustries @nicks ready for review now!

PR #1410, opened 2019-04-02
hud: better control of webpack shutdown

nicks:
Hello @jazzdan,

Please review the following commits I made in branch nicks/ch2025/webpack:

132d5926a755d49301300b54ea51b050ddd96afe (2019-04-02 13:06:27 -0400)
hud: better control of webpack shutdown

### DIFF
--- a/internal/store/subscriber.go
+++ b/internal/store/subscriber.go
@@ -17,31 +17,62 @@ type Subscriber interface {
 	OnChange(ctx context.Context, st RStore)
 }
 
-type subscriberEntry struct {
-	subscriber Subscriber
-	mu         sync.Mutex
+// Some subscribers need to do teardown.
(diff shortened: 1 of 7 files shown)


internal/store/subscriber.go
```diff
@@ -37,13 +48,33 @@ func (l *subscriberList) Remove(s Subscriber) error {
 	for i, current := range l.subscribers {
 		if s == current.subscriber {
 			l.subscribers = append(l.subscribers[:i], l.subscribers[i+1:]...)
+			if l.setup {
+				current.maybeTeardown(context.TODO())
```
jazzdan: Should we fill out this TODO or are you leaving it for now?

nicks: good point! updated it to use a real context

PR #1411, opened 2019-04-02
tiltfile: make live_update work with unqualified image names

landism:
When we say `docker_build("foo", "foo")`, we normalize that image name and store it as `"docker.io/library/foo"`.

These image names were not working with `live_update` because `live_update` was just doing straight string matching on the image name we gave it (e.g., if we had `docker_build("foo", "foo")`, we'd have to say `live_update("docker.io/library/foo", ...)`).

The obvious options are to match against the unqualified name or to also normalize the image names stored in live_update.

When we're using `"foo"` with `default_registry`, we're going to transform that into, e.g., `"gcr.io/windmilleng/foo"` before publishing, and it feels like a bug if we ever show `"docker.io/library/foo"` to the user, so I've opted to not store `live_update` image names in that format.


internal/tiltfile/live_update_test.go
```diff
@@ -134,40 +134,47 @@ live_update('gcr.io/bar',
...
 	}
 	cbManifestOpt := func(f *fixture, expectedLu model.LiveUpdate) {
-		f.assertNextManifest("foo", cb(image("gcr.io/foo"), expectedLu))
+		f.assertNextManifest("foo", cb(imageNormalized("foo"), expectedLu))
 	}
 
 	tests := []testCase{
-		{"docker_build", "docker_build('gcr.io/foo', 'foo')", dbManifestOpt},
-		{"custom_build", "custom_build('gcr.io/foo', 'docker build -t $TAG foo', ['foo'])", cbManifestOpt},
+		{"docker_build", "docker_build('foo', 'foo')", dbManifestOpt},
```
maiamcc: nice! probably good to have a non-normalizable image name (`gcr.io/.../foo` etc.?) w/o default registry as part of this test?

landism: yeah

this was the straw that broke the table test's back, so I ended up restructuring all these tests

maiamcc: RIP camel

☠️ 🐫 😢

PR #1416, opened 2019-04-02
web: fix up error display

nicks:

Please review the following commits I made in branch nicks/error:

39bfcd3ed25ae2cb53b5d187641710f3aaced917 (2019-04-02 17:39:54 -0400)
web: fix up error display


web/src/testdata.tsx
```diff
@@ -0,0 +1,136 @@
+
```
jazzdan: I think this can just be a regular `.ts` file since it doesn't have any jsx in it? Doesn't super matter though.

nicks: but it might have jsx in the future! i don't want to have to change the name of the file every time we change the contents.

jazzdan: yah! fair enough :)

PR #1419, opened 2019-04-03
web: sidebar visible by default with collapse option

yuindustries:


web/src/Sidebar.scss
```diff
@@ -3,21 +3,24 @@
...
+  transform: translateX(0%);
   transition: transform ease-out 200ms;
   font-size: $font-size;
 }
-.Sidebar.is-open {
-  transform: translateX(0%);
+.Sidebar.is-closed {
+  transform: translateX(100%);
   transition: transform ease-in 200ms;
+  right: $spacing-unit * 1.5;
```
nicks: does this feel jumpy? it might be better to do something like:
transform: translateX(calc(100%  - $spacing-unit * 1.5));

web/src/Sidebar.scss
```diff
@@ -29,25 +32,62 @@
...
+
+.Sidebar-toggle svg {
+  fill: $color-white;
+  width: $spacing-unit * 1.5;
+  transform: rotate(180deg);
+}
+
+.Sidebar.is-closed .Sidebar-toggle svg {
+  transform: rotate(0);
+  transition-property: transform;
```
nicks: should the transition fields be on the plain .Sidebar-toggle svg selector above?

web/src/Statusbar.tsx
```diff
@@ -87,7 +86,7 @@ class Statusbar extends PureComponent<StatusBarProps> {
   upPanel(upCount: number, itemCount: number) {
     let upPanelClasses = "Statusbar-panel Statusbar-panel--up"
     let upPanel = (
-      <button className={upPanelClasses} onClick={this.props.toggleSidebar}>
+      <button className={upPanelClasses}>
```
nicks: this probably shouldn't be a button anymore?

PR #1428, opened 2019-04-03
tiltfile: add basic support for k8s resource assembly v2

landism:
This is most of the implementation of the new resource assembly.

The chief missing thing is the ability to specify resource options (i.e., `k8s_resource`), which will is covered by ch2109.

### DIFF
--- a/internal/tiltfile/tiltfile_state.go
+++ b/internal/tiltfile/tiltfile_state.go
@@ -7,6 +7,8 @@ import (
 	"regexp"
 	"strings"
 
+	"github.com/windmilleng/tilt/internal/k8s"
+
 	"github.com/windmilleng/tilt/internal/logger"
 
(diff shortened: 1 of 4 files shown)


internal/tiltfile/tiltfile_state.go
```diff
@@ -289,6 +297,80 @@ func (s *tiltfileState) assembleK8s() error {
...
+		res, err := s.makeK8sResource(resourceName)
+		if err != nil {
+			return err
+		}
+		err = res.addEntities([]k8s.K8sEntity{workload}, s.imageJSONPaths)
+		if err != nil {
+			return err
+		}
+
+		match, rest, err := k8s.FilterByMatchesPodTemplateSpec(workload, s.k8sUnresourced)
```
nicks: can you add comments on what this code does? it took me a while to figure it out

PR #1433, opened 2019-04-04
tiltfile: rejigger live update declaration syntax"

maiamcc:
Hello @landism, @nicks,

Please review the following commits I made in branch maiamcc/live-update-syntax:

13940be1bc1bb1632fe9f5900e70919ce191cc6e (2019-04-04 14:09:50 -0400)
integration test

f8092c6197e4c734a8c3f9b95cf544ed09fe050e (2019-04-04 14:02:04 -0400)
custom build support

de1c7559431243ba8d4d971e436fb75c23b9a3fc (2019-04-04 13:58:09 -0400)
fall_back_on works

0a28d00d15fa997ec6801be35c3b46108d3f6138 (2019-04-04 12:54:30 -0400)
more red tests - ready for fall_back_on

24608d86b280b877f5745cb5bfe3bbba249e124e (2019-04-04 12:47:01 -0400)
tests so far mostly green

d5fcf8f51e2790be4f6f0b35d1b0de613b07efa5 (2019-04-04 12:22:30 -0400)
red tests

### DIFF
--- a/internal/model/live_update.go
+++ b/internal/model/live_update.go
@@ -37,6 +37,13 @@ type LiveUpdateStep interface {
 	liveUpdateStep()
 }
 
+// Specifies that changes to any of the given files should cause the builder to fall back (i.e. do a full image build)
+type LiveUpdateFallBackOnStep struct {
+	Files []string
+}
+
+func (l LiveUpdateFallBackOnStep) liveUpdateStep() {}
+
 // Specifies that changes to local path `Source` should be synced to container path `Dest`
 type LiveUpdateSyncStep struct {
 	Source, Dest string
--- a/internal/model/live_update.go
+++ b/internal/model/live_update.go
@@ -33,6 +33,8 @@ func NewLiveUpdate(steps []LiveUpdateStep, fullRebuildTriggers PathSet) (LiveUpd
 	return LiveUpdate{steps, fullRebuildTriggers}, nil
 }
 
+func (lu LiveUpdate) Empty() bool { return len(lu.Steps) == 0 }
+
 type LiveUpdateStep interface {
 	liveUpdateStep()
 }
--- a/internal/model/live_update.go
+++ b/internal/model/live_update.go
@@ -20,13 +20,13 @@ func NewLiveUpdate(steps []LiveUpdateStep, fullRebuildTriggers PathSet) (LiveUpd
 		switch step.(type) {
 		case LiveUpdateSyncStep:
 			if seenRunStep {
-				return LiveUpdate{}, errors.New("live_update: all sync steps must precede all run steps")
+				return LiveUpdate{}, errors.New("all sync steps must precede all run steps")
 			}
 		case LiveUpdateRunStep:
 			seenRunStep = true
 		case LiveUpdateRestartContainerStep:
 			if i != len(steps)-1 {
-				return LiveUpdate{}, errors.New("live_update: restart container is only valid as the last step")
+				return LiveUpdate{}, errors.New("restart container is only valid as the last step")
 			}
 		}
 	}
--- a/internal/model/live_update.go
+++ b/internal/model/live_update.go
@@ -20,13 +20,13 @@ func NewLiveUpdate(steps []LiveUpdateStep, fullRebuildTriggers PathSet) (LiveUpd
 		switch step.(type) {
 		case LiveUpdateSyncStep:
 			if seenRunStep {
-				return LiveUpdate{}, errors.New("all sync steps must precede all run steps")
+				return LiveUpdate{}, errors.New("live_update: all sync steps must precede all run steps")
 			}
 		case LiveUpdateRunStep:
 			seenRunStep = true
 		case LiveUpdateRestartContainerStep:
 			if i != len(steps)-1 {
-				return LiveUpdate{}, errors.New("restart container is only valid as the last step")
+				return LiveUpdate{}, errors.New("live_update: restart container is only valid as the last step")
 			}
 		}
 	}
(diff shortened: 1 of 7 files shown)


internal/model/live_update.go
```diff
@@ -20,23 +20,32 @@ func NewLiveUpdate(steps []LiveUpdateStep, fullRebuildTriggers PathSet) (LiveUpd
 		switch step.(type) {
 		case LiveUpdateSyncStep:
 			if seenRunStep {
-				return LiveUpdate{}, errors.New("all sync steps must precede all run steps")
+				return LiveUpdate{}, errors.New("live_update: all sync steps must precede all run steps")
```
landism: it seems a bit cleaner to leave the `live_update` off here and attach it in the tiltfile instead, so that function names are all in that package, but nbd

internal/tiltfile/live_update.go
```diff
@@ -181,66 +221,51 @@ func (s *tiltfileState) liveUpdateStepToModel(l liveUpdateStep) (model.LiveUpdat
...
-	for _, v := range frtSlice {
-		str, ok := v.(starlark.String)
-		if !ok {
-			return starlark.None, fmt.Errorf("'full_rebuild_triggers' must only contain strings - got value '%v' of type '%s'", v.String(), v.Type())
+		// HACK(maia): temporary hack for backwards compatibility/making this a smaller PR--
+		// use the existing LiveUpdate constructor. (Soon, we'll treat fall_back_on as
+		// just another step INTERNALLY too (instead of as a non-step property of a LiveUpdate).
+		if fallBackStep, ok := ms.(model.LiveUpdateFallBackOnStep); ok {
+			if len(fallBackOn) != 0 {
+				return model.LiveUpdate{}, fmt.Errorf("live_update: cannot specify more than one "+
```
landism: I was surprised by this and wondered why we only support specifying one path to fall back on!

After a few seconds' thought, I realized `fall_back_on` probably takes a union of str or list[str]. Maybe worth some text here like "just pass all paths to a single step!"?

(or maybe we should support having multiple? that might look cleaner than a long list of strings in a single step?)

internal/tiltfile/live_update.go
```diff
@@ -249,11 +274,5 @@ func (s *tiltfileState) validateLiveUpdates() error {
 		return fmt.Errorf("live_update steps were created that were not used by any live_update: %s", strings.Join(errorStrings, ", "))
 	}
 
-	for k, v := range s.liveUpdates {
-		if !v.matched {
-			return fmt.Errorf("live_update was specified for '%s', but no built resource uses that image", k)
```
landism: yay, this error was awkward

internal/tiltfile/live_update_test.go
```diff
@@ -229,19 +230,20 @@ func (f *liveUpdateFixture) init() {
...
-)`, f.tiltfileCode, f.configuredImageName)
+]`
+	codeToInsert := fmt.Sprintf(f.tiltfileCode, luSteps)
+	tiltfile := fmt.Sprintf(`
+k8s_yaml('foo.yaml')
+%s
+`, codeToInsert)
 	f.file("Tiltfile", tiltfile)
 
+	fmt.Println(tiltfile)
```
landism: if this is intentional rather than temporary debug, it might be better to `fmt.Printf("tiltfile:\n%q\n", tiltfile)`

maiamcc: BOLD ASSUMPTION.

internal/tiltfile/live_update.go
maiamcc: ehh, supporting multiple is gonna get weird when actually store them as steps on the `model.LiveUpdate`. I'll fix the error message.

maiamcc: actually we do the same thing for adds vs runs don't we. kk

internal/tiltfile/live_update.go
```diff
@@ -181,66 +221,47 @@ func (s *tiltfileState) liveUpdateStepToModel(l liveUpdateStep) (model.LiveUpdat
...
-		fullRebuildTriggers: frtStrings,
-	}
+	return model.NewLiveUpdate(modelSteps, model.NewPathSet(fallBackOn, s.absWorkingDir()))
+}
 
-	return starlark.None, nil
+func (s *tiltfileState) consumeLiveUpdateStep(stepToConsume liveUpdateStep) {
+	for i, step := range s.unconsumedLiveUpdateSteps {
+		if step.declarationPosition() == stepToConsume.declarationPosition() {
+			copy(s.unconsumedLiveUpdateSteps[i:], s.unconsumedLiveUpdateSteps[i+1:])
```
nicks: usually when people use this idiom, they want to zero out the value at the end of the slice?

or did you intend to use a simpler idiom like

```
s.unconsumedLiveUpdateSteps = append(s.unconsumedLiveUpdateSteps[i:], s.unconsumedLiveUpdateSteps[i+1:]...)
```
?

cf
https://github.com/golang/go/wiki/SliceTricks

internal/tiltfile/live_update.go
```diff
@@ -181,66 +221,47 @@ func (s *tiltfileState) liveUpdateStepToModel(l liveUpdateStep) (model.LiveUpdat
...
-		steps:               modelSteps,
-		fullRebuildTriggers: frtStrings,
-	}
+	return model.NewLiveUpdate(modelSteps, model.NewPathSet(fallBackOn, s.absWorkingDir()))
+}
 
-	return starlark.None, nil
+func (s *tiltfileState) consumeLiveUpdateStep(stepToConsume liveUpdateStep) {
+	for i, step := range s.unconsumedLiveUpdateSteps {
+		if step.declarationPosition() == stepToConsume.declarationPosition() {
```
nicks: fwiw, i'm not at all convinced that this equality check will behave like you expect. it has unexported pointer fields. and the struct might change a lot in the future

it might be more robust to create your own struct with the equality semantics you want, and use it as a map key to do inserts/deletes

PR #1434, opened 2019-04-04
model: update LiveUpdate internals to s/fullRebuildTriggers/FallBackOn, and include those as steps

maiamcc:
Hello tbd,

Please review the following commits I made in branch maiamcc/fall-back-on:

956afb14ca1911f5e272f9f34f80cbcbbf03df1c (2019-04-04 17:46:54 -0400)
model: update LiveUpdate internals to s/fullRebuildTriggers/FallBackOn, and include those as steps

### DIFF
--- a/internal/model/live_update.go
+++ b/internal/model/live_update.go
@@ -15,27 +15,34 @@ type LiveUpdate struct {
 }
 
 func NewLiveUpdate(steps []LiveUpdateStep, baseDir string) (LiveUpdate, error) {
-	seenRunStep := false
+	// Check that all FallBackOn steps come at the beginning
+	// (Technically could do this in the loop below, but it's
+	// easier to reason about/modify this way.)
 	seenNonFallBackStep := false
-	for i, step := range steps {
+	for _, step := range steps {
 		switch step.(type) {
 		case LiveUpdateFallBackOnStep:
 			if seenNonFallBackStep {
 				return LiveUpdate{}, errors.New("all fall_back_on steps must precede all other steps")
 			}
+		default:
+			seenNonFallBackStep = true
+		}
+	}
+
+	seenRunStep := false
+	for i, step := range steps {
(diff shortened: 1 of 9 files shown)


internal/model/live_update.go
```diff
@@ -5,16 +5,31 @@ import (
...
-	// When files matching any of these paths change, we should fall back to a full rebuild.
-	FullRebuildTriggers PathSet
+	Steps   []LiveUpdateStep
+	BaseDir string // directory where the LiveUpdate was initialized (we'll use this to eval. any relative paths)
 }
 
-func NewLiveUpdate(steps []LiveUpdateStep, fullRebuildTriggers PathSet) (LiveUpdate, error) {
+func NewLiveUpdate(steps []LiveUpdateStep, baseDir string) (LiveUpdate, error) {
+	// Check that all FallBackOn steps come at the beginning
+	// (Technically could do this in the loop below, but it's
```
maiamcc: happy to hear that i should just do all this checking in one big loop -- my reasoning here is that if we did all order validation in a single loop, any time we add a new step, we'll have to remember to set seenNonFallBackStep when we check them. But i don't have strong feelings 🤷‍♀️

nicks: this is probably fine for now, but this might get unwieldy as we get more restrictions. it might be better to have an explicit state machine

internal/model/live_update.go
```diff
@@ -5,16 +5,31 @@ import (
...
+	// easier to reason about/modify this way.)
+	seenNonFallBackStep := false
+	for _, step := range steps {
+		switch step.(type) {
+		case LiveUpdateFallBackOnStep:
```
nicks: is it possible to have more than one fallback step? if not, do you want to assert that?

maiamcc: yep, it IS possible to have more than one!

PR #1436, opened 2019-04-04
docker-compose: parse ports properly

nicks:
Hello @maiamcc,

Please review the following commits I made in branch nicks/ch1121/port:

9aa77b89d9583c95cc5d6dd48ae0392e41599e21 (2019-04-04 18:06:09 -0400)
docker-compose: parse ports properly
fixes https://github.com/windmilleng/tilt/issues/1383


internal/hud/resourceview.go
```diff
@@ -217,10 +217,15 @@ func (v *ResourceView) resourceExpandedDC() rty.Component {
 	l.Add(rty.TextString(" "))
 	l.AddDynamic(rty.NewFillerString(' '))
 
-	// TODO(maia): ports
-
 	st := v.res.DockerComposeTarget().StartTime
 	if !st.IsZero() {
+		if len(v.res.Endpoints) > 0 {
+			for _, endpoint := range v.res.Endpoints {
+				l.Add(rty.TextString(endpoint))
```
maiamcc: we should prooobably split this out into a helper func so that if we want to change the way we display endpoints, we only have to change it in one place? but eh.

internal/tiltfile/docker_compose.go
```diff
@@ -146,7 +146,59 @@ func (v *Volumes) UnmarshalYAML(unmarshal func(interface{}) error) error {
...
+type Ports []Port
+type Port struct {
+	Published int `yaml:"published"`
+}
+
+func (p *Ports) UnmarshalYAML(unmarshal func(interface{}) error) error {
+	var sliceType []interface{}
+	err := unmarshal(&sliceType)
+	if err != nil {
+		return errors.Wrap(err, "unmarshalling volumes")
```
maiamcc:
```suggestion
		return errors.Wrap(err, "unmarshalling ports")
```

nicks: :grimacing:

internal/hud/resourceview.go
nicks: done! tho i don't think the abstraction helped us much

PR #1437, opened 2019-04-04
web: create preview pane

yuindustries:
I tried tab-like buttons for "Logs" and "Preview" view options.

- In this PR, it's not obvious the view options are buttons. I'll address this in my next design pass.
- When we have more than two views, this design will have to change.

2019-04-04 - preview

## Nav Design Alternatives considered

I mocked up a design with the view options to the left of the sidebar, but it was a bit too visually obtrusive.

2019-04-04 - view options mock

### DIFF
--- a/web/src/HUD.scss
+++ b/web/src/HUD.scss
@@ -1,7 +1,5 @@
-@import 'constants';
+@import "constants";
 
 .HUD {
-  display: flex;
-  background-color: $color-navy;
-  min-height: 100vh;
+  height: 100vh;
 }
--- a/web/src/HUD.scss
+++ b/web/src/HUD.scss
@@ -1,5 +1,8 @@
 @import "constants";
 
 .HUD {
-  height: 100vh;
+  display: flex;
+  align-items: stretch;
+  width: 100%;
+  min-height: 100vh;
 }
--- a/web/src/HUD.tsx
+++ b/web/src/HUD.tsx
@@ -43,6 +43,11 @@ type Resource = {
   ShowBuildStatus: boolean
 }
 
+export enum ResourceView {
+  Log,
+  Preview,
+}
+
 type HudState = {
   Message: string
   View: {
@@ -51,7 +56,7 @@ type HudState = {
     LogTimestamps: boolean
     TiltfileErrorMessage: string
   } | null
-  isSidebarClosed: boolean
+  IsSidebarClosed: boolean
 }
 
 // The Main HUD view, as specified in
@@ -74,7 +79,7 @@ class HUD extends Component<HudProps, HudState> {
         LogTimestamps: false,
         TiltfileErrorMessage: "",
       },
-      isSidebarClosed: false,
+      IsSidebarClosed: false,
     }
 
     this.toggleSidebar = this.toggleSidebar.bind(this)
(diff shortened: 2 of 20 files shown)


web/src/PreviewPane.test.tsx
```diff
@@ -0,0 +1,13 @@
+import React from "react"
+import ReactDOM from "react-dom"
+import PreviewPane from "./PreviewPane"
+import renderer from "react-test-renderer"
+
+it("renders without crashing", () => {
+  let div = document.createElement("div")
+  Element.prototype.scrollIntoView = jest.fn()
+  ReactDOM.render(<PreviewPane endpoint="http://localhost:9000" />, div)
```
jazzdan: @yuindustries and I talked and we should add a test here for when no endpoint is passed. The preview component should probably handle that gracefully.

nicks:
one major bug:
when the logs pane is focused, you can scroll with the mousewheel but there's no scrollbar.
when the logs pane is not focused, you can't scroll at all 😭 . and still no scrollbar

some minor issues:
when looking at all logs, and clicking the "Preview" button, i get a "No match" error
when looking at (Tiltfile) logs, and clicking the "Preview" button, I get a blank screen
when the right sidebar is taller than the screen, there's no scrollbar. there used to be a scrollbar before this PR, so this appears to be a regression

web/src/HUD.scss
```diff
@@ -1,7 +1,5 @@
-@import 'constants';
+@import "constants";
 
 .HUD {
-  display: flex;
-  background-color: $color-navy;
-  min-height: 100vh;
-}
\ No newline at end of file
+  height: 100vh;
```
nicks: i'm pretty sure the change from min-height to height here is wrong; why would we want to make the HUD exactly 100vh?

web/src/HUD.tsx
```diff
@@ -139,15 +157,25 @@ class HUD extends Component<HudProps, HudState> {
       combinedLog = view.Log
     }
 
+    let previewRoute = (props: RouteComponentProps<any>) => {
+      let name = props.match.params ? props.match.params.name : ""
+      let endpoint = ""
+      if (view && name !== "") {
+        let r = view.Resources.find(r => r.Name === name)
+        endpoint = r ? r.Endpoints && r.Endpoints[0] : ""
```
nicks: if there's no endpoint, we probably don't want to return a PreviewPane at all? maybe better to fallback to a LogPane?

web/src/LoadingScreen.scss
```diff
@@ -1,8 +1,14 @@
+@import "constants";
+
 .LoadingScreen {
-  min-height: 100vh;
+  position: fixed;
```
nicks: this seems wrong to me? now the loading screen is off-center and looks kind of broken

web/src/LogPane.scss
```diff
@@ -1,16 +1,27 @@
 @import "constants.scss";
 
 .LogPane {
-  margin: $spacing-unit / 2;
+  position: fixed;
```
nicks: hmmm...i understand that using "position: fixed" is a cheap fix to force the layout to do what you want.

But in practice, putting the main content of a page inside a position:fixed container is almost always something you'll regret. It's semantically / accessibility-wise not what we want, and going to break the interaction in lots of subtle and frustrating ways.

Let's talk about this more tomorrow.

PR #1441, opened 2019-04-05
tiltfile: implement v2 k8s_resource, make v1 deprecated

landism:
nb: the `workload` arg to `k8s_resource` currently requires exact match. That will be taken care of in a future PR.


internal/tiltfile/files.go
```diff
@@ -181,8 +181,10 @@ func (s *tiltfileState) recordConfigFile(f string) {
 	s.configFiles = sliceutils.AppendWithoutDupes(s.configFiles, f)
 }
 
-func (s *tiltfileState) readFile(p localPath) ([]byte, error) {
-	s.recordConfigFile(p.path)
+func (s *tiltfileState) readFile(p localPath, recordConfigFile bool) ([]byte, error) {
```
nicks: this seems like a weird api? if you didn't want to record the config file, wouldn't you just call ioutil.ReadFile directly?

internal/tiltfile/k8s.go
```diff
@@ -173,7 +200,39 @@ func (s *tiltfileState) filterYaml(thread *starlark.Thread, fn *starlark.Builtin
...
+		if _, ok := k8sResourceV1OnlyNames[name]; ok {
+			return true, fmt.Sprintf("it was called with kwarg %q, which is deprecated", name)
+		}
+	}
+
+	// check positional args
+	// check if the second arg is yaml (v1) instead of a resource name (v2)
+	if args.Len() >= 2 {
+		// `false` because this isn't actually a config file - we just want to test if it's valid k8s yaml and throw it away
+		_, err := s.yamlEntitiesFromSkylarkValueOrList(args[1], false)
```
nicks: i feel like we could do a much simpler check here? like just check if this is a file with a newline in it?

for example, if the yaml is invalid for some reason (like a missing colon), you still want to print this error message

internal/tiltfile/k8s.go
```diff
@@ -219,9 +274,9 @@ func (s *tiltfileState) k8sResource(thread *starlark.Thread, fn *starlark.Builti
 		return nil, fmt.Errorf("image arg must be a string or fast_build; got %T", imageVal)
 	}
 
-	portForwards, err := s.convertPortForwards(name, portForwardsVal)
+	portForwards, err := convertPortForwards(portForwardsVal)
 	if err != nil {
-		return nil, err
+		return nil, errors.Wrapf(err, "%s %q", fn.Name(), name)
```
nicks: i thought starklark printed the function name by default? not sure why this one is a special case

integration/imagetags/Tiltfile
```diff
@@ -4,5 +4,6 @@ docker_build(repo + '/imagetags-common:stretch', 'common-stretch')
 docker_build(repo + '/imagetags-app', 'app')
 docker_build(repo + '/imagetags-app:stretch', 'app-stretch')
 
-k8s_resource("regular", yaml='app.yaml', image=repo + '/imagetags-app', port_forwards='31000:8000')
-k8s_resource("stretch", yaml='app-stretch.yaml', image=repo + '/imagetags-app:stretch', port_forwards='31001:8000')
+k8s_yaml(['app.yaml', 'app-stretch.yaml'])
+k8s_resource('imagetags', new_name='regular', port_forwards='31000:8000')
```
nicks: it feels like a lot of the code complexity is because we're changing the signature of an existing function

have we considered just using a new function name?

e.g., k8s_workload?

i wish the spec had some "Alternative Considered" section

internal/tiltfile/tiltfile_state.go
```diff
@@ -113,6 +115,7 @@ const (
 	k8sYamlN                    = "k8s_yaml"
 	filterYamlN                 = "filter_yaml"
 	k8sResourceN                = "k8s_resource"
+	k8sResourceOptionsN         = "k8s_resource_options"
```
nicks: this doesn't appear to be used anywhere?

internal/tiltfile/k8s.go
```diff
@@ -45,6 +47,31 @@ type k8sResource struct {
...
+const deprecatedResourceAssemblyV1Warning = "k8s_resource_assembly v1 is deprecated. " +
+	"It will no longer be supported after 2019-05-01. " +
+	"See https://tilt.dev/resource_assembly_migration.html for information on how to migrate."
+
+const deprecatedResourceAssemblyV2Warning = "k8s_resource_assembly is deprecated. " +
+	"version 2 is the default, so you can simply remove this directive! " +
+	"It will no longer be supported after 2019-05-01. " +
+	"See https://tilt.dev/resource_assembly_migration.html for information on how to migrate."
+
+// holds options passed to `k8s_resource_option` until assembly happens
```
nicks: k8s_resource?

internal/tiltfile/tiltfile_state.go
```diff
@@ -308,6 +311,27 @@ func (s *tiltfileState) assembleK8sV2() error {
 		return err
 	}
 
+	for workload, opts := range s.k8sResourceOptions {
+		if r, ok := s.k8sByName[workload]; ok {
+			r.extraPodSelectors = opts.extraPodSelectors
+			r.portForwards = opts.portForwards
+			if opts.newName != "" && opts.newName != r.name {
+				if _, ok := s.k8sByName[opts.newName]; ok {
+					return fmt.Errorf("k8s_resource at %s specified to rename '%s' to '%s', but there is already a resource with that name", opts.tiltfilePosition.String(), r.name, opts.newName)
```
nicks: can you add a test that exercises this error message?

internal/tiltfile/tiltfile_test.go
```diff
@@ -739,20 +761,9 @@ docker_build('gcr.io/c', 'c')
 `)
 	f.load()
 
-	f.assertNextManifest("a", deployment("deployment-a"))
-	f.assertNextManifest("b", deployment("deployment-b"), service("service-b"))
-	f.assertNextManifest("c", deployment("deployment-c"), service("service-c1"), service("service-c2"))
-}
-
-func TestK8sResourceWithoutDockerBuild(t *testing.T) {
```
nicks: this seems like an important test? why did this get deleted?

internal/tiltfile/k8s.go
landism: yeah, I guess I was unconsciously trying to avoid false positives, which is a weird goal

also, I think introducing the possibility of false positives means we need an escape hatch for false positives, i.e. don't check this if the user has explicitly opted into v2, so I'll add that

internal/tiltfile/k8s.go
landism: It does in the case of failed `UnpackArgs`, but not in other cases!

e.g.,
```
k8s_resource('hello-world', image='.sadf')
```

```
  /Users/matt/go/src/github.com/windmilleng/servantes/Tiltfile:114: 
<toplevel>
Error: invalid reference format
```

internal/tiltfile/tiltfile_state.go
landism: That's here: https://github.com/windmilleng/tilt/pull/1441/files#diff-61015eae0b8419289a8f2f1b45e156e6R2705

internal/tiltfile/tiltfile_test.go
landism: unsure. readded.

PR #1445, opened 2019-04-05
build: skip iterating through directories that are ignored

jazzdan:
This results in a significant performance improvements for initial builds that ignore directories with lots of files in them, in the area of 2x-4x improvement on macOS.


jazzdan:
@nicks ack, I'm going to merge this for now without a test to help folks with performance problems.

PR #1446, opened 2019-04-05
web: typography tweaks

yuindustries:
Added some Google Fonts: Inconsolata for monospace, Montserrat for sans.

New Sidebar Expand/Collapse icon.

I also made the Sidebar eaiser to expand when it's collapsed. (The main pane was covering it, making it hard to expand in Firefox)

2019-04-05 - typography tweaks


web/public/index.html
```diff
@@ -3,6 +3,7 @@
   <head>
     <meta charset="utf-8">
     <link rel="shortcut icon" href="https://tilt.build/favicon.ico">
+    <link href="https://fonts.googleapis.com/css?family=Inconsolata:400,700|Montserrat:600" rel="stylesheet">
```
jazzdan: I've read that using Google fonts is really bad for performance. Is that true?

nicks: i wouldn't be too worried about it right now

PR #1451, opened 2019-04-08
rty: improve draw performance a lot when there are nested frames

nicks:
Hello @landism,

Please review the following commits I made in branch nicks/perf:

c40edfd66f36d237b9a9665304c20ab45a9a9467 (2019-04-08 11:08:38 -0400)
rty: improve draw performance a lot when there are nested frames


nicks:
(the rty interactive differences are cases where the foreground changed colors, but the foreground is a space so it doesn't matter anyway)

nicks:
yes. after this change, resource usage goes down substantially

PR #1455, opened 2019-04-08
tiltfile: stream `local` output to Tiltfile log

jazzdan:


internal/tiltfile/files.go
```diff
@@ -275,17 +278,24 @@ func (s *tiltfileState) execLocalCmd(cmd string) (string, error) {
 	// TODO(nick): Should this also inject any docker.Env overrides?
 	c := exec.Command("sh", "-c", cmd)
 	c.Dir = filepath.Dir(s.filename.path)
-	out, err := c.Output()
+	stdoutString := &strings.Builder{}
```
dbentley: Is strings.builder better than bytes.Buffer? Is it new?

jazzdan: Good question. It's new as of Go 1.10. It's 3x faster with lower memory usage according to [these benchmarks](https://medium.com/@felipedutratine/string-concatenation-in-golang-since-1-10-bytes-buffer-vs-strings-builder-2b3081848c45).

jazzdan:
Closing for now until we have a plan to deal with how many logs this could generate.

PR #1456, opened 2019-04-08
engine: record how many times a user has opened a log in stats [ch2194]

jazzdan:
It's unclear to me how best to record this data. Any advice?

### DIFF
--- a/internal/hud/hud.go
+++ b/internal/hud/hud.go
@@ -137,6 +137,7 @@ func (h *Hud) handleScreenEvent(ctx context.Context, dispatch func(action store.
 				// open if we have multiple, or what path to default to on the opened manifest.
 				_, selected := h.selectedResource()
 				if len(selected.Endpoints) > 0 {
+					h.a.Incr("ui.interactions.open_preview", map[string]string{})
 					err := browser.OpenURL(selected.Endpoints[0])
 					if err != nil {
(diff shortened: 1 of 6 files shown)


internal/hud/hud.go
```diff
@@ -196,6 +196,9 @@ func (h *Hud) handleScreenEvent(ctx context.Context, dispatch func(action store.
 			url := h.webURL
 			url.Path = fmt.Sprintf("/r/%s/", r.Name)
 			_ = browser.OpenURL(url.String())
+			dispatch(OpenLogAction{
```
nicks: there's no good reason to put this in EngineState, create actions for it, etc. It's a lot of indirection that buys you nothing. just inject analytics into the HUD and call Incr() directly.

internal/engine/analytics_reporter.go
```diff
@@ -91,4 +91,14 @@ func (ar *AnalyticsReporter) report() {
 	stats["tiltfile.error"] = tiltfileIsInError
 
 	ar.a.Incr("up.running", stats)
+
+	for index := 0; index < st.OpenLogCount; index++ {
+		ar.a.Incr("ui.interactions.log_open", map[string]string{})
+	}
+	st.OpenLogCount = 0
```
nicks: also, for future reference, this assignment is a no-op. subscribers don't have write access to the EngineState.  `st` isn't a pointer. This mutates a copy of the engine state, which is probably not what you intended.

PR #1458, opened 2019-04-09
hud: analytics endpoint

jazzdan:
Part 1/2.

Part 2 will add code to the frontend to call this endpoint at the appropriate times.

Right now it only supports incrementing keys.

### DIFF
--- a/internal/hud/server/server.go
+++ b/internal/hud/server/server.go
@@ -8,21 +8,31 @@ import (
 	"github.com/gorilla/mux"
 	_ "github.com/gorilla/websocket"
 	"github.com/windmilleng/tilt/internal/store"
+	"github.com/windmilleng/wmclient/pkg/analytics"
 )
+
+type analyticsPayload struct {
+	Verb string
+	Name string
+	Tags map[string]string
+}
 
 type HeadsUpServer struct {
 	store  *store.Store
 	router *mux.Router
+	a      analytics.Analytics
 }
 
-func ProvideHeadsUpServer(store *store.Store, assetServer AssetServer) HeadsUpServer {
+func ProvideHeadsUpServer(store *store.Store, assetServer AssetServer, analytics analytics.Analytics) HeadsUpServer {
 	r := mux.NewRouter().UseEncodedPath()
 	s := HeadsUpServer{
 		store:  store,
 		router: r,
+		a:      analytics,
 	}
 
 	r.HandleFunc("/api/view", s.ViewJSON)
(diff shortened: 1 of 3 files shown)


internal/hud/server/server.go
```diff
@@ -44,3 +54,28 @@ func (s HeadsUpServer) ViewJSON(w http.ResponseWriter, req *http.Request) {
...
+		http.Error(w, "must be POST request", http.StatusBadRequest)
+		return
+	}
+
+	var payloads []analyticsPayload
+
+	decoder := json.NewDecoder(req.Body)
+	err := decoder.Decode(&payloads)
+	if err != nil {
+		http.Error(w, fmt.Sprintf("error parsing JSON payload: %v", err), http.StatusInternalServerError)
```
nicks: this should probably be an http.StatusBadRequest
https://httpstatusdogs.com/400-bad-request

internal/hud/server/server.go
```diff
@@ -44,3 +54,28 @@ func (s HeadsUpServer) ViewJSON(w http.ResponseWriter, req *http.Request) {
...
+		return
+	}
+
+	for _, p := range payloads {
+		if p.Verb != "incr" {
+			http.Error(w, "error parsing payloads: only incr verbs are supported", http.StatusInternalServerError)
```
nicks: ditto

internal/hud/server/server.go
```diff
@@ -8,21 +8,31 @@ import (
 	"github.com/gorilla/mux"
 	_ "github.com/gorilla/websocket"
 	"github.com/windmilleng/tilt/internal/store"
+	"github.com/windmilleng/wmclient/pkg/analytics"
 )
 
+type analyticsPayload struct {
+	Verb string
```
nicks: it would usually be more idiomatic to lowercase the fields on serialization `json:"verb"`

internal/hud/server/server_test.go
```diff
@@ -0,0 +1,159 @@
...
+}
+
+func (f *serverFixture) assertIncrement(name string, count int) {
+	contains := assert.Contains(f.t, f.a.increments, name)
+	if contains {
+		assert.Equal(f.t, count, f.a.increments[name])
+	}
+}
+
+type fakeAnalytics struct {
```
nicks: have you considered using the existing MemoryAnalytics?

jazzdan: argh I looked for "FakeAnalytics" and didn't find anything so I assumed it didn't exist.

PR #1460, opened 2019-04-09
tiltfile: support object identifiers in k8s_resource

landism:
This changes `k8s_resource`'s first arg to be a k8s object identifier instead of just a workload name, per [the spec](https://github.com/windmilleng/tilt.specs/blob/master/resource_assembly.md)

### DIFF
--- a/internal/tiltfile/k8s.go
+++ b/internal/tiltfile/k8s.go
@@ -30,7 +30,11 @@ type k8sResource struct {
 	// The name of this group, for display in the UX.
 	name string
 
-	// All k8s resources to be deployed.
+	// The single workload entity for this resource (only applies to v2 resource assembly)
+	hasWorkloadEntity bool
+	workloadEntity    k8s.K8sEntity
+
+	// All extra k8s resources to be deployed, beyond the workload entity.
 	entities []k8s.K8sEntity
 
 	// Image selectors that the user manually asked to be associated with this resource.
@@ -47,6 +51,18 @@ type k8sResource struct {
 	extraPodSelectors []labels.Selector
 
 	dependencyIDs []model.TargetID
+
+	// whether this resource has already had a k8sResourceOptions applied to it
+	k8sResourceOptionsApplied bool
+	appliedOptions            k8sResourceOptions
+}
+
+func (k k8sResource) Entities() []k8s.K8sEntity {
+	if k.hasWorkloadEntity {
+		return append([]k8s.K8sEntity{k.workloadEntity}, k.entities...)
+	} else {
+		return k.entities
+	}
 }
 
 const deprecatedResourceAssemblyV1Warning = "This Tiltfile is using k8s resource assembly version 1, which has been " +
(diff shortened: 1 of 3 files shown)


internal/tiltfile/tiltfile_state.go
```diff
@@ -758,6 +799,15 @@ type k8sObjectSelector struct {
 	namespace *regexp.Regexp
 }
 
+// change a regex to only match full strings
+func reAnchor(s string) string {
+	if s == "" {
+		return s
+	} else {
+		return fmt.Sprintf("^(?:%s)$", s)
```
nicks: this seems very brittle? like if the user made a typo in the regexp, or forgot a paren, the error message would be busted

internal/tiltfile/tiltfile_state.go
```diff
@@ -772,22 +822,22 @@ func newK8SObjectSelector(apiVersion string, kind string, name string, namespace
 		}
 	}
 
-	ret.apiVersion, err = regexp.Compile(makeCaseInsensitive(apiVersion))
+	ret.apiVersion, err = regexp.Compile(reAnchor(makeCaseInsensitive(apiVersion)))
```
nicks: have you considered changing the matches() function to look for a full-string match, rather than mutate the regexp?

internal/tiltfile/tiltfile_state.go
```diff
@@ -395,6 +379,63 @@ func (s *tiltfileState) isWorkload(e k8s.K8sEntity) (bool, error) {
...
+
+	matched := false
+	var matchedRes *k8sResource
+	for _, r := range s.k8s {
+		if r.hasWorkloadEntity && sel.matches(r.workloadEntity) {
+			// ensure that we do not have multiple options per resource, or multiple resources per option
+			if matched {
+				return fmt.Errorf("matched two resources: %q and %q", matchedRes.name, r.name)
+			}
+			if r.k8sResourceOptionsApplied {
```
nicks: what problem is this trying to solve? i can imagine cases where you'd want to override a previously set option. or if you want to set different options on different lines.

internal/tiltfile/k8s.go
```diff
@@ -67,9 +83,23 @@ func (r *k8sResource) addRefSelector(selector container.RefSelector) {
 	r.refSelectors = append(r.refSelectors, selector)
 }
 
+func (r *k8sResource) addWorkloadEntity(workload k8s.K8sEntity, imageJSONPaths func(e k8s.K8sEntity) []k8s.JSONPath) error {
+	if r.hasWorkloadEntity {
+		return fmt.Errorf("tried to add workload %q to resource %q, when it already has workload %q", workload.Name(), r.name, r.workloadEntity.Name())
```
nicks: can you add a test case for this error? i don't totally understand what problem this is trying to prevent.

internal/tiltfile/k8s.go
```diff
@@ -30,7 +30,11 @@ type k8sResource struct {
 	// The name of this group, for display in the UX.
 	name string
 
-	// All k8s resources to be deployed.
+	// The single workload entity for this resource (only applies to v2 resource assembly)
```
nicks: can you add some comments on why you broke out the data model this way? i spent a long time looking at this and couldn't figure it out.

landism:
cutting this feature - it makes things complicated and doesn't provide much value

PR #1465, opened 2019-04-09
web: record clicks to log and preview links

jazzdan:
Not sure if there's a better way to do this in React Router.


nicks:
have you considered history.listen?

https://github.com/ReactTraining/history#listening

trying to annotate every link is going to be a lot of work and error-prone

jazzdan:
@nicks That's the easy thing I was looking for! I'll take a look.

jazzdan:
@nicks updated

web/src/analytics.ts
```diff
@@ -0,0 +1,14 @@
+// Fire and forget all analytics events
+const incr = (name: string, tags: Map<string, string> = new Map()): void => {
+  let url = `http://${window.location.host}/api/analytics`
+  let tagObj: { [key: string]: string } = {}
```
nicks: tags.toObject()?

web/src/HUD.tsx
```diff
@@ -85,12 +88,21 @@ class HUD extends Component<HudProps, HudState> {
     this.toggleSidebar = this.toggleSidebar.bind(this)
   }
 
+  componentWillMount() {
+    this.unlisten = this.history.listen((location, action) => {
+      let tags: Map<string, string> = new Map()
```
nicks: is there some reason why you didn't use a normal dictionary?

web/src/HUD.tsx
```diff
@@ -7,14 +7,11 @@ import Statusbar, { StatusItem } from "./Statusbar"
...
 import K8sViewPane from "./K8sViewPane"
 import PreviewPane from "./PreviewPane"
-import { Map } from "immutable"
-import {
-  BrowserRouter as Router,
-  Route,
-  Switch,
-  RouteComponentProps,
-} from "react-router-dom"
+import { Map as ImmMap } from "immutable"
```
nicks: hmmm...this doesn't seem right, we want to use the immutable map by default

web/src/HUD.tsx
jazzdan: where a normal dictionary is just an object?

web/src/HUD.tsx
jazzdan: the name conflicted with the ES6 map which I was very confused by. The solution might just be to use objects for analytics instead of maps in the first place.

web/src/HUD.tsx
nicks: ya

PR #1466, opened 2019-04-09
hud: move tiltfile to the top

nicks:

Please review the following commits I made in branch nicks/ch2203/tiltfile:

683ce40d634a624cc05776193ff47a490d19ce2b (2019-04-09 18:01:22 -0400)
hud: move tiltfile to the top

### DIFF
--- a/web/src/Sidebar.tsx
+++ b/web/src/Sidebar.tsx
@@ -9,6 +9,7 @@ class SidebarItem {
   name: string
   status: string
   hasWarnings: boolean
+  hasEndpoints: boolean
 
   /**
    * Create a pared down SidebarItem from a ResourceView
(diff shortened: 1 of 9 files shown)


nicks:
Screenshot from 2019-04-09 18-01-58

nicks:
i could go either way on this, but it fixes the immediate issue

web/src/Sidebar.tsx
```diff
@@ -9,6 +9,7 @@ class SidebarItem {
   name: string
   status: string
   hasWarnings: boolean
+  hasEndpoints: boolean
```
landism: is this supposed to be in a different branch?

nicks: nope! now that the tiltfile is at the top, it's not very helpful to open a preview of the tiltfile by default

PR #1467, opened 2019-04-09
hud: change how we truncate inline error messages

nicks:
Hello @landism,

Please review the following commits I made in branch nicks/ch2212/expansion:

a481aaf299f93509d77261346d5940607ac71581 (2019-04-09 19:31:06 -0400)
hud: change how we truncate inline error messages


nicks:
Screenshot from 2019-04-09 19-30-30

nicks:
hopefully this will handle long error lines more gracefully

what idiot called it rty and not rt-cry send tweet

internal/rty/layouts_test.go
```diff
@@ -133,3 +134,39 @@ func TestMinWidthInNestedConcatLayoutOverflow(t *testing.T) {
 	i := NewInteractiveTester(t, screen)
 	i.Run("min width concat text overflow", 8, 1, f2)
 }
+
+func TestTailLayout(t *testing.T) {
+	sc := tcell.NewSimulationScreen("")
+	err := sc.Init()
+	assert.NoError(t, err)
+	sc.SetSize(8, 1)
```
landism: Is it possible these four lines were cargo culted and unnecessary? I wasn't able to tell in a few minutes what purpose they're serving here (and even elsewhere I'm not sure if there's a good reason to do this instead of InteractiveTester)

nicks: ya, i think this is vestigal from before InteractiveTester set the screen size, removed!

PR #1471, opened 2019-04-10
web: render whitespace in logs

jazzdan:


web/src/LogPane.test.tsx
```diff
@@ -21,3 +21,341 @@ it("renders logs", () => {
 
   expect(tree).toMatchSnapshot()
 })
+
+it("renders logs with leading whitespace and ANSI codes", () => {
```
nicks: hmmm...this test doesn't assert what you think it does? the snapshot tests don't look at css at all.

PR #1472, opened 2019-04-10
tiltfile: add workload_to_resource_function

landism:


internal/tiltfile/k8s.go
```diff
@@ -692,6 +783,50 @@ func (s *tiltfileState) k8sResourceAssemblyVersionFn(thread *starlark.Thread, fn
...
+	thread := s.starlarkThread()
+	for i, e := range workloads {
+		id := newK8SObjectID(e)
+		name, err := s.workloadToResourceFunction.fn(thread, id)
+		if err != nil {
+			return nil, errors.Wrapf(err, "error determing resource name for '%s'", id.String())
+		}
+
+		if conflictingWorkload, ok := takenNames[name]; ok {
+			return nil, fmt.Errorf("both '%s' and '%s' mapped to resource name '%s'", newK8SObjectID(e).String(), newK8SObjectID(conflictingWorkload).String(), name)
```
nicks: can you add a test that triggers this error?

landism: that's [here](https://github.com/windmilleng/tilt/blob/95118527ad3f79dc992212664cbdac3842719039/internal/tiltfile/tiltfile_test.go#L2781)

PR #1480, opened 2019-04-11
hud: bring back the tabs

landism:
per ch2252
image


internal/hud/tabview.go
```diff
@@ -1,26 +1,93 @@
...
 	l.Add(log)
+
 	return l
 }
+
+func (v *TabView) log() string {
+	var ret string
+	switch v.tabState {
+	case view.TabAllLog:
+		ret = v.view.Log.String()
```
nicks: can you Tail(mainLogLineCount) these, to avoid the re-introducing the perf problems?

internal/hud/renderer.go
```diff
@@ -101,15 +118,25 @@ func (r *Renderer) maybeAddAlertModal(vs view.ViewState, layout rty.Component) r
...
+		// FullScreen is handled elsewhere, since it's no longer a pane
+		// but we have to set height to something non-0 or rty will blow up
+	}
+	return rty.NewFixedSize(tabView.Build(), rty.GROW, height)
 }
 
-func renderPaneHeader(vs view.ViewState) rty.Component {
-	s := "(l) open logs in browser"
+func renderPaneHeader() rty.Component {
+	s := "X: expand"
```
nicks: should this become "contract" when it's full-screen?

internal/hud/tabview.go
```diff
@@ -1,26 +1,93 @@
...
+		_, resource := selectedResource(v.view, v.viewState)
+		if !resource.CurrentBuild.Empty() {
+			ret = resource.CurrentBuild.Log.String()
+		} else {
+			ret = resource.LastBuild().Log.String()
+		}
+
+	case view.TabPodLog:
+		_, resource := selectedResource(v.view, v.viewState)
+		if resource.ResourceInfo != nil {
```
nicks: should this have a more reasonable error message when there's no resource info? e.g., the Tiltfile pod log doesn't make sense

landism: Just went with "no logs received", which seems good enough for now, rather than trying to get clever about which kinds of logs are expected for which resources

PR #1481, opened 2019-04-11
web: remove path from analytics

jazzdan:


web/src/HUD.tsx
```diff
@@ -89,7 +89,11 @@ class HUD extends Component<HudProps, HudState> {
   componentWillMount() {
     incr("ui.web.init", { ua: window.navigator.userAgent })
     this.unlisten = this.history.listen((location, _) => {
-      let tags = { path: location.pathname }
+      let isLog =
```
nicks: can you add of mappings from pathname -> tags? this feels very brittle

web/src/HUD.tsx
```diff
@@ -89,7 +89,11 @@ class HUD extends Component<HudProps, HudState> {
...
+        location.pathname.startsWith("/r/") &&
+        !location.pathname.endsWith("/preview")
+      let isPreview = location.pathname.endsWith("/preview")
+      let tags = { isLog: isLog.toString(), isPreview: isPreview.toString() }
```
nicks: would it make sense to add a digest if it's available, like we do for other fields? https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/digest

jazzdan: In the future yeah but for now I didn't care about the path names for their own sake I just wanted to be able to tell what kind of paths they were.

web/src/HUD.tsx
jazzdan: Yeah let me merge this in now but I'll make myself a ticket to do that tomorrow.

PR #1483, opened 2019-04-11
tracer: support lightstep

jazzdan:
We might need to support more fields in the `lightStep.Options` struct before I merge this in but I wanted to validate the approach.


landism:
What's there looks fine. The two main questions that come to mind are:

What's it going to look like with more lightstep options? A doc with a half dozen environment variables?

Is synclet configuration explicitly not worth worrying about for now? This leaves us in the mildly weird position of having synclet traces with no parent tilt traces, but I don't know how likely we are to ever do anything with the traces anyway.

jazzdan:
@landism 

> What's it going to look like with more lightstep options? A doc with a half dozen environment variables?
I think if it came to that we would re-evaluate. This is just a short term solution to unblock some folks. In the long term I think all of this config should live in the Tiltfile and it would be especially dope if Tilt could react to tracing being enabled/disabled without having to be restarted.

> Is synclet configuration explicitly not worth worrying about for now? This leaves us in the mildly weird position of having synclet traces with no parent tilt traces, but I don't know how likely we are to ever do anything with the traces anyway.
Yeah it's pretty underutilized (though we should verify that). The folks that are asking for tracing don't use it, either.

PR #1484, opened 2019-04-12
engine: pull subscribers into their own wire provider

nicks:
Hello @jazzdan,

Please review the following commits I made in branch nicks/subscribers:

98196b5329be425f5f9db2fcfce1167cd83ee693 (2019-04-12 11:44:14 -0400)
engine: pull subscribers into their own wire provider


nicks:
(the hope is that this makes it easier to not have `engine` depend on EVERYTHING)

PR #1485, opened 2019-04-12
sail: no-op sail server that just prints "connected"

nicks:
Hello @jazzdan, @landism,

Please review the following commits I made in branch nicks/sail:

6c947fa9453b936293265cede5b2c4dc2345eaae (2019-04-12 12:16:59 -0400)
sail: no-op sail server that just prints "connected"


nicks:
(i think it's likely that we will eventually want to move this into its own repo, but for prototyping just going to iterate in the main repo)

PR #1488, opened 2019-04-12
cli: fix missing dev suffix in version analytics

jazzdan:


internal/cli/build.go
```diff
@@ -31,6 +31,14 @@ func (e BuildInfo) empty() bool {
 	return e == BuildInfo{}
 }
 
+func (e BuildInfo) FullVersion() string {
```
nicks: Can we call it analytics version or smth? "full" is very vague

internal/cli/build_test.go
```diff
@@ -10,3 +10,27 @@ import (
 func TestProvideWebVersion(t *testing.T) {
 	assert.Equal(t, fmt.Sprintf("v%s", devVersion), string(provideWebVersion(provideBuildInfo())))
 }
+
+func TestBuildVersion(t *testing.T) {
```
nicks: This does not seem like a useful test. It's unlikely to catch an actual regression. A better test might be one that tests what analytics is sending

nicks: (a good rule of thumb : would this test have prevented this bug? In this case, the bug is in the interaction between two components, so a test that only exercises one does not really get at the problem)

PR #1489, opened 2019-04-12
sail: stream data from the client to the server

nicks:
Hello @jazzdan, @landism,

Please review the following commits I made in branch nicks/sailclient:

365859eff791544823b415096aff9bc72508b17d (2019-04-12 15:02:12 -0400)
sail: stream data from the client to the server

### DIFF
--- a/internal/sail/client/client.go
+++ b/internal/sail/client/client.go
@@ -5,21 +5,27 @@ import (
 	"net/http"
 	"sync"
 
-	"github.com/gorilla/websocket"
 	"github.com/pkg/errors"
+	"github.com/windmilleng/tilt/internal/hud/server"
+	"github.com/windmilleng/tilt/internal/hud/webview"
+	"github.com/windmilleng/tilt/internal/logger"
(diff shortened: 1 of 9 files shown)


internal/sail/client/client.go
```diff
@@ -71,34 +89,44 @@ func (s *SailClient) Connect(ctx context.Context) error {
...
+		}
 	}
 
-	// TODO(nick): To get an end-to-end connection working, we're just
-	// going to connect to the Sail server on startup. Eventually this
-	// should be changed to connect on user action.
-	err := s.Connect(ctx)
-	if err != nil {
-		st.Dispatch(store.NewErrorAction(errors.Wrap(err, "SailClient")))
+	if !s.isConnected() {
```
jazzdan: If there was no error but we're also not connected is that an error condition?

nicks: yes! tho i expect all this code to go away as the code gets fleshed out, see the big TODO a couple lines above this

jazzdan: Sounds good to me!

PR #1494, opened 2019-04-15
web: cancel auto-scroll on mousewheel up

nicks:
Hello @yuindustries,

Please review the following commits I made in branch nicks/ch2263/scroll:

a6ded38e32463d96a086ec2571b938337ac14cf9 (2019-04-15 12:53:54 -0400)
web: cancel auto-scroll on mousewheel up


nicks:
(also changed this to use a RAF so that it doesn't consume CPU when the tab isn't focused)

PR #1496, opened 2019-04-16
network: clean up port-binding logic a bit

nicks:
Hello @landism,

Please review the following commits I made in branch nicks/localhost:

d9a111a4d442e5831e8b14ffb4a5281efdb9b13d (2019-04-16 14:03:35 -0400)
network: clean up port-binding logic a bit

### DIFF
--- a/internal/network/port_test.go
+++ b/internal/network/port_test.go
@@ -1,0 +1,30 @@
+package network
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestIsLocalhostFree(t *testing.T) {
+	// Find a free port
+	port := 0
+	for port = 10000; port < 10100; port++ {
+		if IsBindAddrFree(LocalhostBindAddr(port)) != nil {
+			break
+		}
+	}
+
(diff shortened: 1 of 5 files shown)


internal/network/port_test.go
```diff
@@ -0,0 +1,30 @@
...
+
+	// bind that port on localhost
+	l, err := bindAddress(LocalhostBindAddr(port))
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer l.Close()
+
+	// assert that the port is now bound
+	err = IsPortFree(port)
```
landism: as you may have seen, this unit test passed on linux and failed on macos

According to [this](https://stackoverflow.com/a/14388707):
* macos allows simultaneously binding `127.0.0.1:1234` and `0.0.0.0:1234` if `SO_REUSEADDR` is set (and it turns out go does set that for listening sockets)
* linux doesn't

we could try turning off `SO_REUSEADDR` for our sockets?

nicks: i just changed the interface to accept a bind address instead of a port, which is probably closer to what we want anyway.

PR #1497, opened 2019-04-16
engine: only inject synclet once

jazzdan:


internal/synclet/sidecar/sidecar_test.go
```diff
@@ -35,3 +35,29 @@ func TestInjectSyncletSidecar(t *testing.T) {
 		t.Errorf("could not find image in yaml (%s):\n%s", SyncletImageName, result)
 	}
 }
+
+func TestInjectSyncletSidecarMultipleContainers(t *testing.T) {
```
jazzdan: This test doesn't exercise this particular bug, but I left it in anyways.

internal/engine/image_build_and_deployer_test.go
```diff
@@ -123,6 +123,38 @@ func TestDeployPodWithMultipleImages(t *testing.T) {
 		"Expected image to appear once in YAML: %s", f.k8s.Yaml)
 }
 
+func TestDeployPodWithMultipleFastBuildImages(t *testing.T) {
```
maiamcc: since we're moving away from FastBuild, might be more useful to test this behavior with LiveUpdate instead

internal/engine/image_build_and_deployer.go
```diff
@@ -259,7 +260,7 @@ func (ibd *ImageBuildAndDeployer) deploy(ctx context.Context, st store.RStore, p
 				if replaced {
 					injectedDepIDs[depID] = true
 
-					if ibd.injectSynclet && needsSynclet {
+					if ibd.injectSynclet && needsSynclet && !injectedSynclet {
```
maiamcc: 😳 good call.

internal/engine/image_build_and_deployer.go
```diff
@@ -211,6 +211,7 @@ func (ibd *ImageBuildAndDeployer) deploy(ctx context.Context, st store.RStore, p
 	deployLabel := k8s.TiltDeployLabel(deployID)
 
 	var targetIDs []model.TargetID
+	injectedSynclet := false
```
nicks: i don't think this will work correctly in the case where you have multiple pods? is that a case we care about?

nicks: it might be good to map this by entity name/kind or smth

jazzdan: @nicks what do you think of this approach where we only inject a synclet at most once for each Kubernetes entity?

Also can you say more about multiple pods? You mean replica sets? Or multiple deployments?

PR #1498, opened 2019-04-16
engine: add newline to a direct call to logger.write

maiamcc:
as mentioned [here](https://github.com/windmilleng/tilt/pull/1454#discussion_r275933560) (#1451), the `write` method doesn't actually append a newline, so we need to do it ourselves.

otherwise, looks like this:
Screen Shot 2019-04-16 at 2 24 25 PM


maiamcc:
gonna land this as Lindsey is doing some debugging around here and I want things to look pretty

PR #1499, opened 2019-04-16
sail: end-to-end sharing

nicks:
Hello @jazzdan, @landism,

Please review the following commits I made in branch nicks/ch2240/sail:

501c173096a11171c736e62e3d7f06562ccf7771 (2019-04-16 14:35:10 -0400)
sail: end-to-end sharing

### DIFF
--- a/internal/sail/server/server.go
+++ b/internal/sail/server/server.go
@@ -1,12 +1,16 @@
 package server
 
 import (
+	"context"
+	"fmt"
 	"log"
 	"net/http"
 	"sync"
 
 	"github.com/gorilla/mux"
 	"github.com/gorilla/websocket"
+
+	hudServer "github.com/windmilleng/tilt/internal/hud/server"
 )
 
 var upgrader = websocket.Upgrader{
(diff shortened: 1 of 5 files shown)


nicks:
with this pr, we're now at a place where we can replicate the data to the sail server, and view it in a UI served by the sail server. it's still super buggy and needs lots of work, particularly in 
- how we deploy the server
- how the share interaction works
- properly setting up and tearing down the connections
- making the ux work correctly

internal/sail/server/server.go
```diff
@@ -43,14 +52,25 @@ func (s SailServer) newRoom(conn *websocket.Conn) *Room {
 
 	room := NewRoom(conn)
 	s.rooms[room.id] = room
+	log.Printf("newRoom: %s", room.id)
```
jazzdan: 👀

jazzdan: There are a bunch of these logs: are they meant to be used long term? If so can we make them structured logs so they'd be easy to pick up in Honeycomb?

web/src/HUD.tsx
```diff
@@ -115,7 +128,15 @@ class HUD extends Component<HudProps, HudState> {
...
 
+  path(relPath: string) {
+    if (relPath[0] != "/") {
+      throw new Error('relPath should start with "/", actual:' + relPath)
+    }
+    return this.rootPath + relPath
+  }
+
   render() {
+    debugger
```
jazzdan: 👀

internal/sail/server/server.go
nicks: yep, the logs are intentional. wanted to get something working and debuggable before we start thinking about logging strategy. a lot will need to change before we can even deploy this

web/src/HUD.tsx
nicks: :grimacing:

PR #1501, opened 2019-04-17
web: tab nav lives above main pane instead of in sidebar

jazzdan:
(Screenshot forthcoming)

### DIFF
--- a/web/src/TabNav.tsx
+++ b/web/src/TabNav.tsx
@@ -25,8 +25,8 @@ class TabNav extends PureComponent<NavProps> {
             >
               Logs
             </Link>
-            </li>
-            <li>
+          </li>
+          <li>
             <Link
               className={`tabLink ${
                 previewIsSelected ? "tabLink--is-selected" : ""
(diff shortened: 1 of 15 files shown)


jazzdan:
Screen Shot 2019-04-17 at 5 02 13 PM


Looks terrible but @yuindustries is going to help me spiff it up.

web/src/TabNav.tsx
```diff
@@ -0,0 +1,50 @@
+import React, { PureComponent } from "react"
+import { Link } from "react-router-dom"
+import { SidebarItem } from "./Sidebar"
+import { ResourceView } from "./HUD"
```
nicks: the circular dependency between TabNav and HUD seems not good, maybe this should be moved to a common file?

web/src/TabNav.tsx
```diff
@@ -0,0 +1,50 @@
...
+import "./TabNav.scss"
+type NavProps = {
+  resourceName: string
+  sidebarItems: Array<SidebarItem>
```
nicks: it seems weird to pass all the sidebar items into TabNav. why should it know about all the sidebar items? maybe just pass in the previewResourceURL?

PR #1502, opened 2019-04-17
k8s: don't inject labels into persistent volume claims

nicks:
Hello @landism,

Please review the following commits I made in branch nicks/ch2291/pvc:

fb95da77a391754563a88e18097286e2543a72d0 (2019-04-17 18:08:24 -0400)
k8s: don't inject labels into persistent volume claims

### DIFF
--- a/internal/k8s/extract.go
+++ b/internal/k8s/extract.go
@@ -4,12 +4,12 @@ import (
 	"fmt"
 	"reflect"
 
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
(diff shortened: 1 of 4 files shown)


internal/k8s/extract.go
```diff
@@ -4,12 +4,12 @@ import (
 	"fmt"
 	"reflect"
 
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
```
landism: I thiiiiink this was a bug in goimports that got fixed?

nicks: hmmmm...i just pulled a new version of goimports and it's still doing it?

landism: apparently it actually went in the other direction!

https://github.com/golang/go/issues/30051

maybe we should have ci verify w/ goimports instead of gofmt

PR #1505, opened 2019-04-18
kustomize: support alternative file names

### LINKED ISSUE #1503: support additional kustomize file names
The main `kustomize` file can come in different filenames.
```
➜ kustomize build /tmp
Error: unable to find one of 'kustomization.yaml', 'kustomization.yml' or 'Kustomization' in directory '/tmp'
```

Right now `tilt` only looks for `kustomization.yaml`.
And if one tries to point to the `kustomize` file directly it won't work.
```
error: 'exit status 1'
       stdout: ''
       stderr: 'Error: got file 'Kustomization', but
       './kustomize/infrastructure/Kustomization' must be a directory to be a
        root
       '
```

**edit:** if I rename the first kustomization file it will just continue with all base resources. This makes tilt unusable for us right now (can't rename all kustomization files).

jazzdan:
Mostly taken from the [kustomize source code](https://github.com/kubernetes-sigs/kustomize/blob/ee68a9c450bc884b0d657fb7e3d62eb1ac59d14f/pkg/target/kusttarget.go#L97) itself.

Fixes #1503

### DIFF
--- a/internal/kustomize/dependencies.go
+++ b/internal/kustomize/dependencies.go
@@ -32,6 +32,7 @@ type configMapGenerator struct {
 	Files []string `yaml:"files"`
 }
 
+// Mostly taken from the [kustomize source code](https://github.com/kubernetes-sigs/kustomize/blob/ee68a9c450bc884b0d657fb7e3d62eb1ac59d14f/pkg/target/kusttarget.go#L97) itself.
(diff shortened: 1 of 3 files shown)


jazzdan:
I'd like to add a couple more test cases.

PR #1506, opened 2019-04-18
ci: check goimports instead of gofmt

landism:
This should reduce the churn on separating groups of imports.

Unfortunately, it won't completely eliminate it. goimports ensures that types of imports are in separate groups, but it doesn't not ensure that they are grouped minimally (i.e., splitting tilt imports into two, or putting each import into its own group is legit by goimports standards).

You'll mostly just want to look at the `Makefile` and `.circleci/` changes. The other changes are simply bringing our source into compliance with `goimports -local github.com/windmilleng/tilt`.


jazzdan:
@landism should we merge this in?

landism:
> @landism should we merge this in?

Yes, probably, once I have time to make it actually work. It's not currently actually running goimports.

Makefile
```diff
@@ -111,11 +114,9 @@ errcheck:
 timing: install
 	./scripts/timing.py
 
+WIRE_PATHS = engine cli synclet sail/client
```
maiamcc: 👌

PR #1509, opened 2019-04-18
tiltfile: make v2 assembly the default

landism:
some of the errors/warnings as they appear in the hud:

image

image

image

### DIFF
--- a/internal/tiltfile/k8s.go
+++ b/internal/tiltfile/k8s.go
@@ -1,9 +1,7 @@
 package tiltfile
 
 import (
-	"bytes"
 	"fmt"
-	"io/ioutil"
 	"sort"
 	"strconv"
 	"strings"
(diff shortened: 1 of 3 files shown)


internal/tiltfile/k8s.go
```diff
@@ -274,7 +274,50 @@ func (s *tiltfileState) k8sResource(thread *starlark.Thread, fn *starlark.Builti
...
+		switch x := args[1].(type) {
+		case starlark.Sequence:
+			return true, "second arg was a sequence"
+		case *blob:
+			return true, "second arg was a blob"
+		case starlark.String:
+			bs, err := ioutil.ReadFile(s.localPathFromString(string(x)).path)
+			if err == nil {
+				if bytes.ContainsRune(bs, '\n') {
+					return true, "second arg was a file containing a newline, which might be yaml, and can't be a valid resource name"
```
maiamcc: ...wait what is this case?

landism: added comments to try to answer this question

PR #1511, opened 2019-04-18
sail: more startup options and better cleanup

nicks:
Hello @maiamcc, @landism,

Please review the following commits I made in branch nicks/sailconfig:

35d3032d3bd041ab89838659d055db50a400fc7b (2019-04-18 18:21:24 -0400)
sail: more startup options and better cleanup

### DIFF
--- a/internal/sail/sail.go
+++ b/internal/sail/sail.go
@@ -5,6 +5,9 @@ import (
 	"log"
 	"net/http"
 	"os"
+	"os/signal"
+	"syscall"
+	"time"
 
 	"github.com/spf13/cobra"
 	hudServer "github.com/windmilleng/tilt/internal/hud/server"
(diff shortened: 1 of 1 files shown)


internal/sail/sail.go
```diff
@@ -24,22 +31,51 @@ func Execute() {
...
 	if err != nil {
 		log.Fatal(err)
 	}
 }
 
-func run(cmd *cobra.Command, args []string) {
+func contextWithCancel() (context.Context, func()) {
 	ctx, cancel := context.WithCancel(context.Background())
-	defer cancel()
+	l := logger.NewLogger(logger.InfoLvl, os.Stderr)
```
maiamcc: this should probably be common between this and `cli/cli.go`

maiamcc: hmm but there's not really a good place to put it, i'll leave it for now

PR #1514, opened 2019-04-19
web: resource timings

jazzdan:
Looks like this:

Screen Shot 2019-04-19 at 3 45 34 PM


jazzdan:
The tests very infuriatingly pass on my machine, but not in CI. Will investigate more on Monday.

PR #1518, opened 2019-04-22
web: statusbar combined message

jazzdan:


web/src/Statusbar.tsx
```diff
@@ -2,14 +2,24 @@ import React, { PureComponent } from "react"
 import { ReactComponent as LogoSvg } from "./assets/svg/logo-imagemark.svg"
 import { combinedStatus, warnings } from "./status"
 import "./Statusbar.scss"
+import { combinedStatusMessage } from "./combinedStatusMessage"
 
 const nbsp = "\u00a0"
 
+type Build = {
+  Error: {} | string | null
+  Edits: Array<string>
```
yuindustries: This is not being used

web/src/combinedStatusMessage.test.ts
```diff
@@ -0,0 +1,44 @@
...
+  oneResourceBuilding,
+  oneResourceFailedToBuild,
+  oneResourceCrashedOnStart,
+  oneResource,
+  oneResourceNoErrors,
+} from "./testdata.test"
+import { combinedStatusMessage } from "./combinedStatusMessage"
+import { StatusItem } from "./Statusbar"
+
+describe("interesting resources", () => {
```
yuindustries: Not so "interesting" anymore, is it?

web/src/combinedStatusMessage.ts
```diff
@@ -0,0 +1,31 @@
...
+  let buildingResources = resources.filter(
+    r => !isZeroTime(r.currentBuild.StartTime)
+  )
+
+  if (buildingResources.length > 0) {
+    return "Building: " + buildingResources[0].name
+  }
+
+  let containerCrashedResources = resources.filter(
+    r => r.podStatus === "CrashLoopBackOff"
```
yuindustries: Note:
If we add an error on startup, we do see "Container crashed". But if we fix that error, we'll see "Building..." and a quick flash of "Container crashed" before the status message clears on success.

PR #1521, opened 2019-04-22
model: new matcher that matches on either exact match of child of given paths (for better tiltfile for Sail)

maiamcc:
Hello @landism, @jazzdan,

Please review the following commits I made in branch maiamcc/file-or-child-matcher:

51f641ee2de47ed80bb016f8d149e36eccd6b9c0 (2019-04-22 17:01:19 -0400)
add directory-based run triggers back to Tiltfile for sail

e0a1f09cfeaa7a00023b163f6bb9d0aa1efe621d (2019-04-22 16:54:31 -0400)
test

a1f1ec8c9836f9ca4b6f19b115677223dea3619a (2019-04-22 16:40:41 -0400)
model: new matcher that matches on either exact match of child of given paths

### DIFF
--- a/Tiltfile
+++ b/Tiltfile
@@ -7,8 +7,8 @@ repo = local_git_repo('.')
 live_update = [
   sync('internal', '/go/src/github.com/windmilleng/tilt/internal'),
   sync('web', '/go/src/github.com/windmilleng/tilt/web'),
-  run('make build-js'),  # trigger=['web'] (when support directory/glob triggers)
-  run('make install-sail'),  # trigger=['internal'] (when support directory/glob triggers)
+  run('make build-js', trigger=['web/']),
+  run('make install-sail', trigger=['internal/']),
   restart_container(),
 ]
 docker_build('gcr.io/windmill-public-containers/sail',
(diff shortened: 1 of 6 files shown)


Tiltfile
```diff
@@ -7,8 +7,8 @@ repo = local_git_repo('.')
 live_update = [
   sync('internal', '/go/src/github.com/windmilleng/tilt/internal'),
   sync('web', '/go/src/github.com/windmilleng/tilt/web'),
-  run('make build-js'),  # trigger=['web'] (when support directory/glob triggers)
-  run('make install-sail'),  # trigger=['internal'] (when support directory/glob triggers)
+  run('make build-js', trigger=['web/']),
```
jazzdan: What happens if you leave the slash off here?

maiamcc: it also works -- i just wanted it to be visually clear that it's a directory. i don't feel strongly, do you have thoughts about what's prettier here?

jazzdan: Nope just wanted to make sure it still worked without the slash. :)

PR #1523, opened 2019-04-23
internal/build: print logs for errors in all cases

dbentley:


internal/build/buildkit_printer.go
```diff
@@ -111,16 +106,14 @@ func (b *buildkitPrinter) parseAndPrint(vertexes []*vertex, logs []*vertexLog) e
...
-				}
+			err := b.flushLogs(b.logger.Writer(logger.InfoLvl), vl)
+			if err != nil {
+				return err
 			}
 		}
 
-		if streamLogs && (vl.vertex.isRun() || vl.vertex.isStep()) {
-			err := b.flushLogs(b.logger.Writer(streamLevel), vl)
+		if vl.vertex.isRun() || vl.vertex.isStep() {
```
landism: maybe make this an `else if` so that we don't log the error twice?

and then we could even keep the `streamLogs` in there, to avoid greatly increasing the amount of output

landism: nm on the second sentence. Looking at the old comment, it seems `streamLogs` was added when we were only streaming on verbose level, and no longer really makes sense.

I think the first still makes sense, though

dbentley: I believe flushLogs is reentrant: it keeps track of what it's flushed, and so will only flush it once.

PR #1524, opened 2019-04-23
Read yaml function

l1redd:
Similar to the `read_json` function. You can pass a `.yaml` file to `read_yaml` in the Tiltfile. The function will parse the yaml and return a dictionary.

### DIFF
--- a/internal/tiltfile/files.go
+++ b/internal/tiltfile/files.go
@@ -10,7 +10,7 @@ import (
 
 	"github.com/windmilleng/tilt/internal/sliceutils"
 
-	pkgyaml "github.com/ghodss/yaml"
+	"github.com/ghodss/yaml"
 	"github.com/pkg/errors"
 	"go.starlark.net/starlark"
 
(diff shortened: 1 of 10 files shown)


internal/tiltfile/files.go
```diff
@@ -401,6 +402,40 @@ func (s *tiltfileState) listdir(thread *starlark.Thread, fn *starlark.Builtin, a
...
+		if os.IsNotExist(err) && defaultValue != nil {
+			return defaultValue, nil
+		}
+		return nil, err
+	}
+
+	var decodedJSON interface{}
+	err = yaml.Unmarshal(contents, &decodedJSON)
+	if err != nil {
+		return nil, fmt.Errorf("YAML to JSON conversion error: %v in %s", err, path.GoString())
```
jazzdan: A user might be confused that reading YAML involves JSON. Should we hide that from them and just call this "error parsing YAML"? I'm not sure.

l1redd: Yeah, I think that makes sense. I'll change it

l1redd: Later on in the function there's another error messages that says "error converting JSON to Starlark..." and it might be useful if I change that one as well?

jazzdan: Yeah sounds good to me.

Issue #1522, opened 2019-04-23
Live update only seems to work for first container in pod


mseashor:
First off - I'm really liking Tilt so far - fast in place builds with live_updates are great!   It just seems I've hit an issue with live_updates for a container which isn't the first container in a pod.  

**Scenario**: Two pods in a container need to support live updates.  FYI: In reality, if two containers with live updates aren't supported, we could probably get away with one container supporting live updates but an error indicating **only** the first container in a pod is supported  would be incredibly helpful (it took me forever to figure out that's why it seems live updates weren't working).

**Test Case**: I've attached the simplest possible test case I can come up with - basically just two apache servers (apache1 and apache2) servers running on different containers/ports in the same pod with live updates to their index.html.  It is a bit contrived, but illustrates the point.

- docker + kubernetes yaml + Tiltfile to reproduce: [only_first_container_live_reloads.zip](https://github.com/windmilleng/tilt/files/3105818/only_first_container_live_reloads.zip).

- Tiltfile for reference (also in zip)
```
k8s_resource_assembly_version(2)
  
docker_build('test.com/apache1','apache1',
  live_update=[
    sync('apache1/public_html', '/usr/local/apache2/htdocs'),
  ]
)
docker_build('test.com/apache2','apache2',
  live_update=[
    sync('apache2/public_html', '/usr/local/apache2/htdocs'),
  ]
)

k8s_yaml('k8s.yaml')

k8s_resource('example-deployment', port_forwards=["8080:80","8081:81"])
```
- Commands to reproduce:
```
unzip only_first_container_live_reloads.zip
cd tilt_issue_only_first_container_live_reloads
tilt up                
#(switch to a new terminal)

echo "apache11" > apache1/public_html/index.html   

#Note that live reload works on apache1 (first container in pod)
# Ex: 
# ──┤ Rebuilding: example-deployment #├──────────────────────────────────────────# ──
#  → Updating container…
# Copying 1 file(s) to container: 0551cadab8

echo "apache22" > apache2/public_html/index.html  

#Note live reload does not work on apache2 and it is fully rebuilt - sometimes getting into a crash loop
```
- The message in tilt when failing to live update is: ```falling back to next update method because: don't have info for deployed container (often a result of the deployment not yet being ready)``` (https://github.com/windmilleng/tilt/blob/master/internal/engine/extractors.go#L90)
 
- This issue seemed similar to: https://github.com/windmilleng/tilt/issues/1336, so I purposely don't set tags on the images.

[...]

PR #1525, opened 2019-04-23
engine: better messaging around the fact that we only support live update for the first container on a pod

maiamcc:
Addresses #1522


maiamcc:
This is the wrong approach, #1530 replaces this. Closing.

PR #1526, opened 2019-04-23
web: error pane

jazzdan:
Screen Shot 2019-04-23 at 3 05 34 PM

### DIFF
--- a/web/src/AnsiLine.tsx
+++ b/web/src/AnsiLine.tsx
@@ -1,0 +1,16 @@
+import React from "react"
+import Ansi from "ansi-to-react"
+
+type AnsiLineProps = {
+  line: string
+}
+
+let AnsiLine = React.memo(function(props: AnsiLineProps) {
+  return (
+    <Ansi linkify={false} useClasses={true}>
+      {props.line}
+    </Ansi>
+  )
+})
+
+export default AnsiLine
--- a/web/src/ErrorPane.scss
+++ b/web/src/ErrorPane.scss
@@ -1,0 +1,22 @@
+@import "constants";
+
+.ErrorPane {
+  margin-top: $tabnav-height;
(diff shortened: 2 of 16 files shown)


web/src/AnsiLine.tsx
```diff
@@ -0,0 +1,16 @@
+import React from "react"
+import Ansi from "ansi-to-react"
+
+type AnsiLineProps = {
+  line: string
+}
+
+let AnsiLine = React.memo(function(props: AnsiLineProps) {
```
yuindustries: Is memoization OK in a shared component? We believe so.

web/src/ErrorPane.scss
```diff
@@ -0,0 +1,22 @@
...
+  padding: $spacing-unit / 2;
+}
+
+.ErrorPane ul {
+  list-style: none;
+}
+
+.ErrorPane li + li {
+  margin-top: $spacing-unit;
+  border-top: 2px solid $color-grey;
```
yuindustries: This color no longer exists in `master`; use `$color-gray`

web/src/ErrorPane.tsx
```diff
@@ -0,0 +1,88 @@
...
+        errorElements.push(
+          <li key={"resourceInfoPodCrash" + r.name}>
+            <p>{`${r.name} has container restarts: ${
+              r.resourceInfo.podRestarts
+            }.`}</p>
+            <p>{`Last log line: ${r.resourceInfo.podLog}`}</p>
+          </li>
+        )
+      }
+      r.buildHistory.forEach((b, i) => {
```
yuindustries: For now, we should only show at most one build error for each resource. 

In the future, we can explore showing all the accumulated errors — but this will require more explanation in the UI, and possibly a "clear" button.

PR #1538, opened 2019-04-25
hud/web: surface tilt forward url in web UI  [ch2294]

maiamcc:
Hello @yuindustries, @nicks,

Please review the following commits I made in branch maiamcc/room-id-in-view:

831dd9a0ed67e5acafd58ce9b6d809b91db11094 (2019-04-25 15:04:04 -0400)
fix tests

39520a623e5cc83f00a28cd5180aee4661bc03f9 (2019-04-25 14:46:08 -0400)
styling, only show link if exists

4f3d0d2e5c19d9dd6920a21bc0530f89ffd77a38 (2019-04-25 14:38:01 -0400)
actually get url into component

b398b8b5c55ca962fbd486581ddc8259369067dc (2019-04-24 17:39:27 -0400)
wip

ca513e6dacdf3f9ea86c446f2a7ec2c7cddc1915 (2019-04-24 17:16:37 -0400)
get it into the view in react

08ae370fd022c0ef7ec6dc9a7e94fdf4fd1b23da (2019-04-24 16:50:37 -0400)
send url or error in SailRoomConnectedAction


internal/store/engine_state.go
```diff
@@ -75,6 +75,8 @@ type EngineState struct {
 	LastTiltfileBuild    model.BuildRecord
 	CurrentTiltfileBuild model.BuildRecord
 	TiltfileCombinedLog  model.Log
+
+	SailUrl string
```
nicks: i thought that this should be SailURL in go style?
https://github.com/golang/go/wiki/CodeReviewComments#initialisms

internal/sail/client/client.go
```diff
@@ -13,6 +14,13 @@ import (
 	"github.com/windmilleng/tilt/internal/store"
 )
 
+type SailRoomConnectedAction struct {
+	ViewUrl string // URL to view the Sail room
```
nicks: ditto, ViewUrl -> ViewURL

internal/engine/upper.go
```diff
@@ -374,6 +377,14 @@ func handleLogTimestampsAction(state *store.EngineState, action hud.SetLogTimest
 	state.LogTimestamps = action.Value
 }
 
+func handleSailRoomConnectedAction(ctx context.Context, state *store.EngineState, action client.SailRoomConnectedAction) {
+	if action.Err != nil {
+		logger.Get(ctx).Infof("Error connecting Sail room: %v\n", action.Err)
```
nicks: it always seems weird to me when reducers log, since that just pushes out another action, tho i'm not totally sure what we should be doing instead

internal/sail/client/client_test.go
```diff
@@ -39,12 +40,29 @@ func TestBroadcast(t *testing.T) {
 	assert.Equal(t, 2, len(f.conn().json.(webview.View).Resources))
 }
 
+func TestSailRoomConnectedAction(t *testing.T) {
+	f := newFixture(t)
+	defer f.TearDown()
+	go f.store.Loop(f.ctx)
+	f.client.OnChange(f.ctx, f.store)
+
+	time.Sleep(10 * time.Millisecond)
```
nicks: are you looking for the waitForActions helper function? (getActions is kind of an idiosyncractic api that generally isn't called directly)

web/src/TabNav.scss
```diff
@@ -24,6 +24,14 @@
   border-left: 1px solid $color-gray-light;
 }
 
+.TabNav span#TabNav-container {
```
nicks: this...doesn't seem like a good idea. can you elaborate a bit on what you're trying to accomplish here?

web/src/HUD.tsx
```diff
@@ -173,7 +176,7 @@ class HUD extends Component<HudProps, HudState> {
       )
     }
 
-    let tabNavRoute = (t: ResourceView, props: RouteComponentProps<any>) => {
+    let tabNavRoute = (t: ResourceView, sailUrl: string, props: RouteComponentProps<any>) => {
```
nicks: you don't need the whole do-si-do of making this an argument and binding it? it's already available in the lexical closure.

internal/store/engine_state.go
maiamcc: ah yep, you're right

internal/engine/upper.go
maiamcc: 🤷‍♀️ gooood question.

internal/sail/client/client_test.go
maiamcc: ooooh! yes, i am!

web/src/TabNav.scss
nicks: (we don't really use IDs in our UI code, or floats...there are good reasons to use them but they aren't everyday tools that I would expect someone to need...also happy to pair on this)

web/src/HUD.tsx
maiamcc: shoot you're right!

web/src/TabNav.scss
maiamcc: Ah! yeah i have no problem using a class -- it just seemed weird to me to have a one-time object be a class instead of an id.

I'm trying to put the tabs at one end of this div and the sail url at another. i'd believe there's another way to do this, i just don't know it, so would love to pair

internal/sail/client/client_test.go
maiamcc: I'm leaving that for another PR though (see #1540 for refactor to make it available)

maiamcc:
Screen Shot 2019-04-25 at 5 38 58 PM

internal/sail/client/client_test.go
maiamcc: nvm, just did!

web/src/TabNav.tsx
```diff
@@ -49,8 +51,22 @@ class TabNav extends PureComponent<NavProps> {
...
+
+    if (this.props.sailUrl) {
+      spans.push(
+        <span className="TabNav-spacer" key="spacer">
+          &nbsp;
+        </span>
+      )
+      spans.push(
+        <span className="sail-url" key="sail-url">
+          Share this view! <a href={this.props.sailUrl}>{this.props.sailUrl}</a>
```
yuindustries: If it's not vital to see the link itself, perhaps "Share this view" can just have the link...thus taking up less horizontal space.

maiamcc: totes! don't wanna wait for tests again but i'll change in my next pass

PR #1539, opened 2019-04-25
engine: fix a bunch of bugs around live update where we needed both nil checks and Empty() checks

nicks:
Hello @maiamcc,

Please review the following commits I made in branch nicks/nullability:

81068ccb63afe3dc1bfb9e881a3f1f1b9d6c91ad (2019-04-25 15:09:08 -0400)
engine: fix a bunch of bugs around live update where we needed both nil checks and Empty() checks

### DIFF
--- a/internal/engine/build_and_deployer_live_update_test.go
+++ b/internal/engine/build_and_deployer_live_update_test.go
@@ -54,7 +54,7 @@ func runTestCase(t *testing.T, f *bdFixture, tCase testCase) {
 	manifest := tCase.baseManifest
 	iTarg := manifest.ImageTargetAt(0)
 	db := iTarg.DockerBuildInfo()
-	db.LiveUpdate = &tCase.liveUpdate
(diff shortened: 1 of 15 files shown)


nicks:
discovered this while i was looking into https://app.clubhouse.io/windmill/story/2351/liveupdate-should-sync-files-from-context-of-dependent-images, and decided to just clean it up

nicks:
@maiamcc sure! for example, this check:
https://github.com/windmilleng/tilt/blob/9bb881b3ffe6f6920924214699928690518f993c/internal/engine/extractors.go#L81

is always false. LiveUpdate is never nil. Because of the way tiltfile construction works, the model will always have an "empty" LiveUpdate{} object, not a nil pointer. does that make sense?

in an earlier version of the pr, i had changed all the nil checks to also check Empty(), but this seemed very likely to regress in the future. the real bug is that the data model has two different ways to express "this does not have a live update object"

PR #1541, opened 2019-04-25
titltfile: fix the impossible live update check when there are undeployed images

nicks:
Hello @maiamcc,

Please review the following commits I made in branch nicks/busted:

c448e728bc2badd0d24054ad2ce36814e9f27d07 (2019-04-25 18:38:27 -0400)
titltfile: fix the impossible live update check when there are undeployed images

### DIFF
--- a/internal/model/manifest.go
+++ b/internal/model/manifest.go
@@ -109,6 +109,16 @@ func (m Manifest) WithDeployTarget(t TargetSpec) Manifest {
 	return m
 }
 
+func (m Manifest) IsImageDeployed(iTarget ImageTarget) bool {
+	id := iTarget.ID()
+	for _, depID := range m.DeployTarget().DependencyIDs() {
+		if depID == id {
+			return true
+		}
+	}
+	return false
+}
+
 func (m Manifest) LocalPaths() []string {
 	// TODO(matt?) DC syncs should probably stored somewhere more consistent with Docker/Fast Build
 	switch di := m.deployTarget.(type) {
--- a/internal/tiltfile/tiltfile_state.go
+++ b/internal/tiltfile/tiltfile_state.go
@@ -711,8 +711,9 @@ func (s *tiltfileState) translateK8s(resources []*k8sResource) ([]model.Manifest
 			return nil, errors.Wrapf(err, "getting image build info for %s", r.name)
 		}
 
-		s.checkForImpossibleLiveUpdates(iTargets)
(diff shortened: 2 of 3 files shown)


internal/model/manifest.go
```diff
@@ -109,6 +109,16 @@ func (m Manifest) WithDeployTarget(t TargetSpec) Manifest {
 	return m
 }
 
+func (m Manifest) IsImageDeployed(iTarget ImageTarget) bool {
+	id := iTarget.ID()
+	for _, depID := range m.DeployTarget().DependencyIDs() {
```
maiamcc: checking my understanding: only images that we care about deploying show up as deps of a DeployTarget (otherwise, they'd be deps of an image target representing the thing we actually want to deploy)?

sorta ambiguous function name, could maybe mean "is the image live in the cluster?" -- consider clarifying with a comment or renaming the func

internal/tiltfile/tiltfile_state.go
```diff
@@ -727,13 +728,29 @@ func (s *tiltfileState) translateK8s(resources []*k8sResource) ([]model.Manifest
...
+			continue
+		}
+
+		// TODO(nick): If an undeployed base image has a live-update component, we
+		// should probably emit a different kind of warning.
+		if !isDeployed {
+			continue
+		}
+
+		if !isFirstDeployedImage {
```
maiamcc: hmm we probably don't want this whoooole error string for EVERY subsequent liveUpdated image -- maybe either `break` out after we find one impossible liveUpdate, or collect the image names and drop them into a single warning

internal/model/manifest.go
nicks: hmmmm...this "isDeployed" terminology is already used a bunch of other places in the codebase. if we decide to change it here, let's do it everywhere

"Live" seems like a much worse name, because it's likely to get confused with live update

internal/tiltfile/tiltfile_state.go
nicks: done!

internal/model/manifest.go
maiamcc: oh oh, I was NOT suggesting that we use "live" -- i was saying that "isDeployed" is confusing btwn "this is an image that we deploy, as opposed to an image that we don't" and "this is an image that is deployed to the cluster, as opposed to one that has not yet been sent to the cluster"

but sure, I don't have strong feelings

PR #1546, opened 2019-04-26
model: add a helper struct for traversing image graphs

nicks:
Hello @maiamcc,

Please review the following commits I made in branch nicks/liveupdategraph:

83c6780b60f53b11fc5d8957bef11985c6b857b8 (2019-04-26 15:20:40 -0400)
model: add a helper struct for traversing image graphs

### DIFF
--- a/internal/model/target_graph.go
+++ b/internal/model/target_graph.go
@@ -1,0 +1,123 @@
+package model
+
+import "fmt"
+
+type TargetGraph struct {
+	// Targets in topological order.
+	sortedTargets []TargetSpec
+	byID          map[TargetID]TargetSpec
+}
+
+func NewTargetGraph(targets []TargetSpec) (TargetGraph, error) {
+	sortedTargets, err := TopologicalSort(targets)
+	if err != nil {
+		return TargetGraph{}, err
+	}
+
+	return TargetGraph{
+		sortedTargets: sortedTargets,
+		byID:          MakeTargetMap(sortedTargets),
+	}, nil
+}
+
+// In Tilt, Manifests should always be DAGs with a single root node
+// (the deploy target). This is just a quick sanity check to make sure
+// that's true, because many of our graph-traversal algorithms won't work if
+// it's not true.
+func (g TargetGraph) IsSingleSourceDAG() bool {
+	seenIDs := make(map[TargetID]bool, len(g.sortedTargets))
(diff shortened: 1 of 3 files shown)


internal/model/target_graph.go
```diff
@@ -0,0 +1,119 @@
...
+			return err
+		}
+		visitedIDs[current.ID()] = true
+		return nil
+	}
+
+	return helper(root)
+}
+
+func (g TargetGraph) DepsOf(t TargetSpec) ([]TargetSpec, error) {
```
maiamcc: so this, like `t.DependencyIDs()`, is only getting the DIRECT deps of this target? and the only difference btwn this and `t.DependencyIDs()` is that this also verifies that the dep is in the graph?

internal/model/target_graph.go
```diff
@@ -0,0 +1,119 @@
...
+	depIDs := t.DependencyIDs()
+	result := make([]TargetSpec, len(depIDs))
+	for i, depID := range depIDs {
+		dep, ok := g.byID[depID]
+		if !ok {
+			return nil, fmt.Errorf("Dep not found in graph: %s", depID)
```
maiamcc: nit: "Dep %s not found in graph" (otherwise it looks like you're printing the graph when you're just printing the dep id.)

internal/model/target_graph.go
```diff
@@ -0,0 +1,119 @@
...
+	}
+	return result, nil
+}
+
+func (g TargetGraph) IsDeployedImage(iTarget ImageTarget) bool {
+	id := iTarget.ID()
+	for _, t := range g.sortedTargets {
+		switch t := t.(type) {
+		case K8sTarget, DockerComposeTarget:
+			for _, depID := range t.DependencyIDs() {
```
maiamcc: Logic kinda opaque, maybe a comment. "If an image target is a direct dependency of a deploy target..."

internal/model/target_graph.go
nicks: t.DependencyIDs() can't give you the TargetSpec

internal/model/target_graph.go
nicks: done

internal/model/target_graph.go
nicks: done

PR #1548, opened 2019-04-26
assets: move AssetServer to its own package because it's shared

maiamcc:
Hello @nicks, @landism,

Please review the following commits I made in branch maiamcc/asset-server-pkg:

82f19200dde4f38e8bf6e6b5d4845b068aadbf8f (2019-04-26 16:12:07 -0400)
assets: move AssetServer to its own package because it's shared

### DIFF
--- a/internal/sail/sail.go
+++ b/internal/sail/sail.go
@@ -10,7 +10,7 @@ import (
 	"time"
 
 	"github.com/spf13/cobra"
-	hudServer "github.com/windmilleng/tilt/internal/hud/server"
+	assets2 "github.com/windmilleng/tilt/internal/assets"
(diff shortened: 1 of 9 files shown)


internal/sail/sail.go
```diff
@@ -10,7 +10,7 @@ import (
 	"time"
 
 	"github.com/spf13/cobra"
-	hudServer "github.com/windmilleng/tilt/internal/hud/server"
+	assets2 "github.com/windmilleng/tilt/internal/assets"
```
landism: was this intentional or done by a refactor rename?

(it seems weird to rename the package rather than the var)

PR #1549, opened 2019-04-26
engine: A BuildState's FileChanged field should only contain files that the image depends on. Fix up the tests to accurately reflect this

nicks:
Hello @maiamcc,

Please review the following commits I made in branch nicks/stateset:

ff1b4fc29647b002e1ce05daad79e7a48359658e (2019-04-26 17:06:57 -0400)
engine: A BuildState's FileChanged field should only contain files that the image depends on. Fix up the tests to accurately reflect this

### DIFF
--- a/internal/engine/build_and_deployer_live_update_test.go
+++ b/internal/engine/build_and_deployer_live_update_test.go
@@ -8,7 +8,6 @@ import (
 
 	"github.com/windmilleng/tilt/internal/k8s"
 	"github.com/windmilleng/tilt/internal/model"
-	"github.com/windmilleng/tilt/internal/store"
 	"github.com/windmilleng/tilt/internal/synclet/sidecar"
 )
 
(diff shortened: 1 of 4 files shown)


internal/engine/build_and_deployer_live_update_test.go
```diff
@@ -42,13 +43,36 @@ type testCase struct {
 }
 
 func runTestCase(t *testing.T, f *bdFixture, tCase testCase) {
-	var bs store.BuildStateSet
-	if len(tCase.changedFiles) > 0 {
-		changed := make([]string, len(tCase.changedFiles))
-		for i, file := range tCase.changedFiles {
-			changed[i] = f.WriteFile(file, "blah")
+	bs := store.BuildStateSet{}
```
maiamcc: nice! consider putting this in a helper function so it doesn't look like part of the test logic.

nicks: great idea! done

PR #1550, opened 2019-04-26
sail/web: button to trigger sail room join [ch2282]

maiamcc:
Styling tdb, but...

tiltfwd button

### DIFF
--- a/internal/hud/server/controller.go
+++ b/internal/hud/server/controller.go
@@ -32,7 +32,7 @@ func (s *HeadsUpServerController) Teardown(ctx context.Context) {
 
 func (s *HeadsUpServerController) OnChange(ctx context.Context, st store.RStore) {
 	if s.hudServer.sailCli != nil {
-		s.hudServer.sailCli.MaybeBroadcast(ctx, st)
+		s.hudServer.sailCli.MaybeBroadcast(st)
 	}
 
 	defer func() {
--- a/internal/hud/server/controller.go
+++ b/internal/hud/server/controller.go
@@ -32,7 +32,7 @@ func (s *HeadsUpServerController) Teardown(ctx context.Context) {
 
 func (s *HeadsUpServerController) OnChange(ctx context.Context, st store.RStore) {
 	if s.hudServer.sailCli != nil {
-		s.hudServer.sailCli.MaybeBroadcast(st)
+		s.hudServer.sailCli.MaybeBroadcast(ctx, st)
 	}
 
 	defer func() {
--- a/internal/hud/server/controller.go
+++ b/internal/hud/server/controller.go
@@ -31,6 +31,10 @@ func (s *HeadsUpServerController) Teardown(ctx context.Context) {
 }
 
 func (s *HeadsUpServerController) OnChange(ctx context.Context, st store.RStore) {
+	if s.hudServer.sailCli != nil {
+		s.hudServer.sailCli.MaybeBroadcast(ctx, st)
+	}
+
 	defer func() {
 		s.initDone = true
 	}()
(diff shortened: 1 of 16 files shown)


internal/hud/server/controller.go
```diff
@@ -31,6 +31,10 @@ func (s *HeadsUpServerController) Teardown(ctx context.Context) {
 }
 
 func (s *HeadsUpServerController) OnChange(ctx context.Context, st store.RStore) {
+	if s.hudServer.sailCli != nil {
+		s.hudServer.sailCli.MaybeBroadcast(st)
```
nicks: it seems very weird for the hudServer to be responsible for this...can you elaborate a bit on what was wrong with making sailCli a Subscriber?

internal/sail/client/client.go
```diff
@@ -69,11 +88,11 @@ func (s *SailClient) broadcast(ctx context.Context, view webview.View) {
 
 	err := s.conn.WriteJSON(view)
 	if err != nil {
-		logger.Get(ctx).Infof("broadcast(%s): %v", s.addr, err)
+		logger.Get(s.ctx).Infof("broadcast(%s): %v", s.addr, err)
```
nicks: i'm pretty sure this context won't have a logger attached, so this will be an error

nicks: as a rule of thumb, "Context" is for passing along information about the current call context, like tracing information.

If you're storing a context on an object that's alive for the entire program, you're probably using it wrong. In all of the places you're using it in this file, it'd be better to pass a context in

internal/hud/server/controller.go
maiamcc: ...nothing? like, it felt a little weird and i wanted a gut check, and then I thought _you_ suggested putting the sail cli on the hud server? I can put it back.

internal/sail/client/client.go
maiamcc: yeah this was all an artifact of trying to get a ctx into these functions that were being called from inside the hud http handler.

internal/hud/server/controller.go
nicks: afair, you asked me how to trigger a Connect(), and I suggested calling Connect() from hudServer. I wouldn't expect that to imply moving all method calls here

internal/sail/client/client.go
nicks: fwiw, there should be a req.Context() that you can get in the http handler

maiamcc: oh, amazing!

web/src/SailInfo.tsx
```diff
@@ -19,7 +28,9 @@ class SailInfo extends PureComponent<SailProps> {
 
       return (
         <span className="SailInfo">
-          <button type="button">Share me!</button>
+          <button type="button" onClick={SailInfo.newSailRoom}>
```
yuindustries: For the actual link, you can do this to have it open in a new window:

```
<a target="_blank" href={this.props.sailUrl}>Share this view!</a>
```

internal/hud/server/server.go
```diff
@@ -80,3 +86,16 @@ func (s HeadsUpServer) HandleAnalytics(w http.ResponseWriter, req *http.Request)
...
+
+func (s HeadsUpServer) HandleSail(w http.ResponseWriter, req *http.Request) {
+	if req.Method != http.MethodPost {
+		http.Error(w, "must be POST request", http.StatusBadRequest)
+		return
+	}
+
+	err := s.sailCli.Connect(req.Context(), s.store)
+	if err != nil {
+		s.store.Dispatch(store.NewErrorAction(errors.Wrap(err, "sailClient")))
```
nicks: this will make Tilt crash, which is probably not what you want. would be better to return an http.Error (like on the line above) and have the UI handle the error properly.

maiamcc: 😅

PR #1554, opened 2019-04-29
k8s: add a work-around for bugs in container status image refs

nicks:
Hello @maiamcc,

Please review the following commits I made in branch nicks/badtag:

aea59f869406825b1bb9d444ba64883268c5c678 (2019-04-29 17:09:34 -0400)
k8s: add a work-around for bugs in container status image refs
https://github.com/kubernetes/kubernetes/issues/51017

### DIFF
--- a/internal/k8s/container.go
+++ b/internal/k8s/container.go
@@ -6,7 +6,7 @@ import (
 	"strings"
 
 	"github.com/pkg/errors"
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
 
 	"github.com/windmilleng/tilt/internal/container"
 	"github.com/windmilleng/tilt/internal/logger"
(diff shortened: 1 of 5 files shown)


internal/k8s/container.go
```diff
@@ -6,7 +6,7 @@ import (
 	"strings"
 
 	"github.com/pkg/errors"
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
```
maiamcc: ?

internal/k8s/container_test.go
```diff
@@ -7,12 +7,33 @@ import (
 	"time"
 
 	"github.com/stretchr/testify/assert"
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
```
maiamcc: ?

internal/k8s/container.go
```diff
@@ -99,6 +101,37 @@ func IsUnschedulable(pod v1.PodStatus) (bool, string) {
 	return false, ""
 }
 
+// Kubernetes has a bug where the image ref in the container status can be wrong.
```
maiamcc: mayyybe worth mentioning
> Although it doesn't really make a difference as both image tags share the same digest it is technically wrong.

(i.e. "though this bug does NOT result in the container running unexpected code"). but also, 🤷‍♀

nicks: done

internal/k8s/container.go
nicks: this is just what goimports does

PR #1556, opened 2019-04-30
web: show resource info

yuindustries:


internal/hud/server/view.go
```diff
@@ -62,6 +62,11 @@ func StateToWebView(s store.EngineState) webview.View {
 
 		endpoints := store.ManifestTargetEndpoints(mt)
 
+		var podIDs []string
+		for id, _ := range ms.PodSet.Pods {
```
nicks: in other places in the UI, we've used GetMostRecentPod() to show the most relevant pod, rather than just showing all of them

internal/hud/webview/view.go
```diff
@@ -78,6 +78,7 @@ type Resource struct {
 	PendingBuildSince  time.Time
 
 	Endpoints []string
+	PodIDs    []string
```
nicks: this should probably be a `[]k8s.PodID`? the serializer will do the right thing

web/src/HUD.tsx
```diff
@@ -198,11 +199,22 @@ class HUD extends Component<HudProps, HudState> {
           ? props.match.params.name
           : ""
       let logs = ""
+      let endpoints: Array<string> = []
+      let podIDs: Array<string> = []
       if (view && name !== "") {
         let r = view.Resources.find(r => r.Name === name)
         logs = r ? r.CombinedLog : ""
+        endpoints = r ? r.Endpoints : []
```
nicks: fwiw, the `(r && r.Endpoints) || []` idiom is probably closer to what you want

web/src/LogPane.tsx
```diff
@@ -110,6 +112,17 @@ class LogPane extends Component<LogPaneProps, LogPaneState> {
       )
     }
 
+    let resourceInfo: React.ReactElement
+
+    let endpoints = ""
+    if (this.props.endpoints) {
+      endpoints = this.props.endpoints.join(", ")
```
nicks: it would be good to make each of these into links, instead of just text

web/src/LogPane.tsx
```diff
@@ -110,6 +112,17 @@ class LogPane extends Component<LogPaneProps, LogPaneState> {
...
+    }
+    let podIDs = ""
+    if (this.props.podIDs) {
+      podIDs = this.props.podIDs.join(", ")
```
nicks: similarly, it would be good to make each of these `user-select: all`

internal/hud/server/view.go
yuindustries: Huh, this entire file seems to be gone now...

internal/hud/webview/view.go
yuindustries: Ok, so `PodIDs` is type `[]k8s.PodID`, but `Endpoints` is OK as `[]string`?

nicks: ya, though i would be fine with creating an Endpoint typedef

web/src/LogPane.tsx
yuindustries: Agreed!

PR #1557, opened 2019-04-30
sail client: trigger OnChange when new room created so we connect/surface the url in the UI right away

maiamcc:
whoooops!

### DIFF
--- a/internal/engine/upper.go
+++ b/internal/engine/upper.go
@@ -170,6 +170,8 @@ var UpperReducer = store.Reducer(func(ctx context.Context, state *store.EngineSt
 		handleStopProfilingAction(state)
 	case hud.SetLogTimestampsAction:
 		handleLogTimestampsAction(state, action)
+	case client.SailNewRoomAction:
+		// nothing to handle, we just need to trigger OnChange
 	case client.SailRoomConnectedAction:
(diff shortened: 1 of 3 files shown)


nicks:
this doesn't seem like a good idea. 

long-term, we expect the engine state to diff the model itself, and not fire onchange events for things that haven't changed.

shouldn't the UI redraw itself after it makes the POST? why does it need a message from the engine state?

maiamcc:
...doyyyy. I forgot where in the code I was. thank you.

maiamcc:
actually wait, I think it might be more complicated than that. right now we need the OnChange to _connect_ the room (in [sailClient.OnChange](https://github.com/windmilleng/tilt/blob/a3e3df17191c0daab2f518d4cf66c99fa94c1862/internal/sail/client/client.go#L92)), so just redrawing before the room is CONNECTED will do us no good

previously we both created the room and connected to it from the client on the hud server, but I had to move the actual connection logic b/c it requires a persistent context, not a short-lived one like you get from `req.Context`.

Am I missing something here?

nicks:
let's talk about this in-person, maybe i'm misunderstanding something. 

Under the current architecture, firing an empty action makes no sense and is a pretty big red flag that there's a bug.

PR #1558, opened 2019-04-30
store: add optional SetUp method for subscribers

maiamcc:
Hello @nicks,

Please review the following commits I made in branch maiamcc/subscriber-setup:

2f2613ba1668f3acacf8a0c332dbf6c1fb54d52e (2019-04-30 14:33:55 -0400)
store: add optional SetUp method for subscribers

### DIFF
--- a/internal/assets/server.go
+++ b/internal/assets/server.go
@@ -28,7 +28,7 @@ import (
 type Server interface {
 	http.Handler
 	Serve(ctx context.Context) error
-	Teardown(ctx context.Context)
+	TearDown(ctx context.Context)
 }
 
 type devServer struct {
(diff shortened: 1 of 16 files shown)


maiamcc:
Unsure if the behavior in the second commit is right, would love feedback.

maiamcc:
...was trying to follow the naming conventions already there but it's maybe simpler to call `s.SetUp()` when on `Add(s)` and not call any individual `SetUp` methods during `subscriberList.SetUp()` -- objections?

nicks:
makes sense to me!

nicks:
not sure i really understand your last question

maiamcc:
🤷‍♀ then whatever, i'll leave it as is! i don't feel strongly

PR #1560, opened 2019-04-30
local dev: make instructions for sail image with go cache

maiamcc:
if you told me to optimize local dev on sail, this is what i'd do.
        is this... what we want to encourage people to do? an anti-pattern?
        do you have other ideas?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -41,6 +41,13 @@ install-debug:
 install-sail:
 	go install ./cmd/sail/...
 
+build-sail-cache:
+	docker build . -t sail-with-go-cache -f deployments/sail.cache.dockerfile
+
+run-sail:
+	@if ! docker inspect --type=image sail-with-go-cache 2> /dev/null 1>/dev/null; then echo "Docker image sail-with-go-cache not found, run make build-sail-cache"; exit 1; fi
+	tilt up --port 10351 --webdev-port 46765
+
 define synclet-build-dev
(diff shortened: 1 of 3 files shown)


nicks:
have you tried using the cache argument to docker_build? maybe we should be encouraging that for people with go projects?

maiamcc:
you mean `--cache-from`? as far as i can tell that's just about images to use as the DOCKER cache, and i assume that at the point at which we change the Go code, docker won't use any layer that has different go source code in it (i.e. won't use a layer that we'd previously run `go build` on?)

or is there another cache arg? can you point me to a blog post or something?

nicks:
we don't really advertise it, because it was obsoleted by fast_build. It's
worth pointing out that your current PR re-invents the iterative image
builds from fast_build.

nicks:
it might be faster for me to just send you a pr

maiamcc:
sure, that'd be great.

basically, yeah -- though i'm trying to optimize initial build here (because live update is pretty good for all subsequent builds) and the iterative build for Fast Build was never for the initial build.

nicks:
@maiamcc here, try this
https://github.com/windmilleng/tilt-golang-demo/pull/1

we need better docs and recommendations on how to use it, but it speeds up the go fetch/compile a lot. should probably switch it over to Go modules and write a tutorial around it.

PR #1566, opened 2019-05-01
web: refine design for StatusBar

yuindustries:
2019-05-01 - Web - footer


web/src/Statusbar.tsx
```diff
@@ -44,52 +44,52 @@ type StatusBarProps = {
...
+            <span className="count">{errorCount}</span> error
+            {errorCount === 1 ? "" : "s"}
+          </p>
+        </div>
+        <div className="err-warn-item">
+          <WarningSvg
+            className={`icon ${warningCount > 0 ? "icon--warning" : ""}`}
+          />
+          <p>
+            <span className="count">{warningCount}</span> warning
```
nicks: i'm worried that having a bunch of small, unnamespaced class names like "count" are likely to become unmanageable fast. it would be better to namespace these in some way

web/src/Statusbar.tsx
```diff
@@ -44,52 +44,52 @@ type StatusBarProps = {
...
-      </span>
-    )
+  progressPanel(upCount: number, itemCount: number) {
     return (
-      <div className={warningPanelClasses}>
-        {icon}&nbsp;{message}
-      </div>
+      <section className="Statusbar-panel progress">
+        <p>
+          <span className="current">{upCount}</span>/{itemCount} running
```
nicks: ditto re:current

web/src/Statusbar.tsx
```diff
@@ -44,52 +44,52 @@ type StatusBarProps = {
...
-        {upCount} / {itemCount} resources up <LogoSvg className="icon" />
-      </span>
+  statusMessagePanel(build: any, editMessage: string) {
+    return (
+      <section className="Statusbar-panel statusMsg">
+        <p className="statusMsg-item">
+          {combinedStatusMessage(this.props.items)}
+        </p>
+        <p className="statusMsg-item statusMsg-item--last-edit">
+          <span className="label">Last Edit:</span>
```
nicks: probably would be better as a label element

web/src/Statusbar.tsx
```diff
@@ -44,52 +44,52 @@ type StatusBarProps = {
...
+          <span className="file">{build ? editMessage : "—"}</span>
```
nicks: ditto re: file

web/src/Statusbar.tsx
yuindustries: I agree!

PR #1569, opened 2019-05-02
web: implement "all" preview view

jazzdan:
It looks extremely basic, but Han and I are going to refine it later.

Screen Shot 2019-05-02 at 2 03 31 PM

### DIFF
--- a/web/src/PreviewList.tsx
+++ b/web/src/PreviewList.tsx
@@ -1,0 +1,32 @@
+import React, { PureComponent } from "react"
+import PathBuilder from "./PathBuilder"
+import { Link } from "react-router-dom"
+import "./PreviewList.scss"
+
+type PreviewListProps = {
+  pathBuilder: PathBuilder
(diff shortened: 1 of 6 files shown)


web/src/PreviewList.tsx
```diff
@@ -0,0 +1,32 @@
...
+}
+
+class PreviewList extends PureComponent<PreviewListProps> {
+  render() {
+    let pb = this.props.pathBuilder
+    if (this.props.resourcesWithEndpoints.length === 0) {
+      return <p>No resources found with endpoints!</p>
+    }
+    let links = this.props.resourcesWithEndpoints.map((r, i) => (
+      <li key={r + i}>
```
nicks: shouldn't the key just be `r`?

jazzdan: I was future proofing a little bit here and imagining a future where there are multiple endpoints for the same name. But then the list wouldn't make sense. So yeah I'll just delete that.

PR #1570, opened 2019-05-02
assets: prod asset server can accept version to serve as query param [ch2284]

maiamcc:
Hello @nicks, @jazzdan,

Please review the following commits I made in branch maiamcc/specify-asset-version:

de425dd03db3c61cc3cac7df5a57ec0e0a3b203e (2019-05-02 13:59:54 -0400)
respect version query param

558a14aafb93c2cbe7ec3bef40a64c08206bff7e (2019-05-02 12:21:31 -0400)
preserve existing behavior constructing with default url

### DIFF
--- a/internal/assets/prod_server_test.go
+++ b/internal/assets/prod_server_test.go
@@ -10,9 +10,9 @@ import (
 )
 
 const (
-	testUrl     = "https://fake.tilt.dev"
-	testVersion = model.WebVersion("v1.2.3")
-	version666  = model.WebVersion("v6.6.6")
+	testUrl        = "https://fake.tilt.dev"
+	versionDefault = model.WebVersion("v1.2.3")
+	version666     = model.WebVersion("v6.6.6")
 )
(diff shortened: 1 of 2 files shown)


nicks:
hmmm...i don't think this does what we want?

we want the instance of Tilt that creates the room to set the version, not the person viewing the room

maiamcc:
so this doesn't do anything for the sail use case yet, it's just ensuring that the prod server can give you whatever version of the assets you ask it for.

I've got a change in progress to a. attach version info to a room when it's created, and b. send that version along to the asset server when actually making the request for assets (i.e. [attaching a query param here](https://github.com/windmilleng/tilt/blob/3ccd424e2fd5e84790a6caadb9a5490c3d958e16/internal/sail/server/server.go#L166). (So, the request to `<sailserver>/view/<room-id>` doesn't need to set any query params, that's all taken care of internally.

does that make sense?

PR #1572, opened 2019-05-02
store: put tiltInfo (i.e. info on the current build of Tilt) on engine state so it's accessible from within the app w/o wiring it individually into components

maiamcc:
Hello @nicks, @jazzdan,

Please review the following commits I made in branch maiamcc/tilt-info-on-store:

14324a3584df3c03835a3a0e298068ee6ef6cccd (2019-05-02 15:40:34 -0400)
store: put tiltInfo (i.e. info on the current build of Tilt) on engine state so it's accessible from within the app w/o wiring it individually into components

### DIFF
--- a/internal/cli/wire.go
+++ b/internal/cli/wire.go
@@ -10,6 +10,7 @@ import (
 	"github.com/docker/docker/api/types"
 	"github.com/google/wire"
 	"github.com/windmilleng/tilt/internal/assets"
+	"github.com/windmilleng/tilt/internal/model"
 	"k8s.io/apimachinery/pkg/version"
 	"k8s.io/client-go/tools/clientcmd/api"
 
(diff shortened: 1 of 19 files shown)


internal/store/engine_state.go
```diff
@@ -254,8 +255,8 @@ type ManifestState struct {
 	ConfigFilesThatCausedChange []string
 }
 
-func NewState() *EngineState {
-	ret := &EngineState{}
+func NewState(b model.TiltBuild) *EngineState {
```
nicks: hmmm...it seems weird to invent a new way to pass in initial data.

have you considered using InitAction?

it might also make sense to inject this data into SailClient and have it pass it up directly; i could see arguments before or against putting this in engine state.

maiamcc: i _did_ consider initAction, went with this instead but not for any particular reason, just figured i already had a `tiltBuild` floating around during wiring and might as well use it. can change.

if you feel strongly i can def put this data straight onto SailClient -- but it's not really a property OF the SailClient, more of the run as a whole, which is why my instinct was EngineState.

internal/cli/wire.go
```diff
@@ -113,12 +114,13 @@ func wireThreads(ctx context.Context) (Threads, error) {
 }
 
 type Threads struct {
-	hud   hud.HeadsUpDisplay
-	upper engine.Upper
+	hud       hud.HeadsUpDisplay
+	upper     engine.Upper
+	tiltBuild model.TiltBuild
```
maiamcc: i... _assume_ this is an appropriate way to use threads? i.e. for stuff you want to be able to access after all the wiring is complete?

PR #1574, opened 2019-05-02
web: misc cleanup

jazzdan:
Using === instead of ==, analytics incr somehow got deleted, etc


web/src/Sidebar.tsx
```diff
@@ -80,6 +81,7 @@ class Sidebar extends PureComponent<SidebarProps> {
         analyticsKey = "ui.interactions.errors"
         link += "/errors"
       }
+      incr(analyticsKey)
```
nicks: hmmm...this doesn't seem right? let's talk about this in-person

PR #1581, opened 2019-05-03
web: bucketize durations under a minute

landism:
A la the TUI

image

### DIFF
--- a/web/src/Sidebar.test.tsx
+++ b/web/src/Sidebar.test.tsx
@@ -77,29 +77,29 @@ describe("sidebar", () => {
   })
 
   it("abbreviates durations under a minute", () => {
-      let items = [4, 9, 19, 29, 39, 49, 54].map(d => {
-          let res = oneResource()
-          res.Name = `resource${d}`
-          res.LastDeployTime = Date.now() - d * 1000
-          return new SidebarItem(res)
-      })
+    let items = [4, 9, 19, 29, 39, 49, 54].map(d => {
+      let res = oneResource()
+      res.Name = `resource${d}`
+      res.LastDeployTime = Date.now() - d * 1000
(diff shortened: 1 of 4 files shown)


web/src/Sidebar.test.tsx
```diff
@@ -11,7 +11,7 @@ let pathBuilder = new PathBuilder("localhost", "/")
 
 describe("sidebar", () => {
   beforeEach(() => {
-    Date.now = jest.fn(() => 1482363367071)
+    Date.now = jest.fn(() => Date.UTC(2017, 11, 21, 15, 36, 7, 0))
```
landism: just to make this human-readable

web/src/testdata.test.tsx
```diff
@@ -49,7 +49,7 @@ function getMockRouterProps<P>(data: P) {
 }
 
 function oneResource(): any {
-  const ts = Date.now().toLocaleString()
+  const ts = Date.now().valueOf()
```
landism: it turns out `.toLocaleString` is inherited from `number`, and returns a human-readable formatting of the number of milliseconds since epoch, which looks like `1,482,363,367,071`.

timeago passes that to `Date.parse`, which rejects it because of the commas, and then kind of assumes that since there are at least three non-contiguous numbers (i.e., `/\d+/` in the string), the second one must be a month, which means it needs to have one subtracted to make it 0-based, and then reconcatenates them ([code](https://github.com/nmn/react-timeago/blob/69b34a295f3c38e53064ce084767c7c294e8ef04/src/dateParser.js#L13)), which is why the old snapshot timestamp was 1942 and timeago rendered it as 75yr

landism:
> Should we add a snapshot that renders a sidebar within these new time durations

done

PR #1583, opened 2019-05-06
tiltfile: improve error message when services on the cli aren't found

nicks:

Please review the following commits I made in branch nicks/msg:

ec3fe8fc76aaa68db319bc0c776bf922cb5fa426 (2019-05-06 11:54:44 -0400)
tiltfile: improve error message when services on the cli aren't found

### DIFF
--- a/internal/tiltfile/tiltfile_state.go
+++ b/internal/tiltfile/tiltfile_state.go
@@ -682,8 +682,10 @@ func match(manifests []model.Manifest, matching map[string]bool) ([]model.Manife
 			}
 		}
 
-		return nil, fmt.Errorf("Could not find resources: %s. Existing resources in Tiltfile: %s",
-			strings.Join(missing, ", "), strings.Join(unmatchedNames, ", "))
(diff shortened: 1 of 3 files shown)


nicks:
someone complained about this error message in k8s slack over the weekend. they had a dangling 'false' at the end of their cli invocation and were confused about what the message was telling them

internal/tiltfile/tiltfile_state.go
```diff
@@ -678,11 +678,12 @@ func match(manifests []model.Manifest, matching map[string]bool) ([]model.Manife
 		var missing []string
 		for k := range matching {
 			if !matched[k] {
-				missing = append(missing, k)
+				missing = append(missing, fmt.Sprintf("%q", k))
```
landism: maybe `%q` `unmatchedNames` a few lines up as well, for consistency?

nicks: i like that, broke it out into a helper

PR #1585, opened 2019-05-06
engine: remove global yaml (instead send it as a regular manifest with a special name)

maiamcc:
Hello tbd,

Please review the following commits I made in branch maiamcc/rm-gyaml:

e046ab197b1f738ebcad6860fcb93c6374561670 (2019-05-06 14:46:45 -0400)
additional find and rm

d971dfc33f8fb795eb3cbd233202dd0c2c61a962 (2019-05-06 14:46:09 -0400)
fix up tests

b2f4424e3d269fb398fcfb8fdde741294787872c (2019-05-06 14:38:19 -0400)
wip

10c0e1df257014c4a0aeead1ca50deab0c268c5c (2019-05-06 13:57:28 -0400)
rewire, rm actions and controllers

bc6add874486ef0dddd88763e46c9ae45ea648a6 (2019-05-06 13:22:09 -0400)
remove global from structs/fix tiltfile


internal/hud/webview/convert.go
```diff
@@ -143,6 +121,11 @@ func tiltfileResourceView(s store.EngineState) Resource {
 }
 
 func resourceInfoView(mt *store.ManifestTarget) ResourceInfoView {
+	if mt.Manifest.Name == model.UnresourcedYAMLManifestName {
```
jazzdan: We've made this check a couple of times. Should we make it a method on the manifest?

nicks: i wonder if the check here shouldn't be for the name, but rather for k8s manifests that have no image targets

maiamcc: i don't think so. we might have k8s manifest w/o image target that ARE still workloads, i.e. we want to display them like any other manifest; we only want this special view for the non-pod-creating leftover yaml.

maiamcc: ooh maybe!

maiamcc:
yeah i agree that the test failures are legitimate, i can (finally) replicate them on local (after wrestling with my integration tests for the better part of an afternoon). looking into it.

maiamcc:
oh damnit it's [this bullshit](https://github.com/windmilleng/tilt/blob/f27767ab67d1d26021a280c04cf0b194a3862917/internal/store/store.go#L185) that i dug into with AdamB a bit ago -- when running integration test (with `--watch=false`), we specify one manifest, we expect one manifest to be built, so we quit after we've built one manifest -- except now k8s_yaml is ALSO a manifest, so we don't build everything that we should.

PR #1589, opened 2019-05-07
k8s: don't accidentally treat 'unable to create new content in terminating namespace' errors as 'immutable field' errors

maiamcc:
Hello @nicks, @landism,

Please review the following commits I made in branch maiamcc/immutable-field-err:

e695be25b6672aa5d888796608608e5850e5f922 (2019-05-07 15:25:38 -0400)
k8s: don't accidentally treat 'unable to create new content in terminating namespace' errors as 'immutable field' errors

### DIFF
--- a/internal/k8s/client.go
+++ b/internal/k8s/client.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io"
 	"net/url"
+	"regexp"
 	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


maiamcc:
I was investigating integration test failures for #1585 and couldn't actually run them on local. Turns out that for whatever reason it was taking a long time to delete the `tilt-integration` namespace, but `tilt down` didn't wait on it -- so the next test would start, try to do apply a resource to that namespace, and get:
> Error from server (Forbidden): error when creating "STDIN": deployments.apps "oneup" is forbidden: unable to create new content in namespace tilt-integration because it is being terminated
except we'd treat it as an immutable field error cuz we were just checking for `contains("forbidden")` lulz. We then tried to force-apply and that didn't work either, so none of the BEHAVIOR would have been different, but the error message was super misleading.

PR #1590, opened 2019-05-07
integration: block integration test teardown on namespace tilt-integration actually being torn down

maiamcc:
I couldn't run integration tests on my local b/c for some reason it was taking a long time to delete the `tilt-integration namespace` (maybe docker being its usual mess of a self? cuz it's better since I :whale::bomb:'d), but `tilt down` didn't wait on it actually being torn down -- so the next test would start, try to apply a resource to that namespace, and get an error message b/c it was still terminating.

this PR blocks teardown until the `tilt-integration` namespace no longe exists

### DIFF
--- a/integration/imagetags/app.yaml
+++ b/integration/imagetags/app.yaml
@@ -1,10 +1,3 @@
-apiVersion: v1
-kind: Namespace
-metadata:
-  name: tilt-integration
-  labels:
-    name: tilt-integration
----
 apiVersion: apps/v1
 kind: Deployment
 metadata:
--- a/integration/imagetags/app.yaml
+++ b/integration/imagetags/app.yaml
@@ -1,3 +1,10 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: tilt-integration
+  labels:
+    name: tilt-integration
+---
 apiVersion: apps/v1
 kind: Deployment
 metadata:
--- a/internal/dockercompose/client.go
+++ b/internal/dockercompose/client.go
@@ -48,7 +48,12 @@ func NewDockerComposeClient(env docker.Env) DockerComposeClient {
 }
 
 func (c *cmdDCClient) Up(ctx context.Context, configPath string, serviceName model.TargetName, shouldBuild bool, stdout, stderr io.Writer) error {
-	args := []string{"-f", configPath, "up", "--no-deps", "-d"}
+	var args []string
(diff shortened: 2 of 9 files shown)


integration/imagetags/app.yaml
```diff
@@ -1,3 +1,10 @@
+apiVersion: v1
```
nicks: i don't understand why we're copy and pasting the namespace into every app.yaml, rather than using https://github.com/windmilleng/tilt/blob/master/integration/namespace.yaml

internal/dockercompose/client.go
```diff
@@ -48,12 +48,7 @@ func NewDockerComposeClient(env docker.Env) DockerComposeClient {
 }
 
 func (c *cmdDCClient) Up(ctx context.Context, configPath string, serviceName model.TargetName, shouldBuild bool, stdout, stderr io.Writer) error {
-	var args []string
```
nicks: why did this change in this pr? this doesn't seem right?

maiamcc: oh, this was just me making the docker compose logs shut up because they make integration test output super gross. if you object i can put this back.

nicks: i'm not sure...though as a user, i would probably expect --verbose to also turn on docker-compose --verbose?

maiamcc: hmm. that's reasonable. i just wish `docker-compose --verbose` wasn't SO DAMN VERBOSE...

maiamcc:
cool, let's do this instead -- it's sliiiiightly less robust because we're only clearing deployments and services out of the namespace, but it saves us a minute on integration tests on CI (and way more on d4m). @nicks, want to re-review?

maiamcc:
(also, suggestions for how to better clear out the namespace? I poked around for a way to scrub EVERYTHING from a namespace but didn't find anything, i guess unless you call `kubectl delete pods,deployments,services,secrets... --all` and just list ALL the possible k8s resource types...? overkill?)

internal/dockercompose/client.go
maiamcc: i put this back

maiamcc:
@nicks i'm gonna go ahead and land this, but lmk if you have any more thoughts on a more robust way to clear out the namespace/feel like it's worth it to nuke every possible k8s resource type.

PR #1591, opened 2019-05-07
integration: test for `tilt up --watch=false` with no resource names

maiamcc:
while I'm here/fixing up [this ticket](https://app.clubhouse.io/windmill/story/1967/tilt-up-watch-false-doesn-t-work-anymore#activity-2469), figured i'd add an integration test


integration/watch_false/Tiltfile
```diff
@@ -0,0 +1,10 @@
+# -*- mode: Python -*-
+
+# If you get push errors, you can change the default_registry.
+# Create tilt_option.json with contents: {"default_registry": "gcr.io/my-personal-project"}
+# (with your registry inserted). tilt_option.json is gitignore'd, unlike Tiltfile
+default_registry(read_json('tilt_option.json', {})
+                 .get('default_registry', 'gcr.io/windmill-test-containers/servantes'))
```
nicks: it seems weird to use servantes for this?

maiamcc: ha it's copypasta from other integration tests

maiamcc:
good call, let me see what a unit test for this might look like.

PR #1593, opened 2019-05-08
web: nicer connection refused error with web-mode local

landism:
I kept finding this jarring and annoying on startup. This still ain't pretty, but it's maybe a little better?

### DIFF
--- a/internal/assets/server.go
+++ b/internal/assets/server.go
@@ -294,7 +294,7 @@ func ProvideAssetServer(ctx context.Context, webMode model.WebMode, webVersion m
 		handler.ErrorHandler = func(writer http.ResponseWriter, request *http.Request, e error) {
 			extra := ""
 			if strings.Contains(e.Error(), "connection refused") {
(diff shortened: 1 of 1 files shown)


internal/assets/server.go
```diff
@@ -289,9 +290,18 @@ func ProvideAssetServer(ctx context.Context, webMode model.WebMode, webVersion m
 		if err != nil {
 			return nil, errors.Wrap(err, "ProvideAssetServer")
 		}
+		handler := httputil.NewSingleHostReverseProxy(loc)
+		handler.ErrorHandler = func(writer http.ResponseWriter, request *http.Request, e error) {
+			extra := ""
+			if strings.Contains(e.Error(), "connection refused") {
+				extra = `"connection refused" expected on startup in self-built Tilt. Refreshing in a few seconds.<meta http-equiv="refresh" content="2">`
```
jazzdan: Can we tell if we're a release or built off of master in the code? If so could we adjust this message appropriately?

landism: maybe, with code changes. would it be cool to just say "web-mode=local" instead of getting into whether it's self-built?

jazzdan: Yeah totally, that's what I meant to say.

PR #1596, opened 2019-05-08
web: show number of errors next to errors tab

jazzdan:
Screen Shot 2019-05-08 at 2 37 09 PM

Future work that will happen in a different PR:
* Styling this number to be a bit better
* When you click on the errors tab the number gets reset to zero.

### DIFF
--- a/web/src/ErrorPane.tsx
+++ b/web/src/ErrorPane.tsx
@@ -30,6 +30,25 @@ class ErrorResource {
       }
     }
   }
+
+  public hasError() {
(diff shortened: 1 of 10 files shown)


yuindustries:
The `(2)` thing should correspond to the resource selected...

Screen Shot 2019-05-08 at 2 50 58 PM

jazzdan:
I refactored the Router out of the HUD to make testing things easier. Namely I now test that if you are looking at an individual resource the errors tab in the tabnav shows only the number of errors for that resource, not the global number of errors for all resources.

Adding @nicks as a reviewer since I talked over this with him.

PR #1597, opened 2019-05-08
engine: fetch version information from github

landism:
This PR ensures that `EngineState` has live version information. That information is not yet used by anything.


internal/engine/upper.go
```diff
@@ -12,10 +12,11 @@ import (
 
 	"github.com/opentracing/opentracing-go"
 	"github.com/pkg/errors"
-	"github.com/windmilleng/tilt/internal/sail/client"
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/labels"
 
+	"github.com/windmilleng/tilt/internal/sail/client"
+
```
jazzdan: Delete this empty line?

landism: That annoys me all the time. I wonder if goimports does that intentionally or if it'd be possible to fix that...

PR #1602, opened 2019-05-09
engine: reset container restarts action

jazzdan:
As a user I want all my resources to be green if they are behaving as expected. If I don't think that a container restart is cause for concern at the moment I should be able to clear it from a given resource.

This PR implements an API call that will be used from the frontend to clear container restart errors. It does that by setting the current container restart number and the old container restart number to the same number, thus hiding them in the UI. Then if another container restart came in we would still show the correct number of total restarts.


internal/cli/subscribers.go
```diff
@@ -0,0 +1,46 @@
...
+
+import (
+	"github.com/windmilleng/tilt/internal/engine"
+	"github.com/windmilleng/tilt/internal/hud"
+	"github.com/windmilleng/tilt/internal/hud/server"
+	"github.com/windmilleng/tilt/internal/sail/client"
+	"github.com/windmilleng/tilt/internal/store"
+)
+
+func ProvideSubscribers(
```
nicks: out of curiosity, why did this move? did you intend to delete the original one in engine, or did you intend to keep two copies of this function?

nicks: also, what happened to TiltVersionChecker? did that get dropped on the floor?

internal/engine/upper.go
```diff
@@ -222,6 +225,19 @@ func handleBuildStarted(ctx context.Context, state *store.EngineState, action Bu
 	removeFromTriggerQueue(state, mn)
 }
 
+func handleResetRestarts(ctx context.Context, engineState *store.EngineState, a server.ResetRestartsAction) {
```
nicks: it seems very weird for this action to be in the `server` package, given that it has nothing to do with servers

jazzdan: It can't go with the other actions in engine because then there's an import cycle:

```
import cycle not allowed
package github.com/windmilleng/tilt/cmd/tilt
	imports github.com/windmilleng/tilt/internal/cli
	imports github.com/windmilleng/tilt/internal/demo
	imports github.com/windmilleng/tilt/internal/engine
	imports github.com/windmilleng/tilt/internal/hud/server
	imports github.com/windmilleng/tilt/internal/engine
```

jazzdan: Open to other suggestions for where to put it though.

nicks: can you put it in store?

would also be open to a store/action package or smth

jazzdan: ah I didn't realize we had already had actions in different packages. Putting it in store worked.

PR #1604, opened 2019-05-09
sail: add a staging server for testing SSL

nicks:
Hello @maiamcc,

Please review the following commits I made in branch nicks/ch2495/staging:

cb9a7c324cf655d1bd9ad4d9278c44bb3c850729 (2019-05-09 16:41:03 -0400)
sail: add a staging server for testing SSL

### DIFF
--- a/internal/model/sail.go
+++ b/internal/model/sail.go
@@ -18,7 +18,13 @@ const (
 	// Local sail server on localhost:10350
 	SailModeLocal SailMode = "local"
 
-	// Production sail server at sail.tilt.dev
+	// Remote sail server on sail-staging.tilt.dev. Useful for testing SSL.
+	// Always uses precompiled JS, because one of the things we want to test
+	// with sail-staging is the JS at head (rather than testing the production JS).
+	SailModeStaging SailMode = "staging"
+
+	// Production sail server at sail.tilt.dev.
+	// Serves production JS according to the version of Tilt that opened the room.
 	SailModeProd SailMode = "prod"
 )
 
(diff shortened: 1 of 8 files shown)


internal/sail/client/roomer.go
```diff
@@ -31,7 +31,7 @@ func (r httpRoomer) NewRoom(ctx context.Context, version model.WebVersion) (room
 
 	resp, err := http.Post(u.String(), "text/plain", bytes.NewReader(reqJson))
 	if err != nil {
-		return "", "", errors.Wrapf(err, "GET %s", u.String())
+		return "", "", err
```
nicks: removed this wrapper because
1) This isn't even a GET, so the message was wrong
2) http.Post errors already include "post [url]", so this led to weird duplication

internal/model/sail.go
```diff
@@ -36,8 +42,10 @@ func (m *SailMode) Set(v string) error {
 		*m = SailModeLocal
 	case string(SailModeProd):
 		*m = SailModeProd
+	case string(SailModeStaging):
+		*m = SailModeStaging
 	default:
-		return UnrecognizedWebModeError(v)
+		return UnrecognizedSailModeError(v)
```
maiamcc: good catch 😅

internal/model/sail.go
```diff
@@ -48,12 +56,12 @@ func (m *SailMode) Type() string {
...
 func (m *SailMode) IsEnabled() bool {
 	mode := *m
-	return mode == SailModeLocal || mode == SailModeProd
+	return mode == SailModeLocal || mode == SailModeProd || mode == SailModeStaging
 }
 
 func UnrecognizedSailModeError(v string) error {
 	return fmt.Errorf("Unrecognized sail mode: %s. Allowed values: %s", v, []SailMode{
-		SailModeDefault, SailModeDisabled, SailModeLocal, SailModeProd,
+		SailModeDefault, SailModeDisabled, SailModeLocal, SailModeProd, SailModeStaging,
```
maiamcc: nit: swap Prod and Staging so they're in a reasonable order

nicks: done

PR #1605, opened 2019-05-09
web: global shortcut for clearing all container restarts

jazzdan:


web/src/HUD.tsx
```diff
@@ -139,6 +140,30 @@ class HUD extends Component<HudProps, HudState> {
...
+  keymap() {
+    return {
+      clearSnackRestarts: "ctrl+shift+9",
+    }
+  }
+
+  handlers() {
+    return {
+      clearSnackRestarts: (event: KeyboardEvent | undefined) => {
+        console.log("Clearing snack restarts!")
```
nicks: O.o

web/src/HUD.tsx
```diff
@@ -139,6 +140,30 @@ class HUD extends Component<HudProps, HudState> {
...
+        debugger
```
nicks: O.o

web/src/HUD.tsx
```diff
@@ -139,6 +140,30 @@ class HUD extends Component<HudProps, HudState> {
...
+        if (this.state.View) {
+          this.state.View.Resources.forEach(r => {
+            fetch(
+              `//${window.location.host}/api/control/reset_restarts?name=${
```
nicks: let's try to use path builder for constructing all URLs?

relatedly, i wonder if there should be a client-side-only way to hide restarts for sail.

PR #1607, opened 2019-05-10
server: only open the browser window if there isn't one open already

nicks:
Hello @jazzdan,

Please review the following commits I made in branch nicks/autoopen:

c1e59cd758da437a3764cd694921446e3c178b31 (2019-05-10 13:26:41 -0400)
server: only open the browser window if there isn't one open already

### DIFF
--- a/internal/hud/server/controller.go
+++ b/internal/hud/server/controller.go
@@ -3,7 +3,9 @@ package server
 import (
 	"context"
 	"net/http"
+	"sync/atomic"
 
+	"github.com/pkg/browser"
 	"github.com/pkg/errors"
 	"github.com/windmilleng/tilt/internal/assets"
 	"github.com/windmilleng/tilt/internal/model"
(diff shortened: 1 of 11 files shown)


nicks:
(i was getting really annoyed at the number of tilt windows that were auto-opening if I ran `tilt up` repeatedly...maybe this will feel better)

internal/hud/server/controller.go
```diff
@@ -13,24 +15,54 @@ import (
...
 	s.assetServer.TearDown(ctx)
 }
 
+func (s *HeadsUpServerController) maybeOpenBrowser(st store.RStore) {
+	if s.webURL.Empty() || s.webLoadDone {
+		return
+	}
+
+	connCount := atomic.LoadInt32(&(s.hudServer.numWebsocketConns))
+	if connCount > 0 {
```
jazzdan: Clever!

PR #1610, opened 2019-05-10
analyticsReporter: when sending data to web UI, check the state for whether we should nudge user about analytics [ch2529, ch2530]

maiamcc:
Hello @landism, @nicks,

Please review the following commits I made in branch maiamcc/needs-nudge-check:

1e64dc521efceb445c6d5b54e47d0692e10e0a56 (2019-05-10 17:10:17 -0400)
analyticsReporter: onChange, check the state for whether we should nudge user about analytics


internal/engine/analytics_reporter.go
```diff
@@ -30,6 +43,13 @@ func (ar *AnalyticsReporter) OnChange(ctx context.Context, st store.RStore) {
 	// wait until state has been kinda initialized
 	if !state.TiltStartTime.IsZero() && state.LastTiltfileError() == nil {
 		ar.started = true
+
+		opt, err := analytics.OptStatus()
```
maiamcc: sooo this isn't great because it's another place to keep the chosen opt consistent (i.e. when user chooses one way or another we need to remember to update their Opt here as well), but seemed better than reading the `choice.txt` file on every OnChange...?

maiamcc:
Wow this is what happens when you write code sleepy. I can't set fields on a read-only state. Making an event.

maiamcc:
Guh okay, ready for review now 😅

internal/engine/analytics_reporter.go
```diff
@@ -13,14 +15,25 @@ import (
...
 type AnalyticsReporter struct {
 	a       analytics.Analytics
 	store   *store.Store
 	started bool
+	opt     analytics.Opt
 }
 
 func (ar *AnalyticsReporter) OnChange(ctx context.Context, st store.RStore) {
 	if ar.started {
+		if newAnalyticsOn() {
```
landism: This might be cleaner as a separate class? Nudging is a pretty separable responsibility from periodically reporting.

This feels a bit weird to me. Previously, `AnalyticsReporter` was unaware of optin status. Now it's aware, but sends data to `ar.a` even if the user has opted out (because `ar.a` itself is aware of the opt-in status and discards data if the user has opted out).

I don't know if this would be considered dirty, but this logic might even be simple enough to just be in a new function called from `handleBuildCompleted`?

internal/engine/analytics_reporter.go
landism: It's fine to leave as-is in this PR since it won't matter until we get to the code that actually modifies the value. I kinda suspect we'll want to pull it into `EngineState`.

internal/engine/analytics_reporter.go
maiamcc: sigh. i hate that you may be right. but you may be right.

internal/engine/analytics_nudge.go
```diff
@@ -0,0 +1,39 @@
...
+)
+
+const newAnalyticsFlag = "TILT_NEW_ANALYTICS"
+
+// TEMPORARY: check env to see if new-analytics flag is set
+func newAnalyticsOn() bool {
+	return os.Getenv(newAnalyticsFlag) != ""
+}
+
+func MaybeSetNeedsNudge(mt *store.ManifestTarget, st *store.EngineState) {
```
nicks: what is this manifest target supposed to be? as a caller, i'm not sure i know what I'd need to pass in here

internal/engine/upper.go
```diff
@@ -293,6 +294,7 @@ func handleBuildCompleted(ctx context.Context, engineState *store.EngineState, c
 		}
 
 		ms.LastSuccessfulDeployTime = time.Now()
+		MaybeSetNeedsNudge(mt, engineState)
```
nicks: it seems very weird that this live in the handleBuildCompletion reducer. why should build completion have to know about analytics?

internal/store/engine_state.go
```diff
@@ -79,6 +80,10 @@ type EngineState struct {
 
 	// from GitHub
 	LatestTiltBuild model.TiltBuild
+
+	// Analytics Info
+	AnalyticsOpt        analytics.Opt // TODO(maia): remember to update this when user opts in/out
+	NeedsAnalyticsNudge bool          // TODO(maia): maybe this can convey more info, e.g. "there was error in /analytics/opt"
```
nicks: i don't understand why you need this state at all?

afaict, the "source of truth" is "have any builds successfully completed yet?"

but you can infer this from the existing engine state. Creating a new state variable makes it more likely that we'll introduce bugs, because the reducer will change and you'll forget to keep this up to date.

it'd be simpler and less buggy to have a function that takes an EngineState and returns if you need an analytics nudge.

does that make sense?

internal/engine/analytics_nudge.go
maiamcc: totally fair!

internal/engine/upper.go
maiamcc: it doesn't have to know, it's just a reasonable place to check for this condition being met -- but you're right that it's more robust just to have a func that checks the engine state, will do that.

PR #1612, opened 2019-05-10
analytics: respect env var to disable analytics, allow us to send metrics in the same session where someone opts in

maiamcc:
1. if $TILT_DISABLE_ANALYTICS is set, don't prompt user to opt in/out
2. create our `Analytics` _after_ user opts out/in, so that it reflects the user's choice (before, we first created the object, then had the user opt, so even if the user opted in, the `Analytics` object didn't reflect that and wouldn't report).

### DIFF
--- a/internal/cli/analytics.go
+++ b/internal/cli/analytics.go
@@ -21,61 +21,71 @@ const disableAnalyticsEnvVar = "TILT_DISABLE_ANALYTICS"
 var analyticsService analytics.Analytics
 
 func initAnalytics(rootCmd *cobra.Command) error {
-	var analyticsCmd *cobra.Command
-	var err error
+	if !isAnalyticsDisabledFromEnv() {
+		status, err := analytics.OptStatus()
+		if err != nil {
+			return err
+		}
 
-	options := []analytics.Option{}
+		if status == analytics.OptDefault {
+			err = promptUserOpt()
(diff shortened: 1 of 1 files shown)


nicks:
Unit test?

maiamcc:
are there particular parts you're concerned about? this code seems like a lot more trouble to test than it's worth b/c it involves spoofing out user input and files in specific directories...

maiamcc:
@nicks bump -- if I refactor some stuff for testability i want to make sure i hit the cases you're concerned about!

maiamcc:
...but also, are tests on this code a good investment atm? this codepath is gonna change soon, but also the way we invoke it in CI means that that env var will always be set, so that's not a super useful thing to test...?

nicks:
hmmmm...this bug seems likely to regress, because changing the order of calls can re-introduce it.

maiamcc:
What bug?

nicks:
if the codepath is going to change soon, that seems like an even stronger reason to invest in tests now? if you're not sure how to test it effectively, let's talk about that! I would probably factor out a struct with the os streams as properties

maiamcc:
Nick and I talked, if this codepath is gonna change soon it doesn't seem worth it to make this change, closing this PR.

PR #1613, opened 2019-05-12
watch: don't watch each individual file

dbentley:


jazzdan:
I know this is a draft, but FWIW this LGTM. I think the macOS failure is test flakiness.

nicks:
i think the test failure is legit? it looks like @dbentley changed TestWatchNonexistentDirectory in a way that changes the meaning of the test?

internal/watch/notify_test.go
```diff
@@ -338,13 +338,14 @@ func TestWatchNonexistentDirectory(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	file := f.JoinPath("root", "parent", "a")
+	parent := f.JoinPath("parent")
+	file := f.JoinPath("parent", "a")
 
-	f.watch(file)
+	f.watch(parent)
```
nicks: i thought this test was particularly for the case where someone is watching a Dockerfile that doesn't exist yet. I don't think this change preserves that meaning?

jazzdan: Whoops, missed that! Good catch @nicks

dbentley: I've now fixed the test and added a test that doesn't work (which matches the original test).

Watching a file within a directory where neither exist is racey on Linux (before this change!), and fixing it would be involved.

internal/watch/notify_test.go
```diff
@@ -338,15 +339,60 @@ func TestWatchNonexistentDirectory(t *testing.T) {
...
+		// for directories that were the root of an Add, we don't report creation, cf. watcher_darwin.go
+		f.assertEvents()
+	} else {
+		f.assertEvents(parent)
+	}
 	f.WriteFile(file, "hello")
-	f.assertEvents(file)
+
+	if runtime.GOOS == "darwin" {
+		// mac doesn't that return dir change as part of file creation
```
jazzdan: I think you accidentally a word in this comment?

internal/watch/notify_test.go
```diff
@@ -338,15 +339,60 @@ func TestWatchNonexistentDirectory(t *testing.T) {
...
+		f.assertEvents(file)
+	} else {
+		f.assertEvents(parent, file)
+	}
 }
 
+// doesn't work on linux
```
jazzdan: Can we make a bug for this so we can come back and uncomment it eventually?

Issue #1614, opened 2019-05-13
helm(): perhaps prevent resources under templates/tests/ to be activated


fkaleo:
If a helm chart contains test resources under templates/tests/ they probably should not be deployed by `tilt up`

nicks:
Good call!

jazzdan:
This should be fixed in https://github.com/windmilleng/tilt/commit/6269fcd766700851214b8ab099bd769ea45c4eae

Thanks for the suggestion @fkaleo!

fkaleo:
You guys are quick :) Thanks!

PR #1619, opened 2019-05-13
tiltfile: don't template tests in helm #1614

jazzdan:


internal/tiltfile/files.go
```diff
@@ -339,7 +339,22 @@ func (s *tiltfileState) helm(thread *starlark.Thread, fn *starlark.Builtin, args
 		return nil, fmt.Errorf("Argument 0 (path): %v", err)
 	}
 
-	yaml, err := s.execLocalCmdArgv("helm", "template", localPath.path)
+	templates, err := ioutil.ReadDir(filepath.Join(localPath.path, "templates"))
+	if err != nil {
+		return nil, fmt.Errorf("Error reading chart template directory at %s: %v", localPath.path, err)
```
nicks: i feel like this error message could be a lot better?

Before this change, the error message would come from Helm and look like "no Chart.yaml exists in directory "/home/nick/src/tilt.build/src"", but not you're short-circuiting that

internal/tiltfile/files.go
```diff
@@ -339,7 +339,22 @@ func (s *tiltfileState) helm(thread *starlark.Thread, fn *starlark.Builtin, args
...
+	for _, t := range templates {
+		name := t.Name()
+		if name == "tests" || name == "NOTES.txt" || filepath.Ext(name) == ".tpl" {
+			continue
+		}
+		cmd = append(cmd, "-x")
+		cmd = append(cmd, filepath.Join("templates", name))
+	}
+
+	yaml, err := s.execLocalCmdArgv(cmd...)
```
nicks: should we also adjust the watch to ignore the test files?

internal/tiltfile/files.go
```diff
@@ -339,7 +339,22 @@ func (s *tiltfileState) helm(thread *starlark.Thread, fn *starlark.Builtin, args
...
-	yaml, err := s.execLocalCmdArgv("helm", "template", localPath.path)
+	templates, err := ioutil.ReadDir(filepath.Join(localPath.path, "templates"))
+	if err != nil {
+		return nil, fmt.Errorf("Error reading chart template directory at %s: %v", localPath.path, err)
+	}
+
+	cmd := []string{"helm", "template", localPath.path}
+	for _, t := range templates {
+		name := t.Name()
+		if name == "tests" || name == "NOTES.txt" || filepath.Ext(name) == ".tpl" {
```
nicks: this seems kind of brittle. how confident are you that this is correct?

nicks: poking at the code a bit, i think the actual rule is "ignore NOTES.txt and files that start with underscore"

https://github.com/helm/helm/blob/master/cmd/helm/template.go#L229

i wonder if we should also file a bug report upstream that `helm template` should skip tests by default

jazzdan: Great find! I need to get better at just looking at what open source code does in cases like this.

internal/tiltfile/files.go
jazzdan: On second thought, there's no good way to watch a directory with exclusions and we definitely want to watch the directory to catch the case where someone adds a new template to a template directory.

internal/tiltfile/tiltfile_test.go
```diff
@@ -3136,6 +3136,26 @@ update_mode(UPDATE_MODE_MANUAL)
 	f.loadErrString("update_mode can only be called once")
 }
 
+func TestHelmSkipsTests(t *testing.T) {
```
nicks: can you also add a test that asserts an error message if helm() is pointing to a bad directory?

internal/tiltfile/files.go
nicks: ok

PR #1623, opened 2019-05-13
web: use runtime instead of the build loader to load assets

nicks:
Hello @jazzdan,

Please review the following commits I made in branch nicks/runtime:

a82182c5a6411c604de6d419a66cfdc20e941ff3 (2019-05-13 17:51:05 -0400)
web: use runtime instead of the build loader to load assets

### DIFF
--- a/web/web.go
+++ b/web/web.go
@@ -20,7 +20,7 @@ func StaticPath() (string, error) {
 		if os.IsNotExist(err) {
 			return "", fmt.Errorf("Could not find Tilt web static files at path: %s", dir)
 		}
-		return "", fmt.Errorf("Coult not find Tilt web static files: %v", err)
+		return "", fmt.Errorf("Could not find Tilt web static files: %v", err)
 	}
 
 	return dir, nil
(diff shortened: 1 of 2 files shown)


nicks:
for background: this came up because i was seeing weird issues where the Go build loader was fetching modules when I ran Tilt, which we definitely don't want.

This means that Tilt dev will only work on the machine it was compiled on, which is honestly probably the behavior we want anyway.

web/web.go
```diff
@@ -0,0 +1,27 @@
...
+	}
+
+	// Double-check that the directory exists on disk and at least contains package.json
+	dir := filepath.Dir(file)
+	_, err := os.Stat(filepath.Join(dir, "package.json"))
+	if err != nil {
+		if os.IsNotExist(err) {
+			return "", fmt.Errorf("Could not find Tilt web static files at path: %s", dir)
+		}
+		return "", fmt.Errorf("Coult not find Tilt web static files: %v", err)
```
jazzdan:
```suggestion
		return "", fmt.Errorf("Could not find Tilt web static files: %v", err)
```

PR #1624, opened 2019-05-13
web: container restarts are warnings

jazzdan:
Screen Shot 2019-05-13 at 5 58 55 PM

This is part 1 of 2 for this work. The next PR will refactor the implementation of errors so that pod restarts aren't considered "errors", but still show up in the errors tab which @yuindustries is soon going to rename to something else.


web/src/testdata.test.tsx
```diff
@@ -68,12 +68,14 @@ function oneResource(): any {
       Edits: ["main.go"],
       StartTime: ts,
     },
-    PodName: "vigoda-pod",
-    PodCreationTime: ts,
-    PodStatus: "Running",
-    PodRestarts: 1,
-    Endpoints: ["1.2.3.4:8080"],
-    PodLog: "1\n2\n3\n4\nabe vigoda is now dead\n5\n6\n7\n8\n",
+    ResourceInfo: {
```
jazzdan: All of this testdata was wrong: pod specific information is always nested under `ResourceInfo`.

PR #1626, opened 2019-05-14
version check: strip leading 'v' from github versions to be consistent

landism:
prior to this:
the "current" build had no leading "v", e.g. "0.8.3"
the "latest" build fetched from github had a leading "v", e.g., "v0.8.3"

1. strip v from github versions, so that we're all the same
2. document that in the TiltBuild class
3. no unit test since implementing a fake github api doesn't seem like a good use of time right now. unless someone has a cheaper idea?

### DIFF
--- a/internal/model/tilt_build.go
+++ b/internal/model/tilt_build.go
@@ -6,6 +6,7 @@ import (
 
 // Information on the current built of the Tilt binary
 type TiltBuild struct {
+	// Version w/o leading "v"
 	Version string
 	Date    string
 	Dev     bool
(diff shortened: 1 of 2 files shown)


internal/model/tilt_build.go
```diff
@@ -6,6 +6,7 @@ import (
 
 // Information on the current built of the Tilt binary
```
maiamcc: luls "current build"

landism:
```suggestion
// Information on a build of the Tilt binary
```

PR #1627, opened 2019-05-14
web/hud: analytics nudge buttons send request to hud server [ch2532]

maiamcc:
Will actually record user opt in a separate PR

### DIFF
--- a/internal/hud/server/server.go
+++ b/internal/hud/server/server.go
@@ -18,7 +18,7 @@ import (
 	"github.com/windmilleng/wmclient/pkg/analytics"
 )
 
-type analyticsStatPayload struct {
+type analyticsPayload struct {
 	Verb string            `json:"verb"`
 	Name string            `json:"name"`
 	Tags map[string]string `json:"tags"`
@@ -46,8 +46,8 @@ func ProvideHeadsUpServer(store *store.Store, assetServer assets.Server, analyti
 	}
 
 	r.HandleFunc("/api/view", s.ViewJSON)
-	r.HandleFunc("/api/analytics/opt", s.HandleAnalyticsOpt)
-	r.HandleFunc("/api/analytics/stat", s.HandleAnalyticsStat)
+	r.HandleFunc("/api/analytics", s.HandleAnalytics)
+	r.HandleFunc("/api/analytics_opt", s.HandleAnalyticsOpt)
 	r.HandleFunc("/api/sail", s.HandleSail)
 	r.HandleFunc("/api/control/reset_restarts", s.HandleResetRestarts)
 	r.HandleFunc("/ws/view", s.ViewWebsocket)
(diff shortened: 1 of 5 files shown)


web/src/AnalyticsNudge.tsx
```diff
@@ -2,6 +2,19 @@ import React, { Component } from "react"
 import "./AnalyticsNudge.scss"
 
 class AnalyticsNudge extends Component {
+  static analyticsOpt(optIn: boolean) {
```
jazzdan: Oooh static, fancy!

web/src/AnalyticsNudge.tsx
```diff
@@ -2,6 +2,19 @@ import React, { Component } from "react"
...
+    let url = `//${window.location.host}/api/analytics/opt`
```
jazzdan: Should this use the `PathBuilder`?

web/src/AnalyticsNudge.tsx
```diff
@@ -2,6 +2,19 @@ import React, { Component } from "react"
...
+
+    let payload = { opt: optIn ? "opt-in" : "opt-out" }
+
+    let j = JSON.stringify(payload)
+    console.log(j)
```
jazzdan: :eyes:

web/src/AnalyticsNudge.tsx
```diff
@@ -14,7 +27,12 @@ class AnalyticsNudge extends Component {
           .)
         </div>
         <div className="AnalyticsNudge-buttons">
-          <button>Yes!</button> <button>Nope!</button>
+          <button onClick={() => AnalyticsNudge.analyticsOpt(true)}>
+            Yes!
+          </button>{" "}
```
jazzdan: Hmm, I'm not sure this whitespace is the way to do what you want, but I defer to @yuindustries.

internal/hud/server/server.go
```diff
@@ -42,7 +46,8 @@ func ProvideHeadsUpServer(store *store.Store, assetServer assets.Server, analyti
 	}
 
 	r.HandleFunc("/api/view", s.ViewJSON)
-	r.HandleFunc("/api/analytics", s.HandleAnalytics)
+	r.HandleFunc("/api/analytics/opt", s.HandleAnalyticsOpt)
+	r.HandleFunc("/api/analytics/stat", s.HandleAnalyticsStat)
```
jazzdan: What does the "stat" in this URL mean? Statistics?

web/src/AnalyticsNudge.tsx
maiamcc: ...you saw nothing.

web/src/AnalyticsNudge.tsx
maiamcc: I think Prettier did this without my consent :-/

web/src/AnalyticsNudge.tsx
maiamcc: ...is it? mostly Prettier just yelled at me until I did it.

jazzdan: Interesting, do you mean TypeScript? I've never seen that before. It's technically correct since `analyticsOpt` isn't doing anything with the instance of the class. I just didn't even realize that TypeScript had static!

web/src/AnalyticsNudge.tsx
jazzdan: Oh weird, I trust it then.

web/src/AnalyticsNudge.tsx
maiamcc: i mean, i don't think so, since it is talking directly to whatever is hosting the page, no need to do weird relativism stuff here -- but i think more importantly, the path builder was relevant for the Sail server since we had to make paths relevant to something other than we had been, but this nudge should never show up on a page served by sail. i'll file a ticket.

PR #1628, opened 2019-05-14
cli: don't ask for analytics opt on startup if newAnalyitcs flag is set

maiamcc:
This will make it easier for me to develop this feature,
though if this seems too risky I can omit.

### DIFF
--- a/internal/cli/analytics.go
+++ b/internal/cli/analytics.go
@@ -12,6 +12,7 @@ import (
 
 	"github.com/spf13/cobra"
 	giturls "github.com/whilp/git-urls"
+	"github.com/windmilleng/tilt/internal/hud/webview"
 	"github.com/windmilleng/wmclient/pkg/analytics"
 )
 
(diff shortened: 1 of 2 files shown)


internal/cli/analytics.go
```diff
@@ -39,40 +40,42 @@ func initAnalytics(rootCmd *cobra.Command) error {
...
-		_, err := fmt.Fprintf(os.Stderr, "Send anonymized usage data to Windmill [y/n]? ")
-		if err != nil {
-			return err
-		}
-
-		buf := bufio.NewReader(os.Stdin)
-		c, _, _ := buf.ReadRune()
-		if c == rune(0) || c == '\n' || c == 'y' || c == 'Y' {
-			err = analytics.SetOpt(analytics.OptIn)
+	if !webview.NewAnalyticsOn() {
```
jazzdan: Maybe we move this up in to its own if statement so that we don't have to re-indent the rest of the code?

```suggestion
	if webview.NewAnalyticsOn() {
	  return nil
	}
```

maiamcc: Yes! I thought this would leave `analyticsService` unset but I forgot that my other PR got abandoned and that `analyticsService` still gets initialized before we do any of the `OptStatus` checks. Good catch!

PR #1643, opened 2019-05-16
web: show timeago for all alert types

jazzdan:


landism:
do we want to use the ["<30s" etc formatter](https://github.com/windmilleng/tilt/blob/4e921aab8b06936e642306f2aa5d2c1f8e9a3ce4/web/src/Sidebar.tsx#L52) for these?

jazzdan:
Oh yes, good call @landism!

web/src/ErrorPane.tsx
```diff
@@ -100,7 +102,9 @@ class AlertPane extends PureComponent<AlertsProps> {
             <header>
               <p>{r.name}</p>
               <p>{`Restarts: ${r.resourceInfo.podRestarts}`}</p>
-              <p>{r.resourceInfo.podCreationTime}</p>
+              <p>
+                <TimeAgo date={r.resourceInfo.podCreationTime} />
```
jazzdan: Using podCreationTime for restarts is really wrong... like it it could say that the restart is 5 minutes old because that's when the pod was created. Let me think about this.

jazzdan: I think we should just omit times for restarts. We don't really have a way of getting them right now.

There are a couple ways we could solve this:
1) In the short term maybe we can grab the _container_ start time from the Kubernetes API?
2) Maybe streaming all the events would let us get this info too.

PR #1644, opened 2019-05-16
web: show number of alerts next to each resource

jazzdan:
Screen Shot 2019-05-17 at 11 39 11 AM

### DIFF
--- a/web/src/Sidebar.scss
+++ b/web/src/Sidebar.scss
@@ -7,7 +7,7 @@
   bottom: $statusbar-height;
   width: $sidebar-width;
   background-color: $color-gray;
-  border-left: 2px solid $color-gray-lightest;
+  border-left: 1px solid rgba($color-white, $translucent-ish);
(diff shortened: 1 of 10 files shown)


web/src/Sidebar.scss
```diff
@@ -53,12 +53,12 @@
   background-color: $color-white;
   color: $color-gray;
 }
-.resLink.has-warnings::after {
-  content: "⚠";
-  margin-top: -5px;
-  margin-left: $spacing-unit / 2;
-  font-size: 30px;
-  color: $color-yellow;
+.resLink.has-warnings::before {
```
maiamcc: nbd for this PR but: do we want to distinguish "warnings" and "alerts" at all?

jazzdan: Yeah the errors/warnings/alerts taxonomy needs to be more fleshed out. cc @yuindustries

PR #1645, opened 2019-05-16
analytics: runtime opt-in/out; record anon stats

landism:
1. upgrade wmclient to pick up [IncrAnonymous](https://github.com/windmilleng/wmclient/pull/27)
2. wrap analytics with TiltAnalytics, which lets us:
  1. manipulate opt-in/out state at runtime
  2. restrict ourselves to anonymous metrics until the user has opted in
3. record an anonymous "hey someone ran tilt without having opted in/out" stat
4. add OptIn/OptOut methods to allow us to implement optin/out in the analytics nudge when we get to that point
  1. also record anonymously when a user opts in/out
  2. we can't actually call these until we change the type we inject from `Analytics` to `TiltAnalytics`, but that should be pretty straightforward


internal/analytics/tilt_analytics.go
```diff
@@ -0,0 +1,72 @@
...
+}
+
+func (ta *TiltAnalytics) IncrIfUnopted(name string, tags map[string]string) {
+	if ta.opt == analytics.OptDefault {
+		ta.a.IncrAnonymous(name, tags)
+	}
+}
+
+func (ta *TiltAnalytics) IncrAnonymous(name string, tags map[string]string) {
+	if ta.opt == analytics.OptIn {
```
maiamcc: wait, I don't think you want this check here 🤔

internal/analytics/tilt_analytics_test.go
```diff
@@ -0,0 +1,223 @@
...
+	initialOpt   analytics.Opt
+	newOpt       analytics.Opt
+	expectRecord bool
+}
+
+func TestOptTransitionIncr(t *testing.T) {
+	for _, test := range []transitionTestCase{
+		{"default -> out", analytics.OptDefault, analytics.OptOut, false},
+		{"default -> in", analytics.OptDefault, analytics.OptIn, true},
+		{"in -> out", analytics.OptIn, analytics.OptOut, false},
```
maiamcc: I don't think we'll ever have to deal with an `in -> out` or `out -> in` transition with the current code -- we only ever surface the nudge if the user isn't opted. tho maybe still worth leaving these cases in idk?

internal/analytics/tilt_analytics_test.go
```diff
@@ -0,0 +1,223 @@
...
+	return nil
+}
+
+func TestCount(t *testing.T) {
+	for _, test := range testCases(true, false, false) {
+		t.Run(test.name, func(t *testing.T) {
+			ma := analytics.NewMemoryAnalytics()
+			os := &optSetting{}
+			a := NewTiltAnalytics(test.opt, os.setOpt, ma)
+			a.Count("foo", map[string]string{}, 1)
```
maiamcc: why not verify that the tags get sent along, too? (and in incr/timer cases as well)

internal/analytics/tilt_analytics_test.go
```diff
@@ -0,0 +1,223 @@
...
+		if err != nil {
+			assert.FailNow(t, err.Error())
+		}
+	}
+
+	if !assert.Equal(t, newOpt, os.opt) {
+		t.FailNow()
+	}
+
+	// wipe out the reports of opt-in/out, since those are just setup for another test
```
maiamcc: ...huh?

internal/cli/analytics.go
```diff
@@ -34,27 +36,33 @@ func initAnalytics(rootCmd *cobra.Command) error {
 	var err error
 
 	options := []analytics.Option{}
-	options = append(options, analytics.WithGlobalTags(globalTags()))
+	options = append(options, analytics.WithGlobalTags(globalTags()), analytics.WithEnabled(true))
```
maiamcc: this looks like we're enabling analytics by default, needs a comment

internal/analytics/tilt_analytics.go
```diff
@@ -0,0 +1,71 @@
...
+	}
+}
+
+func (ta *TiltAnalytics) Incr(name string, tags map[string]string) {
+	if ta.opt == analytics.OptIn {
+		ta.a.Incr(name, tags)
+	}
+}
+
+func (ta *TiltAnalytics) IncrIfUnopted(name string, tags map[string]string) {
```
maiamcc: arguably people can do even less weird stuff w this func if it doesn't accept tags. do we even want to let ppl pass tags here?

internal/analytics/tilt_analytics_test.go
maiamcc: ah, i think you probs want: 
```suggestion
	// wipe out the reports of opt-in/out, since we're testing those separately
```

internal/cli/analytics.go
```diff
@@ -21,7 +23,7 @@ const tiltAppName = "tilt"
 const disableAnalyticsEnvVar = "TILT_DISABLE_ANALYTICS"
 const analyticsURLEnvVar = "TILT_ANALYTICS_URL"
 
-var analyticsService analytics.Analytics
+var analyticsService *tiltanalytics.TiltAnalytics
```
maiamcc: ughhhhh changing this file, and it being untested, makes me antsy. i suppose we'll have to do it soon?? but like, i'm a little inclined to do it when we pull the trigger on newAnalytics so that we don't need to worry about maintaining the old flow and the new flow side by side? 🤷‍♀ 😬

internal/analytics/tilt_analytics_test.go
landism: Yeah, eh, I feel like that's a product-level question that might change at some point, and kind of outside the scope of what this type should know about. If the tests were more expensive (in terms of implementation cost or LOC or execution time), I might worry about it, but I'm inclined to leave them in.

internal/analytics/tilt_analytics.go
landism: Maybe? It was intentional. I could go either way, and I think it's weird either way. I was going to explain my thinking here, but added it as a comment (the best way to paper over poor design) instead. That isn't a doubling down on this choice, and I'm open to being convinced otherwise - I've just seen (and written!) a number of explanations in PR comments that would have been better in the code itself.

maiamcc:
...tho maybe wait till mon to merge juuuust in case? 😬

PR #1646, opened 2019-05-16
k8s: fix regression in namespace-restricted service accounts

nicks:
Hello @jazzdan,

Please review the following commits I made in branch nicks/runtime:

e64451994248ac980ab5a6b598ecdb3dbe902a0b (2019-05-16 17:51:56 -0400)
k8s: fix regression in namespace-restricted service accounts


nicks:
verified this manually against a bespoke cluster i set up with a service account. my bad for not writing a good integration test for this in the first place

PR #1647, opened 2019-05-16
engine: fix service watching

landism:
I'm not entirely sure why this timeout was there, but this code was pretty broken:

The [`defer cancel()`](https://github.com/windmilleng/tilt/blob/e65742e96b9a7b0684a2e75547cc8089f3ac76ba/internal/engine/service_watch.go#L52) here meant we were canceling the watch's context immediately after starting `dispatchServiceChangesLoop`, so it'd exit pretty much immediately.

If we simply remove the `defer cancel()`, then we're still passing an expiring context into `WatchServices`, and `dispatchServiceChangesLoop` exits after `watchTimeout` (10 seconds).

My best guess is that at some point someone thought `WatchServices` was potentially blocking, and added the timeout to ensure we weren't waiting for it forever. Looking at it now, though, I don't see any code in there that would even care if `ctx` expired anyway, so I think it's perfectly safe to remove (well, from the perspective of `ServiceWatcher`, anyway. It's possible that fixing `ServiceWatcher` breaks something else that was only working because `ServiceWatcher` was broken).


landism:
(FWIW, AFAIK no one has actually reported a problem with this - I just happened to run into it when I was copy/pasting code from this for event watching yesterday)

PR #1649, opened 2019-05-16
web: when user opts in/out via nudge, acknowledge it, and vanish the notification [ch2590]

maiamcc:

Please review the following commits I made in branch maiamcc/acknowledge-opt:

d79f671082f9b2c739d80b725b368dcb63f9daec (2019-05-16 18:37:45 -0400)
add enzyme-to-json, test analyticsNudge

7d5497fecac9d32896fd75ef154d0ebcc21659a6 (2019-05-16 17:54:32 -0400)
web: store state on analyticsNudge, reflect endpoint result, disappear after 3s

### DIFF
--- a/web/src/AnalyticsNudge.tsx
+++ b/web/src/AnalyticsNudge.tsx
@@ -1,6 +1,5 @@
 import React, { Component } from "react"
 import "./AnalyticsNudge.scss"
-import { ResourceView } from "./types"
 
 type AnalyticsNudgeProps = {
   needsNudge: boolean
@@ -49,13 +48,24 @@ class AnalyticsNudge extends Component<
   messageElem(): JSX.Element {
     if (this.state && this.state.responseCode) {
       if (this.state.responseCode == 200) {
-        return <span>Okay yay!</span>
+        // Successfully called opt endpt.
+        return <span>Cool, got it! 👍</span>
       } else {
-        return <span>Oh no, something went wrong! hi@windmill.engineering</span>
+        return (
+          // error calling the opt endpt.
+          <span>
+            Oh no, something went wrong!{" "}
+            <a href="https://tilt.dev/contact" target="_blank">
+              Get in touch
+            </a>
+            .
+          </span>
+        )
       }
     }
 
     if (this.state && this.state.requestMade) {
+      // request in progress
       return <span>Okay, we'll inform the robots...</span>
     }
     return (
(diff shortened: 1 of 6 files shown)


maiamcc:
uhh I might have messed something up installing the node modules 😅

maiamcc:
Indeed, I did something goofy with node modules. Should be fixed now, thanks @jazzdan!

web/src/AnalyticsNudge.tsx
```diff
@@ -1,33 +1,75 @@
...
+// constructor for defaults?
+
+class AnalyticsNudge extends Component<
+  AnalyticsNudgeProps,
+  AnalyticsNudgeState
+> {
+  shouldShow(): boolean {
+    return (
+      this.props.needsNudge ||
+      (this.state && this.state.requestMade && !this.state.dismissed)
```
jazzdan: Let's set these to sane defaults in a constructor so we don't have to do these null checks.

web/src/AnalyticsNudge.tsx
```diff
@@ -1,33 +1,75 @@
...
-      console.log("got response -->", response.status) // returns 200
+    }).then((response: Response) => {
+      this.setState({
+        responseCode: response.status,
+      })
+
+      // after 3s, dismiss the nudge
+      setTimeout(() => {
+        this.setState({ dismissed: true })
+      }, 3000)
```
jazzdan: give this number a name, like `const 3Seconds = 3000`?

web/src/AnalyticsNudge.tsx
```diff
@@ -1,33 +1,75 @@
...
     })
   }
-  render() {
-    let classes = ["AnalyticsNudge"]
-    if (this.props.needsNudge) {
-      // or if  already visible...
-      classes.push("is-visible")
+
+  messageElem(): JSX.Element {
+    if (this.state && this.state.responseCode) {
```
jazzdan: We won't need these acrobatics if we set a default in the constructor.

web/src/AnalyticsNudge.tsx
```diff
@@ -1,33 +1,75 @@
...
+            <a href="https://tilt.dev/contact" target="_blank">
+              Get in touch
+            </a>
+            .
+          </span>
+        )
+      }
     }
 
+    if (this.state && this.state.requestMade) {
```
jazzdan: Same here

maiamcc: lol way ahead of you bud

PR #1654, opened 2019-05-17
web: more specific info when showing nudge - whether user opted in/out, http response body if error; finalize copy [ch2553]

maiamcc:
Hello @jazzdan, @landism,

final final copy tbd (db is working on it) but wanted to get this out for review

### DIFF
--- a/web/src/AnalyticsNudge.test.tsx
+++ b/web/src/AnalyticsNudge.test.tsx
@@ -22,18 +22,30 @@ it("shows request-in-progress message", () => {
   expect(component).toMatchSnapshot()
 })
 
-it("shows success message", () => {
+it("shows success message: opt out", () => {
   const component = shallow(<AnalyticsNudge needsNudge={false} />)
 
-  component.setState({ requestMade: true, responseCode: 200 })
+  component.setState({ optIn: false, requestMade: true, responseCode: 200 })
 
   expect(component).toMatchSnapshot()
 })
 
-it("shows failure message", () => {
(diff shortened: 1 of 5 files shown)


web/src/AnalyticsNudge.test.tsx
```diff
@@ -22,18 +22,30 @@ it("shows request-in-progress message", () => {
...
+
+  expect(component).toMatchSnapshot()
+})
+
+it("shows failure message with response body", () => {
+  const component = shallow(<AnalyticsNudge needsNudge={false} />)
+
+  component.setState({
+    requestMade: true,
+    responseCode: 418,
```
jazzdan: Is this [a reference](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/418)? :P

web/src/AnalyticsNudge.tsx
```diff
@@ -2,14 +2,49 @@ import React, { Component } from "react"
 import "./AnalyticsNudge.scss"
 
 const nudgeTimeoutMs = 3000 // 3 seconds
+let nudgeElem = (): JSX.Element => {
```
jazzdan: This could be a `const` too, FWIW.

maiamcc: 🤯

web/src/AnalyticsNudge.test.tsx
maiamcc: but of course!

PR #1670, opened 2019-05-21
web: use build error formatting for other errors too

landism:
The styling for errors in ErrorPane.scss applies to `<code>` elements.
For most of the types of error items we can show in ErrorPanes, we put them in `<p>` elements and *not* `<code>` elements!
It turns out that `<code>` elements are generated by the `AnsiLine` splitting we do for build errors.
This means we get all kinds of fun things, like if a pod crashes with two log lines, we display them both and ignore the newline, so they're concatenated horizontally.

Since build errors have been styled (i.e., we actually define styles for them in the scss) and the other types haven't, I'm assuming we should make everything else match build error styling.

Note this also means I'm dropping the "Last log line" copy, because:
1. we don't have it for build errors, and it's fine
2. it's a lie - we don't truncate to the last line
3. I tried changing it to "Log tail" but it felt weird to have it there with the same styling as the log itself. Given that its absence seemed fine for build errors, I figured it wasn't worth keeping.

image

### DIFF
--- a/web/src/__snapshots__/ErrorPane.test.tsx.snap
+++ b/web/src/__snapshots__/ErrorPane.test.tsx.snap
@@ -1,5 +1,44 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
+exports[`renders multiple lines of a crash log 1`] = `
+<section
+  className="ErrorPane"
+>
+  <ul>
+    <li
+      className="ErrorPane-item"
+    >
+      <header>
+        <p>
+          foo
+        </p>
+        <p>
+          Pod crashed!
+        </p>
+      </header>
+      <section>
+        <code>
+          <span
+            className={null}
+            style={null}
+          >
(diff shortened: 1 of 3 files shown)


web/src/__snapshots__/ErrorPane.test.tsx.snap
```diff
@@ -83,7 +122,14 @@ exports[`renders one container start error 1`] = `
         </p>
       </header>
       <section>
-        Eeeeek there is a problem
+        <code>
+          <span
```
maiamcc: ...I'm confused by the `className={null}` etc. 🤔

landism: Me too!

but the existing snapshots for build errors had them, so I chose not to worry about it.

landism:
> lgtm, tho may I see a screenshot?

added to description, though really it's just exactly how build errors look already

PR #1672, opened 2019-05-21
ui: use crashlog for crashes

landism:
1. We were only populating `CrashLog` in the crash rebuild case. We now also populate it when the restart count increases.
2. We were not using `CrashLog` at all in the web UI, so the web UI was just showing post-restart logs, and it's pretty frustrating for an alert to pull your attention to that tab just to show you logs that are not actually relevant.

on web:
image

in TUI:
image

### DIFF
--- a/web/src/__snapshots__/ErrorPane.test.tsx.snap
+++ b/web/src/__snapshots__/ErrorPane.test.tsx.snap
@@ -39,45 +39,6 @@ exports[`renders multiple lines of a crash log 1`] = `
 </section>
 `;
 
-exports[`renders multiple lines of a crash log 2`] = `
-<section
-  className="ErrorPane"
->
-  <ul>
-    <li
-      className="ErrorPane-item"
-    >
-      <header>
-        <p>
-          foo
-        </p>
-        <p>
-          Pod crashed!
-        </p>
-      </header>
-      <section>
-        <code>
-          <span
-            className={null}
-            style={null}
-          >
-            Eeeeek the container crashed
-          </span>
-        </code>
-        <code>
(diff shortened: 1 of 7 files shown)


web/src/__snapshots__/ErrorPane.test.tsx.snap
```diff
@@ -39,6 +39,45 @@ exports[`renders multiple lines of a crash log 1`] = `
 </section>
 `;
 
+exports[`renders multiple lines of a crash log 2`] = `
```
maiamcc: isn't the snapshot above (`renders multiple lines of a crash log 1`) obsolete now?

landism:
> I still do wanna know why this wasn't implemented in the first place

Do you think it's worth waiting to hear back from someone at kubecon?
(to add to the mystery, we did already have CrashLog in the Resource TS class, and just didn't use it: https://github.com/windmilleng/tilt/blob/1f72191816c0f2c144656956d082ed19cd4f4fde/web/src/HUD.tsx#L31)

> why nick removed CrashLog that one time??

He removed PreRestartLog, not CrashLog, and I think it's because we used PreRestartLog only to ensure we could show logs through restarts, and CombinedLog replaced that functionality.

web/src/__snapshots__/ErrorPane.test.tsx.snap
landism: yessss, (kind of! there are two snapshots because a bad merge led to there being two identical tests with this name, so jest stored them with suffixes 1 and 2, which is why jest didn't delete the obsolete one)

Issue #1650, opened 2019-05-17
k8s_yaml doesn't work where `kubectl apply -f` does


ndarilek:
Trying to use Tilt to do the following:

1. Deploy Helm into a new cluster, Minikube in this instance.
2. Deploy KubeDB and its catalog.
3. Deploy a Postgres database.

I have the following Tiltfile:

```
local("helm init --wait")
local("helm upgrade --install kubedb-operator appscode/kubedb --version 0.12.0  --namespace kube-system")
local("helm upgrade --install kubedb-catalog appscode/kubedb-catalog --version 0.12.0 --namespace kube-system")
k8s_yaml("postgres.yaml")
```

And the following postgres.yaml:

```
apiVersion: kubedb.com/v1alpha1
kind: Postgres
metadata:
  name: postgres
spec:
  version: "11.2"
  storageType: Durable
  storage:
    storageClassName: "standard"
    accessModes:
    - ReadWriteOnce
    resources:
      requests:
        storage: 10Gi
  terminationPolicy: DoNotTerminate
```

Note that you have to run `tilt up` a couple times, because the first Helm installation needs to install CRDs and `--wait` hangs indefinitely.

After a few `tilt up`s, I get this error:

```
      K8s objects not included in the resources above:                                                                                                                                   
      k8sPostgres-postgres                                                                                                                                                               
    ERROR:                                                                                                                                                                               
       Error: Error upserting k8s_yaml: kubectl apply:                                                                                                                                   
       stderr: Error from server (InternalError): error when creating "STDIN": Internal error occurred: failed calling webhook "postgres.mutators.kubedb.com": the server is currently   
       unable to handle the request                                                                                                                                                      
       : exit status 1                                                                                                                                                                   
```

However:

```bash
$ kubectl apply -f postgres.yaml 
postgres.kubedb.com/postgres created
```

I'm not clear on why this fails in Tilt, and there don't seem to be any additional knobs in `k8s_yaml` to tweak.

Thanks.

dbentley:
Interesting bug; thanks for reporting it.

Tilt injects labels before applying yaml so we can connect k8s resources back to Tilt reosurces. I wonder if the labels we're adding make the handler upset?

Do you use kubedb a lot? Are you an expert/implementer in it, or more of a user? (This is my first time hearing of it, so this is a pure guess)

ndarilek:
Just a user.


Is it possible to get logs of what commands are being executed? `tilt up 
[...]

PR #1674, opened 2019-05-22
k8s: log kubectl commands on --klog > 0

landism:
## Problem
In #1650, we're trying to figure out why Tilt's kubectl apply is failing when a manual kubectl apply succeeds. Unfortunately, --klog wasn't useful because it only enables logging on k8s api requests, and we do apply via the command-line.

## Solution
Enable logging of kubectl command/input/output when `--klog` is anything > 0.
It might be cleaner to put it behind another flag, or make it configurable via something other than a command-line flag, but that can be future work.


internal/k8s/logging_kubectl_runner.go
```diff
@@ -0,0 +1,80 @@
...
+
+// wraps a kubectlRunner with logging
+type loggingKubectlRunner struct {
+	logLevel logger.Level
+	runner   kubectlRunner
+}
+
+var _ kubectlRunner = loggingKubectlRunner{}
+
+func (k loggingKubectlRunner) logExecStart(ctx context.Context, args []string, stdin io.Reader) (newStdin io.Reader, err error) {
```
nicks: have you considered interface to take stdin as bytes or string instead of a Reader?

i feel like all the callers pass in a string, and it would make the implementation substantially simpler

internal/cli/up.go
```diff
@@ -168,6 +169,14 @@ func provideLogActions() store.LogActionsFlag {
 	return store.LogActionsFlag(logActionsFlag)
 }
 
+func provideKubectlLogLevel() k8s.KubectlLogLevel {
+	if klogLevel > 0 {
+		return k8s.KubectlLogLevel(logger.InfoLvl)
```
nicks: it's worth pointing out that the --klog flag is supposed to map to klog -v, see
https://github.com/windmilleng/tilt/blob/master/internal/cli/klog.go#L29
so it might make sense to map this -v in the kubectl call as well

nicks: (i.e., `tilt --klog 8` is supposed to be equivalent to `kubectl -v 8`)

internal/k8s/logging_kubectl_runner.go
landism: yep, much better!

internal/cli/up.go
landism: oh, nice!

PR #1675, opened 2019-05-22
integration: add analytics integration tests

landism:


integration/analytics/Tiltfile
```diff
@@ -0,0 +1,11 @@
+# -*- mode: Python -*-
+
+# If you get push errors, you can change the default_registry.
```
maiamcc: last time i wrote an integration test, nick told me that this didn't make sense. so i'll tell you the same thing -- even tho i fully acknowledge that it's copypasta

integration/analytics_test.go
```diff
@@ -0,0 +1,174 @@
...
+		af.t.Fatal(err)
+	}
+	af.tempDir.TearDown()
+	af.k8sFixture.TearDown()
+}
+
+type envVarValue struct {
+	name  string
+	isSet bool
+	val   string
```
maiamcc: maybe `oldVal` for extra clarity? 🤷‍♀

integration/analytics_test.go
```diff
@@ -0,0 +1,174 @@
...
+	defer f.TearDown()
+
+	f.SetOpt(analytics.OptIn)
+
+	f.TiltUp("oneup")
+
+	// ForwardPort will fail if all the pods are not ready.
+	//
+	// We can't use the normal Tilt-managed forwards here because
+	// Tilt doesn't setup forwards when --watch=false.
```
maiamcc: we're not using portforwards, i assume this comment is copypasta (here and in all cases below)

integration/analytics_test.go
```diff
@@ -0,0 +1,174 @@
...
+	f.WaitForAllPodsReady(ctx, "app=oneup")
+
+	var observedEventNames []string
+	for _, c := range f.mss.ma.Counts {
+		observedEventNames = append(observedEventNames, c.Name)
+	}
+
+	// just check that a couple metrics were successfully reported rather than asserting an exhaustive list
+	// the goal is to ensure that analytics is working in general, not to test which specific metrics are reported
+	// and we don't want to have to update this every time we change which metrics we report
```
maiamcc: +100

integration/analytics_test.go
```diff
@@ -0,0 +1,174 @@
...
+
+	// ForwardPort will fail if all the pods are not ready.
+	//
+	// We can't use the normal Tilt-managed forwards here because
+	// Tilt doesn't setup forwards when --watch=false.
+	ctx, cancel := context.WithTimeout(f.ctx, time.Minute)
+	defer cancel()
+	f.WaitForAllPodsReady(ctx, "app=oneup")
+
+	assert.Equal(t, 1, len(f.mss.ma.Counts))
```
maiamcc: all the below should be wrapped in an `if assert.Equal(t, 1, len...)` or else `f.mss.ma.Counts[0]` will throw an index error if something went wrong and it'll be confusing.

integration/analytics_test.go
```diff
@@ -0,0 +1,174 @@
...
+	c := f.mss.ma.Counts[0]
+	var tagKeys []string
+	for k := range c.Tags {
+		tagKeys = append(tagKeys, k)
+	}
+	assert.Equal(t, []string{"time"}, tagKeys)
+	assert.Equal(t, "tilt.analytics.up.optdefault", c.Name)
```
maiamcc: ehhhh it's not greeeaaaat that if we add any additional `IncrIfUnopted` calls that we have to update this integration test... can you take an approach like in `TestOptedIn` where you assert on a non-exhaustive list (say, assert that `tilt.analytics.up.optdefault` is in there, and assert that none of the stats have a `user` tag)?

integration/analytics_test.go
```diff
@@ -0,0 +1,174 @@
...
+	v := saveEnvVar(WindmillDirEnvVarName)
+	err := os.Setenv(WindmillDirEnvVarName, af.tempDir.Path())
+	if err != nil {
+		af.t.Fatal(err)
+	}
+	err = analytics.SetOpt(opt)
+	if err != nil {
+		af.t.Fatal(err)
+	}
+	err = restoreEnvVar(v)
```
maiamcc: wait a sec, if you restore the env var after you record the `choice.txt` file... how does tilt know where to look for the choice file when we `tilt up`?

landism: https://github.com/windmilleng/tilt/blob/241f7a83d70483b6178b225dbb19f39dcc872a64/integration/analytics_test.go#L33

maiamcc: 👍

integration/analytics_test.go
landism: hmm, I figured we're more sensitive to adding optdefault metrics so it's fine to raise the barrier on adding them, but yeah it just feels like busy work to force people to come edit this.

PR #1676, opened 2019-05-22
webview/engine: send stat when we surface nudge || ctx: analytics stored on context like logger is [ch2618]

maiamcc:
Hello @landism, @nicks,

Please review the following commits I made in branch maiamcc/metric-surface-nudge:

0ebdde1c93e7861fda140b44e442566a022c127a (2019-05-22 16:55:25 -0400)
move TiltAnalyticsSubscriber to engine pkg to avoid import cycles

852e8badcde71fe05e91c6cc6eb54a068d7c5c83 (2019-05-22 16:32:54 -0400)
send stat when we surface analytics nudge (only once)

c5da15ff115ec6df0f9e4f1196bebcd2940ec0cd (2019-05-22 16:32:13 -0400)
attach analytics to context


internal/hud/server/websocket.go
```diff
@@ -67,6 +67,11 @@ func (ws WebsocketSubscriber) Stream(ctx context.Context, store *store.Store) {
 func (ws WebsocketSubscriber) OnChange(ctx context.Context, s store.RStore) {
 	state := s.RLockState()
 	view := webview.StateToWebView(state)
+
+	if view.NeedsAnalyticsNudge && !state.AnalyticsNudgeSurfaced {
+		// Nudge surfaced for the first time!
```
landism: There's sorrrt of a race here since we could call this OnChange multiple times before the AnalyticsNudgeSurfacedAction gets handled.

This isn't a bug since the handler checks `!state.AnalyticsNudgeSurfaced`, but it's probably worth tidying up the language here (not necessarily the first time) and maybe adding a comment to the handle func (to help readers understand any weirdness they might encounter), and maybe even worth storing a bool on `WebsocketSubscriber` to keep things clean and only send this action once. Or...I guess there are multiple subscribers, so that doesn't help. blurgh.

internal/store/actions.go
```diff
@@ -60,3 +60,10 @@ type AnalyticsOptAction struct {
 }
 
 func (AnalyticsOptAction) Action() {}
+
+// Indicates nudge surfaced for the first time
+type AnalyticsNudgeSurfacedAction struct {
+	Opt analytics.Opt
```
landism: this looks unused?

maiamcc: ...oh weird how did that get in there?

PR #1677, opened 2019-05-22
tiltfile: deprecation warning for FastBuild

maiamcc:
Hello @landism, @jazzdan,

Please review the following commits I made in branch maiamcc/fast-build-deprecation-warn:

873ed75d9d799615774dace32223edb6041111b2 (2019-05-22 17:51:28 -0400)
tiltfile: deprecation warning for FastBuild

### DIFF
--- a/internal/tiltfile/docker.go
+++ b/internal/tiltfile/docker.go
@@ -14,6 +14,11 @@ import (
 	"github.com/windmilleng/tilt/internal/ospath"
 )
 
+const fastBuildDeprecationWarning = "FastBuild (`fast_build`; `add_fast_build`; " +
+	"`docker_build(...).add(...)`, etc.) will be deprecated soon; you can use Live " +
+	"Update instead! See https://docs.tilt.dev/live_update_tutorial.html for more " +
+	"information. If Live Update doesn't fit your use case, let us know."
+
 type dockerImage struct {
 	baseDockerfilePath localPath
 	baseDockerfile     dockerfile.Dockerfile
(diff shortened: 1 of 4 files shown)


internal/tiltfile/docker.go
```diff
@@ -604,3 +609,15 @@ func (s *tiltfileState) dockerignoresForImage(image *dockerImage) []model.Docker
...
 	return s.dockerignoresForPaths(paths)
 }
+
+func (s *tiltfileState) checkForFastBuilds(manifests []model.Manifest) {
+Loop:
+	for _, m := range manifests {
+		for _, iTarg := range m.ImageTargets {
+			if !iTarg.AnyFastBuildInfo().Empty() {
+				s.warnings = append(s.warnings, fastBuildDeprecationWarning)
+				break Loop
```
jazzdan: I find this style of breaking/continuing to labels very confusing. This could be rewritten as:

```golang
for _, m := range manifests {
                foundWarningForManifest := false
		for _, iTarg := range m.ImageTargets {
                        if foundWarningForManifest {
                                break
                        }
			if !iTarg.AnyFastBuildInfo().Empty() {
				s.warnings = append(s.warnings, fastBuildDeprecationWarning)
				foundWarningForManifest = true
                                break
```

This will work as it is, that's just my preference from previous label scar tissue (granted not in Go, though)

landism: or s/break Loop/return/?

maiamcc: ... 😅

PR #1683, opened 2019-05-24
cli: add 'tilt docker' and let the user manually enable/disable buildkit

nicks:
Hello @landism, @maiamcc,

67204d7ce08e947d1f6338ccdccc5d99792b44f8 (2019-05-24 13:32:43 +0100)
cli: add 'tilt docker' and let the user manually enable/disable buildkit


nicks:
(i'm finding that when i need to help people debug docker builds, i spend a lot of time just showing them how to run it in the right environment, so this should help with that)

PR #1687, opened 2019-05-24
upper: log 'detected a container change' message to ms.CombinedLog

landism:
## Problem
When one:
1. deploys a change to a tilt resource via live_update
2. triggers a crash in that resource

One observes:
1. "Detected a container change for docs-site. We could be running stale code. Rebuilding and deploying a new image." is logged in the "all resources" log
2. The same message is not logged in the resource-specific (e.g., `fortune`) log, so that log just shows builds happening without reason

## Solution
Also log the "detected a container change" message to the resource-specific log.

(There's probably some logging api cleanup we should do here, like having a clear hierarchy of logs (all -> resource-specific -> resource+build-specific), and writing to a lower log also writes to its ancestors, or something)

### DIFF
--- a/internal/engine/upper.go
+++ b/internal/engine/upper.go
@@ -8,16 +8,14 @@ import (
 	"strconv"
 	"time"
 
-	tiltanalytics "github.com/windmilleng/tilt/internal/analytics"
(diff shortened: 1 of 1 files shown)


nicks:
agreed that eventually we need some better log primitives

PR #1690, opened 2019-05-28
k8s: clean up serializers so that they ever print status objects

nicks:
Hello @maiamcc, @jazzdan,

Please review the following commits I made in branch nicks/poddisruption:

5b4b73e027e0d372fc4b24008671b53aa37ef647 (2019-05-28 10:44:19 -0400)
k8s: clean up serializers so that they ever print status objects
Fixes https://github.com/windmilleng/tilt/issues/1667

### DIFF
--- a/internal/k8s/serialize_test.go
+++ b/internal/k8s/serialize_test.go
@@ -25,7 +25,7 @@ func TestTracerYAML(t *testing.T) {
 		t.Errorf("Unexpected entities: %+v", entities)
 	}
 
-	result, err := SerializeYAML(entities)
+	result, err := SerializeSpecYAML(entities)
 	if err != nil {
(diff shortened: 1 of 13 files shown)


internal/k8s/serialize_test.go
```diff
@@ -91,12 +91,23 @@ func TestCRDYAML(t *testing.T) {
 		"projects.example.martin-helmich.de",
 		"example-project",
 	}, names)
+}
+
+func TestPodDisruptionBudgetYAML(t *testing.T) {
```
maiamcc: add a comment about what behavior this is supposed to test?

nicks: done!

Issue #1669, opened 2019-05-21
Tilt does not get logs of short-running pods


landism:
This was reported in #tilt by Jim P, who is using Tilt and k8s jobs to run unit tests.

repro: in Servantes, `tilt up echo-hi`
expected: tilt log shows aeirugherg
observed: tilt log shows no pod output

It looks like once Tilt gets a pod event for a manifest, it starts listening to the logs for that pod. If the pod is not in the `Running` state, Tilt [skips the pod](https://github.com/windmilleng/tilt/blob/07dfa735d24223e0207031680f40a7c9182346b5/internal/engine/podlogmanager.go#L64). In the case of a short-running job, it's possible for the job to be in the `Completed` state before Tilt sees the pod event, so Tilt never gets the logs.

PR #1693, opened 2019-05-28
engine: get logs from pods if succeeded or failed (as well as if running) so we can get logs for short-lived pods (jobs, etc.) -- addresses #1669

maiamcc:
addresses #1669

### DIFF
--- a/internal/engine/podlogmanager_test.go
+++ b/internal/engine/podlogmanager_test.go
@@ -199,6 +199,45 @@ func TestContainerPrefixes(t *testing.T) {
 	f.AssertOutputDoesNotContain(cNameNoPrefix.String())
 }
 
+func TestLogsByPodPhase(t *testing.T) {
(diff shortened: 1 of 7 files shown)


internal/engine/podlogmanager_test.go
```diff
@@ -271,6 +310,7 @@ func (f *plmFixture) AssertOutputContains(s string) {
 }
 
 func (f *plmFixture) AssertOutputDoesNotContain(s string) {
+	time.Sleep(10 * time.Millisecond)
```
jazzdan: 👀 is this necessary?

maiamcc: ya, otherwise we risk calling this before there is any output and it'll always return true, i.e. it's a useless assertion. (case in point: without this sleep, I couldn't make this test go red/green.)

PR #1695, opened 2019-05-29
integration: use restricted RBAC credentials for some tests

jazzdan:


jazzdan:
For an explanation of what I'm scripting here see [this article](https://jeremievallee.com/2018/05/28/kubernetes-rbac-namespace-user.html).

integration/k8s_fixture_test.go
```diff
@@ -168,8 +179,125 @@ func (f *k8sFixture) ClearNamespace() {
...
+	cmd.Dir = packageDir
+	err := cmd.Run()
+	if err != nil {
+		f.t.Fatalf("Error creating user: %v. Logs:\n%s", err, outWriter.String())
+	}
+}
+
+func (f *k8sFixture) getSecrets() {
+	outWriter := bytes.NewBuffer(nil)
+	cmdStr := `kubectl get secrets -n tilt-integration -o json | jq -r '.items[] | select(.metadata.name | startswith("tilt-integration-user-token-")) | .data.token'`
```
maiamcc: 😢

maiamcc: diff between these two blocks is not immediately apparent, consider comments for readability

integration/k8s_fixture_test.go
```diff
@@ -168,8 +179,125 @@ func (f *k8sFixture) ClearNamespace() {
...
+	cmd := exec.CommandContext(f.ctx, "bash", "-c", cmdStr)
+	cmd.Stderr = outWriter
+	cmd.Dir = packageDir
+	token, err := cmd.Output()
+	if err != nil {
+		f.t.Fatalf("Error getting secrets: %v. Cmd: %s", err, cmdStr)
+	}
+
+	cmdStr = `kubectl get secrets -n tilt-integration -o json | jq -r '.items[] | select(.metadata.name | startswith("tilt-integration-user-token-")) | .data["ca.crt"]'`
```
maiamcc: if this kubectl command takes longer than a few hundred ms, might make sense to run the `kubectl get secrets` bit only once and run the output of that through `jq` -- i don't want to add multiple seconds of overhead to every integration test if we can help it

integration/k8s_fixture_test.go
```diff
@@ -168,8 +179,125 @@ func (f *k8sFixture) ClearNamespace() {
...
+	}
+
+	f.token = string(token)
+	f.cert = string(cert)
+}
+
+func (f *k8sFixture) SetRestrictedCredentials() {
+	f.getSecrets()
+
+	err := ioutil.WriteFile("/tmp/cert", []byte(f.cert), 0644)
```
maiamcc: worth a note about why we have to write it to disk instead of passing it to the cmd

maiamcc: ...also it makes me a little sad that we're doing this unnecessarily for local integration tests, but 🤷‍♀

integration/k8s_fixture_test.go
jazzdan: This code is only called if you enable restricted credentials in the test, so it's opt in.

jazzdan:
@nicks I didn't, but I think that's a good idea!

integration/k8s_fixture_test.go
```diff
@@ -168,8 +179,125 @@ func (f *k8sFixture) ClearNamespace() {
...
+	if err != nil {
+		f.t.Fatalf("Error writing cert: %v", err)
+	}
+	outWriter := bytes.NewBuffer(nil)
+	cmd := exec.CommandContext(f.ctx, "kubectl", "config", "set-credentials", "tilt-integration-user", "--client-certificate=/tmp/cert", fmt.Sprintf("--token=%s", f.token))
```
nicks: modifying the user's kubeconfig seems very dangerous and likely to break something (e.g., if the user ctrl-c's a test)

have you considered setting the KUBECONFIG env var to use a custom kubeconfig?

integration/k8s_fixture_test.go
nicks: are you aware of client-certificate-data so you can embed the data in the kubeconfig file?

jazzdan: Yes, but you can't specify that via `kubectl config set-credentials` and I'm trying to avoid having to parse/write a kubeconfig.

I don't think I can avoid it though, tragically.

integration/k8s_fixture_test.go
jazzdan: Yeah I think that would be a better approach and once I get this working at all I will do it that way instead.

integration/k8s_fixture_test.go
nicks: have you tried `kubectl config set`? its instructions explicitly mention certificate-authority-data

jazzdan: Aha, no! Thank you for the tip

internal/k8s/watch.go
```diff
@@ -67,6 +68,7 @@ func (kCli K8sClient) WatchPods(ctx context.Context, ls labels.Selector) (<-chan
 		o.LabelSelector = ls.String()
 	}))
 
+	logger.Get(ctx).Infof("DERP namespace: %s", ns)
```
jazzdan: ignore this for now 😬

jazzdan:
This is still very much a work in progress, but it should be much different than the last time you looked at it @nicks @maiamcc, but I have to head out for the day.

integration/k8s_fixture_test.go
```diff
@@ -154,6 +171,9 @@ func (f *k8sFixture) ClearResource(name string) {
 func (f *k8sFixture) CreateNamespaceIfNecessary() {
 	outWriter := bytes.NewBuffer(nil)
 	cmd := exec.CommandContext(f.ctx, "kubectl", "apply", "-f", "namespace.yaml")
+	if f.usingOverriddenConfig {
```
maiamcc: instead of having to remember this every time you make a new `Command` object, consider an `f.CommandContext(...)` method that does this automatically (and maybe even passes the ctx for you). Otherwise, seems like a recipe of someone invoking a command wrong and breaking stuff in mysterious ways

integration/k8s_fixture_test.go
maiamcc: ah gotcha 👍

PR #1698, opened 2019-05-29
k8s events: fix panic from trying to index into array when context canceled

maiamcc:
Idk exactly how it worked out so, but everything in selectCases corresponds to an elem in `gvrs` (i.e. is a channel for events for a specific kind of k8s group+version+resource), except the very last case, which represents ctx close.

When we got a ctx close and also `!ok` (i don't understand why this would or wouldn't happen, but it happened at least sometimes) we tried to print an error message referencing `gvrs[chosen]`, i.e. the group+version+resource at the index of the select statement that made things weird. If the select statement that tripped it was `<- ctx.Done()` (i.e. the very last `SelectStatement`), it wouldn't correspond to something in the `gvrs` array, and we'd get a panic:

```
panic: runtime error: index out of range

goroutine 1297 [running]:
github.com/windmilleng/tilt/internal/k8s.watchEverythingLoop(0x5e75cc0, 0xc0006d29f0, 0xc000e19b00, 0xc000c04800, 0x48, 0x80, 0xc000bd4800, 0x48, 0x80)
    /Users/maia/code/go/src/github.com/windmilleng/tilt/internal/k8s/watch.go:329 +0x805
created by github.com/windmilleng/tilt/internal/k8s.K8sClient.WatchEverything
    /Users/maia/code/go/src/github.com/windmilleng/tilt/internal/k8s/watch.go:296 +0x4e8
```

### DIFF
--- a/internal/k8s/watch.go
+++ b/internal/k8s/watch.go
@@ -311,27 +311,28 @@ func watchEverythingLoop(ctx context.Context, ch chan<- watch.Event, watchers []
 
 	for {
 		chosen, value, ok := reflect.Select(selectCases)
+
 		// the last selectCase is ctx.Done
-		if chosen == len(selectCases)-1 || !ok {
-			if !ok {
-				// XXX DEBUG
-				// for some reason, we're getting ok = false for fission resources (e.g. fission.io/v1, Resource=environments)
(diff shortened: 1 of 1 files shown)


maiamcc:
Uh. Moderate pain. Left myself a note.

PR #1702, opened 2019-05-31
k8s: error if we can't watch _any_ resources

jazzdan:


internal/k8s/watch.go
```diff
@@ -294,6 +294,10 @@ func (kCli K8sClient) watchableGroupVersionResources(ctx context.Context) ([]sch
 		}
 	}
 
+	if len(ret) == 0 {
```
maiamcc: didn't we say that it made more sense to check the number of things returned in the caller of this func? though meh i don't have strong feels

internal/k8s/watch_test.go
```diff
@@ -259,6 +264,11 @@ func (tf *watchTestFixture) runServices(input []runtime.Object, expectedOutput [
 	assert.Equal(tf.t, expectedOutput, observedServices)
 }
 
+func (tf *watchTestFixture) watchEverythingExpectError(expectedErr string) {
+	_, err := tf.kCli.WatchEverything(tf.ctx, []model.LabelPair{})
```
maiamcc: I'm not sure I get why this errors -- can you add a comment about why?

internal/k8s/watch.go
jazzdan: Yeah, the problem is the thing that calls this returns a channel, and you can't exactly check the length of a channel. I could move it up one level though.

internal/k8s/watch_test.go
jazzdan: Sure!

PR #1703, opened 2019-06-03
logger: better support for logging during dependency-injection

nicks:
Hello @jazzdan, @maiamcc,

Please review the following commits I made in branch nicks/logs:

b5bf11e8d53caeb40d633b7b98ae0a8341912b7e (2019-05-31 16:51:45 -0700)
logger: better support for logging during dependency-injection


maiamcc:
I'm not sure I understand why this is necessary. What's the current behavior?

nicks:
if you log during DI, it will print to stdout, and then immediately reset the screen for the HUD.

PR #1704, opened 2019-06-03
k8s: better error reporting when tilt doesn't have permission to read nodes

nicks:
Hello @jazzdan, @maiamcc,

Please review the following commits I made in branch nicks/node-error:

46e8f237c9de53436845bf49883191f62732924d (2019-06-03 11:06:13 -0400)
k8s: better error reporting when tilt doesn't have permission to read nodes

### DIFF
--- a/internal/k8s/errors.go
+++ b/internal/k8s/errors.go
@@ -1,0 +1,18 @@
+package k8s
+
+import (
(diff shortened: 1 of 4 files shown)


jazzdan:
FYI there's a story for this: ch2660

PR #1707, opened 2019-06-03
engine/k8s: tests for k8s event code [ch2646]

maiamcc:
Hello @jazzdan, @nicks,

Please review the following commits I made in branch maiamcc/k8s-evts-test-coverage:

d490217d4023ed8b7c540454d2416ce0b8d51c9b (2019-06-03 17:23:10 -0400)
add comment

a54dddf62ef4cd12536d7b25272b8512e6c3eba0 (2019-06-03 16:44:43 -0400)
eventWatchManager and DeleteUID tests

f8a286978369f782cd1028fa4228c207daaf6b09 (2019-06-03 11:27:21 -0400)
uid map manager tests


internal/engine/event_watch_manager_test.go
```diff
@@ -0,0 +1,131 @@
...
+	"github.com/stretchr/testify/assert"
+	"github.com/windmilleng/tilt/internal/k8s"
+	"github.com/windmilleng/tilt/internal/model"
+	"github.com/windmilleng/tilt/internal/store"
+	"github.com/windmilleng/tilt/internal/testutils/output"
+	v1 "k8s.io/api/core/v1"
+)
+
+func TestEventWatchManager_dispatchesEvent(t *testing.T) {
+	f := newEWMFixture(t)
```
nicks: EWM sounds like the next music craze. Electronic Western Music?

internal/k8s/fake_client.go
```diff
@@ -120,7 +120,8 @@ func (c *FakeK8sClient) WatchEvents(ctx context.Context) (<-chan *v1.Event, erro
 }
 
 func (c *FakeK8sClient) EmitEvent(ctx context.Context, evt *v1.Event) {
-	c.eventsCh <- evt
+	// Do this async so it doesn't block if we haven't set up watch on this channel yet.
```
nicks: have you considered setting up the eventsCh on startup?

i appreciate the temptation to fix race conditions with more async code, but it often just leads to even more race conditions.

internal/engine/event_watch_manager_test.go
```diff
@@ -0,0 +1,131 @@
...
+
+func (f *ewmFixture) assertNoK8sEventActions() {
+	f.assertObservedK8sEventActions()
+}
+
+func (f *ewmFixture) assertObservedK8sEventActions(expectedActions ...store.K8SEventAction) {
+	if len(expectedActions) == 0 {
+		// assert no k8s event actions -- sleep briefly
+		// to give any actions a chance to get into the queue
+		time.Sleep(10 * time.Millisecond)
```
nicks: should this sleep be after the for loop, to catch the case where you get too many actions?

internal/engine/event_watch_manager_test.go
```diff
@@ -0,0 +1,131 @@
...
+		}
+	}
+
+	var observedActions []store.K8SEventAction
+	for _, a := range f.getActions() {
+		sca, ok := a.(store.K8SEventAction)
+		if !ok {
+			f.t.Fatalf("got non-%T: %v", store.K8SEventAction{}, a)
+		}
+		observedActions = append(observedActions, sca)
```
nicks: out of curiosity, why do you copy the actions to a new slice? i would have expected assert.Equal to work fine with f.getActions(), but maybe i'm misunderstanding

internal/engine/upper_test.go
```diff
@@ -1939,6 +1939,50 @@ func TestK8sEventLogTimestamp(t *testing.T) {
...
+		_, ok := st.ObjectsByK8SUIDs[k8s.UID(entityUID)]
+		return ok
+	})
+
+	f.kClient.EmitEverything(ls, k8swatch.Event{
+		Type:   k8swatch.Deleted,
+		Object: e.Obj,
+	})
+
+	time.Sleep(10 * time.Millisecond)
```
nicks: this should probably be a WaitUntil?

internal/k8s/fake_client.go
maiamcc: ooh ya that makes sense, will do

internal/engine/event_watch_manager_test.go
maiamcc: ha this is copypasta from matt's code, so I don't know -- but i'm actually rejiggering it a bit because i changed it to `assert.Equal` but turned out that that didn't work with ErrorActions b/c of pointer nonsense.

internal/engine/upper_test.go
maiamcc: ooh yeah you're totally right

internal/engine/event_watch_manager_test.go
maiamcc: ...wait what?

so the thinking is that if we check immediately and get no actions, that doesn't necessarily mean anything cuz maybe the actions took a few ms to get into the queue. we wait a few ms so that we can be more certain that we ACTUALLY have no actions, rather than that we just checked too fast and the actions haven't show up yet.

nicks: ya, i understand that. i'm saying that
 "assert 0 actions but have to wait a few ms to see if any more actions get in the queue" 
is a special case of
"assert N actions but have to wait a few ms to see if N + 1 actions get in the queue"
and this assertion would be simpler and better if it handled the general case.

internal/k8s/fake_client.go
maiamcc: oh! I remembered why I did it this way -- because if i create the eventsCh when we call `WatchEvts` it's easier to close it, i.e.:
```
go func() {
		<-ctx.Done()
		close(c.eventsCh)
	}()
```

I... have the sense that this is best practice but don't have a good sense of how necessary this is for tests. Thoughts?

internal/engine/event_watch_manager_test.go
maiamcc: OH! yes you're right, good call

internal/k8s/fake_client.go
nicks: shouldn't we do this in TearDown?

internal/engine/event_watch_manager_test.go
maiamcc: ...though then we wait 200ms every time we assert 0 actions, rather than waiting just 10 ms... i might bump that timeout down abit.

internal/k8s/fake_client.go
maiamcc: you mean adding a teardown func to the fake client? i'd be fine with that -- though seems kind of a pain for all the tests to remember to call it (i.e. for all the fixtures to remember to call it in their respective Teardown funcs). I'd kind of rather add a ctx to the call to `NewFakeClient`.

But that doesn't actually answer my question, which is: IS this sort of clean up in fact necessary in tests?

maiamcc: (I guess I'm asking if you have objections to me putting that async channel-closing call in `NewFakeClient` only b/c you explicitly suggested putting it in `TearDown`)

nicks: two separate points:

1) ya, you'll probably want to do this sort of cleanup in tests. it can be a pain to debug a test if there are a bunch of random goroutines from other tests hanging around

2) I worry that you're missing the high-level advice:

You added a goroutine to do teardown. 

Then you realized that this introduced a race condition, so you needed to add a new goroutine to handle the race condition. 

My high-level advice is that this is an understandable impulse, but likely to lead to even more race conditions. It might be better to consider how you could accomplish the goal with fewer goroutines

you can take or leave that advice

maiamcc: right right, that's v useful advice. Thanks!

PR #1710, opened 2019-06-04
tiltfile: tiltfile respects ignore parameter

jazzdan:


internal/tiltfile/tiltfile_test.go
```diff
@@ -3239,42 +3239,44 @@ func TestDockerbuildIgnoreAsString(t *testing.T) {
...
 	f.file("Tiltfile", `
 k8s_resource_assembly_version(2)
-docker_build('fooimage', 'foo', ignore="package.json")
+docker_build('gcr.io/foo', '.', ignore="*.txt")
 k8s_yaml('foo.yaml')
 `)
 
-	f.load()
-
+	f.load("foo")
```
maiamcc: hmm any particular reason why you're only loading "foo"? (i.e. why you're not leaving this call as `f.load()`?)

internal/tiltfile/docker.go
```diff
@@ -590,6 +591,11 @@ func (s *tiltfileState) dockerignoresForPaths(paths []string) []model.Dockerigno
 			continue
 		}
 
+		result = append(result, model.Dockerignore{
+			LocalPath: path,
+			Contents:  ignoresToDockerignoreContents(ignores),
```
maiamcc: afaict, this func can return however many dockerIgnores you want, and each elem of `ignore` is a complete `Dockerignore` in itself -- why bother to write all the `ignores` to one file?

i.e. why not:
```
for _, ignore := range ignores {
    result = append(result, model.Dockerignore{
        LocalPath: path,
        Contents: ignore,
    }
}
```

internal/tiltfile/docker.go
```diff
@@ -576,7 +577,7 @@ func (s *tiltfileState) defaultRegistry(thread *starlark.Thread, fn *starlark.Bu
 	return starlark.None, nil
 }
 
-func (s *tiltfileState) dockerignoresForPaths(paths []string) []model.Dockerignore {
+func (s *tiltfileState) dockerignoresFromPathsAndIgnores(paths []string, ignores []string) []model.Dockerignore {
```
maiamcc: this function name makes me sad, tho i can't think of a better one :(

internal/tiltfile/tiltfile_test.go
jazzdan: Nope no real reason, just copy pasta :D

internal/tiltfile/docker.go
jazzdan: No real reason, but you made me realize that we can create the contents once outside of the for loop because it's the same for every path, which might be getting at some of your unease. Let me try that.

PR #1711, opened 2019-06-05
tiltfile: add only parameter to docker_build

jazzdan:


internal/tiltfile/tiltfile_test.go
```diff
@@ -3295,6 +3295,88 @@ k8s_yaml('foo.yaml')
...
+}
+
+func TestDockerbuildOnlyAndIgnore(t *testing.T) {
+	f := newFixture(t)
+	defer f.TearDown()
+
+	f.dockerfile("Dockerfile")
+	f.yaml("foo.yaml", deployment("foo", image("gcr.io/foo")))
+	f.file("Tiltfile", `
+docker_build('gcr.io/foo', '.', ignore="**/*.md", only=["common/**", "myservice/**"])
```
jazzdan: Are there any other weird interactions I should add tests for here?

dbentley: There are probably a bunch. It would be good to say in the docs that only is built on dockerignore, and so may have issues. 

What if you put the same file in both ignore and only?

What if you have elements that start with ! in only?

What if you have \n in strings?

I'm not sure it's worth trying to catch all of these right now, vs. being clear that if you need to be extremely dlever you should talk to us.

jazzdan:
@dbentley that's true and I think that would be the job of a test in the [build/container](https://github.com/windmilleng/tilt/blob/f5a87394c4a7f07c2329204b809665570e26cd44/internal/build/container_test.go) test suite.

My instinct would be to put that in a separate PR.

internal/tiltfile/tiltfile_test.go
jazzdan: Oooh these are great. I'm going to add some tests for them, and make myself a TODO to write up some docs about precedence rules and such.

dbentley: For things that don't block this epic, but we should do, it can be better to leave it as a TODO in code than in your notebook so that if you end up working on something else, your thoughts can help the next person to take it up.

dbentley:
Fine with another PR; it seems like important testing to have some version of before closing this epic

dbentley:
The build/container wouldn't include the Tiltfile piece, right?
I'm thinking end-to-end means being able to write a test like:

f.tiltfile("docker_build('foo', '.', only=['foo'])")
f.assertInContext('foo/main.go')
f.assertNotInContext('bar/main.go')

PR #1715, opened 2019-06-06
k8s: integration tests use roles instead of cluster roles, events watcher and uid watcher respect namespace [ch2686]

maiamcc:
Hello tbd,

Please review the following commits I made in branch maiamcc/watch-permissions-for-role:

a77b82679269585a8cb6463e71e7d5c2bbf389c9 (2019-06-06 11:29:41 -0400)
k8s: integration tests use roles instead of cluster roles, events watcher and uid watcher respect namespace [ch2686]

### DIFF
--- a/internal/k8s/watch.go
+++ b/internal/k8s/watch.go
@@ -32,7 +32,10 @@ type watcher interface {
 
 type gvrWithNamespaced struct {
 	schema.GroupVersionResource
-	namespaced bool // if true, this resource can exist within a namespace; if false, cluster-wide
+
+	// if true, this resource can exist within a namespace; if false, resource is cluster-wide
+	// e.g. pods ARE namespaced; nodes are NOT namespaced
+	namespaced bool
 }
 
 func (kCli K8sClient) makeWatcher(f watcherFactory, ls labels.Selector) (watch.Interface, Namespace, error) {
@@ -64,7 +67,6 @@ func (kCli K8sClient) makeWatcher(f watcherFactory, ls labels.Selector) (watch.I
 
 		watcher, err := w.Watch(metav1.ListOptions{LabelSelector: ls.String()})
 		if err == nil {
-			fmt.Println("👌")
 			return watcher, kCli.configNamespace, nil
 		}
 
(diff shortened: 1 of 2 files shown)


internal/k8s/watch.go
```diff
@@ -317,16 +337,24 @@ func (kCli K8sClient) WatchEverything(ctx context.Context, lps []model.LabelPair
 	}
 
 	var watchers []watch.Interface
+	fmt.Printf("🔎 found %d watchable gvrs\n", len(gvrs))
 	for _, gvr := range gvrs {
+		fmt.Println("👀 making watcher for", gvr.String())
```
jazzdan: 👀

internal/k8s/watch.go
```diff
@@ -317,16 +337,24 @@ func (kCli K8sClient) WatchEverything(ctx context.Context, lps []model.LabelPair
 	}
 
 	var watchers []watch.Interface
+	fmt.Printf("🔎 found %d watchable gvrs\n", len(gvrs))
```
jazzdan: 👀

internal/k8s/watch.go
```diff
@@ -30,9 +30,19 @@ type watcher interface {
 	Watch(options metav1.ListOptions) (watch.Interface, error)
 }
 
+type gvrWithNamespaced struct {
+	schema.GroupVersionResource
+	namespaced bool // if true, this resource can exist within a namespace; if false, cluster-wide
```
jazzdan: if the comment goes above the struct member then certain tools/IDEs will pull that comment in when you hover over a reference to it.

internal/k8s/watch.go
maiamcc: that's uh. that's supposed to be there!

internal/k8s/watch.go
maiamcc: ALSO SUPPOSED TO BE THERE WHAT ARE YOU TALKING ABOUT

maiamcc:
well integration tests may not run right locally for me, but looks like i can manifest the fix locally! (left is master, right is this branch)
Screen Shot 2019-06-06 at 12 55 30 PM

Issue #1713, opened 2019-06-05
warnings and errors without any clarifying info


nicks:
We received this screenshot in the Tilt channel in k8s slack:

<img width="1676" alt="Screen Shot 2019-06-05 at 2 40 56 PM" src="https://user-images.githubusercontent.com/278641/58994003-27559680-87bd-11e9-9ed5-66146ced9916.png">

<img width="1679" alt="Screen Shot 2019-06-05 at 2 33 41 PM" src="https://user-images.githubusercontent.com/278641/58994071-54a24480-87bd-11e9-9b57-bc1b9e54e8ac.png">

There are a bunch of independent problems here, but the obvious ones that pop out at me:
- the resource names are overflowing the sidebar
- there's an error, but no actual info on what the error is
- there's a warning, but no actual info on what the warning is
- not clear what the relationship is between the Alert and the error/warning

nicks:
(trying to come up with repro cases for this, might need to be independent bugs)

nicks:
I think one of these pods does a `wget` and then exits immediately. And we don't really surface the pod termination state in the UI

nicks:
an easy way to repro this is to change the docker_build in https://github.com/windmilleng/tilt-helm-demo to use an image name that doesn't exist

nicks:
(i think you can find the actual warning in the ALL/logs pane. we're definitely not surfacing it correctly)

nicks:
ya, it looks like the warning indicator in the sidebar got removed here:
https://github.com/windmilleng/tilt/commit/67898e18d4143b384ff8f5e604b8515746cdc114#diff-0daacb7ac4bda035d3d7aaf9503b63c9R56
it looks like we tried to change it to something else, but made a mistake?

jazzdan:
<img width="356" alt="Screen Shot 2019-06-06 at 12 20 03 PM" src="https://user-images.githubusercontent.com/452453/59049197-76e7a100-8855-11e9-9d28-ca49921c1377.png">

It _does_ add this yellow indicator to the resource. I'm not super familiar with the css here, was this an intentional change @yuindustries?

nicks:
That's way too subtle. Neither of us (me or the user) even noticed that in the initial screenshot, even though it's a critical problem that would have helped them identify the issue.

yuindustries:
[...]

PR #1716, opened 2019-06-06
web: warnings count as alerts

jazzdan:
Addresses parts of #1713

Alerts now show up in the UI like this:

Screen Shot 2019-06-06 at 2 27 32 PM

### DIFF
--- a/web/src/AlertPane.tsx
+++ b/web/src/AlertPane.tsx
@@ -170,7 +170,7 @@ class AlertPane extends PureComponent<AlertsProps> {
                 <TimeAgo date={lastBuild.FinishTime} formatter={formatter} />
               </p>
             </header>
-            <section>{w}</section>
+            <section>{logToLines(w)}</section>
           </li>
         )
       })
--- a/web/src/AlertPane.tsx
+++ b/web/src/AlertPane.tsx
@@ -32,6 +32,7 @@ class AlertResource {
     }
   }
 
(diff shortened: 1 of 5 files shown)


nicks:
the formatting on that error seems all wrong? shouldn't there be newlines?

web/src/AlertPane.tsx
```diff
@@ -33,7 +33,12 @@ class AlertResource {
   }
 
   public hasAlert() {
-    return this.podStatusIsError() || this.podRestarted() || this.buildFailed()
+    return (
+      this.podStatusIsError() ||
```
nicks: i'm worried that the hasAlert() codepath and the render alerts codepath are so different, and likely to get out of sync.

have you considered having a getAlertElements() that returns what's currently in AlertPane's errorElements? then hasAlerts() just returns `getAlertElements().length > 0`, and we're back to a single source of truth

web/src/AlertPane.tsx
```diff
@@ -64,8 +69,18 @@ class AlertResource {
...
     }
 
+    num += this.warnings().length
+
     return num
   }
+
+  public warnings(): Array<string> {
+    if (this.buildHistory.length > 0) {
+      return this.buildHistory[0].Warnings ? this.buildHistory[0].Warnings : []
```
nicks: this.buildHistory[0].Warnings || []

web/src/AlertPane.tsx
jazzdan: That's a good idea, I'll add it as a TODO and get to it today.

PR #1717, opened 2019-06-06
web: refactor alerts so number of alerts and alert rendering are the same codepath

jazzdan:


web/src/status.tsx
```diff
@@ -28,6 +28,7 @@ function warnings(res: any): string[] {
   let buildHistory = res.BuildHistory || []
   let lastBuild = buildHistory[0]
   let warnings = (lastBuild && lastBuild.Warnings) || []
+  warnings = Array.from(warnings)
```
jazzdan: I was accidentally modifying this array in place 😬

PR #1718, opened 2019-06-06
engine: bring some consistency to how we do logging

nicks:
Hello @jazzdan, @maiamcc,

Please review the following commits I made in branch nicks/log:

9e7f60eb736946b50d3d8fde958fa31b082c9b29 (2019-06-06 18:46:29 -0400)
engine: bring some consistency to how we do logging

### DIFF
--- a/internal/engine/upper.go
+++ b/internal/engine/upper.go
@@ -6,6 +6,7 @@ import (
 	"io/ioutil"
 	"path/filepath"
 	"strconv"
+	"strings"
 	"time"
 
 	"github.com/davecgh/go-spew/spew"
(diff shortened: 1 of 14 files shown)


nicks:
The big idea here is that we used to have several different strategies for how we multiplexed logs:
- MultiWriter to send logs to multiple writers
- ActionWriters to create multiple actions
- Adding a reducer that copied the log to multiple places

And each of these strategies had their own sub-strategies on how to handle:
- log prefixing to figure out where the log came from
- copying byte slices to avoid mutation

I think the long-term strategy here should be to have a single logs store that has all the metadata about where a log came from, and separate out "how logs are displayed" from "how logs are stored"

The short-term strategy is simply to push all of this logic into a single reducer, handleLogAction, that sends things to both the global Log and the manifest-specific Log

internal/engine/upper.go
```diff
@@ -196,6 +200,9 @@ var UpperReducer = store.Reducer(func(ctx context.Context, state *store.EngineSt
 		handleAnalyticsNudgeSurfacedAction(ctx, state)
 	case UIDUpdateAction:
 		handleUIDUpdateAction(state, action)
+	case store.LogEvent:
+		// handled as a LogAction, do nothing
```
jazzdan: Why not use a `fallthrough` instead of what happens on L142?

nicks: hmmmm...that wouldn't work. we never want to fallthrough to a subsequent case.

PR #1724, opened 2019-06-07
web: wire TriggerMode through to web frontend

jazzdan:


jazzdan:
@maiamcc Like in Go they end up as numbers in the end.

PR #1725, opened 2019-06-07
hud server api endpoint to trigger build, if possible [ch2670]

maiamcc:

Please review the following commits I made in branch maiamcc/trigger-update-endpt:

0cbfda50385c2c21f9a0fb0d1eb5e751d6b7e533 (2019-06-07 17:24:57 -0400)
hud server api endpoint to trigger build, if possible [ch2670]

### DIFF
--- a/internal/hud/server/server.go
+++ b/internal/hud/server/server.go
@@ -31,7 +31,7 @@ type analyticsOptPayload struct {
 }
 
 type triggerPayload struct {
-	ManifestName string `json:"manifest_name"`
+	ManifestNames []string `json:"manifest_names"`
 }
 
 type HeadsUpServer struct {
@@ -162,7 +162,12 @@ func (s *HeadsUpServer) HandleTrigger(w http.ResponseWriter, req *http.Request) 
 		return
 	}
 
-	err = MaybeSendToTriggerQueue(s.store, payload.ManifestName)
+	if len(payload.ManifestNames) != 1 {
+		http.Error(w, fmt.Sprintf("/api/trigger currently supports exactly one manifest name, got %d", len(payload.ManifestNames)), http.StatusBadRequest)
+		return
+	}
+
+	err = MaybeSendToTriggerQueue(s.store, payload.ManifestNames[0])
 	if err != nil {
 		http.Error(w, err.Error(), http.StatusBadRequest)
 		return
--- a/internal/hud/server/server.go
+++ b/internal/hud/server/server.go
@@ -6,6 +6,7 @@ import (
 	"log"
 	"net/http"
 
+	"github.com/windmilleng/tilt/internal/model"
 	"github.com/windmilleng/wmclient/pkg/analytics"
 
 	"github.com/gorilla/mux"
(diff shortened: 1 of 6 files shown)


internal/hud/server/server.go
```diff
@@ -140,3 +146,44 @@ func (s *HeadsUpServer) HandleSail(w http.ResponseWriter, req *http.Request) {
 	}
 
 }
+
+func (s *HeadsUpServer) HandleTrigger(w http.ResponseWriter, req *http.Request) {
+	if req.Method != http.MethodPost {
```
jazzdan: Since the input is so simple you could also structure this API like:

```
GET /api/trigger/MANIFEST_NAME
```

And parse the URL for the manifest name. Up to you though!

internal/hud/server/server.go
```diff
@@ -140,3 +146,44 @@ func (s *HeadsUpServer) HandleSail(w http.ResponseWriter, req *http.Request) {
...
+		return
+	}
+}
+
+func MaybeSendToTriggerQueue(st store.RStore, name string) error {
+	mName := model.ManifestName(name)
+
+	state := st.RLockState()
+	m, ok := state.Manifest(mName)
+	st.RUnlockState()
```
jazzdan: Should this be `defer`ed?

internal/store/testing_store.go
```diff
@@ -63,3 +63,22 @@ func WaitForAction(t testing.TB, typ reflect.Type, getActions func() []Action) A
...
+	start := time.Now()
+	timeout := 150 * time.Millisecond
+
+	for time.Since(start) < timeout {
+		actions := getActions()
+		for _, a := range actions {
+			if reflect.TypeOf(a) == typ {
+				t.Fatalf("Found action of type %s where none was expected: %+v", typ.Name(), a)
+			} else if la, ok := a.(LogAction); ok {
+				fmt.Println(string(la.Message()))
```
jazzdan: Is this supposed to be here?

internal/hud/server/server.go
maiamcc: oh shit wait, you're making me realize that i did it this way specifically because i wanted it to be easier to trigger multiple manifests at once (e.g. if we have a button for "trigger all dirty") -- i guess by THAT logic, i should keep it as a payload and make it take a _list_ of manifest names...

internal/hud/server/server.go
maiamcc: why should it be? we don't need anything off the state after ln 176, i don't see any point to holding a lock on it for longer than we need to.

internal/store/testing_store.go
maiamcc: I believe so -- compare to [WaitForAction](https://github.com/windmilleng/tilt/blob/0cbfda50385c2c21f9a0fb0d1eb5e751d6b7e533/internal/store/testing_store.go#L57). I DID  just copy that function over so I'm not positive if this bit is necessary/necessary for my test case, but if memory serves, we need it to see what we're logging when we're not running a real reducer, cuz otherwise those log actions don't get processed.

PR #1732, opened 2019-06-10
web: Sidebar icons respect update/build control mode [ch2705]

jazzdan:
This review does a couple things:

1. Move all the CSS icons to the `SidebarIcon` React Component that returns SVGs.
2. Systematize the different runtime states we have a bit more ("error", "pending", "ok) and their corresponding colors in the sidebar icon
3. Introduce the concept of build control in to the sidebar
4. Lots of tests

In this screenshot "snack" is under manual build control, and its last build failed.

Screen Shot 2019-06-10 at 6 08 25 PM


web/src/SidebarIcon.test.tsx
```diff
@@ -0,0 +1,173 @@
+import React from "react"
+import { mount } from "enzyme"
+import SidebarIcon, { IconType } from "./SidebarIcon"
+import { RuntimeStatus, TriggerMode } from "./types"
+import { Color } from "./constants"
+
+type Ignore = boolean
+
+const cases: Array<
+  [
```
maiamcc: maybe comment what each of these fields are?

web/src/SidebarIcon.tsx
```diff
@@ -0,0 +1,159 @@
...
+  hasWarning: boolean
+  isBuilding: boolean
+}
+
+export enum IconType {
+  DotAuto = "dotAuto",
+  DotAutoPending = "dotAutoPending",
+  DotAutoBuilding = "dotAutoBuilding",
+  DotManual = "dotManual",
+  DotManualBuilding = "dotManualBuilding",
```
maiamcc: doesn't have to be in this PR, but we should probably have a `ManualPending` option (e.g. you've clicked "build" for resource X but there are other things ahead of it in the queue so it hasn't started building yet)

maiamcc: (except, wait, i don't mean "pending" as you use it below for "this resource hasn't come up yet at all"... i think there needs to be a way to indicate "waiting to build this" without scrubbing the color that shows you the current status of the resource.)

web/src/SidebarIcon.tsx
```diff
@@ -0,0 +1,159 @@
...
+        className={`${IconType.DotAutoBuilding} auto`}
+        height="10"
+        viewBox="0 0 10 10"
+        width="10"
+        xmlns="http://www.w3.org/2000/svg"
+        fill={fill}
+        style={style}
+      >
+        <path
+          d="m9.65354132 6.83246601-1.85421206-.75144721c.12960419-.33538852.20067074-.69990442.20067074-1.0810188 0-1.65685425-1.34314575-3-3-3s-3 1.34314575-3 3 1.34314575 3 3 3c.25896905 0 .51027414-.03281333.74998437-.09450909l.37116491 1.96832235c-.360418.08256902-.73568991.12618674-1.12114928.12618674-2.76142375 0-5-2.23857625-5-5s2.23857625-5 5-5 5 2.23857625 5 5c0 .64686359-.12283756 1.26503693-.34645868 1.83246601z"
```
maiamcc: ...gesundheit

web/src/status.tsx
```diff
@@ -1,11 +1,12 @@
 import { isZeroTime } from "./time"
+import { RuntimeStatus } from "./types"
 
 // A combination of runtime status and build status over a resource view.
 // 1) If there's a current or pending build, this is "pending".
 // 2) Otherwise, if there's a build error or runtime error, this is "error".
 // 3) Otherwise, we fallback to runtime status.
-function combinedStatus(res: any): string {
-  let runtimeStatus = res.RuntimeStatus
+function combinedStatus(res: any): RuntimeStatus {
```
maiamcc: ...i'm confused about whether runtime status and build status are two separate things or not. this comment indicates that this func returns some combination of build status and runtime status, but returns a `RuntimeStatus`, and i'm confused.

jazzdan: Yeah let me rename this.

web/src/SidebarIcon.tsx
```diff
@@ -0,0 +1,159 @@
...
+    if (props.isBuilding) {
+      return this.dotManualBuilding(fill)
+    }
+
+    return this.dotManualPending(fill)
+  }
+
+  dotAuto(fill: Color) {
+    return (
+      <svg
```
nicks: i'm surprised you inlined the svg? it might be easier to import the svg from a separate file, then set these properties on the component. makes it easier to view them independently

i can show you how to do this if it's unclear

yuindustries: I can take a crack at this. We sort of had a roundabout journey where we were looking at inlining SVG in a CSS `content` declaration, so we could design the icons purely in CSS. But gave up on that, but kept the inline SVG habit. I agree that we can import them as we do with other SVGs elsewhere.

PR #1735, opened 2019-06-11
engine: fix opt-in analytics event

landism:
### Problem

There are way more "opt in" events in our analytics logs than there should be, many of which do not follow an "analytics nudge" event.

`EngineState.AnalyticsOpt` is initialized by `handleInitAction`.
`TiltAnalyticsSubscriber.ta.Opt()` is initialized from the user's .windmill directory.

In theory, if a user has already opted in, TiltAnalyticsSubscriber [never sees](https://github.com/windmilleng/tilt/blob/4cc4b4fd246bd834618b5eb55cfd20f49636a14e/internal/engine/tilt_analytics.go#L22) a change in EngineState, and thus never reports an "opt in" event.

Unfortunately, `InitAction` is not the first action handled by `Store`. Anything we log before calling `Upper.Init` results in `logEvent` actions being dispatched, which get reduced before `InitAction`.

This means `TiltAnalyticsSubscriber.OnChange` is called with an uninitialized `EngineState` where `AnalyticsOpt` == `OptDefault`), and then once we handle `InitAction`, it's called again with an `EngineState` where `AnalyticsOpt` == `OptIn`, which it thinks means the user has just opted in, so it sends an "opt in" event to the analytics server.

This issue is significantly mitigated by the batching of events. I was actually unable to reproduce this bug using a master build (I ran Tilt 50 times and never generated an opt-in event). As long as the `logEvent`s and the `InitAction` are within 1ms of each other, they'll be batched together, and `TiltAnalyticsSubscriber.OnChange` will only ever see the initialized `EngineState`, so everything is fine. I was able to induce the issue by adding a sleep to `Upper.Init` to prevent batching, which at least proves there's a race here, even if we don't know whether/how often it occurs in reality.

### Solution

Only report the "opt in" event to analytics when we get an `AnalyticsOptAction`. This action is only dispatched when a user takes an action to indicate an analytics preference and isn't subject to the whims of `EngineState` initialization order.

It's possible the race doesn't actually happen in the wild and there's some other cause for this problem, but: 1) I haven't seen another way it could be happening, 2) I think this is an improvement regardless, and 3) we can check in a few days and see if "opt in" events in our logs seem more reasonable.

### DIFF
--- a/internal/engine/upper.go
+++ b/internal/engine/upper.go
@@ -12,6 +12,7 @@ import (
 	"github.com/davecgh/go-spew/spew"
 	"github.com/opentracing/opentracing-go"
 	"github.com/pkg/errors"
+	"github.com/windmilleng/tilt/internal/hud/server"
 	"github.com/windmilleng/wmclient/pkg/analytics"
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/labels"
@@ -21,7 +22,6 @@ import (
 	"github.com/windmilleng/tilt/internal/container"
 	"github.com/windmilleng/tilt/internal/dockercompose"
 	"github.com/windmilleng/tilt/internal/hud"
-	"github.com/windmilleng/tilt/internal/hud/server"
 	"github.com/windmilleng/tilt/internal/k8s"
 	"github.com/windmilleng/tilt/internal/logger"
 	"github.com/windmilleng/tilt/internal/model"
@@ -193,7 +193,7 @@ var UpperReducer = store.Reducer(func(ctx context.Context, state *store.EngineSt
 	case LatestVersionAction:
 		handleLatestVersionAction(state, action)
 	case store.AnalyticsOptAction:
-		handleAnalyticsOptAction(ctx, state, action)
+		handleAnalyticsOptAction(state, action)
 	case store.AnalyticsNudgeSurfacedAction:
 		handleAnalyticsNudgeSurfacedAction(ctx, state)
 	case UIDUpdateAction:
@@ -1005,11 +1005,7 @@ func handleTiltfileLogAction(ctx context.Context, state *store.EngineState, acti
 	state.TiltfileCombinedLog = model.AppendLog(state.TiltfileCombinedLog, action, state.LogTimestamps, nil)
 }
 
-func handleAnalyticsOptAction(ctx context.Context, state *store.EngineState, action store.AnalyticsOptAction) {
-	// only logging on opt-in, because, well, opting out means the user just told us not to report data on them!
-	if action.Opt == analytics.OptIn {
-		tiltanalytics.Get(ctx).IncrIfUnopted("analytics.opt.in")
-	}
+func handleAnalyticsOptAction(state *store.EngineState, action store.AnalyticsOptAction) {
 	state.AnalyticsOpt = action.Opt
 }
 
--- a/internal/engine/upper.go
+++ b/internal/engine/upper.go
@@ -12,7 +12,6 @@ import (
 	"github.com/davecgh/go-spew/spew"
 	"github.com/opentracing/opentracing-go"
 	"github.com/pkg/errors"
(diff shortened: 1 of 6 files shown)


internal/engine/upper_test.go
```diff
@@ -2450,32 +2450,35 @@ func TestSetAnalyticsOpt(t *testing.T) {
...
 
 	f.Start([]model.Manifest{}, true, opt)
-	f.store.Dispatch(store.AnalyticsOptAction{Opt: analytics.OptOut})
-	f.WaitUntil("opted out", func(state store.EngineState) bool {
-		return state.AnalyticsOpt == analytics.OptOut
-	})
-
-	// if we don't wait for 1 here, it's possible the state flips to out and back to in before the subscriber sees it,
-	// and we end up with no events
-	f.opter.waitUntilCount(t, 1)
```
maiamcc: is this no longer a concern?

internal/engine/upper.go
```diff
@@ -1005,7 +1005,11 @@ func handleTiltfileLogAction(ctx context.Context, state *store.EngineState, acti
 	state.TiltfileCombinedLog = model.AppendLog(state.TiltfileCombinedLog, action, state.LogTimestamps, nil)
 }
 
-func handleAnalyticsOptAction(state *store.EngineState, action store.AnalyticsOptAction) {
+func handleAnalyticsOptAction(ctx context.Context, state *store.EngineState, action store.AnalyticsOptAction) {
+	// no logging on opt-out, because, well, opting out means the user just told us not to report data on them!
```
maiamcc: nit (feel free to ignore): this comment feels mis-placed, because it's not about the clause it's on top of. consider:
```
if action.Opt...
    Incr...
}
// no logging on opt-out...
```

landism: I feel like it is even less about the line below the place you suggest, and your review comment would be even *more* applicable after that change!

I was thinking it's about the clause it's on top of since the clause could kind of similarly be `if action.Opt != analytics.OptOut` (except for OptDefault, which hopefully never appears here)

I guess I could move it above the `func` line entirely? but then a reader would expect it to describe the function as a whole.

how about changing "no logging on opt-out" to "only logging on opt-in"?

internal/engine/upper_test.go
landism: it is, but I was kinda on the fence about whether it was worth the space

added it back with better wording

internal/engine/upper.go
```diff
@@ -1005,7 +1005,11 @@ func handleTiltfileLogAction(ctx context.Context, state *store.EngineState, acti
 	state.TiltfileCombinedLog = model.AppendLog(state.TiltfileCombinedLog, action, state.LogTimestamps, nil)
 }
 
-func handleAnalyticsOptAction(state *store.EngineState, action store.AnalyticsOptAction) {
+func handleAnalyticsOptAction(ctx context.Context, state *store.EngineState, action store.AnalyticsOptAction) {
+	// only logging on opt-in, because, well, opting out means the user just told us not to report data on them!
+	if action.Opt == analytics.OptIn {
+		tiltanalytics.Get(ctx).IncrIfUnopted("analytics.opt.in")
```
nicks: it seems very sus to be sending requests from inside a reducer loop because this is holding the lock...would it be better to do this where the action gets fired?

internal/engine/upper.go
maiamcc: fair nough, I trust your judgement!

internal/engine/upper.go
landism: yeah, it's probably cleaner to do it where the action is fired.

I'm less certain about it being suspect to do this inside the reducer loop - we take pains to ensure that the analytics methods do not send requests synchronously, and I think that's a good contract to have for metrics reporting, so that we can easily add metrics without worrying about where we're doing it.

anyway, I've moved it to the http request handler since that more directly corresponds to the user action we want to record

PR #1736, opened 2019-06-11
show button to trigger update [ch2697]

yuindustries:
When in triggerMode manual, show a button that can be clicked to trigger update.

- If we get a failure response from the API, just console log it for now. In a separate PR, we should handle it more elegantly in the UI so the user knows what's happening.

- There's no indication that there are changes detected in a resource. That will be a separate PR.

## Preview

2019-06-11 - triggerMode manual button


jazzdan:
(Btw I think we have a bad merge here, let's pair on it and figure it out)

yuindustries:
Closing this one since the commit history is a mess due to merge mishaps.

PR #1738, opened 2019-06-12
show button to trigger update [ch2697]

yuindustries:
When in triggerMode manual, show a button that can be clicked to trigger update.

- If we get a failure response from the API, just console log it for now. In a separate PR, we should handle it more elegantly in the UI so the user knows what's happening.

- There's no indication that there are changes detected in a resource. That will be a separate PR.

## Preview

2019-06-11 - triggerMode manual button


yuindustries:
@jazzdan — As discussed, this is the same PR as the one from yesterday. But with a cleaner history.

web/src/SidebarTriggerButton.test.tsx
```diff
@@ -0,0 +1,16 @@
...
+it("renders trigger button in manual mode", () => {
+  const root = mount(
+    <SidebarTriggerButton
+      isSelected={true}
+      resourceName="doggos"
+      triggerMode={TriggerMode.TriggerModeManual}
+    />
+  )
+
+  expect(root.find(".SidebarTriggerButton")).toHaveLength(1)
```
jazzdan: I think this should be a snapshot test. Otherwise it's not really testing that much.

jazzdan: We could, though, test that it makes certain kinds of requests when clicked.

PR #1740, opened 2019-06-12
web: show * next to resources with pending manual mode changes

jazzdan:
Also better support for manual mode pending resources

### DIFF
--- a/web/src/HUD.tsx
+++ b/web/src/HUD.tsx
@@ -35,7 +35,7 @@ type Resource = {
   IsTiltfile: boolean
   LastDeployTime: string
   PathsWatched: Array<string>
-  PendingBuildEdits: any
+  PendingBuildEdits: Array<string>
   PendingBuildReason: number
   ResourceInfo: {
     PodCreationTime: string
@@ -49,6 +49,7 @@ type Resource = {
   RuntimeStatus: string
   ShowBuildStatus: boolean
   TriggerMode: TriggerMode
+  HasPendingChanges: boolean
 }
 
 type HudState = {
--- a/web/src/SidebarIcon.tsx
+++ b/web/src/SidebarIcon.tsx
@@ -1,5 +1,5 @@
 import React, { PureComponent } from "react"
-import { TriggerMode, ResourceStatus } from "./types"
+import { TriggerMode, ResourceStatus, Build } from "./types"
 import { Color } from "./constants"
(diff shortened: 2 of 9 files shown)


jazzdan:
Screen Shot 2019-06-12 at 12 36 21 PM

jazzdan:
I imagine we will change the styling later on, but wanted to get the correct functionality in first.

web/src/HUD.tsx
```diff
@@ -49,6 +49,7 @@ type Resource = {
   RuntimeStatus: string
   ShowBuildStatus: boolean
   TriggerMode: TriggerMode
+  HasPendingChanges: boolean
```
maiamcc: is there a reason this has to be a separate boolean rather than just `PendingBuildEdits.len > 0`?

web/src/SidebarIcon.tsx
```diff
@@ -31,6 +34,8 @@ export default class SidebarIcon extends PureComponent<SidebarIconProps> {
       fill = Color.red
     } else if (props.hasWarning) {
       fill = Color.yellow
+    } else if (props.isDirty && props.lastBuild && props.lastBuild.Error) {
```
maiamcc: comment this case

web/src/HUD.tsx
jazzdan: Yes, take this scenario:

1. Make a change to a manual resource
2. The resource errors
-- At this point `PendingBuildEdits` doesn't get emptied because a completed built did not go through.

Therefore if we relied only on `PendingBuildEdits` we wouldn't be able to reliably distinguish between "I changed a file" and "a build failed so there's still files that haven't been updated in the cluster".

Does that make sense?

maiamcc: ah yes gotcha!

maiamcc: ...though i might disambiguate the names a little in that case? I assume the difference to the user is that if `PendingBuildEdits.len > 0 && !HasPendingChanges` then we don't offer them a "play" button?

jazzdan: Yeah, thing is that `HasPendingChanges` is a term with [specific meaning in the backend](https://github.com/windmilleng/tilt/blob/16349b8ffdb6b152865db699ab902d07e9bd2a57/internal/store/engine_state.go#L387). If we change it here we should change it everywhere, and then that feels like a separate PR.

PR #1741, opened 2019-06-12
model: changes to a manifest's TriggerMode via config change actually make it into engine state [ch2703]

maiamcc:

Please review the following commits I made in branch maiamcc/set-trigger-mode-from-tiltfile:

0dce522f86a002503d75a0b8aed6cbfa51eb06ed (2019-06-12 16:07:37 -0400)
actually test this code change

b315ef613b6db748286d0aa9c926058cb440b8ae (2019-06-12 16:01:24 -0400)
clean up some existing config change tests

f32a355fd79311fa5c7b3e30e4b3d71d99e27d3e (2019-06-11 16:27:43 -0400)
check triggerMode when checking manifest equality (so we attach the new manifest to the state if we only change triggerMode)

### DIFF
--- a/internal/engine/upper.go
+++ b/internal/engine/upper.go
@@ -495,18 +495,14 @@ func handleConfigsReloaded(
 
 		configFilesThatChanged := state.LastTiltfileBuild.Edits
 		old := mt.Manifest
-		if !m.Equal(old) {
-			old := mt.Manifest
-			mt.Manifest = m
-
-			if model.ChangesInvalidateBuild(old, m) {
-				// Manifest has changed such that the current build is invalid;
-				// ensure we do an image build so that we apply the changes
-				state := mt.State
-				state.BuildStatuses = make(map[model.TargetID]*store.BuildStatus)
-				state.PendingManifestChange = time.Now()
-				state.ConfigFilesThatCausedChange = configFilesThatChanged
-			}
+		mt.Manifest = m
+		if model.ChangesInvalidateBuild(old, m) {
+			// Manifest has changed such that the current build is invalid;
+			// ensure we do an image build so that we apply the changes
+			state := mt.State
+			state.BuildStatuses = make(map[model.TargetID]*store.BuildStatus)
+			state.PendingManifestChange = time.Now()
+			state.ConfigFilesThatCausedChange = configFilesThatChanged
 		}
 		state.UpsertManifestTarget(mt)
 	}
--- a/internal/engine/upper.go
+++ b/internal/engine/upper.go
@@ -494,14 +494,19 @@ func handleConfigsReloaded(
 		newDefOrder[i] = m.ManifestName()
 
 		configFilesThatChanged := state.LastTiltfileBuild.Edits
-		if !m.Equal(mt.Manifest) {
+		old := mt.Manifest
+		if !m.Equal(old) {
+			old := mt.Manifest
 			mt.Manifest = m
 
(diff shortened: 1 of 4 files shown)


internal/engine/upper_test.go
```diff
@@ -691,101 +691,54 @@ k8s_resource('doggos', new_name='quux')  # rename "doggos" --> "quux"
 	f.assertAllBuildsConsumed()
 }
 
-func TestNoOpChangeToDockerfile(t *testing.T) {
+func TestConfigChange_NoOpChange(t *testing.T) {
```
maiamcc: while I was here I cleaned up some old configChange tests to make them
a. more in line with how we currently test stuff
b. easier to read
c. not use fast build, since we'll deprecate it eventually

internal/engine/upper.go
```diff
@@ -494,14 +494,19 @@ func handleConfigsReloaded(
 		newDefOrder[i] = m.ManifestName()
 
 		configFilesThatChanged := state.LastTiltfileBuild.Edits
-		if !m.Equal(mt.Manifest) {
+		old := mt.Manifest
+		if !m.Equal(old) {
+			old := mt.Manifest
```
jazzdan: Duplicate of L497?

internal/model/manifest.go
```diff
@@ -177,7 +177,23 @@ func (m Manifest) Validate() error {
...
 func (m1 Manifest) Equal(m2 Manifest) bool {
-	primitivesMatch := m1.Name == m2.Name
+	primitivesEq, dockerEq, k8sEq, dcEq := m1.fieldGroupsEqual(m2)
+	return primitivesEq && dockerEq && k8sEq && dcEq
+}
+
+// ChangesInvalidateBuild checks whether the changes from old => new manifest
+// invalidate our build of the old one; i.e. if we're replacing `old` with `new`,
+// should we perform a full rebuild?
+func ChangesInvalidateBuild(old, new Manifest) bool {
```
jazzdan: Is this change necessary to fix the bug in ch2703? I was under the impression that it was just the change to the manifest equality function.

internal/engine/upper.go
```diff
@@ -494,14 +494,19 @@ func handleConfigsReloaded(
...
 			mt.Manifest = m
```
landism: Moving this outside the `if` would reduce the impact of similar bugs in the future, though with the downside that the remaining impact would be harder to notice.

internal/engine/upper.go
maiamcc: lol yep

internal/model/manifest.go
maiamcc: I meannnn strictly speaking it's not necessary for just the bug fix -- however, without this code, every change to TriggerMode also causes the manifest to rebuild, which is unnecessary and so makes it a pain to change TriggerMode.

jazzdan: If a resource goes from manual to auto, shouldn't it trigger a rebuild since there might be pending changes?

Or is that handled by the build controller?

internal/engine/upper.go
maiamcc: 😬 imma do it!

internal/model/manifest.go
maiamcc: Good question! I believe it's handled by the build controller but i'll write a test

maiamcc: Test added -- good catch @jazzdan! (it IS indeed handled by the build controller.)

PR #1746, opened 2019-06-13
web: disambiguate k8s status from Tilt status

jazzdan:
This lays the groundwork for a series of changes that will disentangle the idea of "the status in kubernetes" from "where is this resource in Tilt's flow", in to two different statuses: `RuntimeStatus` (for k8s) and `ResourceStatus` (for Tilt).

In a future PR I will remove the `combinedStatus` function and call two different functions: `runtimeStatus` and `resourceStatus` as appropriate in the various places that `combinedStatus` is used.

All this PR does is:
1. Introduce `RuntimeStatus` and `ResoruceStatus`
2. Refactor tests to use the appropriate types

### DIFF
--- a/web/src/HUD.test.tsx
+++ b/web/src/HUD.test.tsx
@@ -71,7 +71,7 @@ it("doesn't re-render the sidebar when the logs change", async () => {
   let resourceView = oneResourceView()
   hud.setState({ View: resourceView })
   let oldDOMNode = root.find(".Sidebar").getDOMNode()
-  resourceView.Resources[0].PodLog += "hello world\n"
+  resourceView.Resources[0].ResourceInfo.PodLog += "hello world\n"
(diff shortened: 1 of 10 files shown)


web/src/HUD.test.tsx
```diff
@@ -71,7 +71,7 @@ it("doesn't re-render the sidebar when the logs change", async () => {
   let resourceView = oneResourceView()
   hud.setState({ View: resourceView })
   let oldDOMNode = root.find(".Sidebar").getDOMNode()
-  resourceView.Resources[0].PodLog += "hello world\n"
+  resourceView.Resources[0].ResourceInfo.PodLog += "hello world\n"
```
jazzdan: This test was actually doing nothing because `PodLog` is on the ResourceInfo, not on the Resource 😭

PR #1747, opened 2019-06-13
watch: use informer for WatchEverything

### LINKED ISSUE #1744: channels closing when watching k8s objects for events
This generates a log message like `Failed watching K8S objects of type apps/v1, Resource=deployments. This shouldn't happen. Please report a bug at: https://github.com/windmilleng/tilt/issues`, and then tilt stops watching objects of that type.

This is discussed a bit more [here](https://github.com/windmilleng/tilt/blob/2667f959005e82d69e24817b8c011ed62256b0f3/internal/k8s/watch.go#L395). It's worth noting the first line of that comment supposing it's CRD-specific is evidently wrong, as users have observed this affecting `Deployment`. This was committed as it was since no one else was able to reproduce the problem.

landism:
This hopefully fixes #1744.

I haven't figured out a 100% repro for #1744, but when running with this change and `--klog 8`, I was able to observe a bunch of logs like this, which certainly looks like the problem is reproing (i.e., channels are closing) and being handled (i.e., watches are restarted) by the informer.
```
I0613 18:25:11.948358   60899 reflector.go:370] github.com/windmilleng/tilt/internal/k8s/watch.go:377:
Watch close - *unstructured.Unstructured total 0 items received
I0613 18:25:11.948551   60899 round_trippers.go:416] GET
https://localhost:6443/apis/fission.io/v1/canaryconfigs?resourceVersion=1190535&timeoutSeconds=582&watch=t
rue
```

It's still troubling that there are so many dropped connections in the first place, but that's an issue to address separately.

bonus: the informer code is much prettier than all of that reflection

### DIFF
--- a/internal/k8s/watch.go
+++ b/internal/k8s/watch.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"fmt"
 	"net/http"
-	"reflect"
 	"time"
 
 	"github.com/pkg/errors"
@@ -13,13 +12,14 @@ import (
 	apiErrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/watch"
+	"k8s.io/client-go/dynamic/dynamicinformer"
 	"k8s.io/client-go/informers"
 	authorizationv1client "k8s.io/client-go/kubernetes/typed/authorization/v1"
 	"k8s.io/client-go/tools/cache"
 
-	"github.com/windmilleng/tilt/internal/analytics"
 	"github.com/windmilleng/tilt/internal/logger"
 	"github.com/windmilleng/tilt/internal/model"
 )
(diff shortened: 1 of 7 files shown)


landism:
Also, no tests, but there were already no tests for WatchEverything? Maybe I should spend half the day tomorrow on that.

internal/k8s/watch.go
```diff
@@ -345,80 +345,42 @@ func (kCli K8sClient) WatchEverything(ctx context.Context, lps []model.LabelPair
...
-}
-
-func watchEverythingLoop(ctx context.Context, ch chan<- watch.Event, watchers []watch.Interface, gvrs []gvrWithNamespaced) {
-	var selectCases []reflect.SelectCase
-	for _, w := range watchers {
-		selectCases = append(selectCases, reflect.SelectCase{
-			Dir:  reflect.SelectRecv,
-			Chan: reflect.ValueOf(w.ResultChan()),
+	for _, gvr := range gvrs {
+		informer := dynamicinformer.NewDynamicSharedInformerFactory(kCli.dynamic, 0).ForResource(gvr.GroupVersionResource).Informer()
```
nicks: i'm pretty sure this won't work with namespaces?

nicks:
Also, fwiw, I've found the k8s client-go testing infra to be pretty good for exercising these sorts of issues?

landism:
> Also, fwiw, I've found the k8s client-go testing infra to be pretty good for exercising these sorts of issues?

So have I! I was kinda fishing for context on whether there was a reason we skipped them in the first place.

landism:
putting this on hold until we have more time to push events over the finish line

PR #1749, opened 2019-06-14
web: correctly coerce RuntimeStatus string to proper type

jazzdan:


web/src/status.tsx
```diff
@@ -22,7 +22,17 @@ function combinedStatus(res: any): RuntimeStatus {
   if (lastBuildError) {
     return RuntimeStatus.Error
   }
-  return runtimeStatus
+
```
landism: This is a bit confusing to me... My best guess - is the intent of this change to make it so that if we add a new status and don't update this, we convert it to Unknown? Or is this where the actual coercion is happening and I don't understand TS enums?

also, ffti code golf (well, maybe slightly better than code golf since it's a bit more DRY-y, or maybe slightly worse since it's using case fallthroughs):
```
case RuntimeStatus.Error:
case RuntimeStatus.Pending:
case RuntimeStatus.Ok:
  return status
default:
  return RuntimeStatus.Unknown
```

jazzdan: Yeah, the `Resource` typo only uses primitives since we can't know the type of what we get from the server (everything on the internet is a string).

Right now that conversion happens here but we should probably have a step further up where we validate/add types to the websocket response.

Issue #1728, opened 2019-06-08
Error launching in TMUX


dotellie:
I tried to launch tilt in tmux using just `tilt up` and I got this error:

```
Error: setting up screen: Tilt does not support TERM="tmux-256color". This is not a common Terminal config. If you expect that you're using a common terminal, you might have misconfigured $TERM in your .profile.
```

Launching with `TERM="screen-256color" tilt up` meanwhile worked without problem (and I still have color).

Quite frankly, I have no idea what's going on but maybe someone can help me out here? :sweat_smile:

nicks:
Thanks for the report!

The upstream terminal library we use has a very large database of popular TERM values, and managing them all is a complex engineering problem in itself. See: https://github.com/gdamore/tcell/issues/252

They're working on fixing it, but it's a complex change. 

In the meantime, is putting `export TERM="screen-256color"` in your `.profile` a reasonable work-around?

dotellie:
Thanks for the quick answer! That's a totally reasonable solution until they get a fix in. Maybe this could be put in some sort of FAQ for other newbies like me?

nicks:
good idea! Will add it to the FAQ

jazzdan:
Fixed in 178bde683ced66effebcf327650d4838ae98f276

PR #1750, opened 2019-06-14
vendor: bump tcell

landism:
This picks up a fix for #1728 from tcell.

Tested by:

export GOPATH=/tmp
export TERM=xterm-color

before the upgrade, this repro'd the error in #1728 
after the upgrade, tilt ran fine


jazzdan:
Hmm looks like you have some dependency problems?

PR #1751, opened 2019-06-14
engine: unresourced YAML always builds first

### LINKED ISSUE #1515: Tilt fails to apply namespace first
A user tried to load yaml containing both namespaces and objects declared in those namespaces.

Theoretically, tilt applies the namespaces first, so things work out.

For some reason, in this case, the objects declared in those namespaces are applied first, and yield an error because the namespaces have not yet been created. If you re-run tilt, the namespaces still exist, and things work fine.

Repro:
Save [this](https://gist.githubusercontent.com/fentas/bab98a9559368c4fef3b2fcb33a3356c/raw/f103d080cbe8445bf47cbc1a7d4f307f15c21c40/gistfile1.txt) into gist.yaml
make a Tiltfile that just has `k8s_yaml('gist.yaml')`
`tilt up`

expected:
the resources come up successfully
observed:
rook-ceph-operator fails to apply with:
```
       ERROR: kubectl apply:
       stderr: Error from server (NotFound): error when creating "STDIN":
       namespaces "rook-ceph-system" not found
```
(note, if reproing multiple times, you'll need to `tilt down` between attempts, to ensure the namespace gets deleted)

The same problem probably applies to CRDs.

### LINKED ISSUE #1737: Order of k8s_yaml commands
My `Tiltfile` is as following:
```python
k8s_yaml('path/to/yaml/that/creates/namespace')

k8s_yaml('path/to/deployment/that/created/in/namespace/above')
```

When I'm running `tilt up` I got the namespace created but error on creating deployment that the namespace doesn't exist.

The second time it runs ok.

jazzdan:
Fixes #1515
Fixes #1737

### DIFF
--- a/internal/engine/buildcontroller.go
+++ b/internal/engine/buildcontroller.go
@@ -40,9 +40,12 @@ func nextTargetToBuild(state store.EngineState) *store.ManifestTarget {
 		return nil
 	}
 
-	// put no-build manifests first since they're more likely to be
-	// 1. fast and 2. dependencies of other services (e.g., redis)
+	// unresourced YAML goes first
 	targets := append([]*store.ManifestTarget{}, state.Targets()...)
+	findUnresourcedYAMLAndPutItFirst(targets)
+
+	// put no-build manifests next since they're more likely to be
+	// 1. fast and 2. dependencies of other services (e.g., redis)
 	sort.Sort(newNoBuildsManifestsFirst(targets))
 
 	// First, go through all the manifests in order.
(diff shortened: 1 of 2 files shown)


internal/engine/buildcontroller.go
```diff
@@ -92,6 +95,19 @@ func nextTargetToBuild(state store.EngineState) *store.ManifestTarget {
 	return choice
 }
 
+func findUnresourcedYAMLAndPutItFirst(input []*store.ManifestTarget) {
+	for i := range input {
+		if input[i].Manifest.ManifestName() == model.UnresourcedYAMLManifestName {
+			found := input[i]
+			// Remove found
+			input = append(input[:i], input[i+1:]...)
```
landism: I thiiiink:
1. it's not generally safe to assign to a slice param and expect the changes to affect the caller
2. it happens to work most or all of the time in this case due to implementation details

Maybe I'm just not sufficiently comfortable with slice internals, but I'd prefer to either return the modified slice, or limit operations in this function to strictly reordering elements rather than reassigning to `input`, e.g. (untested):
```
found := input[i]
copy(input[1:], append(input[:i], input[i+1:]...))
input[0] = found
```

jazzdan: That works for me!

Issue #1753, opened 2019-06-16
High memory consumption


dotellie:
This is probably more of a question but I'm seeing some pretty extreme memory usage on my system with the `tilt` coming in at about 3.5GB of consumed RAM. Is this normal and what could it possibly be due to? My Tiltfile deploys my Helm app (with `helm template`) which only has two services where one of them is a custom Docker build and the other is Redis.

dotellie:
I noticed that the problem seems to have been that my docker build context included my `target` folder (Rust project) which tilt kept in memory for some reason. That seems like a bug to me but I'm not quite sure :sweat_smile:

landism:
Nope, you're right. Tilt shouldn't be using 3.5GB of ram. We'll have a release out later today that will allow you to grab a heap dump and let us see more clearly what's causing the problem.

Thanks for the report!

landism:
Sorry for the delay on this! We've just released Tilt 0.9.0. As of this version, you can hit ctrl+O in the terminal UI to save a heap dump.

Once you've done that:
1. send us the dump directly (it might be ~3.5GB, so this might not be super convenient for you)
2. run [pprof](https://github.com/google/pprof) to analyze it and send us the analysis (`go tool pprof tilt.heap_profile` and then `dot > heap`) (this also might not be convenient, especially if you don't already have go installed).

If you could do one of these, it'd make it much easier for us to understand why Tilt is misbehaving for you, and more likely that whatever fix we implement is actually fixing your problem rather than an adjacent problem.

If it's a hassle for you, that's totally understandable, and we'll look into other ways to diagnose this.

If you have any questions, feel free to ask here or in #tilt on the K8S Slack.

dotellie:
So I did both and, surprisingly, while gnome system monitor still reported insane memory usage from tilt (now up to about >5GB), the heapfile only ended up at a few kilobytes. Both files you requested are linked to below but I suppose those might not be of much help since they don't correspond with what the rest of the system says.

https://drive.google.com/open?id=1hyE5YXVlRxeLOMhtPU5fJT2EOCckPZlU
https://drive.google.com/open?id=1OCWqMOO9tWbuNwwFPRaJdbelyc7LRltZ

(btw, I'm on the Slack channel you mentioned. Feel free to ping me if you want to debug it in realtime @Ellie)

landism:
I haven't gotten to a diagnosis yet, but...

Silly me. This is pretty trivial to repro:

[...]

PR #1754, opened 2019-06-17
hud: add ^O key to write a heap profile

landism:
to aid in diagnosing #1753

### DIFF
--- a/internal/engine/upper.go
+++ b/internal/engine/upper.go
@@ -4,7 +4,10 @@ import (
 	"context"
 	"fmt"
 	"io/ioutil"
+	"os"
 	"path/filepath"
+	"runtime"
+	"runtime/pprof"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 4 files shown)


internal/engine/upper.go
```diff
@@ -400,6 +405,29 @@ func handleStartProfilingAction(state *store.EngineState) {
 	state.IsProfiling = true
 }
 
+func handleHeapProfileAction(ctx context.Context) {
```
nicks: i don't understand why this is part of the engine state loop at all? it doesn't actually read any engine state

nicks: (seems like it'd be much simpler as a library function we call directly from the HUD)

landism: Hm, yeah, my mental model was this is the "respond to action" loop rather than the "engine state" loop.
It ended up here because: the most cohesive place (from the perspective of this feature, but rather uncohesive from the perspective of hud.go) to put it was in the `case ^O:` right in hud.go. I perhaps prematurely generalized and figured we might want to allow that from the web ui, and making it happen in response to an action seemed like a pretty good way to allow that.

Apparently at some point I started thinking of actions as "make a thing happen" rather than a mechanism for serializing state mutation.

internal/hud/hud.go
```diff
@@ -353,3 +360,26 @@ func selectedResource(view view.View, state view.ViewState) (i int, resource vie
 }
 
 var _ store.Subscriber = &Hud{}
+
+func writeHeapProfile(ctx context.Context) {
+	go func() {
```
nicks: this extra goroutine is probably unnecessary now?

nicks:
ya, i wonder if we can tweak the Action/Reducer API to discourage people from using it this way. e.g., remove Context. or to put it in a package that can't have any dependendencies.

PR #1758, opened 2019-06-21
tiltfile: track the current orchestator, so that we can run integration tests on microk8s

nicks:
Hello @maiamcc, @landism,

Please review the following commits I made in branch nicks/microk8s:

2e1b0a8fc8a4639c85be23881e09ed23b06da32b (2019-06-21 12:10:38 -0400)
tiltfile: track the current orchestator, so that we can run integration tests on microk8s

### DIFF
--- a/internal/tiltfile/tiltfile_test.go
+++ b/internal/tiltfile/tiltfile_test.go
@@ -2591,6 +2591,31 @@ k8s_yaml('foo.yaml')
 		deployment("foo"))
 }
 
+func TestPrivateRegistryDockerCompose(t *testing.T) {
+	f := newFixture(t)
+	defer f.TearDown()
+
+	f.kCli.Registry = "localhost:32000"
+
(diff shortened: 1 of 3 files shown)


internal/tiltfile/tiltfile_test.go
```diff
@@ -2591,6 +2591,31 @@ k8s_yaml('foo.yaml')
...
+docker_build('gcr.io/foo', 'foo')
+docker_compose('docker-compose.yml')
+`)
+
+	f.load()
+
+	// Assert that when we use a docker-compose orchestrator, we don't
+	// use the k8s private registry.
+	f.assertNextManifest("foo",
+		db(image("gcr.io/foo")))
```
maiamcc: I assume that this works b/c using the private registry would change the name of the image?

nicks: ya

PR #1760, opened 2019-06-21
docker: split client providers into a local docker client and a cluster docker client

nicks:
Hello @landism,

Please review the following commits I made in branch nicks/dockerclients:

a3883595b188e34281755575852a386cfd42f702 (2019-06-21 17:48:51 -0400)
docker: split client providers into a local docker client and a cluster docker client


nicks:
so happy to finally remove the dependency of the synclet on minikube code!

this is part of a bigger cleanup of how we use the docker client

PR #1764, opened 2019-06-24
improve rbac integration tests

nicks:
Hello @jazzdan,

Please review the following commits I made in branch nicks/minikubewatch:

3d49718c2e3d64fde6d686713f04a82c63a28ada (2019-06-21 19:29:12 -0400)
improve rbac integration tests

### DIFF
--- a/integration/k8s_fixture_test.go
+++ b/integration/k8s_fixture_test.go
@@ -8,11 +8,10 @@ import (
 	"encoding/base64"
 	"fmt"
 	"io/ioutil"
+	"log"
 	"net/http"
 	"os"
 	"os/exec"
-	"os/user"
-	"path/filepath"
 	"strings"
 	"testing"
 	"time"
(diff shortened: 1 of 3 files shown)


nicks:
(the main change here is to have a ClusterRole that grants access to Nodes, so our service accounts can read the k8s runtime)

nicks:
(we should check with customers with RBAC if they can read Nodes, otherwise we'll need to find some other way to detect runtime)

jazzdan:
I know that some can't, @nicks but in those cases it uses the the default update strategy anyways so it hasn't mattered.

integration/k8s_fixture_test.go
```diff
@@ -181,26 +179,17 @@ func (f *k8sFixture) ClearNamespace() {
...
+	f.kubeconfigPath = f.tempDir.JoinPath("config")
+	f.tempDir.WriteFile(f.kubeconfigPath, string(current))
+	f.fixture.tiltEnviron["KUBECONFIG"] = f.kubeconfigPath
 	f.fixture.tiltEnviron["TILT_K8S_EVENTS"] = "true"
-	f.usingOverriddenConfig = true
-}
-
-func (f *k8sFixture) createUser() {
-	f.runCommandSilently("kubectl", "apply", "-f", "access.yaml")
+	log.Printf("New kubeconfig: %s", f.kubeconfigPath)
```
jazzdan: 👀

nicks: this was deliberate!

PR #1765, opened 2019-06-24
Custom_build ignore

mariabelenvivanco:


internal/tiltfile/docker.go
```diff
@@ -251,6 +249,7 @@ func (s *tiltfileState) customBuild(thread *starlark.Thread, fn *starlark.Builti
 		"disable_push?", &disablePush,
 		"live_update?", &liveUpdateVal,
 		"match_in_env_vars?", &matchInEnvVars,
+		"ignore?", &ignoreVal, // added parameters for ignore
```
jazzdan: I'm not sure this comment adds extra context to the code. What do you think?

internal/tiltfile/docker.go
```diff
@@ -286,6 +285,21 @@ func (s *tiltfileState) customBuild(thread *starlark.Thread, fn *starlark.Builti
 		return nil, errors.Wrap(err, "live_update")
 	}
 
+	// parsing ignore values to strings to be added to dockerimage struct
+	tempIgnores := starlarkValueOrSequenceToSlice(ignoreVal)
```
jazzdan: I noticed that this code is exactly the same as the code on [L173](https://github.com/windmilleng/tilt/blob/6ce179c49633b3aa4a8185df54e3a3ee980ad961/internal/tiltfile/docker.go#L173). Should we factor it out in to a function?

internal/tiltfile/tiltfile_test.go
```diff
@@ -1880,7 +1880,7 @@ hfb.hot_reload()`
 
 	f.loadAssertWarnings(fastBuildDeprecationWarning)
 	f.assertNumManifests(1)
-	f.assertConfigFiles("Tiltfile", ".tiltignore", "foo.yaml", "foo/.dockerignore")
+	f.assertConfigFiles("Tiltfile", ".tiltignore", "foo.yaml", "foo/.dockerignore", ".dockerignore")
```
jazzdan: Hmm why did this test change?

internal/tiltfile/docker.go
```diff
@@ -365,6 +334,25 @@ func (b *customBuild) addFastBuild(thread *starlark.Thread, fn *starlark.Builtin
 	return &fastBuild{s: b.s, img: b.img}, nil
 }
 
+func parseValuesToStrings(value starlark.Value, s string) ([]string, error) {
```
jazzdan: Maybe rename this function to `valueToStringsForParam` so it's more obvious what `s` is for?

```go
func valueToStringsForParam(value starlark.Value, param string)
```

internal/tiltfile/tiltfile_test.go
```diff
@@ -1941,7 +1941,7 @@ hfb = custom_build(
 
 	f.load("foo")
 	f.assertNumManifests(1)
-	f.assertConfigFiles("Tiltfile", ".tiltignore", "foo.yaml", ".dockerignore")
+	f.assertConfigFiles("Tiltfile", ".tiltignore", "foo.yaml", "foo/.dockerignore") //deps
```
jazzdan: Should this comment be there?

PR #1770, opened 2019-06-24
build: change ArchiveBuilder to return io.Reader

jazzdan:
This will make it easier to refactor ArchiveBuilder to not use a
`bytes.Buffer` in the future, which should help address #1753.

### DIFF
--- a/internal/engine/synclet_build_and_deployer.go
+++ b/internal/engine/synclet_build_and_deployer.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"context"
 	"fmt"
+	"io"
 	"strings"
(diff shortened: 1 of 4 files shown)


internal/engine/synclet_build_and_deployer.go
```diff
@@ -170,17 +171,23 @@ func (sbd *SyncletBuildAndDeployer) updateInCluster(ctx context.Context, iTarget
 	return nil
 }
 
+func streamToByte(stream io.Reader) []byte {
```
landism: fwiw this is `ioutil.ReadAll`

PR #1774, opened 2019-06-25
integration: make onewatch work on containerd

nicks:
Hello @jazzdan,

Please review the following commits I made in branch nicks/kind:

fa0f6f27c6ae84618998ce1484e669f96f50d52a (2019-06-25 17:30:17 -0400)
integration: make onewatch work on containerd

### DIFF
--- a/integration/start.sh
+++ b/integration/start.sh
@@ -1,0 +1,26 @@
+#!/bin/sh
+#
+# A helper script to implement restart_container when the docker runtime isn't available.
+#
+# Usage:
+#   Copy start.sh and restart.sh to your container working dir.
+#
+#   Make your container entrypoint:
+#   ./start.sh path-to-binary [args]
+#
+#   To restart the container:
+#   ./restart.sh
+
+while true; do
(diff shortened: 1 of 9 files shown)


nicks:
I'm not super happy with this solution, but it works

Maybe we could do something clever where, if you specify restart_container(), we inject our own entrypoint that does something like this to allow restarts

integration/start.sh
```diff
@@ -0,0 +1,26 @@
...
+#   Make your container entrypoint:
+#   ./start.sh path-to-binary [args]
+#
+#   To restart the container:
+#   ./restart.sh
+
+while true; do
+    rm -f restart.txt
+    
+    $* &
```
jazzdan: FYI I am not very good at bash, and don't feel comfortable reviewing this part.

nicks: lol it's very cryptic but also straightforward.

"$*" means "all the arguments"

"&" means "run this in the background"

so if I run
./start.sh sleep 1000
then this just runs
sleep 1000 &

PR #1775, opened 2019-06-25
integration: all tests pass on containerd/KIND

nicks:
Hello @jazzdan,

Please review the following commits I made in branch nicks/kind2:

4ddb34df3d3bee4a125633d61abe7a9925345d7f (2019-06-25 18:07:26 -0400)
integration: all tests pass on containerd/KIND
- we can delete live_update_test.go, because now it's just a duplicate of onewatch
- live_update_base_image needed the busybox tar workaround

fa0f6f27c6ae84618998ce1484e669f96f50d52a (2019-06-25 17:30:17 -0400)
integration: make onewatch work on containerd


integration/restart.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/sh
```
jazzdan: I think these .sh scripts are in here twice now?

nicks: yep! it's meant to be copied anywhere that needs it

PR #1776, opened 2019-06-25
integration: run tests on Kind

nicks:
Hello @jazzdan, @landism,

Please review the following commits I made in branch nicks/kind3:

d4a708a37501b94f42003271beaa7d2d71040437 (2019-06-25 18:51:51 -0400)
integration: run tests on Kind
Should make the tests faster and more robust, because
we no longer need to push to a remote registry

eac18bf14334d01fcc760c8d32ffa7aeba1741d6 (2019-06-25 18:51:51 -0400)
integration: all tests pass on containerd/KIND
- we can delete live_update_test.go, because now it's just a duplicate of onewatch
- live_update_base_image needed the busybox tar workaround


nicks:
This cuts the integration tests from ~9 min to about ~7 min, and gets rid of the need for gcloud credentials.

I think there are ways we can cut it down even further with https://github.com/windmilleng/kind-local, but that will require some Tilt changes

PR #1784, opened 2019-06-27
start.sh: fix quoting, kill underlying process on exit

landism:
1. we want `"$@"` instead of `$*` to ensure that we run the command with the same quotes as the original command
2. while I was testing that, it turned out that killing the script didn't kill the underlying process, which might not matter for k8s purposes, but is nice to do for local testing

### DIFF
--- a/integration/onewatch/start.sh
+++ b/integration/onewatch/start.sh
@@ -11,10 +11,22 @@
 #   To restart the container:
 #   ./restart.sh
 
+set -euo pipefail
+
+process_id=""
+
+trap quit TERM INT
+
+quit() {
+  if [ -n "$process_id" ]; then
+    kill $process_id
+  fi
+}
+
 while true; do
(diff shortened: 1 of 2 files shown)


jazzdan:
By the way @landism there are multiple coopies of this file in the repo

integration/onewatch/start.sh
```diff
@@ -29,8 +29,7 @@ while true; do
     "$@" &
     process_id=$!
     echo "$process_id" > process.txt
-    wait $process_id
-    EXIT_CODE=$?
+    wait $process_id && EXIT_CODE=$? || EXIT_CODE=$?
```
landism: when we killed the process to restart, `wait` returned a non-0 exit, so the script exited because of `set -e`

nicks: would it make sense to just `set +e` before the `wait`?

PR #1788, opened 2019-06-27
engine: fix exec syncing on busybox

nicks:
Hello @jazzdan,

Please review the following commits I made in branch nicks/tarballs2:

dfa81af83362ddabb0465e48499ad24a9edca99b (2019-06-27 15:36:04 -0700)
engine: fix exec syncing on busybox
Fixes https://github.com/windmilleng/tilt/issues/1611

### DIFF
--- a/integration/live_update_base_image/app/Dockerfile
+++ b/integration/live_update_base_image/app/Dockerfile
@@ -1,8 +1,5 @@
 FROM python:alpine
 
-# TODO(nick): remove this when we have fixed GNU/busybox tar incompatibilites
-RUN apk add tar
-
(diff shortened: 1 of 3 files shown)


nicks:
tbh no idea why this fixes the problem 🤷‍♂

matthiasak:
that's weird @nicks - busybox doesn't respect /dev/stdin? :man_shrugging:

nicks:
@matthiasak ya! it's definitely weird. I suspect the problem is that the busybox tar assumes all files are regular files (whereas /dev/stdin is an abstract file descriptor that doesn't support arbitrary seek)

PR #1792, opened 2019-06-28
Docker Compose Multiple Files 

mariabelenvivanco:
Allows users to enter one or more files into their docker_compose function in their Tiltfile.

### DIFF
--- a/internal/cli/down.go
+++ b/internal/cli/down.go
@@ -37,6 +37,7 @@ func (c *downCmd) run(ctx context.Context, args []string) error {
 	defer a.Flush(time.Second)
 
 	downDeps, err := wireDownDeps(ctx, a)
+
 	if err != nil {
 		return err
 	}
@@ -56,27 +57,17 @@ func (c *downCmd) run(ctx context.Context, args []string) error {
 		logger.Get(ctx).Infof("error deleting k8s entities: %v", err)
 	}
 
-	var dcConfigPath []string
+	var dcConfigPaths []string
 	for _, m := range tlr.Manifests {
 		if m.IsDC() {
-			// TODO(maia): when we support up-ing from multiple docker-compose files, we'll
-			// need to support down-ing as well. For now, we `down` the first one we find.
-			dcConfigPath = m.DockerComposeTarget().ConfigPath
+			dcConfigPaths = m.DockerComposeTarget().ConfigPaths
 			break
 		}
 	}
-
-	for _, config := range dcConfigPath {
-		if config != "" {
-			// TODO(maia): when we support up-ing from multiple docker-compose files, we'll need to support down-ing as well
-			// TODO(maia): a way to `down` specific services?
-			// downing the first one
-			dcc := downDeps.dcClient
-			err = dcc.Down(ctx, dcConfigPath, logger.Get(ctx).Writer(logger.InfoLvl), logger.Get(ctx).Writer(logger.InfoLvl))
-			if err != nil {
-				logger.Get(ctx).Infof("error running `docker-compose down`: %v", err)
-			}
-		}
+	dcc := downDeps.dcClient
+	err = dcc.Down(ctx, dcConfigPaths, logger.Get(ctx).Writer(logger.InfoLvl), logger.Get(ctx).Writer(logger.InfoLvl))
+	if err != nil {
+		logger.Get(ctx).Infof("error running `docker-compose down`: %v", err)
 	}
 	return nil
 }
--- a/internal/cli/down.go
+++ b/internal/cli/down.go
@@ -56,7 +56,7 @@ func (c *downCmd) run(ctx context.Context, args []string) error {
 		logger.Get(ctx).Infof("error deleting k8s entities: %v", err)
 	}
 
-	var dcConfigPath string
+	var dcConfigPath []string
 	for _, m := range tlr.Manifests {
 		if m.IsDC() {
 			// TODO(maia): when we support up-ing from multiple docker-compose files, we'll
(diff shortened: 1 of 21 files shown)


internal/cli/down.go
```diff
@@ -66,14 +66,16 @@ func (c *downCmd) run(ctx context.Context, args []string) error {
...
-
-		dcc := downDeps.dcClient
-		err = dcc.Down(ctx, dcConfigPath, logger.Get(ctx).Writer(logger.InfoLvl), logger.Get(ctx).Writer(logger.InfoLvl))
-		if err != nil {
-			logger.Get(ctx).Infof("error running `docker-compose down`: %v", err)
+	for _, config := range dcConfigPath {
+		if config != "" {
+			// TODO(maia): when we support up-ing from multiple docker-compose files, we'll need to support down-ing as well
+			// TODO(maia): a way to `down` specific services?
+			// downing the first one
```
jazzdan: It looks like we're not downing the first one, but downing all of them?

internal/dockercompose/client.go
```diff
@@ -19,13 +19,13 @@ import (
...
 
 type DockerComposeClient interface {
-	Up(ctx context.Context, configPath string, serviceName model.TargetName, shouldBuild bool, stdout, stderr io.Writer) error
-	Down(ctx context.Context, configPath string, stdout, stderr io.Writer) error
-	StreamLogs(ctx context.Context, configPath string, serviceName model.TargetName) (io.ReadCloser, error)
-	StreamEvents(ctx context.Context, configPath string) (<-chan string, error)
-	Config(ctx context.Context, configPath string) (string, error)
-	Services(ctx context.Context, configPath string) (string, error)
-	ContainerID(ctx context.Context, configPath string, serviceName model.TargetName) (container.ID, error)
+	Up(ctx context.Context, configPath []string, serviceName model.TargetName, shouldBuild bool, stdout, stderr io.Writer) error
```
jazzdan: Maybe rename these parameters to `configPaths`?

internal/engine/docker_compose_event_watcher.go
```diff
@@ -36,7 +36,7 @@ func (w *DockerComposeEventWatcher) OnChange(ctx context.Context, st store.RStor
 	configPath := state.DockerComposeConfigPath()
 	st.RUnlockState()
 
-	if configPath == "" {
+	if configPath == nil {
```
jazzdan: I think this is slightly wrong. This will catch cases where the slice is nil, but not cases where the slice is empty. It's the difference between:

```golang
slice := nil
slice := []string{}
```

I think instead this should be:

```golang
if len(configPath) == 0 {
```

See this example https://play.golang.org/p/gOogOof4XI0

internal/engine/docker_compose_log_manager.go
```diff
@@ -111,7 +111,7 @@ func (m *DockerComposeLogManager) consumeLogs(watch dockerComposeLogWatch, st st
 	}()
 
 	name := watch.name
-	readCloser, err := m.dcc.StreamLogs(watch.ctx, watch.dc.ConfigPath, watch.dc.Name)
+	readCloser, err := m.dcc.StreamLogs(watch.ctx, watch.dc.ConfigPath, watch.dc.Name) //todo(maria)
```
jazzdan: 👀

internal/model/deploy_info.go
```diff
@@ -136,7 +136,7 @@ func (dc DockerComposeTarget) Validate() error {
 		return fmt.Errorf("[Validate] DockerCompose resource missing name:\n%s", dc.YAMLRaw)
 	}
 
-	if dc.ConfigPath == "" {
+	if dc.ConfigPath == nil {
```
jazzdan: Same thing here: we should do a `len` check instead.

internal/tiltfile/docker_compose_test.go
```diff
@@ -11,6 +11,7 @@ import (
 )
 
 // ParseConfig must return services topologically sorted wrt dependencies.
+
```
jazzdan: I think there are some extra spaces in this file.

mariabelenvivanco:
@jazzdan  The other commit that isn't included is where I fixed some of the tests that failed. I will make the changes and push the commits as one.

internal/dockercompose/client.go
```diff
@@ -148,21 +165,22 @@ func (c *cmdDCClient) StreamEvents(ctx context.Context, configPath string) (<-ch
 		err = cmd.Wait()
 		if err != nil {
 			logger.Get(ctx).Infof("[DOCKER-COMPOSE WATCHER] exited with error: %v", err)
+			logger.Get(ctx).Infof("command running %v", cmd)
```
jazzdan: 👀

internal/dockercompose/client.go
```diff
@@ -177,8 +195,13 @@ func (c *cmdDCClient) dcCommand(ctx context.Context, args []string) *exec.Cmd {
 	return cmd
 }
 
-func (c *cmdDCClient) dcOutput(ctx context.Context, configPath string, args ...string) (string, error) {
-	args = append([]string{"-f", configPath}, args...)
+func (c *cmdDCClient) dcOutput(ctx context.Context, configPath []string, args ...string) (string, error) {
+
+	var tempArgs []string
```
jazzdan: Hmm why do we need this `tempArgs` variable?

internal/dockercompose/fake_client.go
```diff
@@ -26,7 +26,7 @@ type FakeDCClient struct {
 
 // Represents a single call to Up
 type UpCall struct {
-	PathToConfig string
+	PathToConfig []string
```
jazzdan: This variable name seems misleading now. Maybe rename it to `ConfigPaths`?

PR #1796, opened 2019-07-01
feature: simple feature flag framework backed by static map

jazzdan:


internal/feature/flags.go
```diff
@@ -0,0 +1,59 @@
...
+
+type staticMapFeature struct {
+	flags map[string]bool
+	mu    *sync.Mutex
+}
+
+func (f *staticMapFeature) IsEnabled(flag string) bool {
+	f.mu.Lock()
+	defer f.mu.Unlock()
+	return f.flags[flag]
```
nicks: what happens if we typo a flag name in IsEnabled?

jazzdan: It will return false, but I think it should return an error. Good call.

internal/feature/flags.go
```diff
@@ -0,0 +1,59 @@
...
+	IsEnabled(flag string) bool
+}
+
+type Writer interface {
+	Enable(flag string) error
+	Disable(flag string) error
+}
+
+func NewStaticMapFeature() *staticMapFeature {
+	return &staticMapFeature{flags: flags, mu: &sync.Mutex{}}
```
nicks: it seems bad for all instances of this class to share a single global map? was that intentional? seems like we should dependency-inject this

jazzdan: It was intentional: there are only two different kinds of feature flag sets we will ever want. One for testing and one for everything else. As a result it didn't seem necessary to dependency inject the flag set because we want to discourage people from providing their own.

I don't feel super strongly though and I'm happy to add it.

nicks: talked about this offline. his will cause problems, because one test will enable the flag, and then the global state will affect other tests. let's copy the map so that we don't rely on global state

internal/feature/flags.go
```diff
@@ -0,0 +1,72 @@
...
+	newMap := map[string]bool{}
+	for key, value := range defaults {
+		newMap[key] = value
+	}
+	return &staticMapFeature{flags: newMap, mu: &sync.Mutex{}}
+}
+
+type staticMapFeature struct {
+	flags map[string]bool
+	mu    *sync.Mutex
```
landism: nit: this doesn't need to be a pointer since the method receivers are pointers (which then also means this field doesn't need to be initialized)

PR #1798, opened 2019-07-01
engine: wire up feature, check in events code

jazzdan:
This commit does two things:
1) Wire up feature in to the place where we will set it (tiltfile), and
the place where we will read it initially (event watch manager)
2) Check the feature flag correctly in event watch manager

A future commit will add syntax to the tiltfile that allow features
to be modified


internal/engine/event_watch_manager.go
```diff
@@ -33,18 +34,21 @@ type EventWatchManager struct {
...
 		clock:   clock,
+		f:       f,
 	}
 }
 
 func (m *EventWatchManager) needsWatch(st store.RStore) bool {
-	if !k8sEventsFeatureFlagOn() {
+	enabled, err := m.f.IsEnabled("events")
+	if !enabled || err != nil {
 		return false
```
landism: FFTI since this is good enough as is, but something I wish I'd thought of on the previous PR:
Ignoring the error here doesn't feel great. Do you think it's worth going with types instead of strings, or is that annoying to do right now and/or likely to scale poorly at scales we're likely to want? e.g.:
```
type Feature interface {
  Enable()
  IsEnabled() bool
  Disable()
}

type FeatureConfig struct {
  Events Feature
}
```

landism: We should probably also handle the case of enabled going from true to false (i.e., stop the watch).

jazzdan: Not annoying to do right now and I don't foresee any scaling problems. What problem are you looking to solve? How does that distinguish between features that are off and features that don't exist any better than this?

internal/engine/event_watch_manager.go
```diff
@@ -33,18 +34,21 @@ type EventWatchManager struct {
...
 		kClient: kClient,
 		uidMap:  make(map[k8s.UID]uidMapEntry),
 		clock:   clock,
+		f:       f,
 	}
 }
 
 func (m *EventWatchManager) needsWatch(st store.RStore) bool {
-	if !k8sEventsFeatureFlagOn() {
+	enabled, err := m.f.IsEnabled("events")
```
nicks: maybe IsEnabled should just panic if the key doesn't exist? seems wrong to drop the error silently

jazzdan: Yeah I'm down with a panic, good idea.

internal/engine/event_watch_manager.go
landism: it'd be a compiler error instead of a runtime error

landism: errr, it'd be a runtime error on startup, when parsing feature configs, but the code *using* the configs wouldn't have to worry about it

jazzdan: Yeah that's what I ended up doing

jazzdan: OH now I understand what this is saying. Basically make it a typed map, rather than dynamic.

jazzdan: I think I'm going to leave this as is so that we have flexibility in the future to have this backed by something else, like a networked backed store, as much as I like the compile time safety this solution offers.

PR #1810, opened 2019-07-05
cli: make sure tilt down fails properly if the kubectl/docker-compose commands fail

nicks:
Hello @maiamcc,

Please review the following commits I made in branch nicks/ch2853:

ff12f2afc3121f203b042401f1f6dfc76067b071 (2019-07-05 13:40:06 -0400)
cli: make sure tilt down fails properly if the kubectl/docker-compose commands fail

### DIFF
--- a/internal/cli/down.go
+++ b/internal/cli/down.go
@@ -2,7 +2,6 @@ package cli
 
 import (
 	"context"
-	"fmt"
 	"time"
 
 	"github.com/pkg/errors"
@@ -21,7 +20,8 @@ type downCmd struct {
 func (c *downCmd) register() *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "down",
-		Short: "delete kubernetes resources",
+		Short: "delete resources created by 'tilt up'",
+		Args:  cobra.NoArgs,
 	}
 
 	cmd.Flags().StringVar(&c.fileName, "file", tiltfile.FileName, "Path to Tiltfile")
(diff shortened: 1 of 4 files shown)


internal/cli/down.go
```diff
@@ -21,7 +20,8 @@ type downCmd struct {
 func (c *downCmd) register() *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "down",
-		Short: "delete kubernetes resources",
+		Short: "delete resources created by 'tilt up'",
+		Args:  cobra.NoArgs,
```
maiamcc: hang on, this should be able to take args -- e.g. if someone is using the `--file` arg to specify a non-standard Tiltfile path, or if you only `tilt up`'d services A and B, and so only want to `tilt down` those specific ones, lest you get an error.

internal/cli/down.go
```diff
@@ -64,10 +66,14 @@ func (c *downCmd) run(ctx context.Context, args []string) error {
 			break
 		}
 	}
-	dcc := downDeps.dcClient
-	err = dcc.Down(ctx, dcConfigPaths, logger.Get(ctx).Writer(logger.InfoLvl), logger.Get(ctx).Writer(logger.InfoLvl))
-	if err != nil {
-		logger.Get(ctx).Infof("error running `docker-compose down`: %v", err)
+
+	if len(dcConfigPaths) > 0 {
```
maiamcc: ha i was gonna leave this for maria to fix but okay!

nicks: i couldn't fix the bug without also fixing this problem

nicks: (or the integration tests would have failed)

internal/cli/down.go
nicks: re: the --file arg

this still works fine. cobra.NoArgs only applies to positional args

https://godoc.org/github.com/spf13/cobra#PositionalArgs

re: if you only tilt up'd services A and B, and so only want to tilt down those specific ones

that seems reasonable, but that doesn't work in the current code. do you think we should file a feature request for it now, or when someone requests it?

internal/cli/down.go
maiamcc: i figured as much :)

internal/cli/down.go
maiamcc: ah, gotcha. yeah let's wait until someone complains!

Issue #1682, opened 2019-05-24
K8s objects not included in the resources above


Bram-Zijp:
When using Tilt instead of the vanilla way to apply/create your kubernetes files I get errors of K8s objects not being included.

For example `./frontend/configmap.yml`:
```
apiVersion: v1
kind: ConfigMap
metadata:
  name: frontend-configmap
```
 `./frontend/deployment.yml`:
```
...
      env:
        - name: FRONTEND_CONFIG
          valueFrom:
            configMapKeyRef:
              name: frontend-configmap
```

Above example will throw out the following error:
```
     K8s objects not included in the resources above:                                                                                 
      k8sConfigMap-frontend-configmap
```

Is this a bug or can I resolve this by configuration?

nicks:
Oh, I think this is fine? I assume you're seeing something like this:

<img width="639" alt="Screen Shot 2019-05-24 at 2 09 33 PM" src="https://user-images.githubusercontent.com/278641/58329914-a4802380-7e2d-11e9-9d99-fb48bf7dceaa.png">

It's not an error message. It's just trying to inform you that k8s_yaml includes something that it couldn't figure out how to group on the other lines. I agree it's kind of cryptic looking

@yuindustries what do you think?

Bram-Zijp:
@nicks Already had the feeling this was a feature and not a bug. So it just means it won't display as the  `tilt-frontend-demo ------ Running...` line in your screenshot.

My English isn't all that great but maybe it's an idea to change the line to something like:
*K8s objects not shown in the resource list above.*

maiamcc:
Some copy ideas:
- "All other k8s objects"
- "Other k8s objects (those not included in resources above)"

PR #1811, opened 2019-07-05
hud: unify how we compute workload names and uncategorized resource names

nicks:

Please review the following commits I made in branch nicks/yaml:

004f512382bf3f49f2c43ceaf207e54ec5d5bd69 (2019-07-05 15:55:54 -0400)
hud: unify how we compute workload names and uncategorized resource names
Fixes https://github.com/windmilleng/tilt/issues/1682


nicks:
it seemed weird to have two different ways to display k8s resources

also tweaked the explanation a bit so that it doesn't sound like an error message
Screenshot from 2019-07-05 15-59-10

jazzdan:
Might also address #1682

internal/k8s/names.go
```diff
@@ -0,0 +1,51 @@
...
+	"strings"
+)
+
+// Calculates names for workloads by using the shortest uniquely matching identifiers
+func UniqueNames(es []K8sEntity, minComponents int) ([]string, error) {
+	ret := make([]string, len(es))
+	// how many resources potentially map to a given name
+	counts := make(map[string]int)
+
+	// count how many entity want each potential name
```
jazzdan:
```suggestion
	// count how many entities want each potential name
```

Issue #1817, opened 2019-07-08
File watch leak


jazzdan:
We've had at least one report of the number of watches reported by the expvars server increasing steadily after each save. From @dotellie in the #tilt channel:

> Hm so I have that page open right now while doing some development. Refreshing it periodically, I can see that the number of watches steadily increase after each save. I've never worked with watches myself before, but unless this is picking up on build artifacts from RLS (which it shouldn't since I have ignores on that basically everywhere), I would suspect this is some kind of leak.

This shouldn't happen. First step is to try to reproduce the problem in a test project.

nicks:
to be clear, is this on macos or linux?

dotellie:
In my case, it's Linux :penguin:

jazzdan:
I've got a repro case! It's very alarming. On Linux:

```
$ git clone --branch dmiller/add-tiltfile https://github.com/jazzdan/Rocket.git
$ cd Rocker
$ code .
# change `example/hello_world/src/main.rs`
# tilt starts a rebuild and then crashes with the inotify event queue overflow error
```

nicks:
this should be fixed in the next release (by pr https://github.com/windmilleng/tilt/pull/1865)

PR #1819, opened 2019-07-08
watch: record which dirs are watched in expvars

jazzdan:
To help diagnose #1817

Do we think that this will cause memory issues if it's enabled for everyone?

### DIFF
--- a/internal/watch/watcher_naive.go
+++ b/internal/watch/watcher_naive.go
@@ -36,7 +36,8 @@ type naiveNotify struct {
 }
 
 var (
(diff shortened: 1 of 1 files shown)


nicks:
hmmmmm...have you been able to reproduce the problem locally? have you tried setting up a local Rust project in Tilt yet?

adding these kinds of metrics seems like a last-resort after other routes have been exhausted

jazzdan:
@nicks haven't tried with Rust, no. Will do.

jazzdan:
Closing because I have a repro case now, described in #1817

nicks:
:clap:

PR #1823, opened 2019-07-08
watch: don't watch ignored paths

### LINKED ISSUE #1817: File watch leak

jazzdan:
Fixes #1817

Not super confident in the API changes to notify for this. Going to sleep on it.

### DIFF
--- a/internal/engine/watchmanager.go
+++ b/internal/engine/watchmanager.go
@@ -195,8 +195,19 @@ func (w *WatchManager) OnChange(ctx context.Context, st store.RStore) {
 			continue
 		}
 
+		filter, err := ignore.CreateFileChangeFilter(target)
+		if err != nil {
+			st.Dispatch(NewErrorAction(err))
+			return
+		}
+		tiltIgnoreFilter, err := dockerignore.DockerIgnoreTesterFromContents(tiltRoot, w.tiltIgnoreContents)
+		if err != nil {
+			st.Dispatch(NewErrorAction(err))
+		}
+		filter = model.NewCompositeMatcher([]model.PathMatcher{filter, tiltIgnoreFilter})
+
 		for _, d := range target.Dependencies() {
-			err = watcher.Add(d)
+			err = watcher.Add(d, filter)
 			if err != nil {
 				st.Dispatch(NewErrorAction(err))
 			}
@@ -204,7 +215,7 @@ func (w *WatchManager) OnChange(ctx context.Context, st store.RStore) {
 
 		ctx, cancel := context.WithCancel(ctx)
 
-		go w.dispatchFileChangesLoop(ctx, target, watcher, st, tiltRoot)
+		go w.dispatchFileChangesLoop(ctx, target, watcher, st, tiltRoot, filter)
 		newWatches[target.ID()] = targetNotifyCancel{target, watcher, cancel}
 	}
 
@@ -231,19 +242,8 @@ func (w *WatchManager) dispatchFileChangesLoop(
 	target WatchableTarget,
 	watcher watch.Notify,
 	st store.RStore,
-	tiltRoot string) {
-
-	filter, err := ignore.CreateFileChangeFilter(target)
-	if err != nil {
-		st.Dispatch(NewErrorAction(err))
-		return
-	}
-	tiltIgnoreFilter, err := dockerignore.DockerIgnoreTesterFromContents(tiltRoot, w.tiltIgnoreContents)
-	if err != nil {
-		st.Dispatch(NewErrorAction(err))
-	}
-	filter = model.NewCompositeMatcher([]model.PathMatcher{filter, tiltIgnoreFilter})
-
+	tiltRoot string,
+	filter model.PathMatcher) {
 	eventsCh := coalesceEvents(w.timerMaker, watcher.Events())
 
 	for {
--- a/internal/engine/watchmanager.go
+++ b/internal/engine/watchmanager.go
@@ -270,7 +270,14 @@ func (w *WatchManager) dispatchFileChangesLoop(
 					st.Dispatch(NewErrorAction(err))
 					continue
 				}
-				watchEvent.files = append(watchEvent.files, path)
+				isIgnored, err := filter.Matches(path, false)
+				if err != nil {
+					st.Dispatch(NewErrorAction(err))
+					continue
+				}
+				if !isIgnored {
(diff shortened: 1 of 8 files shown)


internal/engine/watchmanager.go
```diff
@@ -195,16 +195,27 @@ func (w *WatchManager) OnChange(ctx context.Context, st store.RStore) {
...
+		}
+		tiltIgnoreFilter, err := dockerignore.DockerIgnoreTesterFromContents(tiltRoot, w.tiltIgnoreContents)
+		if err != nil {
+			st.Dispatch(NewErrorAction(err))
+		}
+		filter = model.NewCompositeMatcher([]model.PathMatcher{filter, tiltIgnoreFilter})
+
 		for _, d := range target.Dependencies() {
-			err = watcher.Add(d)
+			err = watcher.Add(d, filter)
```
nicks: would it make more sense to pass `filter` in to fsWatcherMaker?

internal/engine/watchmanager.go
```diff
@@ -195,16 +195,27 @@ func (w *WatchManager) OnChange(ctx context.Context, st store.RStore) {
...
 			if err != nil {
 				st.Dispatch(NewErrorAction(err))
 			}
 		}
 
 		ctx, cancel := context.WithCancel(ctx)
 
-		go w.dispatchFileChangesLoop(ctx, target, watcher, st, tiltRoot)
+		go w.dispatchFileChangesLoop(ctx, target, watcher, st, tiltRoot, filter)
```
nicks: passing `filter` in here should be redundant now, since the watcher should now filter these events internally?

internal/watch/watcher_darwin.go
```diff
@@ -71,12 +73,20 @@ func (d *darwinNotify) loop() {
...
 }
 
-func (d *darwinNotify) Add(name string) error {
+func (d *darwinNotify) Add(name string, filter model.PathMatcher) error {
 	d.sm.Lock()
 	defer d.sm.Unlock()
 
 	es := d.stream
 
+	isIgnored, err := filter.Matches(name, false)
```
nicks: i don't think this will work correctly in the case where subpaths are watched

internal/watch/watcher_naive.go
```diff
@@ -184,8 +186,15 @@ func (d *naiveNotify) shouldNotify(path string) bool {
 	return false
 }
 
-func (d *naiveNotify) add(path string) error {
-	err := d.watcher.Add(path)
+func (d *naiveNotify) add(path string, filter model.PathMatcher) error {
+	isIgnored, err := filter.Matches(path, false)
```
nicks: ditto

internal/engine/watchmanager.go
jazzdan: Yes, I like that approach better.

landism:
I was a bit surprised that this is fixing a leak when it doesn't seem to be cleaning up any watches, so I figured I'd try to state my understanding:
1. tilt never cleans up its watches
2. rust writes a lot of files to the `target` directory, using new paths for each build
3. (1)+(2) means after every new rust build, tilt has a lot of new watches, and they never go away
4. this could be mitigated by specifying ignore='target' in the Tiltfile, except ignore doesn't stop Tilt from watching files because of the issue discussed in #1580 
5. This PR seeks simply to re-allow the mitigation mentioned in (4)

Do you have a guess as to whether cleaning up watches is hard? IIUC, the approach in this PR:
1. Requires the user to recognize this problem and then change their config to work around it.
2. Doesn't address the same problem in other cases where the new path names aren't ignorable.

I don't mean this to oppose this PR, and having a workaround is better than not, and this solves an actual user problem, so that's good. I'm just looking to understand.

jazzdan:
All of that is correct.

To be clear this isn't fixing a leak. I don't think there is a leak _in this case_. I think there's just a lot of files being written.

A leak is possible, if files are removed we don't remove watches on Linux. I don't think that's what's happenin gin this case though.

> Do you have a guess as to whether cleaning up watches is hard?
No, I haven't looked in to it at all. It wouldn't solve the problem in this case, right? The watches are legit.

internal/engine/watchmanager.go
jazzdan: It _should_ but the watchmanger tests depend on the watchmanager having this behavior in `dispatchFileChangesLoop` and trying to move that behavior in to the fake notifier is proving frustrating/flakey for reasons I'm not sure of.

jazzdan:
After talking with @landism and @nicks last night I agree that the better approach is to lift [Docker's strategy](https://github.com/moby/moby/blob/master/pkg/archive/archive.go#L833) wholesale. This addresses a bunch of scenarios that are currently problematic in our system now, such as: 

```
target
!target/*.go
```

However the way we are handling filters right now makes this system hard to adopt. Right now we are just storing an [interface](https://github.com/windmilleng/tilt/blob/a44beb3b199ce0db4839596b7948c4de47907367/internal/model/matcher.go#L12-L14) in memory to decide whether things should be included in a build and trigger builds for a given target. To use Docker's approach we would need to store a list of the include/exclude patterns and introspect them as we traverse the directory structure. (see https://github.com/moby/moby/blob/master/pkg/archive/archive.go#L850-L858)

I propose that we go with this PR's solution for now, to solve problems on Linux where there are too many files to watch, and schedule time to adopt the approach that Docker uses later.

jazzdan:
After talking with @landism we think a workaround could be to somehow encode whether there any ignore _exclusions_, ie:

```
directory
!directory/*.go
```

Which is read as “ignore directory EXCEPT for any Go files in directory”. If the ignores do have such an exclusion, which I expect to be uncommon, then we revert back to watching everything. If there are no exclusions we can be moderately smart as this code does.

nicks:
fwiw, i don't think we can move forward with this PR as-is without introducing unacceptable bugs. i can discuss more detail in-person if necessary

jazzdan:
Closing this in favor of another approach @nicks is trying.

Issue #1785, opened 2019-06-27
Web UI crashes if resource doesn't exist 


mariabelenvivanco:
![image](https://user-images.githubusercontent.com/43042150/60284417-7e8ce980-98d9-11e9-9559-0fbc26366f64.png)

1.  Click on resource on the sidebar 
2. Removed that resource from the Tiltfile 

OR 
1. Go to the URL of a resource that doesn't exist. For example: http://localhost:10350/r/removedresource

PR #1826, opened 2019-07-09
bug: fixing web UI crashes whenever resource has been deleted

mariabelenvivanco:
As a response to  #1785 , the webUI crashes whenever a resource is deleted. This pull redirects to the main page and shows error whenever the user tries to reach the deleted resource's URL.


nicks:
@mariabelenvivanco looks like you need to run `make prettier` to format the JS

PR #1828, opened 2019-07-09
tiltfile: print output of local() to the logger

nicks:

Please review the following commits I made in branch nicks/running:

4799f142e3dbcf2fd64ab677a119e2f5a7cb839a (2019-07-09 15:02:18 -0400)
tiltfile: print output of local() to the logger

### DIFF
--- a/internal/tiltfile/docker.go
+++ b/internal/tiltfile/docker.go
@@ -25,7 +25,7 @@ type dockerImage struct {
 	baseDockerfile     dockerfile.Dockerfile
 	configurationRef   container.RefSelector
 	deploymentRef      reference.Named
-	syncs              []sync
+	syncs              []pathSync
 	runs               []model.Run
 	entrypoint         string
 	cachePaths         []string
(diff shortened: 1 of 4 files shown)


nicks:
Screenshot from 2019-07-09 15-01-47

nicks:
i was trying to debug some problems and having a lot of trouble...this seems generally more useful

internal/tiltfile/docker.go
```diff
@@ -25,7 +25,7 @@ type dockerImage struct {
 	baseDockerfile     dockerfile.Dockerfile
 	configurationRef   container.RefSelector
 	deploymentRef      reference.Named
-	syncs              []sync
+	syncs              []pathSync
```
jazzdan: Are these changes supposed to be here?

nicks: yes. you can't have an `import` and a `type` with the same name. the go compiler yells at you

jazzdan: Ah, gotcha makes sense

PR #1829, opened 2019-07-09
engine: print out the resources we're deploying

nicks:

Please review the following commits I made in branch nicks/apply:

b35bffe3508bb0c1032ec8f5aab5ca80b4055819 (2019-07-09 15:31:09 -0400)
engine: print out the resources we're deploying


nicks:
Screenshot from 2019-07-09 15-30-43

Issue #1821, opened 2019-07-08
Flag to disable automatic opening of WebUI 


mariavivanco:
Every time Tilt starts up, it opens a new tab for the web UI even if there is a Tilt tab open already as reported by @dotellie in the #tilt channel:
> I have the tab permanently pinned in Chrome so I don't really want another one to open 
 
Edit: Tilt should have a flag that can disable the automatic opening of new Tilt WebUI tabs.

dotellie:
This isn't really a bug as much as it's a limitation of current browsers. Tilt detects that a tab is already open but that's only if Chrome is still actively keeping that tab alive. As an example, if I exit Tilt, keep my computer on during the night with another tab open in Chrome, Chrome will decide that the Tilt tab isn't important anymore and put it to sleep which means that Tilt can't detect that a tab is already open, causing it to open another tab. What really is needed here is just a flag to disable the automatic opening outright :wink:

nicks:
This has been available for a while as `--no-browser`

tumregels:
Just checked on `v0.18.9` the flag `--no-browser` is not there

```
$ tilt version
v0.18.9, built 2021-02-12
$ tilt up --no-browser
...
unknown flag: --no-browser
```

nicks:
Hi @tumregels ! We removed the flag because Tilt doesn't open a browser by default anymore. Is there something else you were looking for?

tumregels:
Hi @nicks. Sorry, just realized that can connect to `http://localhost:10350` after `tilt up` with using space.

kahirokunn:
How can I rather open the browser automatically now?
Thx 🙏

PR #1830, opened 2019-07-09
Flag for opening browser

mariabelenvivanco:
As a response to #1821 , this PR adds a command line flag `-- no-browser `to stop webUI from starting up automatically on startup.


internal/cli/up.go
```diff
@@ -211,6 +214,9 @@ func provideWebPort() model.WebPort {
 	return model.WebPort(webPort)
 }
 
+func provideNoBrowserFlag() bool {
```
nicks: can you create a `type NoBrowser bool` for this?

it's not just any boolean, it's a boolean for a specific thing

internal/hud/server/controller.go
```diff
@@ -28,8 +29,9 @@ type HeadsUpServerController struct {
 	initDone    bool
 }
 
-func ProvideHeadsUpServerController(port model.WebPort, hudServer *HeadsUpServer, assetServer assets.Server, webURL model.WebURL) *HeadsUpServerController {
+func ProvideHeadsUpServerController(noBrowser bool, port model.WebPort, hudServer *HeadsUpServer, assetServer assets.Server, webURL model.WebURL) *HeadsUpServerController {
```
nicks: try running `make shorttest` so you can find all the test code that calls this function

internal/cli/up.go
maiamcc: ☝️ to add context to this: remember when we were looking at the generated wire code and i was so surprised that it put the boolean in the right place? that's because wire is basically guessing what function output to pass to what function, based on type. there's probably some magic that makes it more accurate but it's still kind of a guess that THIS boolean gets passed to THAT function.

by making a type alias, it's just a boolean under the hood, but wire knows that this function returns a `NoBrowser` and this other function accepts a `NoBrowser`, and they're the only two functions that do that, so it's really clear what code to generate. Does that make sense?

mariabelenvivanco: ooo ok noted! Thank you

PR #1832, opened 2019-07-10
build: stream tar to docker

### LINKED ISSUE #1753: High memory consumption

jazzdan:
The basic strategy is to have the tar archiver be responsible only for _writing_ to a `io.Writer`. Whenever we call the tar archiver we need to either

1. pass in a byte buffer and [deal with it](https://gph.is/1lDpHWH)
2. wire up an `io.Pipe`, pass the writer end in to the archiver and pass the reader end to Docker

This makes the tar package API truly gross, IMO. But that's a problem for a separate PR.

Memory allocations during tar before:

```
(pprof) top
Showing nodes accounting for 1.11GB, 99.72% of 1.11GB total
Dropped 42 nodes (cum <= 0.01GB)
Showing top 10 nodes out of 20
      flat  flat%   sum%        cum   cum%
    1.11GB 99.72% 99.72%     1.11GB 99.72%  bytes.makeSlice
         0     0% 99.72%     1.11GB 99.72%  archive/tar.(*Writer).Write
         0     0% 99.72%     1.11GB 99.72%  archive/tar.(*regFileWriter).Write
         0     0% 99.72%     1.11GB 99.72%  bytes.(*Buffer).Write
         0     0% 99.72%     1.11GB 99.72%  bytes.(*Buffer).grow
         0     0% 99.72%     1.11GB 99.72%  github.com/windmilleng/tilt/internal/build.(*ArchiveBuilder).ArchivePathsIfExist
         0     0% 99.72%     1.11GB 99.72%  github.com/windmilleng/tilt/internal/build.(*ArchiveBuilder).writeEntry
         0     0% 99.72%     1.11GB 99.72%  github.com/windmilleng/tilt/internal/build.(*dockerImageBuilder).BuildDockerfile
         0     0% 99.72%     1.11GB 99.72%  github.com/windmilleng/tilt/internal/build.(*dockerImageBuilder).buildFromDf
         0     0% 99.72%     1.11GB 99.72%  github.com/windmilleng/tilt/internal/build.tarContextAndUpdateDf
```

After:

```
(pprof) top
Showing nodes accounting for 5408.26kB, 100% of 5408.26kB total
Showing top 10 nodes out of 34
      flat  flat%   sum%        cum   cum%
 2804.01kB 51.85% 51.85%  2804.01kB 51.85%  github.com/windmilleng/tilt/vendor/github.com/gdamore/tcell.(*CellBuffer).Resize
  553.04kB 10.23% 62.07%   553.04kB 10.23%  github.com/windmilleng/tilt/vendor/github.com/gogo/protobuf/proto.RegisterType
  513.56kB  9.50% 71.57%   513.56kB  9.50%  github.com/windmilleng/tilt/vendor/k8s.io/apimachinery/pkg/api/meta.(*DefaultRESTMapper).AddSpecific
  513.50kB  9.49% 81.06%   513.50kB  9.49%  github.com/windmilleng/tilt/vendor/github.com/rivo/tview.init.ializers
  512.12kB  9.47% 90.53%   512.12kB  9.47%  runtime.mcommoninit
  512.02kB  9.47%   100%   512.02kB  9.47%  fmt.Sprintf
         0     0%   100%   512.02kB  9.47%  fmt.Errorf
         0     0%   100%   513.56kB  9.50%  github.com/windmilleng/tilt/internal/cli.(*upCmd).run
         0     0%   100%  2804.01kB 51.85%  github.com/windmilleng/tilt/internal/cli.(*upCmd).run.func1
         0     0%   100%   513.56kB  9.50%  github.com/windmilleng/tilt/internal/cli.Execute
```

Fixes #1753

### DIFF
--- a/cmd/buildkitapi/main.go
+++ b/cmd/buildkitapi/main.go
@@ -41,20 +41,25 @@ func run() error {
 
 	d.NegotiateAPIVersion(ctx)
 
-	archive, err := build.TarPath(ctx, ".")
-	if err != nil {
-		return err
-	}
+	pr, pw := io.Pipe()
+	go func() {
+		err := build.TarPath(ctx, pw, ".")
+		if err != nil {
+			pw.CloseWithError(err)
+		} else {
+			pw.Close()
+		}
+	}()
 
 	opts := types.ImageBuildOptions{}
 	opts.Version = types.BuilderBuildKit
 	opts.Dockerfile = "Dockerfile"
-	opts.Context = archive
+	opts.Context = pr
 	if !useCache {
 		opts.NoCache = true
 	}
 
-	response, err := d.ImageBuild(ctx, archive, opts)
+	response, err := d.ImageBuild(ctx, pr, opts)
 	if err != nil {
 		return err
 	}
--- a/cmd/buildkitapi/main.go
+++ b/cmd/buildkitapi/main.go
@@ -45,9 +45,9 @@ func run() error {
 	go func() {
 		err := build.TarPath(ctx, pw, ".")
 		if err != nil {
-			pw.CloseWithError(err)
+			_ = pw.CloseWithError(err)
 		} else {
-			pw.Close()
+			_ = pw.Close()
 		}
 	}()
 
--- a/internal/build/image_builder.go
+++ b/internal/build/image_builder.go
@@ -319,7 +319,6 @@ func (d *dockerImageBuilder) buildFromDf(ctx context.Context, ps *PipelineState,
 
 	ps.StartBuildStep(ctx, "Building image")
 	spanBuild, ctx := opentracing.StartSpanFromContext(ctx, "daemon-ImageBuild")
-	// pass pr here
 	imageBuildResponse, err := d.dCli.ImageBuild(
 		ctx,
 		pr,
--- a/internal/build/image_builder.go
+++ b/internal/build/image_builder.go
@@ -311,9 +311,9 @@ func (d *dockerImageBuilder) buildFromDf(ctx context.Context, ps *PipelineState,
 	go func() {
 		err := tarContextAndUpdateDf(ctx, pw, df, paths, filter)
 		if err != nil {
-			pw.CloseWithError(err)
+			_ = pw.CloseWithError(err)
 		} else {
-			pw.Close()
+			_ = pw.Close()
 		}
 	}()
 
--- a/internal/build/image_builder.go
+++ b/internal/build/image_builder.go
@@ -307,17 +307,23 @@ func (d *dockerImageBuilder) buildFromDf(ctx context.Context, ps *PipelineState,
 		}
 	}
 
-	archive, err := tarContextAndUpdateDf(ctx, df, paths, filter)
-	if err != nil {
-		return nil, err
-	}
+	pr, pw := io.Pipe()
+	go func() {
+		err := tarContextAndUpdateDf(ctx, pw, df, paths, filter)
+		if err != nil {
+			pw.CloseWithError(err)
+		} else {
+			pw.Close()
+		}
+	}()
 
 	ps.StartBuildStep(ctx, "Building image")
 	spanBuild, ctx := opentracing.StartSpanFromContext(ctx, "daemon-ImageBuild")
+	// pass pr here
 	imageBuildResponse, err := d.dCli.ImageBuild(
 		ctx,
-		archive,
-		Options(archive, buildArgs),
+		pr,
+		Options(pr, buildArgs),
 	)
 	spanBuild.Finish()
 	if err != nil {
--- a/internal/build/tar_test.go
+++ b/internal/build/tar_test.go
@@ -3,10 +3,10 @@ package build
 import (
 	"archive/tar"
 	"context"
+	"io"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
-
 	"github.com/windmilleng/tilt/internal/dockerfile"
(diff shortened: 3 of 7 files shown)


cmd/buildkitapi/main.go
```diff
@@ -41,20 +41,25 @@ func run() error {
...
 	d.NegotiateAPIVersion(ctx)
 
-	archive, err := build.TarPath(ctx, ".")
-	if err != nil {
-		return err
-	}
+	pr, pw := io.Pipe()
+	go func() {
+		err := build.TarPath(ctx, pw, ".")
+		if err != nil {
```
nicks: you can get rid of this `if` branch. it's worth pointing out that Close() is literally the same as CloseWithError(nil)
https://golang.org/src/io/pipe.go?s=3046:3098#L128

internal/build/tar_test.go
```diff
@@ -17,18 +17,22 @@ import (
...
+	pr, pw := io.Pipe()
+	go func() {
+		ab := NewArchiveBuilder(pw, model.EmptyMatcher)
+		defer ab.Close()
+		defer f.tearDown()
+
+		df := dockerfile.Dockerfile(dfText)
+		err := ab.archiveDf(f.ctx, df)
+		if err != nil {
+			t.Fatal(err)
```
nicks: you can't use t.Fatal in a goroutine. it won't work correctly and weird things will happen

https://golang.org/pkg/testing/#T

internal/build/tar_test.go
```diff
@@ -17,18 +17,22 @@ import (
...
-	defer ab.close()
-	defer f.tearDown()
 
 	dfText := "FROM alpine"
-	df := dockerfile.Dockerfile(dfText)
-	err := ab.archiveDf(f.ctx, df)
-	if err != nil {
-		t.Fatal(err)
-	}
+	pr, pw := io.Pipe()
```
nicks: it would be a lot simpler to use a bytes.Buffer in all these tests. it should work fine as both a reader and a writer. and then you don't need the goroutine.

certainly, we should have at least one test that exercises concurrency, but you don't need all of them to do the concurrency dance.

internal/build/image_builder.go
```diff
@@ -307,17 +307,23 @@ func (d *dockerImageBuilder) buildFromDf(ctx context.Context, ps *PipelineState,
...
+		if err != nil {
+			pw.CloseWithError(err)
+		} else {
+			pw.Close()
+		}
+	}()
 
 	ps.StartBuildStep(ctx, "Building image")
 	spanBuild, ctx := opentracing.StartSpanFromContext(ctx, "daemon-ImageBuild")
+	// pass pr here
```
landism: 👀

internal/engine/synclet_build_and_deployer.go
```diff
@@ -130,15 +131,13 @@ func (sbd *SyncletBuildAndDeployer) updateInCluster(ctx context.Context, iTarget
 	containerPathsToRm := build.PathMappingsToContainerPaths(toRemove)
 
 	// archive files to copy to container
-	ab := build.NewArchiveBuilder(ignore.CreateBuildContextFilter(iTarget))
+	// TODO(dmiller) maybe don't do this?
```
nicks: ?

internal/build/tar_test.go
landism: t.Fatal doesn't work in goroutines. I thought go vet caught this...maybe I'm thinking of `golangci-lint`

internal/build/tar_test.go
```diff
@@ -17,18 +17,22 @@ import (
...
+	go func() {
```
landism: for tests I think we could just use a `bytes.Buffer` for the `Writer` and skip the pipe and goroutine

internal/engine/synclet_build_and_deployer.go
landism: would this just be a pretty straightforward application of the pipe/go used elsewhere in this PR?

jazzdan: The problem is both `updateViaExec` and `updateViaSynclet` take the paths that we archived as a parmater which the current API doesn't expose to us before the archiving is done, and now the archiving is done asynchronously in a separate goroutine.

jazzdan: https://github.com/windmilleng/tilt/pull/1832#discussion_r302265678

landism: ah...It looks like `updateViaExec` only uses it for logging purposes. Maybe we can drop that param and then add a `-v` to the `tar xf` [here](https://github.com/windmilleng/tilt/blob/c0a9ce2d6b88f5f79a32aeac9ca50d81ce6332f3/internal/engine/synclet_build_and_deployer.go#L226).

And then IIUC, `updateViaSynclet` only takes the list of files to delete, not the list of files that we archived, so that's not a problem.

jazzdan: Yeah I'm down for that. Let me fix these other test failure first.

jazzdan: I'm also tempted not to fix this since it seems like the synclet may be getting deprecated?

PR #1833, opened 2019-07-10
tiltfile: add 'entrypoint' opt arg to docker_build and custom_build, translate into image target [ch2886]

maiamcc:
Hello @landism, @nicks, @mariabelenvivanco,

Please review the following commits I made in branch maiamcc/tiltfile-override-entrypoint:

7f8a4527c3f97b522e35b7566f0c6d71babfab2d (2019-07-10 17:18:09 -0400)
rename a fastBuild-related field to free up better names

29bc4349a2f670a6635d702e36828307caa0d80e (2019-07-10 17:16:15 -0400)
tests

79181b35f8e53117620beaa4aa1f5239a69d1b44 (2019-07-10 17:16:01 -0400)
tiltfile: add 'entrypoint' opt arg to docker_build and custom_build, translate into image target

### DIFF
--- a/internal/tiltfile/docker.go
+++ b/internal/tiltfile/docker.go
@@ -29,7 +29,7 @@ type dockerImage struct {
 	matchInEnvVars     bool
 	ignores            []string
 	onlys              []string
-	entrypoint         model.Cmd // optional: override container entrypoint/command with this
+	entrypoint         model.Cmd // optional: if specified, we override the image entrypoint/k8s command with this
 
 	// fast-build properties -- will be deprecated
 	syncs        []pathSync
--- a/internal/tiltfile/docker.go
+++ b/internal/tiltfile/docker.go
@@ -25,15 +25,17 @@ type dockerImage struct {
 	baseDockerfile     dockerfile.Dockerfile
 	configurationRef   container.RefSelector
 	deploymentRef      reference.Named
-	syncs              []pathSync
-	runs               []model.Run
-	entrypoint         string
 	cachePaths         []string
-	hotReload          bool
 	matchInEnvVars     bool
 	ignores            []string
 	onlys              []string
-	entrypointOverride model.Cmd // optional: override container entrypoint/command with this
+	entrypoint         model.Cmd // optional: override container entrypoint/command with this
+
+	// fast-build properties -- will be deprecated
+	syncs        []pathSync
+	runs         []model.Run
+	fbEntrypoint string
+	hotReload    bool
 
 	dbDockerfilePath localPath
 	dbDockerfile     dockerfile.Dockerfile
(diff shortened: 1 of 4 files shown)


internal/tiltfile/docker.go
```diff
@@ -25,14 +25,17 @@ type dockerImage struct {
...
 	deploymentRef      reference.Named
-	syncs              []pathSync
-	runs               []model.Run
-	entrypoint         string
 	cachePaths         []string
-	hotReload          bool
 	matchInEnvVars     bool
 	ignores            []string
 	onlys              []string
+	entrypoint         model.Cmd // optional: if specified, we override the image entrypoint/k8s command with this
```
maiamcc: I would have called this "command" to mirror what we call it [on the ImageTarget](https://github.com/windmilleng/tilt/blob/b258a687d3a6bd1e283806d63406d59917434972/internal/model/docker_info.go#L23) / because it's most accurately overriding the k8s yaml `container.command` field, BUT that arg name is already taken by `custom_build` :-/ suggestions welcome

internal/tiltfile/tiltfile_docker_compose_test.go
```diff
@@ -490,6 +490,20 @@ dc_resource('no-svc-with-this-name-eek', 'gcr.io/foo')
 	f.loadErrString("no Docker Compose service found with name")
 }
 
+func TestDockerComposeDoesntSupportEntrypointOverride(t *testing.T) {
```
mariabelenvivanco: How do you handle invalid entrypoints for dockerBuild/customBuild? For example if a given entry point doesn't exist, how do we handle it?

internal/tiltfile/docker.go
maiamcc: ah, I would call it "command" not to make it obvious what the type is, but so it more closely matched the place in the system where we're injecting it (i.e. the `container.command` field of k8s yaml).

I think the general wisdom is that you should NOT try to name things after their types, b/c in a typesafe language like go, it should be obvious from inspecting any variable what type it is. e.g. calling something
```
var nameString string
```
is kind of redundant. (I still do it sometimes, but i'm probably not supposed to :P )

internal/tiltfile/docker.go
```diff
@@ -25,14 +25,17 @@ type dockerImage struct {
...
+
+	// fast-build properties -- will be deprecated
+	syncs        []pathSync
+	runs         []model.Run
+	fbEntrypoint string
```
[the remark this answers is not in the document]
maiamcc: you're totally correct! the only reason i'm doing it here is b/c the `fbEntrypoint` field ought to be gone from this code within the month 🤞

internal/tiltfile/tiltfile_docker_compose_test.go
maiamcc: i'm not sure what you mean by "doesn't exist"? but to answer your question, we don't try to validate that here -- we let the container try to run the entrypoint, and report any errors back. trying to validate the entrypoint here would be like trying to validate a user's dockerfile before we ran it -- potennnntially useful but kind of a waste of time.

PR #1835, opened 2019-07-10
watch: change the watcher interface to better match how we actually use it

nicks:
Hello @jazzdan,

Please review the following commits I made in branch nicks/watchapi:

5f5ac9d41d7258670ea7553b0f6d0bdf20609cc6 (2019-07-10 18:42:12 -0400)
watch: change the watcher interface to better match how we actually use it

### DIFF
--- a/internal/engine/fake_notify_test.go
+++ b/internal/engine/fake_notify_test.go
@@ -24,12 +24,12 @@ func newFakeMultiWatcher() *fakeMultiWatcher {
 	return r
 }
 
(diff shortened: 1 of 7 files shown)


nicks:
for background: originally, we wrote this data structure to be persistent and mutable, so you could add new watches while it's running.

in practice, we never want this to be mutable. we want to throw away the old one and create a new one whenever the watches change. But the interface had never been updated to reflect this usage pattern.

This should make the watches easier to change going forward, because we don't need to keep the pause/restart logic up-to-date.

PR #1854, opened 2019-07-15
tiltfile: handle duplicate resources without erroring

nicks:
Hello @landism,

Please review the following commits I made in branch nicks/duplicate:

33d4bb143e0e7db76dd8e3ce6143afe563ecd4b5 (2019-07-15 10:21:19 -0400)
tiltfile: handle duplicate resources without erroring
fixes https://github.com/windmilleng/tilt/issues/1853
helps with https://github.com/windmilleng/tilt/issues/1852, until we find a better way to address this

### DIFF
--- a/internal/k8s/names.go
+++ b/internal/k8s/names.go
@@ -6,7 +6,7 @@ import (
 )
 
 // Calculates names for workloads by using the shortest uniquely matching identifiers
-func UniqueNames(es []K8sEntity, minComponents int) ([]string, error) {
+func UniqueNames(es []K8sEntity, minComponents int) []string {
 	ret := make([]string, len(es))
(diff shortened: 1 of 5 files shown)


landism:
well, that's a bummer. is there a valid use case for duplicate resources, or is this purely a matter of "well kubectl accepts my yaml - why won't tilt?"

nicks:
ya, i think there are probably a lot of cases where people have scripts that pull in the same resource multiple times, and it's generally ok.

It's only really a problem if the resources differ in some way

maybe we should do a deep comparison and de-dupe

AleF83:
In my case I call kubectl apply to yaml with namespace definition first and then kubectl apply to the whole folder that contains several objects,  including the namespace.

Issue #1857, opened 2019-07-15
docker-compose: FSEventStreamCreate: _FSEventStreamCreate: ERROR: could not allocate 0 bytes for array of path strings


landism:
Repro:

build tilt @ 0f6d36e
in the [analytics repo](https://github.com/windmilleng/analytics), run `tilt up --hud=false`

printed to the screen is this:
```
2019-07-15 13:31 tilt[52973] (FSEvents.framework) FSEventStreamCreate: _FSEventStreamCreate: ERROR: could not allocate 0 bytes for array of path strings
2019-07-15 13:31 tilt[52973] (FSEvents.framework) FSEventStreamScheduleWithRunLoop(): failed assertion 'streamRef != NULL'

2019-07-15 13:31 tilt[52973] (FSEvents.framework) FSEventStreamStart(): failed assertion 'streamRef != NULL'
```

(these logs are printed directly to stdout, so if one runs w/o `--hud=false, the messages get garbled on top of the hud)

landism:
fixed by #1861

PR #1858, opened 2019-07-15
watch: fix behavior for unbuilt resources

landism:
### Problem

#1857 

This was introduced by #1835 - the analytics repo has an unbuilt resource, which means it has 0 dependencies and we pass a list of empty paths to the fsevents API, which then writes errors directly to stdout.

### Solution

Don't add watchers for empty lists of paths, matching the pre-#1835 behavior.

### DIFF
--- a/internal/engine/watchmanager.go
+++ b/internal/engine/watchmanager.go
@@ -203,7 +203,12 @@ func (w *WatchManager) OnChange(ctx context.Context, st store.RStore) {
 			continue
 		}
 
-		watcher, err := w.fsWatcherMaker(target.Dependencies(), ignore, logger)
+		deps := target.Dependencies()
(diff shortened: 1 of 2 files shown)


internal/engine/watchmanager.go
```diff
@@ -203,7 +203,12 @@ func (w *WatchManager) OnChange(ctx context.Context, st store.RStore) {
 			continue
 		}
 
-		watcher, err := w.fsWatcherMaker(target.Dependencies(), ignore, logger)
+		deps := target.Dependencies()
```
nicks: i wonder if this filtering should happen further up, like in watchableTargetsForManifest?

landism:
looks like this was taken care of by #1861

PR #1859, opened 2019-07-15
tiltfile/k8s: sort uncategorized yaml so that namespaces and then crds come first [ch2909]

maiamcc:
Harvested from [old code](https://github.com/windmilleng/tilt/pull/1233/files) in the global yaml controller.

When applying unresourced YAML (and in fact, probably all yaml), we should first apply namespaces, then crd definitions, then everything else (so we don't get an error when trying to create an entity in a nonexistent namespace, or an instance of a nonexistent CRD)

This PR sorts the unresourced yaml by these rules before attaching it to the manifest (rather than the old approach, to sort the entities every time we applied them)


internal/k8s/entity_test.go
```diff
@@ -173,6 +174,45 @@ func TestHasKind(t *testing.T) {
...
+	eCRD := mustParseYAML(t, testyaml.CRDYAML)[0]
+	eNamespace := mustParseYAML(t, testyaml.MyNamespaceYAML)[0]
+
+	for _, test := range []struct {
+		name           string
+		input          []K8sEntity
+		expectedOutput []K8sEntity
+	}{
+		{"namespace first",
+			[]K8sEntity{eDeploy, eNamespace, eService},
```
nicks: fuzz-testing is sometimes a more fun way to test these kinds of things. Just specify the correct input, call rand.Shuffle N times, and verify that the sort gets you back to the input.
https://golang.org/pkg/math/rand/#Shuffle

internal/tiltfile/tiltfile.go
```diff
@@ -166,6 +167,10 @@ func (tfl tiltfileLoader) Load(ctx context.Context, filename string, matching ma
 
 	yamlManifest := model.Manifest{}
 	if len(unresourced) > 0 {
+		// We apply Namespaces first b/c subsequent entities may depend on the NS existing.
+		// We apply CRDs first so we don't try to create an instance of that CRD before it's defined.
+		sort.Sort(k8s.NamespacesAndCRDsFirst(unresourced))
```
nicks: would it make sense to do this in NewK8sTarget, so that all K8sTargets get this treatment? i'm pretty sure you can write a Tiltfile that puts the Namespace/CRD in a named manifest

nicks: alternatively, you could do the Sort in Upsert, which is the only place this really matters

landism: Yeah, `Upsert` is probably the right place to do this. IIUC, `Upsert` moves all `Job`s and `Pod`s in front of everything else, so any sorting we do prior to that won't help us if we have both a pod and a namespace.

landism: (which was also a bug in the previous solution to this problem!)

maiamcc: oh womp. will put this in upsert, good tip!

internal/k8s/entity_test.go
maiamcc: yeah, I'd considered that! but i also wanted to ensure the behavior with e.g., only namespaces/crds, no namespaces/crds, etc.

internal/tiltfile/tiltfile.go
maiamcc: blurgh because of the way we upsert, sorting does us no good -- we pluck stuff out of the entities list to apply. fixing now.

internal/k8s/client.go
```diff
@@ -241,6 +241,15 @@ func (k K8sClient) Upsert(ctx context.Context, entities []K8sEntity) error {
 	span, ctx := opentracing.StartSpanFromContext(ctx, "daemon-k8sUpsert")
 	defer span.Finish()
 
+	// First apply all the entities on which something else might depend
+	withDeps, entities := EntitiesWithDependenciesAndRest(entities)
+	if len(withDeps) > 0 {
+		_, stderr, err := k.actOnEntities(ctx, []string{"apply"}, withDeps)
```
nicks: i wonder if this will ever need the retry logic below if the apply is rejected

internal/k8s/entity.go
```diff
@@ -158,6 +158,26 @@ func (e K8sEntity) DeepCopy() K8sEntity {
 	}
 }
 
+// EntitiesWithDependenciesAndRest returns two lists of k8s entities: those that may have dependencies,
+// which we will therefore want to apply first (i.e. namespaces and CRDs -- e.g. trying to create a
+// pod in a nonexistent namespace causes an error); and the rest of the entities.
+func EntitiesWithDependenciesAndRest(entities []K8sEntity) (withDeps, rest []K8sEntity) {
```
nicks: fwiw, if you have a Pod that depends on a namespace, I would usually interpret "Entity with a dependency" to mean the Pod

PR #1861, opened 2019-07-15
watch: skip setup if nothing is being watched

nicks:
Hello @jazzdan,

Please review the following commits I made in branch nicks/zerodeps:

9f588501593be67067cf92a7c22aa05384e94b00 (2019-07-15 18:46:39 -0400)
watch: skip setup if nothing is being watched


nicks:
(this basically replicates the old behavior, where the watcher doesn't even start if you don't call `Add()`)

PR #1868, opened 2019-07-17
internal/feature: noop and obsolete flags let us deprecate flags

dbentley:
We want to remove feature flags.

If Alice and Bob are on adjacent Tilt versions where the latter removes the flag, Bob will want to remove the flag, but then Alice will lose the feature.

Allow marking a flag "noop" (it's set to a fixed-value, but Bob doesn't see anything), then "obsolete" (Bob sees an error, removes it, and Alice isn't affected), and then removing it (where setting a value would be an error)

### DIFF
--- a/internal/feature/flags_test.go
+++ b/internal/feature/flags_test.go
@@ -6,9 +6,11 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
-func TestGetReturnsFalseIfKeyDoesntExist(t *testing.T) {
+func TestGetPanicsIfKeyDoesntExist(t *testing.T) {
 	m := FeatureSet{}
-	assert.False(t, m.Get("foo"))
+	assert.Panics(t, func() {
+		m.Get("foo")
+	})
 }
 
 func TestIsEnabled(t *testing.T) {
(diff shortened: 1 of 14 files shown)


internal/feature/flags_test.go
```diff
@@ -6,54 +6,45 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
-func TestIsEnabledReturnsAnErrorIfKeyDoesntExist(t *testing.T) {
-	m := newStaticMapFeature(map[string]bool{})
-	assert.Panics(t, func() {
-		m.IsEnabled("foo")
-	})
+func TestGetReturnsFalseIfKeyDoesntExist(t *testing.T) {
```
jazzdan: Do we want this to silently return false in our code? If we typo a feature checkflag name?

dbentley: We shouldn't be typo'ing checkflag names because we shoudl use feature.Foo instead of "foo". Maybe instead of map[string]Value it should be map[*struct{}]Value so that you *have* to use the right key?

Added the panic back. WDYT?

jazzdan: Ahh I see, yeah this LGTM. Thanks!

PR #1871, opened 2019-07-17
integration: create namespaces with Tilt, because namespace-creation seems like it breaks a lot

nicks:
Hello @maiamcc,

Please review the following commits I made in branch nicks/namespace:

ed6368789ce4b32a09354f1ef29e643258391c79 (2019-07-17 15:46:32 -0400)
integration: create namespaces with Tilt, because namespace-creation seems like it breaks a lot

### DIFF
--- a/integration/analytics/Tiltfile
+++ b/integration/analytics/Tiltfile
@@ -1,5 +1,10 @@
 # -*- mode: Python -*-
 
+# HACK: load namespaces on `tilt up` but not on `tilt down`
+load_namespace = not str(local('echo -n $SKIP_NAMESPACE'))
+if load_namespace:
+  k8s_yaml('../namespace.yaml')
+
 # If you get push errors, you can change the default_registry.
 # Create tilt_option.json with contents: {"default_registry": "gcr.io/my-personal-project"}
 # (with your registry inserted). tilt_option.json is gitignore'd, unlike Tiltfile
--- a/integration/k8s_fixture_test.go
+++ b/integration/k8s_fixture_test.go
@@ -36,7 +36,6 @@ func newK8sFixture(t *testing.T, dir string) *k8sFixture {
 	td := tempdir.NewTempDirFixture(t)
 
 	kf := &k8sFixture{fixture: f, tempDir: td}
-	kf.CreateNamespaceIfNecessary()
 	kf.ClearNamespace()
 	return kf
 }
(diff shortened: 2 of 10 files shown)


integration/k8s_fixture_test.go
```diff
@@ -161,18 +160,6 @@ func (f *k8sFixture) ClearResource(name string) {
 	}
 }
 
-func (f *k8sFixture) CreateNamespaceIfNecessary() {
```
maiamcc: this seems cheap, is it worth leaving in?

```
$ time kubectl apply -f ns.yaml
namespace/mynamespace unchanged

real	0m0.145s
user	0m0.120s
sys	0m0.037s
```

integration/analytics/Tiltfile
```diff
@@ -1,5 +1,9 @@
 # -*- mode: Python -*-
 
+load_namespace = not str(local('echo -n $SKIP_NAMESPACE'))
```
maiamcc: can you add comments here in all the Tiltfiles so future devs know it's part of the infra stuff and must be in every integ. test Tiltfile, rather than it being specific to this test?

integration/k8s_fixture_test.go
maiamcc: ...okay hmm I guess it's not thaaat cheap when it's a no-op... do you know when/why it was originally put in?

nicks:
hmmm...it might be more productive to talk about this in-person. this is addressing https://github.com/windmilleng/tilt/issues/1853

i.e., our tilt integration tests should be testing tilt functionality

nicks:
talked about this offline. added additional comments to explain the trade-offs we're making by not deleting the namespace on `tilt down`

nicks:
(and to be clear, this is not a new trade-off)

mallyvai:
FYI - just ran into this:

```[Tiltfile] unable to find a unique resource name for 'deployapi-legacy:deployment:mallyvai:apps'```

Here's what was running:

```
$ lc ps
POD_NAME           PROVISION            CONTAINER_NAME     VERSION                                           STATUS     AGE          RESTARTS     REASON
deployapi.legacy   successful           deployapi-legacy   legacy.a3ff29424382d77eab814b50eb86ba3b3cbc230a   Running    10 minutes   0
dynamodb.server    successful           dynamodb-server    5cc47d338caaa578e0424d9390b193faf07df80b          Running    a day        0
envoy.server       pending              envoy-discovery    c2b903421fa9f5947057aaced30aaad5f2e03688          Running    a day        0
envoy.server       pending              envoy-server       c2b903421fa9f5947057aaced30aaad5f2e03688          Running    a day        0
jenkins.server     successful           jenkins-server     f70b39c560bdda8f0d9cf30ff741a0f8cef67b39          Running    a day        0
redis.server       successful           redis-server       0036d010c106a16c8c40ca6d90a0bae6ee2b2240          Running    a day        0
sqs.server         successful           sqs-server         acc4d674f0a12196941783fbaf8dd5be61ddd3e0          Running    a day        0
```

PR #1875, opened 2019-07-18
integration: remove dependency on windmilleng/windmill

nicks:
Hello @landism,

Please review the following commits I made in branch nicks/integration:

ad37dd9706ab0d104338fd983678393ec8aad899 (2019-07-18 13:15:34 -0400)
integration: remove dependency on windmilleng/windmill


nicks:
looks like this slipped in accidentally :grimacing:

landism:
> looks like this slipped in accidentally 😬

Nope, this was intentional, with the aim of making it easy to keep the in-memory StatsServer's logic / api in sync with the production StatsServer's. Does that come with an unreasonable cost?

nicks:
talked about this offline, the real fix needs to be to move some of these bits to a public repo like windmilleng/stats

PR #1878, opened 2019-07-18
log: fix appending multiple lines at once w/ prefix

landism:
### Problem

The main log sometimes ends up with gobbledygook that does not appear in the resource log, e.g.:
```
alert-inges…┊   WORKDIR /usr/src/app
alert-inges…┊ 
 Ealert-inges…┊   COPY package.json ./
alert-inges…┊   COPY yarn.lock ./
alert-inges…┊ 
 Ealert-inges…┊ 
 EXPOSE 30alert-inges…┊ 
 Ealert-inges…┊ 
 EXPOSE 30alert-inges…┊ 
 EXPOSE 3000
alert-inges…┊   CMD ["yarn", "start"]
alert-inges…┊ 
```

The bug lies in [log.Append](https://github.com/windmilleng/tilt/blob/a6d13f0bb25655c6179629669ba991b26f6924e0/internal/model/log.go#L120). We're appending to `prefix` and sticking it in `logLines`, which means if we're given a `prefix` with enough cap, we could end up reusing the same backing array for multiple lines. From playing around with a unit test, this seems to only repro when 1) multiple lines are written in the same action and 2) the resource name is long enough to require truncating. I don't know about the internals of `append` or `fmt.Sprintf`, but, empirically, it looks like we don't get extra cap in the `prefix` `[]byte` when there's no truncation.

### Solution
pass `prefix` as a `string` to remove the possibility for `[]byte` shenanigans (also, just, like, a nicer API)

### DIFF
--- a/internal/engine/upper.go
+++ b/internal/engine/upper.go
@@ -16,12 +16,11 @@ import (
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/labels"
 
-	"github.com/windmilleng/tilt/internal/hud/server"
-
 	tiltanalytics "github.com/windmilleng/tilt/internal/analytics"
 	"github.com/windmilleng/tilt/internal/container"
 	"github.com/windmilleng/tilt/internal/dockercompose"
 	"github.com/windmilleng/tilt/internal/hud"
+	"github.com/windmilleng/tilt/internal/hud/server"
 	"github.com/windmilleng/tilt/internal/k8s"
 	"github.com/windmilleng/tilt/internal/logger"
 	"github.com/windmilleng/tilt/internal/model"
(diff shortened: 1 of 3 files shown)


nicks:
unit test?

landism:
> unit test?

I tried to think of a useful unit test for this before posting this PR and failed. I'm open to suggestions.

All of my work on this was writing a unit test and iterating on it.
It evolved like this:
1. I started off by calling it with the same prefix `[]byte` that I observed when I ran tilt, but the problem failed to repro, because it didn't have a sufficiently large cap.
2. I then successfully repro'd the issue by generating the prefix `[]byte` in the same way `Upper` does, resulting in the extra cap and allowing the corruption.
3. I then changed the signature so that there was no more `[]byte`, and the unit test just passed a string.

The only way in which test versions (1) and (2) differed was the caps of their `prefix`es, which is a trait that no longer exists, since it's now a string. Since we no longer have a cap, any unit test we try to write to repro the problem is effectively just version (1), which didn't repro the problem in the first place.
By the time I got to (3), the test didn't seem to meaningfully differ from the [existing prefix test](https://github.com/windmilleng/tilt/blob/9c95b88cca8bd28c98bbfe7addcab157ad68bc33/internal/model/log_test.go#L64), so I deleted it.

landism:
thanks!

PR #1879, opened 2019-07-18
integration: fix namespace creation on macos

landism:
whatever sh-compatibility mode mac's `/bin/sh` uses doesn't support `-n` on its `echo` 🤷‍♂ 
```
$ sh -c 'echo -n hello'
-n hello
```


nicks:
lol

PR #1883, opened 2019-07-19
engine: merge codepaths for container updates into a single BuildAndDeployer [ch2962]

maiamcc:
apologies for the size of this PR (a lot of it is moving files around/making methods public).

Esp. would love thoughts on behavior that I've failed to test.


internal/containerupdate/fake_container_updater.go
```diff
@@ -22,6 +26,15 @@ type UpdateContainerCall struct {
 	HotReload  bool
 }
 
+func (cu *FakeContainerUpdater) ValidateSpecs(specs []model.TargetSpec, env k8s.Env) error {
+	var err error
```
landism: if you wanna golf, this is all the same as:
```
err := cu.ValidateErr
cu.ValidateErr = nil
return err
```

internal/mode/update_mode.go
```diff
@@ -1,10 +1,10 @@
-package engine
+package mode
```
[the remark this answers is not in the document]
maiamcc: yeahhh it used to live in `engine` but iirc that didn't work because it needed to be used in `containerupdate`, and then i tried to put it in `model` but that didn't work b/c it needs some `build` stuff...? I... think? let me look and see what I can do.

internal/containerupdate/container_updater.go
```diff
@@ -2,13 +2,70 @@ package containerupdate
...
+	"github.com/windmilleng/tilt/internal/mode"
 	"github.com/windmilleng/tilt/internal/model"
 	"github.com/windmilleng/tilt/internal/store"
+	"github.com/windmilleng/tilt/internal/target"
 )
 
 type ContainerUpdater interface {
 	UpdateContainer(ctx context.Context, deployInfo store.DeployInfo,
 		archiveToCopy io.Reader, filesToDelete []string, cmds []model.Cmd, hotReload bool) error
+	ValidateSpecs(specs []model.TargetSpec, env k8s.Env) error
```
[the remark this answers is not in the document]
maiamcc: done!

internal/containerupdate/synclet_updater_test.go
```diff
@@ -0,0 +1,71 @@
...
+
+	"github.com/windmilleng/tilt/internal/k8s"
+	"github.com/windmilleng/tilt/internal/model"
+	"github.com/windmilleng/tilt/internal/testutils"
+)
+
+func TestSyncletUpdater_ValidateSpecs(t *testing.T) {
+	f := newSyncletFixture(t)
+
+	// iTarg := model.ImageTarget{}
```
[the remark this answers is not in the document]
maiamcc: erk, removed!

internal/containerupdate/fake_container_updater.go
maiamcc: done

internal/containerupdate/synclet_updater_test.go
```diff
@@ -0,0 +1,71 @@
...
+	// k8sTarg := model.K8sTarget{}
+	// dcTarg := model.DockerComposeTarget{}
+
+	for _, test := range []struct {
+		name        string
+		specs       []model.TargetSpec
+		env         k8s.Env
+		expectedErr error
+	}{
+		// {"can't update docker compose",
```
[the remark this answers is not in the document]
maiamcc: 😅 removed

internal/engine/image_build_and_deployer.go
```diff
@@ -80,14 +86,14 @@ func NewImageBuildAndDeployer(
 }
 
 // Turn on synclet injection. Should be called before any builds.
-func (ibd *ImageBuildAndDeployer) SetInjectSynclet(inject bool) {
+func (ibd *ImageBuildAndDeployer) NewWithInjectSynclet(inject bool) {
```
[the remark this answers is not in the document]
maiamcc: it should still be `SetInjectSynclet`

internal/engine/live_update_build_and_deployer.go
```diff
@@ -7,8 +7,11 @@ import (
 
 	"github.com/pkg/errors"
 
+	errors2 "github.com/windmilleng/tilt/internal/engine/errors"
```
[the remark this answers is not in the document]
maiamcc: 😅👍

internal/containerupdate/exec_updater.go
```diff
@@ -0,0 +1,74 @@
...
+	return validateSpecsOnlyImagesDeployedToK8s(specs, "ExecUpdater")
+}
+
+func (cu *ExecUpdater) UpdateContainer(ctx context.Context, deployInfo store.DeployInfo,
+	archiveToCopy io.Reader, filesToDelete []string, cmds []model.Cmd, hotReload bool) error {
+	span, ctx := opentracing.StartSpanFromContext(ctx, "ExecUpdater-UpdateContainer")
+	defer span.Finish()
+
+	if !hotReload {
+		// If we're using kubectl exec syncing, it implies a non-Docker runtime,
```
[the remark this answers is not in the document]
maiamcc: good call

internal/engine/build_and_deployer.go
```diff
@@ -87,34 +104,24 @@ func (composite *CompositeBuildAndDeployer) BuildAndDeploy(ctx context.Context,
...
-	if updMode == UpdateModeImage || updMode == UpdateModeNaive {
-		return BuildOrder{dcbad, ibad}
+func DefaultBuildOrderForK8s(liveUpdBAD k8sLiveUpdBAD, ibad *ImageBuildAndDeployer, dcbad *DockerComposeBuildAndDeployer, updMode mode.UpdateMode) K8sOrder {
+	if updMode == mode.UpdateModeImage || updMode == mode.UpdateModeNaive {
+		return K8sOrder(BuildOrder{ibad})
 	}
 
-	if updMode == UpdateModeKubectlExec {
-		return BuildOrder{sbad, dcbad, ibad}
+	k8sLiveUpdater := LiveUpdateBuildAndDeployer(*liveUpdBAD)
```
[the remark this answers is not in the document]
maiamcc: good call, done!

internal/mode/update_mode.go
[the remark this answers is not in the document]
maiamcc: oh right! `mode` needs `k8s` for [ProvideUpdateMode](https://github.com/windmilleng/tilt/blob/f0a4915e77040559b6da1e2acc6f340038b4dc5f/internal/mode/update_mode.go#L45).

internal/containerupdate/container_updater.go
```diff
@@ -2,13 +2,71 @@ package containerupdate
...
+}
+
+// Differentiate between the CU we'll need for k8s updates (may be docker || synclet || exec)
+// and the one we'll need for DC updates (always docker)
+type K8sContainerUpdater ContainerUpdater
+type DCContainerUpdater ContainerUpdater
+
+func ProvideK8sContainerUpdater(kCli k8s.Client, dCli docker.Client, sm SyncletManager, env k8s.Env, updMode mode.UpdateMode, runtime container.Runtime) K8sContainerUpdater {
+	if updMode == mode.UpdateModeImage || updMode == mode.UpdateModeNaive {
+		return NewExplodingContainerUpdater()
```
nicks: i worry that this ExplodingContainerUpdater is a code smell that our abstractions are wrong

internal/containerupdate/container_updater.go
```diff
@@ -2,13 +2,71 @@ package containerupdate
...
+func ProvideDCContainerUpdater(dCli docker.Client, updMode mode.UpdateMode, env k8s.Env, runtime container.Runtime) DCContainerUpdater {
+	if updMode == mode.UpdateModeImage || updMode == mode.UpdateModeNaive {
+		return NewExplodingContainerUpdater()
+	}
+
+	return NewDockerContainerUpdater(dCli, env, runtime)
+}
+
+// validation func shared by ExecUpdater and SyncletUpdater
+func specsAreOnlyImagesDeployedToK8s(specs []model.TargetSpec, updaterName string) (ok bool, msg string) {
```
nicks: i don't think this is accurate? we should be able to update base images fine

internal/containerupdate/container_updater.go
```diff
@@ -2,13 +2,71 @@ package containerupdate
...
+	"github.com/windmilleng/tilt/internal/mode"
 	"github.com/windmilleng/tilt/internal/model"
 	"github.com/windmilleng/tilt/internal/store"
+	"github.com/windmilleng/tilt/internal/target"
 )
 
 type ContainerUpdater interface {
 	UpdateContainer(ctx context.Context, deployInfo store.DeployInfo,
 		archiveToCopy io.Reader, filesToDelete []string, cmds []model.Cmd, hotReload bool) error
+	SupportsSpecs(specs []model.TargetSpec) (supported bool, msg string)
```
nicks: The responsibility of the SupportsSpecs function seems poorly delineated. In other words, it seems like the decision about "am i allowed to live update in this container" is spread out across a few different helpers, and it's poorly defined which helpers are responsible for which parts of the decision

internal/containerupdate/container_updater.go
maiamcc: so it's based on [this check from the old syncletBaD](http://bit.ly/2Y1VtEb), and it makes sense -- how can you do a live update on a base image, if the essence of a live update is that it works on a running container?

anyway, figured out that integration tests were failing b/c I was passing the wrong set of specs in -- we care about the specs extracted for LiveUpdate, which is a subset of all the specs passed into the BaD

nicks: ya, that's what i was worried about: i think the only way this API works is if SupportsSpecs() is called with an idiosyncratic and non-obvious set of inputs, and bugs are going to leak at that API boundary

internal/containerupdate/exec_updater_test.go
```diff
@@ -0,0 +1,126 @@
...
+	f := newExecFixture(t)
+
+	// No files to delete
+	err := f.ecu.UpdateContainer(f.ctx, TestDeployInfo, newReader("boop"), nil, cmds, true)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	for _, call := range f.kCli.ExecCalls {
+		if sliceutils.StringSliceStartsWith(call.Cmd, []string{"rm", "-rf"}) {
```
nicks: this test isn't super robust. it will stop testing the thing you're looking for if we change the flags to `rm`. 

a better way to test this would be to have a helper like f.getRmExecCalls() and use that helper in multiple places

or better yet, to do something like
f.assertExecCallsArg0([]string{"rm"})

internal/engine/build_and_deployer.go
```diff
@@ -87,34 +87,15 @@ func (composite *CompositeBuildAndDeployer) BuildAndDeploy(ctx context.Context,
...
-		}
-		return BuildOrder{sbad, dcbad, ibad}
-	}
-
-	if env.IsLocalCluster() && runtime == container.RuntimeDocker {
-		return BuildOrder{cbad, dcbad, ibad}
-	}
-
-	if runtime == container.RuntimeDocker {
+	if updMode == UpdateModeSynclet || (!env.IsLocalCluster() && runtime == container.RuntimeDocker) {
```
nicks: i wonder if LiveUpdateBuildAndDeployer should have a canUseSynclet method, since it's important that the logic here and the logic in containerUpdaterForSpecs are in-sync, and right now it's hard to prove that they are

internal/containerupdate/exec_updater_test.go
maiamcc: mmm good call. will change.

internal/sliceutils/sliceutils.go
```diff
@@ -44,6 +44,22 @@ func StringSliceEquals(a, b []string) bool {
 	return true
 }
 
+// StringSliceStartsWith returns true if slice A starts with (in order)
+// the elements of slice B
+func StringSliceStartsWith(a, b []string) bool {
```
landism: `a, prefix` would make it clear which one's which without having to read comments/code

internal/containerupdate/exec_updater_test.go
```diff
@@ -0,0 +1,124 @@
...
+	f := newExecFixture(t)
+
+	// No files to delete
+	err := f.ecu.UpdateContainer(f.ctx, TestDeployInfo, newReader("boop"), nil, cmds, true)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	for _, call := range f.kCli.ExecCalls {
+		if len(call.Cmd) >= 1 && call.Cmd[0] == "rm" {
```
landism: this seems like a good use for that sliceutils.StringSliceStartsWith function that someone just added!

internal/containerupdate/exec_updater_test.go
```diff
@@ -0,0 +1,124 @@
...
+	// Two files to delete
+	err = f.ecu.UpdateContainer(f.ctx, TestDeployInfo, newReader("boop"), toDelete, cmds, true)
+	if err != nil {
+		t.Fatal(err)
+	}
+	var rmCmd []string
+	for _, call := range f.kCli.ExecCalls {
+		if len(call.Cmd) >= 1 && call.Cmd[0] == "rm" {
+			rmCmd = call.Cmd
+			break
```
landism: instead of breaking I might put above this an assertion that rmCmd is nil, to catch the case where we erroneously have two rm cmds

either that or change it to rmCmds, keep a list of all of them, and then assert.Equals on the list (which could later lend itself better to like a `func (kCli *Client) GetExecCalls(cmd string) [][]string` and `rmCmds := f.kCli.GetExecCalls("rm")`)

internal/containerupdate/exec_updater.go
```diff
@@ -0,0 +1,68 @@
...
+		return fmt.Errorf("ExecUpdater does not support `restart_container()` step. If you ran Tilt " +
+			"with `--updateMode=exec`, omit this flag. If you are using a non-Docker container runtime, " +
+			"see https://github.com/windmilleng/rerun-process-wrapper for a workaround")
+	}
+
+	l := logger.Get(ctx)
+	w := l.Writer(logger.InfoLvl)
+
+	if len(filesToDelete) > 0 {
+		err := cu.kCli.Exec(ctx,
```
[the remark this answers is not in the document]
maiamcc: nah, it's still used below

maiamcc:
😬 🤞 here we goooo!

PR #1893, opened 2019-07-23
web: refactoring alertResource to just be Resource type with alerts array

mariavivanco:
Hello @jazzdan, 

This PR is the refactoring me and @landism paired on - where we deleted AlertResource and made there be one Resource type that has an alert array. 

Please review the commits made in branch post_alerts.

### DIFF
--- a/internal/engine/upper.go
+++ b/internal/engine/upper.go
@@ -196,6 +196,7 @@ var UpperReducer = store.Reducer(func(ctx context.Context, state *store.EngineSt
 		handleAnalyticsOptAction(state, action)
 	case store.AnalyticsNudgeSurfacedAction:
 		handleAnalyticsNudgeSurfacedAction(ctx, state)
+
 	case store.LogEvent:
 		// handled as a LogAction, do nothing
 
--- a/internal/engine/upper.go
+++ b/internal/engine/upper.go
@@ -196,7 +196,6 @@ var UpperReducer = store.Reducer(func(ctx context.Context, state *store.EngineSt
 		handleAnalyticsOptAction(state, action)
 	case store.AnalyticsNudgeSurfacedAction:
 		handleAnalyticsNudgeSurfacedAction(ctx, state)
-
 	case store.LogEvent:
 		// handled as a LogAction, do nothing
 
--- a/internal/hud/server/server.go
+++ b/internal/hud/server/server.go
@@ -3,6 +3,7 @@ package server
 import (
 	"encoding/json"
 	"fmt"
+	"io/ioutil"
 	"log"
 	"net/http"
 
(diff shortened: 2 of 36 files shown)


internal/hud/server/server.go
```diff
@@ -105,6 +106,28 @@ func (s *HeadsUpServer) HandleAnalyticsOpt(w http.ResponseWriter, req *http.Requ
 	s.store.Dispatch(store.AnalyticsOptAction{Opt: opt})
 }
 
+//func (s *HeadsUpServer) HandleAlertsNotification(w http.ResponseWriter, req *http.Request) {
```
jazzdan: 👀

internal/engine/upper.go
```diff
@@ -196,6 +196,7 @@ var UpperReducer = store.Reducer(func(ctx context.Context, state *store.EngineSt
 		handleAnalyticsOptAction(state, action)
 	case store.AnalyticsNudgeSurfacedAction:
 		handleAnalyticsNudgeSurfacedAction(ctx, state)
+
```
jazzdan: 👀

web/src/AppController.ts
```diff
@@ -44,6 +46,27 @@ class AppController {
       this.tryConnectCount = 0
 
       let data = JSON.parse(event.data)
+
+      data.Resources = data.Resources.map((r: any) => {
+        // function(r) { }
```
jazzdan: 👀

web/src/HUD.tsx
```diff
@@ -98,6 +102,7 @@ class HUD extends Component<HudProps, HudState> {
 
   componentDidMount() {
     this.controller.createNewSocket()
+    console.log("HUD componentDidMount")
```
jazzdan: 👀

web/src/HUD.tsx
```diff
@@ -130,6 +135,7 @@ class HUD extends Component<HudProps, HudState> {
   }
 
   render() {
+    console.log("HUD render")
```
jazzdan: 👀

web/src/__snapshots__/AlertPane.test.tsx.snap
```diff
@@ -168,6 +206,36 @@ exports[`renders one container start error 2`] = `
         </code>
       </section>
     </li>
+    <li
```
jazzdan: as we talked about in person, I think the fact that the snapshot is rendering two alerts for the container restart is a problem.

web/src/alerts.ts
```diff
@@ -0,0 +1,149 @@
...
+  titleMsg: string
+}
+
+export const PodRestartErrorType = "PodRestartError"
+export const PodStatusErrorType = "PodStatusError"
+export const ResourceCrashRebuildErrorType = "ResourceCrashRebuild"
+export const BuildFailedErrorType = "BuildError"
+export const WarningErrorType = "Warning"
+
+//these functions can be moved to wear the resources type is defined
```
jazzdan:
```suggestion
//these functions can be moved to where the resources type is defined
```

web/src/constants.ts
```diff
@@ -19,7 +19,7 @@ function podStatusIsCrash(status: string) {
   return status === podStatusError || status === podStatusCrashLoopBackOff
 }
 
-function podStatusIsError(status: string) {
+function podStatusErrorFunction(status: string) {
```
jazzdan: I would rename this to `podStatusError` because you can tell by its type that it's a function

```suggestion
function podStatusError(status: string) {
```

web/src/HUD.tsx
```diff
@@ -65,6 +67,7 @@ class HUD extends Component<HudProps, HudState> {
     this.state = {
       Message: "",
       View: {
+        alertInitialized: false,
```
jazzdan: Is this ever used/set to true?

mariabelenvivanco: oops didn't mean to keep this there!

PR #1894, opened 2019-07-23
docker: even if DOCKER_BUILDKIT=1, don't use buildkit on platforms that don't support it

nicks:
Hello @jazzdan, @maiamcc,

Please review the following commits I made in branch nicks/issue1892:

2d153b85b5344a19cceae4efeadc34e43a026917 (2019-07-23 11:39:09 -0400)
docker: even if DOCKER_BUILDKIT=1, don't use buildkit on platforms that don't support it


nicks:
this addresses https://github.com/windmilleng/tilt/issues/1892

PR #1895, opened 2019-07-23
engine: use fake DC client in upper tests

landism:
(because reproing a flaky DC upper test is really annoying when it takes 2 seconds)

### DIFF
--- a/internal/engine/upper_test.go
+++ b/internal/engine/upper_test.go
@@ -19,13 +19,12 @@ import (
 	"github.com/jonboulle/clockwork"
 	"github.com/stretchr/testify/assert"
 	"github.com/windmilleng/wmclient/pkg/analytics"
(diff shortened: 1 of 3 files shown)


internal/engine/upper_test.go
```diff
@@ -3018,6 +3016,24 @@ func (f *testFixture) setupDCFixture() (redis, server model.Manifest) {
...
 	f.WriteFile("Tiltfile", `docker_compose('docker-compose.yml')`)
 
+	var dcConfig tiltfile.DcConfig
+	err = yaml.Unmarshal(dcpc, &dcConfig)
+	if err != nil {
+		f.T().Fatal(err)
+	}
+
+	svc := dcConfig.Services["server"]
+	svc.Build.Context = f.Path()
```
maiamcc: oh! noiiiice!

PR #1907, opened 2019-07-24
doctor: print both cluster and local

nicks:
Hello @maiamcc,

Please review the following commits I made in branch nicks/doctor:

26f09f203a0c1fe5f9d184811b06faa78c0e24be (2019-07-24 15:31:00 -0400)
doctor: print both cluster and local

### DIFF
--- a/internal/cli/doctor.go
+++ b/internal/cli/doctor.go
@@ -32,24 +32,54 @@ func (c *doctorCmd) run(ctx context.Context, args []string) error {
 	ctx, cancel := context.WithTimeout(ctx, 5*time.Second)
 	defer cancel()
 
-	fmt.Println("---")
-	fmt.Println("Docker")
+	clusterDocker, clusterDockerErr := wireDockerClusterClient(ctx)
+	localDocker, localDockerErr := wireDockerLocalClient(ctx)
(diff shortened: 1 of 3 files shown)


nicks:
current output for minikube:

```
Tilt: v0.9.3-dev, built 2019-07-24
System: linux-amd64
---
Docker
- Host: tcp://192.168.99.100:2376
- Version: 1.39
- Builder: 1
---
Docker (local)
- Host: [default]
- Version: 1.40
- Builder: 2
---
Kubernetes
- Env: minikube
- Context: minikube
- Cluster Name: minikube
- Namespace: default
- Container Runtime: docker
- Version: v1.15.0
---
Please send this info along when filing bug reports. 💗
```

internal/cli/doctor.go
```diff
@@ -35,22 +35,45 @@ func (c *doctorCmd) run(ctx context.Context, args []string) error {
 	fmt.Println("---")
 	fmt.Println("Docker")
```
maiamcc: Oh wait this should def be `Docker (cluster)` to differentiate from `Docker (local)`

nicks: done!

PR #1908, opened 2019-07-24
engine: internal representation of pod (store.pod) contains info on multiple containers [ch3006]

maiamcc:
Currently when we populate a `*store.Pod`, we bless a single container and collect info on that container ([see code](https://github.com/windmilleng/tilt/blob/f159b30e7f20a923c96981c393aab0d447c145e4/internal/engine/upper.go#L599)).

First step to enabling us to LiveUpdate multiple containers is to have info for all (relevant) containers on the store.Pod.

This PR only changes structs/data storage -- all logic continues to use info from a single blessed container, to minimize the size of this change.

### DIFF
--- a/internal/engine/pod.go
+++ b/internal/engine/pod.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"strconv"
 
+	"github.com/pkg/errors"
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/labels"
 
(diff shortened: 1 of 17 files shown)


internal/engine/pod.go
```diff
@@ -151,71 +130,96 @@ func ensureManifestTargetWithPod(state *store.EngineState, pod *v1.Pod) (*store.
...
+			Blessed:  cID == blessedCID,
+		}
+		podInfo.Containers = append(podInfo.Containers, c)
 	}
-	podInfo.ContainerInfos = cInfos
 }
 
+func getBlessedContainerID(iTargets []model.ImageTarget, pod *v1.Pod) (container.ID, error) {
+	if len(iTargets) > 0 {
+		fmt.Println("~~~ GETTING BLESSED CONTAINER ID ~~~")
```
nicks: :eyes:

internal/engine/pod.go
```diff
@@ -151,71 +130,96 @@ func ensureManifestTargetWithPod(state *store.EngineState, pod *v1.Pod) (*store.
...
+		}
+
+		forwards := PopulatePortForwards(manifest, *podInfo)
+		if len(forwards) < len(manifest.K8sTarget().PortForwards) {
+			logger.Get(ctx).Infof(
+				"WARNING: Resource %s is using port forwards, but no container ports on pod %s",
+				manifest.Name, podInfo.PodID)
+		}
+
+		fmt.Printf("Container %s (%s) is blessed? %t\n", cName, cID, cID == blessedCID)
```
nicks: :eyes:

internal/engine/pod.go
```diff
@@ -151,71 +130,96 @@ func ensureManifestTargetWithPod(state *store.EngineState, pod *v1.Pod) (*store.
...
 			continue
 		}
 
-		cID, err := k8s.ContainerIDFromContainerStatus(cStat)
+		cName := k8s.ContainerNameFromContainerStatus(cStatus)
+
+		cID, err := k8s.ContainerIDFromContainerStatus(cStatus)
 		if err != nil {
 			logger.Get(ctx).Debugf("Error parsing container ID: %v", err)
 			return
```
nicks: should this be a `continue` rather than a `return`? 

alternatively, it might be worth moving this to a helper function so that it's harder to make this mistake

internal/engine/pod.go
```diff
@@ -151,71 +130,96 @@ func ensureManifestTargetWithPod(state *store.EngineState, pod *v1.Pod) (*store.
...
 }
 
+func getBlessedContainerID(iTargets []model.ImageTarget, pod *v1.Pod) (container.ID, error) {
+	if len(iTargets) > 0 {
+		fmt.Println("~~~ GETTING BLESSED CONTAINER ID ~~~")
+		// Get status of (first) container matching (an) image we built for this manifest.
+		for _, iTarget := range iTargets {
+			cStatus, err := k8s.ContainerMatching(pod, container.NameSelector(iTarget.DeploymentRef))
+			if err != nil {
+				fmt.Println("~~~ Something went wrong!")
```
nicks: :eyes:

internal/engine/pod.go
```diff
@@ -151,71 +130,96 @@ func ensureManifestTargetWithPod(state *store.EngineState, pod *v1.Pod) (*store.
...
+				return "", errors.Wrapf(err, "Error matching container for target: %s",
+					iTarget.DeploymentRef.String())
+			}
+			if cStatus.Name != "" {
+				fmt.Println("~~~ Got it!", cStatus.ContainerID)
```
nicks:
~~~

PR #1911, opened 2019-07-25
build: remove iterative image builds

nicks:
Hello @maiamcc,

Please review the following commits I made in branch nicks/imagebuild:

af0bfd16e440a7250a53124ec89ae19c7d24b7d7 (2019-07-25 14:32:01 -0400)
build: remove iterative image builds
This was only used in fast build. This was too finicky and not very successful.
Removing this allows us to simplify the state passed between builds.

### DIFF
--- a/internal/engine/image_and_cache_builder.go
+++ b/internal/engine/image_and_cache_builder.go
@@ -56,43 +56,18 @@ func (icb *imageAndCacheBuilder) Build(ctx context.Context, iTarget model.ImageT
 
 		go icb.maybeCreateCacheFrom(ctx, cacheInputs, ref, state, iTarget, cacheRef)
 	case model.FastBuild:
-		if !state.HasImage() || icb.updateMode == UpdateModeNaive {
-			// No existing image to build off of, need to build from scratch
-			ps.StartPipelineStep(ctx, "Building from scratch: [%s]", userFacingRefName)
-			defer ps.EndPipelineStep(ctx)
+		ps.StartPipelineStep(ctx, "Building from scratch: [%s]", userFacingRefName)
+		defer ps.EndPipelineStep(ctx)
 
-			df := icb.baseDockerfile(bd, cacheRef, iTarget.CachePaths())
-			runs := bd.Runs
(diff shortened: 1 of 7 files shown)


internal/engine/image_and_cache_builder.go
```diff
@@ -56,43 +56,19 @@ func (icb *imageAndCacheBuilder) Build(ctx context.Context, iTarget model.ImageT
...
-				return nil, err
-			}
-			n = ref
+		// Build the image from scratch.
+		ps.StartPipelineStep(ctx, "Building from scratch: [%s]", userFacingRefName)
+		defer ps.EndPipelineStep(ctx)
+
+		df := icb.baseDockerfile(bd, cacheRef, iTarget.CachePaths())
+		runs := bd.Runs
+		ref, err := icb.ib.BuildImageFromScratch(ctx, ps, refToBuild, df, bd.Syncs, ignore.CreateBuildContextFilter(iTarget), runs, bd.Entrypoint)
```
maiamcc: nbd but we could now `s/BuildImageFromScratch/BuildImage`

internal/engine/image_and_cache_builder.go
```diff
@@ -56,43 +56,19 @@ func (icb *imageAndCacheBuilder) Build(ctx context.Context, iTarget model.ImageT
...
-			ps.StartPipelineStep(ctx, "Building from existing: [%s]", userFacingRefName)
-			defer ps.EndPipelineStep(ctx)
-
-			runs := bd.Runs
-			ref, err := icb.ib.BuildImageFromExisting(ctx, ps, state.LastResult.Image, cf, ignore.CreateBuildContextFilter(iTarget), runs)
-			if err != nil {
-				return nil, err
-			}
-			n = ref
+		// Build the image from scratch.
```
maiamcc: this comment doesn't make sense anymore

internal/engine/image_and_cache_builder.go
nicks: good point! done

internal/engine/image_and_cache_builder.go
nicks: removed

PR #1917, opened 2019-07-25
tiltfile: the helm() command should include requirements

nicks:
Hello @jazzdan,

Please review the following commits I made in branch nicks/issue1913:

272885b23e9721f50e977b9472128fe4521bf2f7 (2019-07-25 19:37:43 -0400)
tiltfile: the helm() command should include requirements
Fixes https://github.com/windmilleng/tilt/issues/1913


nicks:
i also considered linking in helm (rather than shelling out). The Helm Go API isn't that complicated, and might be easier to work with than the string-parsing we're doing here. But it might be worth preserving the ability to upgrade helm independently.

PR #1920, opened 2019-07-26
Updated: web: refactoring alertResource to just be Resource type with alerts array

mariavivanco:
Hello @jazzdan,

Fixed the previous refactoring PR that had wrong behavior - where on a resource's Alert Pane, all of the other resources alerts were shown alongside of the resource's alerts. This is the updated PR. 

Previous PR message:
This PR is the refactoring me and @landism paired on - where we deleted AlertResource and made there be one Resource type that has an alert array.

### DIFF
--- a/web/src/AlertPane.test.tsx
+++ b/web/src/AlertPane.test.tsx
@@ -1,8 +1,8 @@
 import React from "react"
 import AlertPane from "./AlertPane"
 import renderer from "react-test-renderer"
-import {oneResourceUnrecognizedError} from "./testdata.test"
-import {Resource, ResourceInfo, TriggerMode} from "./types";
+import { oneResourceUnrecognizedError } from "./testdata.test"
+import { Resource, ResourceInfo, TriggerMode } from "./types"
 import {
   Alert,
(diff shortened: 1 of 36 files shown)


web/src/HUD.tsx
```diff
@@ -19,7 +19,7 @@ import AlertPane from "./AlertPane"
 import PreviewList from "./PreviewList"
 import AnalyticsNudge from "./AnalyticsNudge"
 import NotFound from "./NotFound"
-import { getResourceAlerts, numberOfAlerts } from "./alerts"
+import { getResourceAlerts, hasAlert, numberOfAlerts } from "./alerts"
```
jazzdan: I think `hasAlert` is imported here but not used?

web/src/AlertPane.test.tsx
```diff
@@ -18,58 +18,8 @@ beforeEach(() => {
 })
 
 it("renders no errors", () => {
-  let resources: Array<Partial<Resource>> = [
```
jazzdan: Hmm why did we delete these tests?

PR #1922, opened 2019-07-26
store/engine: BuildResult can have multiple container IDs

maiamcc:
Hello @dbentley, @nicks,

Please review the following commits I made in branch maiamcc/build-result-multi-container:

a86e80e3a1bb62a42644c8378e8836e6d11b7b17 (2019-07-26 15:26:13 -0400)
store/engine: BuildResult can have multiple container IDs


maiamcc:
interesting, say more? We need some way to tell the engine state what containers to expect (i.e. if container X crashes, we've lost state -- needs to know which containers have extra state on them, i.e. which containers we built on top of.)

PR #1923, opened 2019-07-26
Errors have button to get Alert URL

hyu:
Hidden under a feature flag
(Add `enable_feature("team_alerts")` to your Tiltfile to see this WIP feature)

We were using `titleMsg` and `header` for the little bit of additional meta text we send along with errors. I standardized to the latter.

2019-07-26 - Alert button to get link

### DIFF
--- a/internal/hud/server/server.go
+++ b/internal/hud/server/server.go
@@ -1,6 +1,7 @@
 package server
 
 import (
+	"context"
 	"encoding/json"
 	"fmt"
 	"log"
@@ -9,9 +10,6 @@ import (
 	"github.com/gorilla/mux"
 	_ "github.com/gorilla/websocket"
 	"github.com/windmilleng/wmclient/pkg/analytics"
-
-	"context"
-
 	tiltanalytics "github.com/windmilleng/tilt/internal/analytics"
 	"github.com/windmilleng/tilt/internal/assets"
 	"github.com/windmilleng/tilt/internal/hud/webview"
--- a/internal/hud/server/server.go
+++ b/internal/hud/server/server.go
@@ -265,18 +265,19 @@ func templateAlertURL(id tft.AlertID) string {
 }
 
 type tsAlert struct {
-	AlertType string `json:"alertType"`
-	Msg       string `json:"msg"`
-	Timestamp string `json:"timestamp"`
-	TitleMsg  string `json:"titleMsg"`
+	AlertType    string `json:"alertType"`
+	Header       string `json:"header"`
+	Msg          string `json:"msg"`
+	Timestamp    string `json:"timestamp"`
+	ResourceName string `json:"resourceName"`
 }
 
 func tsAlertToBackendAlert(alert tsAlert) tft.Alert {
 	return tft.Alert{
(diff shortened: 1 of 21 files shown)


internal/hud/server/server.go
```diff
@@ -10,15 +10,20 @@ import (
 	_ "github.com/gorilla/websocket"
 	"github.com/windmilleng/wmclient/pkg/analytics"
 
+	"context"
```
jazzdan: This should move up with the other built in packages

web/src/HUD.tsx
```diff
@@ -130,20 +137,51 @@ class HUD extends Component<HudProps, HudState> {
...
+  }
+
   render() {
     let view = this.state.View
     let sailEnabled = view && view.SailEnabled ? view.SailEnabled : false
     let sailUrl = view && view.SailURL ? view.SailURL : ""
     let needsNudge = view ? view.NeedsAnalyticsNudge : false
     let message = this.state.Message
     let resources = (view && view.Resources) || []
+    let featureFlags = (view && view.FeatureFlags) || {}
```
jazzdan: I think this variable is unused?

web/src/__snapshots__/AlertPane.test.tsx.snap
```diff
@@ -15,14 +15,14 @@ exports[`renders multiple lines of a crash log 1`] = `
         <p>
           Pod crashed
         </p>
-        <p>
+        <time>
```
landism: `<time>` inside `<time>` seems weird, as does a `<time>` with no `dateTime` attribute. Can we just get rid of the outer one? Was the `<p>` screwing things up?

internal/tft/client/client.go
```diff
@@ -0,0 +1,61 @@
...
+
+	if resp.StatusCode != http.StatusOK {
+		return "", fmt.Errorf("Expected 200 response code from backend, got %d", resp.StatusCode)
+	}
+
+	body, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return "", err
+	}
+	fmt.Println(string(body))
```
landism: 👀

internal/hud/server/server.go
```diff
@@ -219,3 +227,62 @@ func MaybeSendToTriggerQueue(st store.RStore, name string) error {
...
+
+	decoder := json.NewDecoder(req.Body)
+	var alert tsAlert
+	err := decoder.Decode(&alert)
+	if err != nil {
+		http.Error(w, fmt.Sprintf("error decoding request: %v", err), http.StatusBadRequest)
+		return
+	}
+
+	ctx := context.TODO()
```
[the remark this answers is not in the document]
jazzdan: I don't think so, unless we make context part of the wire tree which seems weird? We should definitely have a timeout on this though, I'll do that.

web/src/__snapshots__/AlertPane.test.tsx.snap
hyu: Good catch! When we added relative time we neglected to remove this wrapping HTML.

PR #1924, opened 2019-07-26
engine: fix regression from #1908 that caused spam about 'no container ports' when trying to port fwd (tho port forwarding still worked)

maiamcc:
Hello @nicks, @dbentley,

Please review the following commits I made in branch maiamcc/fix-no-container-ports-spam:

bbe6e5397090628f20900071fafe59a26a3266a3 (2019-07-26 17:17:38 -0400)
engine: fix regression from #1908 that caused spam about 'no container ports' when trying to port fwd (tho port forwarding still worked)

### DIFF
--- a/internal/engine/pod.go
+++ b/internal/engine/pod.go
@@ -46,6 +46,12 @@ func handlePodChangeAction(ctx context.Context, state *store.EngineState, pod *v
 		return
 	}
 
+	fwdsValid := portForwardsAreValid(manifest, *podInfo)
+	if !fwdsValid {
+		logger.Get(ctx).Infof(
+			"WARNING: Resource %s is using port forwards, but no container ports on pod %s",
+			manifest.Name, podInfo.PodID)
+	}
 	checkForPodCrash(ctx, state, ms, *podInfo)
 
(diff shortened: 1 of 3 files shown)


maiamcc:
fix regression from #1908

internal/engine/pod.go
```diff
@@ -46,6 +46,12 @@ func handlePodChangeAction(ctx context.Context, state *store.EngineState, pod *v
 		return
 	}
 
+	fwdsValid := portForwardsAreValid(manifest, *podInfo)
```
nicks: it seems very bad for a reducer to be emitting warnings, because this means that the warning will be emitted whenever there's a state change. can you leave a todo for me about this?

maiamcc: sure -- or we could move that warning to `PortForwardController.OnChange`?

maiamcc:
> also: unit tests?

All the behavior of `populatePortForwards` is tested in `portforwardcontroller_test.go`. unless you just want me to write a test that we output this particular warning?

nicks:
ya, or test the new function

PR #1926, opened 2019-07-26
engine: LiveUpdateBaD runs updates on all containers it knows about [ch3021]

maiamcc:
Integration test to follow but wanted some eyes on this first


internal/engine/live_update_build_and_deployer.go
```diff
@@ -159,20 +159,22 @@ func (lubad *LiveUpdateBuildAndDeployer) buildAndDeploy(ctx context.Context, cu
...
 		}
 	}
 
-	err = cu.UpdateContainer(ctx, cInfo, pr, build.PathMappingsToContainerPaths(toRemove), boiledSteps, hotReload)
-	if err != nil {
-		if build.IsUserBuildFailure(err) {
-			return WrapDontFallBackError(err)
+	for _, cInfo := range state.RunningContainers {
+		err = cu.UpdateContainer(ctx, cInfo, pr, build.PathMappingsToContainerPaths(toRemove), boiledSteps, hotReload)
+		if err != nil {
```
nicks: i think we need to keep updating all the containers so they're in a consistent state?

i feel like when there are multiple containers, we probably need a much more aggressive fallback strategy than when there's one container

maiamcc: if one update fails, they definitionally won't be in a consistent state -- and the effect here would be that ALL the targets for this manifest fall back to the Image Build and Deployer. (idk what happens if we succeed at one container update and fail at the next, fall back to ibad, and THAT fails... but this is all i can think of right now? any other ideas?)

nicks: you said "ALL the targets for this manifest fall back"

did you see the line "WrapDontFallBackError" two lines below this? i'm 90% sure that doesn't fall back

maiamcc: okay, mulled this over some, have some thoughts. (first of all, we should rename that function -- it means that a user's _run_ step failed. but whatever.) lmk what you think:

1. a container update fails, it's NOT a `UserRunFailure` -- e.g. we couldn't connect to the synclet or something. We fall back to image build. Doesn't matter if any container updates succeeded previously.
2. a container update fails with a `UserRunFailure` (e.g. `go build` failed b/c syntax error), run the container update on all remaining containers -- if there's a syntax error and all of the updates fail, we at least want the same fs state on all the containers
2a. if all the updates fail with `UserRunFailure`, great, don't fall back. (We can assume that it's the same failure -- which is probably good enough for now)
2b. if any update _succeeds_ (or exits with `!UserRunFailure`), we have potentially disparate state, fall back to image build.

internal/engine/build_and_deployer_test.go
```diff
@@ -836,6 +837,18 @@ func (f *bdFixture) assertContainerRestarts(count int) {
 		"checking for expected # of container restarts")
 }
 
+// Total number of restarts, regardless of which container.
+func (f *bdFixture) assertTotalContainerRestarts(count int) {
+	// Ensure that MagicTestContainerID was the only container id that saw
+	// restarts, and that it saw the right number of restarts.
+	expected := map[string]int{}
```
nicks: this appears to be a dead assignment? am i missing something?

internal/engine/live_update_build_and_deployer.go
```diff
@@ -159,20 +160,40 @@ func (lubad *LiveUpdateBuildAndDeployer) buildAndDeploy(ctx context.Context, cu
...
 			l.Infof("- %s", pm.PrettyStr())
 		}
 	}
 
-	err = cu.UpdateContainer(ctx, cInfo, pr, build.PathMappingsToContainerPaths(toRemove), boiledSteps, hotReload)
-	if err != nil {
-		if build.IsUserBuildFailure(err) {
-			return WrapDontFallBackError(err)
+	var lastUserBuildFailure error
+	for _, cInfo := range state.RunningContainers {
```
nicks: what does this code currently do if there are two different images running in two different containers? can you add a test for this case?

I'm pretty sure that the old code worked correctly, since it would just fall back to the image build codepath if there was more than one container

internal/engine/live_update_build_and_deployer.go
```diff
@@ -159,20 +160,40 @@ func (lubad *LiveUpdateBuildAndDeployer) buildAndDeploy(ctx context.Context, cu
...
+		err = cu.UpdateContainer(ctx, cInfo, pr, build.PathMappingsToContainerPaths(toRemove), boiledSteps, hotReload)
+		if err != nil {
+			if build.IsUserRunFailure(err) {
+				// Keep running updates -- we want all containers to have the same files on them
+				// even if the Runs don't succeed
+				lastUserBuildFailure = err
+				logger.Get(ctx).Infof("  → FAILED TO UPDATE CONTAINER %s with user run error: %v", cInfo.ContainerID, err)
```
nicks: "user run error" seems like a weird thing to say to the user. This is something they expect to happen frequently in normal operation, right?

maiamcc:
bump! @nicks @dbentley this is ready for review (lmk if i should loop in someone else)

nicks:
@maiamcc did you see my last round of review comments? i didn't see them addressed

nicks:
relatedly, it seems like this PR is spiraling a bit out of control...concerned that it's introducing lots of new bugs due to the size and complexity.

is there a way we could split this into smaller testable pieces? I can suggest a few options if you're not sure how.

maiamcc:
oh whoops, @nicks i didn't get notified -_- sorry, will look into them tomorrow AM!

internal/engine/live_update_build_and_deployer.go
maiamcc: you mean if the user is trying to update their app img and their sidecar img at the same time, or it tilt things that c1 and c2 are both running `appimage:tilt-123` but in fact, one is running that image and another is running `foobar:tilt-456`? or something else?

nicks: tilt should behave correctly in both cases. it might make sense to add integration tests for them first

maiamcc: @nicks bump ☝️

internal/engine/build_and_deployer_test.go
maiamcc: It gets called when you want to assert total number of restarts rather than asserting X restarts of `MagicTestContainerID`. I'm not in love with this naming/distinction but i didn't want to rewrite all the tests that use the latter logic.

internal/engine/live_update_build_and_deployer.go
maiamcc: So the image target is responsible for any image with name=`appimage` and won't differentiate between in-place updating `appimage:123` and `appimage:456` (see: [we match running containers by name, not name+tag](https://github.com/windmilleng/tilt/blob/3403400979e5db559280dd7782088e040ee69c00/internal/store/build_result.go#L261)). This seems like reasonable behavior to me, but do you think it should be different? (regardless, I'd probs write that test  for `RunningContainersForTarget` rather than for this func)

nicks:
:raised_hands:

PR #1932, opened 2019-07-29
store: Split LiveUpdate container ids from DockerCompose container ids

nicks:
Hello @maiamcc,

Please review the following commits I made in branch nicks/liveupdated:

8317fa5bd419ed57e824b718e5e1cf9f4e63de49 (2019-07-29 16:35:02 -0400)
store: Split LiveUpdate container ids from DockerCompose container ids
These should have always been separate fields, since they trigger very different
behavior. Now that multiple containers can be updated with live_update, but
docker compose can only bring up one container per service, it seems like a good
excuse to fix this.

### DIFF
--- a/internal/engine/upper_test.go
+++ b/internal/engine/upper_test.go
@@ -125,7 +125,11 @@ type fakeBuildAndDeployer struct {
 
 	// Set this to simulate a container update that returns the container ID
 	// it updated.
-	nextBuildContainer container.ID
+	nextLiveUpdateContainerID container.ID
+
+	// Inject the container ID of the container started by Docker Compose.
+	// If not set, we will auto-generate an ID.
+	nextDockerComposeContainerID container.ID
 
 	nextDeployID model.DeployID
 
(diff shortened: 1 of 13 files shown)


internal/store/engine_state.go
```diff
@@ -250,8 +250,10 @@ type ManifestState struct {
 	// The last `BuildHistoryLimit` builds. The most recent build is first in the slice.
 	BuildHistory []model.BuildRecord
 
-	// If the pod isn't running this container then it's possible we're running stale code
-	ExpectedContainerID container.ID
+	// If a container has been updated in-place, then its contents have diverged
+	// from the image it's built on.
+	LiveUpdatedContainerID container.ID
```
maiamcc: i wonder if this is still usefully called "ExpectedContainerID", since its use in code is "we expect to see container(s) X: if we don't, something is wrong."

I might comment this field like:
> The container ID that we've run a LiveUpdate on, if any. Its contents have diverged from the image it's built on. If this container doesn't appear on the pod, we've lost that state and need to rebuild.

internal/engine/upper_test.go
```diff
@@ -128,7 +128,8 @@ type fakeBuildAndDeployer struct {
 
 	// Set this to simulate a container update that returns the container ID
 	// it updated.
-	nextBuildContainer container.ID
+	nextLiveUpdateContainerID    container.ID
+	nextDockerComposeContainerID container.ID
```
maiamcc: `nextDockerComposeContainerID` should go in a different paragraph with a different comment

internal/store/build_result.go
```diff
@@ -21,26 +21,38 @@ type BuildResult struct {
 	// The tag is derived from a content-addressable digest.
 	Image reference.NamedTagged
 
-	// If this build was a container build, containerIDs we built on top of
-	ContainerIDs []container.ID
+	// When we `docker-compose up` a service, we wait synchronously for the container to start.
```
maiamcc: this and the comment below don't describe WHAT the field represents, which imo is more helpful.

> The ID of the docker compose container we deployed to (when we `docker-compose up`...)

> IDs of containers we LiveUpdated (i.e. containers whose contents have diverged...)

internal/store/engine_state.go
nicks: description added.

The name "ExpectedContainerID" is going to be incorrect once we have multiple container IDs, some of which may not have been live updated yet.

internal/store/build_result.go
nicks: done

internal/engine/upper_test.go
nicks: done

internal/store/engine_state.go
maiamcc: I mean, container IDs will only get added to this list once they've been live updated. But sure.

PR #1934, opened 2019-07-30
engine: factor out a PodBuilder for creating fake pods

nicks:
Hello @landism, @maiamcc,

Please review the following commits I made in branch nicks/podbuilder2:

3b87839785dab842df0382e685ade1acd4efa58e (2019-07-30 11:11:03 -0400)
engine: factor out a PodBuilder for creating fake pods

### DIFF
--- a/internal/testutils/podbuilder/podbuilder.go
+++ b/internal/testutils/podbuilder/podbuilder.go
@@ -1,0 +1,207 @@
+package podbuilder
+
+import (
+	"fmt"
+	"testing"
+	"time"
+
+	"github.com/docker/distribution/reference"
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/validation"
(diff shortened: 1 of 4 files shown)


internal/testutils/podbuilder/podbuilder.go
```diff
@@ -0,0 +1,207 @@
...
+	b.podID = podID
+	return b
+}
+
+func (b PodBuilder) WithPhase(phase string) PodBuilder {
+	b.phase = phase
+	return b
+}
+
+func (b PodBuilder) WithImage(image string, index int) PodBuilder {
```
maiamcc: maybe call this `WithImageAtIndex` and a separate func to set the image (/container id) for Container[0], since imagine that'll be the majority of invocations of this function, and the extra parameter for index is kinda confusing

nicks:
will address comments in a follow-up pr!

PR #1953, opened 2019-07-31
engine: Handle pod crashes when there are multiple containers

nicks:
Hello @maiamcc,

Please review the following commits I made in branch nicks/ch2905/crashes:

5b59a49d352acf630b5fe7c23f0e294c9c8be8e6 (2019-07-31 12:14:03 -0400)
engine: Handle pod crashes when there are multiple containers

6b9f39db6434f83ad7c32b59d5d383cc03f1242e (2019-07-30 18:49:10 -0400)
podbuilder: by default, use the image ref name from the YAML rather than making one up

### DIFF
--- a/internal/engine/buildcontroller_test.go
+++ b/internal/engine/buildcontroller_test.go
@@ -11,6 +11,7 @@ import (
 	v1 "k8s.io/api/core/v1"
 
 	"github.com/windmilleng/tilt/internal/hud/server"
+	"github.com/windmilleng/tilt/internal/k8s/testyaml"
 	"github.com/windmilleng/tilt/internal/testutils/manifestbuilder"
 	"github.com/windmilleng/tilt/internal/testutils/podbuilder"
 
@@ -161,7 +162,7 @@ func TestBuildControllerTwoContainers(t *testing.T) {
 	assert.Equal(t, "pod-id", c0.PodID.String(), "pod ID for cInfo at index 0")
 	assert.Equal(t, "pod-id", c1.PodID.String(), "pod ID for cInfo at index 1")
 
-	assert.Equal(t, podbuilder.FakeContainerID, c0.ContainerID, "container ID for cInfo at index 0")
+	assert.Equal(t, podbuilder.FakeContainerID(), c0.ContainerID, "container ID for cInfo at index 0")
 	assert.Equal(t, "cID-same-image", c1.ContainerID.String(), "container ID for cInfo at index 1")
 
 	assert.Equal(t, "sancho", c0.ContainerName.String(), "container name for cInfo at index 0")
@@ -218,7 +219,7 @@ func TestBuildControllerCrashRebuild(t *testing.T) {
 	assert.Equal(t, []string{}, call.oneState().FilesChanged())
 	f.waitForCompletedBuildCount(1)
 
-	f.b.nextLiveUpdateContainerID = podbuilder.FakeContainerID
+	f.b.nextLiveUpdateContainerIDs = []container.ID{podbuilder.FakeContainerID()}
 	f.podEvent(f.testPod("pod-id", manifest, "Running", time.Now()))
 	f.fsWatcher.events <- watch.NewFileEvent(f.JoinPath("main.go"))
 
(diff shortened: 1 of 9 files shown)


internal/engine/buildcontroller_test.go
```diff
@@ -245,6 +246,51 @@ func TestBuildControllerCrashRebuild(t *testing.T) {
...
+		Build())
+	f.fsWatcher.events <- watch.NewFileEvent(f.JoinPath("main.go"))
+
+	call = f.nextCall()
+	f.waitForCompletedBuildCount(2)
+	f.withManifestState("sancho", func(ms store.ManifestState) {
+		assert.Equal(t, 2, len(ms.LiveUpdatedContainerIDs))
+	})
+
+	// Restart one container id but not the other.
```
maiamcc: i don't think you mean "restart"...? maybe:

> // Simulate pod event where one of the containers has been replaced

?

internal/engine/pod.go
```diff
@@ -218,21 +218,34 @@ func getBlessedContainerID(iTargets []model.ImageTarget, pod *v1.Pod) (container
 	}
 	return "", nil
 }
-func checkForPodCrash(ctx context.Context, state *store.EngineState, ms *store.ManifestState, podInfo store.Pod) {
+
+func checkForPodCrash(ctx context.Context, state *store.EngineState, mt *store.ManifestTarget) {
```
maiamcc: nit: we're not JUST checking for pod crash now -- it's pod-or-container-crash. idk what i'd rename it to tho

internal/engine/upper.go
```diff
@@ -309,6 +308,24 @@ func handleBuildCompleted(ctx context.Context, engineState *store.EngineState, c
 		}
 	}
 
+	// Track the container ids that have been live-updated whether the
+	// build succeeds or fails.
```
maiamcc: ...why would we track a live-updated CID if the build failed?

internal/store/build_result.go
```diff
@@ -101,27 +110,6 @@ func (set BuildResultSet) OneAndOnlyLiveUpdatedContainerID() container.ID {
 	return id
 }
 
-// NOTE(maia): this is used only to populate ms.LiveUpdatedContainerID, and will
-// be removed in http://bit.ly/2LFAPDb when we implement crash detection
-// for multiple containers
-func (set BuildResultSet) OneLiveUpdatedContainerIDForOldBehavior() container.ID {
```
maiamcc: 🙌

internal/engine/buildcontroller_test.go
```diff
@@ -245,6 +246,51 @@ func TestBuildControllerCrashRebuild(t *testing.T) {
 	f.assertAllBuildsConsumed()
 }
 
+func TestCrashRebuildTwoContainersOneImage(t *testing.T) {
```
maiamcc: is it worth testing the behavior when it's two containers on a single pod, each with a different image? should work the same, but maybe good for the sake of coverage.

internal/engine/buildcontroller_test.go
nicks: AFAICT, "restart" is consistent with the terminology that Kubernetes uses

if you kill the main container process, Kubernetes creates a new container ID and increments the restartCount field to 1

internal/engine/pod.go
nicks: changed to checkForContainerCrash

internal/engine/upper.go
nicks: This is the distinction we spent a long time talking about on the other PR. If a user Run step fails on a syntax error, that is normal and expected. The container files have still diverged

internal/engine/buildcontroller_test.go
nicks: i thought that didn't work yet?

internal/engine/buildcontroller_test.go
maiamcc: Ah. That makes sense -- but fwiw I read "restart" like `restart_container`, i.e. it's the same container, we just re-exec the entrypoint. So i might consider replacing cuz that word has a different meaning in our codebase. but nbd

internal/engine/buildcontroller_test.go
maiamcc: OH! yeah that building behavior doesn't work, you're right -- was imaging that you could directly set up the pod/build-result state. but that's probably not worth it. leave a todo for that test tho, and i'll fill it in

internal/engine/upper.go
maiamcc: oh i see! well shoot, [we don't return a result in that case :/ ](https://github.com/windmilleng/tilt/blob/96e48fcb02bced80fe58c0e975f1d52655fe49e9/internal/engine/live_update_build_and_deployer.go#L205) -- can you make a ticket for this?

nicks: filed as https://app.clubhouse.io/windmill/story/3094/live-update-should-still-work-if-run-steps-fail

internal/engine/buildcontroller_test.go
nicks: filed as https://app.clubhouse.io/windmill/story/3095/add-tests-for-live-update-two-containers-with-different-images

PR #1964, opened 2019-08-01
store: remove Blessed containers

nicks:
Hello @maiamcc,

Please review the following commits I made in branch nicks/ch3091/blessed:

73b4094384f416acb1c19fa012fe643dc71d9ee4 (2019-08-01 12:51:30 -0400)
store: remove Blessed containers

5fd2b168ab94c1c8f2eef4cd00f5c360d1db754a (2019-07-31 18:15:14 -0400)
store: make some Pod methods work with multiple containers

### DIFF
--- a/internal/demo/script.go
+++ b/internal/demo/script.go
@@ -93,7 +93,7 @@ func (m *podMonitor) OnChange(ctx context.Context, st store.RStore) {
 			m.healthy = false
 		}
 
-		if pod.ContainerRestarts > 0 {
+		if pod.AllContainerRestarts() > 0 {
 			m.hasPodRestart = true
 			m.healthy = false
 		}
--- a/internal/engine/pod.go
+++ b/internal/engine/pod.go
@@ -39,7 +39,8 @@ func handlePodChangeAction(ctx context.Context, state *store.EngineState, pod *v
 
 	defer prunePods(ms)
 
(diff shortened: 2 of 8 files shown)


internal/engine/podlogmanager.go
```diff
@@ -121,6 +110,29 @@ func (m *PodLogManager) diff(ctx context.Context, st store.RStore) (setup []PodL
...
 }
 
+func (m *PodLogManager) shouldWatchPodLogs(pod store.Pod) bool {
+	if pod.PodID == "" || len(pod.Containers) == 0 {
+		return false
+	}
+
+	// If an ID or name for the containers hasn't been created yet,
+	// weird things will happen when we try to store them in the `m.watches` map
+	for _, container := range pod.Containers {
```
maiamcc: should we watch all the containers we CAN watch rather than bailing on the whole pod at the first un-populated container? (idk the right answer, just seems vaguely weird)

nicks: this is an edge case that should only last for very short period of time (like sub-millisecond). It isn't worth optimizing for. better to just do the safe thing and not worry about it.

nicks: (left a comment about this)

maiamcc: sgtm!

PR #1968, opened 2019-08-01
if we touched containers in LiveUpdate, return/record touched container IDs even if the BuildAndDeploy call fails [ch3105]

maiamcc:
if a user's Run step(s) fails, we don't fall back to an image build (b/c
        presumably the same thing will fail on the image build)
        however, any sync's have run successfully, which means
        that state on the running container has diverged, and we need
        to monitor it for crashes/do a crash rebuild if that container
        disappears. that means that we need to get that container's ID
        onto the manifest state, regardless of the fact that the
        BuildAndDeploy call as a whole has failed.

### DIFF
--- a/internal/engine/upper_test.go
+++ b/internal/engine/upper_test.go
@@ -194,12 +194,6 @@ func (b *fakeBuildAndDeployer) BuildAndDeploy(ctx context.Context, st store.RSto
 		logger.Get(ctx).Infof("fake building %s", ids)
 	}()
 
-	err := b.nextBuildFailure
-	if err != nil {
-		b.nextBuildFailure = nil
-		return store.BuildResultSet{}, err
-	}
-
 	dID := podbuilder.FakeDeployID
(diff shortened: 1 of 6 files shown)


internal/engine/upper_test.go
```diff
@@ -2299,25 +2296,6 @@ func TestDockerComposeBuildCompletedSetsStatusToUpIfSuccessful(t *testing.T) {
 	})
 }
 
-func TestDockerComposeBuildCompletedDoesntSetStatusIfNotSuccessful(t *testing.T) {
```
nicks: why did this test get deleted?

maiamcc: It isn’t testing anything in our code, all it’s testing is the behavior of the fake build and deployer. 

 If we want to enforce that, when we handle a completed build, we don’t record docker compose state if there is an error, that’s a separate question – though it seems moot since the DC BAD Only ever returns empty results along side errors

PR #1970, opened 2019-08-02
engine: can LiveUpdate multiple images for same manifest [ch3005]

maiamcc:
Hello @nicks, @jazzdan,

Please review the following commits I made in branch maiamcc/live-update-multi-image:

04f913506b05d7ea143d32d3185905de89269620 (2019-08-02 11:08:45 -0400)
fix error handling & test

4fc089e00b408a8b2a5d8cbf26dd89dfed152f74 (2019-08-01 15:14:44 -0400)
wip

e11ff97109e1de79d72640f220e583a23940c49f (2019-08-01 14:37:59 -0400)
some tests

fbc8ac330bcfc9a1cd8044d74de435defb9d70e0 (2019-08-01 12:51:43 -0400)
another crash rebuild test

1a88d75dfc413bccdeab1f6284ce97e16d4b1ce8 (2019-08-01 12:30:35 -0400)
initial implementation of multi-image live update for same pod

5d4db2b9a8d0b9d9447b745b09ad9015158023b0 (2019-08-01 12:30:02 -0400)
Red test

### DIFF
--- a/internal/engine/build_and_deployer_test.go
+++ b/internal/engine/build_and_deployer_test.go
@@ -138,6 +138,24 @@ func TestNamespaceGKE(t *testing.T) {
 	assert.Equal(t, "sancho-ns", string(f.sCli.Namespace))
 }
 
+func TestYamlManifestDeploy(t *testing.T) {
+	f := newBDFixture(t, k8s.EnvGKE, container.RuntimeDocker)
+	defer f.TearDown()
+
+	manifest := manifestbuilder.New(f, "some_yaml").
+		WithK8sYAML(testyaml.TracerYAML).Build()
+	targets := buildTargets(manifest)
+	_, err := f.bd.BuildAndDeploy(f.ctx, f.st, targets, store.BuildStateSet{})
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	assert.Equal(t, 0, f.sCli.UpdateContainerCount)
+	assert.Equal(t, 0, f.docker.BuildCount)
+	assert.Equal(t, 0, f.docker.PushCount)
+	f.assertK8sUpsertCalled(true)
+}
+
 func TestContainerBuildLocal(t *testing.T) {
 	f := newBDFixture(t, k8s.EnvDockerDesktop, container.RuntimeDocker)
 	defer f.TearDown()
--- a/internal/engine/build_and_deployer_test.go
+++ b/internal/engine/build_and_deployer_test.go
@@ -807,7 +807,6 @@ func TestLiveUpdateMultipleImagesSamePod(t *testing.T) {
 
 	// one for each container update
 	assert.Equal(t, 2, f.sCli.UpdateContainerCount)
-
 }
 
 // The API boundaries between BuildAndDeployer and the ImageBuilder aren't obvious and
--- a/internal/engine/build_and_deployer_test.go
+++ b/internal/engine/build_and_deployer_test.go
@@ -11,15 +11,16 @@ import (
 	"testing"
 	"time"
 
-	"github.com/windmilleng/tilt/internal/build"
-	"github.com/windmilleng/tilt/internal/container"
-	"github.com/windmilleng/tilt/internal/dockercompose"
-	"github.com/windmilleng/tilt/internal/ospath"
-	"github.com/windmilleng/tilt/internal/store"
-
 	"github.com/docker/docker/api/types"
 	"github.com/opencontainers/go-digest"
 	"github.com/stretchr/testify/assert"
+
+	"github.com/windmilleng/tilt/internal/build"
+	"github.com/windmilleng/tilt/internal/container"
+	"github.com/windmilleng/tilt/internal/dockercompose"
+	"github.com/windmilleng/tilt/internal/k8s/testyaml"
+	"github.com/windmilleng/tilt/internal/ospath"
+	"github.com/windmilleng/tilt/internal/store"
 
 	"github.com/windmilleng/wmclient/pkg/dirs"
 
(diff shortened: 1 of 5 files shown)


maiamcc:
alright tests should be fixed, this should be ready for review!

internal/engine/build_and_deployer_test.go
```diff
@@ -793,6 +810,167 @@ func TestLiveUpdateWithRunFailureReturnsContainerIDs(t *testing.T) {
...
+	// should NOT have run live update
+	assert.Equal(t, 0, f.sCli.UpdateContainerCount)
+
+	// expect image build (2x images) when we fall back from failed LiveUpdate
+	assert.Equal(t, 2, f.docker.BuildCount)
+	assert.Equal(t, 2, f.docker.PushCount)
+	f.assertK8sUpsertCalled(true)
+}
+
+func multiImageLiveUpdateManifestAndBuildState(f *bdFixture) (model.Manifest, store.BuildStateSet) {
```
nicks: seems weird to couple these into one function

internal/engine/live_update_build_and_deployer.go
```diff
@@ -47,86 +47,108 @@ func NewLiveUpdateBuildAndDeployer(dcu *containerupdate.DockerContainerUpdater,
...
+	if len(liveUpdateStateSet) == 0 {
+		return nil, RedirectToNextBuilderInfof("no targets for LiveUpdate found")
 	}
 
-	liveUpdateState := liveUpdateStateSet[0]
-	iTarget := liveUpdateState.iTarget
-	state := liveUpdateState.iTargetState
-	filesChanged := liveUpdateState.filesChanged
+	for i, liveUpdateState := range liveUpdateStateSet {
+		iTarget := liveUpdateState.iTarget
```
nicks: can you break this out into a separate function? all the nesting seems hard to reason about

internal/engine/live_update_build_and_deployer.go
```diff
@@ -47,86 +47,108 @@ func NewLiveUpdateBuildAndDeployer(dcu *containerupdate.DockerContainerUpdater,
...
-	var hotReload bool
+			runs = luInfo.RunSteps()
+			hotReload = !luInfo.ShouldRestart()
+		}
 
-	if fbInfo := iTarget.AnyFastBuildInfo(); !fbInfo.Empty() {
-		changedFiles, err = build.FilesToPathMappings(filesChanged, fbInfo.Syncs)
-		if err != nil {
-			return store.BuildResultSet{}, err
+		liveUpdInfos[i] = liveUpdInfo{
```
nicks: what does code this do in the case where one image has live update info and the other doesn't?

internal/engine/live_update_state_tree.go
```diff
@@ -15,27 +15,29 @@ type liveUpdateStateTree struct {
 }
 
 // Create a successful build result if the live update deploys successfully.
-func (t liveUpdateStateTree) createResultSet() store.BuildResultSet {
-	iTargetID := t.iTarget.ID()
-	state := t.iTargetState
-	res := state.LastResult
+func createResultSet(trees []liveUpdateStateTree) store.BuildResultSet {
```
nicks: it might be easier to leave the existing method as is, and have a method for merging BuildResultSets?

internal/engine/build_and_deployer_test.go
maiamcc: I mean, it's just giving me a specific handful of (pretty interconnected) test data. seems less weird than putting it in two different functions but still having the data it returns be so interconnected that the build state only makes sense for the specific manifest (cuz then the functions risk being called separately?) but also 🤷‍♀

internal/engine/live_update_build_and_deployer.go
maiamcc: that should never happen, `extractImageTargetsForLiveUpdates` [should only return targets with LiveUpdate or FastBuild](https://github.com/windmilleng/tilt/blob/d92acf8542722d6361421d19dacd58c51e7859b8/internal/engine/extractors.go#L63). I suppose I can put an extra check here tho

internal/engine/live_update_state_tree.go
maiamcc: i mean, i'd much rather call a single "resultSet" func at the end of BaD execution than call it once for each stateTree in the for loop and keep track of the results to merge later. if you're saying that i should have implemented this function by using the old func as a helper and just merging all the results at the end... yeah you're probably right 🤷‍♀ but feel like it would take as much work now to put back as it would have saved me before. unless you have a _style_ objection to this, i'd just leave it

internal/engine/live_update_build_and_deployer.go
```diff
@@ -210,6 +188,64 @@ func (lubad *LiveUpdateBuildAndDeployer) buildAndDeploy(ctx context.Context, cu
...
+		if anyMatch {
+			return liveUpdInfo{}, RedirectToNextBuilderInfof(
+				"detected change to fall_back_on file '%s'", file)
+		}
+
+		runs = luInfo.RunSteps()
+		hotReload = !luInfo.ShouldRestart()
+	} else {
+		// We should have validated this when generating the LiveUpdateStateTrees, but double check!
+		panic(fmt.Sprintf("found neither FastBuild nor LiveUpdate info on target %s, "+
```
maiamcc: this seems extreme, but at the same time, this ought to never happen, so 🤷‍♀ wdyt?

internal/engine/live_update_state_tree.go
nicks: hmmm...ok, though in general we should be trying to converge towards code that's independently testable and reusable. writing lots of bespoke functions like this one means we're going to spend lots of time deleting and rewriting them.

PR #1972, opened 2019-08-02
pod watch: delete dead pods

### LINKED ISSUE #1862: "error streaming ___ logs" spam after pod terminates
Repro:
1. In [tilt.build](https://github.com/windmilleng/tilt.build), `tilt up`
2. Wait for all the services to be green
3. Edit [deploy/serve.yaml](https://github.com/windmilleng/tilt.build/blob/master/deploy/serve.yaml): remove the 'e' from the end of blog-site's spec.selector.matchLabels
4. Wait a few minutes

Observed:
unending spam that tilt cannot find the pod:
```
Error streaming blog-site logs: pods "blog-site-84f5dd4c8c-nqcrq" not found
Error streaming blog-site logs: pods "blog-site-84f5dd4c8c-nqcrq" not found
Error streaming blog-site logs: pods "blog-site-84f5dd4c8c-nqcrq" not found
Error streaming blog-site logs: pods "blog-site-84f5dd4c8c-nqcrq" not found
Error streaming blog-site logs: pods "blog-site-84f5dd4c8c-nqcrq" not found
Error streaming blog-site logs: pods "blog-site-84f5dd4c8c-nqcrq" not found
Error streaming blog-site logs: pods "blog-site-84f5dd4c8c-nqcrq" not found
Error streaming blog-site logs: pods "blog-site-84f5dd4c8c-nqcrq" not found
Error streaming blog-site logs: pods "blog-site-84f5dd4c8c-nqcrq" not found
Error streaming blog-site logs: pods "blog-site-84f5dd4c8c-nqcrq" not found
```

Expected:
there's no more pod so there's nothing to display

landism:
fixes #1862 

### Problem

If a manifest loses its pod and doesn't get a new pod to replace it, we never remove that pod from its podset, so every time `PodLogController`'s `OnChange` gets called, it tries to watch the old pod and immediately fails and logs an error, resulting in spam.

Our normal mechanism for removing dead pods from a manifest's podset only kicks in when the podset has one more than one pod (which is normally the case because normally a pod gets deleted when a new pod gets created).

### Solution

Delete pods from podset as soon as we see they've been deleted.

### DIFF
--- a/internal/engine/pod.go
+++ b/internal/engine/pod.go
@@ -102,6 +102,11 @@ func ensureManifestTargetWithPod(state *store.EngineState, pod *v1.Pod) (*store.
 	ns := k8s.NamespaceFromPod(pod)
 	hasSynclet := sidecar.PodSpecContainsSynclet(pod.Spec)
 
+	if pod.DeletionTimestamp != nil && !pod.DeletionTimestamp.IsZero() && ms.PodSet.Pods != nil {
+		delete(ms.PodSet.Pods, podID)
+		return nil, nil
+	}
+
 	// CASE 1: We don't have a set of pods for this DeployID yet
 	if ms.PodSet.DeployID == 0 || ms.PodSet.DeployID != deployID {
 		ms.PodSet = store.PodSet{
--- a/internal/engine/pod.go
+++ b/internal/engine/pod.go
@@ -32,12 +32,12 @@ func handlePodChangeAction(ctx context.Context, state *store.EngineState, pod *v
 	}
 
 	// Update the status
-	podInfo.Deleting = pod.DeletionTimestamp != nil
+	podInfo.Deleting = pod.DeletionTimestamp != nil && !pod.DeletionTimestamp.IsZero()
 	podInfo.Phase = pod.Status.Phase
 	podInfo.Status = podStatusToString(*pod)
 	podInfo.StatusMessages = podStatusErrorMessages(*pod)
 
-	defer prunePods(ms)
+	prunePods(ms)
 
 	oldRestartTotal := podInfo.AllContainerRestarts()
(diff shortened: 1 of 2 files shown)


internal/engine/pod.go
```diff
@@ -102,6 +102,11 @@ func ensureManifestTargetWithPod(state *store.EngineState, pod *v1.Pod) (*store.
 	ns := k8s.NamespaceFromPod(pod)
 	hasSynclet := sidecar.PodSpecContainsSynclet(pod.Spec)
 
+	if pod.DeletionTimestamp != nil && !pod.DeletionTimestamp.IsZero() && ms.PodSet.Pods != nil {
```
nicks: i thought we had a separate `prunePods` function for removing pods? does this logic make more sense there?

internal/engine/pod.go
```diff
@@ -102,6 +102,11 @@ func ensureManifestTargetWithPod(state *store.EngineState, pod *v1.Pod) (*store.
...
+		delete(ms.PodSet.Pods, podID)
```
nicks: hmmm...will the short-lived pod log stuff still work if we remove the pod immediately?

landism: depends on what you mean by "short-lived pod log stuff"

The case covered by the unit test is a Job, where "short-lived" simply means it isn't in state "Running" for very long - it transitions to "Completed" or "Failed", but doesn't actually get deleted - it still shows up in `kubectl get pods`, and doesn't have a `DeletionTimestamp`.

IIUC, the case that might stop working is if a pod gets deleted, we have a chance to grab its logs while it's in the "Terminating" state, but getting logs from a "Terminating" pod is always going to be a race condition, because it might be gone by the time we try to get the logs.

PR #1973, opened 2019-08-02
containerupdater: fix kubectl exec updater to handle RunStepFailure

nicks:
Hello @maiamcc,

Please review the following commits I made in branch nicks/liveupdate:

9e16268684663664fdf5c3302a2bd6a415520873 (2019-08-02 13:14:18 -0400)
containerupdater: fix kubectl exec updater to handle RunStepFailure

### DIFF
--- a/internal/containerupdate/exec_updater_test.go
+++ b/internal/containerupdate/exec_updater_test.go
@@ -3,11 +3,14 @@ package containerupdate
 import (
 	"bytes"
 	"context"
+	"fmt"
 	"io"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
+	"k8s.io/client-go/util/exec"
 
+	"github.com/windmilleng/tilt/internal/build"
 	"github.com/windmilleng/tilt/internal/sliceutils"
 
 	"github.com/windmilleng/tilt/internal/k8s"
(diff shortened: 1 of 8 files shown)


nicks:
thinking about this, I think we need to move away from RunStepFailure as an implementation of Error

Most build implementations I've worked with are super careful about separating "user command failed" from "error", and often give them very obtuse names just to make sure people understand that distinction.

For example, here's a snippet from Bazel that calls "user command failed" a "NON_ZERO_EXIT" but often calls it a "success" case
https://github.com/bazelbuild/bazel/blob/master/src/main/java/com/google/devtools/build/lib/actions/SpawnResult.java#L36

internal/containerupdate/exec_updater_test.go
```diff
@@ -101,6 +104,16 @@ func TestUpdateContainerRunsCommands(t *testing.T) {
 	}
 }
 
+func TestUpdateContainerRunsFailure(t *testing.T) {
```
maiamcc: is it worth also testing that, if one run errors, the subsequent ones don't execute? (could write a new test, or make that happen with this test by having the FIRST exec call throw the error, and then assert on the number of exec calls).

nicks: added!

nicks:
re: result: we could probably do it quickly?

maiamcc:
I had in fact left myself a note this morning of "maybe we should rework naming for `DontFallBackError` to make it clearer what's going on/what it represents" -- so yes i would love it if we make that change! Do you want to or shall I?

nicks:
re: DontFallbackError: I think that's a separate thing, but also something we should rename. 

i've been working on this doc (more for myself than for anyone else) about how build systems work and how we might update the execution model to better capture DontFallback

https://docs.google.com/document/d/1HsoOTpdQ0BejJrsY9_iV1f3yHbuDh6ZzjwWiv3F2w8o/edit#

where "Fallback" is more like the build system notion of "this target is changed and all downstream dependencies need to be rebuilt" and "DontFallback" is like the notion of "we rebuilt this target and nothing changed, so no need to build things downstream"

PR #1977, opened 2019-08-02
Alert Link shows URL to work around Pop-up detection

hyu:
Instead of immediately opening a new window, which is often blocked by browser pop-up detection, let's show the URL first.

Screen Shot 2019-08-02 at 5 19 20 PM


jazzdan:
Looks like we have some prettier problems though.

PR #1985, opened 2019-08-05
synclet/engine: handle UserRunFailures in synclet [ch3094]

maiamcc:
Hello @nicks, @jazzdan,

This code ensure that
a. `UserRunErrors` are wrapped/surfaced correctly from the synclet, and
b. that the LiveUpdBaD can correctly read wrapped `UserRunErrors`

(also adds tests for all three container updaters surfacing `UserRunErrors` correctly)


internal/build/error.go
```diff
@@ -56,8 +56,28 @@ func (e RunStepFailure) Error() string {
 }
 
 func IsRunStepFailure(err error) bool {
-	_, ok := err.(RunStepFailure)
+	_, ok := MaybeRunStepFailure(err)
 	return ok
 }
 
+func MaybeRunStepFailure(err error) (RunStepFailure, bool) {
```
maiamcc: this seems a little clunky, is there a more elegant way to do this? (Alternately, it seems maybe overkill to recursively drill into causes rather than just checking the very first cause and leaving it at that... but also, ideally no matter how many times we wrap an error, it should still be detectable as a rsf, sooo 🤷‍♀ )

nicks:
I'm confused, how does this work with grpc boundaries? Presumably there needs to be code that converts the type in client_adapter and server_adapter

internal/build/error.go
nicks: fyi, the Go2 errors proposal has a library for this
https://godoc.org/golang.org/x/exp/errors#Is

maiamcc:
> I'm confused, how does this work with grpc boundaries? Presumably there needs to be code that converts the type in client_adapter and server_adapter

@nicks ...I'd think so too, but the tests pass?? I just assume grpc is doing ✨ magic ✨ when it relays the error across. Anyway, I'm not sure _why_ it works, but it works -- is there something different I should be doing, or anything that I should verify before I merge this?

maiamcc:
fwiw, if I put back the OLD error type check in the LiveUpdBaD (`if runFail, ok := err.(build.RunStepFailure); ok`), then `TestLiveUpdateSyncletUserRunFailureDoesntFallBack` fails -- so it is the new code that's making the difference.

nicks:
Have you tested it with a real sync let?

maiamcc:
OH DUH I understand. Let me verify.

maiamcc:
@nicks this is ready for review--i'll make the tests actually touch useful behavior after you land #1992

internal/synclet/client_adapter.go
```diff
@@ -69,6 +71,14 @@ func (s *SyncletCli) UpdateContainer(
 	for {
 		reply, err := stream.Recv()
 
+		if reply != nil && reply.FailedRunStep != nil {
+			frs := reply.FailedRunStep
+			return build.RunStepFailure{
```
nicks: this is going to leak a goroutine. from the GRPC docs:

```
NewStream creates a new Stream for the client side. This is typically called by generated code. ctx is used for the lifetime of the stream.

To ensure resources are not leaked due to the stream returned, one of the following actions must be performed:

1. Call Close on the ClientConn.
2. Cancel the context provided.
3. Call RecvMsg until a non-nil error is returned. A protobuf-generated
   client-streaming RPC, for instance, might use the helper function
   CloseAndRecv (note that CloseSend does not Recv, therefore is not
   guaranteed to release all resources).
4. Receive a non-nil, non-io.EOF error from Header or SendMsg.
```

nicks: (probably best thing to do is save this in a variable, and only return it after the EOF)

maiamcc: went with cancelling the context instead because it seemed potentially gnarly to rely on getting an EOF--what if something DIFFERENT went wrong with the stream? I assume we'd still want to return the `RunStepFailure` error...) -- but lmk if this is wrong and i will go the storing-on-a-variable route

maiamcc: ...nvm it was giving me lint errors, I just went with your suggestion.

PR #1995, opened 2019-08-06
tiltfile: disallow deploys to remote kube by default

### LINKED ISSUE #1096: Should Tilt allow deploys to non-local kubecontexts by default?
@thrawny [in Kubernetes Slack](https://kubernetes.slack.com/archives/CESBL84MV/p1549034803134300) had production downtime caused by inadvertently allowing Tilt to deploy dev configs to a production Kubernetes cluster. This is something we'd very much like to avoid happening to others.

A use case we want to support is for people to run Tiltfiles with a large number of services against a Kubernetes cluster that exists in the cloud, so that their laptop doesn’t melt. For this use case the kubecontext could be anything. However, it's my believe that comparatively few people are using Tilt in this way.

I think we should disable deploying to non-local (minikube, docker-for-mac, possibly others) kubecontexts by default. If Tilt is started with such a non-local kubecontext we could display a message like this:

```
It looks like you're running Tilt against a remote Kubernetes cluster (KUBECONTEXT_NAME_HERE). Tilt supports this, but we want to make sure you're not inadvertently deploying to prod! To enable remote Kubernetes deploys simply add `enable_remote_deploy()` to your Tiltfile.
```

Perhaps if you are using a remote kubecontext and you haven't set `enable_remote_deploy()` we still perform other local operations (docker builds, locals)?

I would love other's thoughts! @thrawny let me know if I got any details wrong.

landism:
Resolves #1096 

Brief spec / thoughts [here](https://github.com/windmilleng/tilt.specs/blob/master/safer_remote_clusters.md).

The error message is kinda wordy, and is a wall of text in the TUI:
image


internal/tiltfile/k8s.go
```diff
@@ -887,3 +888,57 @@ func newK8sObjectID(e k8s.K8sEntity) k8sObjectID {
...
+		switch val := c.(type) {
+		case starlark.String:
+			s.whitelistedK8SContexts = append(s.whitelistedK8SContexts, k8s.KubeContext(val))
+		default:
+			return nil, fmt.Errorf("allow_k8s_contexts contexts must be a string or a sequence of strings; found a %T", val)
+
+		}
+	}
+
+	s.whitelistedK8SContexts = append(s.whitelistedK8SContexts)
```
nicks: this line is a no-op?

internal/tiltfile/k8s.go
```diff
@@ -887,3 +888,57 @@ func newK8sObjectID(e k8s.K8sEntity) k8sObjectID {
...
+func (s *tiltfileState) validateK8SContext(kubeConfig *api.Config) error {
+	contextName := kubeConfig.CurrentContext
+	clusterName := kubeConfig.Contexts[contextName].Cluster
+	server := kubeConfig.Clusters[clusterName].Server
+	url, err := url.Parse(server)
+	if err != nil {
+		return errors.Wrapf(err, "running in kube context %q, which specifies kube api url %q. error parsing that url", contextName, server)
+	}
+
+	ipaddr, err := net.ResolveIPAddr("ip", url.Hostname())
```
nicks: it seems very odd for Tiltfile execution to be making DNS requests. can we just whitelist a few hard-coded hosts (127.0.0.1, localhost)?

internal/tiltfile/k8s.go
```diff
@@ -887,3 +888,57 @@ func newK8sObjectID(e k8s.K8sEntity) k8sObjectID {
...
+		return nil
+	}
+
+	for _, c := range s.whitelistedK8SContexts {
+		if c == k8s.KubeContext(contextName) {
+			return nil
+		}
+	}
+
+	return fmt.Errorf("Current kube context %q is neither whitelisted nor known to be local. If this is not the "+
```
nicks: This error message is pretty inscrutable

I like looking at Chrome error messages for inspiration
https://lifehacker.com/report-suspicious-sites-to-google-using-this-chrome-ext-1835650376

how about something like

"
Stop! X is a production cluster.
If you're sure that this is a dev cluster, add add
allow_k8s_contexts('X')
to your Tiltfile
```

No need to even give them an API link

internal/tiltfile/k8s.go
landism: Hm, it feels sloppy to me to poorly reimplement loopback detection, and I think I'd rather move the dns request outside of Tiltfile execution, but it's probably not worth the effort given some hard-coding is probably 99% good enough, so ok let's do that and not worry about it unless someone notices.

(My main hangup here is that at a former employer, dev machines were configured such that 127.0.0.1 resolved to `$USER.local`, so tools would detect that and talk to `$USER.local` instead of `localhost`. I have no idea why they did that or if that's common elsewhere, and I'm probably overindexing on that experience).

internal/tiltfile/k8s.go
landism: That's better, thanks.

Here's what it looks like now:
image

internal/tiltfile/k8s.go
nicks: fwiw, we're not trying to cover the case of "urls that J. Random Engineer has added manually", we're covering the case of "urls that Kubernetes tools will add to kubeconfig", which is probably much more limited.

PR #1996, opened 2019-08-06
tiltfile: add load() and include()

nicks:
Hello @jazzdan,

Please review the following commits I made in branch nicks/load:

fb52ea09ab175cce141d3738b7318b6c71240513 (2019-08-06 17:41:26 -0400)
tiltfile: add load() and include()

### DIFF
--- a/internal/tiltfile/include.go
+++ b/internal/tiltfile/include.go
@@ -1,0 +1,67 @@
+package tiltfile
+
+import (
+	"fmt"
+	"path/filepath"
+
+	"go.starlark.net/starlark"
+)
+
+type loadCacheEntry struct {
+	status  loadStatus
+	exports starlark.StringDict
+	err     error
+}
+
+type loadStatus int
+
+const (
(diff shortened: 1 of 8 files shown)


internal/tiltfile/include.go
```diff
@@ -0,0 +1,67 @@
...
+
+type loadStatus int
+
+const (
+	loadStatusNone loadStatus = iota
+	loadStatusExecuting
+	loadStatusDone
+)
+
+// Tiltfiles support both load() and include().
```
jazzdan: As of this PR it only supports `include()`. Is this comment forward looking?

nicks: did you see `TestLoadFunction`?

jazzdan: Hmm yes I see that now, but I was looking for `load` in the list of built-in functions https://github.com/windmilleng/tilt/pull/1996/files#diff-187cac6214111b4c1831b65825c135f8R180

PR #2003, opened 2019-08-07
integration: add tests for crash rebuilds

nicks:
Hello @maiamcc,

Please review the following commits I made in branch nicks/integration:

d148c8d761a197d9b19e5b83534396177ad56862 (2019-08-07 13:07:40 -0400)
integration: add tests for crash rebuilds

### DIFF
--- a/integration/onewatch_test.go
+++ b/integration/onewatch_test.go
@@ -17,17 +17,9 @@ func TestOneWatch(t *testing.T) {
 
 	f.TiltWatch()
 
-	// ForwardPort will fail if all the pods are not ready.
-	//
-	// For the purposes of this integration tests, we want the test to fail if the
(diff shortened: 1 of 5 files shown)


integration/onewatch_test.go
```diff
@@ -51,4 +43,19 @@ func TestOneWatch(t *testing.T) {
 	// Assert that the pods were changed in-place, and not that we
 	// created new pods.
 	assert.Equal(t, oneUpPods, twoUpPods)
+
+	// Delete the pod and make sure it got replaced with something that prints the
+	// same thing (crash rebuild).
```
maiamcc: nbd but I think this comment should go above the `runCommandSilently` call instead of here (made me think the length check was somehow deleting pods for a sec)

nicks: done!

PR #2006, opened 2019-08-07
engine: handle case where changed file matches sync for one image target but not another [ch3116]

maiamcc:
If `a.txt` is in the docker ctx for both `appimg` and `sidecarimg`, and the docker_build for `appimg` syncs `a.txt` and sidecar does not, we’ll get a “file matches no syncs” error when making path mappings for sidecar and fall back to an image build.

after this PR, we do NOT fall back to image build: instead, we can see that only the app has changed files, and do a LiveUpdate for the app but not the sidecar

### DIFF
--- a/internal/store/build_result.go
+++ b/internal/store/build_result.go
@@ -1,7 +1,6 @@
 package store
 
 import (
-	"log"
 	"sort"
 
 	"github.com/docker/distribution/reference"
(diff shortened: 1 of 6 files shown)


internal/store/build_result.go
```diff
@@ -270,7 +269,6 @@ func RunningContainersForTarget(iTarget model.ImageTarget, deployID model.Deploy
 	for _, c := range pod.Containers {
 		// Only return containers matching our image
 		if c.ImageRef == nil || iTarget.DeploymentRef.Name() != c.ImageRef.Name() {
-			log.Println("SKIPPING", iTarget.DeploymentRef, c.ImageRef)
```
nicks: :grimacing:

PR #2011, opened 2019-08-07
web: can init from static HTTP response

jazzdan:
I couldn't really test this without some sort of working snapshot frontend server. I tested this by changing the URL that it loads from to `localhost:10350/api/view` manually.

### DIFF
--- a/web/src/AppController.ts
+++ b/web/src/AppController.ts
@@ -124,19 +124,12 @@ class AppController {
   setStateFromSnapshot(): void {
     let url = this.url
     fetch(url)
-      .then(resp =>
-        resp
-          .json()
-          .then(data => {
-            data.Resources = this.setDefaultResourceInfo(data.Resources)
-            // @ts-ignore
-            this.component.setAppState({ View: data })
-          })
-          .catch(err => {
-            // TODO(dmiller): set app state with an error message
-            console.error(err)
-          })
-      )
+      .then(resp => resp.json())
+      .then(data => {
+        data.Resources = this.setDefaultResourceInfo(data.Resources)
+        // @ts-ignore
+        this.component.setAppState({ View: data })
+      })
       .catch(err => {
         // TODO(dmiller): set app state with an error message
         console.error(err)
--- a/web/src/AppController.ts
+++ b/web/src/AppController.ts
@@ -45,25 +45,29 @@ class AppController {
 
       let data = JSON.parse(event.data)
 
-      data.Resources = data.Resources.map((r: any) => {
-        if (r.ResourceInfo === null) {
-          r.ResourceInfo = {
-            PodName: "",
-            PodCreationTime: "",
-            PodUpdateStartTime: "",
-            PodStatus: "",
-            PodStatusMessage: "",
-            PodRestarts: 0,
-            PodLog: "",
-            YAML: "",
-            Endpoints: [],
-          }
-        }
-        r.Alerts = getResourceAlerts(r)
-        return r
-      })
+      data.Resources = this.setDefaultResourceInfo(data.Resources)
       // @ts-ignore
       this.component.setAppState({ View: data })
+    })
+  }
+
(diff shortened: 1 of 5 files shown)


web/src/HUD.tsx
```diff
@@ -107,11 +104,17 @@ class HUD extends Component<HudProps, HudState> {
...
+    if (this.pathBuilder.isSnapshot()) {
+      this.controller.setStateFromSnapshot()
+    } else {
+      this.controller.createNewSocket()
+    }
   }
 
   componentWillUnmount() {
-    this.controller.dispose()
+    if (!this.pathBuilder.isSnapshot()) {
```
jazzdan: Maybe there should be a function like `isWebsocket` instead of doing this inverse thing.

web/src/HUD.tsx
```diff
@@ -321,6 +324,23 @@ class HUD extends Component<HudProps, HudState> {
         )
       }
     }
+    let snapshotRoute = () => {
+      return (
+        <div className="SnapshotMessage">
+          <h1>Welcome to a Tilt snapshot!</h1>
+          <p>
+            In here you can look around and check out a "snapshot" of a Tilt
```
jazzdan: This is not the final copy :P

web/src/AppController.ts
```diff
@@ -45,28 +45,32 @@ class AppController {
...
-        r.Alerts = getResourceAlerts(r)
-        return r
-      })
+      data.Resources = this.setDefaultResourceInfo(data.Resources)
       // @ts-ignore
       this.component.setAppState({ View: data })
     })
   }
 
+  setDefaultResourceInfo(resources: Array<any>): Array<any> {
```
nicks: i wonder how long it will be until we just switch over to js protobufs, or some other message format that automatically does the default population for us

web/src/AppController.ts
```diff
@@ -116,6 +120,28 @@ class AppController {
       this.createNewSocket()
     }, timeout)
   }
+
+  setStateFromSnapshot(): void {
+    let url = this.url
+    fetch(url)
+      .then(resp =>
+        resp
```
nicks: nested promise chains like this are strongly discouraged. a good rule of them is that whenever you have a `then` call inside a `then` handler, you're probably doing something wrong

try something like:
```
fetch(url).then(resp => resp.json()).then(data => ...)
```

web/src/AppController.ts
```diff
@@ -116,6 +120,28 @@ class AppController {
...
+          .json()
+          .then(data => {
+            data.Resources = this.setDefaultResourceInfo(data.Resources)
```
nicks: it seems weird that this both modifies the data in-place and re-assigns it

web/src/HUD.tsx
nicks: :+1:

web/src/HUD.tsx
```diff
@@ -107,11 +104,17 @@ class HUD extends Component<HudProps, HudState> {
...
+      this.controller.setStateFromSnapshot()
+    } else {
+      this.controller.createNewSocket()
+    }
   }
 
   componentWillUnmount() {
-    this.controller.dispose()
+    if (!this.pathBuilder.isSnapshot()) {
+      this.controller.dispose()
```
nicks: i don't understand why the controller is conditionally disposed. seems like it should always be disposed?

web/src/AppController.ts
jazzdan: I would like to investigate solutions for this at some point because it is very painful. I've heard negative things about using protobufs for js but haven't looked in to it myself.

web/src/HUD.tsx
jazzdan: Well it doesn't do anything in non-websocket contexts, but also won't crash if the socket isn't set so I'll remove the conditional here as it is unnecessary.

PR #2012, opened 2019-08-07
unrevert "tiltfile: disallow deploys to remote kube by default"

### LINKED ISSUE #1096: Should Tilt allow deploys to non-local kubecontexts by default?

landism:
(re-)resolves #1096

#1995 was reverted because it turns out the latest Docker for Desktop doesn't use "localhost" as its kube api, but instead adds kubernetes.docker.local to /etc/hosts, and uses that.

This is a two-commit PR:
#1 Revert the revert (i.e., exactly what was already approved)
#2 Add Docker for Desktop context names to the whitelist


internal/tiltfile/tiltfile_state.go
```diff
@@ -94,6 +95,11 @@ const (
 	k8sResourceAssemblyVersionReasonExplicit
 )
 
+// These are k8s context names that we assume are safe to deploy to even if they are neither localhost
+// nor in allow_k8s_contexts. e.g., minikube uses a non-loopback ip on a virtual interface,
+// and some versions of docker-for-desktop use 'kubernetes.docker.local'
+var defaultWhitelistedKubeContexts = []k8s.KubeContext{"minikube", "docker-desktop", "docker-for-desktop"}
```
nicks: can this use the Env constants?

landism: > my main concern is having multiple codepaths that each invent their own definition of "local" cluster.

I'd forgotten about `Env` when writing this. Do you see any downside to just using `Env.IsLocalCluster` and skip checking the url?

I'm not very familiar with KIND, but it seems like the chief current use of `Env.IsLocalCluster` is to check whether can skip the push to docker, and we push to docker for KIND even though it's local? Which is to say, I'm not sure whether this method's goals align with the safety check's. Either way, though, it seems like it makes sense to put this logic in env.go.

landism: OK, new take:
1. renamed the existing `Env.IsLocalCluster` to `Env.UsesLocalDockerRegistry`, since that seems to better match its existing use, and then made a new `Env.IsLocalCluster` that also includes KinD.
2. just use `Env.IsLocalCluster` and don't get clever in the tiltfile code

PR #2025, opened 2019-08-09
upper: rm InitAction.ExecuteFile

landism:
InitAction.ExecuteFile is a flag used by unit tests to bypass Tiltfile loading and pass Manifests in directly via InitAction.

Now, we'll instead use a FakeTiltfileLoader that returns the manifests we want, so all upper tests exercise the normal InitAction / ConfigsReloadedAction flow.

Also deleted some misc nearby fields that were apparently unused.


internal/engine/actions.go
```diff
@@ -93,18 +93,12 @@ func NewBuildCompleteAction(result store.BuildResultSet, err error) BuildComplet
 
 type InitAction struct {
```
nicks: this pr makes me wonder if InitAction should be an empty struct. I feel like a lot of these fields are just arguments and flags that would be easier to handle with dependency injection

PR #2028, opened 2019-08-09
tiltfile: fast build deprecation date

maiamcc:
Feel free to reject the date, or even the fact of setting a hard deadline,
but I would love for us to actually get fast_build out of our codebase.

This PR puts a deprecation deadline in the fast_build deprecation warning.
Come 9/12, we'd remove tiltfile support for `fast_build`, and remove FastBuild internally at our leisure.


jazzdan:
FWIW I've seen these code deadlines come and go maaaany times without any action. I think for this to be effective it also needs to be prioritized in the [Tilt Roadmap](https://github.com/orgs/windmilleng/projects/3).

PR #2033, opened 2019-08-09
Support alerts for docker compose alerts

mariavivanco:
Hi @jazzdan, 

This PR is In response to #2033. Please review changes made to frontend to support Alerts for docker compose 

Screen Shot 2019-08-09 at 3 51 33 PM


web/src/alerts.test.ts
```diff
@@ -211,7 +216,192 @@ describe("getResourceAlerts", () => {
   })
 })
 
-function emptyResource(): Resource {
+//DC Resource Tests
```
jazzdan: This comment is repeated twice

web/src/alerts.ts
```diff
@@ -21,47 +22,57 @@ function hasAlert(resource: Resource) {
...
   return (
     resource.BuildHistory.length > 0 && resource.BuildHistory[0].Error !== null
   )
 }
 
-function numberOfAlerts(resource: Resource): number {
-  return getResourceAlerts(resource).length
+function isK8sResourceInfo(
+  resourceInfo: K8sResourceInfo | DCResourceInfo
+): resourceInfo is K8sResourceInfo {
```
jazzdan: Woah, this function is crazy. What is happening here? I want to learn. :)

mariavivanco: prettier made it ugly >:( but basically it takes in a resourceInfo of either type and checks if a field is defined in there, so in here it checks in the resourceInfo if podName is defined (which it would be for a K8s resource and not a DC resource). Then it returns a boolean based on that :)

jazzdan: What is `is` for?

mariavivanco: its called a "type predicate" found it on here: https://www.typescriptlang.org/docs/handbook/advanced-types.html

PR #2034, opened 2019-08-09
k8s: add k3d env detection

nicks:
Hello @landism,

Please review the following commits I made in branch nicks/k3d:

ce087491490ed5ce73a8d1d32b169551ee18b473 (2019-08-09 17:26:55 -0400)
k8s: add k3d env detection

### DIFF
--- a/internal/k8s/env.go
+++ b/internal/k8s/env.go
@@ -1,11 +1,17 @@
 package k8s
 
 import (
+	"context"
+	"path/filepath"
 	"strings"
 
+	"github.com/mitchellh/go-homedir"
 	"github.com/pkg/errors"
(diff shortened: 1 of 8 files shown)


internal/k8s/env.go
```diff
@@ -1,11 +1,17 @@
 package k8s
 
 import (
+	"context"
+	"path/filepath"
 	"strings"
 
+	"github.com/mitchellh/go-homedir"
```
nicks: i used homedir for consistency with k3d, and it's also supposed to be more robust cross-os?
https://github.com/rancher/k3d/blob/e56be1344eeba451285b26934298f8b6e5da7c5e/cli/cluster.go#L85

PR #2035, opened 2019-08-12
tiltfile/engine: metrics for multi-container usage

maiamcc:
Would love feedback on how to report these stats for best slice-ability.

This PR records info on:
- number of containers running each ref
- number of resources with mutli-container live update

### DIFF
--- a/internal/engine/analytics_reporter.go
+++ b/internal/engine/analytics_reporter.go
@@ -51,17 +51,19 @@ func ProvideAnalyticsReporter(a *analytics.TiltAnalytics, st *store.Store) *Anal
 func (ar *AnalyticsReporter) report() {
 	st := ar.store.RLockState()
 	defer ar.store.RUnlockState()
-	var dcCount, k8sCount, fastbuildBaseCount, anyFastbuildCount, liveUpdateCount, unbuiltCount int
+	var dcCount, k8sCount, fastbuildBaseCount, anyFastbuildCount, liveUpdateCount, unbuiltCount, multiImgLiveUpdate int
 	for _, m := range st.Manifests() {
 		if m.IsK8s() {
 			k8sCount++
 			if len(m.ImageTargets) == 0 {
 				unbuiltCount++
 			}
+			// Open Q: stat for number of resources with multiple image targets? (# targets per resource?)
 		}
 		if m.IsDC() {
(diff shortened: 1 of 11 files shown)


internal/engine/analytics_reporter.go
```diff
@@ -71,9 +72,17 @@ func (ar *AnalyticsReporter) report() {
 				break
 			}
 			if !it.AnyLiveUpdateInfo().Empty() {
-				liveUpdateCount++
+				if !seenLU {
+					seenLU = true
+					liveUpdateCount++
```
maiamcc: I just changed this stat to report the number of LiveUpdate _resources_ (rather than the number of LiveUpdate _targets_, which is now accounted for by the `multiImgLiveUpdate` value). This seems right to me but lmk if you disagree.

internal/tiltfile/tiltfile_state.go
```diff
@@ -815,6 +819,15 @@ func (s *tiltfileState) translateK8s(resources []*k8sResource) ([]model.Manifest
 		}
 
 		result = append(result, m)
+		for _, iTarg := range iTargets {
+			ref := iTarg.ConfigurationRef.String()
+			hasLiveUpd := fmt.Sprintf("%t", !iTarg.AnyLiveUpdateInfo().Empty())
+			count := r.imageRefMap[ref]
+
+			a.Count("containersForRef", map[string]string{
```
maiamcc: re: `containersForRef` -- do we have any casing guidelines for metric names? 🤷‍♀

nicks: this is a bad place to report this data, because it will get executed every time the tiltfile executes. please move this kind of reporting to the engine/analytics_reporter. That is what it exists for.

maiamcc: I did this to avoid storing extra data on the manifest b/c we currently don't know how many times a given image target is referenced in the yaml (unless we re-parse the yaml) -- but sure, can do.

maiamcc:
@nicks this is ready for re-review

maiamcc:
cc @jazzdan -- this is the one I'd love to get into the next release. (I'm heading out now but if this gets approved, feel free to merge on my behalf if I'm not around.)

PR #2037, opened 2019-08-12
DEVELOPING: update version number strategy

jazzdan:


DEVELOPING.md
```diff
@@ -255,4 +255,10 @@ the [default dev version](internal/cli/build.go),
 and the [installer version](scripts/install.sh).
 
 ### Version numbers
-Our rule of thumb pre 1.0 is only bump the minor version if you would write a blog post about it. (We haven't always followed this rule, but we'd like to start!)
+For pre-v1.0:
+* If adding backwards-compatible functionality increment the patch version (0.x.Y)
```
nicks: fwiw, i think that 
1) we're adding backwards-incompatible functionality, and
2) we want to write a blog post about it,
these are going to be very highly correlated

PR #2038, opened 2019-08-12
engine: analytics reporter should report as soon as it starts up (currently reports hourly starting 1 hr after startup)

maiamcc:
Hello @nicks, @jazzdan,

Please review the following commits I made in branch maiamcc/earlier-analytics:

edb937503d02ebaf9ee435d804669f76a20aa4e2 (2019-08-12 17:40:57 -0400)
engine: analytics reporter should report as soon as it starts up (currently reports hourly starting 1 hr after startup)


maiamcc:
> maybe it should be like 15minutes?

Nick, you mean the reporting interval should be 15 minutes, or that we should wait 15 minutes before the first report?

maiamcc:
@nicks ☝️ can you clarify?

nicks:
the interval

PR #2042, opened 2019-08-13
engine: log changed config files in manifest builds

landism:
### Problem
When we change a config file and it causes a manifest to rebuild, the log would just say, e.g. "rebuilding snack", without any indication why it was happening

### Solution
log, e.g., "1 changed: [snack/Dockerfile]"

### DIFF
--- a/internal/engine/buildcontroller.go
+++ b/internal/engine/buildcontroller.go
@@ -21,6 +21,7 @@ type buildEntry struct {
 	name          model.ManifestName
 	targets       []model.TargetSpec
 	buildStateSet store.BuildStateSet
+	filesChanged  []string
 	buildReason   model.BuildReason
 	firstBuild    bool
 }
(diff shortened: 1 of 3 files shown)


internal/engine/buildcontroller.go
```diff
@@ -212,14 +214,13 @@ func (c *BuildController) OnChange(ctx context.Context, st store.RStore) {
 		}
 		ctx := logger.WithLogger(ctx, logger.NewLogger(logger.Get(ctx).Level(), actionWriter))
 
-		filesChanged := entry.buildStateSet.FilesChanged()
```
maiamcc: pretty sure `bss.FilesChanged` is obsolete now

landism: it's used up in `needsBuild` now: https://github.com/windmilleng/tilt/pull/2042/files#diff-fe6c9f90c8a07f06670c56f531886cedR192

PR #2049, opened 2019-08-13
src: the http bootstrap has the entire state, not the websocket

jazzdan:


web/src/AppController.ts
```diff
@@ -126,7 +126,7 @@ class AppController {
     fetch(url)
```
mariavivanco: where is this function used?

jazzdan: It's a function invocation, not a function definition. Is that what you mean?

jazzdan: https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API

PR #2054, opened 2019-08-13
TFT: Tilt has a button that allows user to create a snapshot and share the link

mariavivanco:
Screen Shot 2019-08-13 at 5 40 52 PM
Screen Shot 2019-08-13 at 5 41 01 PM

finally works end to end ! 😎


internal/hud/server/server.go
```diff
@@ -266,3 +271,81 @@ func tsAlertToBackendAlert(alert tsAlert) tft.Alert {
...
+	return fmt.Sprintf("https://%s/snapshot/%s", tiltAlertsDomain, id)
+}
+
+func (s *HeadsUpServer) HandleNewSnapshot(w http.ResponseWriter, req *http.Request) {
+	if req.Method != http.MethodPost {
+		http.Error(w, "must be POST request", http.StatusBadRequest)
+		return
+	}
+
+	request, err := http.NewRequest(http.MethodPost, "https://alerts.tilt.dev/api/snapshot/new", req.Body) // New request
```
jazzdan:
```suggestion
	request, err := http.NewRequest(http.MethodPost, "https://alerts.tilt.dev/api/snapshot/new", req.Body)
```

internal/hud/server/server_test.go
```diff
@@ -21,6 +24,16 @@ import (
...
 )
 
+var originalWD string
+
+func init() {
+	wd, err := os.Getwd()
+	if err != nil {
+		panic(err)
+	}
+	originalWD = wd
```
jazzdan: Hmm why do we need to do this? Does the HUD server test muck with the current working directory?

internal/tft/client/client.go
```diff
@@ -54,6 +54,8 @@ func (t *tftClient) SendAlert(ctx context.Context, alert Alert) (AlertID, error)
 	return AlertID(r.ID), nil
 }
 
+//TODO TFT: Create send snapshot function: returns snapshot ID and error
```
jazzdan: Are we still planning on doing this?

web/src/HUD.tsx
```diff
@@ -167,6 +182,25 @@ class HUD extends Component<HudProps, HudState> {
       .then(err => console.error(err))
   }
 
+  sendSnapshot(snapshot: Snapshot) {
+    let url = `//${window.location.host}/api/snapshot/new`
+    fetch(url, {
+      method: "post",
+      body: JSON.stringify(snapshot),
+    })
+      .then(res => {
```
jazzdan: I think this promise could be re-written more idiomatically as:

```typescript
fetch()
  .then(res => res.json)
  .then((value: NewSnapshotResponse) => {
            this.setState({
              SnapshotLink: value.url,
            })
  })
  .catch(err => console.error(err)
```

then the same catch will handle everything.

web/src/TopBar.tsx
```diff
@@ -35,4 +48,32 @@ class TopBar extends PureComponent<TopBarProps> {
...
+        <button
+          title="Open link in new tab"
+          onClick={() => window.open(snapshotURL)}
+        >
+          Open
+        </button>
+      </section>
+    )
+  }
+  //TODO TFT - formatting the button
```
jazzdan: Hmm the button looks formatted to me!

internal/hud/server/server_test.go
hyu: Haha, as far as I know we were just trying to read the testdata dump with all the JSON, and found this snippet in some SO answer.

PR #2055, opened 2019-08-14
k8s: Allow GetByReference to return structured objects as well as unstructured ones

nicks:
Hello @jazzdan, @landism,

Please review the following commits I made in branch nicks/entity:

1959f82b8f6ad23e67df8388ae5c034aba8235e6 (2019-08-14 10:57:20 -0400)
k8s: Allow GetByReference to return structured objects as well as unstructured ones

### DIFF
--- a/internal/engine/pod_watch.go
+++ b/internal/engine/pod_watch.go
@@ -85,7 +85,7 @@ func (w *PodWatcher) OnChange(ctx context.Context, st store.RStore) {
 			st.Dispatch(NewErrorAction(err))
 			return
 		}
-		go dispatchPodChangesLoop(ctx, ch, st)
+		go w.dispatchPodChangesLoop(ctx, ch, st)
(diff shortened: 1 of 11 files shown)


internal/engine/pod_watch.go
```diff
@@ -104,7 +104,7 @@ func (w *PodWatcher) removeWatch(toRemove PodWatch) {
 	}
 }
 
-func dispatchPodChangesLoop(ctx context.Context, ch <-chan *v1.Pod, st store.RStore) {
+func (w *PodWatcher) dispatchPodChangesLoop(ctx context.Context, ch <-chan *v1.Pod, st store.RStore) {
```
jazzdan: Hmm, does this change do anything?

nicks: nope! but it will be important in a bit

PR #2064, opened 2019-08-15
internal/build: also log last message from docker

dbentley:


dbentley:
Have any ideas on how to test this well? I think for something like this I care more about getting it in, given that it's only in a case where things are already bad

nicks:
it's worth pointing out that most buildkit messages are binary-encoded. i think this is safe to try, but there's a risk it might just spew binary gibberish at people in most cases

nicks:
this should be obsolete now, see discussion in https://github.com/windmilleng/tilt/issues/1967

PR #2014, opened 2019-08-08
integration: test LiveUpdate multiple images one manifest [ch3115]

maiamcc:
Hello @,

Please review the following commits I made in branch maiamcc/integ-test-two-images-one-manifest:

38865671b3db8ac4fe26ae66ebb7b7f00678fc9d (2019-08-08 11:24:36 -0400)
integration: test LiveUpdate multiple images one manifest [ch3115]

Code review reminders, by giving a LGTM you attest that:

* Commits are adequately tested
* Code is easy to understand and conforms to style guides
* Incomplete code is marked with TODOs
* Code is suitably instrumented with logging and metrics

PR #2072, opened 2019-08-19
integration: bump timeout

landism:
I think #2014 made it so we're now hitting the integration test timeout, causing us to occasionally fail in master. At some point we might want to speed this up, but for now let's not waste our time investigating fake failures.


maiamcc:
sorryyyy 😢

Issue #2077, opened 2019-08-20
implicit port forwarding should only fall back to the first available port if the local port is not exposed


omrikiei:
When only providing an int/string in the port_forwards attribute(k8s_resource), tilt should attempt to map the local port as the container port if such port is exposed, and only if there isn't one default to the first exposed port.

nicks:
fixed by https://github.com/windmilleng/tilt/pull/2078

PR #2078, opened 2019-08-20
#2077 fix: port forwarding should first try to match local port to the expo...

omrikiei:
…sed container port(issue #2077)


jazzdan:
Thanks @omrikiei!

@nicks can you take a look at this?

nicks:
ya, this makes sense to me. thanks! lgtm

PR #2083, opened 2019-08-20
fixup for my previous code(#2077)

omrikiei:


nicks:
thanks! it's important to save those cpu cycles!

PR #2085, opened 2019-08-20
web: pathbuilder.path handles snapshot URLs

jazzdan:


jazzdan:
Tested end to end using https://github.com/windmilleng/tilt/pull/2086

PR #2086, opened 2019-08-20
hud/server: create snapshot HTTP bootstrap simulacrum

jazzdan:
With this change someone working on Tilt locally can navigate to
http://localhost:10350/snapshot/aaaaaa and Tilt will make a request to
localhost:10350/api/snapshot/aaaaaa which will serialize the current
Tilt engine state in to the snapshot format that the frontend understands
and return it.

Tilt will bootstrap off of this HTTP request thus simulating snapshots.

This addresses only part of ch3234

### DIFF
--- a/internal/hud/server/server.go
+++ b/internal/hud/server/server.go
@@ -64,6 +64,7 @@ func ProvideHeadsUpServer(store *store.Store, assetServer assets.Server, analyti
 	r.HandleFunc("/api/sail", s.HandleSail)
 	r.HandleFunc("/api/trigger", s.HandleTrigger)
 	r.HandleFunc("/api/snapshot/new", s.HandleNewSnapshot)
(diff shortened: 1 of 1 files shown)


internal/hud/server/server.go
```diff
@@ -64,6 +64,7 @@ func ProvideHeadsUpServer(store *store.Store, assetServer assets.Server, analyti
 	r.HandleFunc("/api/sail", s.HandleSail)
 	r.HandleFunc("/api/trigger", s.HandleTrigger)
 	r.HandleFunc("/api/snapshot/new", s.HandleNewSnapshot)
+	r.HandleFunc("/api/snapshot/{snapshot_id}", s.SnapshotJSON)
```
landism: is `{snapshot_id}` actually used here, is it only to match the path used by the snapshot server?

is this endpoint only for local testing? seems worth a comment

jazzdan: Yes and yes, will add a comment.

PR #2089, opened 2019-08-20
engine: tilt should generate a token on startup if one doesn't exist

jazzdan:
For use with teams and for possible use as a new primary key in analytics.

Open questios:
* Is it OK to just keep regenerating this over and over again if people delete the file?
* Is there a meaningful test to write here that I'm missing?


internal/engine/upper_test.go
```diff
@@ -2819,7 +2820,7 @@ func newTestFixture(t *testing.T) *testFixture {
 	tvc := NewTiltVersionChecker(func() github.Client { return ghc }, tiltVersionCheckTimerMaker)
 
 	subs := ProvideSubscribers(fakeHud, pw, sw, plm, pfc, fwm, bc, ic, cc, dcw, dclm, pm, sm, ar, hudsc, sc, tvc, tas, ewm)
-	ret.upper = NewUpper(ctx, st, subs)
+	ret.upper = NewUpper(ctx, st, subs, token.Token(""))
```
landism: I might make this `"unit test token"` or something, just in case at some point someone accidentally changes code leading to a test making a real request and we end up with a bunch of unit test requests in our logs / analytics

internal/token/token.go
```diff
@@ -0,0 +1,41 @@
...
+)
+
+const tokenPath = "token"
+
+type Token string
+
+func GetOrCreateToken(dir *dirs.WindmillDir) (Token, error) {
+	token, err := getExistingToken(dir)
+	if os.IsNotExist(err) {
+		u := uuid.New()
```
landism: Could we just use a random string instead? Or do you know if Google's UUID package attempt to be unguessable? The [docs](https://godoc.org/github.com/google/uuid) don't make mention of guessability, but references [RFC 4122](https://tools.ietf.org/html/rfc4122), which seems pretty explicit: "Do not assume that UUIDs are hard to guess; they should not be used as security capabilities (identifiers whose mere possession grants access), for example."

internal/token/token.go
```diff
@@ -0,0 +1,41 @@
+package token
+
+import (
+	"os"
+
+	"github.com/google/uuid"
+	"github.com/windmilleng/wmclient/pkg/dirs"
+)
+
+const tokenPath = "token"
```
landism: nit: more of a filename than a path

internal/token/token.go
jazzdan: I looked in to this and these UUIDs _are_ random strings.

google/uuid uses crypto/rand under the hood which is what we would use to generate a random string. It in turn uses `getrandom(2)` on Linux and `getentropy(2)` on macOS, which as far as I can tell from the man pages are secure/unguessable.

Furthermore google/uuid [reads 122 random bits](https://github.com/google/uuid/blob/master/version4.go#L24).

PR #2095, opened 2019-08-21
 web: move share snapshot in to modal, show register token URL 

jazzdan:
It's pretty ugly, but:

Screen Shot 2019-08-21 at 5 49 00 PM

After clicking "Share Snapshot":

Screen Shot 2019-08-21 at 5 49 10 PM

Will probably need to rope @nicks in for some styling help 😅

### DIFF
--- a/internal/tft/url.go
+++ b/internal/tft/url.go
@@ -1,0 +1,14 @@
+package tft
+
+import(
+	"fmt"
+
+	"github.com/windmilleng/tilt/internal/token"
+)
+
+//TODO TFT: change snapshot url to be cloud.tilt.dev
+const Domain = "alerts.tilt.dev"
+
+func RegisterTokenURL(t token.Token) string {
+	return fmt.Sprintf("https://%s/register_token?token=%s", Domain, t)
+}
--- a/internal/tft/url.go
+++ b/internal/tft/url.go
@@ -1,6 +1,6 @@
 package tft
 
-import(
+import (
 	"fmt"
 
 	"github.com/windmilleng/tilt/internal/token"
--- a/web/src/HUD.tsx
+++ b/web/src/HUD.tsx
@@ -43,7 +43,7 @@ type HudState = {
   } | null
   IsSidebarClosed: boolean
   SnapshotLink: string
-  ShowSnapshotModal: boolean
+  showSnapshotModal: boolean
 }
 
 type NewSnapshotResponse = {
@@ -95,7 +95,7 @@ class HUD extends Component<HudProps, HudState> {
       },
       IsSidebarClosed: false,
       SnapshotLink: "",
-      ShowSnapshotModal: false,
+      showSnapshotModal: false,
     }
 
     this.toggleSidebar = this.toggleSidebar.bind(this)
@@ -206,7 +206,7 @@ class HUD extends Component<HudProps, HudState> {
       )
     }
 
-    let handleOpenModal = () => this.setState({ ShowSnapshotModal: true })
+    let handleOpenModal = () => this.setState({ showSnapshotModal: true })
 
     let topBarRoute = (t: ResourceView, props: RouteComponentProps<any>) => {
       let name =
@@ -353,8 +353,8 @@ class HUD extends Component<HudProps, HudState> {
         <AnalyticsNudge needsNudge={needsNudge} />
         <ShareSnapshotModal
           handleSendSnapshot={this.sendSnapshot.bind(this, this.state)}
-          handleClose={() => this.setState({ ShowSnapshotModal: false })}
-          show={this.state.ShowSnapshotModal}
+          handleClose={() => this.setState({ showSnapshotModal: false })}
+          show={this.state.showSnapshotModal}
           snapshotUrl={this.state.SnapshotLink}
           registerTokenUrl={registerTokenUrl}
         />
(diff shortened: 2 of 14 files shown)


internal/tft/url.go
```diff
@@ -0,0 +1,14 @@
+package tft
```
dbentley: Maybe "cloud" instead of "tft"?

web/src/ShareSnapshotModal.tsx
```diff
@@ -0,0 +1,59 @@
...
+    ? "modal display-block"
+    : "modal display-none"
+  const hasLink = props.snapshotUrl !== ""
+  if (!hasLink) {
+    return (
+      <div className={showHideClassName}>
+        <section className="modal-main">
+          <section className="modal-snapshotUrlWrap">
+            <button onClick={props.handleClose}>close</button>
+            <button onClick={() => props.handleSendSnapshot(props.state)}>
```
nicks: it seems weird to pass modal the entire state. usually you would bind the snapshot to the function, so that this was just invoked as props.sendSnapshot()

web/src/ShareSnapshotModal.tsx
```diff
@@ -0,0 +1,59 @@
...
+}
+export default function ShareSnapshotModal(props: props) {
+  const showHideClassName = props.show
+    ? "modal display-block"
+    : "modal display-none"
+  const hasLink = props.snapshotUrl !== ""
+  if (!hasLink) {
+    return (
+      <div className={showHideClassName}>
+        <section className="modal-main">
```
nicks: this isn't what `section` is for, see 
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/section#Usage_notes

```
Do not use the <section> element as a generic container; this is what <div> is for, especially when the sectioning is only for styling purposes. A rule of thumb is that a section should logically appear in the outline of a document.
```

web/src/ShareSnapshotModal.tsx
```diff
@@ -0,0 +1,59 @@
...
+      </div>
+    )
+  }
+
+  return (
+    <div className={showHideClassName}>
+      <section className="modal-main">
+        <section className="modal-snapshotUrlWrap">
+          <button onClick={props.handleClose}>close</button>
+          <button onClick={() => window.open(props.snapshotUrl)}>
```
nicks: can you reduce the duplication here by assigning this button to a variable?

e.g.
let button = <button...>
if (hasLink) {
  button = <button ...>
}

web/src/ShareSnapshotModal.scss
```diff
@@ -0,0 +1,60 @@
...
+  background: $color-white;
+  color: $color-gray-darkest;
+  width: 80%;
+  height: auto;
+  top:50%;
+  left:50%;
+  transform: translate(-50%,-50%);
+}
+
+.display-block {
```
nicks: i would usually call this u-displayBlock / u-displayNone

in BEM convention, "display-none" typically means "in the display component, there's a sub-element called block", which is obviously not what you mean here

web/src/ShareSnapshotModal.scss
```diff
@@ -0,0 +1,60 @@
...
+  position: fixed;
+  top: 0;
+  left: 0;
+  width:100%;
+  height: 100%;
+  background: rgba(0, 0, 0, 0.6);
+}
+
+.modal-main {
+  position:fixed;
```
nicks: if you've got a fixed element inside another fixed element, that usually means you're confused and not using it correctly. tho i can clean this up in a follow-up pr

web/src/TopBar.tsx
```diff
@@ -46,33 +40,12 @@ class TopBar extends PureComponent<TopBarProps> {
...
-        <button
-          title="Open link in new tab"
-          onClick={() => window.open(snapshotURL)}
-        >
-          Open
-        </button>
-      </section>
-    )
-  }
+const renderSnapshotModal = (handleOpenModal: () => void) => {
```
nicks: this would make more sense as a method of TopBar? that way it has access to this.props. It creates a sub-element of topbar and doesn't seem generally useful outside it

web/src/ShareSnapshotModal.tsx
```diff
@@ -0,0 +1,59 @@
...
+          </button>
+        </section>
+        <hr />
+        {loggedOutTiltCloudCopy("")}
+      </section>
+    </div>
+  )
+}
+
+const loggedOutTiltCloudCopy = (registerTokenURL: string) => (
```
nicks: it might make more sense if ShareSnapshotModal was a class, so this could be a method

web/src/ShareSnapshotModal.tsx
nicks: also, this should really be an `<a>` with an href

web/src/HUD.tsx
```diff
@@ -38,24 +39,18 @@ type HudState = {
     RunningTiltBuild: TiltBuild
     LatestTiltBuild: TiltBuild
     FeatureFlags: { [featureFlag: string]: boolean }
+    RegisterTokenURL: string
   } | null
   IsSidebarClosed: boolean
   SnapshotLink: string
+  ShowSnapshotModal: boolean
```
nicks: fine for this PR, but it seems weird that all these fields are capitalized

jazzdan: They're capitalized because they come from Go and we don't override the JSON serialization.

jazzdan: They're capitalized because they come from Go and we don't override the JSON serialization.

nicks: this property doesn't come from Go, though, right? or am i misunderstanding?

jazzdan: Ohh you're right @nicks, good call. I can lowercase it.

PR #2097, opened 2019-08-22
engine: make the ServiceWatcher responsible for mapping services to manifests

nicks:
Hello @maiamcc, @jazzdan,

Please review the following commits I made in branch nicks/uidexp:

3b43542b3dd4cf42c36ff9205d6da7b3e154cf40 (2019-08-22 13:13:54 -0400)
engine: make the ServiceWatcher responsible for mapping services to manifests

### DIFF
--- a/internal/engine/actions.go
+++ b/internal/engine/actions.go
@@ -31,14 +31,19 @@ func NewPodChangeAction(pod *v1.Pod) PodChangeAction {
 }
 
 type ServiceChangeAction struct {
-	Service *v1.Service
-	URL     *url.URL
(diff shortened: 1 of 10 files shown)


nicks:
i had a crazy idea last night to move all the logic for mapping K8s resources -> Manifest out of the Upper reducers and into the individual watchers. 

It think this is going to end up being much more robust, and make it easier to handle things like watching non-deployed resources

nicks:
(i'm doing it with ServiceWatch first because it's simpler, than going to try it out with PodWatch)

PR #2104, opened 2019-08-23
engine: triage pods by owner UID rather than by manifest label

nicks:
Hello @landism, @maiamcc,

Please review the following commits I made in branch nicks/podwatch:

30334e83b6572cba08adf6ff55f986c3fc236f31 (2019-08-23 14:52:47 -0400)
engine: triage pods by owner UID rather than by manifest label

### DIFF
--- a/internal/engine/pod_watch.go
+++ b/internal/engine/pod_watch.go
@@ -7,10 +7,12 @@ import (
 
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/types"
 
 	"github.com/pkg/errors"
 
 	"github.com/windmilleng/tilt/internal/store"
+	"github.com/windmilleng/tilt/pkg/logger"
 	"github.com/windmilleng/tilt/pkg/model"
 
 	"github.com/windmilleng/tilt/internal/k8s"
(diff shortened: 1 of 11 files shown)


internal/engine/pod_watch.go
```diff
@@ -60,36 +75,57 @@ func subtract(a, b []PodWatch) []PodWatch {
...
+		for _, ls := range mt.Manifest.K8sTarget().ExtraPodSelectors {
+			if !ls.Empty() {
+				neededWatches = append(neededWatches, PodWatch{name: mt.Manifest.Name, labels: ls})
+			}
+		}
+
+		for id := range mt.State.K8sRuntimeState().DeployedUIDSet {
+			oldName := w.knownDeployedUIDs[id]
+			if name != oldName {
+				newUIDs[id] = name
```
maiamcc: 🤔 when would this happen? (i agree it's good to have this check here to be safe, i'm just trying to imagine when it might happen)

internal/engine/pod_watch.go
```diff
@@ -60,36 +75,57 @@ func subtract(a, b []PodWatch) []PodWatch {
...
+		name := mt.Manifest.Name
+		atLeastOneK8s = true
+		for _, ls := range mt.Manifest.K8sTarget().ExtraPodSelectors {
+			if !ls.Empty() {
+				neededWatches = append(neededWatches, PodWatch{name: mt.Manifest.Name, labels: ls})
+			}
+		}
+
+		for id := range mt.State.K8sRuntimeState().DeployedUIDSet {
+			oldName := w.knownDeployedUIDs[id]
```
maiamcc: more comments pls, for the sake of the next person to work in this file? esp. here, but throughout

internal/engine/pod_watch.go
```diff
@@ -126,19 +166,80 @@ func (w *PodWatcher) removeWatch(toRemove PodWatch) {
...
+
+func (w *PodWatcher) triagePodUpdate(pod *v1.Pod, objTree k8s.ObjectRefTree) model.ManifestName {
+	w.mu.Lock()
+	defer w.mu.Unlock()
+
+	uid := pod.UID
+	oldPod, ok := w.knownPods[uid]
+
+	// In "real" code, if we get two pod updates with the same resource version,
+	// we can safely ignore the new one. But dispatching a spurious event
```
maiamcc: do you mean "NOT dispatching a spurious event"?

internal/engine/pod_watch.go
nicks: every time we deploy something, the new UID will get added to the newUID map. So this will happen all the time, pretty regularly

internal/engine/pod_watch.go
nicks: nope! this code dispatches an event when oldPod.ResourceVersion == pod.ResourceVersion

maiamcc: OH! at a glance, the code looked like it was accounting for the case where a UID changed manifest associations o_0

internal/engine/pod_watch.go
nicks: done

PR #2105, opened 2019-08-23
remove some dead event code

nicks:
Hello @landism,

Please review the following commits I made in branch nicks/events:

fd83f6a3059e23f89863b6fc73a3f591c5dc6b75 (2019-08-23 16:07:16 -0400)
remove some dead event code


nicks:
ya, i need to fix event watch manager to not use labels, and it was easier to remove this code than to fix it to work with uids

PR #2147, opened 2019-09-04
tilt: update cloud username when registering token

landism:


internal/cloud/cloud_username_manager.go
```diff
@@ -63,6 +126,15 @@ func (c *CloudUsernameManager) OnChange(ctx context.Context, st store.RStore) {
...
 	c.mu.Unlock()
 
+	// if a refresh has been induced, then do a long_get lookup, so that we get the username as soon as
+	// the user has finished the process
+	if state.TiltCloudUsernameNeedsRefresh && !currentlyMakingRequest {
+		go c.CheckUsername(ctx, st, true)
+		return
+	}
+
+	// otherwise, we're not necessarily expecting a username, so just get the current state
```
maiamcc: see also: note about naming "refresh" vs. "wait for register and get username", but this comment is confusing. Maybe:
> otherwise, we don't think the user is actively in the process of registering their token; just get the current state, w/o waiting for a registration flow to finish

internal/cloud/cloud_username_manager.go
```diff
@@ -63,6 +126,15 @@ func (c *CloudUsernameManager) OnChange(ctx context.Context, st store.RStore) {
 	currentlyMakingRequest := c.currentlyMakingRequest
 	c.mu.Unlock()
 
+	// if a refresh has been induced, then do a long_get lookup, so that we get the username as soon as
+	// the user has finished the process
```
maiamcc: see also: note about naming "refresh", but this comment is confusing. consider like, `state.NeedTiltCloudUsernameAfterRegister` and:
> If the user is currently registering a token, then do a long_get... has finished the registration process

landism:
> mostly lgtm, but the naming is a little confusing. i would assume a "refresh" would be FAST, a simple check of the current state. you seem to be using "refresh" where i would use "wait for register" or "get username after register" or something.

Good call. That was named in simpler times.

internal/engine/upper.go
```diff
@@ -727,6 +727,9 @@ func handleAnalyticsNudgeSurfacedAction(ctx context.Context, state *store.Engine
 }
 
 func handleTiltCloudUserLookedUpAction(state *store.EngineState, action store.TiltCloudUserLookedUpAction) {
+	if state.TiltCloudUsername != "" {
+		state.TiltCloudUsernameNeedsRefresh = false
```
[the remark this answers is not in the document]
landism: I think I've solved this now by adding `action.IsPostRegistrationLookup`.

This also caught a related race condition:
1. CloudUsernameManager starts doing a normal username lookup
2. We trigger a long get lookup. CloudUsernameManager does nothing because there's already a request in process.
3. The lookup from (1) completes, setting TiltCloudUsernameNeedsRefresh to false
4. We never do the lookup from (2).

PR #2154, opened 2019-09-05
k8s: assert that the watchers emit all existing resources at startup

nicks:
Hello @nicks,

Please review the following commits I made in branch nicks/ch3277/list:

9c0fab812bc002366ced4d4d41ba85964995d5a8 (2019-09-05 13:28:52 -0400)
k8s: assert that the watchers emit all existing resources at startup


nicks:
this is mostly just moving test code around

PR #2160, opened 2019-09-06
web: design snapshot modal

hyu:
Made a pass at the copy, too.

## Token Not Registered
2019-09-05 - Snapshot modal - sign up

## Token Registered
2019-09-05 - Snapshot modal


web/src/__snapshots__/ShareSnapshotModal.test.tsx.snap
```diff
@@ -28,24 +28,59 @@ exports[`ShareSnapshotModal renders with modal open w/ known username 1`] = `
...
+      Share a Shapshot
+    </h2>
     <div
       className="ShareSnapshotModal-pane"
     >
-      <a
-        href="http://test.com"
-        target="_blank"
+      <p>
+        Let anyone explore your current Tilt session with an interactive snapshot.
```
landism: I think it would be good to emphasize:
1) read-only
2) not live

to provide assurance that they're only sharing a limited thing (don't need to worry about anyone else modifying your stuff, don't need to worry about keeping your Tilt in the state you want the user to see, and don't need to worry about whether any secrets logged in the future are now available via the snapshot). Maybe some of that could be inferred from "snapshot", but more explicitness seems like it'd yield more feelings of safety.

stab: "Let anyone explore a read-only copy of the current state of your Tilt session with an interactive snapshot"

The "anyone" also kinda gives me pause, like "if I share this, can the whole WORLD see it!?"
We could say "anyone with the link", a la google docs. We do at least provide that level of privacy, and are not currently communicating it anywhere. That might be too wordy.

Copy isn't the point of this PR though. If you don't wanna mess with this, LMK and I'll file a ticket for us to deal with it later.

web/src/ShareSnapshotModal.tsx
```diff
@@ -12,62 +12,108 @@ type props = {
...
+            target="_blank"
+            method="POST"
+            onSubmit={ShareSnapshotModal.notifyTiltOfRegistration}
+          >
+            <p>You'll need a GitHub account</p>
+            <span className="ShareSnapshotModal-tiltCloudButtonForm-spacer" />
+            <input
+              className="ShareSnapshotModal-tiltCloudButton"
+              type="submit"
+              value="Sign Up"
```
maiamcc: this copy isn't quiiiite right. needs to be like "sign up / link account" or something (cuz they may already be signed up for tilt cloud, and it's just associated with this Tilt instance yet)

web/src/ShareSnapshotModal.tsx
```diff
@@ -12,62 +12,108 @@ type props = {
...
-                value={cookies.get("Tilt-Token")}
-              />
-            </form>
-            to associate this copy of Tilt with your TiltCloud account.
-          </div>
-          <div>(You'll just need to link your GitHub Account)</div>
-        </div>
+        <section>
+          <p>
+            Register on TiltCloud to share under your name, view, and delete
```
landism: oh man
potential can o' worms time
have we settled on "TiltCloud" vs "Tilt Cloud"? I've been using "Tilt Cloud"

(not important for this PR)

web/src/ShareSnapshotModal.tsx
```diff
@@ -12,62 +12,108 @@ type props = {
...
+            snapshots.
```
landism: *your snapshots

(or maybe "to share under your name and manage your snapshots"...the "view" feels weird to me, since you can view them without registering)

web/src/ShareSnapshotModal.tsx
```diff
@@ -12,62 +12,108 @@ type props = {
...
         onRequestClose={this.props.handleClose}
         isOpen={this.props.isOpen}
         className="ShareSnapshotModal"
       >
-        <div className="ShareSnapshotModal-pane">{link}</div>
+        <h2 className="ShareSnapshotModal-title">Share a Shapshot</h2>
+        <div className="ShareSnapshotModal-pane">
+          <p>
+            Let anyone explore your current Tilt session with an interactive
+            snapshot.
```
maiamcc: i don't feel strongly but maybe this gets more clearly at the fact that the state, while interactive, is static:
"Let anyone explore your dev environment with an interactive snapshot of your current Tilt state."

web/src/ShareSnapshotModal.tsx
```diff
@@ -12,62 +12,108 @@ type props = {
...
+              value={cookies.get("Tilt-Token")}
+            />
+          </form>
+        </section>
       )
     } else {
-      return <div>Sharing snapshots as {this.props.tiltCloudUsername}.</div>
+      return (
+        <p>
+          View and delete snapshots on{" "}
```
landism: as above

web/src/ShareSnapshotModal.tsx
maiamcc: > Connect your Tilt install with a TiltCloud account...

web/src/__snapshots__/ShareSnapshotModal.test.tsx.snap
hyu: This is great! This is precisely what I was thinking through as well, and trying to get a succinct version that sounds accessible and breezy is...tricky. I won't tinker further in this PR but I think we can have a story for "final round of copyediting" to address these branding and technical accuracy issues.

web/src/ShareSnapshotModal.tsx
hyu: Duly noted...I am ingesting this useful feedback and mulling on it. I think we should just have one last copy round in a separate PR.

web/src/ShareSnapshotModal.tsx
hyu: I agree, the "Sign Up" is misleading. I will consider a better way to express that you're linking your Tilt instance.

web/src/ShareSnapshotModal.tsx
hyu: I was playing with "session" vs "state" and you just nudged me into thinking the latter may be better.

PR #2176, opened 2019-09-10
events: fix k8s case in log message

landism:


landism:
ehhh, I thought it was definitely K8S, but turns out people are all over the place on the web.

PR #2177, opened 2019-09-10
engine: record baseline restarts when we first see a pod

nicks:
Hello @maiamcc, @landism,

Please review the following commits I made in branch nicks/ch3417/baseline:

4dc0b7b9d1e0d8f5cc3f0be1a2e8abaff08e6c2f (2019-09-10 11:51:35 -0400)
engine: record baseline restarts when we first see a pod

### DIFF
--- a/internal/engine/pod.go
+++ b/internal/engine/pod.go
@@ -16,14 +16,15 @@ import (
 )
 
 func handlePodChangeAction(ctx context.Context, state *store.EngineState, action PodChangeAction) {
-	pod := action.Pod
-	mt, podInfo := ensureManifestTargetWithPod(state, action)
-	if mt == nil || podInfo == nil {
+	mt := matchPodChangeToManifest(state, action)
+	if mt == nil {
 		return
 	}
 
+	pod := action.Pod
 	ms := mt.State
 	manifest := mt.Manifest
+	podInfo, isNew := trackPod(ms, action)
 	podID := k8s.PodIDFromPod(pod)
 	if podInfo.PodID != podID {
 		// This is an event from an old pod.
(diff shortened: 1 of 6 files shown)


internal/store/runtime_state.go
```diff
@@ -102,9 +102,9 @@ type Pod struct {
 
 	Containers []Container
 
-	// We want to show the user # of restarts since pod has been running current code,
-	// i.e. OldRestarts - Total Restarts
-	OldRestarts int // # times the pod restarted when it was running old code
+	// We want to show the user # of restarts since some baseline time
+	// i.e. BaselineRestarts - Total Restarts
```
landism: ha, this was backwards before:
Total Restarts - BaselineRestarts

internal/engine/pod.go
```diff
@@ -40,6 +41,11 @@ func handlePodChangeAction(ctx context.Context, state *store.EngineState, action
 
 	oldRestartTotal := podInfo.AllContainerRestarts()
 	podInfo.Containers = podContainers(ctx, pod)
+	if isNew {
```
landism: this block read pretty confusingly to me: "if isNew" followed by "we want to do stuff for existing pods" while "new" and "existing" sound pretty antonym-y

maybe something like "// we want to discount any restarts that occurred before we first saw the pod"?

internal/store/runtime_state.go
nicks: done

internal/engine/pod.go
nicks: i tried to expand the comment a bit

PR #2178, opened 2019-09-10
tilt: adjust snapshot modal copy

landism:
This attempts to address comments made on #2160.

It now looks like this:

image

### DIFF
--- a/web/src/ShareSnapshotModal.tsx
+++ b/web/src/ShareSnapshotModal.tsx
@@ -26,8 +26,8 @@ export default class ShareSnapshotModal extends PureComponent<props> {
         <h2 className="ShareSnapshotModal-title">Share a Shapshot</h2>
         <div className="ShareSnapshotModal-pane">
           <p>
-            Get a link to share a snapshot of the current state of your Tilt
-            session.
+            Get a link to a snapshot -- a browsable, sharable view of the
+            current state of your Tilt session.
           </p>
           {link}
           <p className="ShareSnapshotModal-user">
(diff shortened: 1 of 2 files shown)


web/src/ShareSnapshotModal.tsx
```diff
@@ -80,7 +80,7 @@ export default class ShareSnapshotModal extends PureComponent<props> {
       return (
         <section>
           <p>
-            Register on TiltCloud to share under your name, view, and delete
+            Connect Tilt to TiltCloud to share under your name and manage your
```
maiamcc: Maybe: "Connect this Tilt install to TiltCloud"?

web/src/ShareSnapshotModal.tsx
```diff
@@ -26,8 +26,8 @@ export default class ShareSnapshotModal extends PureComponent<props> {
         <h2 className="ShareSnapshotModal-title">Share a Shapshot</h2>
         <div className="ShareSnapshotModal-pane">
           <p>
-            Let anyone explore your current Tilt session with an interactive
-            snapshot.
+            Get a link to share a snapshot of the current state of your Tilt
```
maiamcc: Seems obtuse to me. Maybe:

> Get a link to a snapshot -- a browsable, sharable view of the current state of your Tilt session

Or something.

web/src/ShareSnapshotModal.tsx
landism: I'd considered that but figured it felt excessively precise. Like, it's only relevant if you switch to another install, and at that point maybe you're not surprised if you need to again click connect to TiltCloud?

PR #2185, opened 2019-09-11
docker: rejigger the docker API a bit so that engine tests don't connect to the docker server for auth info

nicks:
Hello @jazzdan,

92f830e71f31fbf06e4b302e1844d913cca180db (2019-09-11 14:47:55 -0400)
docker: rejigger the docker API a bit so that engine tests don't connect to the docker server for auth info


nicks:
also tweaked the logging a bit while i was here

PR #2199, opened 2019-09-16
tiltfile: remove fast_build

dbentley:


internal/tiltfile/docker.go
```diff
@@ -207,11 +207,7 @@ func (s *tiltfileState) dockerBuild(thread *starlark.Thread, fn *starlark.Builti
 		return nil, err
 	}
 
-	// NOTE(maia): docker_build returns a fast build that users can optionally
-	// populate; if populated, we use it for in-place updates of this image
-	// (but use DockerBuild info for image builds)
-	fb := &fastBuild{s: s, img: r}
-	return fb, nil
+	return starlark.None, nil
```
maiamcc: if there are any users still relying on this behavior, this will break them in a hard-to-diagnose way:
> NoneType has no .add field or method
While more annoying, the more thorough way to do this would be to still have this return a `fast_build`, but have every method on that fast build throw an error.

internal/tiltfile/docker.go
```diff
@@ -350,25 +346,15 @@ func (b *customBuild) Hash() (uint32, error) {
...
 	default:
 		return nil, nil
 	}
 }
 
 func (b *customBuild) AttrNames() []string {
-	return []string{addFastBuildN}
-}
-
-func (b *customBuild) addFastBuild(thread *starlark.Thread, fn *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
```
maiamcc: not adding a deprecation error to this function whenever it's called?

internal/tiltfile/docker.go
```diff
@@ -390,61 +376,6 @@ func parseValuesToStrings(value starlark.Value, param string) ([]string, error)
 	return ignores, nil
 
 }
-func (s *tiltfileState) fastBuild(thread *starlark.Thread, fn *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
```
maiamcc: again: leave this method in but have it throw a deprecation error with a link to LiveUpdate docs?

landism:
>  IMO we should leave all the fast build methods in and have them throw "FastBuild has been deprecated, use LiveUpdate instead " errors.

+1, I think the deprecation process should always include this step. Warnings will get ignored indefinitely. This gives users who have been ignoring the warning a next step now that it's an error and they can't ignore it.

maiamcc:
bump!

internal/tiltfile/docker.go
```diff
@@ -350,25 +346,15 @@ func (b *customBuild) Hash() (uint32, error) {
...
 	}
 }
 
 func (b *customBuild) AttrNames() []string {
-	return []string{addFastBuildN}
-}
-
-func (b *customBuild) addFastBuild(thread *starlark.Thread, fn *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
-	return &fastBuild{s: b.s, img: b.img}, nil
+	return []string{}
```
maiamcc: should leave this method in and have it throw the deprecation error

internal/tiltfile/docker.go
```diff
@@ -15,10 +15,7 @@ import (
 	"github.com/windmilleng/tilt/pkg/model"
 )
 
-const fastBuildDeprecationWarning = "FastBuild (`fast_build`; `add_fast_build`; " +
-	"`docker_build(...).add(...)`, etc.) will be deprecated **ON THURS. 9/12/19**; " +
-	"you can use Live Update instead! See https://docs.tilt.dev/live_update_tutorial.html " +
-	"for more information. If Live Update doesn't fit your use case, let us know."
+var fastBuildDeletedErr = fmt.Errorf("fast_build is deleted; check out live_update (e.g. https://docs.tilt.dev/live_update_tutorial.html")
```
landism: at the risk of getting soapboxy:

This message feels very "we did a thing, deal with it."-y, which, as a user, gives me low confidence that tool maintainers are invested in / aware of the pain their users experience.

how about: "Sorry, fast_build is no longer supported. This functionality is now provided via live_update: https://docs.tilt.dev/live_update_tutorial.html. If you run into problems, let us know: https://tilt.dev/contact"?

internal/tiltfile/docker.go
maiamcc: currently you get:
>        Error: custom_build has no .add_fast_build field or method

dbentley: How's this? It will error when you ask for "add_fast_build".

PR #2201, opened 2019-09-16
model: LocalTarget [ch3471]

maiamcc:
Hello @dbentley,

Please review the following commits I made in branch maiamcc/local-target-model:

ac9499f4254454b3684cec14b1dca1d6df12aa42 (2019-09-16 17:20:29 -0400)
model: LocalTarget [ch3471]

### DIFF
--- a/pkg/model/local_target.go
+++ b/pkg/model/local_target.go
@@ -54,7 +54,7 @@ func (lt LocalTarget) Dependencies() []string {
 }
 
 func (lt LocalTarget) LocalRepos() []LocalGitRepo {
-	return lt.LocalRepos()
+	return lt.repos
 }
 
 func (lt LocalTarget) Dockerignores() []Dockerignore {
--- a/pkg/model/local_target.go
+++ b/pkg/model/local_target.go
@@ -1,0 +1,66 @@
+package model
+
+import (
+	"fmt"
+
+	"github.com/windmilleng/tilt/internal/sliceutils"
+)
+
+type LocalTarget struct {
+	Cmd  Cmd
+	Deps []string
+
+	dockerignores []Dockerignore
+	repos         []LocalGitRepo
+}
+
+var _ TargetSpec = LocalTarget{}
+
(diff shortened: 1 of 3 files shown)


pkg/model/local_target.go
```diff
@@ -0,0 +1,66 @@
...
+
+import (
+	"fmt"
+
+	"github.com/windmilleng/tilt/internal/sliceutils"
+)
+
+type LocalTarget struct {
+	Cmd  Cmd
+	Deps []string
```
nicks: i'm surprised this is a public field?

pkg/model/local_target.go
```diff
@@ -0,0 +1,66 @@
...
+
+	dockerignores []Dockerignore
+	repos         []LocalGitRepo
+}
+
+var _ TargetSpec = LocalTarget{}
+
+func NewLocalTarget(cmd Cmd, deps []string) LocalTarget {
```
nicks: it would be good to clarify that `deps` have to be absolute directories (since they're fulfilling the Dependencies() contract)

pkg/model/local_target.go
```diff
@@ -0,0 +1,66 @@
...
+	return LocalTarget{Cmd: cmd, Deps: deps}
+}
+
+func (lt LocalTarget) WithRepos(repos []LocalGitRepo) LocalTarget {
+	lt.repos = append(append([]LocalGitRepo{}, lt.repos...), repos...)
+	return lt
+}
+
+func (lt LocalTarget) WithDockerignores(dockerignores []Dockerignore) LocalTarget {
```
dbentley: Are Dockerignores needed? Or GitRepos?

pkg/model/local_target.go
maiamcc: oh, yeah fair.

pkg/model/local_target.go
maiamcc: yeah, because we set up watches for the deps just like we do for custom_build.deps

Issue #2180, opened 2019-09-11
tilt doesn't work if the docker server isn't running


nicks:
from the K8s tilt channel:

> I'm trying to use minikube on mac as my k8s cluster.  When I run tilt up, I get the following error
Error: NewDockerVersion: Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?

confirmed that it works on linux but fails on macos, so not sure what's going on

landism:
The error is coming from [NewDockerClient](https://github.com/windmilleng/tilt/blob/5e29b5101a86fdda88f5f1dc8d528eb71da8a820/internal/docker/client.go#L127) via [ProvideLocalCli](https://github.com/windmilleng/tilt/blob/8c7200765a66752c8ccc602183422ffc9e76c9ca/internal/docker/clients.go#L21)

I'm not familiar with this code and am not sure of its purpose, but AIUI, we instantiate both a LocalClient and a ClusterClient on startup, and then combine them into a SwitchClient that picks whether to use LocalClient or ClusterClient based on whether we're the orchestrator is k8s.

In the case of minikube, we don't actually *use* the LocalClient, but we instantiate it anyway. Since there is no local docker daemon running, that errors, and fatally.

Two options that come to mind:
* give the switch client cluster/local factories and don't create the client until after the switch has determined which one it wants
* change ProvideLocalCli to, on error, return an exploding client instead of erroring immediately.

> confirmed that it works on linux but fails on macos, so not sure what's going on

Does it work on linux if you stop your local docker daemon? If my understanding is correct, this is not a macos vs linux issue, but a "is there a local docker daemon running or not" issue.

nicks:
oh, wait, i'm confused. so it works ok on MacOS if your local docker client is running?

landism:
If I start docker for desktop, then I can run Tilt.
If I stop docker for desktop, then I get the error in this issue.
If I stop docker for desktop, and run `eval $(minikube docker-env)`, then I can run Tilt.

nicks:
OK, so there's a couple separate requirements that are conflicting in a weird way:
1) Sometimes Docker gets hosed for some reason. Users really wanted us to do a Docker healthcheck on startup, so that Tilt wouldn't just hang.
2) Tilt doesn't know which Docker client you want to use (the docker-compose client or the k8s client) until after we've executed the Tiltfile

I don't really want to delay the healthcheck until after Tiltfile loads (since this could take a long time).

[...]

PR #2203, opened 2019-09-17
docker: If the docker server fails to connect, don't report it until we're sure we need it. Fixes issue #2180

nicks:
Hello @jazzdan, @landism,

Please review the following commits I made in branch nicks/issue2180:

d0fca8b009e41707845cd4acec2b0fd2f217f4c9 (2019-09-17 12:40:20 -0400)
docker: If the docker server fails to connect, don't report it until we're sure we need it. Fixes issue #2180

### DIFF
--- a/internal/docker/client.go
+++ b/internal/docker/client.go
@@ -58,6 +58,8 @@ var sessionSharedKey = identity.NewID()
 
 // Create an interface so this can be mocked out.
 type Client interface {
+	CheckConnected() error
+
 	// If you'd like to call this Docker instance in a separate process, these
 	// are the environment variables you'll need to do so.
 	Env() Env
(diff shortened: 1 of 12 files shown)


internal/docker/client.go
```diff
@@ -116,29 +118,30 @@ type Cli struct {
 	env       Env
 }
 
-func NewDockerClient(ctx context.Context, env Env) (*Cli, error) {
+func NewDockerClient(ctx context.Context, env Env) Client {
 	opts, err := CreateClientOpts(ctx, env)
 	if err != nil {
-		return nil, errors.Wrap(err, "CreateClientOpts")
+		return newExplodingClient(err)
```
landism: hm, is it it not worth `errors.Wrap`s throughout to add context to the client explosion?

internal/docker/clients.go
```diff
@@ -10,14 +10,26 @@ type ClusterClient Client
...
+		cClient = ClusterClient(lClient)
+	} else {
+		cClient = NewDockerClient(ctx, Env(cEnv))
 	}
-	result, err := NewDockerClient(ctx, Env(cEnv))
-	return ClusterClient(result), err
+
+	// The LocalClient is the docker server from docker env variables.
+	// The ClusterClient is the docker server from kubectl configs.
+	// If neither of them work, we can fail on startup.
```
landism: clever!

landism: ...at least until we have a tiltfile with only local resources. oy.

Issue #1850, opened 2019-07-12
Feature request: ability to dismiss alerts


epotex:
Since right now we have no way to control the order of the containers, We get alerts from applications that wait for DB  to start - when the DB start all is working but the alert is still there

will it be possible to add a feature that will allow users in the web UI to dismiss alerts?

nicks:
Oooh, we've talked about this a bunch. It's a pretty common use-case (see https://github.com/windmilleng/tilt/issues/1846 for example).

The thing we get stuck on is how persistent we should be about the dismissal. If I dismiss the alert once, and the same error happens again, should the error come back? Or should it be snoozed for N minutes? How smart do we need to be about determining it's the "same" error? And is there a way to snooze it in the Tiltfile so that everyone on your team knows not to worry about it?

But maybe we're overthinking it and there's something simple we can do that works for most cases.

drubin:
The problem right now that I can see is you use the cluster state as the source of truth for alerts, This coupled with the "restarted" flag makes it impossible.
 
I had two ideas.

1) Store an  immutable event log. This way when you dismiss the alerts you dismiss them from a point in time and all future events (i.e increase in restart would create another event)

2) You base your healthy state from the k8s readiness/liveness checks which were added in https://github.com/windmilleng/tilt/commit/ad34add39b7d6d9c50605b4b7529e2e6f0ff2f7f 

If your application is currently healthy the alert could automatically be dismissed. 

The issue here is that in a distributed system these types of dependancies are normal and I don't believe you should solve this by trying to hide this race condition. Docker compose and Garden try solve this with "depends on" but the issue with this approach is they only "hide" the explicitly race condition and allow developers to ignore this. i.e avoiding proper health checks and error handling. 

just my 2 cents.

drubin:
btw we are hacking this state by deleting the pods manually(via a script) that have restarts so we can then easily see if a restart even happens later. 

This is horrid but its the only way the tilt dashboard is useful as part of a normal distributed startup flow with an application with liveness/readiness checks

nicks:
ugh, sorry to hear that, let's see if we can do something quick just to solve the "reset restarts" case

nicks:
We added a button to dismiss restart alerts once the pod is running; we'll see how that works.

PR #2205, opened 2019-09-17
web: dismiss pod restart alerts. Fixes issue #1850

nicks:
Hello @hyu, @jazzdan,

Please review the following commits I made in branch nicks/issue1850:

5510bd0b9b3f36710b2bd533a843c27ef5866831 (2019-09-17 15:21:43 -0400)
web: dismiss pod restart alerts. Fixes issue #1850

### DIFF
--- a/internal/hud/server/server.go
+++ b/internal/hud/server/server.go
@@ -15,6 +15,7 @@ import (
 	tiltanalytics "github.com/windmilleng/tilt/internal/analytics"
 	"github.com/windmilleng/tilt/internal/cloud"
 	"github.com/windmilleng/tilt/internal/hud/webview"
+	"github.com/windmilleng/tilt/internal/k8s"
 	"github.com/windmilleng/tilt/internal/store"
 	"github.com/windmilleng/tilt/pkg/assets"
 	"github.com/windmilleng/tilt/pkg/model"
@@ -35,6 +36,13 @@ type analyticsOptPayload struct {
 
 type triggerPayload struct {
 	ManifestNames []string `json:"manifest_names"`
+}
+
+type actionPayload struct {
+	Type            string             `json:"type"`
+	ManifestName    model.ManifestName `json:"manifest_name"`
+	PodID           k8s.PodID          `json:"pod_id"`
+	VisibleRestarts int                `json:"visible_restarts"`
 }
 
 type HeadsUpServer struct {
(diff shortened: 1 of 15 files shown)


nicks:
Screenshot from 2019-09-17 15-19-36

Need to write some tests for this but here's what it looks like

nicks:
might change the text to be red or glowing or smth to call more attention to it

web/src/AlertPane.tsx
```diff
@@ -53,13 +26,56 @@ class AlertPane extends PureComponent<AlertsProps> {
...
     }
 
     return <section className="AlertPane">{el}</section>
   }
+
+  renderAlerts() {
+    let formatter = timeAgoFormatter
+    let alertElements: Array<JSX.Element> = []
+    let resources = this.props.resources
+    let isLocal = this.props.pathBuilder.isLocal()
```
jazzdan: What does `isLocal` mean here?

web/src/alerts.ts
```diff
@@ -7,6 +7,7 @@ export type Alert = {
   msg: string
   timestamp: string
   resourceName: string
+  dismissHandler?: () => void
```
jazzdan: When is this not set on the alert?

internal/hud/server/server.go
```diff
@@ -175,6 +184,30 @@ func (s *HeadsUpServer) HandleAnalytics(w http.ResponseWriter, req *http.Request
 	}
 }
 
+func (s *HeadsUpServer) DispatchAction(w http.ResponseWriter, req *http.Request) {
```
jazzdan: Interesting idea taking all of the action dispatchers in to one endpoint!

Are we worried about how non-homogeneous the inputs to this endpoint will be? What happens when `actionPayload` has a bunch of possible fields which are used for different actions that it can dispatch?

nicks:
ya i think i like it a bit better in red
Screenshot from 2019-09-17 16-01-29

web/src/AlertPane.tsx
nicks: on localhost

web/src/alerts.ts
nicks: most alerts don't have a way to dismiss them

internal/hud/server/server.go
nicks: nope! keep in mind that we can guarantee that the client and server are always in-sync. So it doesn't really matter. We can break backwards-compatibility

nicks:
ok, updated with a bunch of tests and red test, PTAL?

nicks:
here's what it looks like now:
Screenshot from 2019-09-18 12-32-06

PR #2206, opened 2019-09-17
hud: local targets show up in webview

jazzdan:


pkg/model/manifest.go
```diff
@@ -85,6 +85,11 @@ func (m Manifest) IsLocal() bool {
 	return ok
 }
 
+func (m Manifest) WithLocalTarget(lTarget LocalTarget) Manifest {
```
maiamcc: can't you use `WithDeployTarget` instead? (`LocalTarget` implements `TargetSpec`)

jazzdan: Yup, good call!

jazzdan:
re: conversation with @dbentley elsewher, this does cause the target to show up in the UI:

Screenshot from 2019-09-18 10-04-12

PR #2214, opened 2019-09-18
tiltfile: add local_resource

jazzdan:


internal/tiltfile/tiltfile_state.go
```diff
@@ -1177,3 +1181,24 @@ func (s *tiltfileState) setTeam(thread *starlark.Thread, fn *starlark.Builtin, a
...
+	var result []model.Manifest
+
+	for _, r := range s.localResources {
+		mn := model.ManifestName(r.name)
+		tm, err := starlarkTriggerModeToModel(s.triggerModeForResource(r.triggerMode))
+		if err != nil {
+			return nil, err
+		}
+		lt := model.NewLocalTarget(r.cmd, r.deps)
+		m := model.Manifest{
```
maiamcc: should populate repos/dockerignores [the same way we do for image targets](https://github.com/windmilleng/tilt/blob/009ef1c491001eeb9cb329d97db25378cdbaa251/internal/tiltfile/tiltfile_state.go#L1023) -- the `deps` are the things we feed into the filewatcher, so if one of the deps is e.g. a git repo, we want to ignore the `.git` directory.

internal/tiltfile/tiltfile_test.go
```diff
@@ -3900,6 +3900,23 @@ local('echo hi')
 	}
 }
 
+func TestLocalResource(t *testing.T) {
+	f := newFixture(t)
+	defer f.TearDown()
+
+	f.file("Tiltfile", `
+local_resource("test", "echo hi", ["/foo/bar"])
```
maiamcc: imo more interesting to pass a relative dependency and make sure it gets converted to an abs path.

internal/tiltfile/tiltfile_state.go
maiamcc: (and check in a test that they got populated correctly)

jazzdan: Hmm, I don't think it makes sense to popular dockerignores. This local resource could be anything, and I might want it to be able to operate on files that I _don't_ want to explicitly include in my build contexts.

PR #2215, opened 2019-09-18
dockercompose: move config yaml parsing into the dockercompose package

nicks:
Hello @maiamcc,

11af908036cee0006e66db183d5fd0093f434837 (2019-09-18 13:01:50 -0400)
dockercompose: move config yaml parsing into the dockercompose package


nicks:
I think the code was originally very simple, so made more sense colocated with the code that used it. But it's obviously grown a lot!

Issue #1935, opened 2019-07-30
auto-scrolling should stop when the user scrolls up


nicks:
in the #tilt channel, Chuck reported that he was using the Tilt logs view. The service has lots of logs. He's using Safari

When he scrolled up, the UI would autoscroll back down to the new logs.

Expected behavior: when you scroll up, the UI should stop autoscrolling

I have not been able to confirm this behavior, but I do not have my Mac with me today.

jazzdan:
Hmm, it works for me on Safari.

jazzdan:
I'm going to see @chuckha in person tomorrow and I'll try to figure out what's going on.

chuckha:
does not reproduce when dan is watching

feel free to close, thank you for looking into it !

maiamcc:
Multiple other folks have reported this :-/

from #tilt :
```tmc Yesterday at 4:00 PM
running 0.10.7 — if a container is spewing a ton of output the web view becomes basically unusable since contents will scroll off the page
9 replies

maia  21 hours ago
oh dear.
a. is this behavior new, or were you experiencing it before you updated?
b. what browser/version are you on?
c. so when you scroll up even just a bit, the auto-scroll behavior is SUPPOSED to turn off so that you stay looking at the point at the logs you’re currently at. I assume that’s not the case for you? (I.e. if you scroll up a bit, logs continue spewing/scrolling you back down?)

Matt Brown  17 hours ago
I have the same browser behavior. auto-scrolling continues even when you scroll up.
v. 0,10,6
Firefox browser (edited) 

maia  1 hour ago
:disappointed: :disappointed: :disappointed:

maia  1 hour ago
@dmiller did you say you could replicate this?

maia  1 hour ago
also @Michael Hatch iirc you were seeing this autoscrolling bug — what browser are you on? (are you still seeing it?)

Michael Hatch  1 hour ago
chrome

dmiller  1 hour ago
I was able to replicate it sometimes on Firefox

dmiller  1 hour ago
maybe it has to do with the rate of messages

Matt Brown  5 minutes ago
I also noticed it when the messages were scrolling super fast
```

landism:
this might be picking nits, and is really a technical distinction between two cases that look pretty much the same to users, but I think it's at least worth improving clarity among Tilt devs:
1. autoscrolling firing incorrectly - autoscrolling is where the UI modifies the scroll position to make the last line of the log visible. The scrollbar changes position to the bottom.
[...]

PR #2221, opened 2019-09-19
logs: slow down truncation frequency

landism:
Mitigates at least one case of #1935 

### Problem

When a log exceeds the max buffer size, we remove old lines until we're under the limit. This often happens when a service is rapidly logging a lot of messages.
This has the effect that every time the server logs a new message, we shift the full log buffer. The log view in the web UI scrolls very quickly and it is difficult to read and pretty much impossible to copy/paste.

### Solution

When a log exceeds the max buffer size, instead of truncating to be just under the limit, we truncate to be at half of the limit. This means instead of constantly (with every log message) scrolling the whole buffer a little, we more rarely (with every 60k written) scroll the buffer a whole lot. even if users are logging an extreme 1k/second, this will still give them a minute between scrolls, which might still be suboptimal but should at least make interaction with the log possible.

### DIFF
--- a/pkg/model/log.go
+++ b/pkg/model/log.go
@@ -12,6 +12,12 @@ import (
 // A render time of ~40ms was about when the interface started being noticeably laggy to me.
 const maxLogLengthInBytes = 120 * 1000
 
+// After a log hits its limit, we need to truncate it to keep it small
+// we do this by cutting a big chunk at a time, so that we have rarer, larger changes, instead of
+// a small change every time new data is written to the log
+// https://github.com/windmilleng/tilt/issues/1935#issuecomment-531390353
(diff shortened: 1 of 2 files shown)


pkg/model/log.go
```diff
@@ -143,13 +149,17 @@ type LogEvent interface {
 }
 
 func ensureMaxLength(lines []logLine) []logLine {
-	bytesLeft := maxLogLengthInBytes
+	bytesSpent := 0
```
nicks: I wonder if we should have a line limit instead of a bytes limit

landism: Yeah, this feels like one of those things where, given the choice, that state would be preferable, but it's not clearly worth the time to switch to that state.

PR #2222, opened 2019-09-19
tiltfile: remove local_git_repo

jazzdan:


jazzdan:
There are still a lot of users for this function and open source code that references it. I'm going to let it be for now.

Issue #2210, opened 2019-09-18
Docker Compose depends_on


fgallese:
Hello,

I am trying to use tilt with our existing docker-compose files, and I am having issues with dependencies.

For example, I have some apps that declare a "depends_on" in order to not be started before a required database. See: https://docs.docker.com/compose/compose-file/compose-file-v2/#depends_on

I see that this is not being respected by tilt. Is there a way for me to make tilt build/start a docker-compose image/service before another ?

I am new with tilt so maybe I am missing something ? If so please let me know.. I didn't find anything while searching previous issues and in the doc.

Thank you.

nicks:
Hmmm... I'm surprised that Tilt isn't respecting the `depends_on`. 

I have a suspicion that this should be easy to fix by just changing the builder to sort by `depends_on` here: https://github.com/windmilleng/tilt/blob/master/internal/engine/buildcontroller.go#L36

Lemme poke a bit at it

nicks:
Hmmm...I was not able to reproduce this bug

As far as I can tell, Tilt uses `docker-compose config --services` to sort the services and start them in that order. In every sample project I tried, this properly handled `depends_on` (e.g., in https://github.com/dockersamples/example-voting-app, Tilt will start 'redis' and 'db' before 'worker')

@fgallese Do you have a sample project I could try that demonstrates what you're seeing?

fgallese:
Thank you for your time reviewing this.

Well, after some more testing, I think I've realized what is going on.. and it is probably caused by my own misunderstanding of the way Tilt works.

**In short**: I am setting up a Tiltfile where I define some docker-compose files using `docker_compose()`. Also, in the same Tiltfile I define a `docker_build()` for a service _myServiceB_.
In one of the docker-compose files, I have a service _myServiceA_ with a `depends_on` _myServiceB_ (both of this services are included in one of the docker-compose files loaded with  `docker_build()`)

What I was expecting to see, is that _myServiceA_ is not run until _myServiceB_ is built and up. But this is not happening. I guess that when Tilt has a docker_build() definition it always runs it _after_ every other service from docker_compose(). Is this correct ?

This is a simplified version of my setup:

docker-compose.myServiceA.yml:

``` docker-compose.yml
services:
  myServiceA:
    extends:
      file: docker-compose.base.yml
      service: myServiceA_base # here I define some stuff like the image (myRegistry/myServiceA)
    restart: unless-stopped
    networks:
      - a_network
    ports:
      - 3000:7000
[...]

PR #2223, opened 2019-09-19
engine: respect the native docker-compose order. Fixes part of issue #2210

nicks:
Hello @maiamcc,

Please review the following commits I made in branch nicks/issue2210/order:

92f95f64a721e13cd855de220b4c7b85fb565020 (2019-09-19 17:07:12 -0400)
engine: respect the native docker-compose order. Fixes part of issue #2210

### DIFF
--- a/internal/engine/buildcontroller_test.go
+++ b/internal/engine/buildcontroller_test.go
@@ -655,3 +655,32 @@ func TestBuildControllerUnresourcedYAMLFirst(t *testing.T) {
 	}
 	assert.Equal(t, expectedBuildOrder, observedBuildOrder)
 }
+
+func TestBuildControllerRespectDockerComposeOrder(t *testing.T) {
+	f := newTestFixture(t)
+	defer f.TearDown()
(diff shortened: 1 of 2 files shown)


internal/engine/buildcontroller_test.go
```diff
@@ -655,3 +655,27 @@ func TestBuildControllerUnresourcedYAMLFirst(t *testing.T) {
 	}
 	assert.Equal(t, expectedBuildOrder, observedBuildOrder)
 }
+
+func TestBuildControllerRespectDockerComposeOrder(t *testing.T) {
+	f := newTestFixture(t)
+	defer f.TearDown()
+
+	sancho := NewSanchoFastBuildDCManifest(f)
```
maiamcc: can you make a note somewhere of what the order would be if we weren't respecting DC order and let the k8s-unbuilt logic handle this?

nicks: comment added

PR #2225, opened 2019-09-19
hud: LocalResource correctly shows status in HUD [ch3475]

maiamcc:
Hello @jazzdan, @dbentley,

Please review the following commits I made in branch maiamcc/local-resource-tui:

2f516f695f8b56063f5e738589f4a6ce26e16705 (2019-09-19 18:13:37 -0400)
tests

92830b3f27ea0c4b8736b951717b6d1720465a50 (2019-09-19 17:20:56 -0400)
hud: LocalResource correctly shows status in HUD [ch3475]


internal/hud/view/view.go
```diff
@@ -66,6 +66,17 @@ func (YAMLResourceInfo) resourceInfoView()              {}
...
 
+type LocalResourceInfo struct {
+}
+
+var _ ResourceInfoView = LocalResourceInfo{}
+
+var LocalResourceStatusPlaceholder = "local-resource-ok"
+
+func (LocalResourceInfo) resourceInfoView()     {}
+func (LocalResourceInfo) RuntimeLog() model.Log { return model.Log{} }                    // no runtime log, only build log
```
jazzdan: This is a bit weird, but I'm not sure that it matters, in that really a local resource _only_ has a runtime log, and no build log. But if it's just naming and doesn't semantically matter then it's neither here nor there.

maiamcc: Aaah, i beg to differ! Because of weird semantics, the "build" step is what happens when the resource goes thru our BuildAndDeployer--i.e. we run the command during the BUILD step. That's where any logs/errors will come out. Runtime log/runtime status indicate the life of the resource _beyond_ the "build" step, and in the case of LocalResource currently, there is none.

jazzdan: Right, OK!

internal/store/engine_state.go
```diff
@@ -613,6 +613,16 @@ func resourceInfoView(mt *ManifestTarget) view.ResourceInfoView {
 			PodLog:             pod.CurrentLog,
 			YAML:               mt.Manifest.K8sTarget().YAML,
 		}
+	} else if mt.Manifest.IsLocal() {
```
jazzdan: These doesn't even need to be `else`s since they return. If you reach the end of the function without returning then we can `panic`.

PR #2228, opened 2019-09-20
hud/webview: support local resources

jazzdan:


jazzdan:
Screen Shot 2019-09-20 at 1 06 56 PM

internal/hud/webview/convert.go
```diff
@@ -144,7 +144,11 @@ func resourceInfoView(mt *store.ManifestTarget) ResourceInfoView {
 		dc := mt.Manifest.DockerComposeTarget()
 		dcState := mt.State.DCRuntimeState()
 		return NewDCResourceInfo(dc.ConfigPaths, dcState.Status, dcState.ContainerID, dcState.Log(), dcState.StartTime)
-	} else {
+	}
```
maiamcc: I updated this in my PR -- [now it's a type switch on mt.State](https://github.com/windmilleng/tilt/blob/master/internal/store/engine_state.go#L603) (with a dummy state for LocalResource), which seems cleaner. (and i still return an empty k8sRuntimeState as my base case, instead of panicking -- technically less correct but also less likely to loudly break ppl? idk, I could go either way but @dbentley leaned towards the non-panic so I did that.

internal/hud/webview/view.go
```diff
@@ -73,6 +73,14 @@ func (YAMLResourceInfo) resourceInfoView()              {}
 func (yamlInfo YAMLResourceInfo) RuntimeLog() model.Log { return model.NewLog("") }
 func (yamlInfo YAMLResourceInfo) Status() string        { return "" }
 
+type LocalResourceInfo struct{}
+
+var _ ResourceInfoView = LocalResourceInfo{}
+
+func (LocalResourceInfo) resourceInfoView()     {}
+func (LocalResourceInfo) RuntimeLog() model.Log { return model.NewLog("") }
```
maiamcc: worth commenting why these are nada?

jazzdan:
@maiamcc yeah I'll write some JS tests in a separate PR.

internal/hud/webview/convert.go
jazzdan: I wish Go had tagged unions, then the compiler could enforce that we check all possible values for an enum when we switch over them. In the meantime I like the panic solution since it makes it more obvious where we need to change the code next time we add a resource type, and it's basically impossible for us to not notice it heh.

maiamcc: fair nough. want to change my stupid return statement back to a panic for consistency?

jazzdan: Sure, will do in a separate PR.

PR #2230, opened 2019-09-20
webview: remove yaml

nicks:
Hello @landism,

Please review the following commits I made in branch nicks/logs:

36f6f4ba84f6decefa28068c2fc67a6c26777c64 (2019-09-20 15:26:43 -0400)
webview: remove yaml
If we bring this back, we would want to scrub it for secrets or use a `kubectl
describe`-like formatter that does scrubbing. But right now it's dead code.


nicks:
oops, lemme redo that

PR #2238, opened 2019-09-20
tiltfile: enable feature on partial evaluation

nicks:
Hello @landism, @jazzdan,

Please review the following commits I made in branch nicks/issue2062/error:

4c5bda41b9f6ce2a7b053da5b1986c8b723feb5c (2019-09-20 19:36:47 -0400)
tiltfile: enable feature on partial evaluation


nicks:
(not totally convinced this is a good idea...i went down a rabbit hole with secrets handling, and realizing that if we see a secret in the tiltfile, we want to make sure it gets scrubbed even if there's an error later...and realized that this implied we needed better partial-eval support)

PR #2240, opened 2019-09-23
tilt cloud: don't log http.Response.String()

landism:
no one wants to see this:
```
error checking tilt cloud status: &{500 Internal Server Error 500 HTTP/1.1 1 1 map[Content-Length:[22] Content-Type:[text/plain; charset=utf-8] Date:[Sun, 22 Sep 2019 20:00:32 GMT] X-Content-Type-Options:[nosniff] X-Rate-Limit-Duration:[1] X-Rate-Limit-Limit:[10.00] X-Rate-Limit-Request-Forwarded-For:[] X-Rate-Limit-Request-Remote-Addr:[127.0.0.1:44982]] 0xc000dda700 22 [] false false map[] 0xc0014fff00 <nil>
```

### DIFF
--- a/internal/cloud/cloud_username_manager.go
+++ b/internal/cloud/cloud_username_manager.go
@@ -91,7 +91,13 @@ func (c *CloudUsernameManager) CheckUsername(ctx context.Context, st store.RStor
 	}
 
 	if resp.StatusCode != http.StatusOK {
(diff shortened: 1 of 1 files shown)


nicks:
looks like this pr is causing segfaults? should it be rolled back?

landism:
> should it be rolled back?

done

PR #2244, opened 2019-09-23
tiltfile: make local_resource.deps required arg for now

maiamcc:
before we add support for force-running manual resources, a local_resource
without deps would be useless, so don't let a user do that


jazzdan:
A local resource without deps would run on start up though, correct?

maiamcc:
i mean, yes? I'm not sure it's usage we want to encourage until we actually build out stuff to support it well? idk 🤷‍♀

dbentley:
We already support update_mode=manual, right? It makes sense to allow no deps if update mode is manual, no?

maiamcc:
But you can only run it once, because w/o deps, there's no way to mark it as dirty, and unless it's dirty, clicking "run" does nothing (cuz we have no way to force-update a manifest)

dbentley:
Ah, I didn't understand how deeply through the model it runs that you need to have pending file changes to be triggered.

PR #2255, opened 2019-09-24
store: "permanent" errors dont close ui

jazzdan:
Looks like this:

Screen Shot 2019-09-25 at 12 05 58 PM

If a permanent, now called "fatal", error happens the Upper loop stops but the UI stays up and you can still browse through it.

Nothing is displayed in the web UI, and I'm not sure if that's a feature or a bug.

### DIFF
--- a/internal/engine/upper.go
+++ b/internal/engine/upper.go
@@ -133,6 +133,9 @@ func (u Upper) Init(ctx context.Context, action InitAction) error {
 }
 
 func upperReducerFn(ctx context.Context, state *store.EngineState, action store.Action) {
+	if state.FatalError != nil {
+		return
+	}
 	logAction, isLogAction := action.(store.LogAction)
 	if isLogAction {
 		handleLogAction(state, logAction)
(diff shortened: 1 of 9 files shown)


internal/engine/upper.go
```diff
@@ -133,6 +133,9 @@ func (u Upper) Init(ctx context.Context, action InitAction) error {
 }
 
 func upperReducerFn(ctx context.Context, state *store.EngineState, action store.Action) {
+	if state.FatalError != nil {
```
nicks: do you need to be able to handle the hud.ExitAction even in the error state? The DumpEngineStateAction might be good too

nicks:
can you also fix this to display something reasonable in the web ui? fine to do in a follow-up pr

jazzdan:
@nicks I will do that in a separate PR after you merge #2259

internal/engine/upper.go
jazzdan: ah good point!

nicks:
merged!

PR #2262, opened 2019-09-26
web: show fatal error as a modal in web UI

jazzdan:
Looks like this:

Screen Shot 2019-09-26 at 11 28 12 AM

### DIFF
--- a/internal/hud/webview/view.go
+++ b/internal/hud/webview/view.go
@@ -180,6 +180,8 @@ type View struct {
 
 	TiltCloudUsername   string
 	TiltCloudSchemeHost string
+
+	FatalError string
 }
 
 func (v View) Resource(n model.ManifestName) (Resource, bool) {
--- a/web/src/FatalErrorModal.tsx
+++ b/web/src/FatalErrorModal.tsx
@@ -12,7 +12,7 @@ type props = {
 export default class FatalErrorModal extends PureComponent<props> {
   render() {
     let showModal =
-      this.props.error !== null &&
+      Boolean(this.props.error) &&
       (this.props.showFatalErrorModal === ShowFatalErrorModal.Default ||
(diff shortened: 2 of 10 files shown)


web/src/types.ts
```diff
@@ -103,3 +103,9 @@ export type Snapshot = {
   SnapshotLink: string
   showSnapshotModal: boolean
 }
+
+export enum ShowFatalErrorModal {
+  Default,
+  Show,
```
jazzdan: This value isn't used yet but I'd like, in the future, to provide a way to show the Fatal Error Modal even if it has been dismissed, if there is a fatal error.

internal/hud/webview/view.go
```diff
@@ -180,6 +180,8 @@ type View struct {
 
 	TiltCloudUsername   string
 	TiltCloudSchemeHost string
+
+	FatalError string
```
nicks: i wonder if we should start being better about using omitempty

web/src/FatalErrorModal.tsx
```diff
@@ -0,0 +1,35 @@
...
+type props = {
+  error: string | null
+  showFatalErrorModal: ShowFatalErrorModal
+  handleClose: () => void
+}
+
+export default class FatalErrorModal extends PureComponent<props> {
+  render() {
+    let showModal =
+      this.props.error !== null &&
```
nicks: can we just use `this.props.error &&`?

i feel like we've had a lot of bugs due to bad use of strict equality. for example, in this case, `""` and `undefined` should not show the modal

internal/hud/webview/view.go
jazzdan: Oh interesting! Why?

nicks: 1) reduce payload size
2) reduce diffs when the snapshot json changes

PR #2277, opened 2019-09-30
Let TILT_SYNCLET_IMAGE variable override Docker image

therc:
This is the least invasive way I could think of to plumb a custom image name through, since command line parsing happens far from the `sidecar` package.

Given how `SyncletContainer` is built (once per execution), the only simple test would be for `getImageName()` and that would be trivial, hence no new tests.

All of the above to say that this could be a more complex change, but I will wait until the owners state a preference (if any).

### DIFF
--- a/internal/synclet/sidecar/sidecar.go
+++ b/internal/synclet/sidecar/sidecar.go
@@ -26,7 +26,7 @@ func getImageName() string {
 // When we deploy Tilt for development, we override this with LDFLAGS
 var SyncletTag = "v20190904"
 
-const SyncletImageEnvVar = "SYNCLET_IMAGE"
+const SyncletImageEnvVar = "TILT_SYNCLET_IMAGE"
 const SyncletImageName = "gcr.io/windmill-public-containers/tilt-synclet"
 const SyncletContainerName = "tilt-synclet"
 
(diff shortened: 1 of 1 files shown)


nicks:
can you add a bit more context on what you want to use this for?

is this so that you can inject a custom synclet image when you're developing tilt and/or the synclet?

is it so you can have a team that are all using a custom synclet image for developing apps?

something else?

therc:
I should have mentioned it earlier. This is for cases where you run Tilt
against a Kubernetes cluster with tight policies in terms of what Docker
registries can be pulled from.

therc:
Our use cases in general are perhaps a bit out of the ordinary or novel.
I'll contact Dan about having a chat. We're just South of Bryant Park.  😁

nicks:
aha, that makes sense. 

could you change the env variable to `TILT_SYNCLET_IMAGE`? I'm comfortable merging that, just to unblock you.

long-term, we would probably want to find a better way to plumb the image name, validate it, and log that the substitution is happening. Either via dependency injection (wire) or via a function in the Tiltfile. but a simple env variable hook seems OK for now.

nicks:
thanks! lgtm

PR #2279, opened 2019-09-30
More helpful message from verify_goimports target

therc:
I tried running the partial goimports command output by make. It ran and then
hung for a very long time.

Instead, suggest running `make goimports`.


nicks:
thanks for the fix!

PR #2280, opened 2019-09-30
New custom-synclet-release target to push to custom registry with custom tag

therc:
Let's say you want to build a private synclet image and you want to build it with an arbitrary tag that matches an existing tilt binary, not the current date... this change allows you to do that.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -150,6 +150,11 @@ synclet-release:
 	docker push $(SYNCLET_IMAGE):$(TAG)
 	sed -i 's/var SyncletTag = ".*"/var SyncletTag = "$(TAG)"/' internal/synclet/sidecar/sidecar.go
 
+custom-synclet-release:
+	$(eval TAG := $(if $(SYNCLET_TAG),$(SYNCLET_TAG),$(shell date +v%Y%m%d)))
+	docker build -t $(SYNCLET_IMAGE):$(TAG) -f synclet/Dockerfile .
+	docker push $(SYNCLET_IMAGE):$(TAG)
+
 release:
 	goreleaser --rm-dist
 


nicks:
Hmmm...this makefile target is for doing a synclet release. I don't think we want to make this read env variables.

have you looked at the `install-dev` and `synclet-dev` makefile targets? that seems closer to what you want

therc:
The only intended use case for this is pushing a private build of the synclet from source to a private registry (along with #2277). An alternative would be a docker pull/push command pair to republish the official image.

nicks:
Ah, OK. 

Ya, this would make more sense as a separate Makefile target/script to publish to a custom location (`synclet-custom-release`?) and/or copy a synclet image from one registry to another.

I'm worried that trying to shoehorn this into the existing Makefile target is going to confuse the person doing the release.

Makefile
```diff
@@ -150,6 +150,12 @@ synclet-release:
 	docker push $(SYNCLET_IMAGE):$(TAG)
 	sed -i 's/var SyncletTag = ".*"/var SyncletTag = "$(TAG)"/' internal/synclet/sidecar/sidecar.go
 
+custom-synclet-release:
+	$(eval TAG := $(if $(SYNCLET_TAG),$(SYNCLET_TAG),$(shell date +v%Y%m%d)))
+	docker build -t $(SYNCLET_IMAGE):$(TAG) -f synclet/Dockerfile .
+	docker push $(SYNCLET_IMAGE):$(TAG)
+	sed -i 's/var SyncletTag = ".*"/var SyncletTag = "$(TAG)"/' internal/synclet/sidecar/sidecar.go
```
nicks: hmmm...this `sed` rule doesn't make sense in your use-case, right? like you're not releasing your own tilt binaries with this custom image.

mostly trying to make sure i'm not totally misunderstanding your use-case

therc: I saw that, but if it's the same tag, which would normally be the case, the sed command is a no-op. I can take it out; I'm not sure what anyone else with a custom registry would want to do. :-)

therc: Removed

nicks:
lgtm!

thanks!

PR #2286, opened 2019-10-01
synclet: fix a global init bug introduced by https://github.com/windmilleng/tilt/pull/2277

nicks:
Hello @maiamcc, @jazzdan,

Please review the following commits I made in branch nicks/env2:

36f04931eed5d6549ec7c2876c51e813a7b43f3a (2019-10-01 14:35:12 -0400)
synclet: fix a global init bug introduced by https://github.com/windmilleng/tilt/pull/2277

### DIFF
--- a/internal/synclet/sidecar/sidecar.go
+++ b/internal/synclet/sidecar/sidecar.go
@@ -15,10 +15,12 @@ func syncletPrivileged() *bool {
 	return &val
 }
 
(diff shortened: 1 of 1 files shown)


nicks:
the bug here is that synclet_manager.go reads some of this global state directly to determine if the synclet is ok.

but really we just need to move this init into wire

Issue #2263, opened 2019-09-26
Old pod logs show up on startup


jazzdan:
## repro:
1. tilt up fortune
2. change the startup log message
3. kill tilt
4. tilt up fortune
(potentially repeat a few times)

## observed:
we get the fortune build output, followed by the pod output from multiple fortune pods:
```
fortune     ┊ STEP 3/3 — Deploying
fortune     ┊   │ Injecting images into Kubernetes YAML
fortune     ┊   │ Applying via kubectl:
fortune     ┊   │    matt-fortune:deployment
fortune     ┊ 
fortune     ┊   │ Step 1 - 0.738s
fortune     ┊   │ Step 2 - 0.000s
fortune     ┊   │ Step 3 - 0.227s
fortune     ┊   │ Done in: 0.965s 
fortune     ┊ 
fortune     ┊ 2019/09/23 16:21:19 Starting Fortune Service on :8082
fortune     ┊ 2019/09/23 16:22:55 Starting Fortune Service on :8082!!
fortune     ┊ 2019/09/23 16:24:23 Starting Fortune Service on :8082
```

expected:
we get the fortune build output followed by the pod output from the *current* fortune pod

It's possible it's fine to show pod output from the previous fortune pod, but:
1. it should be prior to the build log in the tilt ui, since it preceded it chronologically. its current position following the build is very confusing
2. I don't think there's an argument for showing pod logs for pods that never existed at the same time as Tilt (and I'm kind of surprised / unsure how Tilt's even managing to get them)

This is observable more dramatically if the service was in a crash loop (you'll get startup logs from every crash!) or if the service had logged a lot (@jazzdan  reported tilt was taking a lot of cpu dealing with old logs)

Originally written by @landism

landism:
> I don't think there's an argument for showing pod logs for pods that never existed at the same time as Tilt (and I'm kind of surprised / unsure how Tilt's even managing to get them)

Silly me.

For this particular repro, these logs are simply coming from the pod that is running when Tilt starts. It has multiple startup messages because there were multiple live updates to the same pod.

I'm not sure what a good solution to this is.

nicks:
fixed by #2287

PR #2287, opened 2019-10-01
engine: start streaming pod log at tthe tilt start time. fixes issue #2263

nicks:
Hello @jazzdan,

Please review the following commits I made in branch nicks/issue2263:

d2399277ea96dbf89abfc608abd168c36ee9235c (2019-10-01 14:47:01 -0400)
engine: start streaming pod log at tthe tilt start time. fixes issue #2263


nicks:
using this in practice, i'm not sure this is what we want. it's just kind of mysterious might need some way to log "attached to existing pod, here are the last N log lines"

Issue #2284, opened 2019-10-01
`helm()` should watch any additional values files


maiamcc:
the `helm()` builtin has a `values` arg that works like the [`-f` or `--values` flags in Helm](https://helm.sh/docs/chart_template_guide/#values-files) -- i.e. you can pass one or more values files that override the `values.yaml` in the chart.

Tilt should watch these values files for changes like we watch files in the chart (b/c if a values file changes, we should regen the yaml).

(See also: #2114 for more helm-related filewatching requests)

Requested by @MattHodge.

PR #2288, opened 2019-10-01
tiltfile: record helm.values args as config files (b/c they affect the helm chart, we should rerun whenever they change)

maiamcc:
addresses #2284

### DIFF
--- a/internal/tiltfile/tiltfile_test.go
+++ b/internal/tiltfile/tiltfile_test.go
@@ -4672,7 +4672,7 @@ func (f *fixture) setupExpand() {
 func (f *fixture) setupHelm() {
 	f.file("helm/Chart.yaml", chartYAML)
 	f.file("helm/values.yaml", valuesYAML)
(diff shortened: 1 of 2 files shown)


internal/tiltfile/tiltfile_test.go
```diff
@@ -4670,7 +4672,7 @@ func (f *fixture) setupExpand() {
 func (f *fixture) setupHelm() {
 	f.file("helm/Chart.yaml", chartYAML)
 	f.file("helm/values.yaml", valuesYAML)
-	f.file("helm/values-dev.yaml", valuesDevYAML)
+	f.file("dev/helm/values-dev.yaml", valuesDevYAML) // make sure we can pull in a values.yaml file from outisde chart dir
```
jazzdan:
```suggestion
	f.file("dev/helm/values-dev.yaml", valuesDevYAML) // make sure we can pull in a values.yaml file from outside chart dir
```

PR #2291, opened 2019-10-02
sidecar: integrate synclet image ref init with wire

nicks:
Hello @jazzdan, @maiamcc,

Please review the following commits I made in branch nicks/syncletinit:

acfa782b747c0a5bf431bad3e3fc179399ef47fa (2019-10-02 07:57:17 -0700)
sidecar: integrate synclet image ref init with wire


nicks:
this is mostly just making sure we do all this init in a deterministic and predictable order, so that we can log things at the right time.

PR #2293, opened 2019-10-02
feature: enable snapshots by default

jazzdan:


jazzdan:
Integration test failure looks unrelated.

PR #2303, opened 2019-10-04
assets: also be able to fetch assets by git SHA

jazzdan:
Now a version can be either `v1.2.3` or `abcdefedba`

### DIFF
--- a/pkg/assets/server.go
+++ b/pkg/assets/server.go
@@ -48,7 +48,8 @@ const errorDivStyle = `
   flex-direction: column;
 `
 
-var versionRe = regexp.MustCompile(`/(v\d+\.\d+\.\d+)/.*`) // matches `/vA.B.C/...`
+var versionRe = regexp.MustCompile(`/(v\d+\.\d+\.\d+)/.*`)       // matches `/vA.B.C/...`
+var shaRe = regexp.MustCompile(`(?m)\/(\b[0-9a-f]{5,40}\b)\/.*`) // matches /8bf2ea29eacff3a407272eb5631edbd1a14a0936/...
 
 type Server interface {
 	http.Handler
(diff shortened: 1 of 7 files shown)


pkg/assets/server.go
```diff
@@ -48,7 +48,8 @@ const errorDivStyle = `
   flex-direction: column;
 `
 
-var versionRe = regexp.MustCompile(`/(v\d+\.\d+\.\d+)/.*`) // matches `/vA.B.C/...`
+var versionRe = regexp.MustCompile(`/(v\d+\.\d+\.\d+)/.*`)       // matches `/vA.B.C/...`
+var shaRe = regexp.MustCompile(`(?m)\/(\b[0-9a-f]{5,40}\b)\/.*`) // matches /8bf2ea29eacff3a407272eb5631edbd1a14a0936/...
```
landism: 1. I don't think we want or need `?m` here?
2. I'm pretty sure the `\b`s are redundant given they're next to `\/`s
3. I suspect this the existing `versionRe` is buggy and we've copied the bug, in that we want a leading `^` to ensure we're only matching the first subpath. Or, at least: the presence of `.*` at the tail makes me suspect the author of `versionRe` thought this was a matching engine in which patterns implicitly require full matches (i.e., `foo` would not match `foobar`, but `foo.*` would), but golang does not have such an engine (i.e., `foo` does match `foobar`, and thus `foo` and `foo.*` are functionally equivalent), so maybe they incorrectly assumed this would only match the start of the path. Not that this is a huge problem, but it's a bit more risky with the new regex since it might match, like `teams/add/` if we ever add such a thing (ok, add doesn't have 5 chars...). Or maybe I'm wrong and we do want to match this anywhere in the string. I'm not certain of the types of URLs we're inputting here.

PR #2306, opened 2019-10-07
model: SHA is separate field

jazzdan:


internal/cli/build_test.go
```diff
@@ -10,7 +10,7 @@ import (
 )
 
 func TestProvideDevWebVersion(t *testing.T) {
-	assert.Equal(t, devVersion, string(provideWebVersion(provideTiltInfo())))
+	assert.Equal(t, fmt.Sprintf("v%s", devVersion), string(provideWebVersion(provideTiltInfo())))
```
nicks: i'm confused, why did this change? this seems like you broke something?

jazzdan: This is actually just [reverting it to the old assertion](https://github.com/windmilleng/tilt/commit/11b373407374c02866748f6049e07727c7f9d268#diff-ebee16632963e82fde104f24ad682044L11)

nicks: oh, ya, i see

PR #2307, opened 2019-10-07
build: label all docker images with builtby:tilt [ch3659]

maiamcc:
Hello @nicks, @jazzdan,

Please review the following commits I made in branch maiamcc/label-images:

877d89e565eb86180df12ce810f3e301e71c1ef6 (2019-10-07 17:26:27 -0400)
build: label all docker images with builtby:tilt [ch3659]

### DIFF
--- a/internal/build/test_utils.go
+++ b/internal/build/test_utils.go
@@ -15,6 +15,7 @@ import (
 	"github.com/docker/docker/api/types"
 	"github.com/docker/docker/api/types/container"
 	"github.com/docker/docker/client"
+	"github.com/stretchr/testify/assert"
 
 	wmcontainer "github.com/windmilleng/tilt/internal/container"
 	"github.com/windmilleng/tilt/internal/docker"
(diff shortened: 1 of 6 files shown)


internal/build/test_utils.go
```diff
@@ -174,6 +175,27 @@ func (f *dockerBuildFixture) assertImageNotExists(ref reference.NamedTagged) {
...
+	if err != nil {
+		f.t.Fatalf("error inspecting image %s: %v", ref.String(), err)
+	}
+
+	if inspect.Config == nil {
+		f.t.Fatalf("'inspect' result for image %s has nil config", ref.String())
+	}
+
+	actual := inspect.Config.Labels
+	for k, expectV := range expected {
```
jazzdan: Since we are only adding one label right now this could be rewritten more simply as (pseudocode):

```
if len(actual) == 1 && require.Contains(builtByTilt) {
```

maiamcc: You're totally right but since this is here i guess i might as well leave it for future? 🤷‍♀

PR #2310, opened 2019-10-07
pseudocode for docker pruner

maiamcc:
Hello @,

Please review the following commits I made in branch maiamcc/docker-pruner-pseudocode:

74d85df8c91fa7aa516e138dea3c3626a7d77349 (2019-10-07 18:42:36 -0400)
tiltfile to tweak settings

393abebb624bbe53583a6946106b3357e35b577b (2019-10-07 18:35:10 -0400)
pseudocode for docker pruner itself

### DIFF
--- a/internal/engine/docker_pruner.go
+++ b/internal/engine/docker_pruner.go
@@ -1,0 +1,76 @@
+package engine
+
+import (
+	"context"
+	"time"
+
+	"github.com/windmilleng/tilt/internal/docker"
+	"github.com/windmilleng/tilt/internal/store"
+)
+
+// How often to prune Docker images while Tilt is running
+// TODO: configurable from Tiltfile for special cases
+const dockerPruneInterval = time.Hour * 12
+
+type DockerPruner struct {
+	dCli docker.Client
+
+	started            bool
+	disabledForTesting bool
+}
+
+func NewDockerPruner(dCli docker.Client) *DockerPruner {
+	return &DockerPruner{dCli: dCli}
+}
+
+func (dp *DockerPruner) DisableForTesting() {
+	dp.disabledForTesting = true
+}
+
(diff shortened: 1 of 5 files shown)


maiamcc:
@nicks @jazzdan this isn't for landing but should be a useful bird's-eye view of what i'm planning to implement / lmk if you see any glaring problems right now

internal/engine/docker_pruner.go
```diff
@@ -0,0 +1,76 @@
+package engine
```
nicks: for future subscribers, can you put them in a subpackage of engine? been trying to reduce the circular dependencies in this code

internal/engine/docker_pruner.go
```diff
@@ -0,0 +1,76 @@
...
+	"context"
+	"time"
+
+	"github.com/windmilleng/tilt/internal/docker"
+	"github.com/windmilleng/tilt/internal/store"
+)
+
+// How often to prune Docker images while Tilt is running
+// TODO: configurable from Tiltfile for special cases
+const dockerPruneInterval = time.Hour * 12
```
nicks: do we really expect tilt to be up for 12 hours?

internal/engine/docker_pruner.go
```diff
@@ -0,0 +1,76 @@
...
+
+	if dp.started {
+		return
+	}
+
+	state := st.RLockState()
+	defer st.RUnlockState()
+
+	// wait until state has been kinda initialized
+	if !state.TiltStartTime.IsZero() && state.LastTiltfileError() == nil {
```
nicks: rather than doing this on a background reaper process, i wonder if we should be more systematic about it? e.g., doing in when there are no pending builds, after N builds complete, etc

The case i'm most worried about is cases like Ellie + Nic where they're editing a file that produces a big binary, so want to reap every N edits, rather than tied to a specific time interval

PR #2314, opened 2019-10-08
docker: set client scheme to http so we can ping server to negotiate version

maiamcc:
Hello @landism, @jazzdan,

Please review the following commits I made in branch maiamcc/set-scheme-docker-client:

f3c4d6fd7808ebbfae84a473f5261273869dd024 (2019-10-08 14:43:35 -0400)
docker: set client scheme to http so we can ping server to negotiate version

without this change, calls to `NegotiateVersion` (which happens if `$APIVersion` isn't set),
we will always return API v1.24:
- [pinging the server](https://github.com/moby/moby/blob/b3be2802d417d05813d6140b5bc177043288db24/client/client.go#L207) will always error
    b/c of unrecognized scheme (cuz nothing in our options has set the scheme yet); `NegotiateAPIVersion` ignores the error and passes along an empty ping to `negotiateAPIVersionPing`
- given an empty ping, [we default to
v1.24](https://github.com/moby/moby/blob/b3be2802d417d05813d6140b5bc177043288db24/client/client.go#L227)

If we can't accurately discern the user's docker version, we can't use
any features introduced in later versions (e.g. I'm fixing this so that
I can use the `filters` param for `ImagesPrune`).

### DIFF
--- a/internal/docker/client.go
+++ b/internal/docker/client.go
@@ -229,6 +229,10 @@ func SupportsBuildkit(v types.Version, env Env) bool {
 func CreateClientOpts(ctx context.Context, env Env) ([]func(client *client.Client) error, error) {
 	result := make([]func(client *client.Client) error, 0)
 
+	// Set scheme = HTTP so we can ping the server (otherwise we can't
+	// accurately negotiate versions with the server below).
+	result = append(result, client.WithScheme("http"))
+
 	if env.CertPath != "" {
 		options := tlsconfig.Options{
 			CAFile:             filepath.Join(env.CertPath, "ca.pem"),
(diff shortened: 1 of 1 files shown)


landism:
Can we do this for just the version check? It seems bad to give up on https across the board.

maiamcc:
I took a swing at doing it just for the version check, but I can't find a way to READ the scheme of the docker client (private field) so i can't get the current value to reset it to that after i change it to http... like what i would want to do is:

```
if client.scheme == "":
    client.WithSchema("http")
    NegotiateAPIVersion()
    client.WithSchema("")
```
I guess I could just set it back to empty after?

landism:
or maybe I'm misunderstanding what's going on?

You say the request fails because we haven't set the scheme yet - do we set the scheme later somehow and then requests start working, but for some reason we're doing the ping earlier than that?

nicks:
ya, this doesn't smell right. 

have you looked at how the docker cli does the version negotiation?

does it work if you switch over to the negotiation option that ships with the cli?
https://godoc.org/github.com/docker/docker/client#WithAPIVersionNegotiation

maiamcc:
This is much cleaner with #2323 + #2325 -- closing this ticket.

PR #2315, opened 2019-10-08
engine: docker pruner (disabled by default) runs on startup and every hour, prunes old objects [ch3633]

maiamcc:
Still needs tests but wanted to push something up for review before I left!


internal/docker/client.go
```diff
@@ -26,12 +27,24 @@ import (
...
 
 	"github.com/windmilleng/tilt/internal/container"
 	"github.com/windmilleng/tilt/pkg/logger"
 	"github.com/windmilleng/tilt/pkg/model"
 )
 
+// Label that we attach to all of the images we build.
+const (
+	BuiltBy = "builtby"
+	Tilt    = "tilt"
```
jazzdan: I think `client.Tilt` is a weird name for a constant that is just "tilt". Is this referenced anywhere else? Does it need to be a constant?

internal/docker/client.go
```diff
@@ -504,3 +520,85 @@ func (c *Cli) ExecInContainer(ctx context.Context, cID container.ID, cmd model.C
 		return nil
 	}
 }
+
+func (c *Cli) Prune(ctx context.Context, age time.Duration) error {
+	l := logger.Get(ctx)
+	if err := c.Client.NewVersionError("1.30", "image | container prune with filter: label"); err != nil {
+		l.Debugf("[Docker Prune] skipping Docker prune, Docker API version too low:\t%v", err)
+		fmt.Printf("[Docker Prune] skipping Docker prune, Docker API version too low:\t%v", err)
```
jazzdan: Both a debug and a printf?

internal/sliceutils/sliceutils.go
```diff
@@ -29,6 +29,18 @@ func QuotedStringList(list []string) string {
 	return strings.Join(result, ", ")
 }
 
+func BulletedIndentedStringList(list []string) string {
```
jazzdan: Is this going to be used outside of the docker package?

internal/docker/client.go
maiamcc: whooops!

internal/cli/docker_prune.go
```diff
@@ -0,0 +1,101 @@
...
+		Short: "run docker prune as tilt does",
+	}
+
+	cmd.Flags().StringVar(&c.untilStr, "until", "12h", "max age of image to keep (as go duration string, e.g. 1h30m, 12h")
+
+	return cmd
+}
+
+func (c *dockerPruneCmd) run(ctx context.Context, args []string) error {
+	// a := analytics.Get(ctx)
```
nicks: ?

internal/cli/docker_prune.go
```diff
@@ -0,0 +1,101 @@
...
+	}
+
+	err = dCli.Prune(ctx, until)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func provideDockerClient(ctx context.Context) (docker.Client, error) {
```
nicks: hmmm...did you write this by hand? can you use the wireDockerClusterClient that `tilt doctor` uses?

internal/cli/docker_prune.go
```diff
@@ -0,0 +1,101 @@
...
+	// tags := tracer.TagStrToMap(c.traceTags)
+
+	// for k, v := range tags {
+	// 	span.SetTag(k, v)
+	// }
+
+	deferred := logger.NewDeferredLogger(ctx)
+	ctx = redirectLogs(ctx, deferred)
+
+	dCli, err := provideDockerClient(ctx)
```
nicks: i wonder if we should print out which docker instance this is talking to, since without executing the Tiltfile, you can't be 100% certain that you're talking to the right one (though it will mostly be ok, `tilt doctor` has a similar problem, which is why it prints out the command)

internal/engine/docker_pruner.go
```diff
@@ -0,0 +1,81 @@
...
+)
+
+// How often to prune Docker images while Tilt is running
+const dockerPruneInterval = time.Hour * 12
+
+// Prune Docker objects older than this
+// TODO: configurable from Tiltfile for special cases
+const dockerUntilInterval = time.Hour * 6
+
+type DockerPruner struct {
```
nicks: let's move new subscribers into sub-packages of engine (see other examples in this directory). Trying to reduce circular deps

internal/engine/docker_pruner.go
```diff
@@ -0,0 +1,81 @@
...
+	"context"
+	"time"
+
+	"github.com/windmilleng/tilt/internal/docker"
+	"github.com/windmilleng/tilt/internal/store"
+	"github.com/windmilleng/tilt/pkg/logger"
+)
+
+// How often to prune Docker images while Tilt is running
+const dockerPruneInterval = time.Hour * 12
```
nicks: i wonder if we should do this once on startup and then after every N builds. it seems unlikely that anyone will be running tilt for 12 hours, given an 8 hour workday ever since the labor rights movement in the 30s

i still think it's likely that some people will want this every N builds, or when their disk space is low (see my comment on the other pr)

internal/cli/docker_prune.go
maiamcc: oh weird idk. good catch.

internal/cli/docker_prune.go
maiamcc: oh yea, I just hacked this file together to test with. will clean this up!

internal/engine/docker_pruner.go
maiamcc: that makes a lot of sense. I'm going to commit this as is (with the pruner disabled) and add that functionality when I make it configurable.

maiamcc:
alright @nicks @jazzdan could you take another pass at this?

still to come: 
- configuration via tiltfile
- ability to prune based on # of completed builds instead of by time

internal/engine/dockerprune/docker_pruner.go
```diff
@@ -0,0 +1,176 @@
...
+	}
+
+	f := filters.NewArgs(
+		filters.Arg("label", docker.BuiltByTiltLabelStr),
+		filters.Arg("until", maxAge.String()),
+	)
+
+	opts := types.BuildCachePruneOptions{
+		All:         true,
+		KeepStorage: 524288000, // 500MB -- TODO: make configurable
```
landism: 1. Is this required? Or can we leave it unset so that we don't muck with it? It seems like setting this is getting out of scope of pruning and maybe breaking something the user has configured elsewhere. Though, I can't find documentation describing what this does, so....I don't know if this is, like, saying "it's ok if you delete all but 500mb".
2. It's generally clearer / less bug-prone to write numbers like this as 500 * 1024 * 1024

internal/engine/dockerprune/docker_pruner.go
```diff
@@ -0,0 +1,176 @@
...
+	"github.com/windmilleng/tilt/internal/store"
+	"github.com/windmilleng/tilt/pkg/logger"
+)
+
+// How often to prune Docker images while Tilt is running
+const dockerPruneInterval = time.Hour
+
+// Prune Docker objects older than this
+// TODO: configurable from Tiltfile for special cases
+const defaultMaxAge = time.Hour * 6
```
landism: Not blocking, but as discussed offline I worry we'll end up in cases where we're deleting the newest build of a resource when the user is using a remote k8s cluster.

internal/engine/dockerprune/docker_pruner.go
```diff
@@ -0,0 +1,176 @@
...
+	return nil
+}
+
+func (dp *DockerPruner) sufficientVersionError() error {
+	return dp.dCli.NewVersionError("1.30", "image | container prune with filter: label")
+}
+
+func prettyPrintCachePruneReport(report *types.BuildCachePruneReport, l logger.Logger) {
+	// TODO: human-readable space reclaimed
+	l.Infof("[Docker Prune] removed %d caches, reclaimed %d bytes", len(report.CachesDeleted), report.SpaceReclaimed)
```
landism: I wonder if these will be useful or distracting in practice, and if we should skip logging if SpaceReclaimed is 0. I guess we'll see!

internal/engine/dockerprune/docker_pruner.go
```diff
@@ -0,0 +1,176 @@
...
+		return nil
+	}
+
+	f := filters.NewArgs(
+		filters.Arg("label", docker.BuiltByTiltLabelStr),
+		filters.Arg("until", maxAge.String()),
+	)
+
+	opts := types.BuildCachePruneOptions{
+		All:         true,
```
nicks: this doesn't seem right. shouldn't we prune the images first, then prune the dangling build cache?

nicks: i.e., `All` should be false by default, as it is in Docker

internal/engine/dockerprune/docker_pruner.go
```diff
@@ -0,0 +1,176 @@
...
+func (dp *DockerPruner) prune(ctx context.Context, maxAge time.Duration) error {
+	l := logger.Get(ctx)
+	if err := dp.sufficientVersionError(); err != nil {
+		l.Debugf("[Docker Prune] skipping Docker prune, Docker API version too low:\t%v", err)
+		return nil
+	}
+
+	f := filters.NewArgs(
+		filters.Arg("label", docker.BuiltByTiltLabelStr),
+		filters.Arg("until", maxAge.String()),
```
nicks: for my education: if you rebuild an image, and all the layers are cached, does it update the timestamp that `until` uses?

maiamcc: =( =( =( no it doesn't

internal/engine/dockerprune/docker_pruner.go
maiamcc: hopefully that'll be addressed by the "prune every X builds" setting coming in my next commit

internal/engine/dockerprune/docker_pruner.go
maiamcc: fair nough

internal/engine/dockerprune/docker_pruner.go
maiamcc: Good calling, will have them not output in that case.

internal/engine/dockerprune/docker_pruner.go
maiamcc: Hmm, that seems right. will change.

PR #2318, opened 2019-10-09
web: fix typescript eslint warnings

jazzdan:


landism:
it looks like if we want to enforce this, we need to:
1. set `CI=true` when we `yarn run` in dev
2. do a `CI=true yarn build` in CI

PR #2323, opened 2019-10-10
vendor: update docker/docker and its deps

maiamcc:
So that I can use [`WithAPIVersionNegotiation`](https://godoc.org/github.com/docker/docker/client#WithAPIVersionNegotiation) in #2314


internal/docker/client.go
```diff
@@ -294,14 +295,18 @@ func (c *Cli) initCreds(ctx context.Context) dockerCreds {
...
 					_ = session.Close()
 				}()
 
 				// Start the server
-				_ = session.Run(ctx, c.Client.DialSession)
+				dialSession := func(ctx context.Context, proto string, meta map[string][]string) (net.Conn, error) {
+					return c.Client.DialHijack(ctx, "/session", proto, meta)
+				}
+				_ = session.Run(ctx, dialSession)
+				_ = session.Run(ctx, dialSession)
```
nicks: did you intend to repeat this line?

maiamcc: noooope i did not!

maiamcc:
> the windows compile error is interesting?

yeahhhh. Looks like [golang broke some stuff](https://github.com/golang/go/issues/34610) -- I added a commit where we get an older version of `x/sys` to work around that.

PR #2331, opened 2019-10-11
engine: put the log managers into their own subpackage

nicks:
Hello @maiamcc,

Please review the following commits I made in branch nicks/runtimelog:

b75d24ab1e41342646b2dcc91259fbc39a41c5d2 (2019-10-11 16:05:08 -0400)
engine: put the log managers into their own subpackage


nicks:
not totally convinced this is the right name, but moving these into a subpackage will help with the complexity

PR #2334, opened 2019-10-11
tiltfile/dockerpruner: configure docker pruner from tiltfile, more options for when we prune [ch3661]

maiamcc:
Hello @jazzdan, @nicks,

Please review the following commits I made in branch maiamcc/tiltfile-docker-prune:

3b8ef4389c13c4b58289ce0bfa89b913961d6e18 (2019-10-11 17:17:45 -0400)
tests

d13689981265d54f895b1e696d422dd889048dc5 (2019-10-11 17:17:42 -0400)
be smarter about when we do/don't prune

6a375340c0e32f3edc5022cbde4b00df80e68b3f (2019-10-11 17:17:38 -0400)
tiltfile tweaks

dd885d0515e2a034026146425790f02eac640034 (2019-10-11 17:17:35 -0400)
apply the settings

94e8746c7b823579f7a8972de6169a5da231db76 (2019-10-11 17:17:30 -0400)
control docker prune settings from tiltfile


internal/tiltfile/tiltfile_state.go
```diff
@@ -1226,6 +1235,34 @@ func (s *tiltfileState) setTeam(thread *starlark.Thread, fn *starlark.Builtin, a
 	return starlark.None, nil
 }
 
+func (s *tiltfileState) dockerPruneSettings(thread *starlark.Thread, fn *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
```
nicks: all new functions should be implemented as extensions, see this pr for an example:
https://github.com/windmilleng/tilt/pull/2330

it should be pretty straightfoward

maiamcc:
oh whoops i just realized i left the state locked for all of pruning, i was gonna fix that.

internal/engine/dockerprune/docker_pruner.go
```diff
@@ -35,57 +36,77 @@ func NewDockerPruner(dCli docker.Client) *DockerPruner {
...
-			}
-		}()
+	// If user doesn't have at least one Docker build, they probably don't care about pruning
+	if !state.HasDockerBuild() {
+		return
+	}
+
+	// Don't prune WHILE we're building something, in case of weird side-effects.
+	// TODO: more robust check here
+	if state.CurrentlyBuilding != "" {
```
nicks: i'm pretty sure this won't behave as you expect. this will just fire as soon as the first build completes. i would probably use BuildController's needsBuild function

internal/cli/down.go
```diff
@@ -48,6 +48,8 @@ func (c *downCmd) down(ctx context.Context, downDeps DownDeps) error {
 		return err
 	}
 
+	tlr.DockerPruneSettings.Disabled = true // don't docker prune on tilt down
```
nicks: i wonder if we should do the opposite: DockerPruneSettings.Enabled, which default to false unless the Tiltfile has run and this is a 'tilt up'

nicks:
high-level question: my understanding is that this still has a major bug where if you haven't run Tilt for a few days, and you run 'tilt up', it will delete all your cached images. is that still an issue?

i'm surprised we're not prioritizing that over making it configurable? That bug seems like a launch-blocker (since it makes Tilt unusable), whereas the config issues do not

internal/cli/down.go
maiamcc: are you saying that the default for the USER should be that DockerPrune is disabled unless they explicitly enable it, or just that in the CODE we should default to `disabled` unless it's a `tilt up`?

nicks: this is just a comment about how to organize the code

internal/tiltfile/tiltfile_state.go
maiamcc: will do in a separate PR to reduce size of this one

internal/cli/down.go
maiamcc: k, done!

maiamcc:
@nicks think i addressed all your comments except:
- this won't work for cluster users -- will address in a separate PR, and be sure to merge the two at the same time
- implement the tiltfile func as an extension -- will address in a separate PR.

Would love a re-review!

maiamcc:
👋 bump!

internal/engine/dockerprune/docker_pruner.go
```diff
@@ -35,57 +38,80 @@ func NewDockerPruner(dCli docker.Client) *DockerPruner {
...
+	dp.disabledForTesting = disabled
 }
 
 func (dp *DockerPruner) OnChange(ctx context.Context, st store.RStore) {
 	if dp.disabledForTesting {
 		return
 	}
 
-	if dp.started {
+	if err := dp.sufficientVersionError(); err != nil {
```
jazzdan: Should we log something in verbose in this case?

nicks:
ah, sorry, this PR has gotten way too big to review, i thought the plan was to break it up?

internal/store/engine_state.go
```diff
@@ -85,15 +85,7 @@ type EngineState struct {
...
-	DockerPruneSettings DockerPruneSettings
-}
-
-type DockerPruneSettings struct {
-	Disabled bool
-}
-
-func defaultDockerPruneSettings() DockerPruneSettings {
-	return DockerPruneSettings{Disabled: true} // for now, always disabled until we can configure it
+	DockerPruneSettings model.DockerPruneSettings
```
jazzdan: This looks weird to me. What's happening here? Is this returning the empty value of `model.DockerPruneSettings`?

internal/tiltfile/tiltfile_state.go
```diff
@@ -1372,6 +1384,34 @@ func (s *tiltfileState) setTeam(thread *starlark.Thread, fn *starlark.Builtin, a
...
+	if err := s.unpackArgs(fn.Name(), args, kwargs,
+		"disable?", &disable,
+		"max_age_mins?", &maxAgeMins,
+		"num_builds?", &numBuilds,
+		"interval_hrs?", &intervalHrs); err != nil {
+		return nil, err
+	}
+
+	if disable && (intervalHrs != 0 || numBuilds != 0 || maxAgeMins != 0) {
+		return nil, fmt.Errorf("can't disable Docker Prune (`disabled=True`) and pass additional settings")
```
jazzdan: Why not?

maiamcc: > ah, sorry, this PR has gotten way too big to review, i thought the plan was to break it up?

idr us discussing that but sure, can do

maiamcc:
> ah, sorry, this PR has gotten way too big to review, i thought the plan was to break it up?

idr us discussing that but sure, can do

internal/store/engine_state.go
maiamcc: Yep: if we don't bubble the TiltfileLoadResults all the way through the engine state, then we get the empty value of `DockerPruneSettings` (where `Enabled = false` -- so in case like `tilt down`, we're not running the pruner at all).

internal/tiltfile/tiltfile_state.go
maiamcc: cuz it doesn't make sense? if we're not ever running docker prune, why would we allow you to configure settings for it? (i don't feel strongly about this tho.)

internal/engine/dockerprune/docker_pruner.go
maiamcc: will do

maiamcc:
closing this PR in favor of three smaller ones:
- #2340 (move relevant `buildcontrol` funcs to their own package)
- #2344 (use `state.DockerPruneSettings` to determine when/how we prune)
- #2345 (configure `DockerPruneSettings` via the tiltfile)

internal/store/engine_state.go
jazzdan: Can we be explicit about what the default values are rather than returning empty values for everything?

internal/tiltfile/tiltfile_state.go
jazzdan: I could imagine it being annoying if you wanted to do something like this:

```python
pruner_disabled = read_config()
docker_prune_settings(disable=pruner_disabled, interval_hrs=2)
```

With how the API is now you'd have to instead rwrite
```python
pruner_disabled = read_config()
if pruner_disabled:
  docker_prune_settings(disable=pruner_disabled, interval_hrs=2)
```

Not a big deal, but that's my logic.

maiamcc: mmm fair.

PR #2335, opened 2019-10-11
web: ability to highlight log lines and store them in state

jazzdan:
Video demo: https://drive.google.com/file/d/1RkYgeQuHsaSpF-fGV0iS4k7_U0YL9Hh7/view?usp=sharing


nicks:
screenshot?

jazzdan:
Screen Shot 2019-10-11 at 5 57 08 PM

web/src/LogPane.tsx
```diff
@@ -119,11 +183,33 @@ class LogPane extends Component<LogPaneProps, LogPaneState> {
...
-      (line: string, i: number): React.ReactElement => {
-        return <AnsiLine key={"logLine" + i} line={line} />
+
+    let sawBeginning = false
+    let sawEnd = false
+    let highlight = this.props.highlight
+    for (let i = 0; i < lines.length; i++) {
+      const l = lines[i]
+      const key = "logLine" + i
+      let classes = [key]
```
nicks: it seems deeply weird for each log line to have a custom css class

can you create a single css class ("logLine") and use a data attribute for the key?

web/src/LogPane.tsx
```diff
@@ -38,6 +46,62 @@ class LogPane extends Component<LogPaneProps, LogPaneState> {
...
+            endingLogID: endingLogLine,
+            highlightedText: selection.toString(),
+          })
+        }
+      }
+    }
+  }
+
+  findLogLineID(el: HTMLElement | null): string | null {
+    if (el === null) {
```
nicks: strict equality is not what you want here. this whole tree would be simpler if you used `if (!el)`

web/src/LogPane.tsx
```diff
@@ -38,6 +46,62 @@ class LogPane extends Component<LogPaneProps, LogPaneState> {
...
+      return null
+    } else if (el && el.className.startsWith("logLine")) {
```
nicks: generally more robust to use the classList api
https://developer.mozilla.org/en-US/docs/Web/API/Element/classList

tho in this case it'd be even better to use a data attribute

web/src/HUD.tsx
```diff
@@ -188,6 +194,18 @@ class HUD extends Component<HudProps, HudState> {
       .catch(err => console.error(err))
   }
 
+  handleSetHighlight(highlight: SnapshotHiglight) {
+    this.setState({
```
nicks: does this get properly cleared when you switch views? seems like you could have a highlight associated with the wrong resource

web/src/types.ts
```diff
@@ -103,6 +103,13 @@ export type Snapshot = {
   SnapshotLink: string
   showSnapshotModal: boolean
   path?: string
+  snapshotHighlight?: SnapshotHiglight | null
+}
+
+export type SnapshotHiglight = {
+  beginningLogID: string
```
nicks: we should probably handle offset as well? traditionally selection state is represented like
startLine
startOffset
endLine
endOffset

jazzdan:
@nicks we're going to hold off on offsets until a separate PR.

web/src/HUD.tsx
jazzdan: We need to look in to this

jazzdan:
This also has some problems crashing because the node isn't mounted when the listener executes when switching views. Investigating.

web/src/HUD.tsx
jazzdan: It does get cleared! When you switch logs the document selection is "collapsed", at which point we clear the highlight from the state.

web/src/HUD.tsx
```diff
@@ -180,6 +188,35 @@ class HUD extends Component<HudProps, HudState> {
...
+  highlightsEnabled(): boolean {
+    return (
+      !this.pathBuilder.isSnapshot() &&
+      this.getFeatures().isEnabled("snapshot_highlights")
+    )
+  }
+
+  private getFeatures(): Features {
+    if (this.state.View) {
+      this.features = new Features(this.state.View.FeatureFlags)
```
nicks: i don't understand why we're storing this as a property of the class at all?

currently, we write the field every time we read the field

jazzdan: Whoops, I meant to memoize this but messed up. Thanks

jazzdan: In the end I decided not to bother memoizing it.

hyu:
@nicks — Yeah, those snapshot tests are usually a hassle. I think the only signal they give is "were DOM changes expected?"

One time I do vaguely recall it helped catch the fact that I hadn't expected HTML changes for something I was doing. But that was a rare instance of it being somewhat useful.

PR #2339, opened 2019-10-14
dockerprune: instead of docker.ImagesPrune, only remove tilt-built images with lastTag younger than MaxAge [ch3707]

maiamcc:
If someone is using Tilt against a cloud cluster, it's not enough to run `docker image prune` -- Image X may be running a container in the cloud, but the local docker daemon doesn't know about it and thinks it's prune-able, but we DON'T want that, as the user is probably actively developing on it.

alternative: list all tilt-built images, and check "last tagged" metadata for a more accurate picture of when the image was last updated; only delete those images where it was last tagged sufficiently long ago.

### DIFF
--- a/internal/engine/dockerprune/docker_pruner.go
+++ b/internal/engine/dockerprune/docker_pruner.go
@@ -128,14 +128,21 @@ func (dp *DockerPruner) prune(ctx context.Context, maxAge time.Duration) error {
 		filters.Arg("until", maxAge.String()),
 	)
 
-	fWithDangling := f.Clone()
-	fWithDangling.Add("dangling", "0") // prune all images, not just dangling ones
-	imgReport, err := dp.dCli.ImagesPrune(ctx, fWithDangling)
+	// PRUNE CONTAINERS
+	containerReport, err := dp.dCli.ContainersPrune(ctx, f)
 	if err != nil {
 		return err
 	}
-	prettyPrintImagesPruneReport(imgReport, l)
+	prettyPrintContainersPruneReport(containerReport, l)
 
+	// PRUNE IMAGES
+	imageReport, err := dp.deleteOldImages(ctx, maxAge)
+	if err != nil {
+		return err
+	}
+	prettyPrintImagesPruneReport(imageReport, l)
+
+	// PRUNE BUILD CACHE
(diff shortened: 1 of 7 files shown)


internal/engine/dockerprune/docker_pruner.go
```diff
@@ -149,13 +156,52 @@ func (dp *DockerPruner) prune(ctx context.Context, maxAge time.Duration) error {
...
+	}
+
+	rmOpts := types.ImageRemoveOptions{PruneChildren: true}
+	var responseItems []types.ImageDeleteResponseItem
+	var reclaimedBytes uint64
+
+	for imgID, bytes := range toDelete {
+		items, err := dp.dCli.ImageRemove(ctx, imgID, rmOpts)
+		// No good way to detect in-use images from `inspect` output, so just silently ignore
+		if err != nil && !strings.Contains(err.Error(), "image is being used by running container") {
```
nicks: if the image is being used by a running container, that means we shouldn't report it as deleted or reclaimed, right?

internal/engine/dockerprune/docker_pruner.go
```diff
@@ -149,13 +156,52 @@ func (dp *DockerPruner) prune(ctx context.Context, maxAge time.Duration) error {
...
+	}
+	imgs, err := dp.dCli.ImageList(ctx, opts)
 	if err != nil {
-		return err
+		return types.ImagesPruneReport{}, err
 	}
-	prettyPrintContainersPruneReport(containerReport, l)
 
-	return nil
+	toDelete := make(map[string]uint64) // map imageID to size in bytes
```
nicks: i'm surprised this isn't just a filtered list; were you seeing duplicates in the list that you were trying to avoid?

internal/engine/dockerprune/docker_pruner.go
maiamcc: whoops, right!

internal/engine/dockerprune/docker_pruner.go
maiamcc: you mean why is it a map instead of a slice? because I needed to keep the image size handy for the report (unless i'm missing an easy way to do that)

PR #2342, opened 2019-10-14
integration: fix the analytics tests to use busybox instead of the golang docker image

nicks:
Hello @nicks,

Please review the following commits I made in branch nicks/busybox:

e97e2eed54d5d880bb96922d9dcf7304b66a43e7 (2019-10-14 19:34:55 -0400)
integration: fix the analytics tests to use busybox instead of the golang docker image


nicks:
I think the real bug here is that the analytics test and the oneup test were stepping on each other's resources, but this also makes things a lot faster in general

PR #2344, opened 2019-10-15
dockerprune: use state.dockerPruneSettings to dictate when/how we prune

maiamcc:
Hello @nicks, @jazzdan,

Please review the following commits I made in branch maiamcc/dp-use-settings:

7daad87da01bfbc7bf5ad4eb8e5c80266ae76391 (2019-10-14 20:50:18 -0400)
log something if version insufficient to prune

8f1cf91a58da6c1223cd8c4e524dd3f3c488b770 (2019-10-14 20:12:11 -0400)
dockerprune: use state.dockerPruneSettings to dictate when/how we prune

### DIFF
--- a/internal/engine/dockerprune/docker_pruner.go
+++ b/internal/engine/dockerprune/docker_pruner.go
@@ -26,7 +26,8 @@ const defaultMaxAge = time.Hour * 6
 type DockerPruner struct {
 	dCli docker.Client
 
-	disabledForTesting bool
+	disabledForTesting           bool
+	insufficientVersionErrLogged bool
 
 	lastPruneBuildCount int
 	lastPruneTime       time.Time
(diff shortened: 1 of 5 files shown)


internal/engine/dockerprune/docker_pruner.go
```diff
@@ -35,57 +39,86 @@ func NewDockerPruner(dCli docker.Client) *DockerPruner {
...
+	if settings.NumBuilds != 0 {
+		buildsSince := curBuildCount - dp.lastPruneBuildCount
+		if buildsSince >= settings.NumBuilds {
+			dp.PruneAndRecordState(ctx, settings.MaxAge, curBuildCount)
+		}
+		return
+	}
+
+	interval := settings.Interval
+	if interval == 0 {
```
nicks: i feel like the defaulting approach here is a big ad-hoc and likely to accidentally miss a place...maybe these should be methods of the settings object? Or there should be a NewDefaultSettings() constructor?

maiamcc: that makes sense. most of the defaulting work will happen in the tiltfile settings i think, so i'll address this in that PR.

PR #2345, opened 2019-10-15
tiltfile: configure docker prune settings

maiamcc:
the tiltfile changes accompanying #2344

### DIFF
--- a/internal/tiltfile/dockerprune/docker_prune.go
+++ b/internal/tiltfile/dockerprune/docker_prune.go
@@ -11,8 +11,7 @@ import (
 	"github.com/windmilleng/tilt/internal/tiltfile/starkit"
 )
 
-// Implements functions for dealing with the Kubernetes context.
-// Exposes an API for other plugins to get and validate the allowed k8s context.
+// Implements functions for dealing with Docker Prune settings.
 type Extension struct {
 	settings model.DockerPruneSettings
 }
--- a/internal/tiltfile/dockerprune/docker_prune.go
+++ b/internal/tiltfile/dockerprune/docker_prune.go
@@ -20,7 +20,9 @@ type Extension struct {
 func NewExtension() *Extension {
 	return &Extension{
 		settings: model.DockerPruneSettings{
-			Enabled: true,
+			Enabled:  true,
+			MaxAge:   model.DockerPruneDefaultMaxAge,
+			Interval: model.DockerPruneDefaultInterval,
 		},
 	}
 }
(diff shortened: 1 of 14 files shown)


maiamcc:
Oh shoot I forgot I was going to make this an extension. Hold off reviewing this for now.

maiamcc:
okay, ready for review!

internal/tiltfile/dockerprune/docker_prune.go
```diff
@@ -0,0 +1,62 @@
...
+
+	"go.starlark.net/starlark"
+
+	"github.com/windmilleng/tilt/pkg/model"
+
+	"github.com/windmilleng/tilt/internal/tiltfile/starkit"
+)
+
+// Implements functions for dealing with the Kubernetes context.
+// Exposes an API for other plugins to get and validate the allowed k8s context.
```
jazzdan: 👀

internal/tiltfile/dockerprune/docker_prune.go
```diff
@@ -0,0 +1,62 @@
...
+	if err := starkit.UnpackArgs(thread, fn.Name(), args, kwargs,
+		"disable?", &disable,
+		"max_age_mins?", &maxAgeMins,
+		"num_builds?", &numBuilds,
+		"interval_hrs?", &intervalHrs); err != nil {
+		return nil, err
+	}
+
+	if disable && (intervalHrs != 0 || numBuilds != 0 || maxAgeMins != 0) {
+		return nil, fmt.Errorf("can't disable Docker Prune (`disabled=True`) and pass additional settings")
```
jazzdan: See my comment on the earlier PR

internal/tiltfile/dockerprune/docker_prune.go
```diff
@@ -0,0 +1,62 @@
...
+	}
+
+	if numBuilds != 0 && intervalHrs != 0 {
```
jazzdan: We could encode this in the type:

```python
docker_prune_settings(frequency: Union[NumBuilds, Interval])
```

might be too much work, but I think it would be a more clean API.

internal/tiltfile/wire.go
```diff
@@ -3,10 +3,13 @@ package tiltfile
...
 
+	"github.com/windmilleng/tilt/internal/tiltfile/dockerprune"
+
 	"github.com/windmilleng/tilt/internal/tiltfile/k8scontext"
 )
 
 var WireSet = wire.NewSet(
 	ProvideTiltfileLoader,
 	k8scontext.NewExtension,
+	dockerprune.NewExtension,
```
nicks: i'm surprised we're injecting the extension instead of just creating it inline? it's not like it has any special state we need to share or inject, so i'm not sure what you're getting from the additional complexity

maiamcc: ...nah, I was just following the k8sContextExt template w/o thinking. will fix.

internal/tiltfile/dockerprune/docker_prune.go
maiamcc: 😅

PR #2347, opened 2019-10-15
cloud: convert BuildRecords into update proto objects

nicks:
Hello @landism,

Please review the following commits I made in branch nicks/update:

350c98c1b831bf8c8b6f2c1f92f73c17e9c6c021 (2019-10-14 21:27:20 -0400)
cloud: convert BuildRecords into update proto objects

### DIFF
--- a/internal/cloud/update_uploader.go
+++ b/internal/cloud/update_uploader.go
@@ -1,0 +1,134 @@
+package cloud
+
+import (
+	"context"
+	"time"
+
+	"github.com/golang/protobuf/ptypes"
+	"github.com/golang/protobuf/ptypes/duration"
+	"github.com/golang/protobuf/ptypes/timestamp"
+
+	"github.com/windmilleng/tilt/internal/feature"
+	"github.com/windmilleng/tilt/internal/store"
+)
+
+type UpdateUploader struct {
+	client HttpClient
+	addr   Address
+
+	lastCompletedBuildCount int
+	lastFinishTime          time.Time
+}
+
+func NewUpdateUploader(client HttpClient, addr Address) *UpdateUploader {
+	return &UpdateUploader{
+		client: client,
+		addr:   addr,
+	}
+}
+
+// TODO(nick): Generate these with protobufs.
+type updateServiceSpec struct {
+	Name string `json:"name"`
+}
+
+type update struct {
+	Service      updateServiceSpec    `json:"service"`
+	StartTime    *timestamp.Timestamp `json:"start_time"`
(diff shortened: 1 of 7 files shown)


internal/cloud/update_uploader.go
```diff
@@ -0,0 +1,128 @@
...
+	}
+
+	// Do a quick check to see if any builds have completed since we last checked.
+	if state.CompletedBuildCount == 0 || state.CompletedBuildCount <= u.lastCompletedBuildCount {
+		return nil
+	}
+
+	// OK, we know we have work to do!
+
+	highWaterMark := u.lastFinishTime
```
landism: is there anything protecting this function from getting invoked in parallel by multiple calls to `OnChange` and processing the same set of builds?

similarly, do we need synchronization on accessing `u`'s fields?

internal/cloud/update_uploader.go
```diff
@@ -0,0 +1,128 @@
...
+// Check the engine state to see if we have any updates.
+func (u *UpdateUploader) makeUpdates(st store.RStore) []update {
+	state := st.RLockState()
+	defer st.RUnlockState()
+
+	// Check if this feature is disabled
+	if !state.Features[feature.UpdateHistory] {
+		return nil
+	}
+
```
landism: let's also return `if state.TeamName == ""`

internal/feature/flags.go
```diff
@@ -21,6 +21,7 @@ const (
 const MultipleContainersPerPod = "multiple_containers_per_pod"
 const Events = "events"
 const Snapshots = "snapshots"
+const UpdateHistory = "update-history"
```
landism: nit: our other feature names are `_` rather than `-`

internal/cloud/update_uploader.go
nicks: from:
https://github.com/windmilleng/tilt/blob/master/internal/store/subscriber.go#L11

```
// Subscribers do not need to be thread-safe. The Store will only
// call OnChange for a given subscriber when the last call completes.
```

internal/cloud/update_uploader.go
nicks: oh, we're not supporting update-history for non-teams?

internal/cloud/update_uploader.go
landism: Whoops, apparently my approach to subscriber concurrency has been off. I've been spinning blocking work off to a goroutine and then dealing with locking, unaware the whole thing is already in a thread-safe goroutine.

PR #2349, opened 2019-10-15
tiltfile: extensions should reset state in OnStart

nicks:
Hello @jazzdan, @maiamcc,

Please review the following commits I made in branch nicks/k8scontext:

17b2e34f0c0a894d7f86b8404cefd689322a3083 (2019-10-15 13:52:26 -0400)
tiltfile: extensions should reset state in OnStart


nicks:
i need to think of a better way to manage extension state...one possibility is that all extensions should be non-stateful, and attach their state to the starlark thread.

PR #2350, opened 2019-10-15
dockerprune: only prune images if they match a running manifest

maiamcc:
it was a pain to run tilt.build b/c the docker pruner on a servantes run had pruned away
all my tilt.build images/caches

with this PR, the docker pruner only prunes images matching refs currently being run.

I BELIEVE this will mean that we only prune docker caches corresponding to those images -- b/c we're not pruning dangling caches, so we shouldn't delete any that relate to extant images? lmk if this is wrong.

for future PR: apply this behavior to `tilt docker-prune` so it gives an accurate preview

### DIFF
--- a/internal/cli/docker_prune.go
+++ b/internal/cli/docker_prune.go
@@ -6,27 +6,12 @@ import (
 
 	"github.com/spf13/cobra"
 
-	"github.com/windmilleng/tilt/internal/docker"
-	"github.com/windmilleng/tilt/internal/tiltfile"
-
 	"github.com/windmilleng/tilt/internal/analytics"
 	"github.com/windmilleng/tilt/internal/engine/dockerprune"
 )
 
 type dockerPruneCmd struct {
 	maxAgeStr string
-}
-
-type dpDeps struct {
-	dCli docker.Client
-	tfl  tiltfile.TiltfileLoader
-}
-
-func newDPDeps(dCli docker.Client, tfl tiltfile.TiltfileLoader) dpDeps {
-	return dpDeps{
-		dCli: dCli,
-		tfl:  tfl,
-	}
 }
 
 func (c *dockerPruneCmd) register() *cobra.Command {
--- a/internal/cli/docker_prune.go
+++ b/internal/cli/docker_prune.go
@@ -6,12 +6,27 @@ import (
 
 	"github.com/spf13/cobra"
 
+	"github.com/windmilleng/tilt/internal/docker"
+	"github.com/windmilleng/tilt/internal/tiltfile"
+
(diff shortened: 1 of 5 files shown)


internal/cli/docker_prune.go
```diff
@@ -14,6 +17,18 @@ type dockerPruneCmd struct {
 	maxAgeStr string
 }
 
+type dpDeps struct {
```
jazzdan: This type and the constructor aren't used?

internal/engine/dockerprune/docker_pruner.go
```diff
@@ -60,6 +62,14 @@ func (dp *DockerPruner) OnChange(ctx context.Context, st store.RStore) {
 	curBuildCount := state.CompletedBuildCount
 	hasDockerBuild := state.HasDockerBuild()
 	nextToBuild := buildcontrol.NextManifestNameToBuild(state)
+	var imgSelectors []container.RefSelector
+	for _, m := range state.Manifests() {
+		for _, iTarg := range m.ImageTargets {
+			sel := iTarg.ConfigurationRef
+			sel.AsNamedOnly()
```
nicks: this is a no-op?

internal/engine/dockerprune/docker_pruner.go
```diff
@@ -60,6 +62,14 @@ func (dp *DockerPruner) OnChange(ctx context.Context, st store.RStore) {
 	curBuildCount := state.CompletedBuildCount
 	hasDockerBuild := state.HasDockerBuild()
 	nextToBuild := buildcontrol.NextManifestNameToBuild(state)
+	var imgSelectors []container.RefSelector
+	for _, m := range state.Manifests() {
+		for _, iTarg := range m.ImageTargets {
+			sel := iTarg.ConfigurationRef
```
nicks: i'm pretty sure this should be DeploymentRef, not ConfigurationRef?

DeploymentRef is what it actually gets tagged with

internal/cli/docker_prune.go
maiamcc: lol ya they belong in a different PR, just removed them

internal/engine/dockerprune/docker_pruner.go
maiamcc: oh yep you're right.

internal/engine/dockerprune/docker_pruner.go
maiamcc: nope, `ConfigurationRef` is actually of type `RefSelector` 😒 but it's irrelevant now that i'm using `DeploymentRef`!

PR #2351, opened 2019-10-15
cli: make docker-prune cli behavior consistent, make logs useful

maiamcc:
`tilt docker-prune` does a docker prune as tilt would (respecting manifests defined in
this dir's tiltfile, etc.)

### DIFF
--- a/internal/engine/dockerprune/docker_pruner.go
+++ b/internal/engine/dockerprune/docker_pruner.go
@@ -62,13 +62,7 @@ func (dp *DockerPruner) OnChange(ctx context.Context, st store.RStore) {
 	curBuildCount := state.CompletedBuildCount
 	hasDockerBuild := state.HasDockerBuild()
 	nextToBuild := buildcontrol.NextManifestNameToBuild(state)
-	var imgSelectors []container.RefSelector
-	for _, m := range state.Manifests() {
-		for _, iTarg := range m.ImageTargets {
-			sel := container.NameSelector(iTarg.DeploymentRef).WithNameMatch()
-			imgSelectors = append(imgSelectors, sel)
-		}
-	}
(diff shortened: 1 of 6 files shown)


internal/engine/dockerprune/docker_pruner.go
```diff
@@ -111,21 +105,21 @@ func (dp *DockerPruner) OnChange(ctx context.Context, st store.RStore) {
...
 
 func (dp *DockerPruner) PruneAndRecordState(ctx context.Context, maxAge time.Duration, imgSelectors []container.RefSelector, curBuildCount int) {
-	dp.Prune(ctx, maxAge, imgSelectors)
+	dp.Prune(ctx, maxAge, imgSelectors, false)
 	dp.lastPruneTime = time.Now()
 	dp.lastPruneBuildCount = curBuildCount
 }
 
-func (dp *DockerPruner) Prune(ctx context.Context, maxAge time.Duration, imgSelectors []container.RefSelector) {
+func (dp *DockerPruner) Prune(ctx context.Context, maxAge time.Duration, imgSelectors []container.RefSelector, verbose bool) {
```
nicks: having a bespoke verbose parameter seems very likely to bite us, given the overlap with the existing logging system and --verbose flag. What problem are you trying to solve with this? Can we either punt on that problem for now, or solve it some better way?

maiamcc: hmm fair -- trying to get at the "nothing happened" case. If we didn't prune anything, we don't wanna bother the user if this ran in the background of tilt, but if they specifically ran `tilt docker-prune` they presumably DO want to know that we didn't find anything to prune. But you're right, i can figure this out from the log level.

maiamcc:
ready for review!

PR #2352, opened 2019-10-15
cloud: send update records

nicks:
Hello @landism,

Please review the following commits I made in branch nicks/ch3667/send:

c314c632d1d83cc2b7fe0c56ea0e53e1f5ac4d6e (2019-10-15 15:45:43 -0400)
cloud: send update records

### DIFF
--- a/internal/cloud/update_uploader.go
+++ b/internal/cloud/update_uploader.go
@@ -1,7 +1,10 @@
 package cloud
 
 import (
+	"bytes"
 	"context"
+	"encoding/json"
+	"net/http"
 	"time"
 
 	"github.com/golang/protobuf/ptypes"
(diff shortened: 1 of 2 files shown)


internal/cloud/update_uploader.go
```diff
@@ -117,16 +140,42 @@ func (u *UpdateUploader) makeUpdates(st store.RStore) []update {
...
+		logger.Get(ctx).Infof("Error encoding updates: %v", err)
+		return
+	}
+
+	request, err := http.NewRequest(http.MethodPost, u.putUpdatesURL(), buf)
+	if err != nil {
+		logger.Get(ctx).Infof("Error sending updates: %v", err)
+		return
+	}
+
```
landism: we want to set the token header here so that the server can get the user id

nicks: done!

PR #2360, opened 2019-10-16
tiltfile: a better pattern for extensions with mutable state

nicks:
Hello @jazzdan,

Please review the following commits I made in branch nicks/state:

e835b35649b297fe2964f9656499ee9b5495c083 (2019-10-16 14:51:50 -0400)
tiltfile: a better pattern for extensions with mutable state
this might be too crazy functional programming bananapants

### DIFF
--- a/internal/tiltfile/starkit/set_state.go
+++ b/internal/tiltfile/starkit/set_state.go
@@ -1,0 +1,62 @@
+package starkit
+
+import (
+	"fmt"
+	"reflect"
+
+	"go.starlark.net/starlark"
+)
+
+// SetState works like SetState in React. It can take a value or a function.
+//
+// That function should transform old state into new state.
+//
+// For example, an extension that accumulated strings might use
+// SetState() like this:
+//
+// err := starkit.SetState(t, func(strings []string) {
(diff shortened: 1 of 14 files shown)


internal/tiltfile/starkit/set_state.go
```diff
@@ -0,0 +1,62 @@
...
+func SetState(t *starlark.Thread, valOrFn interface{}) error {
+	typ := reflect.TypeOf(valOrFn)
+	model, ok := t.Local(modelKey).(Model)
+	if !ok {
+		return fmt.Errorf("Internal error: Starlark not initialized correctly: starkit.Model not found")
+	}
+
+	isFn := typ.Kind() == reflect.Func
+	if !isFn {
+		// If there's already a value with this type in the state store, overwrite it.
```
maiamcc: hmm, does this mean we can't have a state that stores, e.g., two different `int` values?

nicks: ya, i would expect that every extension will have its own type of state

nicks:
ya, i agree, we may have some rough edges to shave off! I'm open to suggestions on how to make the API easier while still making it extensible

PR #2361, opened 2019-10-16
See highlight in Snapshot

hyu:
- make sure the highlight is passed into the Snapshot state
- clear the highlight when navigating between resources

### DIFF
--- a/web/src/LogPane.tsx
+++ b/web/src/LogPane.tsx
@@ -16,6 +16,7 @@ type LogPaneProps = {
   highlight: SnapshotHighlight | null
   highlightsEnabled: boolean
   modalIsOpen: boolean
+  name: string
 }
 type LogPaneState = {
   autoscroll: boolean
@@ -43,7 +44,6 @@ class LogPane extends Component<LogPaneProps, LogPaneState> {
     if (this.lastElement !== null) {
       this.lastElement.scrollIntoView()
     }
-
     window.addEventListener("scroll", this.refreshAutoScroll, { passive: true })
     window.addEventListener("wheel", this.handleWheel, { passive: true })
     if (this.props.highlightsEnabled) {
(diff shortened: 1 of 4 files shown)


web/src/LogPane.tsx
```diff
@@ -53,6 +53,52 @@ class LogPane extends Component<LogPaneProps, LogPaneState> {
...
+    if (this.props.highlightsEnabled && this.props.modalIsOpen) {
+      document.removeEventListener(
+        "selectionchange",
+        this.handleSelectionChange
+      )
+    }
+
+    if (
+      this.props.highlightsEnabled &&
+      prevProps.modalIsOpen &&
```
jazzdan: Does this logic work for the initial view? When there is no prevProps?

web/src/LogPane.tsx
```diff
@@ -53,6 +53,52 @@ class LogPane extends Component<LogPaneProps, LogPaneState> {
     }
   }
 
+  componentDidUpdate(prevProps: LogPaneProps) {
+    if (this.props.highlightsEnabled && this.props.modalIsOpen) {
+      document.removeEventListener(
```
nicks: in general, it's going to be pretty racey/error-prone to add/remove listeners in componentDidUpdate.

have you considered adding these checks to handleSelectionChange, and simply doing nothing in that case?

web/src/LogPane.tsx
```diff
@@ -53,6 +53,52 @@ class LogPane extends Component<LogPaneProps, LogPaneState> {
...
+      !this.props.modalIsOpen
+    ) {
+      document.addEventListener("selectionchange", this.handleSelectionChange, {
+        passive: true,
+      })
+    }
+
+    if (this.props.name !== prevProps.name) {
+      let selection = document.getSelection()
+      this.props.handleClearHighlight()
```
nicks: the way this bounces control back and forth between the child component and the parent component seems hard to reason about and hard to test.

have you considered having the HUD do this in the history.listen handler? that seems a lot closer to what you want - clearing the selection every time the URL changes.

hyu: Oh, I was looking for something which would reliably fire when navigating between resources. But didn't realize it was right here. Thanks for pointing it out!

web/src/LogPane.tsx
hyu: Good point! This simplifies the code nicely.

web/src/LogPane.tsx
hyu: I changed this around per Nick's advice below, so this piece is no longer relevant.

Issue #2348, opened 2019-10-15
PersistentVolumeClaims (and others) are not necessarily applied first


landism:
When we have a bunch of k8s yaml, some objects depend on other objects, and if A depends on B, we need ensure B is applied first (e.g., if A is in namespace B, or A has PersistentVolumeClaim B). Otherwise, A's application may fail because B does not exist.

Tilt currently attempts to do this by putting namespaces and CRDs first, but misses other cases such as PersistentVolumeClaims ([code](https://github.com/windmilleng/tilt/blob/92a183dcf837504191e851c18ec8a7fd0cd7422f/internal/k8s/entity.go#L205)).

Kustomize's approach is [much more thorough](https://github.com/kubernetes-sigs/kustomize/blob/b878cd050d5ffe5d485b34b496c915fa96405db9/pkg/gvk/gvk.go#L82).

Presumably a simple type ordering solves the vast majority of these problems, since it seems fine for Kustomize. I don't know if there are real or theoretical cases where, e.g. an object of type A could depend on an object of type B and also vice versa, at which point we'd need to get into which specific objects depend on which specific objects. 

IIUC, `kubectl apply` solves this problem, but I don't know how. Any solution it has would only work if all the objects are provided in a single invocation, and thus Tilt doesn't get it for free, since Tilt breaks application up into multiple invocations of `kubectl apply`. I looked briefly for any code that does this in k8s and didn't find it.

jazzdan:
>Tilt currently attempts to do this by putting namespaces and CRDs first, but misses other cases such as CRDs 

@landism did you mean to mention CRDs twice here?

landism:
> @landism did you mean to mention CRDs twice here?

fixed

maiamcc:
> IIUC, kubectl apply solves this problem, but I don't know how
I'm not seeing that, fwiw. I'm playing with the examples in the [Persistent Volume Storage tutorial](https://kubernetes.io/docs/tasks/configure-pod-container/configure-persistent-volume-storage/), and it seems (according to `kubectl apply --dry-run` and to observed behavior) that `kubectl apply` respects the order in the yaml.

e.g. if I apply a yaml file containing:
```
<volume claim yaml>
---
<volume yaml>
```

the volume claim comes up in not too much time, but when i `describe` it I see a previous`ProvisioningFailed` error: `storageclass.storage.k8s.io "manual" not found` (implying that it errored while trying to spin up before the Volume existed).

[...]

PR #2364, opened 2019-10-17
k8s: more comprehensive ordering of k8s entities

maiamcc:
addresses #2348 (verified on [this example repo](https://github.com/windmilleng/yaml_ordering))

we still don't print the YAMLs in the order we apply them, which isn't great, but this solves the underlying issue

honestly we should probably sort the yamls once, on Tiltfile execution.


internal/k8s/entity_sort.go
```diff
@@ -0,0 +1,53 @@
+package k8s
+
+// An attempt to order things to help k8s, e.g.
+// a Service should come before things that refer to it.
+// Namespace should be first.
+// In some cases order just specified to provide determinism.
+// Borrowed from Kustomize: https://github.com/kubernetes-sigs/kustomize/blob/b878cd050d5ffe5d485b34b496c915fa96405db9/pkg/gvk/gvk.go#L82
```
nicks: have we considered linking in the Kustomize code directly?

otherwise, we should probably move this to its own file and add appropriate copyright headers, as in 
https://github.com/windmilleng/tilt/blob/d35c73c006f21488b241ccf58e7c3cc529cd0ec8/internal/kustomize/dependencies.go

maiamcc: The Kustomize code doesn't have everything we need, alas -- notably it's missing `PersistentVolume` which is the reason I'm putting in this PR in the first place. i'll add the copyright headers though.

PR #2365, opened 2019-10-17
assets: refactor the asset server so that the asset directory isn't hard-coded

nicks:
Hello @jazzdan, @landism,

Please review the following commits I made in branch nicks/assets:

eb9746688d144702aa63701b54655dcb4cf21649 (2019-10-17 13:07:35 -0400)
assets: refactor the asset server so that the asset directory isn't hard-coded


nicks:
this is 95% just moving code around, the important change is how the provider got moved to the cli package

PR #2368, opened 2019-10-17
web/css: fix logs vanishing under sidebar at small browser widths

### LINKED ISSUE #2276: log line-wrapping doesn't account for the sidebar
When Tilt wraps log lines in the Web UI, it does so at the right edge of the window, as if the side bar didn't exist.

e.g.:
image

and even when the sidebar is collapsed:
image

maiamcc:
fixes #2276

### DIFF
--- a/web/src/LogPane.scss
+++ b/web/src/LogPane.scss
@@ -1,7 +1,7 @@
 @import "constants.scss";
 
 .LogPane {
-  width: 100%;
+  width: calc(100% - #{$sidebar-width});
   margin-top: $tabnav-height + $resourceInfo-height;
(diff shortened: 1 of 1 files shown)


nicks:
though honestly, if we're going to adjust the width of the LogPane when the floating element changes, then it's not totally clear to me why the element floats...

hyu:
Good point @nicks! I initially thought it would be fine for the Sidebar to cover up some log text, since you can always collapse it. But after living with it, that turns out to be annoying and pretty much not a situation you ever want. Your suggestion would be a good jank fix.

Issue #2371, opened 2019-10-18
Buildkit Dockerfile experimental syntax does not work for child images


mvgijssel:
Using the buildkit [Dockerfile frontend experimental syntaxes](https://github.com/moby/buildkit/blob/master/frontend/dockerfile/docs/experimental.md) works for regular `docker_build` statements in the Tiltfile but not when using a base image.

For example:

## Tiltfile
```python
docker_build(
    'localhost:5000/child',
    'child',
    dockerfile="Dockerfile.child",
)

docker_build(
    'localhost:5000/base',
    'base',
    dockerfile="Dockerfile.base",
)
```

## Dockerfile.base
```dockerfile
# syntax = docker/dockerfile:experimental
FROM ubuntu:xenial

RUN --mount=type=cache,target=/var/cache/apt \
    apt-get update && apt-get install -y vim
```

## Dockerfile.child
```dockerfile
# syntax = docker/dockerfile:experimental
FROM localhost:5000/base

RUN --mount=type=cache,target=/var/cache/apt \
    apt-get update && apt-get install -y ruby
```

The `base` image will build just fine, but the `child` image will fail with something like:

```shell
ERROR: ImageBuild: failed to create LLB definition: Dockerfile parse error line 26: Unknown flag: mount
got unexpected error during build/deploy: ImageBuild: failed to create LLB definition: Dockerfile parse error line 26: Unknown flag: mount
```

I believe this has something to do with the fact that docker images using a base image will be parsed by tilt to replace the FROM with the tilt tagged image. 

The parsing step removes all comments, including the buildkit header `# syntax = docker/dockerfile:experimental` which results in build failure.

---

### tilt version
```
v0.10.13, built 2019-10-04
```

### docker version
```
Client: Docker Engine - Community
 Version:           19.03.3
 API version:       1.39 (downgraded from 1.40)
 Go version:        go1.13.1
 Git commit:        a872fc2
 Built:             Tue Oct  8 22:01:16 2019
 OS/Arch:           darwin/amd64
 Experimental:      false

Server: Docker Engine - Community
 Engine:
  Version:          18.09.9
  API version:      1.39 (minimum version 1.12)
  Go version:       go1.11.13
  Git commit:       039a7df9ba
  Built:            Wed Sep  4 16:55:50 2019
  OS/Arch:          linux/amd64
  Experimental:     false
```

maiamcc:
@mvgijssel ty for the very thorough report! Will look into this now.

[...]

PR #2373, opened 2019-10-18
engine: when parsing/unparsing dockerfile to inject image deps, preserve '# syntax =' flag

maiamcc:
Hello @nicks, @landism,

Please review the following commits I made in branch maiamcc/dockerfile-syntax-flag:

34ae0c55498fb2a380bfb5568a340e24d8303bca (2019-10-18 13:49:28 -0400)
engine: when parsing/unparsing dockerfile to inject image deps, preserve '# syntax =' flag

Addresses #2371. Verified on [this example branch](https://github.com/windmilleng/servantes/tree/experimental-dockerfile-child-image).

I'm gonna file an issue with buildkit about pulling in syntax flags during parsing, tho idk if they'll care.


internal/dockerfile/dockerfile.go
```diff
@@ -62,6 +69,17 @@ func (d Dockerfile) RmPaths(pathsToRm []string) Dockerfile {
 	return d.Join(fmt.Sprintf("RUN %s", rmCmd.String()))
 }
 
+// Pull out the `# syntax = docker/xxx` line, if one exists.
+func (d Dockerfile) MaybeSyntaxFlag() Dockerfile {
```
nicks: this is not quite right. can you use the docker parsing code for this?
https://github.com/moby/buildkit/blob/3f8ab160d5079539f9ed971fb069e4205108dd9d/frontend/dockerfile/dockerfile2llb/directives.go#L27

internal/engine/image_build_and_deployer.go
```diff
@@ -391,6 +392,13 @@ func injectImageDependencies(iTarget model.ImageTarget, iTargetMap map[model.Tar
 		return model.ImageTarget{}, errors.Wrap(err, "injectImageDependencies")
 	}
 
+	// HACK(maia): until buildkit/parse supports `# syntax = ` flags, check whether the original
```
nicks: this feels like the wrong place to put this code. Can you move this into the ast and Print() function?

maiamcc: fair, will do

internal/dockerfile/dockerfile.go
maiamcc: amazing, i didn't realize that existed!

internal/dockerfile/ast.go
```diff
@@ -113,6 +116,17 @@ func (a AST) traverseNode(node *parser.Node, visit func(*parser.Node) error) err
 func (a AST) Print() (Dockerfile, error) {
 	buf := bytes.NewBuffer(nil)
 	currentLine := 1
+
+	directiveFmt := "# %s = %s\n"
+	for k, v := range a.directives {
+		dir := fmt.Sprintf(directiveFmt, k, v)
+		_, err := buf.Write([]byte(dir))
```
nicks: you can collapse these two lines with an Fprintf

PR #2379, opened 2019-10-18
tilt: make backend able to send snapshots

landism:
this ended up being surprisingly annoying, mostly coming down to:
1. We had code in the TS that cleaned up the snapshot object before posting it. I wanted to move this into the backend, so that the backend could generate a snapshot.
2. This meant the backend needed to deserialize the object and know the snapshot structure. We have nothing enforcing that the `view` types in the backend and frontend match. It turns out the frontend had an extra `endpoints` field, and set some timestamp fields to empty string, which the default go serializer doesn't like.
3. The view had an `error` field, which is not roundtrip-serializable (it comes back as `nil`)
4. The alert handling is still entirely on the JS side, so now we need client-side types to accept them. NB: this means that snapshots created by the backend won't have alerts.


landism:
> tbh i feel like we're eventually going to have to give up and make webview a protobuf

I don't even consider that giving up - the current situation is messy and bugprone and expensive

PR #2382, opened 2019-10-19
tiltfile: custom host in port forwards

### LINKED ISSUE #2013: Change host used with port_forwards
From @dotellie in #tilt: 

> Is there an easy way to change the host used with `port_forwards`? I'm wanting to connect to my application from my phone which is on the same (trusted) local network which currently doesn't seem possible since Tilt binds to `localhost`.

dolsem:
Fixes #2013.

If this gets merged, I can create a PR to update the docs as well.

### DIFF
--- a/internal/tiltfile/k8s.go
+++ b/internal/tiltfile/k8s.go
@@ -2,6 +2,7 @@ package tiltfile
 
 import (
 	"fmt"
+	"regexp"
 	"sort"
 	"strconv"
 	"strings"
@@ -775,6 +776,7 @@ func (s *tiltfileState) portForward(thread *starlark.Thread, fn *starlark.Builti
 type portForward struct {
 	local     int
 	container int
+	host      *string
 }
 
 var _ starlark.Value = portForward{}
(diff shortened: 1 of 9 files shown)


dolsem:
The way to specify host is with a string like '0.0.0.0:3000:3000', which I believe should be standard.

internal/tiltfile/k8s.go
```diff
@@ -800,35 +802,54 @@ func (f portForward) Hash() (uint32, error) {
...
 	}
 	if n < 0 || n > 65535 {
-		return portForward{}, fmt.Errorf("portForward value %v is not in the range for a port [0-65535]", n)
+		return portForward{}, fmt.Errorf("portForward port value %v is not in the valid range [0-65535]", n)
 	}
 	return portForward{local: int(n)}, nil
 }
 
+const ipReStr = `^(([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])\.){3}([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])$`
+const hostnameReStr = `^(([a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9\-]*[a-zA-Z0-9])\.)*([A-Za-z0-9]|[A-Za-z0-9][A-Za-z0-9\-]*[A-Za-z0-9])$`
```
nicks: where did these regexps come from?

internal/tiltfile/k8s.go
```diff
@@ -800,35 +802,54 @@ func (f portForward) Hash() (uint32, error) {
...
+
+var validHost = regexp.MustCompile(ipReStr + "|" + hostnameReStr)
+
 func stringToPortForward(s starlark.String) (portForward, error) {
-	parts := strings.SplitN(string(s), ":", 2)
-	local, err := strconv.Atoi(parts[0])
+	parts := strings.SplitN(string(s), ":", 3)
```
nicks: where did this syntax come from?

it's probably ok, but seems different from the `kubectl port-forward` syntax (which has a separate `--address` flag). https://github.com/kubernetes/kubectl/blob/master/pkg/cmd/portforward/portforward.go#L87

i guess the advantage of the separate address flag is that you can specify multiple

pkg/model/manifest.go
```diff
@@ -368,12 +368,15 @@ func ToRuns(cmds []Cmd) []Run {
...
-
 	// The port to connect to inside the deployed container.
 	// If 0, we will connect to the first containerPort.
 	ContainerPort int
+
+	// The port to expose on the current machine.
+	LocalPort int
+
+	// Optional host to bind to on the current machine (localhost by default)
+	Host *string
```
nicks: what's the purpose of making it a pointer? would we want to treat the `nil` value differently than the empty string?

internal/tiltfile/k8s.go
dolsem: Not sure what exactly you're asking. These are pretty standard regexps for IPs and hostnames, not sure if it's possible to pinpoint their exact source, cause they are pretty standard, based on relevant RFCs. Here are examples of where they can be encountered online:
IP: http://ipregex.com
Hostname: https://www.regextester.com/23

I've been using them for a while and I don't remember where I took them from, but again, they are pretty standard, so probably shouldn't matter.

internal/tiltfile/k8s.go
dolsem: Just as with the previous discussion, not sure what the exact origin of this syntax is, but it seems to be the most commonly used way to specify hostname for port forwarding, one example is OpenSSH.
You're right, it doesn't allow for multiple addresses, so if we need to support that, we should probably change syntax to something else. That being said, in my experience there's barely ever a need to bind to multiple addresses, you usually want it to either be accessible on a specific network, or accessible on every network (which can be done by specifying `0.0.0.0` as the address).

pkg/model/manifest.go
dolsem: Two thoughts:
A) Even if a piece of information is not currently used, but might be used in the future, and capturing it doesn't require any overhead, imo it's better to capture it now than to refactor later.
B) Right now if the value is `nil` we do port forwarding without addresses, and if the value is empty string we simply pass it to `NewOnAddress`, which I'm assuming explicitly fails with some error (tho I haven't tested it tbh). So this signature makes it slightly easier to catch bugs in case hostname is meant to be supplied, but ended up being empty string. This is probably rather opinionated.

nicks:
ok, sgtm. Two minor points:

1) re: the hostname regexp, you'd be surprised how hard this is to get right! See: https://github.com/golang/go/issues/17659

2) re: ya, I don't think the relative pros vs cons of `*string` vs `string` are that big, it's more important that the codebase is internally consistent. going to to change it to a string for consistency.

thanks for the addition!

dolsem:
@nicks 
1. Interesting, thanks for the link!
2. That's fair.

Thank you for your prompt review and merge!

PR #2385, opened 2019-10-21
web: don't highlight endpoint link after clicking

### LINKED ISSUE #2380: Clicking on endpoint link shouldn't highlight it
When clicking the endpoint link in the web UI the link also becomes highlighted. It makes sense for fields that you would copy, but a click should't do two things in that way. It looks a bit odd when you return to the UI.

I'd propose to add a "copy to clipboard" button, or to simply remove the highlight for the click.

I'll make a PR if it's a change that might be wanted.

And as you might already have heard, this is a damn fine project. Keep up the good work! 🎉

landism:
fixes #2380

### DIFF
--- a/web/src/ResourceInfo.scss
+++ b/web/src/ResourceInfo.scss
@@ -35,7 +35,6 @@
 .resourceInfo-value {
   display: inline-block;
   font-family: $font-monospace;
-  user-select: all;
   margin: 0;
   text-overflow: ellipsis;
   overflow: hidden;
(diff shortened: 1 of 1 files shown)


nicks:
this doesn't seem right. there are other elements of the ui that need this

nicks:
i would probably leave the existing css as-is, and add a rule like: this so that 
```
a.resourceInfo-value {
user-select: auto;
}
```

PR #2391, opened 2019-10-21
web: scroll highlight in to view if it is a snapshot

jazzdan:


web/src/LogPane.test.tsx
```diff
@@ -404,9 +409,34 @@ it("renders highlighted lines", () => {
...
         modalIsOpen={false}
+        isSnapshot={false}
       />
     )
     .toJSON()
 
   expect(tree).toMatchSnapshot()
 })
+
+it("scrolls to highlighted lines in snapshot", () => {
```
jazzdan: Relevant new test buried by large diff caused by my re-arranging the log fixture here.

PR #2395, opened 2019-10-22
web: also store highlighted text in snapshot

jazzdan:


jazzdan:
@nicks I want to be able to include the highlighted text next to the snapshot in the snapshot list server side. This seemed the simplest way to achieve that.

hyu:
I'm not sure if the highlight text in this form is useful. When we use `toString()`, it doesn't appear to preserve the line breaks. (Do you know if this is actually the case, or am I missing something?)

In any instance where I want to display the text of the highlight, I want it to be as close as possible to the original form of the highlight. So it's important to preserve the same line breaks as what we see in the LogPane.

jazzdan:
@hyu it seems to preserve the newline characters to me:

Screen Shot 2019-10-22 at 11 51 33 AM

Issue #1997, opened 2019-08-06
Better error messaging about single pod support for LiveUpdate


maiamcc:
Spent a load of time trying to help a user debug a LiveUpdate issue where they consistently got "don't have info for running container" error and fell back to a full image build.

The issue was that they were running >1 replicas, and when trying to get ContainerInfo for multiple pods at once, [we silently bail](https://github.com/windmilleng/tilt/blob/e2f4ad8c83e826c92312e8fba309028b748e0f08/internal/store/build_result.go#L256) instead of at least raising a useful error message.

maiamcc:
Re-opening this ticket because, while we DO log a descriptive error message when we hit this case, it can quickly scroll off-screen. (e.g. @MattHodge recently missed this message.) We should consider surfacing this error as an Alert -- if the user is trying to live update multiple pods at a time, and LiveUpdate is failing b/c of that, they presumably want to know.

rhyek:
Any plans to support replicas > 1? It is sometimes useful (and depending on the type of service actually required) to test horizontal scaling during development and it would be great to avoid frequent full rebuilds.

PR #2399, opened 2019-10-22
engine: when we fail to get the running container info, print the error on Redirect so it shows up in the right resource view

nicks:
Hello @maiamcc,

Please review the following commits I made in branch nicks/containerrunning:

663ed1dc14fe74d72be56d3544c2174f112882eb (2019-10-22 17:15:23 -0400)
engine: when we fail to get the running container info, print the error on Redirect so it shows up in the right resource view


maiamcc:
I think this addresses #1997, and you can close that issue when you merge this!

PR #2404, opened 2019-10-23
webview: simple GRPC server

jazzdan:
The point of this commit is just to validate the skeleton of the GRPC server,
not to have something useful.

This will be expanded in the future to fully implement all of the APIs
in the HUD webserver (analytics, triggers, snapshots, etc) EXCEPT the
websocket.

Other future work includes generating a TypeScript client for use
in the web frontend.


pkg/webview/view.proto
```diff
@@ -0,0 +1,44 @@
...
+  string tiltCloudUsername = 8;
+  string tiltCloudSchemeHost = 9;
+  string tiltCloudTeamID = 10;
+
+  string fatalError = 11;
+}
+
+message GetViewRequest {}
+
+service ViewService {
```
jazzdan: Is this the right name for this service?

Makefile
```diff
@@ -24,7 +24,7 @@ scripts/protocc/protocc.py: scripts/protocc
 	git submodule update
 
 proto: scripts/protocc/protocc.py
-	python3 scripts/protocc/protocc.py --out go
+	python3 scripts/protocc/protocc.py --out go && toast proto
```
nicks: this isn't right. right now, this makefile rule only depends on scripts/protocc/protocc.py.

can you create a new makefile rule with the correct dependencies?

internal/hud/server/server.go
```diff
@@ -64,11 +68,19 @@ func ProvideHeadsUpServer(
...
 		store:    store,
 		router:   r,
 		a:        analytics,
 		uploader: uploader,
+		grpcMux:  grpcMux,
+	}
+
+	err := proto_webview.RegisterViewServiceHandlerServer(context.TODO(), grpcMux, s)
+	if err != nil {
+		// TODO(dmiller) what should we do here?
```
nicks: let's change this provider to return an error and re-run `wire`. it should handle this correctly.

internal/hud/server/server.go
```diff
@@ -64,11 +68,19 @@ func ProvideHeadsUpServer(
...
+	grpcMux := runtime.NewServeMux()
 	s := &HeadsUpServer{
 		store:    store,
 		router:   r,
 		a:        analytics,
 		uploader: uploader,
+		grpcMux:  grpcMux,
+	}
+
+	err := proto_webview.RegisterViewServiceHandlerServer(context.TODO(), grpcMux, s)
```
nicks: let's change this provider to take a context.Context and re-run `wire`. It should handle this correctly.

toast.yml
```diff
@@ -0,0 +1,30 @@
...
+  apk_update:
+    command: apk update
+  install_protoc_deps:
+    dependencies:
+      - apk_update
+    command: apk add protobuf protobuf-dev git
+  install_protoc_gen_go:
+    dependencies:
+      - install_protoc_deps
+    command: go get -u github.com/golang/protobuf/protoc-gen-go github.com/golang/protobuf/jsonpb github.com/golang/protobuf/proto github.com/grpc-ecosystem/grpc-gateway/protoc-gen-grpc-gateway google.golang.org/grpc
```
nicks: i'm a little bit nervous about these not being vendored properly, and that causing a lot of problems. protocc was meant to solve these problems

Makefile
```diff
@@ -170,3 +173,9 @@ prettier:
 storybook:
 	cd web && yarn install
 	cd web && yarn storybook
+
+tilt-container:
```
nicks: maybe call this tilt-toast-container?

this should also be in the .PHONY declaration

internal/hud/server/server.go
```diff
@@ -56,19 +59,28 @@ type HeadsUpServer struct {
...
+	grpcMux           *runtime.ServeMux
 }
 
 func ProvideHeadsUpServer(
 	store *store.Store,
 	assetServer assets.Server,
 	analytics *tiltanalytics.TiltAnalytics,
-	uploader cloud.SnapshotUploader) *HeadsUpServer {
+	uploader cloud.SnapshotUploader,
+	ctx context.Context) (*HeadsUpServer, error) {
```
nicks: fwiw, i would usually put ctx first, just as a style thing

pkg/webview/view.proto
nicks: ya seems fine

pkg/webview/view.proto
```diff
@@ -0,0 +1,44 @@
...
+package webview;
+
+import "google/api/annotations.proto";
+
+// TODO(dmiller) fill this in
+message Resource {}
+
+message TiltBuild {
+  string version = 1;
+  string commitSHA = 2;
```
nicks: proto naming style is slug case - commit_SHA

pkg/webview/view.proto
```diff
@@ -0,0 +1,44 @@
...
+  string date = 3;
+  bool dev = 4;
+}
+
+message View {
+  string log = 1;
+  repeated Resource resources = 2;
+  bool logTimestamps = 3;
```
nicks: ditto here - log_timestamps

pkg/webview/view.proto
```diff
@@ -0,0 +1,44 @@
...
+
+  map<string, bool> featureFlags = 4;
```
nicks: and so on

PR #2407, opened 2019-10-23
dc: make image actually optional

landism:
this previously generated an "invalid reference format" error, since we were happily passing `nil` to `ParseNamed`

### DIFF
--- a/internal/tiltfile/docker_compose.go
+++ b/internal/tiltfile/docker_compose.go
@@ -93,11 +93,12 @@ func (s *tiltfileState) dcResource(thread *starlark.Thread, fn *starlark.Builtin
 		return nil, fmt.Errorf("dc_resource: `name` must not be empty")
 	}
 
-	var imageRefAsStr string
+	var imageRefAsStr *string
 	switch imageVal := imageVal.(type) {
 	case nil: // optional arg, this is fine
 	case starlark.String:
-		imageRefAsStr = string(imageVal)
+		s := string(imageVal)
+		imageRefAsStr = &s
 	default:
 		return nil, fmt.Errorf("image arg must be a string; got %T", imageVal)
 	}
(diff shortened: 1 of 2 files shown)


internal/tiltfile/docker_compose.go
```diff
@@ -93,11 +93,12 @@ func (s *tiltfileState) dcResource(thread *starlark.Thread, fn *starlark.Builtin
 		return nil, fmt.Errorf("dc_resource: `name` must not be empty")
 	}
 
-	var imageRefAsStr string
+	var imageRefAsStr *string
 	switch imageVal := imageVal.(type) {
 	case nil: // optional arg, this is fine
```
maiamcc: ...wait, does this `case nil` ever trip? seems like the least change here would be to, in the nil case, set `imageRefAsStr = ""` and check below if it's the empty string, no?

internal/tiltfile/docker_compose.go
```diff
@@ -93,11 +93,12 @@ func (s *tiltfileState) dcResource(thread *starlark.Thread, fn *starlark.Builtin
 		return nil, fmt.Errorf("dc_resource: `name` must not be empty")
 	}
 
-	var imageRefAsStr string
+	var imageRefAsStr *string
```
nicks: what's the point of distinguishing `nil` from the empty string? seems like you'd want them to behave the same.

landism:
> ...wait, does this case nil ever trip? seems like the least change here would be to, in the nil case, set imageRefAsStr = "" and check below if it's the empty string, no?

> what's the point of distinguishing nil from the empty string? seems like you'd want them to behave the same.

Since you both kind of asked the same question, replying here instead:

If we treat nil and empty string the same, then `dc_resource("foo", image="")` is the same as `dc_resource("foo")`, and won't generate an error, even though in the former it seems like the user has some intent we're not able to fulfill. This is not a huge deal, but would make it easier to debug a case where a user has `dc_resource("foo", image=myFunc())` and `myFunc` returns the empty string, and then the user spends a bunch of time trying to figure out why Tilt is silently ignoring their parameter before realizing that oh maybe the bug was on their end.

The benefit might be small, but so is the cost.

PR #2408, opened 2019-10-23
tiltfile: pass manifest names from cmdline args around as slice instead of map

landism:
I'm guessing this was a map for historical reasons prior to multiple refactorings, and it's a bit weird for it to still be one now.

### DIFF
--- a/internal/engine/upper.go
+++ b/internal/engine/upper.go
@@ -107,11 +107,6 @@ func (u Upper) Start(
 	}
 
 	var manifestNames []model.ManifestName
-	matching := map[string]bool{}
-	for _, arg := range args {
-		manifestNames = append(manifestNames, model.ManifestName(arg))
-		matching[arg] = true
-	}
 
 	configFiles := []string{absTfPath}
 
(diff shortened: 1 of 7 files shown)


internal/engine/upper.go
```diff
@@ -107,11 +107,6 @@ func (u Upper) Start(
 	}
 
 	var manifestNames []model.ManifestName
```
maiamcc: is this never populated? do we use it anywhere?

internal/tiltfile/tiltfile_state.go
```diff
@@ -1004,28 +1004,33 @@ func (s *tiltfileState) extractEntities(dest *k8sResource, imageRef container.Re
 }
 
 // If the user requested only a subset of manifests, filter those manifests out.
-func match(manifests []model.Manifest, matching map[string]bool) ([]model.Manifest, error) {
-	if len(matching) == 0 {
+func match(manifests []model.Manifest, requestedManifests []model.ManifestName) ([]model.Manifest, error) {
```
maiamcc: ...a nit and i know you didn't write this line of code, but the comment above makes it seem like we're EXcluding the requested manifests. maybe: "...return only those manifests"? but also 🤷‍♀

internal/engine/upper.go
landism: oooops

PR #2409, opened 2019-10-23
web: grpc typescript (WIP)

jazzdan:


web/src/viewServiceClientPb.ts
```diff
@@ -6,55 +6,53 @@
 
 // GENERATED CODE -- DO NOT EDIT!
 
+import * as grpcWeb from "grpc-web"
 
-import * as grpcWeb from 'grpc-web';
+// NOTE(dmiller): why do we have to delete this?
+//import * as google_api_annotations_pb from '../../google/api/annotations_pb';
```
jazzdan: It seems insistent on generating this, but it's never used and that's an error.

web/package.json
```diff
@@ -16,6 +16,8 @@
     "@types/react-timeago": "^4.1.0",
     "@types/styled-components": "^4.1.19",
     "ansi-to-react": "^5.0.0",
+    "google-protobuf": "^3.10.0",
```
jazzdan: Do we need both of these?

web/src/HUD.tsx
```diff
@@ -110,6 +112,14 @@ class HUD extends Component<HudProps, HudState> {
       // we don't want to run any bootstrapping code in the test environment
       return
     }
+    // TODO(dmiller): remove this
```
jazzdan: This is just some test code to see it working end-to-end.

web/src/view_pb.js
```diff
@@ -0,0 +1,956 @@
+/* eslint-disable */
+// NOTE(dmiller): ^I have to add this
```
jazzdan: Needed to add this

jazzdan:
Another problem: the GRPC server responds on http://localhost:10350/api/proto/view but the generated code is making a request to http://localhost:10350/webview.ViewService/GetView

nicks:
hmmm...let's talk about this tomorrow. in TFT, the standard pattern was to add an annotation to the protobuf that declared the url path, so i'm surprised that's not working here

toast.yml
```diff
@@ -1,9 +1,10 @@
...
 tasks:
   proto:
     input_paths:
       - pkg/webview/view.proto
     output_paths:
       - pkg/webview/view.pb.go
       - pkg/webview/view.pb.gw.go
-    command: protoc -I. -I/usr/include -I"$(go env GOPATH)"/src/github.com/grpc-ecosystem/grpc-gateway/third_party/googleapis --grpc-gateway_out=logtostderr=true:. --go_out=plugins=grpc,paths=source_relative:. pkg/webview/view.proto && goimports -local github.com/windmilleng/tft -w pkg/webview/view.pb.go
+      - web/src/
+    command: mkdir -p web/src && protoc -I. -I/usr/include -I"$(go env GOPATH)"/src/github.com/grpc-ecosystem/grpc-gateway/third_party/googleapis --grpc-gateway_out=logtostderr=true:. --go_out=plugins=grpc,paths=source_relative:. --js_out=import_style=commonjs:web/src --grpc-web_out=import_style=typescript,mode=grpcwebtext:web/src pkg/webview/view.proto && goimports -local github.com/windmilleng/tft -w pkg/webview/view.pb.go && mv web/src/Pkg/webview/* web/src && mv web/src/pkg/webview/* web/src && rm -rf web/src/Pkg && rm -rf web/src/pkg
```
nicks: can you move this into a shell script? this is becoming unreadable

nicks: also, i don't totally understand why you have both import_style=commonjs and import_style=typescript. i would expect those to be mutually exclusive

jazzdan: Yeah @nicks ... I was confused about that too. I was just following an example I found with GitHub search, but it seemed wrong to me.

web/src/view_pb.js
```diff
@@ -0,0 +1,954 @@
...
+ * Generated by JsPbCodeGenerator.
+ * @param {Array=} opt_data Optional initial data array, typically from a
+ * server response, or constructed directly in Javascript. The array is used
+ * in place and becomes part of the constructed object. It is not cloned.
+ * If no data is provided, the constructed object will be empty, but still
+ * valid.
+ * @extends {jspb.Message}
+ * @constructor
+ */
+proto.webview.Resource = function(opt_data) {
```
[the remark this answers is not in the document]
jazzdan: Yeah, so it's imported from the typescript code:

```
import {
  GetViewRequest,
  View} from './view_pb';
```

toast.yml
jazzdan: but anyways I moved it in to a "shell script" :)

PR #2416, opened 2019-10-24
tiltfile: add resource_deps arg

landism:
This adds a new `resource_deps` arg to `k8s_resource`, `dc_resource`, and `local_resource`.
It does not yet affect build order.

It does:
1. populate the engine state with those deps so that the build controller will be able to use it in a future PR
2. validate the deps
3. change `tilt up foo` to also get `foo`'s transitive dependencies.

### DIFF
--- a/internal/tiltfile/tiltfile_state.go
+++ b/internal/tiltfile/tiltfile_state.go
@@ -7,14 +7,11 @@ import (
 	"regexp"
 	"strings"
 
-	"go.starlark.net/syntax"
-
-	"github.com/windmilleng/tilt/internal/tiltfile/dockerprune"
-	"github.com/windmilleng/tilt/internal/tiltfile/git"
-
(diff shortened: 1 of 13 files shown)


internal/tiltfile/tiltfile_state.go
```diff
@@ -1404,6 +1450,46 @@ func (s *tiltfileState) translateLocal() ([]model.Manifest, error) {
...
+			}
+			if _, ok := knownResources[b]; !ok {
+				return fmt.Errorf("resource %s specified a dependency on unknown resource %s", m.Name, b)
+			}
+			edges[m.Name] = append(edges[m.Name], b)
+		}
+	}
+
+	// check for cycles
+	connections := tarjan.Connections(edges)
```
dbentley: They named the library after Tarjan? Amazing.

Issue #2415, opened 2019-10-24
`helm --namespace` support doesn't actually define the namespace


maiamcc:
Was digging into #2414 and realized that [we DO in fact have support for `--namespace` in the code](https://github.com/windmilleng/tilt/blob/29ad19f92632fa87b55a20ffffd277c85334a8f9/internal/tiltfile/files.go#L220); however, we inject `namespace: foo` for every resource in the generated yaml, but doesn’t actually inject yaml that defines the namespace. Thus if you use the `helm.namespace` arg and the namespace doesn't yet exist, you get an error:
```
got unexpected error during build/deploy: kubectl apply:
stderr: Error from server (NotFound): error when creating "STDIN": namespaces "beepboop" not found
Error from server (NotFound): error when creating "STDIN": namespaces "beepboop" not found
: exit status 1
```

repro:
- in [`tft/snapshots_frontend`](https://github.com/windmilleng/tft/tree/master/snapshot_frontend), create a new Tiltfile:
```
yaml = helm('./chart', namespace='beepboop')
print('~~~~~\n'+str(yaml)+'\n~~~~~')
k8s_yaml(yaml)
```
- `tilt up`, and observe the above "namespaces 'beepboop' not found" erorr
- if you remove the `namespace` arg, the yaml applies without error.

nicks:
looks like this got fixed in #2418

maiamcc:
Ah yep, was waiting until it went out in a release to follow up.

PR #2418, opened 2019-10-24
tiltfile: if --namespace flag passed, helm builtin should also add yaml defining the namespace

maiamcc:
addresses #2415

### DIFF
--- a/internal/k8s/entity.go
+++ b/internal/k8s/entity.go
@@ -395,3 +395,22 @@ func (e K8sEntity) HasKind(kind string) bool {
 	// TODO(maia): support kind aliases (e.g. "po" for "pod")
 	return strings.ToLower(e.GVK().Kind) == strings.ToLower(kind)
 }
+
+func NewNamespaceEntity(name string) K8sEntity {
+	yaml := fmt.Sprintf(`apiVersion: v1
+kind: Namespace
+metadata:
+  name: %s
+`, name)
+	entities, err := ParseYAMLFromString(yaml)
+
+	// Something is wrong with our format string; this is definitely on us
+	if err != nil {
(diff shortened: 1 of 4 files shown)


internal/k8s/entity.go
```diff
@@ -395,3 +395,22 @@ func (e K8sEntity) HasKind(kind string) bool {
 	// TODO(maia): support kind aliases (e.g. "po" for "pod")
 	return strings.ToLower(e.GVK().Kind) == strings.ToLower(kind)
 }
+
+func NewNamespaceEntity(name string) K8sEntity {
+	yaml := fmt.Sprintf(`apiVersion: v1
```
nicks: this is fine for now, but i'm a bit worried about the case where you have an existing namespace that gets overridden with different values

maiamcc: oh hmm. Yeah good point. (you mean an existing namespace of `name` that also has like, some other metadata?)

added a commit that addresses this, tho if it's getting too finicky i can leave that safeguard off. wdyt?

nicks: ya, exactly. i'm not totally sure what will happen if you overwrite someone's namespace with different metadata

i believe Helm checks the namespace before creating it
https://github.com/helm/helm/blob/release-2.15/pkg/kube/namespace.go#L43

nicks: and fwiw, the issue isn't that you're overwriting the namespace in the chart, the issue is that you're overwriting some previously-installed namespace in the cluster

maiamcc: that makes a lot more sense yeah. I guess I'll leave the safeguard against overwriting a namespace in your chart in, but ugh, that's more complicated...

PR #2422, opened 2019-10-24
web: add generators for typescript defs

nicks:
Hello @jazzdan,

Please review the following commits I made in branch nicks/web-grpc-typescript:

4f5b3ac57227653beda0ab3954f4d1b5a4ee16c2 (2019-10-24 18:02:06 -0400)
web: add generators for typescript defs


web/src/view.d.ts
```diff
@@ -0,0 +1,22 @@
+declare namespace OpenAPI2 {
+  export interface webviewView {
+    log?: string;
```
nicks: this is the main thing to look at

separately, we should strategize on how to substitute these generated types for the handwritten types in types.ts

jazzdan: Aww, it makes all of these optional fields? Why?

toast.yml
```diff
@@ -1,12 +1,35 @@
...
+      
+  proto-ts:
+    dependencies:
+      - proto
+    input_paths:
+      - pkg/webview/view.swagger.json
+    output_paths:
+      - web/src/view.d.ts
+    command: |
+      npx @manifoldco/swagger-to-ts pkg/webview/view.swagger.json --camelcase --output web/src/view.d.ts
```
jazzdan: If we want to continue with using yarn everywhere I think we can change this to use [yarn dlx](https://next.yarnpkg.com/cli/dlx):

```suggestion
      yarn dlx @manifoldco/swagger-to-ts pkg/webview/view.swagger.json --camelcase --output web/src/view.d.ts
```

nicks: done

web/src/view.d.ts
nicks: in proto3, fields can only be optional. there's some good discussion here:
https://github.com/protocolbuffers/protobuf/issues/2497
https://capnproto.org/faq.html#how-do-i-make-a-field-required-like-in-protocol-buffers

PR #2428, opened 2019-10-25
build_control: enforce resource_deps in buildorder

landism:
1. Add `HasEverBeenReady()` state to each `RuntimeState` implementation
2. Don't select a manifest for building unless `HasEverBeenReady()` is true for all of its resource deps

note: `HasEverBeenReady()`'s behavior in practice is still a bit confusing due to the flashing green dot bug

### DIFF
--- a/internal/engine/pod.go
+++ b/internal/engine/pod.go
@@ -54,15 +54,15 @@ func handlePodChangeAction(ctx context.Context, state *store.EngineState, action
 		podInfo.BaselineRestarts = podInfo.AllContainerRestarts()
 	}
 
+	if len(podInfo.Containers) == 0 {
+		// not enough info to do anything else
+		return
+	}
+
 	if podInfo.AllContainersReady() {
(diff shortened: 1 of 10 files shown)


internal/store/engine_state.go
```diff
@@ -434,6 +434,10 @@ func (ms *ManifestState) HasPendingChangesBefore(highWaterMark time.Time) (bool,
 	return ok, earliest
 }
 
+func (ms *ManifestState) BlockDependentManifests() bool {
```
nicks: is this dead code?

internal/engine/pod.go
```diff
@@ -53,6 +54,12 @@ func handlePodChangeAction(ctx context.Context, state *store.EngineState, action
 		podInfo.BaselineRestarts = podInfo.AllContainerRestarts()
 	}
 
+	if podInfo.AllContainersReady() {
```
nicks: seems odd that this lives above the Containers == 0 check

Issue #2423, opened 2019-10-24
docker_build( dockerfile= ) treats tilde (~) as a relative path.


rally25rs:
If you do:

```
docker_build('whatever-name', '.', dockerfile='~/Projects/whatever/Dockerfile')
```

it will treat that as a relative path and give the error

```
Error: error reading dockerfile: open /Users/whoever/path_to_wherever_tilt_was_run/~/Projects/wherever/Dockerfile: no such file or
directory
```

It's be nice if a leading `~` worked the same as a leading `/` and treated it as an absolute path.

maiamcc:
Ooh, this is an interesting one! So we're reluctant to support `~` in paths because it opens the door to Tilt configurations that are more dependent on the individual user's filesystem setup and aren't easily encapsulated in a repo. (I.e. if you're using `~` in your Tiltfile, it's likely to lead to a Tiltfile that would be difficult for someone else to run.) Can you talk a little more about your setup/directory structure why you want to be able to use `~` in your Tiltfile?

More practically speaking: you can around `~` support by using [`os.environ`](https://docs.tilt.dev/api_os.html) to get the value of `$HOME`. (Also, we should have clearer messaging around the fact that we don't support `~/...`, I'll toss that in soon.)

rally25rs:
> Can you talk a little more about your setup/directory structure why you want to be able to use ~ in your Tiltfile?

Certainly! We're moving from a dev environment where we primarily used Vagrant to spin up a VM with everything needed for a project, and deployed to AWS EC2, to one of docker images and k8s.

Our typical workflow with Vagrant was that we would have our VMs mount a `~/Projects` directory back to the host. Most people checkout all their github repos under `~/Projects`.

When I'm building up a k8s cluster for development, our Dockerfiles tend to be in those locations still

```
~/Projects
  +- management_api/Dockerfile
  +- management_ui/Dockerfile
  ...etc
```

In my tiltfile, I left a variable at the top that a dev could change to wherever their project dir was
```
# Customize these variables so setup what you want to work on
env = 'dev'
projectsPath = '~/Projects'

...
# build api
projDir = '{}/management_api/'.format(projectsPath)
docker_build('api', projDir, dockerfile='{}/Dockerfile'.format(projDir))

# build ui
projDir = '{}/management_ui/'.format(projectsPath)
docker_build('ui', projDir, dockerfile='{}/Dockerfile'.format(projDir))
```

[...]

PR #2430, opened 2019-10-25
tiltfile: throw error if user attempts to use '~' in paths in tiltifle

maiamcc:
Error messages should be more explicit about the behavior a user hit in #2423

### DIFF
--- a/internal/tiltfile/starkit/path.go
+++ b/internal/tiltfile/starkit/path.go
@@ -1,18 +1,25 @@
 package starkit
 
 import (
+	"fmt"
 	"path/filepath"
+	"strings"
 
 	"go.starlark.net/starlark"
 )
 
 // We want to resolve paths relative to the dir where the currently executing file lives,
 // not relative to the working directory.
-func AbsPath(t *starlark.Thread, path string) string {
-	if filepath.IsAbs(path) {
-		return path
+func AbsPath(t *starlark.Thread, path string) (string, error) {
+	if strings.HasPrefix(path, "~/") || path == "~" {
+		return "", fmt.Errorf(
+			"Tiltfile does not support expansion of '~', please provide an absolute path (you provided: %q)", path)
 	}
-	return filepath.Join(AbsWorkingDir(t), path)
+
+	if filepath.IsAbs(path) {
+		return path, nil
+	}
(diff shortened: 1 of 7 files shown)


internal/tiltfile/starkit/path.go
```diff
@@ -1,18 +1,25 @@
...
 	"go.starlark.net/starlark"
 )
 
 // We want to resolve paths relative to the dir where the currently executing file lives,
 // not relative to the working directory.
-func AbsPath(t *starlark.Thread, path string) string {
+func AbsPath(t *starlark.Thread, path string) (string, error) {
+	if strings.HasPrefix(path, "~/") || path == "~" {
+		return "", fmt.Errorf(
+			"Tiltfile does not support expansion of '~', please provide an absolute path (you provided: %q)", path)
```
landism: I feel like, as a user, I'd have these thoughts:
1. "You already wrote the code to detect it, it's barely any more work to do the expansion for me. Why don't you?"
2. "I don't *have* an absolute path - if I did, I would have used it"
3. "Why can't I provide a relative path?"

I think (3) is "you can" and that part of this error message is just wrong (IIUC this function takes a potentially relative path and turns it into an absolute path)
I don't really have a good answer for (1), and I think we're kind of inconsistent here - if we disallow `~` expansion, we should probably also disallow accessing files outside of the Tiltfile root. I'm not sure how to communicate this. Maybe we just ignore this.

Dan suggested users can work around with with `os.environ`, which we could technically suggest here, but then it's weird to disallow ~ expansion while encouraging `os.environ`, which is no better.

Maybe those were all wasted words because this error is probably still a net improvement.

Also, `~`-expansion *is* supported in `local` and `run`, which this error could be seen as contradicting.

I also think "Tiltfile" as a subject without "The" feels weird, but I don't know if that's just me.

Maybe "Sorry! The Tiltfile does not support ~-expansion in this context (from %q)"?

Sorry for getting nitty.

maiamcc: Nah, not nitty, super useful. DB said some similar stuff irl, esp re: `~` being supported elsewhere, and us being inconsistent about our error messaging. I'll just abandon this PR

PR #2432, opened 2019-10-25
web: add facets tab

landism:
image

This only appears if you've enable the "facets" feature in the Tiltfile.

Caveats:
* The actual facets aren't very interesting at the moment, though at least now you can see the build log separate from the runtime log.
* The k8s_yaml facet is the yaml we read from disk, pre-injection, so it's not really solving any problems (I guess it's useful in the case of helm, kustomize, or m4). Fixing that will be a separate PR.
* The navigation sucks but at least lets us start adding things. We'll see if it gets in the way of dogfooding.


internal/store/manifest_target.go
```diff
@@ -23,3 +28,50 @@ func (t ManifestTarget) Status() model.TargetStatus {
...
+			histories = histories[:20]
+		}
+		for _, br := range histories {
+			sb.WriteString("Build finished:\n")
+			sb.WriteString(fmt.Sprintf("  Reason: %s\n", br.Reason.String()))
+			sb.WriteString("  Result: ")
+			if br.Error == nil {
+				sb.WriteString("Success")
+			} else {
+				sb.WriteString(fmt.Sprintf("%v", br.Error))
```
maiamcc: do we have access to the error such that we can print it here?

web/src/FacetsPane.tsx
```diff
@@ -0,0 +1,52 @@
...
+    if (!this.props.resource.facets) {
+      return new Array<JSX.Element>()
+    }
+    console.log("facets: " + JSON.stringify(this.props.resource.facets))
+    return this.props.resource.facets.map(facet => {
+      return (
+        <li className="FacetsPane-item">
+          <header>
+            <div className="FacetsPane-headerDiv">
+              <h3 className="FacetsPane-headerDiv-header">{facet.name}</h3>
```
maiamcc: no big deal but i'd probably take out this class name and write a style rule for `div.FacetsPane-headerDiv h3` 🤷‍♀

PR #2433, opened 2019-10-25
webview: implement more of the protobuf version of `/api/view`

jazzdan:


jazzdan:
Right now this is mostly copy pasta to keep things separate. In the future we would remove the JSON deserialization.

internal/hud/server/server.go
```diff
@@ -360,9 +360,10 @@ func (s *HeadsUpServer) HandleNewSnapshot(w http.ResponseWriter, req *http.Reque
 }
 
 func (s *HeadsUpServer) GetView(ctx context.Context, req *proto_webview.GetViewRequest) (*proto_webview.View, error) {
-	return &proto_webview.View{
-		FatalError: "AHHH FATAL ERROR",
-	}, nil
+	state := s.store.RLockState()
```
nicks: Are you missing an unlock?

pkg/webview/view.proto
```diff
@@ -4,8 +4,79 @@ package webview;
...
+enum TriggerMode {
+  TRIGGER_MODE_AUTO = 0;
+  TRIGGER_MODE_MANUAL = 1;
+}
+
+message BuildRecord {
+  repeated string edits = 1;
+  string error = 2;
+  repeated string warnings = 3;
+  string start_time = 4;
```
nicks: I might add a note that this is only a string for backwards compatibility. Future time fields should use the built in proto type

internal/hud/server/server.go
jazzdan: I am! This explains the hanging behavior I was seeing!

nicks:
Maybe we should have something in Dev that makes tilt panic if the reducer
gets stuck waiting for more than a second or two

PR #2439, opened 2019-10-28
pod_watch: only emit events from pods whose specs we have deployed since startup

landism:
### Problem

When a user starts Tilt, Tilt will build a new image with a new hash, create new Deployment YAML, and apply it. If there is already a deployment running, this application will edit the existing deployment, leaving the UID unchanged. Thus, when k8s tells Tilt about the pod the deployment had before Tilt even started, Tilt says "hey, I know that deployment's UID!" and treats it as its own.
This has a few concrete minor effects:
1. Tilt prints logs from pods that predate code the user `tilt up`d with
2. Tilt's "ready" state can flash green for a moment, in between the moment Tilt has completed `kubectl apply` and knows the Deployment's UID and the moment k8s has started deploying the new pod and told Tilt about it. (because the old pod matches the deployment and is healthy)
3. Due to (2), the resource_deps startup behavior is pretty confusing and buggy-seeming: https://github.com/windmilleng/tilt.build/pull/318/files#diff-4605e4a7d790d4cf0f1dd26e49ea8455R47

### Solution

Inject a hash of the pod template spec into every pod template spec we deploy. Then, similar to DeployedUIDSet, track those IDs for each manifest and only pay attention to updates for pods whose template specs we've deployed since startup.


internal/k8s/pod_template.go
```diff
@@ -0,0 +1,47 @@
...
+
+import (
+	"crypto"
+	"fmt"
+
+	"github.com/pkg/errors"
+	v1 "k8s.io/api/core/v1"
+)
+
+const TiltPodTemplateHashLabel = "tilt-pod-template-hash"
```
nicks: maybe name the key something like
tilt.dev/pod-template-hash
per the naming conventions
https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/

internal/engine/image_build_and_deployer.go
```diff
@@ -321,25 +323,35 @@ func (ibd *ImageBuildAndDeployer) createEntitiesToDeploy(ctx context.Context,
...
 
 	for _, depID := range depIDs {
 		if !injectedDepIDs[depID] {
-			return nil, fmt.Errorf("Docker image missing from yaml: %s", depID)
+			return nil, nil, fmt.Errorf("Docker image missing from yaml: %s", depID)
 		}
 	}
 
-	return newK8sEntities, nil
+	return newK8sEntities, podTemplateSpecHashes, nil
```
nicks: seems kind of weird to return the podTemplateSpecHashes separately, given that they're already in the k8s entities

internal/engine/k8swatch/pod_watch.go
```diff
@@ -152,6 +152,37 @@ func (w *PodWatcher) removeWatch(toRemove PodWatch) {
...
+			return true
+		}
+	}
+
+	ms := manifest.State
+	if !ms.IsK8s() {
+		return false
+	}
+
+	hash, ok := pod.Labels[k8s.TiltPodTemplateHashLabel]
```
nicks: i'm pretty sure we want to keep pods that don't have a podTemplate hash labels? otherwise we will break people who are using this with resources that don't have a PodTemplate

internal/engine/k8swatch/pod_watch.go
```diff
@@ -152,6 +152,37 @@ func (w *PodWatcher) removeWatch(toRemove PodWatch) {
...
+// At this point, we want a new pod to be created, and we don't care about the old pod. However, since
+// changing a deployment's PodTemplateSpec does not cause the deployment to get a new UID, the deployment
+// applied in (3) has the same UID as the deployment applied in (1), so when we get an event for the
+// undesired pod from (1), we trace it back to the deployment we applied in (3), and emit an event.
+// To deal with this, we inject a hash of the pod template spec into the pod template itself, and only
+// emit events from pods with known pod template spec hashes.
+func isKnownPodTemplateSpecHash(pod *v1.Pod, manifest *store.ManifestTarget) bool {
+	// if it matches ExtraPodSelectors, don't worry about the hash - for CRDs using ExtraPodSelectors
+	// there isn't necessarily a PodTemplateSpec into which we can put a hash
+	for _, selector := range manifest.Manifest.K8sTarget().ExtraPodSelectors {
```
nicks: i think this is not necessary if you fix the bug below

internal/engine/k8swatch/pod_watch.go
```diff
@@ -160,11 +191,22 @@ func (w *PodWatcher) setupNewUIDs(ctx context.Context, st store.RStore, newUIDs
 	w.mu.Lock()
 	defer w.mu.Unlock()
 
+	state := st.RLockState()
```
nicks: seems generally bad to unlock the state multiple times in a dispatch, because you could get inconsistent state between calls. can you copy out the state you need into the taskList?

internal/engine/k8swatch/pod_watch.go
landism: Hm, my mental model was that all pods are either coming from "normal" things, and have specs, or from CRDs, and have ExtraPodLabels. I hadn't put a lot of thought into that, though. And now that you've made me think about it, I guess it's possible that the UID stuff might solve the CRD/ExtraPodLabel issue without having to specify labels?

I was slightly concerned that we'd risk picking up a pod that was deployed by non/old tilt and didn't have a label, but that doesn't really seem worth worrying about.

internal/k8s/pod_template.go
```diff
@@ -0,0 +1,58 @@
...
+	}
+
+	h := crypto.SHA1.New()
+	h.Write(data)
+	return PodTemplateSpecHash(fmt.Sprintf("%x", h.Sum(nil)[:10])), nil
+}
+
+// Iterate through the fields of a k8s entity and add the pod template spec hash on all
+// pod template specs
+func InjectPodTemplateSpecHash(entity K8sEntity) (K8sEntity, error) {
```
maiamcc: nit: `InjectPodTemplateSpecHashes` plural, no?

internal/engine/image_build_and_deployer_test.go
```diff
@@ -536,6 +537,61 @@ func TestDeployInjectsOverrideCommand(t *testing.T) {
...
 
+func (f *ibdFixture) firstPodTemplateSpecHash() k8s.PodTemplateSpecHash {
+	entities, err := k8s.ParseYAMLFromString(f.k8s.Yaml)
+	if err != nil {
+		f.T().Fatal(err)
+	}
+
+	require.Equal(f.T(), 1, len(entities), "expected only one entity. Yaml contained: %s", f.k8s.Yaml)
+
+	d := entities[0].Obj.(*v1.Deployment)
```
maiamcc: maybe `d, ok :=` here and `require.True(t, ok, "expected entity to be a deployment")` -- if it's NOT a deployment for some reason, this is easier to debug than the weird nil pointer errors you'll get otherwise

maiamcc: (but also why are you assuming the first entity is always a deployment?)

internal/engine/image_build_and_deployer_test.go
```diff
@@ -536,6 +537,61 @@ func TestDeployInjectsOverrideCommand(t *testing.T) {
...
+	manifest := NewSanchoDockerBuildManifest(f)
+
+	resultSet, err := f.ibd.BuildAndDeploy(f.ctx, f.st, buildTargets(manifest), store.BuildStateSet{})
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	hash := f.firstPodTemplateSpecHash()
+
+	require.NotEqual(t, hash, "")
```
maiamcc: maybe worth it to have `firstPodTemplateSpecHash` assert that the hash is non-empty? 🤷‍♀

internal/engine/k8swatch/pod_watch.go
```diff
@@ -152,19 +157,46 @@ func (w *PodWatcher) removeWatch(toRemove PodWatch) {
...
 	defer w.mu.Unlock()
 
 	for uid, mn := range newUIDs {
 		w.knownDeployedUIDs[uid] = mn
 
 		pod, ok := w.knownPods[uid]
+
 		if ok {
+			set, ok := podTemplateSpecHashes[mn]
+			if ok && !hasOKPodTemplateSpecHash(pod, set) {
```
maiamcc: maybe worth a descriptive comment e.g. "the pod template hash doesn't correspond to the most recently deployed one--this pod is probably from an old Tilt run" or something.

internal/store/runtime_state.go
```diff
@@ -210,3 +212,19 @@ func (s UIDSet) Add(uids ...types.UID) {
 func (s UIDSet) Contains(uid types.UID) bool {
 	return s[uid]
 }
+
+type PodTemplateSpecHashSet map[k8s.PodTemplateSpecHash]bool
```
maiamcc: bwahaha, look who's using Go sets now!

nicks:
what's the status of this pr? lost track of whether it's tabled, or needs review, or waiting to be merged, something else?

landism:
> what's the status of this pr? lost track of whether it's tabled, or needs review, or waiting to be merged, something else?

When I left yesterday, it was awaiting approval. I failed to ping you for approval yesterday after addressing your comments.
I now need to address Maia's comments and merge, unless you have any concerns.

internal/engine/k8swatch/pod_watch.go
landism: Do you think the existing comment on `hasOKPodTemplateSpecHash` suffices?
Or do you want to add a descriptive comment to each of its call sites?

maiamcc: oh whoops, i missed that. ya that's fine.

PR #2440, opened 2019-10-28
facets: show the applied k8s yaml

landism:


internal/store/build_result.go
```diff
@@ -105,16 +110,30 @@ type K8sBuildResult struct {
...
+			Name:  "applied yaml",
+			Value: r.AppliedEntitiesText,
+		},
+	}
+}
 
 // For kubernetes deploy targets.
-func NewK8sDeployResult(id model.TargetID, uids []types.UID) BuildResult {
+func NewK8sDeployResult(id model.TargetID, uids []types.UID, appliedEntities []k8s.K8sEntity) BuildResult {
+	appliedEntitiesYaml, _ := k8s.SerializeSpecYAML(appliedEntities)
```
maiamcc: would maybe rather panic here than silently ignore the error, but 🤷‍♀

internal/store/engine_state.go
```diff
@@ -221,6 +221,7 @@ type BuildStatus struct {
 	PendingFileChanges map[string]time.Time
 
 	LastSuccessfulResult BuildResult
+	LastResult           BuildResult
```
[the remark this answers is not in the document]
maiamcc: not a bit deal but i might change BuildState.LastResult to BuildState.LastSuccessfulResult as well to mirror this.

internal/store/build_result.go
landism: better yet, we can stick it in the string!

PR #2441, opened 2019-10-28
Add fallback option for kustomize cmd

### LINKED ISSUE #2436: Support kubectl build in kustomize
Since version 1.14 kubectl ships with kustomize. [kustomize - kubectl integration](https://github.com/kubernetes-sigs/kustomize#kubectl-integration) 

It would be awesome if we had the option to use the build in kustomize.

I've already had a glimpse at the source code and it seems to be an easy enhancement:
[relevant part](https://github.com/windmilleng/tilt/blob/8c77047f781faa97320b4a7d124fccbe20a5e655/internal/tiltfile/files.go#L138)

tiri:
Call "kubectl kustomize" if the kustomize binary is not installed.

Open to suggestions for improvement!

resolves #2436


nicks:
lgtm! thanks

PR #2446, opened 2019-10-29
webview: power by proto instead of JSON marshaling

jazzdan:


internal/hud/webview/convert.go
```diff
@@ -414,12 +407,14 @@ func protoPopulateResourceInfoView(mt *store.ManifestTarget, r *proto_webview.Re
 
 func runtimeStatus(res ResourceInfoView) RuntimeStatus {
 	_, isLocal := res.(LocalResourceInfo)
-	if isLocal {
+	_, isLocalPtr := res.(*LocalResourceInfo)
```
nicks: this seems like it's working around a bug that should be fixed upstream. i don't see a legitimate reason to accept both types here

internal/hud/webview/view.go
```diff
@@ -118,17 +123,27 @@ func ToWebViewBuildRecord(br model.BuildRecord) BuildRecord {
 	}
 }
 
+func timeToProto(t time.Time) *timestamp.Timestamp {
+	ts, err := ptypes.TimestampProto(t)
+	if err != nil {
+		return nil
```
nicks: this seems dangerous. i would be more comfortable with this if it was a `panic()`, tho it really should return an error

jazzdan: @nicks the proto know show to handle `nil` as the empty value for a timestamp, so I think this is OK.

https://github.com/golang/protobuf/blob/master/ptypes/timestamp.go#L95

nicks: oh, my concern isn't that we're returning nil, but that we're quietly dropping validation errors

jazzdan: And in fact empty values is what the web UI expects here.

jazzdan: Ah, okay. I convert all of these things to return error then.

jazzdan:
cc @landism converting the Facets to protobuf

PR #2451, opened 2019-10-29
web: power snapshots with protobufs

jazzdan:


web/src/FatalErrorModal.tsx
```diff
@@ -4,7 +4,7 @@ import { ShowFatalErrorModal } from "./types"
 import "./FatalErrorModal.scss"
 
 type props = {
-  error: string | null
+  error: string | null | undefined
```
jazzdan: swagger generates typescript files with undefined instead of null for some reason?

pkg/webview/view.proto
```diff
@@ -108,10 +108,35 @@ message View {
 
 message GetViewRequest {}
 
+message SnapshotHighlight {
+  string beginning_log_id = 1;
```
nicks: i think this needs to be beginning_logID to preserve the old casing? surprised the test didn't catch this

web/src/HUD.tsx
```diff
@@ -143,7 +143,7 @@ class HUD extends Component<HudProps, HudState> {
     return this.pathBuilder.path(relPath)
   }
 
-  snapshotFromState(state: HudState): Snapshot {
+  snapshotFromState(state: HudState): OpenAPI2.webviewSnapshot {
```
nicks: is there a way we can change the namespace?

pkg/webview/view.proto
jazzdan: Ah well the test doesn't have a highlight so I don't think it would. I'll add a highlight to the test and fix this.

web/src/HUD.tsx
nicks: ya i would just make it `pb` or `data` or smth

PR #2453, opened 2019-10-30
tiltfile: add an analytics_settings for enabling/disabling analytics from the tiltfile

nicks:
Hello @landism, @maiamcc,

Please review the following commits I made in branch nicks/ch3805/analytics:

df5fb930d0e209e474ae94d39412e0c3167fa1e3 (2019-10-29 20:27:13 -0400)
tiltfile: add an analytics_settings for enabling/disabling analytics from the tiltfile

### DIFF
--- a/internal/hud/webview/analytics_nudge.go
+++ b/internal/hud/webview/analytics_nudge.go
@@ -7,7 +7,7 @@ import (
 )
 
 func NeedsNudge(st store.EngineState) bool {
-	if st.AnalyticsOpt != analytics.OptDefault {
+	if st.AnalyticsEffectiveOpt() != analytics.OptDefault {
 		// already opted in/out
 		return false
 	}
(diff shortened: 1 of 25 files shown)


internal/hud/webview/analytics_nudge.go
```diff
@@ -7,7 +7,7 @@ import (
 )
 
 func NeedsNudge(st store.EngineState) bool {
-	if st.AnalyticsOpt != analytics.OptDefault {
+	if st.AnalyticsUserOpt != analytics.OptDefault {
```
landism: probably doesn't matter much, but do you think it'd make sense to skip the nudge if the team is opted in?

nicks: ooh, that's a good point. i could go either way on this, but ya you're right that skipping the nudge should make more sense

PR #2458, opened 2019-10-30
integration test checks if we can do live updates after a crash rebuild (currently fails)

maiamcc:
PR to run tests

### DIFF
--- a/integration/onewatch_test.go
+++ b/integration/onewatch_test.go
@@ -7,7 +7,7 @@ import (
 	"testing"
 	"time"
 
-	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
 func TestOneWatch(t *testing.T) {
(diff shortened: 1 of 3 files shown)


maiamcc:
Still have a unit test to write, but, does this look right?

nicks:
no, this doesn't seem like the right change to me.

why is unioning with the old builds the right move?

maiamcc:
Ah whoops, I forgot you didn't have context here.

We use that set of pod template spec hashes to check that incoming pod events are actually from pods that Tilt cares about. We do this filtering in the watcher code, so pod events that don't match a hash in this set don't even make it into `handlePodChangeAction`.

If we OVERWRITE the set of `PodTemplateSpecHashes` that we know about on every deploy, events from old pods (that Tilt DID deploy, but just isn't currently managing) won't ever hit `handlePodChangeAction` -- most importantly, we'll never catch the Pod Deleted actions for these old pods, which means we won't take them off the manifest, which means we'll always think we have too many pods to do a Live Update. Make sense?

nicks:
Hmmmm...but why even wait until we get the pod deleted actions? we know the
old pods are old, so we can remove them right now?

nicks:
alternatively, maybe the UIDs and PodTemplateSpecHashes should be part of some sort of a BuildHistory-like struct, so we can distinguish between what's old and what's current?

The fundamental mistake here (as I see it, given how i've seen this play out with other build systems), is that you're wiping out important context about the state (the old code threw out the old hashes, and the new code throws out which build they came from)

maiamcc:
After discussion IRL, this PR doesn't accomplish what we want it to -- but @landism, you should probably add this updated integration test when you revert `PodTemplateSpecHash` stuff to make sure that LiveUpdate works properly after a crash rebuild.

PR #2462, opened 2019-10-30
engine: support force update of any resource (backend only) [ch3767]

maiamcc:
here's the BACKEND code for allowing a triggered update to any
resource regardless of
a. whether resource is trigger mode manual/auto, or
b. whether resource has pending changes or not

UI changes to come in a separate PR. the state of the world
in this PR is that only manual resources may be triggered,
but they can be triggered regardless of clean/dirty state.
If the resource is clean, we do a full rebuild.
The resource entry in the sidebar still changes to indicate
when the resource is dirty.


internal/hud/server/server_test.go
```diff
@@ -325,31 +325,38 @@ func TestMaybeSendToTriggerQueue(t *testing.T) {
...
-
-func TestMaybeSendToTriggerQueue_notManualManifest(t *testing.T) {
+func TestSendToTriggerQueue_automaticManifest(t *testing.T) {
 	f := newTestFixture(t)
 
 	mt := store.ManifestTarget{
 		Manifest: model.Manifest{
 			Name:        "foobar",
-			TriggerMode: model.TriggerModeAuto,
+			TriggerMode: model.TriggerModeAuto, // can now enqueue manifests regardless of TriggerMode
```
dbentley: Comments with "non" get outdated quickly. Not sure if this is relevant. If anything, "also test auto" or some such?

maiamcc: fair, i'll just remove it

PR #2463, opened 2019-10-30
tiltfile: add ignore to local_resource

landism:


internal/tiltfile/tiltfile_state.go
```diff
@@ -1410,7 +1410,10 @@ func (s *tiltfileState) translateLocal() ([]model.Manifest, error) {
 		}
 
 		paths := append(r.deps, r.workdir)
-		lt := model.NewLocalTarget(model.TargetName(r.name), r.cmd, r.workdir, r.deps).WithRepos(reposForPaths(paths))
+		ignores := s.dockerignoresFromPathsAndContextFilters(paths, r.ignores, r.onlys)
```
nicks: i don't think this is right?

this reads off all .dockerignores in `paths`, which is not what you want

also, ignores should be evaluated relative to workdir, not relative to deps dirs

internal/tiltfile/tiltfile_test.go
```diff
@@ -3854,6 +3854,39 @@ local_resource("toplvl-local", "echo hello world", ["foo/baz", "foo/a.txt"])
 	}, ltTop.LocalRepos())
 }
 
+func TestLocalResourceIgnoreOnly(t *testing.T) {
+	f := newFixture(t)
+	defer f.TearDown()
+
+	f.file("Tiltfile", `
+local_resource("test", "echo hi", ["foo"], ignore="**/*.a", only="**/bar.*")
```
nicks: i don't understand what the different is between `deps` and `only`? they seem exactly the same to me

internal/tiltfile/tiltfile_state.go
nicks: can you add a test that
1) local_resource does NOT respect .dockerignore, and
2) your ignore pattern is evaluated relative to the correct directory

internal/tiltfile/tiltfile_test.go
landism: maybe the new test array makes things clearer?

not sure if this is what you're asking, but the idea is:
`foo.a` matches ignore but not only
`bar.a` matches ignore *and* only
`bar.b` matches only but not ignore

nicks: also for future reference: ** globs in `only` parameters won't work. they're not implemented correctly in the underlying dockerignore engine. 

This is documented here: https://docs.tilt.dev/api.html#api.docker_build

and our docker_build command will error if you try to use them this way
https://github.com/windmilleng/tilt/blob/master/internal/tiltfile/tiltfile_test.go#L3370

internal/tiltfile/tiltfile_test.go
```diff
@@ -3854,6 +3854,44 @@ local_resource("toplvl-local", "echo hello world", ["foo/baz", "foo/a.txt"])
...
 }
 
+func TestLocalResourceIgnoreOnly(t *testing.T) {
+	f := newFixture(t)
+	defer f.TearDown()
+
+	f.file(".dockerignore", "**/**.c")
+	f.file("Tiltfile", "include('proj/Tiltfile')")
+	f.file("proj/Tiltfile", `
+local_resource("test", "echo hi", ["foo"], ignore=["**/*.a", "foo/bar.d"], only=["**/bar.*", "foo/baz"])
```
nicks: can you make sure this errors if there are globs in the `only`, like `docker_build` does?

landism: hm, I think at that point we might as well just not have 'only'?

nicks: ya, exactly. 

the `only` parameter was added to `docker_build`, because in `docker_build`, there's a meaningful difference between "this is the root of my docker context", and "these are the files included in my docker context". The first one is what dockerignore files are relative to.

what i'm really worried about is a situation where, by accident, "deps" and "only" become subtly different implementations of the same thing, with subtly different bugs and gotchas.

internal/tiltfile/tiltfile_test.go
landism: > also for future reference: ** globs in only parameters won't work. they're not implemented correctly in the underlying dockerignore engine.

Actually, thinking about this last night. The unit test is passing as-is, which, uh, seems like good evidence to the contrary?

This made me try harder to remember the details of this bug, and IIRC underlying dockerignore engine is perfectly fine, and there's an optimization in the *tarring code* that inappropriately uses dockerignore to skip directories. Since we're not tarring for local resource, it should be fine.

nicks: talked about this offline, but adding this comment for posterity

the watchmanager won't watch things correctly, and this test doesn't exercise the watch code at all.

all build systems have this problem, it's not unique to us. 

here's the discussion in Docker:
https://github.com/moby/moby/issues/30018

here's the discussion in Bazel:
https://docs.bazel.build/versions/master/be/functions.html#glob

The general problem is that once you start to have complicated sequences of overlapping include globs and ignore globs, they become impossible to implement without stat-ing the whole file system (and causing performance problems that are hard to diagnose), so all build systems put pretty tight restrictions on how you can use them.

nicks:
what's the status of this pr? was hoping to get this into the release today

PR #2469, opened 2019-10-31
web: UI for trigger button that allows force update [ch3767]

maiamcc:
would esp. appreciate JS/design review from @hyu, i bet there's something
much nicer that can be done to style the `isDirty` trigger button 😅

(Also the lion's share of the JS in this PR was @hyu, I just did a little styling/logic and tests)


maiamcc:
Screen Shot 2019-10-31 at 12 31 20 PM
pending/building resources don't show the trigger button

web/src/SidebarTriggerButton.scss
```diff
@@ -14,19 +12,68 @@
...
+$trigger-button-color: $color-gray-light;
+$trigger-button-hover-color: $color-white;
+$trigger-button-active-color: $color-green;
+$trigger-button-selected-color: $color-gray-dark;
+
 .SidebarTriggerButton.isReady {
-  background-image: url(assets/svg/trigger-button.svg);
+  background-color: $trigger-button-color;
+  -webkit-mask: url(assets/svg/trigger-button.svg) no-repeat 50% 50%;
+  mask: url(assets/svg/trigger-button.svg) no-repeat 50% 50%;
```
hyu: Ooh, interesting solution for changing the SVG background image color! I love the concision, and I see that `mask` is now more browser compatible than it used to be. ("Back in my day..." 🧓🏼)

landism:
1. What motivates the choice of red for the rebuild icon in the "manual build pending changes" case? I feel like red is communicating something is broken.
2. I think we need tooltips for the rebuild buttons, both so that users can tell what these shiny new buttons are going to do without having to actually do it, and so that we can communicate the feature's name to users so they can find docs.

maiamcc:
> What motivates the choice of red for the rebuild icon in the "manual build pending changes" case? I feel like red is communicating something is broken.

Yep, fair. This was 100% a placeholder design that I hoped @hyu would have better ideas for later 😅 but they're swamped so I'm gonna do something better now!

> 2. I think we need tooltips for the rebuild buttons, both so that users can tell what these shiny new buttons are going to do without having to actually do it, and so that we can communicate the feature's name to users so they can find docs.

yep sounds right. i'll do that in a future PR

maiamcc:
force-rebuild-ui

@hyu @landism better?

maiamcc:
BTW @hyu i'm waiting on your green light before merging this!

hyu:
@maiamcc — This is great! Thanks for pulling this together so nicely.

Nits for future PRs:

- I saw that you’re using the old manual button, masked green, for the Dirty state. I thought it would work better if it was using the same minimal icon style, but blue. Maybe a little larger in size if it needs to be more visible. (I wanted to reserve green for the “ok” state. Because we could have a situation where the last build failed and thus the resource is in an error or warn state, but then if the rebuild button is green I think that could be possibly confusing. In the interface, blue is the “neutral” accent color that says “pay attention”, so I think it’s a better choice.)

- I’m biased against using the older icon because I think it’s too different from the new one we’re using, and almost suggests that pressing it would have some other functionality. We’re trying to convey that there’s something that requires your attention, but the button still does the same thing. Also, I imagine that going in and out of a Dirty state will happen a lot, so I think the change can be more subtle.

- It bugs me that the icon disappears when you click it. I kind of think it should become less visible and also rotate when the resource is queued, whether it's building or not. Maybe it can be slowly rotating when queued, and rotate faster when actively building.

PR #2475, opened 2019-11-01
webview: Alert timestamp is string, temporarily

jazzdan:


toast.yml
```diff
@@ -5,7 +5,6 @@ tasks:
       - pkg/webview/view.proto
     output_paths:
       - pkg/webview/view.pb.go
-      - pkg/webview/view.pb.gw.go
```
nicks: ???

jazzdan: We don't need this anymore, we don't use the grpc services in Tilt.

jazzdan: I guess I should delete the generated code too, then. Will do.

PR #2486, opened 2019-11-04
integration: test that we can do a live update after a crash rebuild

maiamcc:
bug fix from #2458 no longer applies, but pull in the integration test
so we can be sure that future commits don't break this behavior

### DIFF
--- a/integration/Tiltfile
+++ b/integration/Tiltfile
@@ -1,5 +1,7 @@
 # -*- mode: Python -*-
 
+docker_prune_settings(disable=True)
+
 # HACK: load namespaces on `tilt up` but not on `tilt down`
 load_namespace = not os.environ.get('SKIP_NAMESPACE', '')
 if load_namespace:
--- a/integration/Tiltfile
+++ b/integration/Tiltfile
@@ -1,7 +1,5 @@
 # -*- mode: Python -*-
 
-docker_prune_settings(disable=True)
-
 # HACK: load namespaces on `tilt up` but not on `tilt down`
 load_namespace = not os.environ.get('SKIP_NAMESPACE', '')
 if load_namespace:
--- a/integration/live_update_after_crash_rebuild_test.go
+++ b/integration/live_update_after_crash_rebuild_test.go
@@ -1,0 +1,76 @@
+//+build integration
+
+package integration
+
+import (
+	"context"
+	"fmt"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestLiveUpdateAfterCrashRebuild(t *testing.T) {
+	f := newK8sFixture(t, "live_update_after_crash_rebuild")
+	defer f.TearDown()
+	f.SetRestrictedCredentials()
+
+	f.TiltWatch()
+
+	ctx, cancel := context.WithTimeout(f.ctx, time.Minute)
+	defer cancel()
+	fmt.Println("> Waiting for pods from initial build")
+	oneUpPods := f.WaitForAllPodsReady(ctx, "app=live-update-after-crash-rebuild")
+
+	ctx, cancel = context.WithTimeout(f.ctx, time.Minute)
+	defer cancel()
+	f.CurlUntil(ctx, "http://localhost:31234", "🍄 One-Up! 🍄")
+
+	// Live update
+	f.ReplaceContents("compile.sh", "One-Up", "Two-Up")
+
+	ctx, cancel = context.WithTimeout(f.ctx, time.Minute)
+	defer cancel()
+	fmt.Println("> Perform a live update")
+	f.CurlUntil(ctx, "http://localhost:31234", "🍄 Two-Up! 🍄")
+
+	// Check that the pods were changed in place, and that we didn't create new ones
+	twoUpPods := f.WaitForAllPodsReady(ctx, "app=live-update-after-crash-rebuild")
+	require.Equal(t, oneUpPods, twoUpPods)
+
+	if len(twoUpPods) != 1 {
+		t.Fatalf("Expected one pod, actual: %v", twoUpPods)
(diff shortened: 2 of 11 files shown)


integration/Tiltfile
```diff
@@ -1,5 +1,7 @@
 # -*- mode: Python -*-
 
+docker_prune_settings(disable=True)
```
nicks: why? can you add a comment?

I would argue that we really do want to run this in the integration tests, because that's how people run tilt normally

integration/onedc_test.go
```diff
@@ -26,7 +26,7 @@ func TestOneDockerCompose(t *testing.T) {
 
 	f.CurlUntil(ctx, "onedc", "localhost:8000", "🍄 One-Up! 🍄")
 
-	f.ReplaceContents("main.go", "One-Up", "Two-Up")
```
nicks: ?

why did this change? this doesn't seem right

integration/Tiltfile
maiamcc: It's just potentially extra log output to wade through--but that's a fair point, I'll put it back

integration/onedc_test.go
maiamcc: ...oh weird, I think this was a weird merge thing. good catch.

maiamcc:
ready for review!

integration/live_update_after_crash_rebuild_test.go
```diff
@@ -0,0 +1,76 @@
...
+	ctx, cancel = context.WithTimeout(f.ctx, time.Minute)
+	defer cancel()
+	f.CurlUntil(ctx, "http://localhost:31234", "🍄 One-Up! 🍄")
+
+	// Live update
+	f.ReplaceContents("compile.sh", "One-Up", "Two-Up")
+
+	ctx, cancel = context.WithTimeout(f.ctx, time.Minute)
+	defer cancel()
+	fmt.Println("> Perform a live update")
```
landism: nit: I feel like this should be above the `ReplaceContents`, so that it's logged before the live update starts, rather than potentially after it's already done

integration/live_update_after_crash_rebuild_test.go
```diff
@@ -0,0 +1,76 @@
...
+	// Delete the pod and make sure it got replaced with something that prints the
+	// same thing (crash rebuild).
+	fmt.Println("> Kill pod, wait for crash rebuild")
+	f.runCommandSilently("kubectl", "delete", "pod", twoUpPods[0], namespaceFlag)
+
+	ctx, cancel = context.WithTimeout(f.ctx, time.Minute)
+	defer cancel()
+	f.CurlUntil(ctx, "http://localhost:31234", "🍄 Two-Up! 🍄")
+
+	// Unfortunately "WaitForAllPodsReady" isn't that accurate and can pull in terminating pods
```
landism: This surprises me. If it pulls in a "terminating" pod, wouldn't that be not ready, and thus either way, it won't return until the terminating pod is gone and the new pod is ready?

Or is that "terminating" with a lower-case t, as in "maybe we killed it but it's still 'Running' for an instant, so we might return from `WaitForAllPodsReady` because the old pod is still running and the new pod doesn't exist yet"? (at which point we should probably say "the pod we just deleted" or something rather than using a term that is also the name of a state returned by kubectl)

In which case, it seems like we should have the sleep before the `CurlUntil` above, because that state would also mean we're getting a Curl back from the old pod, but the goal is to get it from the new pod?

This is maybe all nitty since this is probably copypasta'd from another test that works well enough, and maybe really more of a separate PR issue - just trying to 1) figure out what this is supposed to be doing and 2) figure out if we can make the code more directly match the intent

integration/live_update_after_crash_rebuild_test.go
```diff
@@ -0,0 +1,76 @@
...
+	// too. Sleep here to increase the chance that pods are in the right state when we check.
+	fmt.Println("> (Waiting for dead pods to get into 'terminating' state)")
+	time.Sleep(2 * time.Second)
+
+	newTwoUpPods := f.WaitForAllPodsReady(ctx, "app=live-update-after-crash-rebuild")
+	require.NotEqual(t, twoUpPods, newTwoUpPods)
```
landism: nit: this might be a bit clearer as `postLiveUpdatePods` and `postCrashRebuildPods` or so

integration/live_update_after_crash_rebuild_test.go
```diff
@@ -0,0 +1,76 @@
...
+
+	// Another live update! Make sure that, after the crash rebuild, we're able to run more
+	// live updates (i.e. that we have one and only one pod associated w/ the manifest)
+	fmt.Println("> Perform another live update")
+	f.ReplaceContents("compile.sh", "Two-Up", "Three-Up")
+
+	ctx, cancel = context.WithTimeout(f.ctx, time.Minute)
```
landism: not necessarily a thing for this PR but a `CurlUntilTimeout` would be nice to reduce these repeated three lines to one

integration/live_update_after_crash_rebuild_test.go
maiamcc: yeah i don't get what's going on here in any detail, but here's my report: https://windmill-eng.slack.com/archives/C9SA451CJ/p1572559413240100

maiamcc: it's the second case you mentioned: we killed it but it's still running for an instant. i'll poke around at reordering the sleep/the other stuff and update the comment to make more sense.

integration/live_update_after_crash_rebuild_test.go
maiamcc: yeahhhh that'd be nice but i don't wanna do it nowwww 😬

PR #2491, opened 2019-11-04
web: web UI reflects when builds are queued

maiamcc:
Hello @nicks, @hyu,

1. idk if the CSS is quite right, in particular, moving the animations to a separate file because we're now using the same animation in two places
2. made icon blue instead of green, can go search for a smaller icon later, but @hyu i think you have a much stronger idea here than I do.
3. i'm struggling with the little jitter of the sidebar that you can see in this gif if you look at the scroll bar on the bottom right of the window -- help?

4. is disabling click via CSS the right move? I should maybe do that in code...?


web/src/SidebarTriggerButton.tsx
```diff
@@ -34,23 +35,28 @@ export default class SidebarTriggerButton extends PureComponent<
...
       !props.isBuilding &&
       props.hasBuilt &&
       (!props.hasPendingChanges || isManualTriggerMode)
     let isDirty = props.hasPendingChanges && isManualTriggerMode
 
     return (
       <button
-        onClick={() => triggerUpdate(props.resourceName)}
+        onClick={(e: React.MouseEvent) => {
+          e.preventDefault() // don't select the resource in the sidebar
```
nicks: if you want to prevent the event from bubbling up to the resource handler, use e.stopPropagation()

preventDefault seems wrong here - it's for preventing the default browser ui interactions from kicking in, which is not what we want

web/src/SidebarTriggerButton.scss
```diff
@@ -42,6 +43,15 @@ $trigger-button-active-color: $color-green;
   mask-size: 40%;
 }
 
+
+.SidebarTriggerButton.isQueued {
+  background-color: $trigger-button-color;
+  -webkit-mask: url(assets/svg/trigger-button.svg) no-repeat 50% 50%;
+  mask: url(assets/svg/trigger-button.svg) no-repeat 50% 50%;
+  animation: spin 1s linear infinite;
+  pointer-events: none; // disable click
```
nicks: using pointer-events: none here is pretty bad for a bunch of a11y and UX reasons. you can still press the button with the keyboard for example. and it prevents tooltips from working.

instead, please mark the button disabled, and add a tooltip that explains why it's disabled.

web/src/SidebarTriggerButton.tsx
nicks: here's a pretty good explanation
https://medium.com/@jacobwarduk/how-to-correctly-use-preventdefault-stoppropagation-or-return-false-on-events-6c4e3f31aedb

maiamcc: also already tried that, I think there's some wonky behavior with the react `Link` component? Or else I'm missing something dumb, but here are my attempts to structure the DOM appropriately, and the weird shit they each do:
Screen Shot 2019-11-05 at 2 20 55 PM
Screen Shot 2019-11-05 at 2 20 20 PM

nicks: sent https://github.com/windmilleng/tilt/pull/2497 to move the button out of the link.

also created a storybook so that you can iterate on this outside of tilt

hyu:
Thanks for these fixes, @maiamcc! So much better.

For posterity:
- My quick hypothesis on the jitter you noted is that the rotating square button (with SVG background) is changing the dimensions of the sidebar! When it's 45 degrees, maybe it pushes the whole Resource item wider. To solve…I wonder if it would be possible to only animate the background svg, rather than the entire button element?

- I think your tooltip can probably be more succinct, since I don't expect long strings for that UI element. But that wordsmithing is actually beyond me at the moment. Good for now!

- Pending and building — I think it would work better to have slow rotation for pending, and faster rotation for actually building. It’s jarring to me that there’s a state where it randomly disappears.

PR #2493, opened 2019-11-05
web: introduce HUDGrid

nicks:
Hello @hyu,

Please review the following commits I made in branch nicks/panels:

aaf34c0bfba132400cb8e20c849ac3686f55a1bb (2019-11-04 22:24:18 -0500)
web: introduce HUDGrid
Now, HUDGrid manages the layout and overlap between components, rather than
every component having to know about the other components in the grid.

### DIFF
--- a/web/src/HUDGrid.tsx
+++ b/web/src/HUDGrid.tsx
@@ -1,0 +1,105 @@
+import React from "react"
+import styled from "styled-components"
+import { Height, Width, ZIndex } from "./constants"
+
+// The HUD UI looks like this:
+//
+// ----------------------------------
+// | top bar                    | S |
+// -----------------------------| i |
+// | resource info bar          | d |
+// -----------------------------| e |
+// | main pane                  | b |
+// |                            | a |
+// |                            | r |
+// |                            |   |
+// |                            |   |
+// |                            |   |
+// ----------------------------------
+// | status bar                     |
+// ----------------------------------
+//
+// Our DOM architecture needs to satisfy two competing constraints:
+//
+// 1) "window scroll" needs to scroll the main pane.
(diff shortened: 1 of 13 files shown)


nicks:
Today I learned: creating styled-components in the render() method led to major performance problems. They need to live in the global scope. (This isn't so obvious in TFT, which only renders once, but is much more obvious here)

nicks:
(I did some CSS descender tricks to resolve this problem, but it's something we should keep an eye out for on future code reviews)

web/src/HUDGrid.tsx
```diff
@@ -0,0 +1,105 @@
+import React from "react"
+import styled from "styled-components"
+import { Height, Width, ZIndex } from "./constants"
+
+// The HUD UI looks like this:
+//
+// ----------------------------------
```
hyu: YES! ASCII art layout explanations FTW!

web/src/HUDGrid.tsx
```diff
@@ -0,0 +1,105 @@
...
+`
+
+let MainFrame = styled.div`
+  display: flex;
+  align-items: stretch;
+  width: 100%;
+  padding-bottom: ${Height.statusbar}px;
+  transition: margin-right 300ms ease;
+  padding-top: ${Height.topBar}px;
+  margin-right: ${Width.sidebar}px;
```
hyu: Is it problematic to have width 100% AND a right- (or left-) margin? Could it ever create a hard-to-interpret situation for some browsers?

(I honestly am not sure. This is a thing I've done a lot but recently started to suspect)

nicks: ya, i think making it padding-right is probably more correct here

nicks:
ya, i agree we need to do a naming pass on the different components.

PR #2502, opened 2019-11-05
web: don't show trigger button next to Tiltfile, we currently have no way to force update it

maiamcc:
Hello @hyu, @nicks,

Please review the following commits I made in branch maiamcc/cant-trigger-tiltfile:

4671a148c6a3f88e66d295bb5cd9783a5cc03f7c (2019-11-05 17:59:19 -0500)
web: don't show trigger button next to Tiltfile, we currently have no way to force update it


web/src/Sidebar.test.tsx
```diff
@@ -246,4 +246,31 @@ describe("sidebar", () => {
 
     expect(tree).toMatchSnapshot()
   })
+
+  it("disables trigger button for Tiltfile", () => {
```
nicks: can you elaborate on the value of this test? you don't seem to be asserting anything

web/src/SidebarTriggerButton.tsx
```diff
@@ -46,6 +46,19 @@ export default class SidebarTriggerButton extends PureComponent<
 > {
   render() {
     let props = this.props
+    if (props.resourceName == "(Tiltfile)") {
```
nicks: the protobuf has an isTiltfile property, can you use that instead?

maiamcc: ooh that's much better, will try that, ty

web/src/Sidebar.test.tsx
maiamcc: I mean I'm matching against a snapshot in which the trigger button is `disabled` and has the appropriate tooltip. Is your question about what I am/am not expecting to be in the snapshot, or about the ethos of snapshot tests more broadly?

nicks: this test is a pretty good model:
https://github.com/windmilleng/tilt/blob/master/web/src/SidebarTriggerButton.test.tsx#L38

PR #2519, opened 2019-11-07
engine: restore template spec hashes, use them to decide whether to add a pod to a manifest (but don't use them to filter out pod events entirely -- we still need to know if e.g. an existing pod was deleted, even if it no longer matches the currently deployed pod template spec hash)"

maiamcc:
unit tests for this specific behavior to follow, but nick, is this more
what you had in mind?


internal/engine/pod.go
```diff
@@ -103,17 +103,18 @@ func matchPodChangeToManifest(state *store.EngineState, action k8swatch.PodChang
...
-func trackPod(ms *store.ManifestState, action k8swatch.PodChangeAction) (*store.Pod, bool) {
+func maybeTrackPod(ms *store.ManifestState, action k8swatch.PodChangeAction) (*store.Pod, bool) {
 	pod := action.Pod
 	podID := k8s.PodIDFromPod(pod)
 	startedAt := pod.CreationTimestamp.Time
 	status := k8swatch.PodStatusToString(*pod)
 	ns := k8s.NamespaceFromPod(pod)
 	hasSynclet := sidecar.PodSpecContainsSynclet(pod.Spec)
 	runtime := ms.GetOrCreateK8sRuntimeState()
+	currentDeploy := runtime.HasOKPodTemplateSpecHash(pod)
```
landism: nit: I like putting "is", "has", etc in bool names, e.g. `isCurrentDeploy` (on first skim I thought you were putting the hash in a var called "deploy", which would be weird!)

internal/engine/pod.go
```diff
@@ -122,16 +123,15 @@ func trackPod(ms *store.ManifestState, action k8swatch.PodChangeAction) (*store.
...
-		pod := &store.Pod{
+		podInfo := &store.Pod{
 			PodID:      podID,
 			StartedAt:  startedAt,
 			Status:     status,
 			Namespace:  ns,
 			HasSynclet: hasSynclet,
 		}
-		runtime.Pods[podID] = pod
-		ms.RuntimeState = runtime
```
landism: I think this would be clearer if you keep this line in this `if` block and move it immediately after the `runtime.Pods`. The confusion we had the other day about why we're reassigning back to ms.RuntimeState seems like it'd be exacerbated by moving it further away, into `maybeAttachPodToManifest`

internal/engine/image_build_and_deployer_test.go
```diff
@@ -536,6 +537,64 @@ func TestDeployInjectsOverrideCommand(t *testing.T) {
...
+	f.docker.BuildOutput = docker.ExampleBuildOutput2
+
+	_, err = f.ibd.BuildAndDeploy(f.ctx, f.st, buildTargets(manifest), store.BuildStateSet{})
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	hash2 := f.firstPodTemplateSpecHash()
+
+	require.NotEqual(t, hash1, hash2)
```
nicks: is `require` the new `assert` now?

internal/engine/pod.go
```diff
@@ -122,16 +123,15 @@ func trackPod(ms *store.ManifestState, action k8swatch.PodChangeAction) (*store.
...
-		return pod, true
+		maybeAttachPodToManifest(ms, runtime, podInfo, currentDeploy)
```
nicks: the `maybe` here doesn't seem correct. We've already started mutating the state of the runtime a couple lines above based on this new pod info. Why would we want to start mutating the state, then abandon mid-stream?

in general, it would be better if the currentDeploy check was better unified with the other `if` checks.

internal/engine/upper_test.go
```diff
@@ -1661,6 +1675,95 @@ func TestPodContainerStatus(t *testing.T) {
 	f.assertAllBuildsConsumed()
 }
 
+func TestPodAddedToStateOrNotByTemplateHash(t *testing.T) {
```
maiamcc: @landism or @nicks could you ptal at this test?

internal/engine/upper_test.go
```diff
@@ -1661,6 +1675,95 @@ func TestPodContainerStatus(t *testing.T) {
...
+			ms.RuntimeState = runtime
+			f.store.UnlockMutableState()
+
+			// Dispatch pod event
+			podHash := deployedHash
+			if !test.ptshMatch {
+				podHash = nonMatchingHash
+			}
+			pod := podbuilder.New(f.T(), manifest).
+				WithPodID(podID.String()).
```
landism: it seems weird to always use the same podID here as for the preexisting pod - do you think it's worth testing that dimension?

(I could imagine an implementation of the pod matching logic where we have a good old pod and then a bad new pod comes in and incorrectly wipes out the good old pod, when really the good old pod should be kept, but I think regardless of our behavior in that situation, the way the app works, that case will never arise?)

maiamcc: I mean the case I'm testing is where we already have information about the pod, and we get in a new event about the same pod. good point about the additional dimension tho, let me see how much work it'd be to add that.

PR #2522, opened 2019-11-08
ability to disable upgrade nag #2520

### LINKED ISSUE #2520: Make Upgrade Notification Configurable
It should at least be able to be disabled. In the future we might want to offer other configurations like:

* Min/max versions
* configurable download URL

jazzdan:
This fixes #2520


internal/store/engine_state.go
```diff
@@ -65,7 +65,8 @@ type EngineState struct {
 	TiltfileState ManifestState
 
 	// from GitHub
-	LatestTiltBuild model.TiltBuild
+	LatestTiltBuild     model.TiltBuild
+	CheckVersionUpdates bool
```
nicks: can we put the whole VersionSettings in the engine state, for consistency?

want to move towards a world where we have one Tiltfile result data Model on the engine state, rather than the current world where there are 3 or 4 artisanal data types in-between that are hard to reason about

internal/store/engine_state.go
```diff
@@ -65,7 +65,8 @@ type EngineState struct {
 	TiltfileState ManifestState
 
 	// from GitHub
-	LatestTiltBuild model.TiltBuild
+	LatestTiltBuild     model.TiltBuild
```
maiamcc: nit: the above comment no longer applies to both of these lines
```suggestion
	LatestTiltBuild     model.TiltBuild // from GitHub
```

internal/tiltfile/version/version.go
```diff
@@ -0,0 +1,62 @@
...
+	return Settings{
+		CheckUpdates: true,
+	}
+}
+
+func (Extension) OnStart(env *starkit.Environment) error {
+	return env.AddBuiltin("version_settings", setUpgradeSettings)
+}
+
+func setUpgradeSettings(thread *starlark.Thread, fn *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
```
maiamcc: did you mean "update settings" (rather than "upgrade")?

internal/store/engine_state.go
jazzdan: @nicks would this mean that the struct is duplicated between `internal/tiltfile/analytics` and `internal/model` ?

jazzdan: Or should `internal/tiltfile/analytics` point to `internal/model`?

web/src/Statusbar.test.tsx
```diff
@@ -185,6 +192,32 @@ describe("StatusBar", () => {
...
+            checkVersion={true}
+          />
+        </MemoryRouter>
+      )
+      .toJSON()
+
+    expect(tree).toMatchSnapshot()
+  })
+
+  it("does not render an upgrade badge when the version is out of date if version check is disabled", () => {
```
maiamcc: my understanding is that we're trying to move away from snapshot tests and towards assertions? would prefer this as:
```
const root = mount(...)
let element = root.find(".UpgradeBadge")
expect(element).toHaveLength(0)
```
or whatever it'd be

internal/store/engine_state.go
nicks: take a look at how DockerPruneSettings does it

PR #2525, opened 2019-11-08
upper_test: fix nextCallComplete race condition

landism:
1. test starts the engine with two manifests
2. engine completes initial builds of both manifests
3. test calls `nextCallComplete()` twice

After (2), `EngineState.CompletedBuildCount` is 2
In the first call in (3), `nextCall` returns a `call` with `count == 1`, so `nextCallComplete` blocks until `CompletedBuildCount` is 1, which it will never be.

Since CompletedBuildCount is monotonic, `>=` seems good enough.

### DIFF
--- a/internal/engine/upper_test.go
+++ b/internal/engine/upper_test.go
@@ -3269,7 +3269,7 @@ func (f *testFixture) WaitUntilManifestState(msg string, name model.ManifestName
 
 func (f *testFixture) nextCallComplete(msgAndArgs ...interface{}) buildAndDeployCall {
 	call := f.nextCall(msgAndArgs...)
-	f.waitForCompletedBuildCount(call.count)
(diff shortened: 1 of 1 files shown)


internal/engine/upper_test.go
```diff
@@ -3364,6 +3364,12 @@ func (f *testFixture) notifyAndWaitForPodStatus(pod *v1.Pod, mn model.ManifestNa
 	})
 }
 
+func (f *testFixture) waitForCompletedBuildCountAtLeast(count int) {
+	f.WaitUntil(fmt.Sprintf("%d builds done", count), func(state store.EngineState) bool {
+		return state.CompletedBuildCount >= count
+	})
+}
+
 func (f *testFixture) waitForCompletedBuildCount(count int) {
```
nicks: i think it would have been find to change this to >=

PR #2531, opened 2019-11-11
integration: TestLiveUpdateAfterCrashRebuild: log pod json on failure

landism:


integration/live_update_after_crash_rebuild_test.go
```diff
@@ -14,6 +14,15 @@ import (
 func TestLiveUpdateAfterCrashRebuild(t *testing.T) {
 	f := newK8sFixture(t, "live_update_after_crash_rebuild")
 	defer f.TearDown()
+
+	// TODO(matt) fix flakiness and remove this debug logging
+	succeeded := false
+	defer func() {
+		if !succeeded {
```
nicks: are you familiar with t.Failed()?

https://golang.org/pkg/testing/#T.Failed

it will also work if an assert fails

landism:
it turns out I have local repro, so don't need this in master

integration/live_update_after_crash_rebuild_test.go
landism: that's much nicer, thanks

PR #2534, opened 2019-11-11
integration: TestLiveUpdateAfterCrashRebuild: fix live update

landism:
1. we're using restart.sh but didn't run it with start.sh
2. the dockerfile writes to /app, but the live_update config thinks it's in /src

(note: this only allow live update to actually work - there's still the weird pod flakiness after this)


landism:
closing in favor of #2535

PR #2536, opened 2019-11-11
webui: use linebreaks instead of display: block to separate lines

### LINKED ISSUE #2524: Copy + paste from LogPane in to rich text fields doesn't work
It appears to be missing newlines

>Seems tilt is styling `<code>` (typically a display: inline element) as display: block such that each line is its own `<code>` element.  Copying and pasting that confuses JIRA's rich text editor which doesn't know how to interpret that.
>
>One way to fix would be to remove display: block from the style being applied to the `<code>`elements and inserting `<br>`s in between each code element

>Confusingly, paste into say Vim works, because it uses the plain-text buffer

landism:
Resolves #2524 

### Problem

In some cases, copy/pasting from Tilt into a rich text editor causes linebreaks to be lost.
We render log lines as individual `<div>s` and rely on their `display: block` behavior to get us newline behavior. Apparently some versions of some text editors do not know how to handle this, and so text gets pasted without newlines.

### Solution

Drop the `display: block` and `<div>`s on our log lines, and render then with `<span>` and `<br>`s instead.

I was unfortunately unable to repro this bug in the first place, so I haven't been able to verify the fix but the users who reported the bug also reported that making the analogous change in the chrome developer console resolved the issue.

### DIFF
--- a/web/src/AnsiLine.tsx
+++ b/web/src/AnsiLine.tsx
@@ -8,9 +8,12 @@ type AnsiLineProps = {
 
 let AnsiLine = React.memo(function(props: AnsiLineProps) {
   return (
-    <Ansi linkify={false} useClasses={true} className={props.className}>
-      {props.line}
-    </Ansi>
+    <React.Fragment>
+      <Ansi linkify={false} useClasses={true} className={props.className}>
+        {props.line}
+      </Ansi>
+      <br />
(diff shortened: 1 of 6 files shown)


nicks:
let's keep the snapshot testing discussion off this pr, happy to discuss it offline

PR #2539, opened 2019-11-12
engine: fix force rebuild after failed build

landism:
#### repro
1. `tilt up` and successfully deploy a service that has live_update configured
2. make a change to the service's code such that live_update runs and fails
3. click force update in the web UI

#### observed
the force update is a live update

#### expected
the force update is an image build

This is because we determine whether a triggered build is a forced update by checking if `PendingFileChanges` is non-empty. Since we only clear `PendingFileChanges` on successful builds, this means that if we've made a change and then failed the build, then subsequent build triggers will not set ForceUpdate.

### Solution

Instead of setting ForceUpdate iff there are any files in `PendingFileChanges`, set ForceUpdate iff there are any files in `PendingFileChanges` with a timestamp since the last build's start time.

### DIFF
--- a/internal/engine/buildcontroller_test.go
+++ b/internal/engine/buildcontroller_test.go
@@ -543,6 +543,37 @@ func TestBuildControllerManualTriggerWithoutFileChangesForceUpdates(t *testing.T
 	})
 }
 
+func TestBuildControllerManualTriggerWithFileChangesSinceLastSuccessfulBuildButBeforeLastBuild(t *testing.T) {
+	f := newTestFixture(t)
+	defer f.TearDown()
+	mName := model.ManifestName("foobar")
+
+	manifest := f.newManifest(mName.String())
+	manifests := []model.Manifest{manifest}
+	f.Start(manifests, true)
+
+	f.nextCallComplete()
+
+	f.b.nextBuildFailure = errors.New("build failure!")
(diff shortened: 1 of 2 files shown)


internal/engine/buildcontroller_test.go
```diff
@@ -543,6 +543,37 @@ func TestBuildControllerManualTriggerWithoutFileChangesForceUpdates(t *testing.T
...
+
+	f.b.nextBuildFailure = errors.New("build failure!")
+	f.fsWatcher.events <- watch.NewFileEvent(f.JoinPath("main.go"))
+	f.nextCallComplete()
+
+	f.store.Dispatch(server.AppendToTriggerQueueAction{Name: mName})
+	call := f.nextCallComplete()
+	state := call.oneState()
+	assert.Equal(t, []string{f.JoinPath("main.go")}, state.FilesChanged())
+	assert.True(t, state.NeedsForceUpdate)
```
maiamcc: might put a comment to emphasize that this is the behavior you're specifically trying to test... (OR make the test name longer to say that you expect a force udpate?) but lgtm!

maiamcc: might put a comment to emphasize that this is the behavior you're specifically trying to test... (OR make the test name longer to say that you expect a force udpate?) but lgtm!

PR #2545, opened 2019-11-12
integration: fix up sample projects to actually need restart.sh

nicks:
Hello @maiamcc, @landism,

Please review the following commits I made in branch nicks/integration2:

d6d51602b3e1b629959d0d582b5209bb98fd8ad4 (2019-11-12 15:19:51 -0500)
integration: fix up sample projects to actually need restart.sh


nicks:
In this pattern, we have a compile.sh step (that runs at build-time) and a separate main.sh step (that runs when the pod starts). Both are necessary, or changes to source.txt won't get picked up.

PR #2549, opened 2019-11-13
tilt: remove timing.py

landism:


nicks:
> When you delete code, don’t say “Why was this even here?” or “This code was terrible, I’m glad it’s gone now.” Thank your code for the job it’s done.

“The Life-Changing Magic of Deleting Code” by Jean Hsu https://link.medium.com/Uy709ttSA1

PR #2551, opened 2019-11-13
web: If the ending line is an element, grab the lineID directly

jazzdan:
When you select UP the ending log line is an element, not a text node.

When checking the ending element, see which kind of node it is, and handle it appropriately.


web/src/LogPane.tsx
```diff
@@ -121,9 +127,13 @@ class LogPane extends Component<LogPaneProps, LogPaneState> {
     }
   }
 
+  isElement(n: any): n is Element {
+    return (n as Element).getAttribute !== undefined
```
jazzdan: Is there a better way to do this?

nicks: this seems wrong. what are you trying to accomplish?

web/src/LogPane.tsx
```diff
@@ -108,7 +109,12 @@ class LogPane extends Component<LogPaneProps, LogPaneState> {
         !node.isEqualNode(end)
       ) {
         let beginningLogLine = this.findLogLineID(beginning.parentElement)
```
nicks: let's update findLogLineID to accept a Node, and handle this internally

nicks:
also: this seems like a good time to break this out into a helper method with a unit test. node traversal is hard to get right; these sorts of edge cases are common. We want to make sure that point fixes or changes to the dom structure don't cause regressions.

jazzdan:
@nicks agreed, I'll work on that now.

web/src/findLogLine.ts
```diff
@@ -0,0 +1,19 @@
+const LINE_ID_ATTR_NAME = "data-lineid"
+
+export default function findLogLineID(
+  el: HTMLElement | Node | null
+): string | null {
+  if (el === null) {
+    return null
+  }
+
+  if ((el as HTMLElement).getAttribute(LINE_ID_ATTR_NAME)) {
```
nicks: i think you need to check that getAttribute exists, right? the `as` doesn't do any runtime checks

web/src/findLogLine.test.ts
```diff
@@ -0,0 +1,28 @@
...
+    const dom = new JSDOM(logPaneDOM)
+    const node = dom.window.document.getElementById("start1")
+
+    const actual = findLogLineID(node)
+    expect(actual).toBe("1920")
+  })
+
+  it("returns the value of parent's data-lineid if the element has no data-lineid", () => {
+    const dom = new JSDOM(logPaneDOM)
+    const node = dom.window.document.getElementById("start2")
```
nicks: can you also test that this works with findLogLineID(node.firstChild)?

jazzdan: Yup, working on that now. These tests were WIP. :)

web/src/findLogLine.ts
jazzdan: Yeah I'm trying to figure out the idiomatic way to do this in typescript...

web/src/findLogLine.ts
```diff
@@ -0,0 +1,22 @@
...
+
+export default function findLogLineID(
+  el: HTMLElement | Node | null
+): string | null {
+  if (el === null) {
+    return null
+  }
+
+  if (
+    typeof (el as HTMLElement).getAttribute === "function" &&
```
jazzdan: I think there's probably a better way to do this type juggling but I'm not sure what it is.

jazzdan:
@nicks ready for review!

web/src/findLogLine.ts
nicks: function findLogLineID(node: Node | null)

let el = (node as HTMLElement)
maybe?

it's a little bit weird that you have a variable called `el` that's not an element

PR #2554, opened 2019-11-13
web: allow alerts to be dismissed in all contexts _except_ snapshots

### LINKED ISSUE #2553: Can't dismiss alerts if Tilt is running remotely
If you're running Tilt remotely via the [newly added hosts flag](https://github.com/windmilleng/tilt/issues/2321) you won't be able to dismiss dismissible alerts.

This is because we check [if Tilt is being served from localhost](https://github.com/windmilleng/tilt/blob/a7df4bbc22b7d4803d4d966a35ee007d61637bde/web/src/AlertPane.tsx#L47). If it's not, we don't show the dismiss alert button.

If you're running Tilt remotely though, this will return false and you won't be able to dismiss alerts.

I think this needs to be changed to be:

```typescript
if (alert.dismissHandler && !isSnapshot) {
```

jazzdan:
Fixes #2553

### DIFF
--- a/web/src/PathBuilder.tsx
+++ b/web/src/PathBuilder.tsx
@@ -25,16 +25,8 @@ class PathBuilder {
     return this.snapId !== ""
   }
 
-  private isLocal() {
-    return this.host.indexOf("localhost") === 0
-  }
-
   private snapshotDataUrl(): string {
-    let scheme = "https"
-    if (this.isLocal()) {
-      scheme = "http"
-    }
-    return `${scheme}://${this.host}/api/snapshot/${this.snapId}`
+    return `/api/snapshot/${this.snapId}`
   }
 
   private snapshotPathBase(): string {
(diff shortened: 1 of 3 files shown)


web/src/PathBuilder.tsx
```diff
@@ -21,14 +21,14 @@ class PathBuilder {
...
 
-  isLocal() {
-    return this.host.indexOf("localhost") === 0
-  }
-
   isSnapshot(): boolean {
     return this.snapId !== ""
   }
 
+  private isLocal() {
```
nicks: i'm pretty sure the other places this is used are wrong too

web/src/PathBuilder.tsx
```diff
@@ -21,14 +21,14 @@ class PathBuilder {
...
+    return this.host.indexOf("localhost") === 0
+  }
+
   private snapshotDataUrl(): string {
     let scheme = "https"
     if (this.isLocal()) {
```
nicks: for example, why does this need include the scheme and host at all? it should be able to just use a relative path like "/api/..."

Issue #2374, opened 2019-10-18
tilt sometimes fails to notice containers are ready


landism:
snapshot: https://cloud.tilt.dev/snapshot/AaK2z9oLxm9AvZwOmlI=

Tilt thinks the containers are not ready, and shows the resources as pending, even though k8s says the containers are ready.

We do not currently have a good repro.

This is because the view the backend has handed it says not all containers are ready:
```
$ curl -sSfL https://cloud.tilt.dev/api/snapshot/AaK2z9oLxm9AvZwOmlI= | jq '.View.Resources[] | select(.Name=="web") | .K8sResourceInfo.AllContainersReady'
false
```

however, `kubectl get pod` shows that it's running and ready:
```
linkerd-web-5dcb488644-8f8qs              2/2     Running   0          7m18s
```

and `kubectl describe pod` does not show any non-ready containers:
<details><summary>click to expand</summary><p>

---
```
k -n linkerd describe po linkerd-web-7ffb8fb78-fw8vv                                                        (kubernetes-admin@linkerd/default)
Name:           linkerd-web-7ffb8fb78-fw8vv
Namespace:      linkerd
Priority:       0
Node:           linkerd-control-plane/172.17.0.2
Start Time:     Fri, 18 Oct 2019 10:19:50 -0700
Labels:         linkerd.io/control-plane-component=web
                linkerd.io/control-plane-ns=linkerd
                linkerd.io/proxy-deployment=linkerd-web
                pod-template-hash=7ffb8fb78
Annotations:    linkerd.io/created-by: linkerd/cli stable-2.6.0
                linkerd.io/identity-mode: default
                linkerd.io/proxy-version: stable-2.6.0
Status:         Running
IP:             10.244.0.21
Controlled By:  ReplicaSet/linkerd-web-7ffb8fb78
Init Containers:
  linkerd-init:
    Container ID:  containerd://ed8f7e5fd3bf083c7f08191665c6fef65c4a4cc9401c2cd0f59a325aae1e5895
    Image:         gcr.io/linkerd-io/proxy-init:v1.2.0
    Image ID:      gcr.io/linkerd-io/proxy-init@sha256:c0174438807cdd711867eb1475fba3dd959d764358de4e5f732177e07a75925b
    Port:          <none>
    Host Port:     <none>
    Args:
      --incoming-proxy-port
      4143
      --outgoing-proxy-port
      4140
      --proxy-uid
      2102
      --inbound-ports-to-ignore
      4190,4191
[...]

PR #2555, opened 2019-11-13
watch: ignore resourceVersion

landism:
Potentially addresses #2374 (or, at least, addresses a cause of it, which may or may not be the only cause)

### Problem

As an optimization, when one of Tilt's watchers gets an update for an object from a K8s informer, it checks to see if its resourceVersion is older than the last resourceVersion it saw, and if so, ignores it.
Unfortunately:
1. resourceVersion is a string, and Tilt is using string comparison (so, e.g., if we see resourceVersion "9", we'll then ignore "10" through "89" and "100" through "899")
2. the comment on resourceVersion says clients must treat it as opaque

### Solution

Simply remove the resourceVersion check. Theoretically the K8s informer provides us updates in order, so it does us no good.

### DIFF
--- a/internal/engine/k8swatch/event_watch_manager.go
+++ b/internal/engine/k8swatch/event_watch_manager.go
@@ -117,30 +117,6 @@ func (m *EventWatchManager) setupNewUIDs(ctx context.Context, st store.RStore, n
 	}
 }
 
-// Record the event update, and return true if this is newer than
-// the state we already know about.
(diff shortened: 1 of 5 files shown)


landism:
I cooooould write unit tests for resourceVersion on the other watchers as well, but writing tests for deleted code, especially deleted code that was never tested in the first place, doesn't seem super valuable.

PR #2556, opened 2019-11-14
webui: inline display for code in alerts and facets

landism:
#2536 changed our `<code>`s from `display: block` to `display: inline` and added `<br>`s to get newlines.

However, it failed to change `display` for `<code>`s in the alerts or facets pane, so now those are double-spaced.

These should probably all be one css class, but I don't think the week before kubecon is the right time for a css refactor.


landism:
This is a bigger problem than I realized. Will repost later with more fixes.

PR #2557, opened 2019-11-14
webui: fix facets and alerts linebreaks

landism:
### Problems

1. In #2556 , I changed our `<code>`'s `display` from `block` to `inline`, and inserted `<br>`s to compensate. However, I failed to change `block` to `inline` for the other uses of `<code>`.
2. Also, the alerts pane works by sending the full log to the browser and using CSS to hide all but the last N lines.
  a. We're using nth-last-child to do this, which was inappropriately counting the `<br>`s.
  b. The rule that hides all log lines except the nth-last-child wasn't applying to the `<br>`s, so we ended up with the original log height, and then just the last 4 log lines.

### Solutions

1. Fix the `<code>` style for alerts and facets. These should probably all be one css class, but the week before kubecon doesn't seem like the right time to take on a css refactor.
  a. note that for alert, we need to keep `display: block` for the hidden `code` elements, so that setting their `height` to 0 actually takes effect.
2. Use `\n` instead of `<br>`, since this is inside `white-space: pre-wrap` anyway. There might be Good Web Design reasons for not doing this. I'm interested in hearing. This is mostly just so we can keep using the existing code for hiding all but the last n lines of the log in the alert pane.


web/src/AlertPane.scss
```diff
@@ -66,6 +66,7 @@
 
 // ...only show last few lines
```
hyu: FWIW, this CSS for hiding the lines can be more simple and clear:

```
// Hide all lines by default
.AlertPane-item code {
  white-space: pre-wrap;
  display: none;
}
```

I had `height: 0;` and `opacity: 0;` previously because I was getting ready for a "show more" button that would reveal all the lines with smooth animation. Which I never ended up actually implementing. 

(When you use `display: none` and `display: [something]` for showing and hiding, it takes away your animation options.)

landism: much nicer!

PR #2559, opened 2019-11-14
revert the revert of the revert of the revert of podtemplatespec hashes

landism:
This is the third attempt at #2439. See that PR for the problem description and solution.

The second attempt, #2519, was reverted because it was making an integration test fail pretty consistently. We're now reasonably sure those failures were due to a pre-existing bug resolved by #2555, and not an issue with #2519 itself. In a branch with #2519 and #2555, the test passed 100x in a row.

### DIFF
--- a/internal/engine/pod.go
+++ b/internal/engine/pod.go
@@ -114,7 +114,7 @@ func maybeTrackPod(ms *store.ManifestState, action k8swatch.PodChangeAction) (*s
 	ns := k8s.NamespaceFromPod(pod)
 	hasSynclet := sidecar.PodSpecContainsSynclet(pod.Spec)
 	runtime := ms.GetOrCreateK8sRuntimeState()
-	isCurrentDeploy := runtime.HasOKPodTemplateSpecHash(pod) // pod is from the most recent Tilt deploy
+	isCurrentDeploy := runtime.HasOKPodTemplateSpecHash(pod) // is pod from the most recent Tilt deploy?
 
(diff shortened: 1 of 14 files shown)


internal/engine/pod.go
```diff
@@ -103,35 +103,41 @@ func matchPodChangeToManifest(state *store.EngineState, action k8swatch.PodChang
...
-func trackPod(ms *store.ManifestState, action k8swatch.PodChangeAction) (*store.Pod, bool) {
+func maybeTrackPod(ms *store.ManifestState, action k8swatch.PodChangeAction) (*store.Pod, bool) {
 	pod := action.Pod
 	podID := k8s.PodIDFromPod(pod)
 	startedAt := pod.CreationTimestamp.Time
 	status := k8swatch.PodStatusToString(*pod)
 	ns := k8s.NamespaceFromPod(pod)
 	hasSynclet := sidecar.PodSpecContainsSynclet(pod.Spec)
 	runtime := ms.GetOrCreateK8sRuntimeState()
+	isCurrentDeploy := runtime.HasOKPodTemplateSpecHash(pod) // pod is from the most recent Tilt deploy
```
maiamcc:
```suggestion
	isCurrentDeploy := runtime.HasOKPodTemplateSpecHash(pod) // is pod from the most recent Tilt deploy?
```

landism: betteridge's law?

PR #2565, opened 2019-11-15
web: cleaner/assertive tests for trigger button

maiamcc:
banged this out on the train b/c I didn't have internet. wdyt? is this the style of test we're going for?

note: the tests i moved into `SidebarTriggerButton.test.tsx` DO test some Sidebar logic,
happy to move them back to `Sidebar.test.tsx` but they're 80% about the TriggerButton logic
and thought this made the sidebar test file less cluttered.


web/src/SidebarTriggerButton.test.tsx
```diff
@@ -1,7 +1,32 @@
...
+import { oneResource, twoResourceView } from "./testdata"
+import Sidebar, { SidebarItem } from "./Sidebar"
+import { MemoryRouter } from "react-router"
+import PathBuilder from "./PathBuilder"
+
+let pathBuilder = new PathBuilder("localhost", "/")
+
+let assertTriggerButtonProps = (
+  button: any,
+  expectClickable: boolean,
```
nicks: this seems hard to maintain...it might be easier to pass in a list of classes. @hyu what do you think?

nicks: i guess a more javascript-y way to do this would be to have a type like 
{isCickable?: boolean, isSelected?: boolean} so that the caller only has to pass the ones that are true, tho @hyu might have better ideas

maiamcc:
bump -- @nicks, better?

web/src/SidebarTriggerButton.tsx
```diff
@@ -2,6 +2,13 @@ import React, { PureComponent } from "react"
 import { TriggerMode } from "./types"
 import "./SidebarTriggerButton.scss"
 
+export enum TriggerButtonTooltip {
```
nicks: this seems like a weird thing to make an enum, since it's not really a type. probably better to be a plain object, which is what people usually use for messages

maiamcc: oh, sure, i just don't know the idiom here! Is the idiomatic thing:
```
export const TooltipAlreadyQueued = "Cannot trigger an update if resource is already queued for update."
export const TooltipManualResourcePendingChanges = "This manual resource has pending file changes; click to trigger an update."
...
```
or
```
export const TriggerButtonTooltip = {
  AlreadyQueued: "Cannot trigger an update if resource is already queued for update.",
  ManualResourcePendingChanges: "This manual resource has pending file changes; click to trigger an update.",
  ...
}
```
or something else?

}

nicks: either is fine

PR #2029, opened 2019-08-09
Revert "klog: hide 'resourceVersion too old' warnings (#1955)"

nicks:
Hello @landism,

Please review the following commits I made in branch nicks/klog:

f93dee7eea256326cfae8b9eb7d512ca73f13f30 (2019-08-09 13:26:15 -0400)
Revert "klog: hide 'resourceVersion too old' warnings (#1955)"
This reverts commit 21e97b8d6537f278d1fa0971cc0b754f089f5348.

PR #2575, opened 2019-11-18
klog: ignore too old resource version warnings

### LINKED ISSUE #2406: Tilt should ignore too old resource version errors
Tilt can spew error message like the following:

```
github.com/windmilleng/tilt/internal/k8s/watch.go:172: watch of *v1.Pod ended with: too old resource version: 191906663 (191912819) W1021 14:53:11.799222 68992 reflector.go:299] github.com/windmilleng/tilt/internal/k8s/watch.go:172: watch of *v1.Pod ended with: too old resource version: 191906663 (191912819)
```

These should be ignored similarly to how we ignore `watch of *v1.Event` errors in https://github.com/windmilleng/tilt/pull/1820

landism:
Resolves #2406

We previously fixed a different instance of this.
#1955 used the same approach to hide a very similar, but different warning.
At some point, K8s changed that warning to not be a warning, so #1955 was reverted: #2029
Basically, [in this code](https://github.com/windmilleng/tilt/blob/90aa4750036b819c9f8da2fa4bf90901b8a43019/vendor/k8s.io/client-go/tools/cache/reflector.go#L297). The old warning was on line 297 and the warning addressed by this PR is hitting line 299.

This is apparently because this warning is not setting its `ReasonForError` to `metav1.StatusReasonExpired`, so [`IsResourceExpired`](https://github.com/windmilleng/tilt/blob/90aa4750036b819c9f8da2fa4bf90901b8a43019/vendor/k8s.io/apimachinery/pkg/api/errors/errors.go#L490) returns false.

I spent 15 minutes trying to track down what was failing to set the `ReasonForError`, and figured it wasn't worth the time, especially given I don't have a repro for this. Since the old warning was something only relatively recently fixed, it seems quite possible that we want this behavior to mitigate the warnings in older clusters anyway.

### DIFF
--- a/internal/cli/klog.go
+++ b/internal/cli/klog.go
@@ -55,6 +55,7 @@ func newFilteredWriter(underlying io.Writer, filterFunc func(s string) bool) io.
 }
 
 // everything on google indicates this warning is useless and should be ignored
+// https://github.com/kubernetes/kubernetes/issues/22024
 var isResourceVersionTooOldRegexp = regexp.MustCompile("reflector.go.*watch of.*ended with: too old resource version")
 
 // HACK(nick): The Kubernetes libs we use sometimes use klog to log things to
--- a/internal/cli/klog.go
+++ b/internal/cli/klog.go
@@ -3,13 +3,59 @@ package cli
 import (
 	"flag"
 	"fmt"
+	"io"
 	"log"
 	"os"
+	"regexp"
 
 	"k8s.io/klog"
 )
 
 var klogLevel = 0
+
+type filteredWriter struct {
+	underlying io.Writer
+	filterFunc func(s string) bool
+	leftover   []byte
+}
+
+func (fw *filteredWriter) Write(buf []byte) (int, error) {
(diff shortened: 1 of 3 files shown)


internal/cli/klog.go
```diff
@@ -3,23 +3,73 @@ package cli
...
+
+	return written, nil
+}
+
+// lines matching `filterFunc` will not be output to the underlying writer
+func newFilteredWriter(underlying io.Writer, filterFunc func(s string) bool) io.Writer {
+	return &filteredWriter{underlying: underlying, filterFunc: filterFunc}
+}
+
+// everything on google indicates this warning is useless and should be ignored
```
nicks: Is there a bug upstream yet that we can link to? If not, can you file one?

 In the past, the kubernetes people have seemed pretty receptive to prs that add hooks for silencing warnings.

landism:
k8s bug: https://github.com/kubernetes/kubernetes/issues/22024

PR #2578, opened 2019-11-18
tiltfile: add flags.set_resources

landism:
(part of https://github.com/windmilleng/tilt.build/pull/325)

This adds a `flags.set_resources` builtin that overrides the resource list specified as positional arguments to `tilt up`

### DIFF
--- a/internal/tiltfile/flags/flags.go
+++ b/internal/tiltfile/flags/flags.go
@@ -4,13 +4,13 @@ import (
 	"github.com/pkg/errors"
 	"go.starlark.net/starlark"
 
-	"github.com/windmilleng/tilt/internal/tiltfile/value"
-
 	"github.com/windmilleng/tilt/internal/tiltfile/starkit"
+	"github.com/windmilleng/tilt/internal/tiltfile/value"
+	"github.com/windmilleng/tilt/pkg/model"
 )
 
 type Settings struct {
-	Resources []string
+	resources []string
 }
 
 type Extension struct {
@@ -44,7 +44,7 @@ func setResources(thread *starlark.Thread, fn *starlark.Builtin, args starlark.T
 	}
 
 	err = starkit.SetState(thread, func(settings Settings) Settings {
-		settings.Resources = resources
+		settings.resources = resources
 		return settings
 	})
 	if err != nil {
@@ -68,4 +68,17 @@ func GetState(m starkit.Model) (Settings, error) {
 	var state Settings
 	err := m.Load(&state)
 	return state, err
+}
+
+func (s Settings) Resources(args []model.ManifestName) []model.ManifestName {
+	if s.resources == nil {
+		return args
+	}
+
+	var ret []model.ManifestName
+	for _, r := range s.resources {
+		ret = append(ret, model.ManifestName(r))
+	}
+
+	return ret
 }
--- a/internal/tiltfile/flags/flags.go
+++ b/internal/tiltfile/flags/flags.go
@@ -1,0 +1,71 @@
+package flags
+
+import (
+	"github.com/pkg/errors"
+	"go.starlark.net/starlark"
+
+	"github.com/windmilleng/tilt/internal/tiltfile/value"
+
+	"github.com/windmilleng/tilt/internal/tiltfile/starkit"
+)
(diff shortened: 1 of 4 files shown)


internal/tiltfile/tiltfile_state.go
```diff
@@ -204,6 +206,14 @@ to your Tiltfile. Otherwise, switch k8s contexts and restart Tilt.`, kubeContext
 	}
 	manifests = append(manifests, localManifests...)
 
+	flagsState, _ := flags.GetState(result)
+	if len(flagsState.Resources) > 0 {
+		requested = nil
+		for _, r := range flagsState.Resources {
```
nicks: It might make more sense for this machinery to be a method of flags state? Or better yet, make flag state store manifestnames so you don't need to do this conversion at all?

internal/tiltfile/tiltfile_test.go
```diff
@@ -4226,6 +4226,48 @@ local_resource('e', 'echo e')
 	}
 }
 
+func TestSetResources(t *testing.T) {
```
nicks: Can you make this test part of the subpackage Trying to get to a place where the extensions are independent and it's easier to build tooling on them independent of tilt

landism: Since you mentioned the test but not the logic, I want to make sure you're aware that this is specifically testing the interaction between the `flags` state and the `tilt up` args. We could move the test into the subpackage, but then that interaction would also need to be moved in, and I feel like that makes more sense in tiltfile proper?

(FWIW, the tests of the independent `flags` behavior are in the `flags` package)

landism: Wow, I somehow missed the first sentence of the previous comment.

When working on the next PR after this one, it became apparent that this logic definitely belongs in the extension.

internal/tiltfile/tiltfile_state.go
landism: yeah, that makes this cleaner

internal/tiltfile/tiltfile_state.go
```diff
@@ -204,6 +206,16 @@ to your Tiltfile. Otherwise, switch k8s contexts and restart Tilt.`, kubeContext
 	}
 	manifests = append(manifests, localManifests...)
 
+	// if none were requested, then default to all
+	if requested == nil {
```
maiamcc: ah, i see in a later PR you changed this to "args" which is GOOD.

I'd clarify this comment--the subject is unclear. maybe something like:
> If no args were specified, i.e. user requested no resources via args, then default to all resources. (Resources specified in Tiltfile override resources specified via args.)

internal/tiltfile/flags/flags.go
```diff
@@ -0,0 +1,84 @@
+package flags
```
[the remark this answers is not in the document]
landism: agreed

internal/tiltfile/tiltfile_state.go
landism: Hm, I agree with your comment, but I don't think it's applicable to master, since this code changed.  But, then, your comment is referring to post-later-PR, so I'm not sure how to interpret it.

internal/tiltfile/flags/flags.go
```diff
@@ -0,0 +1,84 @@
...
+
+func (e Extension) NewState() interface{} {
+	return Settings{}
+}
+
+func (Extension) OnStart(env *starkit.Environment) error {
+	return env.AddBuiltin("flags.set_resources", setResources)
+}
+
+func setResources(thread *starlark.Thread, fn *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
```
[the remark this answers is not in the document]
landism: it was originally something like `enable_resources` and I suggested changing it to `set_resources`, so that it was clear it was overwriting existing state, rather than potentially adding to existing state, at which point maybe it needs to be preceded by a `clear_resources` or something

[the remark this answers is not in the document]
landism: discussed offline, going with `set_enabled_resources`

PR #2588, opened 2019-11-21
tiltfile: add flags.define_string_list and flags.parse

landism:
This implements more of https://github.com/windmilleng/tilt.build/blob/183d9b52ecc11a4a834d3ee80f69a04efbb6acde/docs/flags.md

With this PR, `flags.parse` simply returns whatever was passed in after `tilt up` on the current invocation.

It does not have any notion of "sticky" vars per the doc, or have any awareness of tilt_config.json. That will be in a future PR.


internal/tiltfile/flags/flags.go
```diff
@@ -54,31 +98,170 @@ func setResources(thread *starlark.Thread, fn *starlark.Builtin, args starlark.T
...
+}
+
+func (fv *Strings) String() string {
+	return strings.Join(fv.Values, ",")
+}
+
+func (ad ArgsDef) parse(args []string, w io.Writer) (starlark.Value, error) {
+	fs := flag.NewFlagSet("", flag.ContinueOnError)
+	fs.SetOutput(w)
+	stringListValues := make(map[string]*Strings)
```
nicks: this function seems like it's going to get pretty gnarly if you add more types.

i wonder if there's a way it could be more modular, like if the argDef structs knew how to set up their own flags and return values?

internal/tiltfile/starkit/get_state.go
```diff
@@ -0,0 +1,22 @@
+package starkit
+
+import (
+	"fmt"
+	"reflect"
+
+	"go.starlark.net/starlark"
+)
+
+func GetState(t *starlark.Thread, v interface{}) (interface{}, error) {
```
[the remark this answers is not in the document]
landism: Ah, I only added this because I looked for an existing pattern and didn't find one. Even after looking, I only see one place that does this, and your reference to "all the existing code" makes me wonder if I'm not looking at the right thing? All the other calls I see are getting the state from the model from the result of ExecFile, not the thread.

Anyway, yeah, ModelFromThread -> GetState seems cleaner, and we can worry about StateFromThread later if we need it in more places.

internal/tiltfile/flags/flags.go
```diff
@@ -54,31 +98,170 @@ func setResources(thread *starlark.Thread, fn *starlark.Builtin, args starlark.T
...
+		return starlark.None, err
 	}
-	return state
+
+	st, err := starkit.GetState(thread, Settings{})
+	if err != nil {
+		return starlark.None, err
+	}
+
+	return st.(Settings).argDef.parse(e.cmdLineArgs, starkit.NewThreadWriter(thread))
```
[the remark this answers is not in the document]
landism: yeah, that's probably better

internal/tiltfile/starkit/environment.go
```diff
@@ -35,15 +35,18 @@ type Environment struct {
 	print          func(thread *starlark.Thread, msg string)
 	extensions     []Extension
 	fakeFileSystem map[string]string
+	// args that were provided beyond tilt-recognized args
+	extraCmdLineArgs []string
```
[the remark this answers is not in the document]
landism: I guess I don't yet have a good mental model around what the starkit types represent. I figured "execution environment" would provide environmental state, such as the args the file was exec'd with.

Apparently I checked every extension except K8sContext.Extension when looking for an example of another way to do this.

Passing to NewExtension seems cleaner, thanks.

internal/tiltfile/flags/args_def.go
```diff
@@ -0,0 +1,115 @@
...
+		err := ret.SetKey(starlark.String(name), def.starlark())
+		if err != nil {
+			return starlark.None, w.String(), err
+		}
+	}
+
+	return ret, w.String(), nil
+}
+
+// makes a new builtin with the given argValue constructor
```
nicks: can you add some comments on why this is parameterized?

the extra function closure makes the code a lot harder to understand, and i don't understand why you did it this way or what value this adds

internal/tiltfile/starkit/thread_writer.go
```diff
@@ -0,0 +1,34 @@
+package starkit
+
+import (
+	"bytes"
+	"io"
+
+	"go.starlark.net/starlark"
+)
+
+type threadWriter struct {
```
nicks: i think you can delete this now?

internal/tiltfile/flags/args_def.go
landism: I've expanded the comment, but I'm not sure if it helps much.
I suspect the main reason the value isn't apparent is that there's currently only one arg type, and once we add more arg types, it will be clearer.

PR #2594, opened 2019-11-25
build: fix up custom_build to work with non-docker builders like buildah

nicks:
Hello @landism, @maiamcc,

Please review the following commits I made in branch nicks/issue2589:

0ff01c034b3a92ed03b3531dd433b4e147369d71 (2019-11-25 13:00:45 -0500)
build: fix up custom_build to work with non-docker builders like buildah
Currently, custom_build checks to make sure that your local docker registry
has the built image. This doesn't work if you don't have docker installed at all.

Long-term, I think we probably need to rethink the custom_build API to be richer
than a single shell command (and also provide 'tag' and 'push' commands).

### DIFF
--- a/internal/build/custom_builder.go
+++ b/internal/build/custom_builder.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/docker/distribution/reference"
 	"github.com/opencontainers/go-digest"
+	"github.com/pkg/errors"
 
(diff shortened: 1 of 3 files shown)


nicks:
this is supporting https://github.com/windmilleng/tilt/issues/2589

PR #2596, opened 2019-11-25
tiltfile: persist flags to tilt_config.json

landism:
The persistence part of https://github.com/windmilleng/tilt.build/pull/325

When the user calls `flags.parse`:
  1. initial flag values are read from tilt_config.json
  2. cmd-line args are parsed and merged on top of flag values from tilt_config.json
  3. the result of that merge is written back out to tilt_config.json

This means, e.g.:
`tilt up -- --foo bar --baz quu` -> saves `{"foo": ["bar"], "baz": ["quu"]}` to the config
`tilt up -- --foo bar2` -> loads the previous config values, merges in the new args, resulting in `{"foo": ["bar2"], "baz": ["quu"]}`, which is written back out to the config and used as the effective flags for the current run.


internal/engine/upper.go
```diff
@@ -422,13 +423,38 @@ func handleStartProfilingAction(state *store.EngineState) {
...
+
 func handleFSEvent(
 	ctx context.Context,
 	state *store.EngineState,
 	event targetFilesChangedAction) {
 
 	if event.targetID.Type == model.TargetTypeConfigs {
 		for _, f := range event.files {
+			if f == state.FlagsState.ConfigPath {
+				ignored, err := ignoreFlagConfigUpdate(state)
```
nicks: we try pretty hard to make the reducers pure. they shouldn't read file system state. can you move this check into the subscriber?

internal/tiltfile/flags/args_def.go
```diff
@@ -2,66 +2,155 @@ package flags
...
+	usage    string
 }
 
 type ArgsDef struct {
 	positionalArgName string
 	args              map[string]argDef
 }
 
-func (ad ArgsDef) parse(args []string) (v starlark.Value, output string, err error) {
+func (ad ArgsDef) parse(flagsState model.FlagsState, args []string) (v starlark.Value, mergedArgs bool, output string, err error) {
```
nicks: this function is very complex and difficult to follow. 

maybe try separating out the parts that do file i/o from the parts that don't do file i/o?

nicks: alternatively, maybe this cries out for an ArgScanner struct that handles accumulated state, as described in:
https://dave.cheney.net/2019/01/27/eliminate-error-handling-by-eliminating-errors

internal/engine/configs/configs_controller.go
```diff
@@ -87,9 +87,10 @@ func (cc *ConfigsController) loadTiltfile(ctx context.Context, st store.RStore,
 	if !firstBuild {
 		logTiltfileChanges(ctx, filesChanged)
 	}
+	flagsState := state.FlagsState
 	st.RUnlockState()
 
-	tlr := cc.tfl.Load(ctx, tiltfilePath, args)
+	tlr := cc.tfl.Load(ctx, tiltfilePath, args, flagsState)
```
nicks: should FlagsState include `args`?

it feels like you always have to use them together, because `FlagsState` is more like "metadata about where the args came from"

internal/tiltfile/flags/args_def.go
```diff
@@ -2,66 +2,155 @@ package flags
...
+		}
+
+		f, err := os.Create(flagsState.ConfigPath)
+		if err != nil {
+			return nil, false, output, errors.Wrapf(err, "error opening %s for writing", flagsState.ConfigPath)
+		}
+		defer func() {
+			err2 := f.Close()
+			if err2 != nil && err == nil {
+				err = errors.Wrapf(err2, "error closing %s", flagsState.ConfigPath)
```
nicks: i don't think this does what you expect?

this is setting the `err` variable declared inside the `if` block, not the variable declared as a return value. here's an example:
https://play.golang.org/p/0fTWw4vsdPA

in general, the "set a named return value from inside a defer" is pretty error-prone and has lots of little gotchas like this. I'm not sure if there's much value in handling Close() errors at all

internal/tiltfile/flags/flags.go
```diff
@@ -88,9 +109,28 @@ func (e *Extension) parse(thread *starlark.Thread, fn *starlark.Builtin, args st
 		return starlark.None, err
 	}
 
-	ret, out, err := settings.argDef.parse(e.cmdLineArgs)
+	err = io.RecordReadFile(thread, settings.FlagsState.ConfigPath)
+	if err != nil {
+		return starlark.None, errors.Wrapf(err, "recording read file %s", settings.FlagsState.ConfigPath)
```
nicks: if this message is useful, shouldn't it live in RecordReadFile? RecordReadFile already has access to all this information

internal/tiltfile/flags/args_def.go
```diff
@@ -2,66 +2,155 @@ package flags
...
+	m := make(map[string]interface{})
+	err = jsoniter.NewDecoder(r).Decode(&m)
+	if err != nil {
+		return nil, errors.Wrapf(err, "error parsing json from %s", tiltConfigPath)
+	}
+
+	for k, v := range m {
+		def, ok := ad.args[k]
+		if !ok {
+			return nil, fmt.Errorf("%s specified unknown flag name '%s'", tiltConfigPath, k)
```
nicks: are there any tests that exercise these errors? i'm not totally sure i understand when these get triggered

internal/engine/configs/configs_controller.go
maiamcc: +1

nicks:
+1 for this feeling like a weird interaction. 

I feel like build systems have a long history of people trying this and deciding it was untenable (where this = "having the same file be both an input and an output")

have we considered having a separate tilt_config.json and tilt_config.lock.json?

landism:
> Oh wow at first glance, this behavior is SUPER CONFUSING. how would i wipe out previous flag values and start from a clean slate? this seems like potential for all sorts of weird interactions.

I completely agree. DB and I have spent hours debating this and neither has managed to convince the other. DB has talked to users who say this is what they want. Letting users try it out and getting feedback from them is the best resolution that I've thought of. If you have alternative approaches, or want to have a conversation with DB and/or me, I'm definitely open to that.

landism:
> I feel like build systems have a long history of people trying this and deciding it was untenable (where this = "having the same file be both an input and an output")
>
> have we considered having a separate tilt_config.json and tilt_config.lock.json?

Hm, I'm not sure I understand. My understanding was that yarn.lock and Gopkg.lock are both inputs and outputs of their respective tools? Would you be more comfortable with this if it were a sqlite db?

I think the feature as designed requires the same file to be both input and output, and a separate tilt_config.json and tilt_config.json.lock would not suffice to implement the feature. Am I missing something?

internal/engine/configs/configs_controller.go
landism: Yep, I actually already ended up doing this in the next PR. If it's ok, I won't spend the effort to pull that change into this PR.

internal/tiltfile/flags/args_def.go
landism: whoops, good catch

internal/tiltfile/flags/args_def.go
landism: missed those. added. thanks.

nicks:
re: "DB has talked to users who say this is what they want."

We need to break this dynamic of "it's done this way because @dbentley says so". The normal, healthy feedback loops of "let's go back and check on whether users are doing what we expect them to do" turn into "@dbentley feels attacked"

I don't think we can move forward with this feature until we can at least have a written, agreed-upon set of facts on what led us to design the feature this way. Even if we disagree on the design, we should at least agree on the facts that led us here. In particular: what users said, what workflows they have in mind, and how we plan on validating this.

internal/engine/upper.go
landism: Oh, I was unaware of this. I've been calling `time.Now` willy-nilly. Good to know.

landism:
> how would i wipe out previous flag values and start from a clean slate?

A later PR will add a `--clear` option

landism:
starting a new effort via #2653

PR #2598, opened 2019-11-25
build: add skips_local_docker=True to custom_build, for custom_builds that don't use docker at all

nicks:
Hello @jazzdan, @maiamcc,

Please review the following commits I made in branch nicks/ch3983/push:

3edcfc75d7a71cf1653623a336ac5927946cf611 (2019-11-25 17:54:51 -0500)
build: add skips_local_docker=True to custom_build, for custom_builds that don't use docker at all

### DIFF
--- a/pkg/model/image_target.go
+++ b/pkg/model/image_target.go
@@ -277,9 +277,10 @@ type CustomBuild struct {
 	// export $EXPECTED_REF=name:expected_tag )
 	Tag string
 
-	Fast        FastBuild
-	LiveUpdate  LiveUpdate // Optionally, can use LiveUpdate to update this build in place.
-	DisablePush bool
+	Fast             FastBuild
+	LiveUpdate       LiveUpdate // Optionally, can use LiveUpdate to update this build in place.
+	DisablePush      bool
+	SkipsLocalDocker bool
 }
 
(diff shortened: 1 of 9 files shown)


pkg/model/image_target.go
```diff
@@ -277,9 +277,10 @@ type CustomBuild struct {
 	// export $EXPECTED_REF=name:expected_tag )
 	Tag string
 
-	Fast        FastBuild
-	LiveUpdate  LiveUpdate // Optionally, can use LiveUpdate to update this build in place.
-	DisablePush bool
+	Fast             FastBuild
+	LiveUpdate       LiveUpdate // Optionally, can use LiveUpdate to update this build in place.
+	DisablePush      bool
```
jazzdan: Hmm it seems like `DisablePush` is still on this struct but it has been removed as a parameter in the Tiltfile. Was that intentional?

nicks: what makes you say it's been removed?

jazzdan: Oh, I thought the changes in `CustomBuilder` were in the Tiltfile, nevermind.

nicks: ya, fortunately we have lots of good tests for the other places it's used :smile:

PR #2600, opened 2019-11-26
hud: fix a panic when the websocket breaks

nicks:
Hello @jazzdan, @maiamcc,

Please review the following commits I made in branch nicks/issue2599/panic:

9b8f8ca2cea9cefe49e59c69188f969757eaadb2 (2019-11-25 19:06:48 -0500)
hud: fix a panic when the websocket breaks


internal/hud/server/websocket.go
```diff
@@ -86,6 +86,7 @@ func (ws WebsocketSubscriber) OnChange(ctx context.Context, s store.RStore) {
 	w, err := ws.conn.NextWriter(websocket.TextMessage)
 	if err != nil {
 		logger.Get(ctx).Verbosef("getting writer: %v", err)
+		return
```
maiamcc: 😅 good catch!

PR #2604, opened 2019-11-26
submodule: update protocc commit

maiamcc:
i don't actually know how submodules work but this got rid of the
constant "dirty" indicator on my master branch when nothing else
would, so...?


nicks:
this re-adds the submodule? that doesn't seem like a good idea

maiamcc:
😬 🤷‍♀ I believe you, and my dirty indicator on master has gone away, soooo... i'll just uh, delete this then...

nicks:
hahaha, submodule is dark magic, maybe the indicator was wrong? happy to pair on it if you see it again

PR #2606, opened 2019-11-26
tiltfile: support helm 3 in addition to helm 2

jazzdan:


internal/tiltfile/files.go
```diff
@@ -169,6 +185,8 @@ func (s *tiltfileState) helm(thread *starlark.Thread, fn *starlark.Builtin, args
 		cmd = append(cmd, "--set", setArg)
 	}
 
+	s.logger.Infof("Running: %s", cmd)
```
jazzdan: Intentionally left here: I think it's useful to see the resulting helm command.

internal/tiltfile/tiltfile_test.go
```diff
@@ -1804,7 +1804,9 @@ func TestYamlErrorFromHelm(t *testing.T) {
 	f.file("Tiltfile", `
 k8s_yaml(helm('helm'))
 `)
-	f.loadErrString("from helm")
+
+	// TODO(dmiller): there should be a better assertion here
+	f.loadErrString("helm")
```
jazzdan: Yeah this test is silly and asserts on the output of `helm`. Any ideas?

Makefile
```diff
@@ -61,6 +61,16 @@ else
 		gotestsum --format standard-quiet --junitfile test-results/unit-tests.xml -- ./... -p $(GO_PARALLEL_JOBS) -timeout 80s
 endif
 
+test-go-helm-only:
+ifneq ($(CIRCLECI),true)
+		go test -p $(GO_PARALLEL_JOBS) -timeout 80s ./internal/tiltfile -run "(?i)(.*)Helm(.*)"
```
jazzdan: How bananas is this? Any other ideas?

nicks: tbh, we should probably move Helm and its tests to its own plugin, and then a lot of this weirdness would go away

internal/tiltfile/tiltfile_test.go
nicks: i don't think that's silly. an easy mistake to make is to accidentally swallow the output from the tool

have you considered using the getHelmVersion function, and have different asserts per version?

vendor/github.com/golang/protobuf/jsonpb/jsonpb.go
```diff
@@ -818,7 +818,7 @@ func (u *Unmarshaler) unmarshalValue(target reflect.Value, inputValue json.RawMe
 
 			t, err := time.Parse(time.RFC3339Nano, unq)
 			if err != nil {
-				return fmt.Errorf("bad Timestamp: %v (%+v)", err, prop)
+				return fmt.Errorf("bad Timestamp: %v", err)
```
nicks: was this intentional?

jazzdan: I ran a `dev ensure` when I thought I was going to be adding a dependency, but then I ended up going a different direction and removed the dep. Proto was apparently updated with a `dev ensure` so I left it in (this seems to be the only change?) but I'm just as happy to remove it.

internal/tiltfile/tiltfile_test.go
jazzdan: That's a good idea, I'll do that!

Issue #2597, opened 2019-11-25
High CPU Usage


osener:
I'm getting started Tilt and bumped into a high CPU usage scenario I can reproduce using [Gloo helm chart](https://docs.solo.io/gloo/latest/installation/gateway/kubernetes/#installing-on-kubernetes-with-helm). This consistently gets Tilt to use at least 40% CPU on my 8 core laptop. I'm using Tilt v0.10.21 with [k3d](https://github.com/rancher/k3d) on Manjaro Linux.

I'm new to k8s so please let me know if you need more details, but I'm hoping these steps will help you reproduce the issue:

Tiltfile:
```python
k8s_yaml(helm('gloo', name='gloo-gateway', namespace='gloo-system'))
k8s_yaml(helm('gloo', name='gloo-knative', namespace='gloo-system', values=['./gloo/values-knative.yaml']))
```

```sh
$ helm repo add gloo https://storage.googleapis.com/solo-public-helm
$ helm fetch gloo/gloo
$ k3d create cluster --image rancher/k3s:v0.10.2-amd64
$ sleep 10
$ export KUBECONFIG="$(k3d get-kubeconfig --name='k3s-default')"
$ tilt up
```

In the first run, this chart seems to log _a lot_ of diagnostic errors, but even if I stop tilt & run `tilt up` again it quickly goes back to using a lot of resources even though I don't see much output on the console.

nicks:
Thanks for the report! This sounds pretty bad, @maiamcc are you able to repro?

maiamcc:
Oh eek, yep!
![Screen Shot 2019-11-25 at 5 21 06 PM](https://user-images.githubusercontent.com/6332648/69583160-3f59e600-0fa8-11ea-9085-1350221a3380.png)

I had trouble getting k3d running but your Tiltfile+chart gives me plenty of Tilt-eating-my-CPU on K8s for Docker for Mac, I'll start looking into it over here.

Just to verify I'm not debugging the wrong problem, can you profile your Tilt run for me? [See profiling instructions here](https://github.com/windmilleng/tilt/blob/master/CONTRIBUTING.md#go-profile), and then share the `tilt.profile` file you create. Thanks!

osener:
That seems to be along the lines of what I saw. I’ll gather some profiling data tomorrow when I’m back on my laptop, thanks for the quick response!

maiamcc:
[...]

PR #2607, opened 2019-11-26
hud: if hud not running, don't generate the whole view

maiamcc:
this fixes a small amt of the perf problems in #2597, tho not the bulk of them.
unsure if it's worth the complexity, figured I'd toss up a PR. thoughts?

### DIFF
--- a/internal/hud/hud.go
+++ b/internal/hud/hud.go
@@ -306,22 +306,6 @@ func (h *Hud) setView(ctx context.Context, view view.View) error {
 	return h.refresh(ctx)
 }
 
-// setViewLogOnly prints the new logs to stdout. (Should only be invoked if !h.isRunning)
-func (h *Hud) setViewLogOnly(ctx context.Context, log model.Log) error {
-	if h.isRunning {
-		return fmt.Errorf("cannot call setViewLogOnly if hud is running")
-	}
-
-	logLen := log.Len()
-	if !h.isRunning && h.currentViewState.ProcessedLogByteCount < logLen {
-		fmt.Print(log.String()[h.currentViewState.ProcessedLogByteCount:])
-	}
-
-	h.currentViewState.ProcessedLogByteCount = logLen
-
-	return h.refresh(ctx)
-}
-
 // Must hold the lock
 func (h *Hud) setViewState(ctx context.Context, currentViewState view.ViewState) error {
 	h.currentViewState = currentViewState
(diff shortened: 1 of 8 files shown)


nicks:
can you give a bit more detail on what this addresses?

The `isRunning` field is _supposed_ to be for handling special cases where the HUD is in the middle of being set up or torn down. A frequent mistake we've made in the past is "oops! we hit a fatal error, then screwed up the teardown codepath, so that the user can't even see the error".

If `isRunning` is false when Tilt is in steady-state, that would surprise me, and is something we should dig into more.

nicks:
(admittedly, `isRunning` is not a great name for this field)

maiamcc:
Oh sorry, this is for cases when `--hud=false` -- `hud.OnChange` is eating up a non-trivial amount of CPU even when the hud is disabled via flag. This attempts to fix that by not doing the whole view conversion if the hud is disabled. That said, idk if this is the right way to go about it or if the CPU savings are worth the complexity introduced. thoughts?

nicks:
can we make it so that the HUD doesn't get installed at all if it's not running? or just make OnChange a no-op? i don't totally understand why the HUD needs to do any work at all in this case.

maiamcc:
It's because we smooshed the responsibility for "if we're not doing any rty stuff, print all the logs since the last change to the terminal" into the hud, which isn't great.

```
logLen := log.Len()
if !h.isRunning && h.currentViewState.ProcessedLogByteCount < logLen {
	fmt.Print(log.String()[h.currentViewState.ProcessedLogByteCount:])
}

h.currentViewState.ProcessedLogByteCount = logLen
```

Idk if this properly ought to be a separate subscriber, or what? I can certainly split it out into its own subscriber (or its own implementation of the `Hud` interface??)

maiamcc:
Looking through, it seemed easier to just do the rearch. now -- wdty?

internal/hud/hud.go
```diff
@@ -293,6 +306,22 @@ func (h *Hud) setView(ctx context.Context, view view.View) error {
 	return h.refresh(ctx)
 }
 
+// setViewLogOnly prints the new logs to stdout. (Should only be invoked if !h.isRunning)
+func (h *Hud) setViewLogOnly(ctx context.Context, log model.Log) error {
```
nicks: i think this function is dead now?

maiamcc: good catch!

maiamcc:
hudfix before and after

Well this isn't the _most_ useful, but it did something, at least...

PR #2610, opened 2019-12-02
web: don't set selection change event handler if snapshot

jazzdan:


jazzdan:
I tried to write a test for this that looked like:

1. mount the component with `isSnapshot` set to false
2. assert the the event listener _isn't_ set

But it seems like you can't check the window and see if an event listener is set or not. All you can do is set or unset an event listener.

Open to any ideas though!

web/src/LogPane.tsx
```diff
@@ -52,11 +52,16 @@ class LogPane extends Component<LogPaneProps, LogPaneState> {
...
+
+    if (!this.props.isSnapshot) {
+      window.addEventListener("scroll", this.refreshAutoScroll, {
+        passive: true,
+      })
+      document.addEventListener("selectionchange", this.handleSelectionChange, {
+        passive: true,
+      })
+    }
     window.addEventListener("wheel", this.handleWheel, { passive: true })
```
nicks: i think we can skip this one too in snapshots?

and disable autoscroll

jazzdan:
For posterity: I figured out a way to test this

PR #2617, opened 2019-12-03
logstore: fork model.Log into logstore.LogStore

nicks:
Hello @jazzdan,

Please review the following commits I made in branch nicks/logs2:

071eb7d657c0d41296e61445326ad4ab02d77b47 (2019-12-03 13:52:08 -0500)
logstore: fork model.Log into logstore.LogStore
The main difference between them is that model.Log stores the logs
with prefixes, and logstore.LogStore stores the logs without prefixes,
then formats them with prefixes on String()

### DIFF
--- a/pkg/model/logstore/doc.go
+++ b/pkg/model/logstore/doc.go
@@ -1,0 +1,5 @@
+// A central logstore.
+//
+// Motivation and spec:
+// https://github.com/windmilleng/tilt.specs/pull/19
+package logstore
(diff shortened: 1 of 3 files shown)


nicks:
this architecture is described in https://github.com/windmilleng/tilt.specs/pull/19

i originally wanted this to replace model.Log in-place, but there were enough little api differences that i wasn't confident i could do it correctly.

PR #2618, opened 2019-12-03
tiltfile: address comments from #2578

landism:
addressing (hopefully) Maia's post-merge comments on #2578


internal/tiltfile/config/args_def.go
```diff
@@ -1,4 +1,4 @@
-package flags
+package config
```
maiamcc: 🙌

PR #2620, opened 2019-12-03
build: execute custom build cmd in Tiltfile directory

### LINKED ISSUE #2616: When using include, Tiltfile working directory should not be parent's
As reported in [Kubernetes > Tilt Channel](https://kubernetes.slack.com/archives/CESBL84MV/p1575385873369500): 

> when using include , any commands run for custom builds in the imported Tiltfiles appear to be run with the same working directory as the parent.
> this is making composition with makefile driven workflows a little tricky.
> is there a way to change the working directory to match the Tiltfile  containing the command?

jazzdan:
Fixes #2616

### DIFF
--- a/internal/build/custom_builder.go
+++ b/internal/build/custom_builder.go
@@ -16,7 +16,7 @@ import (
 )
 
 type CustomBuilder interface {
-	Build(ctx context.Context, ref reference.Named, command string, expectedTag string, skipsLocalDocker bool) (reference.NamedTagged, error)
+	Build(ctx context.Context, ref reference.Named, tiltfilePath string, command string, expectedTag string, skipsLocalDocker bool) (reference.NamedTagged, error)
 }
 
 type ExecCustomBuilder struct {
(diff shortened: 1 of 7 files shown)


internal/build/custom_builder.go
```diff
@@ -16,7 +16,7 @@ import (
 )
 
 type CustomBuilder interface {
-	Build(ctx context.Context, ref reference.Named, command string, expectedTag string, skipsLocalDocker bool) (reference.NamedTagged, error)
+	Build(ctx context.Context, ref reference.Named, tiltfilePath string, command string, expectedTag string, skipsLocalDocker bool) (reference.NamedTagged, error)
```
nicks: i wonder if we should pass in the whole CustomBuild model? it seems weird to break it apart

pkg/model/image_target.go
```diff
@@ -267,7 +267,8 @@ func (FastBuild) buildDetails()  {}
 func (fb FastBuild) Empty() bool { return reflect.DeepEqual(fb, FastBuild{}) }
 
 type CustomBuild struct {
-	Command string
+	TiltfilePath string
```
nicks: can we call this workdir or something? To me, "TiltfilePath" means "/path/to/dir/Tilfile", but this is actually "/path/to/dir". And workdir matches the name in local_resource

PR #2621, opened 2019-12-03
engine: docker-compose log actions should not contain timestamps or service prefixes

nicks:
Hello @maiamcc,

47aca0da54791dbfa8a14436b4ca8cce5ca7dcbb (2019-12-03 18:26:04 -0500)
engine: docker-compose log actions should not contain timestamps or service prefixes
Per https://github.com/windmilleng/tilt.specs/pull/19, we're moving towards a system
where the engine stores structured logs, and adds the prefixes later.


nicks:
yup! exactly, you got it

PR #2627, opened 2019-12-04
logstore: add a thread-safe log reader

nicks:
Hello @jazzdan,

Please review the following commits I made in branch nicks/logstoremutex:

7842dd295c5430cbf9200d51c9e60241de3b4628 (2019-12-04 13:02:54 -0500)
logstore: add a thread-safe log reader


nicks:
also, the tests for this helped me find a bug where ContinuingString() was mutating the original span map, so i fixed that while i was here

PR #2630, opened 2019-12-05
web: upgrade to typescript 3.7, create-react-app 3.3.0

jazzdan:
This is all to enable us to use [optional chaining](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#optional-chaining) and [nullish coalescing](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#nullish-coalescing), the value of which I tried to demonstrate with some changes to HUD.tsx


web/src/SidebarIcon.test.tsx
```diff
@@ -6,9 +6,12 @@ import { Color } from "./constants"
 
 type Ignore = boolean
 
-const cases: Array<
-  [string, ResourceStatus, Color | Ignore, IconType | Ignore]
-> = [
+const cases: Array<[
```
jazzdan: In order to use the features I wanted from TypeScript 3.7 I had to upgrade prettier, which caused this formatting change.

web/src/HUD.tsx
```diff
@@ -207,16 +202,16 @@ class HUD extends Component<HudProps, HudState> {
   render() {
     let view = this.state.view
 
-    let needsNudge = view ? view.needsAnalyticsNudge : false
-    let resources = (view && view.resources) || []
+    let needsNudge = view?.needsAnalyticsNudge
+    let resources = view?.resources
```
nicks: are you sure this is right? i think you still need the `|| []` if resources is null

jazzdan: I agree that this isn't a 1 to 1 translation of this behavior, however according to the type definitions the resources field can't be null, and it worked fine in my testing locally.

jazzdan: Oh wait, I misunderstood this feature. if `view` is `undefined|null` `view?.` returns undefined. So yes we still need to handle that. Thanks for second guessing me!

PR #2636, opened 2019-12-06
web: add a client-side logstore

nicks:
Hello @jazzdan, @hyu,

Please review the following commits I made in branch nicks/clientlogs:

c8bf2935da179bfa38f0f7dae2564f4ffcf2e7ec (2019-12-06 14:31:52 -0500)
web: add a client-side logstore
currently, this is a straight port of the server-side logstore

over time, we expect this logstore to have much richer methods
for filtering logs by manifest, log level, etc

### DIFF
--- a/web/src/LogStore.ts
+++ b/web/src/LogStore.ts
@@ -1,0 +1,191 @@
+// Client-side log store, which helps client side rendering and filtering of logs.
+//
+// Loosely adapted from the data structures in
+// pkg/model/logstore/logstore.go
+// but with better support for incremental updates and rendering.
+
+type LogSpan = {
+  manifestName: string
+  lastSegmentIndex: number
+  firstSegmentIndex: number
+}
+
+class LogSegment {
+  spanId: string
+  time: string
+  text: string
+  continuesLine: boolean
+
+  constructor(seg: Proto.webviewLogSegment) {
+    this.spanId = seg.spanId ?? ""
+    this.time = seg.time ?? ""
+    this.text = seg.text ?? ""
+    this.continuesLine = false
+  }
+
+  startsLine() {
+    return !this.continuesLine
+  }
+
+  isComplete() {
+    return this.text[this.text.length - 1] == "\n"
+  }
+}
+
+class LogStore {
+  spans: { [key: string]: LogSpan }
+  segments: LogSegment[]
+
+  constructor() {
+    this.spans = {}
+    this.segments = []
+  }
+
+  append(logList: Proto.webviewLogList) {
+    let newSpans = logList.spans as { [key: string]: Proto.webviewLogSpan }
+    for (let key in newSpans) {
+      let exists = this.spans[key]
+      if (!exists) {
+        this.spans[key] = {
+          manifestName: newSpans[key].manifestName ?? "",
+          firstSegmentIndex: -1,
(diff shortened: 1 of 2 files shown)


web/src/LogStore.ts
```diff
@@ -0,0 +1,191 @@
+// Client-side log store, which helps client side rendering and filtering of logs.
+//
+// Looslely adapted from the data structures in
```
jazzdan:
```suggestion
// Loosely adapted from the data structures in
```

web/src/LogStore.ts
```diff
@@ -0,0 +1,191 @@
...
+}
+
+class LogSegment {
+  spanId: string
+  time: string
+  text: string
+  continuesLine: boolean
+
+  constructor(seg: Proto.webviewLogSegment) {
+    this.spanId = seg.spanId || ""
```
jazzdan: We can use `??` here now
```suggestion
    this.spanId = seg.spanId ?? ""
```

ditto on the ones below

web/src/LogStore.ts
```diff
@@ -0,0 +1,191 @@
...
+  spans: { [key: string]: LogSpan }
+  segments: LogSegment[]
+
+  constructor() {
+    this.spans = {}
+    this.segments = []
+  }
+
+  append(logList: Proto.webviewLogList) {
+    let newSpans = logList.spans as { [key: string]: Proto.webviewLogSpan }
```
jazzdan: Why do we cast this instead of handling the case where `logList.spans` isn't defined?

web/src/LogStore.ts
```diff
@@ -0,0 +1,191 @@
...
+        }
+      }
+    }
+
+    let segments = logList.segments ?? []
+    segments.forEach(segment => {
+      let newSegment = new LogSegment(segment)
+      let span = this.spans[newSegment.spanId]
+      if (!span) {
+        // If we don't have the span for this log, we can't meaninfully print it,
```
jazzdan:
```suggestion
        // If we don't have the span for this log, we can't meaningfully print it,
```

web/src/LogStore.ts
```diff
@@ -0,0 +1,191 @@
...
+    })
+  }
+
+  allLog() {
+    return this.manifestLog("")
+  }
+
+  manifestLog(mn: string) {
+    let lastLineCompleted = false
+    let allLogs = mn == ""
```
jazzdan: should we use `===` here and on the line below?

web/src/LogStore.ts
nicks: done

web/src/LogStore.ts
nicks: done

web/src/LogStore.ts
nicks: logList.spans is always defined. this isn't about runtime behavior. this is about compile-time behavior. The type definition is "object" and needs to be cast to the correct type. i don't know why the typescript definitions use "object"

web/src/LogStore.ts
nicks: done

web/src/LogStore.ts
nicks: done

PR #2647, opened 2019-12-09
logstore: prep for multiple spans per manifest

nicks:
Hello @jazzdan,

Please review the following commits I made in branch nicks/spans:

1c944efa1878319e7546e40794990b9cda804ed4 (2019-12-09 18:32:20 -0500)
logstore: prep for multiple spans per manifest

### DIFF
--- a/pkg/model/logstore/logstore.go
+++ b/pkg/model/logstore/logstore.go
@@ -84,17 +84,11 @@ type LogEvent interface {
 	Message() []byte
 	Time() time.Time
 
-	// Ideally, all logs should be associated with a source.
-	//
-	// In practice, not all logs have an obvious source identifier,
-	// so this might be empty.
-	//
-	// Right now, that source is a ManifestName. But in the future,
-	// this might make more sense as another kind of identifier (like SpanID).
-	//
-	// (As of this writing, we have TargetID as an abstract build-time
-	// source identifier, but no generic run-time source identifier)
-	Source() model.ManifestName
+	// The manifest that this log is associated with.
+	ManifestName() model.ManifestName
+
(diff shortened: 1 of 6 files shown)


pkg/model/logstore/logstore.go
```diff
@@ -371,6 +371,16 @@ func (s *LogStore) String() string {
 	return s.ManifestLog("")
 }
 
+func (s *LogStore) spansInManifest(mn model.ManifestName) map[SpanID]*Span {
```
jazzdan: This wording is confusing to me. I'd maybe rename it to `spansForManifest`?

web/src/LogStore.ts
```diff
@@ -106,13 +117,28 @@ class LogStore {
...
+      }
+
+      if (earliestStartIndex == -1) {
         return ""
       }
 
-      startIndex = span.firstSegmentIndex
-      lastIndex = span.lastSegmentIndex
+      startIndex = earliestStartIndex
+      lastIndex = latestEndIndex
```
jazzdan: /me resists make a snide web assembly comment

nicks: ?

pkg/model/logstore/logstore.go
nicks: done!

web/src/LogStore.ts
nicks: if the comment is "why don't we compile the go code to webassembly"...i think we _might_ go down that path eventually, but it's too early to say. these two data structures might end up diverging a lot because the needs of the UI are going to be different...

jazzdan: Oh it was a joke, sorry!

PR #2649, opened 2019-12-10
web: use structured logs to build a data model

nicks:
Hello @jazzdan, @hyu,

Please review the following commits I made in branch nicks/ch4097/datamodel:

1b060025abb0e5439fdabd6352b04788ba27d182 (2019-12-10 13:40:57 -0500)
web: use structured logs to build a data model

### DIFF
--- a/web/src/LogPane.tsx
+++ b/web/src/LogPane.tsx
@@ -3,13 +3,15 @@ import { ReactComponent as LogoWordmarkSvg } from "./assets/svg/logo-wordmark-gr
 import AnsiLine from "./AnsiLine"
 import "./LogPane.scss"
 import ReactDOM from "react-dom"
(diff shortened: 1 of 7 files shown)


web/src/LogPane.tsx
```diff
@@ -211,31 +247,26 @@ class LogPane extends Component<LogPaneProps, LogPaneState> {
       }
 
       let el = (
-        <span
+        <LogLineComponent
           ref={i === 0 ? this.highlightRef : null}
```
nicks: @jazzdan do you understand what this `i == 0` check is trying to accomplish? i spent some time poking at it and couldn't figure it out. it seems wrong

jazzdan: I agree, that seems wrong.

PR #2650, opened 2019-12-10
config: rename args and flags

landism:
The current code conflates args and flags, and really flags is a bad name anyway, so new names:

* flags and args are things that come from the command line
* tilt_config.json is the config file
* the config is the in-memory combination of args and the config file
* an individual configurable value in a config is a config setting


maiamcc:
discussed offline: "args" are the things that come from the command line, the "config" is the dict generated in the tiltfile (by merging args and the contents of the config file).

internal/tiltfile/config/config.go
```diff
@@ -10,9 +10,9 @@ import (
 
 type Settings struct {
 	enabledResources []model.ManifestName
-	argDef           ArgsDef
+	configDef        ConfigDef
 
-	flagsParsed bool
+	configParsed bool
```
maiamcc: i think this one SHOULD be `argsParsed`?

maiamcc: or rather: would you call `configs.parse` even if no args were passed? Hmm, I guess maybe yes...

internal/tiltfile/config/config_test.go
```diff
@@ -205,7 +205,7 @@ cfg = config.parse()
 
 	_, err := f.ExecFile("Tiltfile")
 	require.Error(t, err)
-	require.Equal(t, "positional args were specified, but none were expected (no arg defined with args=True)", err.Error())
+	require.Equal(t, "positional args were specified, but none were expected (no flag defined with args=True)", err.Error())
```
maiamcc: ...hm where would `args=True` be specified? I'm confused.

internal/tiltfile/config/config.go
landism: In general I don't expect Tiltfile to make the `config.parse` call dependent on whether there are flags (and I don't think it's even non-trivial to write such a Tiltfile) - so, yes, with two sample use cases:
* you want to enable the "`tilt up` with no args starts all services", but with `config.parse` and `config.set_enabled_resources`. When there are no args, you still call flags.parse and get an empty array. (or maybe it's fairer to say `args` always has a value, but sometimes it's an empty array?)
* (not in master) you have a tilt_config.json and want the values from that

An explanation of this bool's purpose:
1. If a Tiltfile does not call `config.parse`, the args are used to select which resources to start.
2. If a Tiltfile does call `config.parse`, then we assume the Tiltfile does useful things with the args, and Tilt does not do anything with the args other than make sure they're an input to the handling of `config.parse` (with which the Tiltfile author might possibly then turn around and call `set_enabled_resources`)

Maybe `configParseCalled` would be clearer?

internal/tiltfile/config/config_test.go
landism: This is an argument to `config.define_string_list`, e.g. https://github.com/windmilleng/tilt.build/pull/325/files#diff-0f16fdfc562a71a859b08af2266710ffR33

PR #2651, opened 2019-12-10
web: fix selection behavior by using margin instead of padding

nicks:
Hello @hyu,

Please review the following commits I made in branch nicks/ch4102/webui:

4aebdf9f2c5ac94e1362ced1e349357a90c9a7ec (2019-12-10 15:42:21 -0500)
web: fix selection behavior by using margin instead of padding
Fixes https://github.com/windmilleng/tilt/issues/2645

### DIFF
--- a/web/src/LogPane.scss
+++ b/web/src/LogPane.scss
@@ -1,7 +1,7 @@
 @import "constants.scss";
 
 .LogPane {
-  padding: $spacing-unit / 2 0;
+  margin: $spacing-unit / 2 0;
   width: 100%;
 }
 
@@ -20,7 +20,7 @@
 .logEnd {
   animation: blink 1s infinite;
   animation-timing-function: ease;
-  padding-left: $spacing-unit / 2;
+  margin-left: $spacing-unit / 2;
 }
 
 .LogPane-empty {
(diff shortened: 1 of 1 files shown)


nicks:
do you have any insight into how this works???

hyu:
Haha, this is a mess right now. So, the ideal highlight behavior is...

1. There's one highlight that shows the lines associated with whatever you've selected.
2. There's another highlight that shows precisely what you've selected.

The idea behind (1) is that when you share a Snapshot with a highlight, the minimum unit of the highlight isn't chars, it's lines. Ex: if I share some lines with you on GitHub you see a neat [yellow highlight](https://github.com/windmilleng/tilt/blob/master/internal/cli/analytics.go#L38-L40)

But, we have (2) because you still want to see what your selection is, so you can copy the precise bit that you want. Ex: you can select and copy the word `initAnalytics` within the [yellow highlight](https://github.com/windmilleng/tilt/blob/master/internal/cli/analytics.go#L38-L40)

Sadly, the behavior for (1) broke during Matt's attempt to fix some of Lyft's log copy-pasting issues (with Jira? I think line breaks were lost): [webui: use linebreaks instead of display: block to separate lines](https://github.com/windmilleng/tilt/pull/2536)

So, @nicks — your fix above makes things less ugly than they currently are, but ideally I'd like to restore the original behavior above.

nicks:
straightened this out offline! this is a chrome-only issue, so Han was confused

PR #2653, opened 2019-12-10
user config: also read from tilt_config.json

landism:
This is a scoped-down #2596. 

The behavior implemented by this PR is, when `config.parse` is called:
1. If tilt_config.json specifies any settings, load those.
2. If there are any args on the command line, parse those and have them take precedence over anything in tilt_config.json.
3. in starlark, return whatever config we're left with

Tilt does not write to tilt_config.json.

### DIFF
--- a/internal/tiltfile/config/config.go
+++ b/internal/tiltfile/config/config.go
@@ -1,26 +1,34 @@
 package config
 
 import (
+	"path/filepath"
+
 	"github.com/pkg/errors"
 	"go.starlark.net/starlark"
 
+	"github.com/windmilleng/tilt/internal/tiltfile/io"
 	"github.com/windmilleng/tilt/internal/tiltfile/starkit"
 	"github.com/windmilleng/tilt/pkg/model"
 )
+
+const UserConfigFileName = "tilt_config.json"
 
 type Settings struct {
 	enabledResources []model.ManifestName
 	configDef        ConfigDef
 
+	configPath string
+
 	configParseCalled bool
+	UserConfigState   model.UserConfigState
 }
 
 type Extension struct {
-	cmdLineArgs []string
+	UserConfigState model.UserConfigState
 }
 
-func NewExtension(args []string) *Extension {
-	return &Extension{cmdLineArgs: args}
+func NewExtension(userConfigState model.UserConfigState) *Extension {
+	return &Extension{UserConfigState: userConfigState}
 }
 
 func (e *Extension) NewState() interface{} {
@@ -30,6 +38,19 @@ func (e *Extension) NewState() interface{} {
 }
 
 var _ starkit.StatefulExtension = &Extension{}
+
+func (e *Extension) OnExec(t *starlark.Thread, path string) error {
+	dir := filepath.Dir(path)
+	configPath := filepath.Join(dir, UserConfigFileName)
+
+	return starkit.SetState(t, func(settings Settings) Settings {
+		settings.UserConfigState = e.UserConfigState
+		settings.configPath = configPath
+		return settings
+	})
+}
+
+var _ starkit.OnExecExtension = &Extension{}
 
 func MustState(model starkit.Model) Settings {
 	state, err := GetState(model)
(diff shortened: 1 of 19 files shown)


internal/tiltfile/starkit/fixture.go
```diff
@@ -39,6 +41,7 @@ func (f *Fixture) OnStart(e *Environment) error {
 
 	e.SetPrint(func(t *starlark.Thread, msg string) {
 		_, _ = fmt.Fprintf(f.out, "%s\n", msg)
+		_, _ = fmt.Printf("%s\n", msg)
```
nicks: ?

i'm not sure that all tests would want this?

internal/tiltfile/config/config.go
```diff
@@ -31,6 +39,19 @@ func (e *Extension) NewState() interface{} {
 
 var _ starkit.StatefulExtension = &Extension{}
 
+func (e *Extension) OnExec(t *starlark.Thread, path string) error {
+	dir := filepath.Dir(path)
```
nicks: so...every time you sub-include a Tiltfile, it creates a new tilt_config.json in that Tiltfile's directory and overwrites the current UserConfigState.

are you sure that's the behavior you want? can you add a test for this?

I would have expected this to be an OnStart extension that only affects the main Tiltfile load, OnExec extensions are pretty rare

landism: > are you sure that's the behavior you want?

Nope, I failed to account for loads/includes in this PR. Thanks. This interaction should be tested regardless of what the behavior is.

> I would have expected this to be an OnStart extension that only affects the main Tiltfile load, OnExec extensions are pretty rare

If we want Tiltfiles to be composable, I think we'll need this to be an OnExec extension so that we can get each config path separately (or maybe a thread-local that tells us the current tiltfile path), but yeah ok I'll find another way to get the path.

> so...every time you sub-include a Tiltfile, it creates a new tilt_config.json in that Tiltfile's directory and overwrites the current UserConfigState.

(FWIW, just to make sure we're on the same page, the intent is that Tilt is no longer creating tilt_config.json)

nicks: re: thread-local that tells us the current tiltfile path: https://github.com/windmilleng/tilt/blob/master/internal/tiltfile/starkit/path.go#L18

internal/tiltfile/starkit/fixture.go
landism: hm, I figured:
1. if the test passes, test runners will hide the output anyway
2. if the test fails, then the output is generally nice to have around
3. tests with Tiltfiles that generate an unmanageable amount of output are likely to be rare

I added this in reaction to a couple of failing tests where the information I needed to debug them was in the tiltfile output. Is there a cost I'm missing that offsets that benefit?

nicks: the upper tests are a good example of where this logic leads you - you end up with lots of unrelated output that obscures what failed.

nicks: i'm pretty sure we already have Tiltfiles that already print a ton, so (3) is already not true.

landism: > the upper tests are a good example of where this logic leads you - you end up with lots of unrelated output that obscures what failed.

Interesting. I feel like the upper tests are a good example of this logic paying off! In my experience, the relevant failure info is almost always at the end of the test output, where I'd expect it, and then more information is available above if I want it. (aside from that period where it was dumping a stack trace in a defer so it got printed after the failure). I guess we have different tastes.

> i'm pretty sure we already have Tiltfiles that already print a ton

Hmm, I was thinking unmanageable as in "kick the start of the test out of your scrollback buffer". and I haven't seen or conceived of any need to have tests that generate overbearing logs. Maybe I haven't been working on the fun parts of the tiltfile.

Anyway, this is unrelated to this purpose of this PR, so I'll take it out and consider revisiting it if I continue to find it annoying.

PR #2656, opened 2019-12-11
webview: remove logs from webclient build history and pod status

nicks:
Hello @jazzdan, @hyu,

Please review the following commits I made in branch nicks/deadcode2:

74558e54b0f7c304cf05337f3d0fde56ebaa08be (2019-12-11 11:51:40 -0500)
webview: remove logs from webclient build history and pod status
we don't even use these anyway!

we'll eventually move these to a system where we have a single log store, and a
method for "get all the logs from this pod", or "where does this error appear in
the log stream"


pkg/webview/view.proto
```diff
@@ -14,7 +14,8 @@ message BuildRecord {
   repeated string warnings = 3;
   google.protobuf.Timestamp start_time = 4;
   google.protobuf.Timestamp finish_time  = 5;
-  string log = 6;
+  reserved "log";
```
nicks: today i learned: this is the proto3 way to mark a field as deleted
https://developers.google.com/protocol-buffers/docs/proto3#reserved

PR #2657, opened 2019-12-11
engine: telemetry controller

jazzdan:


internal/engine/telemetrycontroller.go
```diff
@@ -0,0 +1,82 @@
...
+	"github.com/windmilleng/wmclient/pkg/dirs"
+
+	"github.com/windmilleng/tilt/internal/build"
+	"github.com/windmilleng/tilt/internal/engine/configs"
+	"github.com/windmilleng/tilt/internal/store"
+	"github.com/windmilleng/tilt/internal/tracer"
+	"github.com/windmilleng/tilt/pkg/model"
+)
+
+type TelemetryController struct {
```
nicks: all new subscribers (and their actions) should go in a subpackage of engine

internal/engine/telemetrycontroller.go
```diff
@@ -0,0 +1,82 @@
...
+func NewTelemetryController(lock tracer.Locker, clock build.Clock, dir *dirs.WindmillDir) *TelemetryController {
+	return &TelemetryController{
+		lock:  lock,
+		clock: clock,
+		dir:   dir,
+	}
+}
+
+func (t *TelemetryController) OnChange(ctx context.Context, st store.RStore) {
+	state := st.RLockState()
```
nicks: can you move this into a function so that other things in this method don't accidentally read `state`? see PodLogManager#diff for an example of this pattern

internal/engine/telemetrycontroller.go
```diff
@@ -0,0 +1,82 @@
...
+	// clear the file if the telemetry command succeeded
+	if cmdSucceeded {
+		if err = t.dir.WriteFile(tracer.OutgoingFilename, ""); err != nil {
+			logError(st, err)
+		}
+	}
+}
+
+func logError(st store.RStore, err error) {
+	st.Dispatch(configs.TiltfileLogAction{
```
nicks: TiltfileLogAction doesn't seem right, this isn't really about tiltfile execution? and if you sync your pr, this won't work anymore because now we have span ids.

internal/engine/telemetrycontroller.go
```diff
@@ -0,0 +1,82 @@
...
+		return
+	}
+	defer func() {
+		err := file.Close()
+		if err != nil {
+			logError(st, err)
+		}
+	}()
+
+	cmd.Stdin = file
```
nicks: where does the Stdout/Stderr of this command go?

i wonder if we'll need eventually need a new kind of manifest to handle all this

internal/tracer/otel.go
```diff
@@ -0,0 +1,33 @@
...
+	}
+
+	tp, err := sdktrace.NewProvider(
+		sdktrace.WithConfig(sdktrace.Config{DefaultSampler: sdktrace.AlwaysSample()}))
+	if err != nil {
+		return nil, err
+	}
+	tp.RegisterSpanProcessor(exporter)
+	global.SetTraceProvider(tp)
+	tracer = tp.Tracer("tilt.dev/usage") // global
```
nicks: why is this global?

afaict, it should be straightforward to dependency-inject the tracer. global state is going to fall apart pretty quickly

internal/tracer/exporter.go
```diff
@@ -0,0 +1,129 @@
...
+	spanDataCh chan *exporttrace.SpanData
+	stopCh     chan struct{}
+
+	// members only accessed by the loop() goroutine
+	queue      []*exporttrace.SpanData
+	needsFlush bool
+}
+
+// TODO(dbentley): write tests for this
+func newExporter(ctx context.Context, dir *dirs.WindmillDir) (*exporter, error) {
```
nicks: when would this return an error?

internal/tracer/exporter.go
```diff
@@ -0,0 +1,129 @@
...
+
+func (e *exporter) loop(ctx context.Context) {
+	var flushingCh chan struct{}
+	var timerCh <-chan time.Time
+	for {
+		select {
+		case sd := <-e.spanDataCh:
+			e.queue = append(e.queue, sd)
+			if timerCh == nil {
+				timerCh = time.NewTimer(5 * time.Second).C
```
nicks: this use of timer isn't quite right, see the comments on 
https://golang.org/pkg/time/#After

usually, you'd either keep the Timer around (and call Stop() when you don't need it anymore), or you'd use time.After()

have you considered keeping a single Timer outside of the loop that fires every 5 sec and does a flush()? this throttling algorithm feels a lot more complicated than it needs to be at this point.

internal/tracer/exporter.go
```diff
@@ -0,0 +1,129 @@
...
+	}
+}
+
+const maxFileSize = 32 * 1024 // 128 MiB
+
+func (e *exporter) flush(ctx context.Context, flushingCh chan struct{}) {
+	e.outgoingMu.Lock()
+	defer e.outgoingMu.Unlock()
+	defer close(flushingCh)
+	q := e.queue
```
nicks: This read of the queue isn't thread-safe, since the queue can be simultaneously modified by your control loop

have you considered having two mutexes: one on the queue and one on the outgoing file? i think that would be simpler to reason about than your sendData channel

internal/engine/telemetrycontroller.go
jazzdan: I think that's likely, but not something I wanted to do immediately.

The stdout is purposefully hidden to not distract end users with telemetry that they certainly don't care about. stderr should be surfaced if there's an error, though. Good call.

internal/engine/telemetrycontroller.go
jazzdan: I guess I need a new kind of LogAction?

internal/tracer/otel.go
dbentley: Just for similarity to opentracing, where we use a global tracer and haven't had an issue.

internal/tracer/exporter.go
dbentley: I think locks can be simpler in easy cases but get more complicated quickly, vs. using a loop in a goroutine a la https://blog.golang.org/share-memory-by-communicating .

internal/tracer/otel.go
nicks: ?

in opentracing, you usually inject the tracer with server middleware, see this example:
https://github.com/grpc-ecosystem/go-grpc-middleware/blob/master/tracing/opentracing/client_interceptors.go#L21
https://github.com/grpc-ecosystem/go-grpc-middleware/blob/master/tracing/opentracing/options.go#L51

it looks like opentracing is encouraging people to use global.TraceProvider().Trace() when you need a new tracer, see this example:
https://github.com/open-telemetry/opentelemetry-go/blob/master/example/grpc/middleware/tracing/tracing.go#L51

```
tr := global.TraceProvider().Tracer("example/grpc")
```

can you do that instead?

internal/tracer/exporter.go
nicks: ya, i was suggesting a possible fix to the race condition here. i don't care how you fix it.

I think you misunderstood that "Share memory by communicating" blog post. They're saying that you should pass the object to the channel, not "use channels to simulate mutually exclusive access to fields on a struct".

internal/tracer/exporter.go
```diff
@@ -0,0 +1,146 @@
...
+	exporttrace "go.opentelemetry.io/otel/sdk/export/trace"
+	sdktrace "go.opentelemetry.io/otel/sdk/trace"
+)
+
+// Exporter does 3 things:
+// 1) Accepts spans from OpenTelemetry.
+// 2) Stores spans (for now, in memory)
+// 3) Allows consumers to read spans they might want to send elsewhere
+// Numbers 2 and 3 access the same data, and so it's a concurrency issue.
+type Exporter struct {
```
nicks: I might call this SpanCollector or something now...it doesn't really do any exporting

internal/tracer/exporter.go
```diff
@@ -0,0 +1,146 @@
...
+			// add to the queue
+			queue = appendAndTrim(queue, sd)
+		case <-e.readReqCh:
+			// send the queue to the reader
+			if len(queue) == 0 {
+				// There's nothing pending, so send a nil reader and a channel the client can write to (because it's buffered with length 1)
+				e.readRespCh <- readResp{doneCh: make(chan bool, 1)}
+				break
+			}
+			r, err := e.makeReader(queue)
```
nicks: not a big deal, but i'm surprised that the serialization logic live inside the collector. Most APIs would push that out to a separate layer. And then a lot of this serialization/error-handling wouldn't block the collector loop

internal/tracer/exporter.go
```diff
@@ -0,0 +1,146 @@
...
+				// oh wait, there's a problem encoding?
+				// tell the reader, and then clear the queue
+				e.readRespCh <- readResp{err: err}
+				queue = nil
+				break
+			}
+
+			// there's now a read in flight
+			readDoneCh = make(chan bool)
+			beingRead, queue = queue, nil
```
nicks: this seems pretty brittle, and will break badly if two readers try to access this API at the same time.

have you considered having the Done channel read back the span that failed to send?

internal/tracer/exporter.go
```diff
@@ -0,0 +1,146 @@
...
+	"encoding/json"
+	"fmt"
+	"io"
+	"strings"
+
+	exporttrace "go.opentelemetry.io/otel/sdk/export/trace"
+	sdktrace "go.opentelemetry.io/otel/sdk/trace"
+)
+
+// Exporter does 3 things:
```
nicks: feel free to tell me to F off, but have you considered using BatchSpanProcessor?
https://github.com/open-telemetry/opentelemetry-go/blob/1ab645fedb49f300145c671bdab3fe57564bf45a/sdk/trace/batch_span_processor.go

most of the batching this class does seems pretty generalizable, and not Tilt-specific

internal/tracer/otel.go
```diff
@@ -0,0 +1,23 @@
...
+func InitOpenTelemetry(ctx context.Context, exporter sdktrace.SpanProcessor) (apitrace.Tracer, error) {
+	tp, err := sdktrace.NewProvider(
+		sdktrace.WithConfig(sdktrace.Config{DefaultSampler: sdktrace.AlwaysSample()}))
+	if err != nil {
+		return nil, err
+	}
+	if exporter != nil {
+		tp.RegisterSpanProcessor(exporter)
+	}
+	tracer = tp.Tracer("tilt.dev/usage")
```
nicks: is this a dead assign now?

nicks:
lots of good progress! i made some minor comments where we might be able to simplify things, the only blocking one is the one about the engine state

internal/tracer/exporter.go
dbentley: Good idea; I'll move it around in a bit.

internal/tracer/exporter.go
dbentley: Yup, great idea. I'll go do that after my meetings.

PR #2666, opened 2019-12-12
webview: Send down the span ids for runtime logs and build logs

nicks:
Hello @jazzdan,

Please review the following commits I made in branch nicks/spanlog2:

94e5922f96e3356200feddea346e92e3cc44dbbf (2019-12-12 13:06:14 -0500)
webview: Send down the span ids for runtime logs and build logs


nicks:
this is not as big as the diff makes it look! most of it is the protobuf changes

nicks:
(which i'm happy to move to a separate pr if that's easier)

PR #2668, opened 2019-12-12
web: standardize on T[] instead of Array<T>

jazzdan:
I thought we should standardize on one way of notating arrays, and this way uses fewer characters
and more closely aligns with how slices are annotated in Go.

I ran:

```
jscodeshift -t ~/development/typescript-generic-array-to-square-brackets/transform.js --extensions=ts --parser=ts src/
```

The code for the jscodeshift transformer lives [here](https://github.com/jazzdan/typescript-generic-array-to-square-brackets).


nicks:
seems fine to me, wdyt Han?

PR #2671, opened 2019-12-12
tilt: add args subcommand

landism:
This also works on tilt instances that are not using `config` at all.

I've found the help kind of awkward to write and am super interested in advice on making it less awkward. I wrote up some of my thinking on naming issues [here](https://docs.google.com/document/d/1JXoIR2UN4zKMBdcSZiPd7AzdQM7CVOiEL453BQtI8a0/edit?usp=sharing).

### DIFF
--- a/internal/cli/args.go
+++ b/internal/cli/args.go
@@ -1,0 +1,80 @@
+package cli
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+)
+
+type argsCmd struct {
+	webPort int
+	clear   bool
+	post    httpPoster
+}
+
+func newArgsCmd() *argsCmd {
+	return &argsCmd{post: http.Post}
+}
+
+func (c *argsCmd) register() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "args [<tiltfile args>]",
+		Short: "Changes the Tiltfile args in use by a running Tilt",
+		Long: `Changes the Tiltfile args in use by a running Tilt.
+
+Note that this does not affect built-in Tilt args (e.g. --hud, --host), but rather the extra args that come after,
+i.e., those specifying which resources to run and/or handled by a Tiltfile calling config.parse.
+
+To provide args starting with --, insert a standalone --, e.g.:
+
+tilt args -- --foo=bar frontend backend
+`,
+	}
+
+	cmd.Flags().IntVar(&c.webPort, "port", DefaultWebPort, "Web port for the Tilt whose args should change")
+	cmd.Flags().BoolVar(&c.clear, "clear", false, "Clear the Tiltfile args, as if you'd run tilt with no args")
+
+	return cmd
+}
+
+type httpPoster func(url string, contentType string, body io.Reader) (*http.Response, error)
+
+func (c *argsCmd) run(ctx context.Context, args []string) error {
+	// require --clear instead of an empty args list to ensure an experimental `tilt flags` doesn't unintentionally wipe state
+	if len(args) == 0 {
+		if !c.clear {
+			return errors.New("no args specified. If your intent is to empty the args, run `tilt args --clear`.")
+		}
+	} else {
+		if c.clear {
+			return errors.New("--clear cannot be specified with other values. either use --clear to clear the args or specify args to replace the args with a new (non-empty) value")
+		}
+	}
+	url := fmt.Sprintf("http://localhost:%d/api/set_tiltfile_args", c.webPort)
+	body := &bytes.Buffer{}
+	err := json.NewEncoder(body).Encode(args)
+	if err != nil {
+		return errors.Wrap(err, "failed to encode args as json")
+	}
+	res, err := c.post(url, "application/json", body)
+	if err != nil {
+		return errors.Wrapf(err, "could not connect to Tilt at %s", url)
+	}
+	defer func() {
+		_ = res.Body.Close()
+	}()
+
+	if res.StatusCode != http.StatusOK {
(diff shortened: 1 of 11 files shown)


internal/cli/args.go
```diff
@@ -0,0 +1,80 @@
...
+	post    httpPoster
+}
+
+func newArgsCmd() *argsCmd {
+	return &argsCmd{post: http.Post}
+}
+
+func (c *argsCmd) register() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "args [<tiltfile args>]",
```
maiamcc: not a big deal but could consider making the command `tfargs` to be even clearer about what it does/doesn't allow you to set

internal/cli/args.go
```diff
@@ -0,0 +1,80 @@
...
+		Short: "Changes the Tiltfile args in use by a running Tilt",
```
maiamcc: maybe slightly more correct to say "for a running Tilt _instance_"? (I... think?) if you change it here, also change in the description for "port" below

internal/cli/args.go
```diff
@@ -0,0 +1,80 @@
...
+	cmd.Flags().BoolVar(&c.clear, "clear", false, "Clear the Tiltfile args, as if you'd run tilt with no args")
+
+	return cmd
+}
+
+type httpPoster func(url string, contentType string, body io.Reader) (*http.Response, error)
+
+func (c *argsCmd) run(ctx context.Context, args []string) error {
+	// require --clear instead of an empty args list to ensure an experimental `tilt flags` doesn't unintentionally wipe state
+	if len(args) == 0 {
```
maiamcc: nice!!!

internal/cli/args.go
```diff
@@ -0,0 +1,80 @@
...
+		if !c.clear {
+			return errors.New("no args specified. If your intent is to empty the args, run `tilt args --clear`.")
+		}
+	} else {
+		if c.clear {
+			return errors.New("--clear cannot be specified with other values. either use --clear to clear the args or specify args to replace the args with a new (non-empty) value")
+		}
+	}
+	url := fmt.Sprintf("http://localhost:%d/api/set_tiltfile_args", c.webPort)
```
maiamcc: ah so this is going through a request to the web UI -- does that mean that someone running Tilt with `--port=0` can't use `tilt args`?

internal/cli/args.go
```diff
@@ -0,0 +1,80 @@
...
+	res, err := c.post(url, "application/json", body)
+	if err != nil {
+		return errors.Wrapf(err, "could not connect to Tilt at %s", url)
+	}
+	defer func() {
+		_ = res.Body.Close()
+	}()
+
+	if res.StatusCode != http.StatusOK {
+		return fmt.Errorf("error connecting to Tilt at %s: %d", url, res.StatusCode)
```
maiamcc: maybe include the status result body in case it has useful info?

internal/cli/up.go
```diff
@@ -50,8 +50,18 @@ type upCmd struct {
 
 func (c *upCmd) register() *cobra.Command {
 	cmd := &cobra.Command{
-		Use:   "up [<resource_name1>] [<resource_name2>] [...]",
-		Short: "stand up one or more resources (if no resource names specified, stand up all known resources specified in Tiltfile)",
+		Use:   "up [<Tiltfile args>]",
```
maiamcc: like we discussed, maybe we want to tell EVERYONE to separate tilt args from tiltfile args with a `--` even if not strictly necessary, to help ppl grok the division between the two types of args?
```
use: "up [<tilt built-in args>] -- [<Tiltfile Args>]"
```

internal/cli/up.go
```diff
@@ -50,8 +50,18 @@ type upCmd struct {
...
+By default:
+1) Tiltfile args are interpreted as the list of services to start, e.g. tilt up frontend backend.
+2) Running with no args starts all services defined in the Tiltfile
+
+This default behavior does not apply if the Tiltfile uses config.parse or config.set_enabled_resources.
+In that case, see (hopefully) comments in your Tiltfile and/or https://tilt.dev/user_config.html
+`,
 	}
 
 	cmd.Flags().BoolVar(&c.watch, "watch", true, "If true, services will be automatically rebuilt and redeployed when files change. Otherwise, each service will be started once.")
```
maiamcc: maybe not quite this wording, but consider something like:
"The args/options appearing in this help text are Tilt built-in args, which are distinct from Tiltfile args (provided after `--`, processed by your tiltfile)"

internal/cli/args.go
landism: good call

internal/cli/args.go
landism: Yes, with the potential nit that the docs say `--port` controls the "Tilt HTTP server", not the web UI.

internal/cli/args.go
landism: > This is a nit and I'm not sure how important it is to clarify for users, but Tilt doesn't run on a port, the webserver runs on a port. (also, i feel like this implementation isn't compatible with someone running Tilt with `--port=0`, is that important?)

At the risk of getting arguey, I'm not sure I agree with the value or correctness of that distinction. If someone mentioned the port their docker was running on and someone else replied "docker doesn't run on a port, its HTTP API runs on a port", is that reducing confusion? I'm not even sure what "run on a port" would mean at that point. Maybe one could make an argument about the extent to which a program's interface is via the port.

I suspect that almost none of our users use --port, which would make discussion of that noise, and for the users who do use --port, they're presumably running like `tilt up --port 3248` and "running Tilt on a port" feels like a good match to that usage?

aaaand, I guess we're still logging "running on a port" to users who aren't specifying a --port, so that's not as clean as I thought. How about `fmt.Printf("changed config args for Tilt running at %s to %v\n", schemeHostPort, args)`?

PR #2672, opened 2019-12-12
web: Add new HUDLayout to Storybook in prep for new page hierarchy

hyu:
I'm proposing a change to the visual hierarchy of our main page, in preparation for creating more space in our layout for helpful Log navigation tools. 

First step: Create a new HUDLayout which will eventually replace HUDGrid:

Screen Shot 2019-12-12 at 4 56 04 PM

### DIFF
--- a/web/src/HUDLayout.stories.tsx
+++ b/web/src/HUDLayout.stories.tsx
@@ -20,42 +20,41 @@ let Main = styled.main`
   border-right: 1px dashed ${Color.white};
   border-bottom: 1px dashed ${Color.white};
   padding: ${SizeUnit(0.5)};
-  white-space: pre;
 `
 
-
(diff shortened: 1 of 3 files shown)


web/src/HUDLayout.stories.tsx
```diff
@@ -0,0 +1,87 @@
...
+  background-color: ${Color.white};
+  color: ${Color.gray};
+  padding: ${SizeUnit(0.25)};
+`
+
+let Main = styled.main`
+  border-right: 1px dashed ${Color.white};
+  border-bottom: 1px dashed ${Color.white};
+  padding: ${SizeUnit(0.5)};
+  white-space: pre;
```
nicks: are you sure you want this to be white-space: pre? this makes sense for, e.g., LogPane, but not so much for other kinds of main panes.

PR #2683, opened 2019-12-17
internal/tiltfile: add experimental_telemetry_cmd

jazzdan:


internal/tiltfile/tiltfile_state.go
```diff
@@ -453,6 +455,7 @@ func (s *tiltfileState) OnStart(e *starkit.Environment) error {
 		{disableFeatureN, s.disableFeature},
 		{disableSnapshotsN, s.disableSnapshots},
 		{setTeamN, s.setTeam},
+		{telemetryCmdN, s.setTelemetryCmd},
```
nicks: please don't add new built-ins here, make it an extension

jazzdan: Got it, will change.

jazzdan: LMK what you think now @nicks!

internal/tiltfile/telemetry/telemetry.go
```diff
@@ -0,0 +1,80 @@
...
+import (
+	"fmt"
+
+	"go.starlark.net/starlark"
+
+	"github.com/windmilleng/tilt/internal/tiltfile/starkit"
+	"github.com/windmilleng/tilt/pkg/model"
+)
+
+type Settings struct {
```
nicks: fwiw, I've been defining these sorts of settings objects in pkg/model, and handing them directly to EngineState, rather than make tiltfileLoader do custom unpacking logic per-model

I think this starts to point to a general Tilt extension model where you define a (1) Starkit extension, (2) a data model, and (3) a subscriber to that model that kinda parallels the CRD spec / status / operator mindset

internal/tiltfile/telemetry/telemetry.go
```diff
@@ -0,0 +1,80 @@
...
+	if err != nil {
+		return starlark.None, err
+	}
+
+	if len(cmd) == 0 {
+		return starlark.None, fmt.Errorf("cmd cannot be empty")
+	}
+
+	var innerErr error
+	err = starkit.SetState(thread, func(settings Settings) Settings {
```
nicks: The SetState callback is allowed to return an error, see how config.Extension does it

internal/tiltfile/telemetry/telemetry.go
```diff
@@ -0,0 +1,80 @@
...
+		if len(settings.Cmd.Argv) > 0 {
+			innerErr = fmt.Errorf("%v called multiple times; already set to %v", fn.Name(), settings.Cmd)
+			return settings
+		}
+
+		settings.Cmd = model.ToShellCmd(cmd)
```
nicks: should the Telemetry command also inherit the working directory of the Tiltfile? That would be easy to add and dovetail nicely with L's work. Then you could load() a remote github repo with your Tiltfile + telemetry script, which would be useful for orgs with multiple repos.

PR #2684, opened 2019-12-17
internal/engine/telemetry: refactor in response to review comments

dbentley:
1) rename Exporter -> SpanCollector
2) move serialization out of main goroutine
3) change interface to requeue the spans when rejected


internal/tracer/span_collector.go
```diff
@@ -0,0 +1,120 @@
...
+	}
+	go r.loop(ctx)
+	return r
+}
+
+func (c *SpanCollector) loop(ctx context.Context) {
+	// spans that have come in and are waiting to be read by a consumer
+	var queue []*exporttrace.SpanData
+
+	for {
```
jazzdan: Should this goroutine ever stop? Maybe when Shutdown is called?

dbentley: Just added it; lmk if you want to chat through it.

jazzdan: LGTM!

PR #2686, opened 2019-12-17
vendor: update to kubernetes 1.17

nicks:
Hello @nicks,

Please review the following commits I made in branch nicks/k8s117:

7cb037b9652562b37b6cad7911422f3fdfe3894f (2019-12-17 16:37:59 -0500)
vendor: update to kubernetes 1.17


nicks:
something broke in how the client-go test fixture handles namespace, adding namespace explicitly in places where we use the test fixture fixes the issue. I verified that revering the client-go fixture changes also fixes the problem, just to be sure it's not a client-go "prod" issue.

PR #2690, opened 2019-12-18
web: go back to wrapping-by-default and add section dividers

nicks:
Hello @hyu,

Please review the following commits I made in branch nicks/wrap:

c2a234a79043d7579ad9d763c0527423dd19c48d (2019-12-18 13:39:24 -0500)
web: go back to wrapping-by-default and add section dividers

### DIFF
--- a/web/src/LogPane.tsx
+++ b/web/src/LogPane.tsx
@@ -33,6 +33,7 @@ type LogLineComponentProps = {
   lineId: number
   shouldHighlight: boolean
   showManifestPrefix: boolean
+  isContextChange: boolean
 }
 
 let LogLinePrefixRoot = styled.span`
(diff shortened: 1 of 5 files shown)


nicks:
Screenshot from 2019-12-18 13-39-56

web/src/LogPane.tsx
```diff
@@ -43,6 +44,7 @@ let LogLinePrefixRoot = styled.span`
   box-sizing: border-box;
   display: inline-block;
   background-color: ${color.grayDark};
+  border-top: 1px dotted transparent;
   border-right: 1px dotted ${color.grayLightest};
```
hyu: I was playing with this on my machine and think it's clearer if the vertical line separating the prefixes from logs is more prominent than the horizontal lines separating each context.

So, the root styles change like so:
```
border-right: 1px solid ${color.grayLightest};
```
> ~~border-top: 1px dotted transparent;~~ (To be handled with negative margin, below)

...and `is-contextChange` gets these:

```
    margin-top: -1px; // Don't let border change height of element
    border-top: 1px dotted ${color.grayLightest};
```

nicks: done!

nicks:
here's what it looks like now:

Screenshot from 2019-12-18 14-21-44

PR #2692, opened 2019-12-18
web: default span ids to _. Fixes firestore issues with empty map keys

nicks:
Hello @jazzdan,

Please review the following commits I made in branch nicks/ch4132/firestore:

2162402d2520fa86f0fad7de897a405170655bab (2019-12-18 14:33:44 -0500)
web: default span ids to _. Fixes firestore issues with empty map keys


nicks:
it probably makes sense to do this in the tilt engine as well, but this fixes the immediate problem

PR #2693, opened 2019-12-19
model: remove logs from builds and pods

nicks:
Hello @maiamcc, @jazzdan,

Please review the following commits I made in branch nicks/logevent:

6adef23272d9d1a7d6f4a015fddaaa4fd84cd1e7 (2019-12-18 22:09:49 -0500)
model: remove logs from builds and pods


pkg/model/build_status.go
```diff
@@ -24,7 +24,7 @@ type BuildRecord struct {
 
 	// TODO(nick): Delete Log and use SpanID to load
 	// the log from the logstore.
-	Log Log `testdiff:"ignore"`
+	//Log Log `testdiff:"ignore"`
```
jazzdan: 👀

nicks: hah, removed!

nicks:
re: porting logstore changes. Most of the logstore changes are around the Tail() function. I'm...not sure. I know Han had some ideas around in-context logs in the web UI, so an InContext() (like grep --context) might make more sense than a Tail() (which makes more sense in a terminal context)

PR #2695, opened 2019-12-19
web: Fix up the Facets protocol so that it doesn't copy huge strings on every change

nicks:
Hello @maiamcc, @jazzdan,

Please review the following commits I made in branch nicks/facet:

48ae5c447abc39bcc0220d87341482502f1264dd (2019-12-19 13:06:23 -0500)
web: Fix up the Facets protocol so that it doesn't copy huge strings on every change


nicks:
I also fixed some annoying ux bugs while i was here

PR #2696, opened 2019-12-19
convert to go modules

jazzdan:


nicks:
:birthday: 

would it make sense to update circleci, etc to use go 1.13 first? just so we don't get bitten by go 1.12 vs 1.13 modules differences

jazzdan:
FYI @nicks I switched to a different branch https://github.com/windmilleng/tilt/pull/2697

But yes I did this work on Go 1.13 and I'm updating CI accordingly.

PR #2697, opened 2019-12-19
convert to go modules take 2

jazzdan:


jazzdan:
This is ready for review @nicks

nicks:
looks like there was a backwards-incompatible change to go here: https://github.com/golang/go/issues/34610

and we'll need to pull up some docker changes to fix

PR #2700, opened 2019-12-19
logstore: add an api for pulling warnings out of the logstore

nicks:
Hello @jazzdan,

Please review the following commits I made in branch nicks/warning2:

7612954c737aa781be50f53a81e5c2d3a6d9424f (2019-12-19 17:42:42 -0500)
logstore: add an api for pulling warnings out of the logstore


pkg/model/logstore/logstore.go
```diff
@@ -41,6 +41,15 @@ type LogSegment struct {
 
 	// Continues a line from a previous segment.
 	ContinuesLine bool
+
+	// When we store warnings in the LogStore, we break them up into lines and
+	// store them as a series of line segments. Atom marks the beginning of a
```
jazzdan: What does Atom mean here?

nicks: oooooops, Atom was an earlier name for this property before I changed it to 'anchor' (i changed it a few times)

nicks: good catch, fixed!

PR #2704, opened 2019-12-20
engine: remove logs from the runtime data model and load them from the logstore

nicks:
Hello @maiamcc, @jazzdan,

Please review the following commits I made in branch nicks/runtimelog:

4cca650332357d0bced0b1f08b9aff5c560f5f07 (2019-12-20 17:21:13 -0500)
engine: remove logs from the runtime data model and load them from the logstore
Note that this changes functionality slightly. Before this PR,
we would clear the pod logs every time the pod crashed. In the LogStore,
there's a single span ID for a pod's logs, no matter how many times it crashed.

I don't *think* that changing this behavior will have any bad side-effects
but @maiamcc might be able to think of problems that I don't remember right now.

Long-term, I think we should move towards a system where every container ID
has its own log span ID, and that span ID has a parent span ID for the whole
pod's logs (similar to how spans work in opentracing)

### DIFF
--- a/internal/dockercompose/state.go
+++ b/internal/dockercompose/state.go
@@ -55,9 +55,6 @@ type State struct {
 	Status      Status
 	ContainerID container.ID
 
-	// TODO(nick): Delete this and use SpanID to look up the log.
-	CurrentLog model.Log
-
 	StartTime     time.Time
 	IsStopping    bool
 	LastReadyTime time.Time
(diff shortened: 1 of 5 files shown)


nicks:
also, i don't remember why we originally cleared the logs this way, and we might need to withdraw this pr if someone does remember!

PR #2706, opened 2019-12-21
web: HUD layout allows log filtering in correct hierarchy

hyu:
This moves info about the resources above the tabbed navigation. This way, we can add log filtering controls below the Log tab.

## Screenshot — at top of view
Screen Shot 2019-12-20 at 11 32 01 PM

## Screenshot — scrolled down
Screen Shot 2019-12-20 at 11 32 03 PM

## Known Issues
1. When tailing logs or just when scrolling down, the content in the header is not visible. Our current HUD layout makes it annoying to both tail logs and also see the header when you have more than a screenful of logs. Possible ways to address this:

- the header and nav can both be persistently visible and just take space from the logs. This makes sense if we believe that we need a layer of hierarchy above logs, which will grow as we experiment with other less log-centric ways of representing resource state.

- we can squish everything into a single persistent nav. Currently, this is possible because we still only show limited data about each resource. This makes the view navigation more simple. But we'll have less space to work with, and thus need to squeeze endpoints and Pod ID into dropdown menus.

2. The Create Snapshot button is in the header, and also not visible during times we'd want it to be (ex: when highlighting lines). If we want to keep the prominence of this feature, we could...

- make the button part of the sticky nav
- expand the nav when scrolling up, so we can see this Snapshot button again (a la [headroom](https://wicky.nillia.ms/headroom.js/))
- move the button elsewhere (such as in the sidebar, which makes more sense than its current location)


hyu:
I have a failing test and I can't figure out why it's failing.

#### In HUD.test.tsx:
```
it("renders snapshot button if snapshots are enabled and this isn't a snapshot view", async () => {
  ...
  let button = root.find(".snapshotButton")
  expect(button.exists()).toBe(true)

  button.simulate("click")
```

So the test complains...

> Method “simulate” is meant to be run on 1 node. 3 found instead.

...but I'm quite certain that the class name only appears once in the entire DOM. I tried `button.debug()` to see what jest sees, but that didn't seem to work.

nicks:
hold on, lemme poke at it and see if i can figure out why it's failing

web/src/HUDHeader.scss
```diff
@@ -1,5 +1,7 @@
 @import "constants";
 
+// DELETE THIS
```
nicks: ?

nicks:
@hyu you need to change `root.find(".snapshotButton")` to `root.find("button.snapshotButton")`

The issue is that styled-components use a bunch of wrapper elements in the virtual DOM with the same class name.

for future reference, `console.error(wrapper.debug())` will print out the debug info of a ReactWrapper https://airbnb.io/enzyme/docs/api/ReactWrapper/debug.html

web/src/HUDHeader.tsx
```diff
@@ -0,0 +1,143 @@
...
+import styled from "styled-components"
+import * as s from "./style-helpers"
+import { SnapshotHighlight } from "./types"
+import { ReactComponent as SnapshotSvg } from "./assets/svg/snapshot.svg"
+
+type HUDHeaderProps = {
+  podID?: string
+  endpoints?: string[]
+  podStatus?: string
+  showSnapshotButton?: boolean
```
nicks: i'm surprised these 3 fields (showSnapshotButton, handleOpenModal, highlight) are optional. they seem like ones that you never want to leave off

nicks:
anecdotally, in the new UI, I found it fairly difficult to make the port-forward URL appear, even when I knew where it was. i predict we would get a lot of complaints about this. I vote that we should make the topnav a fixed height and not do the disappearing thing until we have a better solution.

web/src/HUDHeader.scss
hyu: Oops! Had a bit more cleanup to do here.

hyu:
Thanks for the careful review, @nicks!

- I agree that the disappearing header is too much for now. I will make sure it remains visible.
- Sorry about cramming in the changes to styles-helper. I'll be sure to separate such changes into a separate PR in the future.
- Really good to know about why my test was failing. That all makes sense!

PR #2713, opened 2020-01-02
engine test: fakeBuildAndDeployer can now resolve builds manually

maiamcc:
this PR doesn't change any behavior right now, but sets me up
to better test parallel builds--can now start builds A, B, and C,
and complete just build C

### DIFF
--- a/internal/engine/upper_test.go
+++ b/internal/engine/upper_test.go
@@ -70,6 +70,8 @@ import (
 
 var originalWD string
 
+type buildCompletionChannel chan bool
+
 func init() {
 	wd, err := os.Getwd()
 	if err != nil {
@@ -156,9 +158,8 @@ type fakeBuildAndDeployer struct {
 	t     *testing.T
 	calls chan buildAndDeployCall
 
-	resolveBuildsManually    bool
-	buildsAwaitingResolution map[int]bool
-	resolveBuildsMu          sync.Mutex
+	completeBuildsManually bool
+	buildCompletionChans   sync.Map // map[int]buildCompletionChannel; close channel at buildCompletionChans[n] to complete build #n
 
 	buildCount int
 
@@ -199,15 +200,16 @@ func (b *fakeBuildAndDeployer) nextBuildResult(iTarget model.ImageTarget, deploy
 
 func (b *fakeBuildAndDeployer) BuildAndDeploy(ctx context.Context, st store.RStore, specs []model.TargetSpec, state store.BuildStateSet) (brs store.BuildResultSet, err error) {
 	b.buildCount++
-	count := b.buildCount
+	index := b.buildCount
+	b.registerBuild(index)
 
-	if !b.resolveBuildsManually {
+	if !b.completeBuildsManually {
 		// resolve builds automatically: mark the build for resolution now,,
 		// so we complete the build as soon as we start it
-		b.resolveBuildAtIndex(count)
+		b.resolveBuild(index)
 	}
 
-	call := buildAndDeployCall{count: count, specs: specs, state: state}
+	call := buildAndDeployCall{count: index, specs: specs, state: state}
 	if call.dc().Empty() && call.k8s().Empty() && call.local().Empty() {
 		b.t.Fatalf("Invalid call: %+v", call)
 	}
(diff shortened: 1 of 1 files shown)


maiamcc:
For reference, here's what a test looks like that uses manual build resolution:
```
func TestBuildControllerWontBuildManifestIfNoSlotsAvailable(t *testing.T) {
	f := newTestFixture(t)
	defer f.TearDown()
	f.setMaxBuildSlots(2)
	f.b.resolveBuildsManually = true

	manA := f.newDockerBuildManifestWithBuildPath("manA", f.JoinPath("a"))
	manB := f.newDockerBuildManifestWithBuildPath("manB", f.JoinPath("b"))
	manC := f.newDockerBuildManifestWithBuildPath("manC", f.JoinPath("c"))
	f.StartAndWaitForInitialBuilds([]model.Manifest{manA, manB, manC}, true)

	// start builds for all manifests (we only have 2 build slots)
	f.fsWatcher.events <- watch.NewFileEvent(f.JoinPath("a/main.go"))
	f.assertManifestBuilding("manA")
	manABuildIndex := f.b.buildCount
	f.fsWatcher.events <- watch.NewFileEvent(f.JoinPath("b/main.go"))
	f.assertManifestBuilding("manB")
	f.fsWatcher.events <- watch.NewFileEvent(f.JoinPath("c/main.go"))
	f.assertManifestNotBuilding("manC")

	// Complete one build -- now there's a free build slot for 'manC'
	f.b.resolveBuildAtIndex(manABuildIndex)
	call := f.nextCall("expect manA build complete")
	assert.Equal(t, manA.ImageTargetAt(0).ID(), call.firstImgTarg().ID())
	assert.Equal(t, call.oneState().FilesChanged(), []string{f.JoinPath("a/main.go")})
	f.assertManifestNotBuilding("manA")
	f.assertManifestBuilding("manC")

	err := f.Stop()
	assert.NoError(t, err)
	f.assertAllBuildsConsumed()
}
```

internal/engine/upper_test.go
```diff
@@ -272,16 +283,41 @@ func (b *fakeBuildAndDeployer) BuildAndDeploy(ctx context.Context, st store.RSto
 		b.resultsByID[key] = val
 	}
 
+	// block until we know we're supposed to resolve this build
+	for {
+		if _, ok := b.buildsAwaitingResolution[count]; ok {
```
nicks: did you intend to also lock the mutex here? otherwise you'll have race conditions

i might also break this into a waitUntilBuildResolves(index) helper, just to cut down on the size of this method

internal/engine/upper_test.go
```diff
@@ -156,6 +156,10 @@ type fakeBuildAndDeployer struct {
 	t     *testing.T
 	calls chan buildAndDeployCall
 
+	resolveBuildsManually    bool
```
landism: I was a bit confused by 'resolve' - 'complete' feels clearer to me. WDYT?

internal/engine/upper_test.go
maiamcc: good catch! and that makes sense, will do

landism: fine as is, but if the values were chans instead of bools, this might be a bit cleaner (no loop needed). up to you.

internal/engine/upper_test.go
```diff
@@ -195,8 +199,15 @@ func (b *fakeBuildAndDeployer) nextBuildResult(iTarget model.ImageTarget, deploy
 
 func (b *fakeBuildAndDeployer) BuildAndDeploy(ctx context.Context, st store.RStore, specs []model.TargetSpec, state store.BuildStateSet) (brs store.BuildResultSet, err error) {
 	b.buildCount++
+	count := b.buildCount
```
landism: might be clearer to change "count" to "index"

internal/engine/upper_test.go
```diff
@@ -272,16 +283,41 @@ func (b *fakeBuildAndDeployer) BuildAndDeploy(ctx context.Context, st store.RSto
...
-		resultsByID:    store.BuildResultSet{},
+		t:                        t,
+		calls:                    make(chan buildAndDeployCall, 20),
+		buildsAwaitingResolution: make(map[int]bool),
+		buildLogOutput:           make(map[model.TargetID]string),
+		resultsByID:              store.BuildResultSet{},
+	}
+}
+
+func (b *fakeBuildAndDeployer) resolveBuildAtIndex(i int) {
```
landism: could rename the arg to `index` and drop `AtIndex` from the function name, if you wanted

internal/engine/upper_test.go
maiamcc: oooh neat! ty

maiamcc:
made some nontrivial changes for code review comments--since concurrency is gnarly, could you take another look @nicks @landism ? ty!

PR #2722, opened 2020-01-03
server: use the standard golang pprof hooks

nicks:
Hello @jazzdan, @maiamcc,

Please review the following commits I made in branch nicks/httppprof:

28f8722b83c95ab2c897a47f99eae882a6288fe6 (2020-01-03 15:13:52 -0500)
server: use the standard golang pprof hooks


nicks:
ya, i think we should remove it on the next release

PR #2723, opened 2020-01-03
hud: only redraw on clock tick, not OnChange

nicks:
Hello @maiamcc,

Please review the following commits I made in branch nicks/perf:

436891c58a8befc147b34c8bd56a017f13dad11f (2020-01-03 15:19:18 -0500)
hud: only redraw on clock tick, not OnChange


nicks:
i think this will address a lot of the remaining issues in https://github.com/windmilleng/tilt/issues/2597

PR #2724, opened 2020-01-03
engine/tests: fix race condition in fakeBuildAndDeployer concurrency

maiamcc:
this solution swiped from https://go101.org/article/channel-closing.html

### DIFF
--- a/internal/engine/upper_test.go
+++ b/internal/engine/upper_test.go
@@ -72,6 +72,16 @@ var originalWD string
 
 type buildCompletionChannel chan bool
 
+func (bcc buildCompletionChannel) isClosed() bool {
+	select {
+	case <-bcc:
+		return true
+	default:
+	}
+
+	return false
+}
+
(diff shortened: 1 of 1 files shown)


internal/engine/upper_test.go
```diff
@@ -3739,7 +3751,9 @@ func (f *testFixture) TearDown() {
 	close(f.fsWatcher.errors)
 	f.b.buildCompletionChans.Range(func(key, value interface{}) bool {
 		if ch, ok := value.(buildCompletionChannel); ok {
-			close(ch)
+			if !ch.isClosed() {
```
nicks: this still has race conditions, because your check is non-atomic.

tbh i wish we hadn't used channels at all here -- it's not really a good fit for what we're doing.

nicks: alternatively, just remove this Range() call entirely, it's not really necessary

PR #2725, opened 2020-01-03
engine/tests: a few more fixes to fakeBuildAndDeployer for concurrency safety

maiamcc:
just got hit with a "concurrent map writes" complaint in my other branch
for b.ResultsByID so here's this.

### DIFF
--- a/internal/engine/upper_test.go
+++ b/internal/engine/upper_test.go
@@ -156,6 +156,7 @@ func (c buildAndDeployCall) oneState() store.BuildState {
 
 type fakeBuildAndDeployer struct {
 	t     *testing.T
+	mu    sync.Mutex
 	calls chan buildAndDeployCall
 
 	completeBuildsManually bool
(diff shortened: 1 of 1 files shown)


internal/engine/upper_test.go
```diff
@@ -340,9 +345,9 @@ func newFakeBuildAndDeployer(t *testing.T) *fakeBuildAndDeployer {
 func (b *fakeBuildAndDeployer) completeBuild(index int) {
 	ch, ok := b.getBuildCompletionChan(index)
 	if !ok {
-		// TODO(maia): maybe this should just create and close a channel for this key
-		b.t.Fatalf("tried to mark build #%d for for completion, but no completion channel registered (b.registerBuild(index) "+
-			"should already have been called--probably by b.BuildAndDeploy)", index)
+		// If build chan doesn't exist, create and store it
```
nicks: this will cause race conditions. in async programming, you usually want to use LoadOrStore to atomically create something that doesn't exist
https://golang.org/pkg/sync/#Map.LoadOrStore

nicks: i would probably change getBuildCompletionChan to getOrCreateBuildCompletionChan

maiamcc:
good catch, ty

PR #2729, opened 2020-01-06
engine/build_control: enable parallel build (off by default) [ch4065]

maiamcc:
someday i'll redo the fakeBaD manual build completion logic with
a map instead of channels, but today is not that day :(


internal/engine/buildcontroller_test.go
```diff
@@ -1047,3 +1048,303 @@ func TestLogsLongResourceName(t *testing.T) {
...
+	if len(indexes) != len(manifests) {
+		f.t.Fatalf("can only complete + check builds for manifests when passed an equal number of indexes and manifests")
+	}
+
+	for _, i := range indexes {
+		f.b.completeBuild(i)
+	}
+
+	targetsFromCalls := f.imageTargsForNextNCalls(len(manifests))
+	require.Len(f.t, targetsFromCalls, len(manifests))
```
nicks: this check doesn't seem correct, and only works in particular tests

internal/engine/upper_test.go
```diff
@@ -3501,13 +3524,20 @@ func (f *testFixture) WaitForNoExit() error {
...
-	// by the state machine.
-	f.WaitUntil("build complete processed", func(state store.EngineState) bool {
-		return len(state.CurrentlyBuilding) == 0
+	// Before setting the nextBuildFailure, make sure that any in-flight builds (state.BuildStartedCount)
+	// have hit the buildAndDeployer (f.b.buildCount); by the time we've incremented buildCount and
+	// the fakeBaD mutex is unlocked, we've already grabbed the nextBuildFailure for that build,
+	// so we can freely set it here for a future build.
+	f.WaitUntil("any in-flight builds have hit the buildAndDeployer", func(state store.EngineState) bool {
+		f.b.mu.Lock()
+		f.b.mu.Unlock()
```
nicks: is this supposed to be `defer Unlock()`?

internal/engine/upper_test.go
```diff
@@ -3407,6 +3407,23 @@ func (f *testFixture) Start(manifests []model.Manifest, watchFiles bool, initOpt
...
 
+func (f *testFixture) StartAndWaitForInitialBuilds(manifests []model.Manifest, initOptions ...initOption) {
+	f.startWithInitManifests(nil, manifests, true, initOptions...)
+	if f.b.completeBuildsManually {
+		for i := range manifests {
+			f.b.completeBuild(i + 1) // builds are 1-indexed
+		}
+	}
+
+	expectedTargs := make([]model.ImageTarget, len(manifests))
```
nicks: this check doesn't seem generally correct, because it assumes 1 image target per manifest

can you use the waitForCompletedBuildCount() check, for consistency with other tests?

internal/engine/upper_test.go
```diff
@@ -3407,6 +3407,23 @@ func (f *testFixture) Start(manifests []model.Manifest, watchFiles bool, initOpt
 	f.startWithInitManifests(nil, manifests, watchFiles, initOptions...)
 }
 
+func (f *testFixture) StartAndWaitForInitialBuilds(manifests []model.Manifest, initOptions ...initOption) {
```
nicks: I really don't want to create a new Start method variant every time we have a new batch of tests with new defaults.

Can you change your tests to call the existing start method?

internal/engine/upper_test.go
```diff
@@ -3630,6 +3660,15 @@ func (f *testFixture) nextCall(msgAndArgs ...interface{}) buildAndDeployCall {
 	}
 }
 
+func (f *testFixture) imageTargsForNextNCalls(n int, msgAndArgs ...interface{}) []model.ImageTarget {
```
nicks: this helper function seems problematic.

It invents a new way to assert things about calls, rather than using the existing assert conventions. And it bakes in some assumptions that aren't true in all test cases (1 ImageTarget per Manifest), then hides that assumption.

internal/engine/buildcontroller_test.go
```diff
@@ -1047,3 +1048,303 @@ func TestLogsLongResourceName(t *testing.T) {
...
+			return ms.LastBuild().Error == errA
+		})
+	})
+
+	// complete the rest
+	f.completeAndCheckBuildsForManifests([]int{indexB, indexC}, []model.Manifest{manB, manC})
+	f.WaitUntilManifestState("last manB build reflects expected error", "manB", func(ms store.ManifestState) bool {
+		return ms.LastBuild().Error == errB
+	})
+	time.Sleep(10 * time.Millisecond)
```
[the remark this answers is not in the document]
maiamcc: it takes a few ms for the BuildComplete action to bubble through, so we could potentially get a false positive if we just checked for "manifest C last build has no error" before a BuildComplete action had made it through the pipe. (Granted this is LESS likely because we know the manifest B BuildComplete has already made it through, but seemed prudent.)

will add a comment, unless you think this is unnecessary

internal/engine/upper_test.go
maiamcc: good call, will do

internal/engine/upper_test.go
maiamcc: This is less a new start method and more a wrapper -- notice that we DO call the existing start method, and just do some checks after it completes. do you have suggestions for how to rename this so it's clearer what's going on?

internal/engine/upper_test.go
maiamcc: 😅yes.

internal/engine/buildcontroller_test.go
```diff
@@ -1047,3 +1048,303 @@ func TestLogsLongResourceName(t *testing.T) {
...
+		_, ok := st.CurrentlyBuilding[name]
+		require.True(f.t, ok, "expected EngineState to reflect that %q is currently building", name)
+	})
+}
+
+func (f *testFixture) waitUntilBuildCountAtLeast(n int) {
+	ctx, cancel := context.WithTimeout(f.ctx, time.Millisecond*200)
+	defer cancel()
+	for {
+		if f.b.buildCount >= n {
```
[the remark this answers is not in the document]
maiamcc: synchronizing on which `buildCount`?

so here's an example of where I'm using this func:
```
func (f *testFixture) editFileAndWaitForManifestBuilding(name model.ManifestName, path string) (buildIndex int) {
        startBuildCount := f.b.buildCount
        f.fsWatcher.events <- watch.NewFileEvent(f.JoinPath(path))
        f.waitUntilManifestBuilding(name)
        f.waitUntilBuildCountAtLeast(startBuildCount + 1)
        return startBuildCount + 1
}
```
`f.waitUntilManifestBuilding(name)` checks that the manifest is building _according to the engine state_, i.e. that we've processed the `BuildStartedAction`. The point of `waitUntilBuildCountAtLeast` (which I should probably rename) is to make sure that this build has hit the BaD (and that the index that we return definitely corresponds to this build).

Without this check, I ran into problems with something like:

```
f.setMaxBuildSlots(3)
indexA := f.editFileAndWaitForManifestBuilding("manA", "a/main.go")
indexB := f.editFileAndWaitForManifestBuilding("manB", "b/main.go")
indexC := f.editFileAndWaitForManifestBuilding("manC",  "c/main.go")
f.b.completeBuild(indexB)
```
because of the order in which these manifests actually hit the buildAndDeployer, the indexes we associated with them were NOT a simple matter of `f.b.buildCount++` for each; e.g. here's a test result where resolving the build at `indexB` didn't actually correspond to a build of manifestB:
```

--- FAIL: TestCurrentlyBuildingMayExceedMaxBuildCount (0.01s)
    buildcontroller_test.go:1147: 
                Error Trace:        buildcontroller_test.go:1147
                Error:              Not equal: 
                                    expected: model.TargetID{Type:"image", Name:"manb"}
                                    actual  : model.TargetID{Type:"image", Name:"mana"}
                                    
                                    Diff:
                                    --- Expected
                                    +++ Actual
                                    @@ -1,2 +1,2 @@
                                    -(model.TargetID) image:manb
                                    +(model.TargetID) image:mana
                                     
                Test:               TestCurrentlyBuildingMayExceedMaxBuildCount
```
BUT I would love a better way of doing this!

internal/engine/buildcontroller_test.go
```diff
@@ -1047,3 +1048,303 @@ func TestLogsLongResourceName(t *testing.T) {
...
+	f.waitUntilBuildCountAtLeast(startBuildCount + 1)
+	return startBuildCount + 1
+}
+
+func (f *testFixture) editFileAndAssertManifestNotBuilding(name model.ManifestName, path string) {
+	f.fsWatcher.events <- watch.NewFileEvent(f.JoinPath(path))
+	f.waitUntilManifestNotBuilding(name)
+}
+
+func (f *testFixture) runAndWaitForManifestBuilding(name model.ManifestName, fn func()) (buildIndex int) {
```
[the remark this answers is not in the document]
maiamcc: hmm fair, let me see how clunky that makes the test body.

i assume from your lack of comments that you feel ok about `editFileAndWaitForManifest(Not)Building`?

internal/engine/buildcontroller_test.go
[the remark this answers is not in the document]
nicks: hmmm...but that means if the behavior is totally broken, then this test will pass? right?

i'm not opposed to adding Sleeps(). But in this case, the sleep is a code smell that the test has deeper problems.

internal/engine/upper_test.go
nicks: i would just create a separate WaitForInitialBuilds method and call that in the top-level test cases

internal/engine/buildcontroller_test.go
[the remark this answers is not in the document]
nicks: you're reading f.b.buildCount, so you need to f.b.mu.Lock() the mutex.

tho this whole discussion makes me wonder if the real problem is that we've structured the test harness wrong. Using indices as identifiers means you have to do a ton of work to enforce ordering. But the whole point of this code is that things should happen in parallel, and we're not actually exercising the parallelism.

what if we changed registerBuild/completeBuild to use target ids instead of indices? would that simplify things?

internal/engine/upper_test.go
maiamcc: i can fix the 1 target per manifest thing for sure -- but i need some order-agnostic way of checking calls, which is what this attempted to be. let me see if i can make it cleaner.

internal/engine/buildcontroller_test.go
maiamcc: fixed!

internal/engine/upper_test.go
maiamcc: sure

PR #2730, opened 2020-01-06
model: local resource should support serve

jazzdan:
Just setting aside the names for the two different types of commands that local resources will be able to have going forward:

1. `UpdateCmd` - is what we have been referring to as a `LocalResource.Cmd` today
2. `ServerCmd` - this is something like `rails s`: a long running command that we want to keep running _after_ the `UpdateCmd` has completed.


pkg/model/local_target.go
```diff
@@ -7,27 +7,28 @@ import (
 )
 
 type LocalTarget struct {
-	Name    TargetName
-	Cmd     Cmd
-	Workdir string   // directory from which this Cmd should be run
-	deps    []string // a list of ABSOLUTE file paths that are dependencies of this target
-	ignores []Dockerignore
+	Name      TargetName
+	UpdateCmd Cmd
```
maiamcc: comment the two commands pls

pkg/model/local_target.go
```diff
@@ -7,27 +7,28 @@ import (
...
+	deps      []string // a list of ABSOLUTE file paths that are dependencies of this target
+	ignores   []Dockerignore
 
 	repos []LocalGitRepo
 }
 
 var _ TargetSpec = LocalTarget{}
 
-func NewLocalTarget(name TargetName, cmd Cmd, workdir string, deps []string) LocalTarget {
+func NewLocalTarget(name TargetName, updateCmd Cmd, workdir string, deps []string, serveCmd Cmd) LocalTarget {
```
maiamcc: this is nbd but logically I'd expect the two commands to be grouped together:
> func NewLocalTarget(name TargetName, updateCmd Cmd, serveCmd Cmd, workdir string, deps []string)

pkg/model/local_target.go
```diff
@@ -7,27 +7,28 @@ import (
...
 	return LocalTarget{
-		Name:    name,
-		Cmd:     cmd,
-		Workdir: workdir,
-		deps:    deps,
+		Name:      name,
+		UpdateCmd: updateCmd,
+		Workdir:   workdir,
+		deps:      deps,
```
maiamcc: also, did you mean to actually attach the `ServeCmd` here?

pkg/model/local_target.go
```diff
@@ -7,27 +7,28 @@ import (
...
 	}
 }
 
-func (lt LocalTarget) Empty() bool { return lt.Cmd.Empty() }
+func (lt LocalTarget) Empty() bool { return lt.UpdateCmd.Empty() }
```
maiamcc: either `UpdateCmd.Empty() || ServeCmd.Empty()` or put a TODO

jazzdan:
@maiamcc I just wanted something to differentiate the two commands.

Updates aren't necessarily builds, so the word build is out. A local resource could be:

```python
local_resource("drop everything", "mysql -u user < ./drop_all_tables.sql")
local_resource("enable JIT", "echo JIT_ENABLED=1 > php.conf && systemctl restart apache)
```

It feels wrong to me to refer to those as builds

PR #2731, opened 2020-01-07
tiltfile: refactor local target tests to use assertNextManifest style tests

jazzdan:


internal/tiltfile/tiltfile_test.go
```diff
@@ -4769,6 +4739,16 @@ func (f *fixture) assertNextManifest(name model.ManifestName, opts ...interface{
 			assert.Equal(f.t, opt.deps, m.ResourceDependencies)
 		case funcOpt:
 			assert.True(f.t, opt(f.t, m))
+		case ltHelper:
+			lt := m.LocalTarget()
+			for _, matcher := range opt.matchers {
+				switch matcher := matcher.(type) {
+				case updateCmdHelper:
+					assert.Equal(f.t, matcher.cmd, lt.UpdateCmd.String())
+				case depsHelper:
```
nicks: this needs a default case that calls Fatal()

internal/tiltfile/tiltfile_test.go
```diff
@@ -3950,30 +3934,16 @@ local_resource("toplvl-local", "echo hello world", ["foo/baz", "foo/a.txt"])
...
-		f.JoinPath("nested/more_nested/repo"),
-	}, ltNested.Dependencies())
 	f.assertRepos([]string{
 		f.JoinPath("nested"),
 		f.JoinPath("nested/more_nested/repo"),
 	}, ltNested.LocalRepos())
 
-	mTop := f.loadResult.Manifests[1]
-	require.Equal(t, "toplvl-local", mTop.Name.String())
+	mTop := f.assertNextManifest("toplvl-local", lt(updateCmd("echo hello world"), deps(f.JoinPath("foo/baz"), f.JoinPath("foo/a.txt"))))
```
nicks: it might be simpler if the fixture does the JoinPaths?

internal/tiltfile/tiltfile_test.go
```diff
@@ -5136,6 +5116,22 @@ func disablePush(disable bool) disablePushHelper {
...
 
+type updateCmdHelper struct {
+	cmd string
+}
+
+func updateCmd(cmd string) updateCmdHelper {
+	return updateCmdHelper{cmd}
+}
+
+type ltHelper struct {
```
nicks: can we spell this out as localTarget? i'm not sure i'd ever remember what lt meant without digging around. i would assume it meant "less than"

pkg/model/manifest.go
```diff
@@ -161,6 +161,8 @@ func (m Manifest) LocalPaths() []string {
 	switch di := m.deployTarget.(type) {
 	case DockerComposeTarget:
 		return di.LocalPaths()
+	case LocalTarget:
```
nicks: should we panic() in the default case, to prevent this from getting missed in the future?

PR #2733, opened 2020-01-07
tiltfile: ability to specify serve commands in tiltfile

jazzdan:
Note: this is backwards incompatible because I made a formerly required field optional. It would be possible to do this in other ways. For example, we could leave updateCmd as required and just let people who only want to do serve cmds leave it empty:

```python
local_resource("foo", "", serveCmd="rails s")
```

Or we could make it a boolean parameter, like so:

```python
local_resource("build_foo", "go install .")
local_resource("serve_foo", "./main", serve=True)
```

Or make it an entirely different function.

### DIFF
--- a/internal/tiltfile/local_resource.go
+++ b/internal/tiltfile/local_resource.go
@@ -41,7 +41,7 @@ func (s *tiltfileState) localResource(thread *starlark.Thread, fn *starlark.Buil
 		"resource_deps?", &resourceDepsVal,
 		"ignore?", &ignoresVal,
 		"auto_init?", &autoInit,
-		"serveCmd?", &serveCmdStr,
+		"serve_cmd?", &serveCmdStr,
 	); err != nil {
 		return nil, err
 	}
@@ -72,7 +72,7 @@ func (s *tiltfileState) localResource(thread *starlark.Thread, fn *starlark.Buil
 	serveCmd := model.ToShellCmd(serveCmdStr)
 
 	if updateCmd.Empty() && serveCmd.Empty() {
-		return nil, fmt.Errorf("local_resource must have an updateCmd and/or a serveCmd, but both were empty")
+		return nil, fmt.Errorf("local_resource must have a cmd and/or a serve_cmd, but both were empty")
 	}
 
 	res := localResource{
--- a/internal/tiltfile/local_resource.go
+++ b/internal/tiltfile/local_resource.go
@@ -35,13 +35,13 @@ func (s *tiltfileState) localResource(thread *starlark.Thread, fn *starlark.Buil
 
 	if err := s.unpackArgs(fn.Name(), args, kwargs,
 		"name", &name,
-		"updateCmd?", &updateCmdStr,
-		"serveCmd?", &serveCmdStr,
+		"cmd?", &updateCmdStr,
 		"deps?", &deps,
 		"trigger_mode?", &triggerMode,
 		"resource_deps?", &resourceDepsVal,
 		"ignore?", &ignoresVal,
 		"auto_init?", &autoInit,
+		"serveCmd?", &serveCmdStr,
 	); err != nil {
 		return nil, err
 	}
--- a/internal/tiltfile/local_resource.go
+++ b/internal/tiltfile/local_resource.go
@@ -72,13 +72,13 @@ func (s *tiltfileState) localResource(thread *starlark.Thread, fn *starlark.Buil
 	serveCmd := model.ToShellCmd(serveCmdStr)
 
 	if updateCmd.Empty() && serveCmd.Empty() {
-		return nil, fmt.Errorf("local_resource must have an updateCmd or a serveCmd, but both were empty")
+		return nil, fmt.Errorf("local_resource must have an updateCmd and/or a serveCmd, but both were empty")
 	}
 
 	res := localResource{
 		name:         name,
-		updateCmd:    model.ToShellCmd(updateCmdStr),
-		serveCmd:     model.ToShellCmd(serveCmdStr),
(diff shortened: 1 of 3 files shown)


landism:
> Note: this is backwards incompatible because I made a formerly required field optional

I was only able to think of one instance of backwards incompatibility - if the TIltfile specifies a named `cmd` arg to `local_resource`, which seems pretty easily handlable. Am I missing others?

internal/tiltfile/local_resource.go
```diff
@@ -66,9 +68,17 @@ func (s *tiltfileState) localResource(thread *starlark.Thread, fn *starlark.Buil
...
+	serveCmd := model.ToShellCmd(serveCmdStr)
+
+	if updateCmd.Empty() && serveCmd.Empty() {
+		return nil, fmt.Errorf("local_resource must have an updateCmd or a serveCmd, but both were empty")
+	}
+
 	res := localResource{
 		name:         name,
-		cmd:          model.ToShellCmd(cmd),
+		updateCmd:    model.ToShellCmd(updateCmdStr),
```
landism: can just use `updateCmd` and `serveCmd` here

internal/tiltfile/local_resource.go
```diff
@@ -66,9 +68,17 @@ func (s *tiltfileState) localResource(thread *starlark.Thread, fn *starlark.Buil
 		return nil, err
 	}
 
+	updateCmd := model.ToShellCmd(updateCmdStr)
+	serveCmd := model.ToShellCmd(serveCmdStr)
+
+	if updateCmd.Empty() && serveCmd.Empty() {
+		return nil, fmt.Errorf("local_resource must have an updateCmd or a serveCmd, but both were empty")
```
landism: maybe nitty, but maybe `and/or` or "at least one of updateCmd or serveCmd", to make clear it's an inclusive or?

internal/tiltfile/local_resource.go
jazzdan: whoops, yep that was a refactoring mistake

internal/tiltfile/local_resource.go
jazzdan: yes good call!

jazzdan:
@landism right because, and I should have pointed this out, I also changed the order of fields.

dbentley:
1) ugh, wish we had kwarg-only params in starlark
2) I think we should encourage using kwargs instead of positional
3) therefore, I wouldn't change the order of args
4) also, to be precise, changing the name of cmd to update cmd could break people who pass it as a kwarg

nicks:
+1 for changing the args parser to allow kwarg-only params. the starkit stuff we added lets you specify your own unpacker!

jazzdan:
I'm down to change the order back.

I'm down to rename the parameter back to cmd.

I'm down to implement a kwarg-only unpacker in a separate PR

internal/tiltfile/local_resource.go
```diff
@@ -34,12 +35,13 @@ func (s *tiltfileState) localResource(thread *starlark.Thread, fn *starlark.Buil
...
 	if err := s.unpackArgs(fn.Name(), args, kwargs,
 		"name", &name,
-		"cmd", &cmd,
+		"cmd?", &updateCmdStr,
 		"deps?", &deps,
 		"trigger_mode?", &triggerMode,
 		"resource_deps?", &resourceDepsVal,
 		"ignore?", &ignoresVal,
 		"auto_init?", &autoInit,
+		"serveCmd?", &serveCmdStr,
```
nicks: all args should be slug-cased, so this should be `serve_cmd`

nicks: (if you end up deciding to keep it as-is, ok, i'll stop commenting now!)

jazzdan: Nope this is great, thank you!

dbentley:
https://github.com/google/starlark-go/issues/249 discussed a bit; I think the unpack'ing for positional-only would be .5-1 days of work to get right, but would love to be surprised.

I think it'd be easy to get it accepted upstream, too.

internal/tiltfile/tiltfile_test.go
```diff
@@ -3934,15 +3977,15 @@ local_resource("toplvl-local", "echo hello world", ["foo/baz", "foo/a.txt"])
 	f.load()
 
 	f.assertNumManifests(2)
-	mNested := f.assertNextManifest("nested-local", lt(updateCmd("echo nested"), deps("nested/foo/bar", "nested/more_nested/repo")))
+	mNested := f.assertNextManifest("nested-local", localTarget(updateCmd("echo nested"), deps("nested/foo/bar", "nested/more_nested/repo")))
```
landism: 💯

internal/tiltfile/tiltfile_test.go
```diff
@@ -4745,11 +4788,13 @@ func (f *fixture) assertNextManifest(name model.ManifestName, opts ...interface{
...
 				case updateCmdHelper:
 					assert.Equal(f.t, matcher.cmd, lt.UpdateCmd.String())
+				case serveCmdHelper:
+					assert.Equal(f.t, matcher.cmd, lt.ServeCmd.String())
 				case depsHelper:
 					deps := f.JoinPaths(matcher.deps)
 					assert.ElementsMatch(f.t, deps, lt.Dependencies())
 				default:
-					f.t.Fatal("unknown matcher for local target")
+					f.t.Fatal(fmt.Sprintf("unknown matcher for local target %T", matcher))
```
landism: I think this can be `f.t.Fatalf`

PR #2735, opened 2020-01-08
internal/engine/local: controller for local serve

dbentley:
This is a draft.

Things that are missing that should be here before merging:
* integrate with DI/wire
* upper/reducer logic

For a next PR:
* actually run processes


internal/engine/local/controller.go
```diff
@@ -0,0 +1,171 @@
...
+				SequenceNum:  sequenceNum,
+				Status:       status,
+			})
+		}
+	}()
+
+	proc.doneCh = c.execer.Start(ctx, spec.ServeCmd, logger.Get(ctx).Writer(logger.InfoLvl), statusCh)
+}
+
+// lastProcess represents the last process
```
jazzdan: Can we expand on this comment? Why do we need a sentinel struct to know when we've reached the end of the processes?

dbentley: Oh, it's not a sentinel struct. It's where we store info about the most recent exec, if any. Ideas for better names?

jazzdan: Ohh I see. `mostRecentProcess`? Or the extra verbose, non-Go-like `mostRecentlyExecedProcess`?

dbentley: Went for currentProcess, but happy to change.

PR #2744, opened 2020-01-08
engine: simple local process execer

jazzdan:


internal/engine/local/execer.go
```diff
@@ -39,3 +39,43 @@ func fakeRun(ctx context.Context, cmd model.Cmd, w io.Writer, statusCh chan Stat
...
+		// TODO(dmiller): should this be different status than when the command fails? Unknown?
+		statusCh <- Error
+		return
+	}
+
+	statusCh <- Running
+
+	err = c.Wait()
+	if err != nil {
+		statusCh <- Error
```
landism: we should probably log `err` here

internal/engine/local/execer.go
```diff
@@ -39,3 +39,43 @@ func fakeRun(ctx context.Context, cmd model.Cmd, w io.Writer, statusCh chan Stat
...
+		return
+	}
+
+	statusCh <- Done
```
landism: In my mind, I figured we'd want to `statusCh <- Error` when the process exits, whether it exited with an error or not - if a service quits with exit code 0, it's still probably not what the user wanted?

At least, that seems true for the "run a service" case. I don't know if we have other use cases where it's more like a job than a service, but I think those would be taken care of by a local resource with a cmd and no serve_cmd. But I haven't put that much thought into it and maybe I missed something.

dbentley: I think the Execer should just say it finished (and perhaps we need to includ estatus code), and then it's up to the Controller how to turn that into something that makes sense for Tilt. I've found that's a decent way to separate responsibility.

jazzdan: I'll leave this as-is for now and we can come back to it later.

PR #2745, opened 2020-01-09
tiltfile: set max build slots from tiltfile [ch4065]

maiamcc:
by all rights, `state.MaxBuildSlots` should be `engine.MaxUpdateSlots` or something,
b/c I think we've been trying to move away from "build" and towards "update"--
I tried to keep the user-facing language consistent with "update", and am happy
to change backend language in a separate PR

### DIFF
--- a/internal/tiltfile/tiltfile_test.go
+++ b/internal/tiltfile/tiltfile_test.go
@@ -4345,21 +4345,21 @@ func TestMaxParallelUpdates(t *testing.T) {
 		{
 			name:                  "default max parallel updates",
 			tiltfile:              "print('hello world')",
-			expectedMaxBuildSlots: model.DefaultMaxBuildSlots,
+			expectedMaxBuildSlots: model.DefaultMaxParallelUpdates,
 		},
 		{
 			name:                  "set max parallel updates",
-			tiltfile:              "max_parallel_updates(42)",
+			tiltfile:              "update_settings(max_parallel_updates=42)",
 			expectedMaxBuildSlots: 42,
 		},
 		{
 			name:                "NaN error",
-			tiltfile:            "max_parallel_updates('boop')",
+			tiltfile:            "update_settings(max_parallel_updates='boop')",
 			expectErrorContains: "got string, want int",
 		},
 		{
 			name:                "must be positive int",
-			tiltfile:            "max_parallel_updates(-1)",
+			tiltfile:            "update_settings(max_parallel_updates=-1)",
 			expectErrorContains: "must be >= 1",
 		},
 	} {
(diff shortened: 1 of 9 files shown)


internal/tiltfile/tiltfile_test.go
```diff
@@ -4321,6 +4321,52 @@ local_resource('e', 'echo e')
 	}
 }
 
+func TestMaxParallelUpdates(t *testing.T) {
```
nicks: can you put this test in the subpackage? i don't want this test file to have the union of all extensions

internal/tiltfile/updatesettings/update_settings.go
```diff
@@ -0,0 +1,62 @@
...
+	"fmt"
+
+	"go.starlark.net/starlark"
+
+	"github.com/windmilleng/tilt/pkg/model"
+
+	"github.com/windmilleng/tilt/internal/tiltfile/starkit"
+)
+
+// Implements functions for dealing with Docker Prune settings.
```
nicks: :eyes:

pkg/model/build_settings.go
```diff
@@ -0,0 +1,11 @@
+package model
+
+var DefaultMaxBuildSlots = 1
+
+type UpdateSettings struct {
```
nicks: re: Update vs Build naming

if it's called "update" does that mean people will think they can set max_parallel_updates=1 and that will make pods come up in serial? (i.e., that it will force Tilt to wait until pod A is ready before it even starts building image B?) i'm not sure if it matters but it's something to watch for

maiamcc: oof. i... don't know. maybe? but on the flip-side, the word "build" is less and less accurate as we get into live_update, local_resource, etc. I think we've settled on "update" as the verb for that? but also 🤷‍♀️

internal/tiltfile/updatesettings/update_settings.go
```diff
@@ -0,0 +1,62 @@
...
+func NewExtension() Extension {
+	return Extension{}
+}
+
+func (e Extension) NewState() interface{} {
+	return model.DefaultUpdateSettings()
+}
+
+func (e Extension) OnStart(env *starkit.Environment) error {
+	return env.AddBuiltin("max_parallel_updates", e.maxParallelUpdates)
```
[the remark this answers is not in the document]
maiamcc: will do (see #2754 for a rename that will keep the user-facing naming more consistent)

pkg/model/build_settings.go
maiamcc: hmm yeah I'm not sure -- I'll make sure to clarify in the docs

PR #2747, opened 2020-01-09
engine: hook local resource serve controller up to upper

landism:
This is working well enough that we can run stuff. I've added some TODOs for bugs I noticed. It seems worth getting this in and then we can work on those in parallel.

### DIFF
--- a/internal/engine/local/controller.go
+++ b/internal/engine/local/controller.go
@@ -75,6 +75,7 @@ func (c *Controller) update(ctx context.Context, specs []ServeSpec, st store.RSt
 	// stop old ones
 	for _, name := range toStop {
 		c.stop(name)
+		delete(c.procs, name)
 	}
 
 	// now start them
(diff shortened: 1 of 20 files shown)


internal/engine/local/controller.go
```diff
@@ -131,29 +136,56 @@ func (c *Controller) start(ctx context.Context, spec ServeSpec, st store.RStore)
...
+			runtimeStatus := status.ToRuntime()
+			if runtimeStatus != "" {
+				// TODO(matt) when we get an error, the dot is red in the web ui, but green in the TUI
+				st.Dispatch(LocalServeStatusAction{
+					ManifestName: manifestName,
+					SequenceNum:  sequenceNum,
+					Status:       runtimeStatus,
+				})
+			}
+		case <-stoppedCh:
```
dbentley: Once it's closed, you[ll want to nil it out, or else you'll be at 100% CPU going through this loop?

landism: yeah, though this should be mitigated by it always happening in close proximity to the closing of statusCh

anyway, I've got another branch ready to get rid of stoppedCh and just use sequenceNum

PR #2758, opened 2020-01-09
max number of parallel updates is 3 by default [ch4065]

maiamcc:
here's the PR that will turn on parallel updates by default; merge it
whenever you like!


nicks:
LGTM!

gonna merge now so we can play around with it and see how it feels

PR #2771, opened 2020-01-13
web: modal should fit on screen when it's overflowing

### LINKED ISSUE #2760: long fatal errors do not show up correctly in the Tilt UI
Reported in the #tilt channel. If you have a fatal error message that's very long, it scrolls off the screen and isn't visible.

Screen Shot 2020-01-09 at 10 35 44 PM

(We eventually figured out that it was an issue with how their gcloud auth credentials were setup; the main bug here is that Tilt should display the full error)

jazzdan:
Fixes #2760

Screenshot:
Screen Shot 2020-01-13 at 5 06 26 PM

Video:
https://drive.google.com/file/d/1PzCvWW6KIQ5zZQB8kV42R4A4SnsLxxjs/view?usp=sharing

### DIFF
--- a/web/src/FatalErrorModal.scss
+++ b/web/src/FatalErrorModal.scss
@@ -9,6 +9,8 @@
   padding-bottom: $spacing-unit;
   padding-left: $spacing-unit;
   padding-right: $spacing-unit;
+  height: 80vh;
+  overflow-y: auto;
 }
 
 .FatalErrorModal-title {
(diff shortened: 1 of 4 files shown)


web/src/FatalErrorModal.tsx
```diff
@@ -11,8 +11,9 @@ type props = {
 
 export default class FatalErrorModal extends PureComponent<props> {
   render() {
+    let error = "ERROR\n".repeat(500)
```
jazzdan: whoops! one sec

web/src/FatalErrorModal.scss
```diff
@@ -9,6 +9,8 @@
   padding-bottom: $spacing-unit;
   padding-left: $spacing-unit;
   padding-right: $spacing-unit;
+  height: 80vh;
```
nicks: should this be max-height?

PR #2774, opened 2020-01-14
tiltfile: record required helm local subcharts as dependencies

### LINKED ISSUE #2114: `helm()` should watch files in `requirements.yaml`
As it says on the tin. It's confusing that Tilt doesn't understand this.

jazzdan:
Fixes #2114


jazzdan:
Ugh Windows build failure...

internal/tiltfile/helm_test.go
```diff
@@ -61,3 +61,36 @@ func TestHelmUnknownVersion(t *testing.T) {
 
 	assert.Equal(t, expected, actual)
 }
+
+const fileRequirementsYAML = `dependencies:
+  - name: foobar 
+    version: 1.0.1
+    repository: file://./foobar`
+
+func TestLocalSubchartFileDependencies(t *testing.T) {
```
maiamcc: i don't find this to be a super compelling test case. I'd like to see multiple dependencies,  some with `repository` values that match the prefix and some that don't, and when you check the output of `localSubchartDependencies`, assert the length as well to make sure that nothing got in there that shouldn't have.

maiamcc: oh huh i see, you tested things that don't match the prefix below.

internal/tiltfile/helm_test.go
```diff
@@ -61,3 +61,36 @@ func TestHelmUnknownVersion(t *testing.T) {
...
+}
+
+const remoteRequirementsYAML = `
+dependencies:
+- name: etcd
+  version: 0.6.2
+  repository: https://kubernetes-charts-incubator.storage.googleapis.com/
+  condition: etcd.deployChart`
+
+func TestSubchartRemoteDependencies(t *testing.T) {
```
maiamcc: this test name wasn't descriptive for me. maybe `TestLocalSubchartDependenciesDoesntReturnRemoteDependencies`?

go.mod
```diff
@@ -130,23 +109,23 @@ require (
...
+	k8s.io/klog v1.0.0
+	k8s.io/kubectl v0.0.0-20191016120415-2ed914427d51
 	sigs.k8s.io/kustomize v2.0.3+incompatible
 	sigs.k8s.io/yaml v1.1.0
 	vbom.ml/util v0.0.0-20180919145318-efcd4e0f9787 // indirect
 )
 
 replace (
 	github.com/Azure/go-autorest => github.com/Azure/go-autorest v12.4.3+incompatible
+	github.com/docker/docker => github.com/moby/moby v0.7.3-0.20190826074503-38ab9da00309
```
nicks: i don't think this change is right? and it breaks things when you try to import the package. if we want to depend on this version of docker, shouldn't that be in the stanza above, rather than overriding it here?

internal/tiltfile/helm.go
```diff
@@ -2,16 +2,24 @@ package tiltfile
...
 	"fmt"
+	"io/ioutil"
+	"os"
 	"os/exec"
+	"path/filepath"
 	"regexp"
 	"runtime"
 	"strings"
+
+	"github.com/ghodss/yaml"
```
nicks: i'm a little bit concerned about how this other yaml library slipped in, as opposed to the one kubernetes uses

internal/tiltfile/helm.go
```diff
@@ -103,3 +111,37 @@ func unableToFindHelmErrorMessage() error {
 
 	return fmt.Errorf("Unable to find Helm installation. Make sure `%s` is on your $PATH.", binaryName)
 }
+
+func localSubchartDependenciesFromPath(chartPath string) ([]string, error) {
+	requirementsPath := filepath.Join(chartPath, "requirements.yaml")
+	_, err := os.Stat(requirementsPath)
```
nicks: idiomatically, you'd just check the os.IsNotExist error from ReadFile, rather than doing a separate Stat syscall.

internal/tiltfile/helm.go
```diff
@@ -103,3 +111,37 @@ func unableToFindHelmErrorMessage() error {
...
+	if err != nil {
+		return deps, err
+	}
+
+	return localSubchartDependencies(dat)
+}
+
+func localSubchartDependencies(dat []byte) ([]string, error) {
+	var deps []string
+	var metadata chart.Metadata
```
nicks: have you considered just writing a new struct with the fields you want to grab?

it seems weird to import Helm just to get 2 fields from a struct

internal/tiltfile/helm.go
jazzdan: Fun fact: it looks like kubernetes/kubernetes uses three different YAML libraries: https://github.com/kubernetes/kubernetes/blob/master/go.mod

jazzdan: (And to be clear: this is one of them)

nicks: :open_mouth:

PR #2775, opened 2020-01-14
web: cache the LogLines to reduce GC thrashing

nicks:
Hello @maiamcc, @hyu,

Please review the following commits I made in branch nicks/logcache:

2ee0653bdbbd1011c83e1c82c17c1172f2c468c4 (2020-01-14 14:56:22 -0500)
web: cache the LogLines to reduce GC thrashing

### DIFF
--- a/web/src/LogStore.ts
+++ b/web/src/LogStore.ts
@@ -14,6 +14,7 @@ type LogSpan = {
   manifestName: string
   lastSegmentIndex: number
   firstSegmentIndex: number
+  lastGlobalLineIndex: number
 }
 
 class LogSegment {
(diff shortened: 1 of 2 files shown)


web/src/LogStore.ts
```diff
@@ -243,57 +264,64 @@ class LogStore {
...
+    span: LogSpan
+  ): LogLine {
+    let spanId = segment.spanId
+    let currentLine = {
+      manifestName: span.manifestName,
+      text: segment.text,
+      level: segment.level,
+    }
+
+    // If this segment is not complete, run ahead and try to complete it.
```
maiamcc: i'm confused why a comment about "if segment is not complete" is right above "if segment.IsComplete()..." -- this comment seems to refer to different code, and should be nearer to the code it refers to

nicks: good question! there's an early exit on line 297. I'll clarify the comment a bit

PR #2777, opened 2020-01-14
engine: s/building/updating in buildcontroller log

jazzdan:
Mostly a PR for discussion. Seems to make sense in all contexts right now. WDYT?

### DIFF
--- a/internal/engine/buildcontroller.go
+++ b/internal/engine/buildcontroller.go
@@ -140,14 +140,14 @@ func (c *BuildController) logBuildEntry(ctx context.Context, entry buildEntry) {
 	l := logger.Get(ctx)
 	var prefix string
 	if firstBuild {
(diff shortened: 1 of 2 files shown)


jazzdan:
@dbentley ping on this, also adding @l-korbes as a reviewer

l-korbes:
Copied from Slack:

> Most tools call it by two separate steps, building and deploying. Since we're merging the two into a general term we don't have to follow those, and IMO "updating" is clearer and more intuitive to end users.

So +1 here.

hyu:
As part of my work on Update vs LiveUpdate, I'm adopting your proposed change, @jazzdan. (And also making some other changes to make this line appear differently in the Web UI.)

Do you mind closing this PR to avoid the possibly of mild merge conflicts?

jazzdan:
You got it @hyu!

PR #2781, opened 2020-01-15
web: disable smoothing and calculate autoscroll without forcing layouts

nicks:
Hello @hyu, @jazzdan,

Please review the following commits I made in branch nicks/windowing2:

94a56d14f2d3da197dc0431903fa14f37280de4e (2020-01-15 09:32:16 -0500)
web: disable smoothing and calculate autoscroll without forcing layouts
These two optimizations speed up scrolling and responsiveness dramatically

4e3e57bf8c8788754188f6a9bbce44d6cc0085aa (2020-01-14 19:27:30 -0500)
web: reset autoscroll when switching views. Fixes https://github.com/windmilleng/tilt/issues/2571

### DIFF
--- a/web/src/LogPane.tsx
+++ b/web/src/LogPane.tsx
@@ -9,8 +9,6 @@ import { SizeUnit, Width } from "./style-helpers"
 import findLogLineID from "./findLogLine"
 import styled from "styled-components"
 
-const WHEEL_DEBOUNCE_MS = 250
-
 type LogPaneProps = {
   manifestName: string
   logLines: LogLine[]
@@ -21,11 +19,6 @@ type LogPaneProps = {
   highlight: SnapshotHighlight | null | undefined
   modalIsOpen: boolean
   isSnapshot: boolean
-}
-
-type LogPaneState = {
-  autoscroll: boolean
-  lastWheelEventTimeMs: number
 }
 
 type LogLineComponentProps = {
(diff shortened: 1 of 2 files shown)


nicks:
Background on this:

I spent a bunch of time experimenting with react-window and react-virtualized for rendering the logs. These two packages try to optimize rendering when you have long lists of things, by only rendering the parts of the list that are currently on-screen.

But when I got them working, rendering got much slower!! Which was a big hint that we had other problems that were causing pathological UI performance issues. I spent a bunch of time digging into the chrome rendering debug tools, and the big things I found were:

1) Autoscroll detection causing a lot of forced layouts
2) Long paint times when scrolling quickly, leaving the screen blank for long periods of time

I rewrote the autoscroll detection to only use window.pageYOffset (which fixes (1)), and turned off some of the font smoothing (which fixes (2)), and I'm much happier with the performance now

hyu:
Also, TIL that font-smoothing can have a dramatic performance impact under certain circumstances! 🤯

nicks:
@hyu what project were you testing with?

web/src/LogPane.tsx
```diff
@@ -227,36 +219,54 @@ class LogPane extends Component<LogPaneProps, LogPaneState> {
       }
     }
   }
-  private handleWheel(event: WheelEvent) {
-    if (event.deltaY < 0) {
-      this.setState({ autoscroll: false, lastWheelEventTimeMs: Date.now() })
+
+  private onScroll() {
+    return
```
nicks: oops :eyes: :laughing: fixed

nicks:
also re: fonts - ya, this was an issue we ran into a bunch at Medium. I have a couple friends who know a lot about the deep internals of how font rendering works. 

My layman's mental model is that they have a "slow path" that's guaranteed to be correct, and then the team goes through and hyper-optimizes "common cases" to use hardware better, etc. Usually if something falls onto the slow path it's because of some combination of css font rendering features that they've failed to optimize for yet (e.g., smoothing paired with pre-wrap)

PR #2792, opened 2020-01-16
build: Add a minimum viable docker-push-progress. Fixes https://github.com/windmilleng/tilt/issues/1983

nicks:
Hello @maiamcc, @hyu,

Please review the following commits I made in branch nicks/push2:

183432229041f42e85e6d7039d33a2fe2aeb2350 (2020-01-16 18:25:05 -0500)
build: Add a minimum viable docker-push-progress. Fixes https://github.com/windmilleng/tilt/issues/1983

### DIFF
--- a/internal/build/image_builder.go
+++ b/internal/build/image_builder.go
@@ -328,6 +328,8 @@ func cleanupDockerBuildError(err string) string {
 	return ret
 }
 
+type dockerMessageID string
+
 // Docker API commands stream back a sequence of JSON messages.
 //
 // The result of the command is in a JSON object with field "aux".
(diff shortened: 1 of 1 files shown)


nicks:
Here's what it looks like:
https://cloud.tilt.dev/snapshot/AdrCh-IL5AWvhEMrEY4=

Screenshot from 2020-01-16 18-27-12

This is not great, and not where we want to go long-term, but it seems like a simple thing to add based on what Han and I puzzled out today.

internal/build/image_builder.go
```diff
@@ -341,6 +341,9 @@ func readDockerOutput(ctx context.Context, reader io.Reader, writer io.Writer) (
 	span, ctx := opentracing.StartSpanFromContext(ctx, "daemon-readDockerOutput")
 	defer span.Finish()
 
+	progressLastPrinted := make(map[string]time.Time)
```
jazzdan: Nit: It would be helpful for code comprehensibility if the key here was an opaque type that described what this string is. I was thinking like:

```golang
type buildkitMessageID = string
progressLastPrinted := make(map[buildkitMessageID]time.Time)
```

nicks: done! tho it's worth pointing out that this isn't from buildkit, it's from docker push

PR #2795, opened 2020-01-17
tiltfile: make kustomize fallback more clear with log statement

jazzdan:


internal/tiltfile/files.go
```diff
@@ -88,6 +88,7 @@ func (s *tiltfileState) kustomize(thread *starlark.Thread, fn *starlark.Builtin,
 
 	_, err = exec.LookPath(cmd[0])
 	if err != nil {
+		s.logger.Infof("Falling back to `kubectl kustomize` since `kubectl` was not found in PATH")
```
khos2ow: Isn't this a typo in the log? I think it should read:

```
Falling back to `kubectl kustomize` since `kustomize` was not found in PATH
```

 If it reaches here, `kustomize` was not found ([this](https://github.com/windmilleng/tilt/blob/master/internal/tiltfile/files.go#L93)).

jazzdan: Good catch! Will fix, thanks @khos2ow!

PR #2800, opened 2020-01-17
engine: fixed typo where the word `with` was repeated twice

robbert229:
Fixed a typo. The word `with` is repeated twice in one of the log statements. 

```failed with with exit code: 1``` -> ```failed with exit code: 1 ```


nicks:
thanks! lgtm

PR #2808, opened 2020-01-22
tiltfile: add quiet parameter to local

### LINKED ISSUE #2807: Add quiet flag to local()
This would allow local() commands that output secret values to not record them in snapshottable output.

tmc:
Fixes #2807


jazzdan:
Thanks @tmc!

PR #2809, opened 2020-01-22
 ci: enable staticcheck, fix problems that were surfaced

jazzdan:


internal/build/tar.go
```diff
@@ -45,7 +45,7 @@ func clearUIDAndGID(h *tar.Header) {
 }
 
 func (a *ArchiveBuilder) archiveDf(ctx context.Context, df dockerfile.Dockerfile) error {
-	span, ctx := opentracing.StartSpanFromContext(ctx, "daemon-archiveDf")
+	span, _ := opentracing.StartSpanFromContext(ctx, "daemon-archiveDf")
```
nicks: can we sink the ctx instead? i.e.,

```
_ = ctx
```

Overwriting the ctx is an important pattern to preserve i think

Makefile
```diff
@@ -117,7 +117,7 @@ verify_goimports:
 benchmark:
 	go test -mod vendor -run=XXX -bench=. ./...
 
-errcheck:
+golangci-lint:
```
landism: is there much downside to folding this into the lint target? having both `lint` and `golangci-lint` feels weird, and I always get annoyed trying to remember what wordbreaking they used (e.g., go-lang-ci-lint or golang-ci-lint or golang-cilint...four words, one hyphen!)

internal/engine/upper_test.go
```diff
@@ -366,6 +366,7 @@ func TestUpper_UpK8sEntityOrdering(t *testing.T) {
 	defer f.TearDown()
 
 	postgresEntities, err := k8s.ParseYAMLFromString(testyaml.PostgresYAML)
+	require.NoError(t, err)
```
landism: 🙌

internal/sliceutils/sliceutils.go
```diff
@@ -34,10 +34,8 @@ func BulletedIndentedStringList(list []string) string {
 		return ""
 	}
 
-	result := make([]string, len(list))
-	for i, s := range list {
-		result[i] = s
-	}
+	var result = make([]string, len(list))
```
landism: mostly just asking so we don't have multiple patterns for the same thing - I think most of our code uses `result := append([]string{}, list...)` to make a copy of a slice - any reason for a difference here?

Makefile
jazzdan: yeah I plan on folding it in soon. We can actually replace all of the existing lint stuff with `golangci-lint`.

internal/sliceutils/sliceutils.go
jazzdan: It's a lint rule that was only triggered here. I think in most other situations we then modify the copied array. I think this is clearer about what's going on.

PR #2812, opened 2020-01-22
web: Implement progress logs

nicks:
Hello @hyu, @maiamcc,

Please review the following commits I made in branch nicks/ch4296/progress:

53f012e4b9b917242b971cf34ef2c045f43e9426 (2020-01-22 16:03:31 -0500)
web: Implement progress logs


nicks:
this is a general API so that you can emit a log that goes back in time and updates older log lines. there are a couple places where i think we'll want to use this. I think we'll also want to tweak the UX a bit (e.g., making the line glow when it updates)

PR #2823, opened 2020-01-23
local_resource: kill pgid on shutdown, add integration test

landism:


landism:
I haven't spent the time to understand why, but it looks like other programs of our nature kill the pgid instead of the pid, and that made the test pass on both linux and macos.

PR #2826, opened 2020-01-23
procutil: don't send signals to nil process

jazzdan:
Before: 

If you `tilt up` this branch Tilt would segfault
https://github.com/windmilleng/enhance/compare/dmiller/crash-tilt?expand=1

After:
It starts up fine and doesn't segfault

### DIFF
--- a/internal/engine/local/execer.go
+++ b/internal/engine/local/execer.go
@@ -141,16 +141,16 @@ func processRun(ctx context.Context, cmd model.Cmd, w io.Writer, statusCh chan s
 	c.Stderr = w
 	c.Stdout = w
 
-	errCh := make(chan error)
+	err := c.Start()
+	if err != nil {
+		return
+	}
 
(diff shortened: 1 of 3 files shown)


internal/engine/local/execer.go
```diff
@@ -141,16 +141,16 @@ func processRun(ctx context.Context, cmd model.Cmd, w io.Writer, statusCh chan s
 	c.Stderr = w
 	c.Stdout = w
 
-	errCh := make(chan error)
+	err := c.Start()
+	if err != nil {
+		return
```
landism: We should log `err` here and have a test for it. It looks like I only tested logging exit errors.

jazzdan: Oh yeah good call, I meant to do that.

landism: oh, I guess splitting this out from errCh means we no longer have one handler for Start and Wait errors. blergh.

jazzdan: Yeah, but we could abstract out a function. :)

PR #2828, opened 2020-01-24
web: implement a simple windowing mechanism to improve UI perf

nicks:
Hello @hyu, @jazzdan,

Please review the following commits I made in branch nicks/ch4315/windowing:

ba0a276347a6ce79a1ba4948d385d2aa98db61bf (2020-01-23 20:24:26 -0500)
web: implement a simple windowing mechanism to improve UI perf


web/src/LogPane.tsx
```diff
@@ -21,6 +21,15 @@ type LogPaneProps = {
...
 }
 
+type LogPaneState = {
+  // The number of log lines to display
+  renderWindow: number
+}
+
+const renderWindowDefault = 50
+const renderWindowStep = 50
+const blankLogLineHeight = 30
```
hyu: Nit: We could use styled-components for the logLine CSS, and use this const value, to ensure it's accurate. For now, though, maybe a comment to note that it's just a value that tracks the computed height of each log line? (Which is currently based on font size and line height.)

nicks: That wouldn't work, because a log line can have multiple lines of text, so the height isn't fixed

nicks:
re: blank artifacts: ya, that is, in many ways, the whole point of this PR. If you forced the browser to render every row off-screen, on the off-chance that the user might scroll fast enough to see it, then you would necessarily make the overall experience much slower.

nicks:
i tweaked the algorithm a bit to expand the render window a bit faster, which should address some of the blank artifact issues.

web/src/LogPane.tsx
nicks: added come comments for clarity

PR #2849, opened 2020-01-27
web: fix an off-by-one error in how we show progress logs

nicks:
Hello @hyu, @maiamcc,

Please review the following commits I made in branch nicks/progress2:

7956c7eb14137b93e16ceca4dd625105687b2736 (2020-01-27 15:42:29 -0500)
web: fix an off-by-one error in how we show progress logs

### DIFF
--- a/web/src/LogStore.ts
+++ b/web/src/LogStore.ts
@@ -191,7 +191,7 @@ class LogStore {
     }
 
     // Iterate backwards and figure out which line to overwrite.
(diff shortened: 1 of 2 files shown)


web/src/LogStore.ts
```diff
@@ -111,6 +111,7 @@ class LogStore {
   }
 
   append(logList: Proto.webviewLogList) {
+    debugger
```
hyu: 👀

nicks: :fearful:

PR #2854, opened 2020-01-28
buildcontrol: don't parallelize local_resource with other commands

nicks:
Hello @landism, @maiamcc,

Please review the following commits I made in branch nicks/ch4318/locals:

958d40ba14fce685182488a2a3e4e309c40131af (2020-01-27 21:12:32 -0500)
buildcontrol: don't parallelize local_resource with other commands

### DIFF
--- a/internal/engine/buildcontrol/build_control_test.go
+++ b/internal/engine/buildcontrol/build_control_test.go
@@ -1,6 +1,7 @@
 package buildcontrol
 
 import (
+	"fmt"
 	"testing"
 	"time"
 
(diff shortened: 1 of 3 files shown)


internal/engine/buildcontrol/build_control_test.go
```diff
@@ -29,6 +30,37 @@ func TestNextTargetToBuildDoesntReturnCurrentlyBuildingTarget(t *testing.T) {
 	f.assertNoTargetNextToBuild()
 }
 
+func TestPodsDisableLocalScheduling(t *testing.T) {
```
maiamcc: This test title is v confusing to me b/c I'm not used to thinking of k8s manifests as "pods". Maybe something like: `TestCurrentlyBuildingK8sResourceDisablesLocalScheduling` and consider renaming the k8s manifests

internal/engine/buildcontrol/build_control_test.go
```diff
@@ -29,6 +30,37 @@ func TestNextTargetToBuildDoesntReturnCurrentlyBuildingTarget(t *testing.T) {
...
+	f.assertNextTargetToBuild("local1")
+
+	pod1.State.CurrentBuild = model.BuildRecord{StartTime: time.Now()}
+	f.assertNextTargetToBuild("pod2")
+
+	pod2.State.CurrentBuild = model.BuildRecord{StartTime: time.Now()}
+	f.assertNoTargetNextToBuild()
+}
+
+func TestLocalDisablesPodScheduling(t *testing.T) {
```
maiamcc: same comment as above (but even more so, because "pod scheduling" is its own thing 😅 )

internal/engine/buildcontrol/build_control_test.go
nicks: done

internal/engine/buildcontrol/build_control_test.go
nicks: done

nicks:
ya, i expect we'll want to revisit parallelization of local_resources, but we'd need some basic primitives to allow synchronization (e.g., when people know that two local_resources shouldn't run at the same time because they'd step on each other's output)

PR #2860, opened 2020-01-28
logstore: Return structured log lines for printing to stdout

nicks:
Hello @jazzdan, @maiamcc,

Please review the following commits I made in branch nicks/continuingstring:

902a694256e1f2cfc52d3e6731360bdf7e767390 (2020-01-28 13:15:50 -0500)
logstore: Return structured log lines for printing to stdout

### DIFF
--- a/pkg/model/logstore/logstore.go
+++ b/pkg/model/logstore/logstore.go
@@ -103,6 +103,20 @@ func segmentsFromBytes(spanID SpanID, time time.Time, level logger.Level, fields
 	return segments
 }
 
+type LogLine struct {
+	Text       string
+	SpanID     SpanID
+	ProgressID string
+}
+
+func linesToString(lines []LogLine) string {
+	sb := strings.Builder{}
+	for _, line := range lines {
+		sb.WriteString(line.Text)
+	}
+	return sb.String()
+}
+
 type LogEvent interface {
 	Message() []byte
 	Time() time.Time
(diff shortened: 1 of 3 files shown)


pkg/model/logstore/logstore.go
```diff
@@ -532,10 +568,41 @@ func (s *LogStore) startAndLastIndices(spans map[SpanID]*Span) (startIndex, last
...
 }
 
-func (s *LogStore) logHelper(spansToLog map[SpanID]*Span, showManifestPrefix bool) string {
+type logOptions struct {
+	spans                       map[SpanID]*Span
+	showManifestPrefix          bool
+	skipFirstLineManifestPrefix bool
+}
+
+func (s *LogStore) logHelper(options logOptions) string {
```
jazzdan: nit: This is pre-existing but I don't think `logHelper` is a good name for this function. `logOptionsToString` maybe?

pkg/model/logstore/logstore.go
```diff
@@ -532,10 +568,41 @@ func (s *LogStore) startAndLastIndices(spans map[SpanID]*Span) (startIndex, last
...
+	return linesToString(s.logHelperLines(options))
+}
+
+// Returns a sequence of lines, including trailing newlines.
+func (s *LogStore) logHelperLines(options logOptions) []LogLine {
```
jazzdan: nit: Same thing here. `logOptionsToLine`?

pkg/model/logstore/logstore.go
nicks: i changed it to `toLogString`... didn't want to imply that we were converting the logOptions struct itself into a string

pkg/model/logstore/logstore.go
nicks: done

Issue #2861, opened 2020-01-28
Occasional "bind: address already in use" errors on Tilt startup


andymartin-sch:
Over the last few months, our team has noticed that rarely but occasionally Tilt will fail to start up with the error:
```
Error: listen tcp 0.0.0.0:10350: bind: address already in use
```

It happens somewhere between 1 and 5% of the time we call `tilt up` and at least 5 engineers have reported seeing it on their workstations at least once. We have never been able to find another process listening to port 10350 after the error is seen and the subsequent `tilt up` call generally always succeeds.

After [some discussion in the Tilt slack channel](https://app.slack.com/client/T09NY5SBT/CESBL84MV/thread/CESBL84MV-1580227203.120300), it looks like this is a TOCTOU-like bug but one where the _check_ occasionally interferes with the _use_ of the address being bound to.

I was able to use auditd to confirm that the `bind` syscall is being called twice per `tilt up` invocation, and that it's the second call that fails with `EADDRINUSE` when the bind error is seen:
```
type=PROCTITLE msg=audit(01/28/2020 18:04:55.198:235752) : proctitle=tilt up --hud=false --no-browser 
type=SOCKADDR msg=audit(01/28/2020 18:04:55.198:235752) : saddr={ fam=inet laddr=127.0.0.1 lport=10350 } 
type=SYSCALL msg=audit(01/28/2020 18:04:55.198:235752) : ... tty=pts2 ses=4524 comm=tilt exe=/usr/local/bin/tilt key=(null) 
----
...cutting one unrelated event out...
----
type=PROCTITLE msg=audit(01/28/2020 18:04:55.198:235754) : proctitle=tilt up --hud=false --no-browser 
type=SOCKADDR msg=audit(01/28/2020 18:04:55.198:235754) : saddr={ fam=inet laddr=127.0.0.1 lport=10350 } 
type=SYSCALL msg=audit(01/28/2020 18:04:55.198:235754) : arch=x86_64 syscall=bind success=no exit=EADDRINUSE(Address already in use)  ... tty=pts2 ses=4524 comm=tilt exe=/usr/local/bin/tilt key=(null) 
```

On my Ubuntu 18.04 machine, I can reliably reproduce the issue with the following bash command:
```
$ while true; do timeout 1s tilt up --hud=false --no-browser | grep "bind"; echo "could not reproduce. trying again..."; done |& tee out.log
could not reproduce. trying again...
could not reproduce. trying again...
could not reproduce. trying again...
[...]

PR #2862, opened 2020-01-28
allow extra time for OS to free bind address after IsBindAddrFree check

andymartin-sch:
See #2861 for more info.

This is a kludgy workaround -- I was still able to reproduce the underlying issue after sleeping for 10ms but have not seen it in ~2k `tilt up` executions after bumping the sleep time up to 100ms.

Some other options might be:
- to add the sleep directly to `IsBindAddrFree` [here](https://github.com/windmilleng/tilt/blob/8ea4b95e0f1fb39203faa2e95537337a211c19a5/internal/network/port.go#L25)
- eliminating all uses of `IsBindAddrFree` and always gracefully catching/handling the "address already in use" errors instead

No worries at all if you'd like to go with one of the above approaches (or something else entirely) instead of this PR.

### DIFF
--- a/internal/hud/server/controller.go
+++ b/internal/hud/server/controller.go
@@ -110,6 +110,10 @@ func (s *HeadsUpServerController) OnChange(ctx context.Context, st store.RStore)
 		return
 	}
 
+	// We need to pause for a bit to ensure the OS has a chance to free the socket
+	// from the test above (https://github.com/windmilleng/tilt/issues/2861)
+	time.Sleep(100 * time.Millisecond)
+
 	httpServer := &http.Server{
 		Addr:    network.BindAddr(string(s.host), int(s.port)),
 		Handler: http.DefaultServeMux,


andymartin-sch:
I'm going to hold off messing around with debugging the issue with the CircleCI test until I hear whether or not this is the approach you all want to go with

jazzdan:
Hey @andymartin-sch, don't worry about the integration test failure, they don't pass when kicked off by people outside of our org.

jazzdan:
^ scratch that, apparently my information is outdated

nicks:
The integration test failure is semi-legitimate. 

Basically, your change delays the server initialization.

tilt_args_test.go has a race condition because it doesn't wait for server initialization. That's a bug we should fix (see https://github.com/windmilleng/tilt/pull/2867). The race was unlikely to happen before, but with your change it happens more often.

andymartin-sch:
Got it. That was my guess but I wasn't sure. Thanks for confirming!

nicks:
ok, the integration test fix is in, but i think you need to rebase your changes on top of it

i also liked your suggestion of removing IsBindAddrFree and just handing the "address already in use" errors instead, so you could also wait on https://github.com/windmilleng/tilt/pull/2874

A lot of the network.IsBindAddrFree code is leftover from the days when we didn't have good internal APIs for shutting down Tilt gracefully, so the code for listening the code for error-handling lived in different places.

andymartin-sch:
Yeah I think https://github.com/windmilleng/tilt/pull/2874 looks like a better solution than this PR - will leave it open for you all to decide but I'm totally in favor of going with that one instead.

Thank you for looking into this so quickly!

andymartin-sch:
I verified #2874 resolves the underlying issue (#2861) and left some details in [this comment](https://github.com/windmilleng/tilt/issues/2861#issuecomment-579946982).

Going to close this out - thanks again!

PR #2865, opened 2020-01-28
circleci: cache golangci-lint parsing

jazzdan:


landism:
I don't have a great understanding of circleci caching, but IIUC it's pretty limited and doesn't really work here, and even if it does, we should write comments explaining what's going on (I'm not certain I know what's going on).

> Cache is immutable on write: once a cache is written for a particular key like node-cache-master, it cannot be written to again.

My understanding is the first time we run this we'll cache golangci-lint output with whatever the state of master is at the time, and then it'll just get progressively slower or maybe entirely ineffective the more master changes from that point. Or maybe there's a decent chunk of the golangci-lint stuff that is just normal init stuff and independent of our code and this will be useful caching forever? It might be worth a comment in the golangci-lint yaml that we should bump the cache key when changing it?

The [docs](https://circleci.com/docs/2.0/caching/#writing-to-the-cache-in-workflows) talk about hashing the input and putting that in the cache key, to get a better hit rate, but since our input is "all of the go files", it just seems like we'd be invalidating it constantly.

jazzdan:
Oh, I didn't realize that @landism. That's unfortunate, let me think about it.

PR #2867, opened 2020-01-28
integration: fix a race condition in the tilt args integration test

nicks:
Hello @landism, @jazzdan,

Please review the following commits I made in branch nicks/argsfailure:

e6102eb5983a0a4e15381940278dce5b75dd9765 (2020-01-28 18:26:52 -0500)
integration: fix a race condition in the tilt args integration test

### DIFF
--- a/integration/tilt_args_test.go
+++ b/integration/tilt_args_test.go
@@ -3,6 +3,7 @@
 package integration
 
 import (
+	"fmt"
 	"testing"
 	"time"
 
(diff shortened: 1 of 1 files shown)


nicks:
another option would be to add some sort of a health check.

integration/tilt_args_test.go
```diff
@@ -22,8 +23,19 @@ func TestTiltArgs(t *testing.T) {
 
 	f.logs.Reset()
 
+	// Just run
```
landism: not sure what this means - is it cruft?

integration/tilt_args_test.go
```diff
@@ -22,8 +23,19 @@ func TestTiltArgs(t *testing.T) {
...
+
 	err = f.tilt.Args([]string{"bar"}, f.LogWriter())
-	require.NoError(t, err)
+	if err != nil {
+		// Currently, Tilt starts printing logs before the webserver has bound to a port.
+		// If this happens, just sleep for a second and try again.
+		duration := 2 * time.Second
+		fmt.Printf("Error setting args. Sleeping (%s): %v\n", duration, err)
+
+		time.Sleep(2 * time.Second)
```
landism:
```suggestion
		time.Sleep(duration)
```

integration/tilt_args_test.go
nicks: maybe it was trying to be motivational? :joy:

integration/tilt_args_test.go
nicks: done

PR #2871, opened 2020-01-29
tiltfile: error in default_registry if registry is not canonical

### LINKED ISSUE #2768: cryptic 'repository name must be canonical' error
Repro steps:
1) In any Tiltfile, add:

```
default_registry('rose-quartz')
```

2) Start a `kind` cluster
3) Run `tilt up`

Expected result:
A sensible error + stack trace pointing to where the bad default registry was set

Actual result:
https://cloud.tilt.dev/snapshot/AYDc3-ELDA5kK4OKJFs=

"repository name must be canonical" with no context

Thanks to Ralph Squillace on Twitter for reporting! https://twitter.com/ralph_squillace/status/1216507879118770176

jazzdan:
This causes a stack trace to be printed during Tiltfile execution
which makes it more obvious where this error is coming from.

Previously the error would only surface when the default registry
was being transformed in to the engine state model, at which point
we don't know what line it occurred on.

Fixes #2768

### DIFF
--- a/internal/container/default_registry.go
+++ b/internal/container/default_registry.go
@@ -5,6 +5,7 @@ import (
 	"regexp"
 
 	"github.com/docker/distribution/reference"
+	"github.com/pkg/errors"
 )
 
 // The Host of a container registry where we can push images.
@@ -30,7 +31,7 @@ func ReplaceRegistry(defaultRegistry Registry, rs RefSelector) (reference.Named,
 	newNs := fmt.Sprintf("%s/%s", defaultRegistry, escapeName(rs.RefFamiliarName()))
 	newN, err := reference.ParseNamed(newNs)
 	if err != nil {
-		return nil, fmt.Errorf("Error parsing %s after applying default registry %s: %w", newNs, defaultRegistry, err)
+		return nil, errors.Wrapf(err, "Error parsing %s after applying default registry %s", newNs, defaultRegistry)
 	}
 
 	return newN, nil
--- a/internal/container/default_registry.go
+++ b/internal/container/default_registry.go
@@ -30,7 +30,7 @@ func ReplaceRegistry(defaultRegistry Registry, rs RefSelector) (reference.Named,
 	newNs := fmt.Sprintf("%s/%s", defaultRegistry, escapeName(rs.RefFamiliarName()))
 	newN, err := reference.ParseNamed(newNs)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("Error parsing %s after applying default registry %s: %w", newNs, defaultRegistry, err)
 	}
 
 	return newN, nil
--- a/internal/tiltfile/tiltfile_test.go
+++ b/internal/tiltfile/tiltfile_test.go
@@ -2560,8 +2560,8 @@ func TestTwoDefaultRegistries(t *testing.T) {
 	defer f.TearDown()
 
 	f.file("Tiltfile", `
-default_registry("foo")
-default_registry("bar")`)
+default_registry("gcr.io/foo")
+default_registry("gcr.io/bar")`)
 
 	f.loadErrString("default registry already defined")
 }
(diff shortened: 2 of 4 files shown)


internal/tiltfile/tiltfile_test.go
```diff
@@ -2588,13 +2589,13 @@ func TestDefaultRegistryAtEndOfTiltfile(t *testing.T) {
 	f.file("Tiltfile", `
 docker_build('gcr.io/foo', 'foo')
 k8s_yaml('foo.yaml')
-default_registry('bar.com')
+default_registry('bar.com/baz')
```
nicks: why did this change? is bar.com not a valid registry?

jazzdan: It isn't, see all of the tests here:

https://github.com/moby/moby/blob/master/registry/registry_test.go#L309

landism: it sure looks like it is? https://github.com/moby/moby/blob/master/registry/registry_test.go#L415

or https://docs.docker.com/registry/ has
`docker push localhost:5000/myfirstimage`

IIUC this change is validating the registry name by parsing it as an image name, which probably isn't what we want. Maybe we could append a dummy image name?

jazzdan: ah yeah, okay. I wonder if there's a function that validates a repository...

nicks: ya, i think you want to use the DomainRegexp here: https://godoc.org/github.com/docker/distribution/reference#pkg-variables

nicks: actually, scratch that, because i guess you can also have example.com/my-bucket be a registry?

landism: I think it just has to contain a `.` or a `:` to avoid the error: https://github.com/windmilleng/tilt/blob/6297a6f3aeb93ee5a8dcaaed5921acfe67455060/vendor/github.com/docker/distribution/reference/normalize.go#L64-L63

but even then it's complicated:
1. it might [fail later if it the resulting name is too long](https://github.com/windmilleng/tilt/blob/6297a6f3aeb93ee5a8dcaaed5921acfe67455060/vendor/github.com/docker/distribution/reference/reference.go#L203)
2. "rose-quartz" might be a valid registry name (I couldn't find the rules for valid registry names) that we should allow, and the issue is we're assembling combining it with the image name in an invalid way
3. regardless of the other stuff, it seems like we have a bug somewhere that we're logging this error without any context (resource or image name), and maybe if we just fix that, users can figure it out?

landism: > and maybe if we just fix that, users can figure it out?

nm. it's easy to change the error message, to, e.g.
`error parsing rose-quartz/vigoda after applying default registry rose-quartz: repository name must be canonical`

but I'd still have no idea what that means

jazzdan:
@landism @nicks WDYT now?

internal/container/default_registry.go
```diff
@@ -30,7 +30,7 @@ func ReplaceRegistry(defaultRegistry Registry, rs RefSelector) (reference.Named,
 	newNs := fmt.Sprintf("%s/%s", defaultRegistry, escapeName(rs.RefFamiliarName()))
 	newN, err := reference.ParseNamed(newNs)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("Error parsing %s after applying default registry %s: %w", newNs, defaultRegistry, err)
```
landism: nit: `errors.Wrapf`

internal/tiltfile/tiltfile_test.go
```diff
@@ -2576,6 +2576,7 @@ default_registry("foo")
 docker_build('gcr.io/foo', 'foo')
 `)
 
+	f.loadErrString("Traceback ")
```
landism: doesn't matter that much, but each call to `loadErrString` will execute the Tiltfile

loadErrString takes a `...string`, so you can just `f.loadErrString("Traceback ", "repository name must be canonical")`

also, maybe change the default registry to "bar" (so that it doesn't match the image name) and throw "bar" in the list of expected error text

PR #2872, opened 2020-01-29
model: ImageTarget has both localRef and clusterRef [ch4292]

maiamcc:
not ever populating these refs with different things yet; that behavior
and tests to follow


pkg/model/image_target.go
```diff
@@ -4,17 +4,14 @@ import (
...
 	"github.com/windmilleng/tilt/internal/container"
 	"github.com/windmilleng/tilt/internal/sliceutils"
 )
 
 type ImageTarget struct {
-	ConfigurationRef container.RefSelector
-	DeploymentRef    reference.Named
-	BuildDetails     BuildDetails
-	MatchInEnvVars   bool
+	Refs           container.RefSet
```
nicks: out of curiosity, why group them together into a sub-struct? it doesn't seem like you ever use them together as a group

nicks: it might make sense to add ConfigurationRef() etc accessors

pkg/model/manifest_test.go
```diff
@@ -141,29 +141,43 @@ var equalitytests = []struct {
...
+		Manifest{}.WithImageTarget(ImageTarget{Refs: container.RefSet{LocalRef: img1.AsNamedOnly()}}),
+		Manifest{}.WithImageTarget(ImageTarget{Refs: container.RefSet{LocalRef: img2.AsNamedOnly()}}),
 		false,
 		true,
 	},
 	{
-		"ImageTarget.DeploymentRef equal",
-		Manifest{}.WithImageTarget(ImageTarget{DeploymentRef: img1.AsNamedOnly()}),
-		Manifest{}.WithImageTarget(ImageTarget{DeploymentRef: img1.AsNamedOnly()}),
+		"ImageTarget.LocalRef equal",
```
nicks: fwiw, it doesn't seem like adding tests here is adding a ton of value. Now that we've added the comparator library, they're basically just testing the comparator library.

pkg/model/image_target.go
maiamcc: in a future PR i pass refsets around to builders -- i don't feel strongly about it, it just seemed a way to reduce the number of args to keep track of 🤷‍♀

pkg/model/manifest_test.go
maiamcc: the what now?

PR #2873, opened 2020-01-29
hud: when printing push output, make sure we always print the last "finished" line

nicks:
Hello @jazzdan, @hyu,

Please review the following commits I made in branch nicks/mustprint:

19cb7fbec692022e29906047ae2eba0b9a2f68a9 (2020-01-29 13:21:55 -0500)
hud: when printing push output, make sure we always print the last "finished" line

### DIFF
--- a/pkg/model/logstore/logstore.go
+++ b/pkg/model/logstore/logstore.go
@@ -107,7 +107,15 @@ type LogLine struct {
 	Text       string
 	SpanID     SpanID
 	ProgressID string
-	Time       time.Time
+
+	// Most progress lines are optional. For example, if a bunch
+	// of little upload updates come in, it's ok to skip some.
+	//
+	// ProgressMustPrint indicates that this line must appear in the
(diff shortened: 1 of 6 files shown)


nicks:
fwiw, i don't have an overarching plan for what this progress-logging API should look like. I'm mostly just adding progress-logging to different places in Tilt, playing with the output, and seeing what features it needs.

pkg/model/logstore/logstore.go
```diff
@@ -104,10 +104,11 @@ func segmentsFromBytes(spanID SpanID, time time.Time, level logger.Level, fields
...
 
 type LogLine struct {
-	Text       string
-	SpanID     SpanID
-	ProgressID string
-	Time       time.Time
+	Text              string
+	SpanID            SpanID
+	ProgressID        string
+	ProgressMustPrint bool
```
maiamcc: maybe worth a comment

nicks: added!

PR #2876, opened 2020-01-29
container: make clusterref optional, return localref if no clusterref provided

maiamcc:
not strictly necessary but i think this fits better with our mental models
of localRef vs. clusterRef -- i.e. clusterRef is only specified in very specific
situations, and otherwise, we assume that localRef is usaable everywhere

### DIFF
--- a/internal/container/default_registry.go
+++ b/internal/container/default_registry.go
@@ -18,33 +18,34 @@ type Registry struct {
 	// The Host of a container registry where we can push images. e.g.:
 	//   - localhost:32000
 	//   - gcr.io/windmill-public-containers
-	PushHost string
+	Host string
 
-	// The prefix we use with image names when attempt to pull them from the registry.
-	// In most cases, this is equivalent to PushHost (the host of the container registry that we push to),
-	// but sometimes users will specify a pullHost separately (e.g. using a local registry with KIND:
-	// YAMLs will specify the image as `registry:5000/my-img`, so the pullHost will be `registry:5000`).
-	pullHost string
+	// The prefix we use with image names when referring to them from inside the cluster.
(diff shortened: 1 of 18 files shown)


internal/container/reference.go
```diff
@@ -19,29 +19,55 @@ type RefSet struct {
...
+	clusterRef reference.Named
 }
 
-// SimpleRefSet makes a ref set for the given selector, assuming that LocalRef
-// and ClusterRef are equal.
+func NewRefSet(confRef RefSelector, localRef, clusterRef reference.Named) RefSet {
+	return RefSet{
+		ConfigurationRef: confRef,
+		LocalRef:         localRef,
+		clusterRef:       clusterRef,
```
nicks: seems fine, tho this makes me wonder if you should store the localRegistry / clusterRegistry instead of the localRef / clusterRef.

that way, if you can't update the LocalRef and forget to update the clusterRef

maiamcc: oooh I like that!

internal/container/default_registry.go
```diff
@@ -8,31 +8,67 @@ import (
...
 func escapeName(s string) string {
 	return string(escapeRegex.ReplaceAll([]byte(s), []byte("_")))
 }
 
-// Produces a new image name that is in the specified registry.
+type Registry struct {
+	// The Host of a container registry where we can push images. e.g.:
+	//   - localhost:32000
+	//   - gcr.io/windmill-public-containers
+	PushHost string
```
nicks: are you sure you don't want to keep this as HostFromLocal vs HostFromCluster?

PushHost vs PullHost isn't quite accurate - for example, when you have image dependencies, you'll both push and pull from the PushHost

PR #2882, opened 2020-01-30
engine: simplify pod registration a bit

nicks:
Hello @landism, @maiamcc,

Please review the following commits I made in branch nicks/pod:

e6535c08ea75f0965d5bcf29c81a852f03eeef36 (2020-01-30 17:13:21 -0500)
engine: simplify pod registration a bit

### DIFF
--- a/internal/engine/pod.go
+++ b/internal/engine/pod.go
@@ -28,16 +28,19 @@ func handlePodChangeAction(ctx context.Context, state *store.EngineState, action
 	ms := mt.State
 	manifest := mt.Manifest
 	podInfo, isNew := maybeTrackPod(ms, action)
-	podID := k8s.PodIDFromPod(pod)
-	if podInfo.PodID != podID {
-		// This is an event from an old pod.
+	if podInfo == nil {
+		// This is an event from an old pod that has never been tracked.
 		return
 	}
 
 	// Update the status
+	podInfo.StartedAt = pod.CreationTimestamp.Time
+	podInfo.Status = k8swatch.PodStatusToString(*pod)
+	podInfo.Namespace = k8s.NamespaceFromPod(pod)
+	podInfo.HasSynclet = sidecar.PodSpecContainsSynclet(pod.Spec)
+	podInfo.SpanID = runtimelog.SpanIDForPod(podInfo.PodID)
 	podInfo.Deleting = pod.DeletionTimestamp != nil && !pod.DeletionTimestamp.IsZero()
 	podInfo.Phase = pod.Status.Phase
-	podInfo.Status = k8swatch.PodStatusToString(*pod)
 	podInfo.StatusMessages = k8swatch.PodStatusErrorMessages(*pod)
 	podInfo.Conditions = pod.Status.Conditions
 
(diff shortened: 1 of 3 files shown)


nicks:
moving towards a system where we have Deployments and ReplicaSets in our data model, and so want to separate out the general-purpose resource tracking code from pod-specific stuff

internal/engine/pod.go
```diff
@@ -114,35 +117,30 @@ func matchPodChangeToManifest(state *store.EngineState, action k8swatch.PodChang
...
+	// Only attach a new pod to the runtime state if it's from the current deploy;
+	// if it's from an old deploy/an old Tilt run, we don't want to be checking it
+	// for status etc.
+	if !isCurrentDeploy {
+		return runtime.Pods[podID], false
+	}
+
 	// Case 1: We haven't seen pods for this ancestor yet.
 	ancestorUID := action.AncestorUID
 	isAncestorMatch := ancestorUID != ""
```
maiamcc: this line has confused me in the past -- I think this variable is more appropriately named `hasAncestor`?

maiamcc: this line has confused me in the past -- I think this variable is more appropriately named `hasAncestor`?

internal/engine/pod.go
```diff
@@ -114,35 +117,30 @@ func matchPodChangeToManifest(state *store.EngineState, action k8swatch.PodChang
...
-			// for status etc.
-			runtime.PodAncestorUID = ancestorUID
-			runtime.Pods = make(map[k8s.PodID]*store.Pod)
-			runtime.Pods[podID] = podInfo
-			ms.RuntimeState = runtime
+			PodID: podID,
 		}
+
+		runtime.PodAncestorUID = ancestorUID
+		runtime.Pods = make(map[k8s.PodID]*store.Pod)
```
maiamcc: any reason not to initialize this map in `GetOrCreateK8sRuntimeState`?

internal/engine/pod.go
```diff
@@ -151,19 +149,11 @@ func maybeTrackPod(ms *store.ManifestState, action k8swatch.PodChangeAction) (*s
...
-			Namespace:  ns,
-			HasSynclet: hasSynclet,
-			SpanID:     runtimelog.SpanIDForPod(podID),
-		}
-		if isCurrentDeploy {
-			// Only attach a new pod to the runtime state if it's from the current deploy;
-			// if it's from an old deploy/an old Tilt run, we don't want to be checking it
-			// for status etc.
-			runtime.Pods[podID] = podInfo
+			PodID: podID,
```
maiamcc: these two cases are increasingly similar, is there any way to consolidate? (I have an instinct that yes but I'm not seeing exactly how, so 🤷‍♀)

internal/engine/pod.go
nicks: nope, that's not what this field means. There are more comments on it where the PodChangeAction struct is defined. I can change the name of the field if that helps.

internal/engine/pod.go
nicks: we're clearing it, not initializing it. will add a comment on that too

internal/engine/pod.go
nicks: done

PR #2883, opened 2020-01-30
container: RefSet calculates local/clusterRef from Registry to keep things from getting out of sync

maiamcc:
local/ClusterRefs shouldn't really be hardcoded, so here's this

### DIFF
--- a/internal/build/custom_builder_test.go
+++ b/internal/build/custom_builder_test.go
@@ -18,6 +18,8 @@ import (
 	"github.com/windmilleng/tilt/pkg/model"
 )
 
+var TwoURLRegistry = container.NewRegistryWithHostFromCluster("localhost:1234", "registry:1234")
+
 func TestCustomBuildSuccess(t *testing.T) {
 	f := newFakeCustomBuildFixture(t)
 	defer f.teardown()
@@ -39,7 +41,7 @@ func TestCustomBuildSuccessClusterRefTaggedWithDigest(t *testing.T) {
 	sha := digest.Digest("sha256:11cd0eb38bc3ceb958ffb2f9bd70be3fb317ce7d255c8a4c3f4af30e298aa1aab")
 	f.dCli.Images["localhost:1234/foo/bar:tilt-build-1551202573"] = types.ImageInspect{ID: string(sha)}
 	cb := model.CustomBuild{WorkDir: f.tdf.Path(), Command: "true"}
-	refs, err := f.cb.Build(f.ctx, refSetFromStrings("localhost:1234/foo/bar", "registry:1234/foo/bar"), cb)
+	refs, err := f.cb.Build(f.ctx, refSetFromStrings(TwoURLRegistry, "localhost:1234/foo/bar", "registry:1234/foo/bar"), cb)
 	require.NoError(t, err)
 
 	assert.Equal(f.t, container.MustParseNamed("localhost:1234/foo/bar:tilt-11cd0eb38bc3ceb9"), refs.LocalRef)
@@ -63,7 +65,7 @@ func TestCustomBuildSuccessClusterRefTaggedIfSkipsLocalDocker(t *testing.T) {
 	defer f.teardown()
 
 	cb := model.CustomBuild{WorkDir: f.tdf.Path(), Command: "true", SkipsLocalDocker: true}
-	refs, err := f.cb.Build(f.ctx, refSetFromStrings("localhost:1234/foo/bar", "registry:1234/foo/bar"), cb)
+	refs, err := f.cb.Build(f.ctx, refSetFromStrings(TwoURLRegistry, "localhost:1234/foo/bar", "registry:1234/foo/bar"), cb)
 	require.NoError(f.t, err)
 
 	assert.Equal(f.t, container.MustParseNamed("localhost:1234/foo/bar:tilt-build-1551202573"), refs.LocalRef)
@@ -155,10 +157,10 @@ func refSetFromString(s string) container.RefSet {
 	return container.SimpleRefSet(sel)
 }
 
-func refSetFromStrings(local, cluster string) container.RefSet {
+func refSetFromStrings(reg container.Registry, local, cluster string) container.RefSet {
 	localNamed := container.MustParseNamed(local)
 	clusterNamed := container.MustParseNamed(cluster)
-	return container.NewRefSet(container.NameSelector(localNamed), localNamed, clusterNamed)
+	return container.NewRefSet(container.NameSelector(localNamed), reg, localNamed, clusterNamed)
 }
 
 func (f *fakeCustomBuildFixture) teardown() {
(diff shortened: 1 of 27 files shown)


internal/tiltfile/tiltfile_state.go
```diff
@@ -1068,8 +1056,20 @@ func (s *tiltfileState) imgTargetsForDependencyIDsHelper(ids []model.TargetID, c
 		}
 		claimStatus[id] = claimPending
 
+		registry := s.defaultRegistryHost
+		if s.orchestrator() == model.OrchestratorK8s && !s.privateRegistry.Empty() {
+			// If we've found a private registry in the cluster at run-time,
+			// use that instead of the one in the tiltfile
+			s.logger.Infof("Auto-detected private registry from environment: %s", s.privateRegistry)
+			registry = s.privateRegistry
```
maiamcc: hmm, I would expect something specified in the Tiltfile to override something found in the environment--do you have thoughts?

nicks: Nope, if you're using a local registry (like microk8s), you always want Tilt to use that, regardless of what's in the Tiltfile.

maiamcc: 👍 cool thanks

internal/build/custom_builder_test.go
```diff
@@ -37,13 +39,13 @@ func TestCustomBuildSuccessClusterRefTaggedWithDigest(t *testing.T) {
 	defer f.teardown()
 
 	sha := digest.Digest("sha256:11cd0eb38bc3ceb958ffb2f9bd70be3fb317ce7d255c8a4c3f4af30e298aa1aab")
-	f.dCli.Images["localhost:1234/foo/bar:tilt-build-1551202573"] = types.ImageInspect{ID: string(sha)}
+	f.dCli.Images["localhost:1234/foo_bar:tilt-build-1551202573"] = types.ImageInspect{ID: string(sha)}
```
nicks: this change isn't right.

I think you broke something here, and it's revealing a deeper problem with this PR?

The ReplaceRegistry function you're using isn't bijective, and it's leading to problems where going from
imageName -> (registry, path) -> imageName
is making a different image name pop out.

The ReplaceRegistry function you're using isn't bijective -> i.e.,

internal/tiltfile/tiltfile_state.go
```diff
@@ -1068,8 +1056,20 @@ func (s *tiltfileState) imgTargetsForDependencyIDsHelper(ids []model.TargetID, c
...
+		}
+		refs, err := container.NewRefSet(image.configurationRef, registry)
```
nicks: would it make more sense to do this substitution during tiltfile execution? So RefSet would go on the dockerImage struct.

The problem with this approach is that if there's an error, you'll lose the stack trace, as in this bug: #2768

so we try hard not to return errors from this function without lots of context about what went wrong and how to fix it.

internal/tiltfile/tiltfile_state.go
```diff
@@ -1068,8 +1056,20 @@ func (s *tiltfileState) imgTargetsForDependencyIDsHelper(ids []model.TargetID, c
 		}
 		claimStatus[id] = claimPending
 
+		registry := s.defaultRegistryHost
+		if s.orchestrator() == model.OrchestratorK8s && !s.privateRegistry.Empty() {
+			// If we've found a private registry in the cluster at run-time,
+			// use that instead of the one in the tiltfile
+			s.logger.Infof("Auto-detected private registry from environment: %s", s.privateRegistry)
```
nicks: this is going to print once for every image, which doesn't seem right.

internal/k8s/registry.go
```diff
@@ -94,7 +94,11 @@ func (r *registryAsync) Registry(ctx context.Context) container.Registry {
 		}
 
 		portSpec := portSpecs[0]
-		r.registry = container.NewRegistry(fmt.Sprintf("localhost:%d", portSpec.NodePort))
+		host := fmt.Sprintf("localhost:%d", portSpec.NodePort)
+		r.registry, err = container.NewRegistry(host)
```
nicks: this sets r.registry if there's an error, which i'm pretty sure will cause problems?

internal/container/reference.go
```diff
@@ -1,73 +1,112 @@
...
-	LocalRef reference.Named
-
-	// The image name (minus the Tilt tag) as referenced from the cluster (in k8s YAML,
-	// etc.) (Often this will be the same as the LocalRef, but in some cases they diverge:
-	// e.g. when using a local registry with KIND, the image localhost:1234/my-image
-	// (LocalRef) is referenced in the YAML as http://registry/my-image (clusterRef).
-	// clusterRef is optional; if not provided, we assume LocalRef == clusterRef.
-	clusterRef reference.Named
+	// (Optional) registry to prepend to ConfigurationRef to yield ref to use in update and deploy
+	Registry Registry
```
nicks: i wonder if these should be private fields so that you can't change them after they've been validated

internal/container/reference_test.go
```diff
@@ -0,0 +1,81 @@
...
+			assert.Equal(t, tc.expectedClusterRef, clusterRef.String())
+		})
+	}
+}
+
+func assertNewRefSetError(t *testing.T, selector RefSelector, reg Registry, expectedErr string) {
+	_, err := NewRefSet(selector, reg)
+	require.Error(t, err)
+	require.Contains(t, err.Error(), expectedErr)
+	fmt.Println(err)
```
nicks: ?

internal/build/custom_builder_test.go
nicks: sorry! you were right, i misread the test

internal/tiltfile/tiltfile_state.go
maiamcc: ah gotcha, will do.

maiamcc: ...hmm wait, i don't know if we can do that -- we don't guarantee that `default_registry` is called at any particular point in the tiltfile, right? so we can't assume that by the time the user calls `docker_build`, we've already set `s.registry`.

internal/k8s/registry.go
maiamcc: good catch

internal/container/reference_test.go
maiamcc: 😅

internal/container/reference.go
maiamcc: will do

internal/tiltfile/tiltfile_state.go
nicks: ah, good point, maybe just add additional context to the error?

maiamcc: will do

maiamcc:
okay @nicks I think addressed all your comments

PR #2885, opened 2020-01-30
tiltfile: can optionally specify host_from_cluster as arg to default_registry in Tiltfile [ch4292]

maiamcc:
Tiltfile change that should make all the previous engine changes go 🤞
engine tests and integration tests to follow


nicks:
i think the preceding this PR that this builds on has problems, let's talk over there

internal/container/reference.go
```diff
@@ -1,73 +1,112 @@
...
-	LocalRef reference.Named
-
-	// The image name (minus the Tilt tag) as referenced from the cluster (in k8s YAML,
-	// etc.) (Often this will be the same as the LocalRef, but in some cases they diverge:
-	// e.g. when using a local registry with KIND, the image localhost:1234/my-image
-	// (LocalRef) is referenced in the YAML as http://registry/my-image (clusterRef).
-	// clusterRef is optional; if not provided, we assume LocalRef == clusterRef.
-	clusterRef reference.Named
+	// (Optional) registry to prepend to ConfigurationRef to yield ref to use in update and deploy
+	Registry Registry
```
nicks: i wonder if these should be private fields so that you can't change them after they've been validated

internal/container/reference_test.go
```diff
@@ -0,0 +1,81 @@
...
+			assert.Equal(t, tc.expectedClusterRef, clusterRef.String())
+		})
+	}
+}
+
+func assertNewRefSetError(t *testing.T, selector RefSelector, reg Registry, expectedErr string) {
+	_, err := NewRefSet(selector, reg)
+	require.Error(t, err)
+	require.Contains(t, err.Error(), expectedErr)
+	fmt.Println(err)
```
nicks: ?

internal/k8s/registry.go
```diff
@@ -94,7 +94,11 @@ func (r *registryAsync) Registry(ctx context.Context) container.Registry {
 		}
 
 		portSpec := portSpecs[0]
-		r.registry = container.NewRegistry(fmt.Sprintf("localhost:%d", portSpec.NodePort))
+		host := fmt.Sprintf("localhost:%d", portSpec.NodePort)
+		r.registry, err = container.NewRegistry(host)
```
nicks: this sets `r.registry` if there's an error, which i'm pretty sure will cause problems?

internal/tiltfile/tiltfile_state.go
```diff
@@ -1068,8 +1056,20 @@ func (s *tiltfileState) imgTargetsForDependencyIDsHelper(ids []model.TargetID, c
 		}
 		claimStatus[id] = claimPending
 
+		registry := s.defaultReg
+		if s.orchestrator() == model.OrchestratorK8s && !s.privateRegistry.Empty() {
+			// If we've found a private registry in the cluster at run-time,
+			// use that instead of the one in the tiltfile
+			s.logger.Infof("Auto-detected private registry from environment: %s", s.privateRegistry)
```
nicks: this is going to print once for every image, which doesn't seem right.

internal/tiltfile/tiltfile_state.go
```diff
@@ -1068,8 +1056,20 @@ func (s *tiltfileState) imgTargetsForDependencyIDsHelper(ids []model.TargetID, c
...
+			registry = s.privateRegistry
+		}
+		refs, err := container.NewRefSet(image.configurationRef, registry)
```
nicks: would it make more sense to do this substitution during tiltfile execution? So RefSet would go on docker_image.

The problem with this approach is that if there's an error, you'll lose the stack trace, as in this bug: https://github.com/windmilleng/tilt/issues/2768

so we try hard not to return errors from this function without lots of context about what went wrong and how to fix it.

nicks:
oops, just realized i left comments on the wrong pr, i'll transfer them over to the other one

internal/container/reference_test.go
maiamcc: 😅

internal/k8s/registry.go
maiamcc: good catch

nicks:
does this need to be rebased?

maiamcc:
@nicks this is ready for review!

PR #2886, opened 2020-01-31
local_resource: expand on sleep comment

landism:


internal/engine/local_target_build_and_deployer.go
```diff
@@ -38,9 +38,20 @@ func (bd *LocalTargetBuildAndDeployer) BuildAndDeploy(ctx context.Context, st st
 	}
 
 	if state := stateSet[targ.ID()]; state.IsEmpty() {
-		// HACK(maia): on initial build, give the file change a little extra time to
-		// propagate through, to increase the chance that we pick it up before we start
-		// the next build (otherwise, we may build that next resource twice).
+		// HACK(maia) If target A generates file X and target B depends on file X, it was common that on Tilt startup,
```
nicks: anecdotally, i've seen this Sleep call be insufficient when inotify is under load. 

One possible way to fix with is to introduce some kind of an fsync() primitive (like we have elsewhere) where we make some change in a temp directory and wait for it to propagate thru all relevant listeners (in this case, the Store state loop)

PR #2888, opened 2020-01-31
engine: handle pod deletion properly

nicks:
Hello @jazzdan, @maiamcc,

Please review the following commits I made in branch nicks/deletepod:

5294b2349e7d2b3f4546cbd79ac85b8558fa5113 (2020-01-31 14:47:02 -0500)
engine: handle pod deletion properly
This is part of a larger change to make our watch code more
generalizable, so that we can watch ReplicaSets and Deployments

### DIFF
--- a/internal/k8s/watch.go
+++ b/internal/k8s/watch.go
@@ -20,6 +20,50 @@ import (
 	"github.com/windmilleng/tilt/pkg/logger"
 )
 
+var PodGVR = v1.SchemeGroupVersion.WithResource("pods")
+var ServiceGVR = v1.SchemeGroupVersion.WithResource("services")
+var EventGVR = v1.SchemeGroupVersion.WithResource("events")
+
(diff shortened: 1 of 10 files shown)


internal/k8s/watch.go
```diff
@@ -20,6 +20,42 @@ import (
...
+
+// A wrapper object around SharedInformer objects, to make them
+// a bit easier to use correctly.
+type ObjectUpdate struct {
+	obj      interface{}
+	isDelete bool
+}
+
+func (r ObjectUpdate) AsPod() (*v1.Pod, bool) {
+	if r.isDelete {
```
maiamcc: I want a comment (here? on `AsDeletedKey`? At the top of this file?) that deletion events are processed agnostic to the type of resource they're deleting.

nicks: added some comments on AsDeletedKey. Basically, the poor informer is trying to do the best it can with limited information

PR #2891, opened 2020-01-31
web: Update vs Live Update is more obvious in Logs

hyu:
This introduces a new visual treatment so it's more obvious when an update is triggered. When Live Update fallbacks occur, the message is now more prominent:

2020-02-04 - LogPane update vs live_update
--------------------------------------------------

This PR had an early version where I cleaned up the command line log output (best seen with `--hud=false`). I made it less noisy by removing border characters and constraining ourselves to indentation and limited color to help users parse the multiplexed logs.

However, this broke too many tests and made this PR excessively massive, so I will do that work in a separate PR. 

Screen Shot 2020-01-31 at 5 07 15 PM
Screen Shot 2020-01-31 at 5 11 23 PM


internal/engine/build_and_deployer.go
```diff
@@ -74,8 +74,9 @@ func (composite *CompositeBuildAndDeployer) BuildAndDeploy(ctx context.Context,
 		}
 
 		if redirectErr, ok := err.(buildcontrol.RedirectToNextBuilder); ok {
-			s := fmt.Sprintf("falling back to next update method because: %v\n", err)
-			logger.Get(ctx).Write(redirectErr.Level, []byte(s))
+			l := logger.Get(ctx).WithFields(logger.Fields{logger.FieldNameBuildEvent: "1"})
+			s := fmt.Sprintf("Live Update temporarily disabled…\nREASON: %v\n", err)
```
nicks: this isn't quite accurate. This codepath handles all redirects, not just live update.

would it make sense to create a new error type (LiveUpdateFailed?), or to add this error message there?

pkg/logger/logger.go
```diff
@@ -77,9 +77,9 @@ func (l Level) AsSevereAs(log Level) bool {
 
 var (
 	NoneLvl    = Level{id: 0, severity: 0}
-	DebugLvl   = Level{id: 3, severity: 100}
-	VerboseLvl = Level{id: 2, severity: 200}
 	InfoLvl    = Level{id: 1, severity: 300}
+	VerboseLvl = Level{id: 2, severity: 200}
```
nicks: haha, i think there's an edit war where @maiamcc wanted these to go in severity order

internal/engine/build_and_deployer.go
```diff
@@ -74,8 +74,9 @@ func (composite *CompositeBuildAndDeployer) BuildAndDeploy(ctx context.Context,
 		}
 
 		if redirectErr, ok := err.(buildcontrol.RedirectToNextBuilder); ok {
-			s := fmt.Sprintf("falling back to next update method because: %v\n", err)
-			logger.Get(ctx).Write(redirectErr.Level, []byte(s))
+			l := logger.Get(ctx).WithFields(logger.Fields{logger.FieldNameBuildEvent: "1"})
```
nicks: can we make this more descriptive than "0" / "1"? maybe "init" and "fallback"?

pkg/model/logstore/logstore.go
```diff
@@ -585,6 +585,22 @@ func (s *LogStore) logHelper(spansToLog map[SpanID]*Span, showManifestPrefix boo
 			}
 		}
 
+		if segment.Fields[logger.FieldNameBuildEvent] == "0" {
```
nicks: i think you're going to need to sync and rewrite this bit, because this rendering pipeline has changed a lot

web/src/LogPaneLine.tsx
```diff
@@ -0,0 +1,159 @@
...
+  overflow: hidden;
+  white-space: nowrap;
+
+  ${LogPaneLineRoot}.is-contextchange > & {
+    margin-top: -${Height.logLineSeparator}px;
+    border-top: ${Height.logLineSeparator}px solid ${Color.gray};
+  }
+`
+
+let LineContent = styled(AnsiLine)`
```
nicks: for what it's worth, in my performance investigation, I was finding that styled-components adds non-trivial overhead because it's basically wrapping every react component in an additional react component. 

It's not a big deal in most cases. But for LogLines (where there might be thousands or tens of thousands of them), I was seeing significant performance slow-down each time we added a new styled(). That's why I kept doing things with spans and CSS classes here.

nicks:
ya, let's meet up monday morning to figure out the best way to land this - i think it's going to be a long slog to merge it as-is (given merge conflicts and test failures) but maybe we can break it down a bit

web/src/LogPaneLine.tsx
hyu: This is a very good point! I had wondered about this. I'll go ahead and use span and CSS here.

pkg/logger/logger.go
hyu: 😬 — I actually vaguely knew about that. I was playing with this and hadn't actually meant to commit it. I withdraw my end of this tug of war.

internal/engine/build_and_deployer.go
hyu: Oh geez, for some reason I got stuck in an earlier era of programming history and thought enums could only be integers. But clearly that's not even an integer, it's a string... 🙄

internal/engine/build_and_deployer.go
hyu: Thanks for catching this! For now, I reverted the error message to the more general "Falling back" notice. I agree that we should have a distinct error type for LiveUpdate. But in the interest of landing this PR more quickly, I will do that in a separate PR.

nicks: hmmm...i don't see an update here? did you forget to push?

pkg/model/logstore/logstore.go
```diff
@@ -679,10 +684,27 @@ func (s *LogStore) toLogLines(options logOptions) []LogLine {
...
+
+			// If the text exceeds lineLength, still add a short border
+			const minBorderLength = 5
+			textPadding := 2
+			borderDashCount := lineLength - len(segment.Text) - textPadding
+			if borderDashCount < minBorderLength {
+				borderDashCount = minBorderLength
+			}
+
+			border := "\n" + strings.Repeat("═", borderDashCount) + "╡ "
```
nicks: This adds a newline character to the middle of a LogLine, which is going to have weird effects.

This probably needs to be

```
sb.WriteString("\n")
maybePushLine()
sb.WriteString(border)

File: web/src/types.ts [hunk 7]
```diff
@@ -67,4 +77,5 @@ export type LogLine = {
   text: string
   manifestName: string
   level: string
+  fields?: Fields | null
```
nicks: will this break React caching? see the comment on line 73 (a couple lines above this)

it might make sense to just put progressID and buildEvent directly on the LogLine

web/src/HUD.scss
```diff
@@ -5,6 +5,10 @@
   align-items: stretch;
   width: 100%;
   min-height: 100vh;
+
+  ::selection {
```
nicks: this doesn't seem right. We don't want to highlight the selection in other panes

pkg/model/logstore/prefix.go
```diff
@@ -11,12 +11,12 @@ func SourcePrefix(n model.ManifestName) string {
 	if n == "" || n == model.TiltfileManifestName {
 		return ""
 	}
-	max := 12
+	max := 13
```
nicks: how hard would it be to move this change to a separate PR? it affects a lot of tests, and doesn't seem related to the rest of the change.

hyu: Sure, I can make this a separate PR!

web/src/HUD.scss
hyu: Oh yeah, I should have explained my thinking here. This adjusts the default word selection color so it's our brand color. I actually thought it was more inconsistent to have the Log Pane use this color, while the rest of the interface has a slightly different (default) blue highlight color. The fact that the two blues were very close, but a little bit off, seemed not very intentional.

That said, this doesn't need to be part of this PR, so I'll revert it for now.

pkg/model/logstore/prefix.go
hyu: Nevermind — discussed this elsewhere with Nick and decided it was too much trouble.

(Many of our tests for logs just test expected hard-coded strings against actual output. We should put these into our snapshot test system so Golden Master for everything can be generated with one command.)

web/src/types.ts
hyu: Oops, I read this comment and subsequently forgot about it. I'll put buildEvent directly on the LogLine.

pkg/model/logstore/logstore.go
hyu: Discussed this IRL with @nicks and decided to move all these changes into a separate PR, since we need to be a little more thoughtful about how we manipulate strings in logstore.

PR #2892, opened 2020-02-03
cli: add separate 'tilt dump engine' and 'tilt dump logstore'

nicks:


nicks:
found this useful when debugging some things. it splits out the logstore from the rest of the engine state, and prints byte[]s as strings instead of as base64 encodings

landism:
woooo, I've been wanting this for a while

Issue #2870, opened 2020-01-29
Live_update: sync destination is not absolute


queil:
Hi, using Tilt on Windows:
```
Tilt: v0.11.3, built 2020-01-24
System: windows-amd64
```
I am getting the following message:
```
 Error: live_update: sync destination '/app' (C:\xxx\yyy\Tiltfile:7:35) is not absolute
```
Tiltfile:

```python
analytics_settings(False)
allow_k8s_contexts('docker-desktop')
default_registry('***redacted***')
k8s_yaml(kustomize('./k8s'))

docker_build('***redacted***', './src',
                live_update=[ sync('./src/bin/Debug/netcoreapp3.1', '/app') ]
                )
```

I am using a Linux container and the path '/app' is indeed absolute:

```Dockerfile
FROM mcr.microsoft.com/dotnet/core/sdk:3.1 AS build-env

# Install the dependencies for Visual Studio Remote Debugger
RUN apt-get update && apt-get install -y --no-install-recommends unzip procps

# Install Visual Studio Remote Debugger
RUN curl -sSL https://aka.ms/getvsdbgsh | bash /dev/stdin -v latest -l ~/vsdbg  

WORKDIR /app


COPY ./*.csproj ./
RUN dotnet restore --source https://api.nuget.org/v3/index.json

COPY ./ ./
RUN dotnet publish -c Release -o ./out --no-restore

FROM ***redacted***/microsoft/dotnet/aspnet:3.1
WORKDIR /app
COPY --from=build-env /app/out ./

RUN groupadd appuser
RUN useradd -r -u 10001 -g appuser appuser
RUN chown appuser:appuser /app
RUN chmod 755 /app

ENV ASPNETCORE_URLS=http://+:5000

EXPOSE 5000
USER appuser
ENTRYPOINT ["dotnet", "***redacted***.dll"]
```

maiamcc:
Oof, @queil, I'm sorry you're running into this problem. Tossed up a PR that I hope will address it, would love for you to give it a try once I merge it into master.

maiamcc:
@queil I merged a PR that shouuuld fix this, you can either build off master and test it now, or wait for our next release.

queil:
@maiamcc thanks, I'll probably wait for the release. In the meantime I've switched to WSL 2 and the linux tilt.

maiamcc:
[...]

PR #2893, opened 2020-02-03
tiltfile: always use unix abs-path check when verifying live update remote path

maiamcc:
hopefully addresses #2870

in so doing, might break windows containers -- which we never purported
to support anyway, so I don't think this is a regression? maybe?

### DIFF
--- a/internal/tiltfile/live_update.go
+++ b/internal/tiltfile/live_update.go
@@ -2,7 +2,6 @@ package tiltfile
 
 import (
 	"fmt"
-	"path/filepath"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


internal/tiltfile/live_update.go
```diff
@@ -203,7 +202,7 @@ func (s *tiltfileState) liveUpdateStepToModel(t *starlark.Thread, l liveUpdateSt
 	case liveUpdateFallBackOnStep:
 		return model.LiveUpdateFallBackOnStep{Files: x.files}, nil
 	case liveUpdateSyncStep:
-		if !filepath.IsAbs(x.remotePath) {
+		if !unixIsAbs(x.remotePath) {
```
jazzdan: I think we want to use [`path.isAbs`](https://golang.org/src/path/path.go?s=5145:5173#L191) (note `path` vs `filepath`):
```suggestion
		if !path.IsAbs(x.remotePath) {
```

This should ensure that a Unix definition of absolute path is used.

nicks: +1

maiamcc:
whoops I accidentally deleted @jazzdan's change request 😮 but imma go do what he says cuz it's smart

PR #2896, opened 2020-02-04
ci: use frozen yarn lockfile

jazzdan:
This will cause CI to fail if `yarn install`
causes an update to `yarn.lock`.

This should prevent drifting lockfiles.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -94,28 +94,18 @@ dev-js:
 	cd web && yarn install && yarn run start
 
 check-js:
-ifneq ($(CIRCLECI),true)
 	cd web && yarn install --frozen-lockfile
-else
-	cd web && yarn install
-endif
-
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -94,18 +94,28 @@ dev-js:
 	cd web && yarn install && yarn run start
 
 check-js:
+ifneq ($(CIRCLECI),true)
```
nicks: maybe all our yarn commands should use a frozen lockfile unless you specifically say otherwise? wdyt?

jazzdan: I'm happy to do that, I just wasn't sure how many folks were using `make` to do JavaScript tasks in their local development. If they are, such a change would break them.

But we can cross that bridge when we get to it. I'll make that change now.

PR #2899, opened 2020-02-04
k8s: auto-detect local registry from node annotations

nicks:
Hello @maiamcc,

Please review the following commits I made in branch nicks/autodetect2:

cc5b8d24210338b7b80fbd99a63affffb8eb1612 (2020-02-04 15:56:06 -0500)
k8s: auto-detect local registry from node labels

d785b12a7bf1b6e50f8157c3397d118bf4b92e9b (2020-02-04 14:44:49 -0500)
k8s: rename private registry to local registry, to reduce confusion

### DIFF
--- a/internal/k8s/registry.go
+++ b/internal/k8s/registry.go
@@ -14,6 +14,9 @@ import (
 	"github.com/windmilleng/tilt/pkg/logger"
 )
 
+const annotationRegistry = "tilt.dev/registry"
+const annotationRegistryFromCluster = "tilt.dev/registry-from-cluster"
+
 const microk8sRegistryNamespace = "container-registry"
 const microk8sRegistryName = "registry"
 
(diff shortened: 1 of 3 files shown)


internal/k8s/registry.go
```diff
@@ -49,66 +54,119 @@ func newRegistryAsync(env Env, core apiv1.CoreV1Interface, runtimeSource Runtime
...
+		fromLocal = fmt.Sprintf("%s:%s", registryHostnameFromLocal, registryPortFromLocal)
+	}
+
+	registryHostnameFromCluster := labels[labelRegistryHostnameFromCluster]
+	registryPortFromCluster := labels[labelRegistryPortFromCluster]
+	if registryHostnameFromCluster != "" && registryPortFromCluster != "" {
+		fromCluster = fmt.Sprintf("%s:%s", registryHostnameFromCluster, registryPortFromCluster)
+	}
+
+	if fromLocal != "" && fromCluster != "" {
```
maiamcc: You can pass `""` as the 2nd arg to `NewRegistryWithHostFromCluster`, that might be easier to follow than these conditionals

i.e.
```
var fromCluster
if registryHostnameFromCluster != "" && registryPortFromCluster != "" {
    fromCluster = fmt.Sprintf("%s:%s", registryHostnameFromCluster, registryPortFromCluster)
}
eg, err := container.NewRegistryWithHostFromCluster(fromLocal, fromCluster)
...
```

internal/k8s/registry.go
```diff
@@ -49,66 +54,119 @@ func newRegistryAsync(env Env, core apiv1.CoreV1Interface, runtimeSource Runtime
...
+	reg, err := container.NewRegistry(host)
+	if err != nil {
+		logger.Get(ctx).Warnf("Error validating private registry host %q: %v", host, err)
+		return container.Registry{}
+	}
+
+	return reg
+}
+
+func (r *registryAsync) inferRegistryFromTiltNodeLabels(ctx context.Context) container.Registry {
```
maiamcc: comment on what this function is supposed to do (i.e. infers the existence of a local registry if it created via a tilt script and thus has these labels on it)

internal/k8s/registry.go
nicks: done

internal/k8s/registry.go
nicks: done

nicks:
also changed this to use annotations instead of labels, which is more inline with how the Kubernetes documentation recommends we use annotations
https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/#attaching-metadata-to-objects

PR #2901, opened 2020-02-05
analytics: update wmclient and use WithoutGlobalTags() instead of IncrAnonymous()

nicks:
Hello @jazzdan,

Please review the following commits I made in branch nicks/wmclinet:

3cc80c034036ce0ad6fe41c845c8c0eb4b6106e5 (2020-02-05 13:32:48 -0500)
analytics: update wmclient and use WithoutGlobalTags() instead of IncrAnonymous()


nicks:
(The analytics anonymous-override situation is already very confusing in the code, and i have to read the code multiple times just to understand it, so wanted to straighten out the names a bit before we go changing it some more)

PR #2902, opened 2020-02-05
analytics: send anonymized analytics by default. update the README and nudge to reflect this

nicks:
Hello @dbentley, @jazzdan, @maiamcc,

Please review the following commits I made in branch nicks/ch4422/analytics:

37c2f58a6f1013f1f59aec7df6a4e376b087d039 (2020-02-05 13:55:08 -0500)
analytics: send anonymized analytics by default. update the README and nudge to reflect this
We're having trouble understanding the early part of the Tilt experience.
Are people making it to the web UI?
Are they able to run a project successfully?
Are they able to set up a local kubernetes cluster?

We're going to try enabling it by default on tilt v0.12 and see
what happens. Previous PRs have cleaned up some of the existing
analytics behavior so that it doesn't pick up CI and respects DO_NOT_TRACK=1,
so we feel more comfortable with this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,7 +65,9 @@ status of running services, errors, logs, and more.
 **Follow along:** [@tilt_dev](https://twitter.com/tilt_dev) on Twitter. Updates
 and announcements on the [Tilt blog](https://blog.tilt.dev).
 
-**Help us make Tilt even better:** Run `tilt analytics opt in` to enable telemetry, so we can improve Tilt on every platform. Details in ["What does Tilt send?"](http://docs.tilt.dev/telemetry_faq.html).
+**Help us make Tilt even better:** Tilt sends anonymized usage data, so we can
+improve Tilt on every platform. Details in ["What does Tilt
+send?"](http://docs.tilt.dev/telemetry_faq.html).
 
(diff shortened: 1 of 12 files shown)


README.md
```diff
@@ -65,7 +65,9 @@ status of running services, errors, logs, and more.
 **Follow along:** [@tilt_dev](https://twitter.com/tilt_dev) on Twitter. Updates
 and announcements on the [Tilt blog](https://blog.tilt.dev).
 
-**Help us make Tilt even better:** Run `tilt analytics opt in` to enable telemetry, so we can improve Tilt on every platform. Details in ["What does Tilt send?"](http://docs.tilt.dev/telemetry_faq.html).
+**Help us make Tilt even better:** Tilt sends anonymized usage data,
```
jazzdan: This bolded section feels a little weird now that it's opt out. Maybe just renaming it to "opt out of analytics"?

nicks: but...we don't want people to opt out of analytics? so it feels like a weird call to action

jazzdan: Yes, but the result of _this_ call to action is that they don't help us make Tilt better because they'd be opting out of analytics.

It's not a blocking comment I just think this reads strangely. Not sure what the answer is though.

nicks: ah, i see. i'll just remove this sentence.

PR #2903, opened 2020-02-05
starkit: allow Load behavior to be modified at runtime

jazzdan:


internal/tiltfile/starkit/fixture.go
```diff
@@ -12,25 +12,37 @@ import (
...
-	fs         map[string]string
-	out        *bytes.Buffer
-	useRealFS  bool // Use a real filesystem
+	tb               testing.TB
+	extensions       []Extension
+	path             string
+	temp             *tempdir.TempDirFixture
+	fs               map[string]string
+	out              *bytes.Buffer
+	useRealFS        bool // Use a real filesystem
```
nicks: this is no longer used?

internal/tiltfile/starkit/fixture.go
```diff
@@ -12,25 +12,37 @@ import (
...
+	loadInterceptors []LoadInterceptor
+	oldDir           string
 }
 
 func NewFixture(tb testing.TB, extensions ...Extension) *Fixture {
+	temp := tempdir.NewTempDirFixture(tb)
+	oldDir, err := os.Getwd()
```
nicks: can we make changing the working directory an option on TempDirFixture?

internal/tiltfile/starkit/environment.go
```diff
@@ -146,31 +152,60 @@ func (e *Environment) start(path string) (Model, error) {
...
+		if newPath != "" {
+			// we found an interceptor that does something with this path, return early
+			return newPath, nil
+		}
+	}
+
+	return AbsPath(t, path), nil
+}
+
+func (e *Environment) doLoad(t *starlark.Thread, path string, localPath string) (starlark.StringDict, error) {
```
nicks: can you elaborate on why this needs two paths? afaict it only uses one

internal/tiltfile/starkit/environment.go
```diff
@@ -146,31 +152,60 @@ func (e *Environment) start(path string) (Model, error) {
...
-	// If the path isn't absolute, the loadCache won't work correctly.
-	if !filepath.IsAbs(path) {
-		return starlark.StringDict{}, fmt.Errorf("internal error: path must be absolute")
+	localPath, err := e.getPath(t, path)
+	if err != nil {
+		e.loadCache[localPath] = loadCacheEntry{
+			status:  loadStatusDone,
+			exports: starlark.StringDict{},
+			err:     err,
+		}
```
nicks: i think you need to return here? can you add a test for the case where the interceptor returns an error?

internal/tiltfile/starkit/fixture.go
jazzdan: It is still used.

Screen Shot 2020-02-05 at 5 00 03 PM

internal/tiltfile/starkit/fixture.go
jazzdan: Yeah! And I would also like to socialize that out to the other tests where we do similar things.

internal/tiltfile/starkit/fixture.go
```diff
@@ -12,23 +12,32 @@ import (
...
+	out              *bytes.Buffer
+	useRealFS        bool // Use a real filesystem
+	loadInterceptors []LoadInterceptor
 }
 
 func NewFixture(tb testing.TB, extensions ...Extension) *Fixture {
+	temp := tempdir.NewTempDirFixture(tb)
+	temp.Chdir()
+
+	_ = os.Chdir(temp.Path())
```
nicks: this is obsolete now? it's covered by the temp.Chdir

jazzdan: Yup thanks

Issue #2894, opened 2020-02-04
Force rebuilding in the cli


rmb938:
In the web ui there is a nice button to force a rebuild, however I can't find a way to do that in the cli.

Can this feature be added to the cli?

maiamcc:
Hi @rmb938! Unfortunately, we're not adding new features to the CLI at this time, so this probably won't happen any time soon.

I'm curious why you'd prefer the CI to the web UI for running a force rebuild, though--can you elaborate a bit on how you use or don't use the two interfaces?

maiamcc:
Hi @rmb938! Potentially--can you tell me more about your use case, and why you want to force a rebuild from the CLI instead of from the web UI?

rmb938:
My workflows tend to be all done via a terminal and an IDE for development. Typically my web browser is never used other than looking at documentation and whatnot. I tend to mostly work on backend applications that have no http endspoints or just a basic api so my browser is never used for interacting with those applications.

Before starting to use tilt I typically ran kubectl and docker commands directly in my terminal. When using tilt I also use the terminal interface via `tilt up --no-browser`. Navigating the CLI interface is much easier for my work flows and being able to switch between all the different log types and whatnot is great.

All the features I currently use are already present in the CLI except the force rebuild button, so having to open the web ui just for that one feature is really annoying.

maiamcc:
This is [live in v0.12.0](https://github.com/windmilleng/tilt/releases/tag/v0.12.0)! Syntax is `tilt trigger [resource_name]`:

![2020-02-07 13 52 58](https://user-images.githubusercontent.com/6332648/74057116-3bd3d980-49b1-11ea-9021-b20f2e6ee40f.gif)

rmb938:
Thanks! I was hoping for more of a keybind, like the `X` key to expend the logs. But this works too.

maiamcc:
Ah, I see. If you feel like submitting a PR, it should be pretty easy: [here's the code to trigger an update](https://github.com/windmilleng/tilt/blob/455f644554aabe6c5ec84b9fe2a2f8c13269ba46/internal/cli/trigger.go#L33), and [here's the code that handles terminal key events](https://github.com/windmilleng/tilt/blob/1521aeac180e5eea664a09773220c6963be5c06d/internal/hud/hud.go#L154). If you end up writing this yourself, let me know if you need any help.

rmb938:
Awesome, I'll take a look. Thanks again!

nicks:
[...]

PR #2904, opened 2020-02-05
cli: can trigger update from cli

maiamcc:
as requested in #2894

### DIFF
--- a/internal/cli/action.go
+++ b/internal/cli/action.go
@@ -1,0 +1,59 @@
+package cli
+
+import (
+	"bytes"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+
+	"github.com/spf13/cobra"
+)
+
+func newActionCmd() *cobra.Command {
+	result := &cobra.Command{
+		Use:   "action",
+		Short: "perform the sepcified action",
+		Long: `Perform the specified action via a call to Tilt's API server.
+`,
+	}
+
+	result.AddCommand(newTriggerCmd())
+
+	return result
+}
+
+func newTriggerCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "trigger [RESOURCE_NAME]",
+		Short: "trigger an update for the specified resource",
+		Long: `Trigger an update for the specified resource.
+
+If no file changes for the resource are pending, this will force a full rebuild.
+
+If it is a manual resource with pending file changes, this will cause a build of those pending changes.
+`,
+		Args: cobra.ExactArgs(1),
+		Run:  triggerUpdate,
+	}
+	cmd.Flags().IntVar(&webPort, "port", DefaultWebPort, "Port for the Tilt HTTP server")
+	return cmd
+}
+
+func triggerUpdate(cmd *cobra.Command, args []string) {
+	resource := args[0]
+	url := fmt.Sprintf("http://localhost:%d/api/trigger", webPort)
+	payload := []byte(fmt.Sprintf(`{"manifest_names":[%q]}`, resource))
+	res, err := http.Post(url, "application/json", bytes.NewBuffer(payload))
+	if err != nil {
+		cmdFail(fmt.Errorf("Could not connect to Tilt at %s: %v", url, err))
+	}
+
+	if res.StatusCode != http.StatusOK {
+		body := "<no response body"
+		b, _ := ioutil.ReadAll(res.Body)
+		if string(b) != "" {
+			body = string(b)
+		}
+		cmdFail(fmt.Errorf("Request to %s failed (status code %d): %s", url, res.StatusCode, body))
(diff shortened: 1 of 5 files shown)


internal/cli/action.go
```diff
@@ -0,0 +1,59 @@
...
+	}
+	cmd.Flags().IntVar(&webPort, "port", DefaultWebPort, "Port for the Tilt HTTP server")
+	return cmd
+}
+
+func triggerUpdate(cmd *cobra.Command, args []string) {
+	resource := args[0]
+	url := fmt.Sprintf("http://localhost:%d/api/trigger", webPort)
+	payload := []byte(fmt.Sprintf(`{"manifest_names":[%q]}`, resource))
+	res, err := http.Post(url, "application/json", bytes.NewBuffer(payload))
```
landism: feel free to ignore this for this PR, but we should probably have a function for making an API request from a tilt command so that we're not re-figuring out/implementing error handling and messaging each time and it's consistent across commands. (at least `dump` and `args` also do this)

internal/cli/action.go
```diff
@@ -0,0 +1,59 @@
...
+	if err != nil {
+		cmdFail(fmt.Errorf("Could not connect to Tilt at %s: %v", url, err))
+	}
+
+	if res.StatusCode != http.StatusOK {
+		body := "<no response body"
```
landism: 👀

internal/cli/action.go
```diff
@@ -0,0 +1,59 @@
...
+		b, _ := ioutil.ReadAll(res.Body)
```
landism: I'm a bit nervous about discarding this error. Are you sure it's safe?

internal/cli/action.go
```diff
@@ -0,0 +1,59 @@
...
+		if string(b) != "" {
+			body = string(b)
+		}
+		cmdFail(fmt.Errorf("Request to %s failed (status code %d): %s", url, res.StatusCode, body))
```
landism: I kinda feel like we should always use `res.Status` over `res.StatusCode` when reporting errors to the user. It's easy to forget since we're always checking res.StatusCode immediately beforehand, or maybe there are arguments I'm missing.

internal/cli/action.go
```diff
@@ -0,0 +1,59 @@
...
+	"io/ioutil"
+	"net/http"
+
+	"github.com/spf13/cobra"
+)
+
+func newActionCmd() *cobra.Command {
+	result := &cobra.Command{
+		Use:   "action",
+		Short: "perform the sepcified action",
```
landism:
```suggestion
		Short: "perform the specified action",
```

internal/cli/action.go
```diff
@@ -0,0 +1,59 @@
...
+import (
+	"bytes"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+
+	"github.com/spf13/cobra"
+)
+
+func newActionCmd() *cobra.Command {
```
landism: This feels kinda weird vs just `tilt trigger`. Does this mean we want to move `tilt dump` and `tilt args` under `tilt action`? I think from the user's perspective, whether a command is an API call is an implementation detail they shouldn't really be exposed to, and "here's where the API calls go!" isn't necessarily a mental model we want to foist on them.

Are there advantages I'm missing?

maiamcc: ...nah, you're totes right, will do

internal/cli/action.go
maiamcc: yeahhhhh not a bad idea.

also whoops I forgot that I was gonna actually use the `payload` struct...

internal/cli/action.go
maiamcc: nnnno. no I'm not.

internal/cli/action.go
maiamcc: ooh, I didn't remember that was a thing, gy

maiamcc:
okay, ready for review! cc @landism @nicks

internal/cli/trigger.go
```diff
@@ -0,0 +1,35 @@
...
+
+func triggerUpdate(cmd *cobra.Command, args []string) {
+	resource := args[0]
+
+	// TODO(maia): this should probably be the triggerPayload struct, but seems
+	//   like a lot of code to move over (to avoid import cycles) for one call.
+	payload := []byte(fmt.Sprintf(`{"manifest_names":[%q]}`, resource))
+
+	body := apiPostJson(webPort, "trigger", payload)
+	_ = body.Close()
```
landism: probably worth giving the user some kind of feedback, like a `fmt.Printf("triggered an update of %s\n", resource)`

maiamcc: good call, will do

PR #2909, opened 2020-02-06
extension: define extension storage API w/ filesystem implementation

jazzdan:
This explicitly only stores the contents of the module's Tiltfile and doesn't deal with any of the other metadata yet, just to unblock work on other parts of functionality.

### DIFF
--- a/internal/tiltfile/extension/store.go
+++ b/internal/tiltfile/extension/store.go
@@ -22,11 +22,11 @@ type Store interface {
 
 // TODO(dmiller): should this include an integrity hash?
 type ModuleContents struct {
-	Name             string
-	TiltfileContents string
-	GitCommitHash    string
-	Source           string
-	TimeFetched      time.Time
+	Name              string
+	TiltfileContents  string
+	GitCommitHash     string
+	ExtensionRegistry string
+	TimeFetched       time.Time
 }
 
 type LocalStore struct {
--- a/internal/tiltfile/extension/store.go
+++ b/internal/tiltfile/extension/store.go
@@ -2,12 +2,12 @@ package extension
 
 import (
 	"context"
-	"fmt"
 	"io/ioutil"
 	"os"
 	"path/filepath"
 	"time"
 
+	"github.com/pkg/errors"
 	"github.com/windmilleng/tilt/internal/tiltfile"
 )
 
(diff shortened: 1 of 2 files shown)


internal/tiltfile/extension/store.go
```diff
@@ -0,0 +1,71 @@
...
+		return "", nil
+	}
+
+	return "", err
+}
+
+func (s *LocalStore) Write(ctx context.Context, contents ModuleContents) (string, error) {
+	moduleDir := filepath.Join(s.baseDir, contents.Name)
+	if err := os.MkdirAll(moduleDir, os.FileMode(0700)); err != nil {
+		return "", fmt.Errorf("couldn't store module %s: %v", contents.Name, err)
```
landism: mind testing manually and/or in a unit test that the error message for this and below include the path that we failed to write?

it'd be super annoying if it were just like "couldn't store module openfaas: permission denied" and there was no pointer to what permissions needed to be fixed

also....errors.Wrapf!

internal/tiltfile/extension/store.go
```diff
@@ -0,0 +1,71 @@
...
+	}
+
+	tiltfilePath := filepath.Join(moduleDir, tiltfile.FileName)
+	// TODO(dmiller): store hash, source, time fetched
+	if err := ioutil.WriteFile(tiltfilePath, []byte(contents.TiltfileContents), os.FileMode(0700)); err != nil {
```
landism: this should probably be 0600

internal/tiltfile/extension/store.go
```diff
@@ -0,0 +1,71 @@
...
+	Stat(ctx context.Context, moduleName string) (string, error)
+	Write(ctx context.Context, contents ModuleContents) (string, error)
+}
+
+// TODO(dmiller): should this include an integrity hash?
+type ModuleContents struct {
+	Name             string
+	TiltfileContents string
+	GitCommitHash    string
+	Source           string
```
landism: ha, when I read "Source", I interpreted it as the source code of the extension, and was trying to figure out how it differed from `TiltfileContents`

maybe we want a `type ExtensionRegistry string`?

internal/tiltfile/extension/store.go
jazzdan: I'll just manually include it.

PR #2911, opened 2020-02-06
tiltfile: move logging about local registry out of for-loop to reduce spam [ch4451]

maiamcc:
Hello @nicks, @hyu,

Please review the following commits I made in branch maiamcc/fix-local-registry-log-spam:

9e7ecd665ae90c5f25162fdd588189680c0f0fd7 (2020-02-06 15:31:51 -0500)
tiltfile: move logging about local registry out of for-loop to reduce spam [ch4451]

### DIFF
--- a/internal/tiltfile/tiltfile_state.go
+++ b/internal/tiltfile/tiltfile_state.go
@@ -884,13 +884,14 @@ func (s *tiltfileState) extractEntities(dest *k8sResource, imageRef container.Re
 	return nil
 }
 
-// registry returns the image registry we should use; if detected, a pre-configured
+// decideRegistry returns the image registry we should use; if detected, a pre-configured
 // local registry; otherwise, the registry specified by the user via default_registry.
 // Otherwise, we'll return the zero value of `s.defaultReg`, which is an empty registry.
-func (s *tiltfileState) registry() container.Registry {
+// It has side-effects (a log line) and so should only be called once.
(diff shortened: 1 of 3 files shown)


internal/tiltfile/tiltfile_state.go
```diff
@@ -911,7 +924,8 @@ func (s *tiltfileState) translateK8s(resources []*k8sResource) ([]model.Manifest
 
 		m = m.WithDeployTarget(k8sTarget)
 
-		iTargets, err := s.imgTargetsForDependencyIDs(r.dependencyIDs)
+		registry := s.registry()
```
nicks: can you move this out of the resources loop?

maiamcc: damnit -_- ha yes will do

internal/tiltfile/tiltfile_state.go
```diff
@@ -884,8 +884,22 @@ func (s *tiltfileState) extractEntities(dest *k8sResource, imageRef container.Re
 	return nil
 }
 
+// registry returns the image registry we should use; if detected, a pre-configured
+// local registry; otherwise, the registry specified by the user via default_registry.
+// Otherwise, we'll return the zero value of `s.defaultReg`, which is an empty registry.
+func (s *tiltfileState) registry() container.Registry {
```
nicks: i feel like this should be called decideRegistry() or smth to make it clear this has side-effects and should only be called once

PR #2914, opened 2020-02-06
tiltfile: allow commands to be specified as arrays

landism:
### Problem

At the moment, any place a command is specified in the Tiltfile, it's specified as a string, which then gets wrapped in an sh -c '', which forces the user to deal with shell evaluation.

This means that any time a user constructs a command from starlark variables, they potentially have to worry about those starlark variables containing characters with special meaning to the shell, and we provide no way to escape them, so there's not really anything to be done.

### Solution

Allow commands to be specified as either shell-interpreted strings or raw argv, [Dockerfile-style](https://docs.docker.com/engine/reference/builder/#run).

This PR pretty much consists of:
1. add a new `value.ValueToCmd` function that takes a `starlark.Value` and converts the `string|[]string` to a `model.Cmd`, as appropriate
2. replace all calls to `model.ToShellCmd` in the tiltfile dir with calls to `value.ValueToCmd`
3. unit tests for above

### DIFF
--- a/internal/tiltfile/local_resource.go
+++ b/internal/tiltfile/local_resource.go
@@ -26,7 +26,8 @@ type localResource struct {
 }
 
 func (s *tiltfileState) localResource(thread *starlark.Thread, fn *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
-	var name, updateCmdStr, serveCmdStr string
+	var name string
+	var updateCmdVal, serveCmdVal starlark.Value
 	var triggerMode triggerMode
 	var deps starlark.Value
 	var resourceDepsVal starlark.Sequence
@@ -35,13 +36,13 @@ func (s *tiltfileState) localResource(thread *starlark.Thread, fn *starlark.Buil
 
 	if err := s.unpackArgs(fn.Name(), args, kwargs,
 		"name", &name,
-		"cmd?", &updateCmdStr,
+		"cmd?", &updateCmdVal,
 		"deps?", &deps,
 		"trigger_mode?", &triggerMode,
 		"resource_deps?", &resourceDepsVal,
 		"ignore?", &ignoresVal,
 		"auto_init?", &autoInit,
-		"serve_cmd?", &serveCmdStr,
+		"serve_cmd?", &serveCmdVal,
 	); err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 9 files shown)


internal/tiltfile/value/value.go
```diff
@@ -112,3 +114,23 @@ func StringSliceToList(slice []string) *starlark.List {
 	}
 	return starlark.NewList(v)
 }
+
+// provides dockerfile-style behavior of:
+// a string gets interpreted as a shell command (like, sh -c 'foo bar $X')
+// an array of strings gets interpreted as a raw argv to exec
+func ValueToCmd(v starlark.Value) (model.Cmd, error) {
+	switch x := v.(type) {
+	case nil:
```
jazzdan: nit: should this be an error?

(Background: I hate null with a passion)

nicks: ya, i think this gets at the problem of doing deserialization outside of unpack, where we can't tell the difference between "the user specified explicit nil" and "the user didn't specify the argument". 

i also wonder if the typed nil problem is going to bite us here

internal/tiltfile/local_resource.go
```diff
@@ -26,7 +26,8 @@ type localResource struct {
 }
 
 func (s *tiltfileState) localResource(thread *starlark.Thread, fn *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
-	var name, updateCmdStr, serveCmdStr string
+	var name string
+	var updateCmdVal, serveCmdVal starlark.Value
```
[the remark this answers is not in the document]
landism: Yeah, I suspect everything would be a lot nicer if we never had to pass a `starlark.Value` to `UnpackArgs`.

internal/tiltfile/value/value.go
landism: I've added a commente explaining why this is here.

FWIW, from the UnpackArgs docs:
```
// Beware: an optional *List, *Dict, Callable, Iterable, or Value variable that is
// not assigned is not a valid Starlark Value, so the caller must
// explicitly handle such cases by interpreting nil as None or some
// computed default.
```

(though ideally that comment would explicitly say that an unspecified optional param is left as `nil`)

I only added this case after unit tests in which optional params were unspecified failed because this value is nil, which gives me some level of confidence that we're ok on typed nils, and if we upgrade starlark and that behavior changes, unit tests will catch it.

Also, probably beside the point of this PR, but IIUC, users cannot specify explicit nil in starlark. As hinted at by the starlark comment above, nil is not a valid starlark value. If they pass, e.g. `serve_cmd=None`, then we'll get a `starlark.None` rather than a nil, which we can distinguish post-deserialization. We've still got this problem if we pass go primitives, e.g. `*int` or `*string` to `UnpackArgs`, but I haven't yet noticed any pain from that.

PR #2915, opened 2020-02-06
cli: make manifest prefixes right-aligned in log output

hyu:
- IMO, makes the multiplexed log view easier to read
- More consistent with the layout on Web UI
- also uses snapshot tests for more of the logstore tests

Screen Shot 2020-02-06 at 4 58 49 PM


internal/build/pipeline_state_test.go
```diff
@@ -11,7 +11,7 @@ import (
 )
 
 // NOTE(dmiller): set at runtime with:
-// go test -ldflags="-X 'github.com/windmilleng/tilt/internal/build.PipelineStateWriteGoldenMaster=1'" github.com/windmilleng/tilt/internal/build -run ^TestPipeline
+// go test -ldflags="-X 'github.com/windmilleng/tilt/internal/build.PipelineStateWriteGoldenMaster=1'" ./internal/build -run TestPipeline
```
maiamcc: was this intentional?

hyu: It was indeed! That bit of regex (^) actually didn’t work for me, and would miss all the tests. It worked after I removed it...having improved upon my own Regex abilities not one bit, though.

landism: FWIW, I'd guess you're using a shell that expands `^` (e.g., zsh), and it was messing up your command, and it could be fixed by wrapping it in single quotes to avoid shell expansion, i.e., `'^TestPipeline'`, but it doesn't really matter

hyu: 😯 — you're exactly correct, @landism! Thank you! TIL.

PR #2923, opened 2020-02-10
integration: run integration tests on kind against local registry

maiamcc:
shaved something like 40s off integration tests--surprised it's not more tbh,
but hey, it's something!


maiamcc:
Screen Shot 2020-02-10 at 11 39 30 AM
Screen Shot 2020-02-10 at 11 21 59 AM

master: 5m23s (tests themselves took 261s)
this branch: 4m30s (tests themselves took 219s)

nicks:
it's worth pointing out that the integration tests running on circleci don't use this script.

was that intentional? or were you going to update circleci in a follow-up PR?

maiamcc:
> it's worth pointing out that the integration tests running on circleci don't use this script.
> 
> was that intentional? or were you going to update circleci in a follow-up PR?

definitely not intentional whoops.

jazzdan:
What's the status of this PR @maiamcc?

maiamcc:
Whoops, it fell off my radar. I didn't realize nick had changed everything to busybox, which makes the potential gains here smaller, so I'll take like <30 mins now to either wrap this up or call it dead.

.circleci/config.yml
```diff
@@ -88,7 +88,7 @@ jobs:
       - run: docker kill portforward || exit 0
       - run: docker rm portforward || exit 0
       - run: /go/portforward.sh start
-      - run: kind create cluster --name=integration
+      - run: KIND_CLUSTER_NAME=integration kind-with-registry.sh && export KUBECONFIG="$(kind get kubeconfig-path --name="integration")"
```
nicks: hmmm...I think what you need to do here is to split kind-with-registry.sh into two scripts: the bit that creates the cluster, and the bit that initializes the cluster.

Right now, you need to do a portforward.sh to the apiserver port (see a couple lines down) to be able to talk to the apiserver. It's failing because the kind-with-registry script tries to talk to the apiserver before that port-forward is set up

maiamcc: urgh you're right. will do

maiamcc:
::exhausted sigh::

PR #2924, opened 2020-02-10
tiltfile: implement extension for Tilt extensions

jazzdan:


internal/tiltfile/extension/extension.go
```diff
@@ -0,0 +1,66 @@
...
+const extensionPrefix = "ext://"
+
+func (e *Extension) LocalPath(t *starlark.Thread, arg string) (string, error) {
+	if !strings.HasPrefix(arg, extensionPrefix) {
+		return "", nil
+	}
+
+	moduleName := strings.TrimPrefix(arg, extensionPrefix)
+	// If the module can't be found we fetch it below
+	localPath, err := e.store.ModulePath(e.ctx, moduleName)
```
nicks: i'm a little bit worried about the contract around `e.ctx` here. It's not clear if this is a global context or a thread-specific context, if it has a logger attached, etc.

would it make sense to attach a context to the starlark thread?

jazzdan: Hm, yeah I think that would make more sense.

PR #2926, opened 2020-02-10
tiltfile: add encode_yaml and encode_json

landism:
the complement of `decode_yaml` and `decode_json`, i.e., from a starlark object to a yaml/json-encoded string (or, really, a `Blob`, to make it easier to pass to `k8s_yaml`)

also: changed `decode_yaml` and `decode_json` to take a `Union[str|Blob]` instead of a `str`, since it's likely most of the inputs to those functions will be `Blob`s already (e.g., from the output of `read_file` or `local`)

I really wanted to name all these `yaml.load` / `yaml.loads` / `yaml.dumps`, but `load` is a keyword in starlark, making that impossible :(

### DIFF
--- a/internal/tiltfile/encoding/json.go
+++ b/internal/tiltfile/encoding/json.go
@@ -1,6 +1,7 @@
 package encoding
 
 import (
+	"bytes"
 	"encoding/json"
 	"fmt"
 	"os"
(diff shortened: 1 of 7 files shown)


internal/tiltfile/encoding/json_test.go
```diff
@@ -118,3 +168,27 @@ test()
 	}
 	require.NoError(t, err)
 }
+
+func TestEncodeJSONNonStringMapKey(t *testing.T) {
```
jazzdan: Is there a typo here and in the test below?

```suggestion
func TestEncodeJSONOnStringMapKey(t *testing.T) {
```

internal/tiltfile/encoding/json.go
```diff
@@ -40,12 +41,22 @@ func readJSON(thread *starlark.Thread, fn *starlark.Builtin, args starlark.Tuple
...
 // reads json from a string
 func decodeJSON(thread *starlark.Thread, fn *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
-	var contents starlark.String
+	var contents starlark.Value
 	if err := starkit.UnpackArgs(thread, fn.Name(), args, kwargs, "json", &contents); err != nil {
 		return nil, err
 	}
 
-	return jsonStringToStarlark(contents.GoString(), "")
+	var s string
```
nicks: this should be equivalent to value.AsString(s)?

internal/tiltfile/encoding/json_test.go
landism: nope! JSON only supports strings as map keys, and this test is for serializing a map that has a map key that's not a string and making sure returns an error that map keys must be strings.

internal/tiltfile/encoding/json.go
landism: oh cool, I looked for something to handle the blob/string union and was surprised I couldn't find it

PR #2927, opened 2020-02-10
tiltile: ability to load extensions

jazzdan:


internal/tiltfile/extension/extension.go
```diff
@@ -46,6 +47,12 @@ func (e *Extension) LocalPath(t *starlark.Thread, arg string) (string, error) {
 		return "", nil
 	}
 
+	loadIsHappeningInTopLevel := t.CallStackDepth() == 1
```
jazzdan: Is this a uhh ... valid check?

internal/tiltfile/extension/store.go
```diff
@@ -41,7 +39,7 @@ func NewLocalStore(baseDir string) *LocalStore {
 }
 
 func (s *LocalStore) ModulePath(ctx context.Context, moduleName string) (string, error) {
-	tiltfilePath := filepath.Join(s.baseDir, moduleName, tiltfile.FileName)
+	tiltfilePath := filepath.Join(s.baseDir, moduleName, "Tiltfile")
```
jazzdan: Had to use the bare string here to prevent a import cycles.

internal/tiltfile/extension/github_fetcher.go
```diff
@@ -0,0 +1,49 @@
...
+type GithubFetcher struct {
+}
+
+func NewGithubFetcher() *GithubFetcher {
+	return &GithubFetcher{}
+}
+
+const githubTemplate = "https://raw.githubusercontent.com/windmilleng/tilt-extensions/master/%s/Tiltfile"
+
+func (f *GithubFetcher) Fetch(ctx context.Context, moduleName string) (ModuleContents, error) {
```
jazzdan: I decided not to write tests for this because I didn't feel like it would be useful. Happy to take pushback, though.

internal/tiltfile/extension/extension.go
```diff
@@ -46,6 +47,12 @@ func (e *Extension) LocalPath(t *starlark.Thread, arg string) (string, error) {
...
+
+	if !loadIsHappeningInTopLevel {
+		return "", fmt.Errorf("Illegal extension load: extensions must be loaded from the root Tiltfile. This extension was loaded from %s", t.CallFrame(t.CallStackDepth()-1).Pos.Filename())
```
nicks: confused about why your error message needs to manually include the filename. shouldn't skylark already print the full stack trace here? do you understand why it doesn't?

internal/tiltfile/extension/github_fetcher.go
```diff
@@ -0,0 +1,49 @@
...
+	c := &http.Client{
+		Timeout: 20 * time.Second,
+	}
+
+	urlText := fmt.Sprintf(githubTemplate, url.PathEscape(moduleName))
```
nicks: should there be some stricter validation for what is a valid module name?

Someone is always going to put something weird in there (like a docker image name+tag), and get confused why it doesn't work

internal/tiltfile/extension/extension.go
jazzdan: Ah it looks like it does, I just wasn't looking far enough to the left.

internal/tiltfile/extension/store.go
landism: probably doesn't matter much, but maybe make an ExtensionTiltfileName const in this package?

internal/tiltfile/tiltfile_state.go
```diff
@@ -168,6 +169,7 @@ func (s *tiltfileState) loadManifests(absFilename string, userConfigState model.
 		telemetry.NewExtension(),
 		updatesettings.NewExtension(),
 		encoding.NewExtension(),
+		extension.NewExtension(extension.NewGithubFetcher(), extension.NewLocalStore(filepath.Dir(absFilename))),
```
landism: really feeling the extension/extension ambiguity here

maybe we should name the package `tiltextension`?

internal/tiltfile/extension/github_fetcher.go
```diff
@@ -0,0 +1,49 @@
...
+	if resp.StatusCode != http.StatusOK {
+		return ModuleContents{}, fmt.Errorf("error fetching Tiltfile %q: %v", urlText, err)
+	}
+
+	body, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return ModuleContents{}, err
+	}
+
+	return ModuleContents{
```
landism: probably worth a TODO to populate the other fields here, or is there a ticket? (or is the plan to do these this week?)

internal/tiltfile/extension/github_fetcher.go
landism: 👍

internal/tiltfile/extension/extension.go
jazzdan: I thought we agreed to ban it for now, though if we want to revisit that decision I'm down.

internal/tiltfile/extension/github_fetcher.go
jazzdan: Yes good call, plan is to do these later, not sure when.

PR #2930, opened 2020-02-11
tiltfile: enable recursion

landism:
what's the worst that could happen?

sample use case:
I want to write a function that traverses k8s yaml and sets namespaces


landism:
> seems fine to me, tho i wonder if we should be giving people some sort of object-patching language, so that they don't need to recursively traverse structs.

Yeah, that'd be nice. I'm not aware of any that'd gracefully handle the "no metadata exists, so create a metadata and then put a namespace in there" for the specific namespace case, though.

PR #2944, opened 2020-02-12
engine: clearer wording about update fallback [ch4492]

maiamcc:
currently we just say "falling back to next update method", which is pretty opaque esp. to someone who doesn't know how our updating works under the hood. In particular, it's not at all obvious that this message means "we can't run a live update, doing a full build instead".

this is the first step towards making this messaging better:
- name the update that we failed to do (so the user knows this is about not being able to LIVE UPDATE e.g.)
- HOPEFULLY being a little clearer about "fall back" means?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -57,3 +57,7 @@ yarn-error.log*
 
 # http caches from running kubectl in integration tests
 integration/.kube
+
+### Files owned by indivual developers
+# I want to be able to play with my Tiltfile w/o committing on unrelated projects
+Tiltfile.maia
--- a/.gitignore
+++ b/.gitignore
@@ -57,7 +57,3 @@ yarn-error.log*
 
 # http caches from running kubectl in integration tests
 integration/.kube
-
-### Files owned by indivual developers
-# I want to be able to play with my Tiltfile w/o committing on unrelated projects
-Tiltfile.maia
--- a/internal/engine/docker_compose_build_and_deployer.go
+++ b/internal/engine/docker_compose_build_and_deployer.go
@@ -55,10 +55,6 @@ func (bd *DockerComposeBuildAndDeployer) extract(specs []model.TargetSpec) ([]mo
 	return iTargets, dcTargets
 }
 
-func (bd *DockerComposeBuildAndDeployer) MethodName() string {
-	return "[Image Build and] Docker Compose Deploy"
-}
-
 func (bd *DockerComposeBuildAndDeployer) BuildAndDeploy(ctx context.Context, st store.RStore, specs []model.TargetSpec, currentState store.BuildStateSet) (store.BuildResultSet, error) {
 	iTargets, dcTargets := bd.extract(specs)
 	if len(dcTargets) != 1 {
--- a/internal/engine/docker_compose_build_and_deployer.go
+++ b/internal/engine/docker_compose_build_and_deployer.go
@@ -55,6 +55,10 @@ func (bd *DockerComposeBuildAndDeployer) extract(specs []model.TargetSpec) ([]mo
 	return iTargets, dcTargets
 }
 
+func (bd *DockerComposeBuildAndDeployer) MethodName() string {
+	return "[Image Build and] Docker Compose Deploy"
+}
+
 func (bd *DockerComposeBuildAndDeployer) BuildAndDeploy(ctx context.Context, st store.RStore, specs []model.TargetSpec, currentState store.BuildStateSet) (store.BuildResultSet, error) {
 	iTargets, dcTargets := bd.extract(specs)
 	if len(dcTargets) != 1 {
--- a/internal/engine/image_build_and_deployer.go
+++ b/internal/engine/image_build_and_deployer.go
@@ -102,8 +102,6 @@ func (ibd *ImageBuildAndDeployer) SetInjectSynclet(inject bool) {
 	ibd.injectSynclet = inject
 }
 
-func (ibd *ImageBuildAndDeployer) MethodName() string { return "[Image Build and] k8s Deploy]" }
-
 func (ibd *ImageBuildAndDeployer) BuildAndDeploy(ctx context.Context, st store.RStore, specs []model.TargetSpec, stateSet store.BuildStateSet) (resultSet store.BuildResultSet, err error) {
 	iTargets, kTargets := extractImageAndK8sTargets(specs)
 	if len(kTargets) != 1 {
--- a/internal/engine/image_build_and_deployer.go
+++ b/internal/engine/image_build_and_deployer.go
@@ -102,6 +102,8 @@ func (ibd *ImageBuildAndDeployer) SetInjectSynclet(inject bool) {
 	ibd.injectSynclet = inject
 }
 
+func (ibd *ImageBuildAndDeployer) MethodName() string { return "[Image Build and] k8s Deploy]" }
+
 func (ibd *ImageBuildAndDeployer) BuildAndDeploy(ctx context.Context, st store.RStore, specs []model.TargetSpec, stateSet store.BuildStateSet) (resultSet store.BuildResultSet, err error) {
 	iTargets, kTargets := extractImageAndK8sTargets(specs)
 	if len(kTargets) != 1 {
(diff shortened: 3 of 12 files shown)


maiamcc:
Screen Shot 2020-02-12 at 2 21 43 PM

internal/engine/docker_compose_build_and_deployer.go
```diff
@@ -55,6 +55,10 @@ func (bd *DockerComposeBuildAndDeployer) extract(specs []model.TargetSpec) ([]mo
 	return iTargets, dcTargets
 }
 
+func (bd *DockerComposeBuildAndDeployer) MethodName() string {
+	return "[Image Build and] Docker Compose Deploy"
```
nicks: what's with the square brackets?

internal/engine/image_build_and_deployer.go
```diff
@@ -102,6 +102,8 @@ func (ibd *ImageBuildAndDeployer) SetInjectSynclet(inject bool) {
 	ibd.injectSynclet = inject
 }
 
+func (ibd *ImageBuildAndDeployer) MethodName() string { return "[Image Build and] k8s Deploy]" }
```
nicks: ditto here, the non-matching square brackets seem odd

internal/engine/live_update_build_and_deployer.go
```diff
@@ -204,14 +206,14 @@ func liveUpdateInfoForStateTree(stateTree liveUpdateStateTree) (liveUpdInfo, err
...
+		var pathsMatchingNoSync []string
+		fileMappings, pathsMatchingNoSync, err = build.FilesToPathMappings(filesChanged, luInfo.SyncSteps())
 		if err != nil {
 			return liveUpdInfo{}, err
 		}
-		if len(skipped) > 0 {
+		if len(pathsMatchingNoSync) > 0 {
 			return liveUpdInfo{}, buildcontrol.RedirectToNextBuilderInfof("found file(s) not matching a LiveUpdate sync, so "+
-				"performing a full build. (Files: %s)", strings.Join(skipped, ", "))
+				"performing a full build. (Files: %s)", strings.Join(pathsMatchingNoSync, ", "))
```
nicks: as long as we're here, have we considered using ospath.FileListDisplayNames to pretty up this file list?

maiamcc:
> But my understanding (from talking to @hyu ) is that the words aren't really the problem, it's that they get lost in a sea of other words, and that we were looking for ways to visually distinguish them

Can one of you tell me the evidence behind this? My instinct is that the difficulty understanding LU fallbacks came from _both_ the difficulty find the log lines AND the uninformative message once you found them. Before this change, it was a crap-shoot whether the fallback message even mentioned the words "live update", and i can't imagine many people were able to decipher it.

nicks:
hmmmm...i wasn't really looking for a debate, was just curious what the current thinking was, happy to chat more about it in person

hyu:
Thanks, @maiamcc — here's my own bits of minor feedback:
- Can we consistently use "Live Update"? I see some variations but we just need to pick one.
- I also think we should pretty up the file list. The absolute path is a lot of characters.
- I do think the error message can be more concise. This is a high ratio of "ink to data":

```
Couldn't perform update via method: Live Update because:
       found file(s) not matching a LiveUpdate sync, so performing a full build
(Files: /blah/blahb/blah/blahb/blahb/blahb/blahb/blahb)

Falling back to next update method
```

I would prefer something like:

```
Live Update didn't work, so trying a Full Build. Reason:
    Found file(s) not matching a LiveUpdate sync: 
    - /blah/blah
```

Also, @nicks — the words are actually still a problem since we don't specifically name "Live Update" fallbacks. My prior PR does some work to visually distinguish all fallbacks, but that was not specific to Live Update. @maiamcc's work here also sets us up to further distinguish Live Update with, say, an icon.

maiamcc:
Yes, I for sure want to find a way to trim down the error text!

> Live Update didn't work, **so trying a Full Build**...

We can't do this at the moment [because of this](https://github.com/windmilleng/tilt/pull/2944/files#diff-7396551d13039e7bceb0b3a6cd301c13R83) -- though I'm inclined to fix this in a later PR.

Give me a sec to go trim copy on the error message itself, though.

internal/engine/live_update_build_and_deployer.go
maiamcc: will do!

maiamcc:
@nicks @hyu ready for re-review.

FYI re: `"[Image Build and] Docker Compose Deploy"` -- going to leave that for now and fix it in a PR I'm making right now, where I calculate the human-readable name of the update method from the specs we're passing.

maiamcc:
> Yes, I for sure want to find a way to trim down the error text!
> 
> > Live Update didn't work, **so trying a Full Build**...
> 
> We can't specify what build is coming next at the moment [because of this](https://github.com/windmilleng/tilt/pull/2944/files#diff-7396551d13039e7bceb0b3a6cd301c13R83) -- though I'm inclined to fix this in a later PR.
> 
> Give me a sec to go trim copy on the error message itself, though.

I'm a doofus, the only thing we fall back to from a Live Update is a full build -- I was trying to make this code too generic, but the only fallback that really needs messaging around is Live Update ->> full build, I'll just special-case that.

maiamcc:
...okay NOW it's ready for review. This should be a lot simpler.

Screen Shot 2020-02-12 at 6 34 29 PM
Screen Shot 2020-02-12 at 5 50 30 PM
Screen Shot 2020-02-12 at 5 50 25 PM

internal/engine/image_build_and_deployer.go
hyu: When I saw the brackets, I also wasn't sure how to interpret them. My perspective is — when we're telling users about what Tilt is doing under the hood, we don't necessarily need to be terse and technical. We can't control the error messages that we just pass along, but we're can be thoughtful with the "voice" of the app when it's messaging about how Tilt is working (or not). 

So, just spitballing along those lines. Instead of a bracket, how about something like....
```
k8s deploy, possibly including an image build
```
?

internal/engine/live_update_build_and_deployer.go
```diff
@@ -204,14 +208,15 @@ func liveUpdateInfoForStateTree(stateTree liveUpdateStateTree) (liveUpdInfo, err
...
 		if err != nil {
 			return liveUpdInfo{}, err
 		}
-		if len(skipped) > 0 {
-			return liveUpdInfo{}, buildcontrol.RedirectToNextBuilderInfof("found file(s) not matching a LiveUpdate sync, so "+
-				"performing a full build. (Files: %s)", strings.Join(skipped, ", "))
+		if len(pathsMatchingNoSync) > 0 {
+			prettyPaths := ospath.FileListDisplayNames(iTarget.LocalPaths(), pathsMatchingNoSync)
+			return liveUpdInfo{}, buildcontrol.RedirectToNextBuilderInfof(
+				"found file(s) not matching any sync (files: %s)", strings.Join(prettyPaths, ", "))
```
hyu: Silly little nit, but: do these have to be lowercase? And for my own edification, where are you setting the amount of indentation?

.gitignore
```diff
@@ -56,4 +56,8 @@ yarn-debug.log*
 yarn-error.log*
 
 # http caches from running kubectl in integration tests
-integration/.kube
\ No newline at end of file
+integration/.kube
+
+### Files owned by indivual developers
+# I want to be able to play with my Tiltfile w/o committing on unrelated projects
+Tiltfile.maia
```
nicks: have you considered using a global gitignore for this?

internal/engine/image_build_and_deployer.go
maiamcc: This has been changed! Lmk if you still have questions or suggestions but it should be much clearer now. Tl/dr: if we're falling back to it from a LiveUpdate, it necessarily includes an image build (b/c we will only try to live update something if it has an image target).

internal/engine/live_update_build_and_deployer.go
maiamcc: I'm just using `\t`; is there a more easily modified way to do that? (I was assuming that if we wanted to style it separately, we'd have to make it its own log line with it's own field identifier...?)

.gitignore
maiamcc: 😳 yes will do

Issue #2943, opened 2020-02-12
Dockerfile image ref injection between dependent images fails if tag is a build arg


elemental-lf:
Image reference injection into a child's `Dockerfile` fails if the tag is a reference to a build argument. The injection works as expected when the tag is missing or is simply a string. 

It also fails if the whole image reference is a build argument (`FROM ${IMAGE_REF}`) even when the build argument references the parent image. But I suspect this might be harder the fix than the first use case.

I tried to work around this limitation by using `custom_build()` and `$EXPECTED_REF` and this works to some degree but fails in the end because there is no way to express the dependency between the two images and so the parent image might not exist when the custom build is started.

This is with version `0.12.1`.

### Tiltfile

```python
docker_build(
    'parent',
    'parent',
    dockerfile='Dockerfile.parent',
)
docker_build(
    'child',
    'child',
    dockerfile='Dockerfile.child',
)
```

### Dockerfile.parent

```dockerfile
FROM scratch
```

### Dockerfile.child
```dockerfile
ARG TAG="latest"
# This reference is not replaced as expected.
FROM parent:${TAG}
```

elemental-lf:
For my unsuccessful `custom_build()` workaround there is some related discussion in https://github.com/windmilleng/tilt/issues/2856 as I've just noticed.

nicks:
Is it possible to rewrite Dockerfile.child to use a FROM with a literal image reference?

Ya, there's a couple of different ways you could approach this:
- Make docker_build smarter about injecting into Dockerfiles with build args
- Make custom_build extensible enough that it can depend on other image builds

It might help to have a little bit more context on what you're trying to do. Is the real dockerfile you're using exactly `FROM parent:${TAG}`, and would making it smart enough to handle that pattern be good enough? Or would it need to be a little smarter?

elemental-lf:
> It might help to have a little bit more context on what you're trying to do. Is the real dockerfile you're using exactly `FROM parent:${TAG}`, and would making it smart enough to handle that pattern be good enough? Or would it need to be a little smarter?

Thank you for the quick response! The use case is actually similar to the one Tilt has. The build pipeline of the project in question builds multiple versions of a set of dependent images. All images belonging to one version have the same tag. To ensure that a child image is based on the correct parent image the build process injects the tag of the parent image to use as a build argument. So as far as I can see support for `FROM parent:${TAG}` would be sufficient to use Tilt together with these `Dockerfile`s.

nicks:
[...]

PR #2950, opened 2020-02-13
dockerfile: handle ARG variables when injecting images. Fixes issue #2943

nicks:
Hello @maiamcc, @landism,

Please review the following commits I made in branch nicks/issue2943:

03bdc9e205a085e36ef39c20567a4e96e04bb3b0 (2020-02-13 00:35:23 -0500)
dockerfile: handle ARG variables when injecting images. Fixes issue #2943

### DIFF
--- a/internal/dockerfile/dockerfile_test.go
+++ b/internal/dockerfile/dockerfile_test.go
@@ -187,3 +187,15 @@ func TestFindImagesCopyFrom(t *testing.T) {
 		assert.Equal(t, "gcr.io/image-a", images[0].String())
 	}
 }
+
+func TestFindImagesWithArg(t *testing.T) {
+	df := Dockerfile(`
+ARG TAG="latest"
+FROM gcr.io/image-a:${TAG}
+`)
+	images, err := df.FindImages()
+	assert.NoError(t, err)
+	if assert.Equal(t, 1, len(images)) {
+		assert.Equal(t, "gcr.io/image-a:latest", images[0].String())
+	}
+}
--- a/internal/dockerfile/ast.go
+++ b/internal/dockerfile/ast.go
@@ -12,6 +12,7 @@ import (
 	"github.com/moby/buildkit/frontend/dockerfile/dockerfile2llb"
 	"github.com/moby/buildkit/frontend/dockerfile/instructions"
 	"github.com/moby/buildkit/frontend/dockerfile/parser"
+	"github.com/moby/buildkit/frontend/dockerfile/shell"
 	"github.com/pkg/errors"
 
 	"github.com/windmilleng/tilt/internal/container"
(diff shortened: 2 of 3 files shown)


internal/dockerfile/dockerfile_test.go
```diff
@@ -174,9 +174,7 @@ func TestFindImagesWeirdSyntax(t *testing.T) {
 	df := Dockerfile(`FROM a b`)
 	images, err := df.FindImages()
 	assert.NoError(t, err)
-	if assert.Equal(t, 1, len(images)) {
-		assert.Equal(t, "docker.io/library/a", images[0].String())
-	}
+	assert.Equal(t, 0, len(images))
```
maiamcc: hmm, why did this change?

nicks: good question. i went back and forth on this. it's a question of how robustly we handle parse errors. i changed my mind and changed it back!

bkcsfi:
Hi @nicks It looks like this change only handles ARGS expansion in the tag part. It looks like the following is not expected to work in Dockerfile:

```docker
ARG SFI_BASE=sfi-base
FROM ${SFI_BASE}
```

I am using the same dockerfiles with both Tilt and docker buildx bake. For the later, I need to explicitly indicate the full FROM to use (registry, base name and tag)

Therefore in my Tilt file, I work around this issue, like this:

```python
    docker_build(
        "zope-server-image",
        ".",
        dockerfile_contents=str(read_file("docker/Dockerfile-zope-server")).replace(
            "FROM ${SFI_BASE}", "FROM sfi-base"
        )
    )
```
I'm ok with this workaround. But I wanted to be sure this issue/PR wasn't intended to also handle the case where the entire FROM is an ARG.

Thanks

nicks:
@bkcsfi ya, if you go back to the original linked issue, there was some discussion about what patterns we needed to support, and I didn't try too hard to cover every case. I can file an issue for the other case if that's exactly what you need.

nicks:
@bkcsfi Hey, I thought about your question a bit more, and I'm super confused! The current code should handle the `FROM ${SFI_BASE}` case correctly, and I'm pretty sure I've written a couple example projects that do exactly what you're doing.

Here's an example:
https://github.com/windmilleng/tilt-example-html/blob/nicks/basearg/2-recommended/Tiltfile
https://github.com/windmilleng/tilt-example-html/blob/nicks/basearg/2-recommended/Dockerfile

is there something I'm misunderstanding about your issue, or some important bit that example doesn't capture?

maiamcc:
Yup, I tried to repro here: https://github.com/windmilleng/dependent-images-via-args and couldn't get it to break. Did you have any luck making a minimal repro case, either on your own or starting from one of the samples Nick or I linked?

PR #2952, opened 2020-02-13
A Tilt Tiltfile

jazzdan:


.vscode/launch.json
```diff
@@ -0,0 +1,17 @@
+{
+  // Use IntelliSense to learn about possible attributes.
```
jazzdan: whoops will delete this

Tiltfile
```diff
@@ -0,0 +1,50 @@
...
+  local_resource("go_lint", "make lint", deps=all_go_files)
+
+def get_all_ts_files(path):
+  return str(local('cd %s && find . -type f -name "*.ts*" | grep -v node_modules | grep -v __snapshots__' % path)).split("\n")
+
+def yarn_install():
+  local_resource("yarn_install", "cd web && yarn", deps=['web/package.json', 'web/yarn.lock'])
+
+def jest(path):
+  local_resource("web_jest", "", serve_cmd="cd %s && yarn run test --bail" % path, resource_deps=["yarn_install"])
```
landism: nit: `cmd` is optional and defaults to empty

```suggestion
  local_resource("web_jest", serve_cmd="cd %s && yarn run test --bail" % path, resource_deps=["yarn_install"])
```

landism: I don't have strong feelings either way, but alternatively to `--bail`, could consider `--notify`, which wouldn't result in the tilt status being updated, but would at least communicate to you when your tests are failing

web/package.json
```diff
@@ -37,6 +37,7 @@
     "build": "react-scripts build",
     "test": "react-scripts test",
     "ci": "react-scripts test --runInBand --reporters=default --reporters=jest-junit",
+    "test-bail": "react-scripts test --bail",
```
landism: is this used, or left over from an earlier iteration?

PR #2954, opened 2020-02-13
k8srollout: tweak pod rollout logs

hyu:
- Use a dotted box-drawing vertical to visually group Pod Rollout events
- The aligned durations makes it clearer when something is Pending, without using an emoji (which doesn't currently render well on RTY)
- No longer using the ❌ emoji, which was alarming

Please note this PR does not include additional visual changes to the Web UI, which may come as a separate PR.

```
 Tracking new pod rollout (blog-site-55bf44cbf9-lfx2n):
      ┊ Scheduled       - <1s
      ┊ Initialized     - (…) Pending
      ┊ Ready           - (…) Pending
```

The "Not.." progress events have a slightly different prefix:

```
      ┊ Scheduled       - <1s
      ┃ Not Ready       - (ContainersNotReady): containers with unready status: [docs-site tilt-synclet]
```

Screen Shot 2020-02-13 at 2 30 59 PM copy

## Previously

Screen Shot 2020-02-06 at 5 07 10 PM

### DIFF
--- a/internal/engine/k8srollout/podmonitor_test.go
+++ b/internal/engine/k8srollout/podmonitor_test.go
@@ -2,11 +2,12 @@ package k8srollout
 
 import (
 	"context"
+	"fmt"
 	"io"
+	"io/ioutil"
 	"testing"
 	"time"
 
-	"github.com/stretchr/testify/assert"
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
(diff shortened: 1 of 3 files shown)


internal/engine/k8srollout/podmonitor_test.go
```diff
@@ -18,6 +19,10 @@ import (
 	"github.com/windmilleng/tilt/pkg/model"
 )
 
+// NOTE(han): set at runtime with:
+// go test -ldflags="-X 'github.com/windmilleng/tilt/internal/engine/k8srollout.PodmonitorWriteGoldenMaster=1'" ./internal/engine/k8srollout
+var PodmonitorWriteGoldenMaster = "0"
```
hyu: I really like this Golden Master testing method that @jazzdan implemented a while back, because it makes testing log output SO much less painful. Originally it was just **pipeline_state_test.go**, then I added it to **logstore_test.go** and now here. So the assertSnapshot function is duplicated in three places. Should / can this be more DRY?

internal/engine/k8srollout/podmonitor_test.go
```diff
@@ -113,3 +115,22 @@ func (s *testStore) Dispatch(action store.Action) {
 		_, _ = s.out.Write(logAction.Message())
 	}
 }
+
+func assertSnapshot(t *testing.T, output string) {
```
hyu: I really like this Golden Master testing method that @jazzdan implemented a while back, because it makes testing log output SO much less painful. Originally it was just pipeline_state_test.go, then I added it to logstore_test.go and now here. So now this ``assertSnapshot` function is duplicated in three places. 

Should / can this be more DRY?

PR #2955, opened 2020-02-13
web: last build duration in sidebar [ch4520]

maiamcc:
This PR adds build duration to the sidebar, adjusting the layout so everything can fit. 

- Status indicator used to be a dot or donut, but it's now a larger color block. This way, we can show the alert badge number here. This frees up some extra space and reduces visual clutter.
- Status items have a new look (more like buttons), a pre-req for the new status indicator design.

2020-02-18 - sidebar with build duration

***
Original message from Maia (2020-02-14):

left some kind of shitty duration formatting b/c I wasn't sure what you wanted there.

CSS styling is similarly just a placeholder


hyu:
As discussed with @maiamcc on Slack, I'm going to take over this branch and make some further changes.

web/src/Sidebar.tsx
```diff
@@ -20,6 +22,7 @@ class SidebarItem {
   isTiltfile: boolean
   status: ResourceStatus
   hasEndpoints: boolean
+  lastBuildDur: Date | null
```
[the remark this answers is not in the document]
hyu: Not sure if this is _the way_, but I tend to think of it case by case. Build Duration is a thing that we optionally show, only when there exists a last build with a successfully completed duration. So I would expect it to be undefined in lots of cases: before a given resource has ever been updated (like on initial startup), or maybe when a last build was aborted (with manual trigger).

So, I think Last Build Duration can be null in these cases where we decide we don't want to show it. (Like, we should tell you how much time elapsed because you canceled the last build, but what's the point of that?)

web/src/Sidebar.tsx
```diff
@@ -5,12 +5,25 @@ import { combinedStatus } from "./status"
...
+  Color,
+  ColorAlpha,
+  ColorRGBA,
+  FontSize,
+  Height,
+  SizeUnit,
+  Width,
+} from "./style-helpers"
+
+const moment = require("moment")
```
maiamcc: i didn't actually look this up, i just blindly copy/pasted -- is this the right way to pull in another library like this?

web/src/Sidebar.tsx
```diff
@@ -33,24 +47,134 @@ class SidebarItem {
...
     this.queued = !!res.queued
-    let buildHistory = res.buildHistory || []
-    if (buildHistory.length > 0) {
-      this.lastBuild = buildHistory[0]
-    }
+    this.lastBuild = lastBuild
   }
 }
 
+const barberpole = keyframes`
```
maiamcc: 👀 what's all this CSS doing in a .tsx file?

web/src/Sidebar.tsx
```diff
@@ -172,6 +288,16 @@ class Sidebar extends PureComponent<SidebarProps> {
   }
 }
 
+function timeDiff(start: string, end: string): Date {
+  let t1 = moment(start)
+  let t2 = moment(end)
+  return t2.diff(t1)
+}
+
+function formatDuration(dur: Date): string {
```
maiamcc: this is pretty hacky, maybe leave a note/todo. i imagine we want at minimum the ability to display durations of 1m+ as like mm:ss

web/src/__snapshots__/AlertPane.test.tsx.snap
```diff
@@ -35,7 +35,7 @@ exports[`renders multiple lines of a crash log 1`] = `
             dateTime="1949-11-18T09:39:00.000Z"
             title="1,555,970,585,039"
           >
-            67yr
+            67yr ago
```
maiamcc: (it's also not strictly necessarily to leave the "ago" in here, i put it in to differentiate "build duration" from "time elapsed since last build", happy for another way to differentiate these that doesn't take up space in the sidebar!)

web/src/Sidebar.tsx
jazzdan: We use ES6 style imports, try:

```typescript
import moment from "moment"
```

web/src/Sidebar.tsx
hyu: Haha, whenever I write styling code I try to do so using [styled-components](https://styled-components.com/). I'm aware that this decision / guideline is not actually documented anywhere. Maybe it's time for that style guide to be written...

web/src/__snapshots__/AlertPane.test.tsx.snap
hyu: Yeah, I also find that "ago" is useful for distinguishing two tiny timestamps at a glance.

PR #2961, opened 2020-02-14
Fix tar creation to accept windows path

pleclech:
- Actualy ADD and COPY failed in Docker file if we use windows path.

The bug is that the path are stored as absolute with full volume name and not as relative as it should be when using unix path

### DIFF
--- a/internal/build/tar.go
+++ b/internal/build/tar.go
@@ -145,11 +145,6 @@ func (a *ArchiveBuilder) entriesForPath(ctx context.Context, source, dest string
 	// (PL) remove volume name + / so if we got a windows path C:\foo it becomes foo
 	dest = strings.TrimPrefix(dest, filepath.VolumeName(dest)+"/")
 
-	destWithSlash := dest
-	if !strings.HasSuffix(destWithSlash, "/") {
-		destWithSlash += "/"
-	}
-
 	result := make([]archiveEntry, 0)
 	err = filepath.Walk(source, func(path string, info os.FileInfo, err error) error {
 		if err != nil {
@@ -199,7 +194,7 @@ func (a *ArchiveBuilder) entriesForPath(ctx context.Context, source, dest string
 				return errors.Wrapf(err, "making rel path source:%s path:%s", source, path)
 			}
 			// ...and live inside `dest`
-			header.Name = destWithSlash + tmp //strings.TrimPrefix(path, source)
+			header.Name = dest + tmp
 		} else if strings.HasSuffix(dest, "/") {
 			header.Name = dest + filepath.Base(path)
 		} else {
--- a/internal/build/tar.go
+++ b/internal/build/tar.go
@@ -120,6 +120,9 @@ type archiveEntry struct {
 // e.g. tarring my_dir --> dest d: d/file_a, d/file_b
 // If source path does not exist, quietly skips it and returns no err
 func (a *ArchiveBuilder) entriesForPath(ctx context.Context, source, dest string) ([]archiveEntry, error) {
+	// (PL) convert \ to / in case of windows path
+	source = filepath.ToSlash(source)
+
 	sourceInfo, err := os.Stat(source)
 	if err != nil {
 		if os.IsNotExist(err) {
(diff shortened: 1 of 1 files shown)


nicks:
Hi @pleclech ! Thanks for the PR! 

Can you give a little more detail on the problem, maybe a repro case we could play with to make sure we don't re-break this?

It's unclear from the PR description if this is a bug with the Tilt Windows binary, or with Windows-based container images.

pleclech:
Hi @nicks 
Here the step to reproduce the issue:

# OS: windows 10 pro
# Tool: Docker for windows 2.2.0.3

#tilt version (v0.12.1, built 2020-02-12)
.\tilt version
  
#create kubernetes cluster with k3d (https://github.com/rancher/k3d/releases/download/v1.6.0/k3d-windows-amd64.exe)
.\k3d-windows-amd64 version
.\k3d-windows-amd64.exe create

# export kubeconfig
$env:KUBECONFIG=..\k3d-windows-amd64.exe get-kubeconfig --name='k3s-default'

#check cluster
kubectl get node -o wide

# clone tilt demo
git clone https://github.com/windmilleng/tiltdemo
cd tiltdemo

#launch tilt up
..\tilt up

# result in error :
error in ADD ./cmd/demoserver1 no such file or directory

windows-path-issue

build-result 112

# check whats in the container
# editing Dockerfile with
ADD . .
RUN ["ls", "-lrt"]

# got
→ total 56
→ -rw-rw-rw- 1 root root   248 Feb 14 21:00 C:\work\kubernetes\tilt-windows-issue\tiltdemo\cmd\demoserver2\main.go
→ drwxrwxrwx 2 root root  4096 Feb 14 21:00 C:\work\kubernetes\tilt-windows-issue\tiltdemo\cmd\demoserver2
→ -rw-rw-rw- 1 root root   329 Feb 14 21:00 C:\work\kubernetes\tilt-windows-issue\tiltdemo\cmd\demoserver1\main.go
→ drwxrwxrwx 2 root root  4096 Feb 14 21:00 C:\work\kubernetes\tilt-windows-issue\tiltdemo\cmd\demoserver1

files-list-result 0

internal/build/tar.go
```diff
@@ -120,6 +120,9 @@ type archiveEntry struct {
 // e.g. tarring my_dir --> dest d: d/file_a, d/file_b
 // If source path does not exist, quietly skips it and returns no err
 func (a *ArchiveBuilder) entriesForPath(ctx context.Context, source, dest string) ([]archiveEntry, error) {
+	// (PL) convert \ to / in case of windows path
+	source = filepath.ToSlash(source)
```
nicks: should there be separate variables for `sourceAsHostOS` and `sourceInContainer`? i want to make sure the `os.Stat` call below works correctly on Windows, and i think converting to slash here might make it not work right?

internal/build/tar.go
```diff
@@ -179,16 +189,18 @@ func (a *ArchiveBuilder) entriesForPath(ctx context.Context, source, dest string
 
 		if sourceIsDir {
 			// Name of file in tar should be relative to source directory...
-			header.Name = strings.TrimPrefix(path, source)
+			tmp, err := filepath.Rel(source, path)
```
nicks: In Go, the `filepath` package is for the host OS paths, and the `path` package is for slash-separated paths paths. so maybe all this code should just use `path.Rel`, `path.Join`, `path.Separator`?

internal/build/tar.go
```diff
@@ -179,16 +189,18 @@ func (a *ArchiveBuilder) entriesForPath(ctx context.Context, source, dest string
...
+			if err != nil {
+				return errors.Wrapf(err, "making rel path source:%s path:%s", source, path)
+			}
 			// ...and live inside `dest`
-			header.Name = filepath.Join(dest, header.Name)
-		} else if strings.HasSuffix(dest, string(filepath.Separator)) {
-			header.Name = filepath.Join(dest, filepath.Base(source))
+			header.Name = filepath.Join(dest, tmp)
+		} else if strings.HasSuffix(dest, "/") {
```
nicks: ya, this should use path.Separator

internal/build/tar.go
pleclech: Hi @nicks

AFAIK Windows will accept path that are in the form c:/xxxx/dsdsd, so just to get a common ground between Linux and Windows i transform everything to forward slash, but if you want to use separate variables there no problems either

internal/build/tar.go
pleclech: @nicks  There is no such function path.Rel this is why i use filepath.Rel.

internal/build/tar.go
pleclech: @nicks As i transform everything to forward slash i can't use filepath.Separator (which will be in windows '\' ), everything now can use only one separator that is '/'

pleclech:
@nicks  Thanks.

PR #2962, opened 2020-02-14
tiltfile: add config.define_{bool,string}

landism:
### Problem

Lots of use cases for config settings don't actually want a list of strings, and it's annoying that we make users deal with that type when it's so cheap to just provide other types (and when pretty much all other arg-parsing APIs provide other types). Especially since for a bool you really want a flag, but the existing syntax makes you say `--foo true`

### Solution

add setting types for bool and string

### DIFF
--- a/internal/tiltfile/config/bool.go
+++ b/internal/tiltfile/config/bool.go
@@ -1,0 +1,72 @@
+package config
+
+import (
+	"bytes"
+	"encoding/json"
+	"flag"
+	"fmt"
+	"strconv"
+
+	"go.starlark.net/starlark"
+)
+
+type boolSetting struct {
+	value bool
+	isSet bool
+}
+
+var _ configValue = &boolSetting{}
+var _ flag.Value = &boolSetting{}
+
+func (s *boolSetting) starlark() starlark.Value {
+	return starlark.Bool(s.value)
+}
+
+func (s *boolSetting) IsSet() bool {
+	return s.isSet
+}
+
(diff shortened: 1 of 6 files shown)


internal/tiltfile/config/bool.go
```diff
@@ -0,0 +1,72 @@
...
+func (s *boolSetting) starlark() starlark.Value {
+	return starlark.Bool(s.value)
+}
+
+func (s *boolSetting) IsSet() bool {
+	return s.isSet
+}
+
+func (s *boolSetting) MarshalJSON() ([]byte, error) {
+	ret := &bytes.Buffer{}
```
nicks: Is this equivalent to json.Marshal(s.value)?

internal/tiltfile/config/string.go
```diff
@@ -0,0 +1,64 @@
...
+func (s *stringSetting) starlark() starlark.Value {
+	return starlark.String(s.value)
+}
+
+func (s *stringSetting) IsSet() bool {
+	return s.isSet
+}
+
+func (s *stringSetting) MarshalJSON() ([]byte, error) {
+	ret := &bytes.Buffer{}
```
nicks: Ditto

internal/tiltfile/config/bool.go
landism: yes, AND we don't need to marshal anyway - this is copypasta of cruft left over from when we were persisting these values

PR #2971, opened 2020-02-17
pkg: export tilt extension name validation for use in CI

jazzdan:


internal/tiltfile/tiltextension/github_fetcher.go
```diff
@@ -6,6 +6,8 @@ import (
 	"io/ioutil"
 	"net/http"
 	"time"
+
+	pkgTiltextension "github.com/windmilleng/tilt/pkg/tiltextension"
```
landism:
```suggestion
	pkgtiltextension "github.com/windmilleng/tilt/pkg/tiltextension"
```

if it's all the same to you, I find the capital T super grating

Issue #2960, opened 2020-02-14
decode_yaml should handle multiple resources


nicks:
Helm usually exports templates as a list of YAML resources separated by "\n---\n"

decode_yaml can only parse one resource at a time, so right now you need to split them manually

It would be nice if there was a build-in for handling multiple resources

For an example use-case, see https://github.com/windmilleng/tilt/issues/2951

PR #2973, opened 2020-02-18
tiltfile: add read/decode/encode_yaml_stream functions

landism:
addresses #2960 

### Problem

Various k8s tools involve YAML streams rather than YAML docs (i.e., multiple YAML docs in the same string, separated by "\n---\n"). `decode_yaml` / `read_yaml` / `encode_yaml` don't handle these (and worse, `decode_yaml` and `read_yaml` just return the first doc in the stream and throw the rest away)

### Solution

1. Add `decode_yaml_stream`, `read_yaml_stream`, `encode_yaml_stream` functions, which operate work on streams of YAML docs
2. Make `decode_yaml` and `read_yaml` fail with an error pointing at the `_stream` variants if they're passed a stream, to protect people from unintentionally silently dropping yaml docs past the first

Another option would have been to just have `decode_yaml` / `read_yaml` return an array of objects when there's a stream, but it seems like it'd be annoying to have it return an array when you know just a single object and annoying to have to variously return a single object or an array depending on what string you pass it (also, then the caller wouldn't be able to distinguish between return values of an array of yaml documents, or a single yaml document with a top-level array). FWIW, both the Ruby and Python YAML libraries have separate _stream functions for handling YAML streams.

### DIFF
--- a/internal/tiltfile/encoding/extension.go
+++ b/internal/tiltfile/encoding/extension.go
@@ -8,18 +8,34 @@ func NewExtension() Extension {
 	return Extension{}
 }
 
+const (
+	readYAMLN         = "read_yaml"
+	readYAMLStreamN   = "read_yaml_stream"
+	decodeYAMLN       = "decode_yaml"
+	decodeYAMLStreamN = "decode_yaml_stream"
+	encodeYAMLN       = "encode_yaml"
+	encodeYAMLStreamN = "encode_yaml_stream"
+
+	readJSONN   = "read_json"
+	decodeJSONN = "decode_json"
+	encodeJSONN = "encode_json"
+)
+
 func (Extension) OnStart(env *starkit.Environment) error {
 	for _, b := range []struct {
 		name string
 		f    starkit.Function
 	}{
-		{"read_yaml", readYAML},
-		{"decode_yaml", decodeYAML},
-		{"encode_yaml", encodeYAML},
+		{readYAMLN, readYAML},
+		{readYAMLStreamN, readYAMLStream},
+		{decodeYAMLN, decodeYAML},
+		{decodeYAMLStreamN, decodeYAMLStream},
+		{encodeYAMLN, encodeYAML},
+		{encodeYAMLStreamN, encodeYAMLStream},
 
-		{"read_json", readJSON},
-		{"decode_json", decodeJSON},
-		{"encode_json", encodeJSON},
+		{readJSONN, readJSON},
+		{decodeJSONN, decodeJSON},
+		{encodeJSONN, encodeJSON},
 	} {
 		err := env.AddBuiltin(b.name, b.f)
 		if err != nil {
--- a/internal/tiltfile/encoding/yaml.go
+++ b/internal/tiltfile/encoding/yaml.go
@@ -16,14 +16,22 @@ import (
 	"github.com/windmilleng/tilt/internal/tiltfile/value"
 )
 
-// reads yaml from a file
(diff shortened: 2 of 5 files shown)


internal/tiltfile/encoding/extension.go
```diff
@@ -8,18 +8,34 @@ func NewExtension() Extension {
 	return Extension{}
 }
 
+const (
+	readYAMLN         = "read_yaml"
```
nicks: out of curiosity, what's the thinking behind moving these to N-suffixed constants?

internal/tiltfile/encoding/yaml.go
```diff
@@ -35,11 +38,26 @@ func readYAML(thread *starlark.Thread, fn *starlark.Builtin, args starlark.Tuple
...
-	return yamlStringToStarlark(string(contents), path.GoString())
+	return yamlStreamToStarlark(string(contents), path.GoString())
+}
+
+func readYAML(thread *starlark.Thread, fn *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
+	v, err := readYAMLStream(thread, fn, args, kwargs)
+	if err != nil {
+		return nil, err
+	}
+	l, ok := v.(*starlark.List)
```
nicks: have you considered an intermediate function that returns the concrete type, so that you don't need to do the type-casting square dance?

internal/tiltfile/encoding/yaml.go
```diff
@@ -50,29 +68,55 @@ func decodeYAML(thread *starlark.Thread, fn *starlark.Builtin, args starlark.Tup
...
-		}
-		return nil, errors.Wrap(err, errmsg)
+		return nil, err
+	}
+	l, ok := v.(*starlark.List)
+	if !ok {
+		return nil, fmt.Errorf("internal error: expected decodeYAMLStream to return a %T, but it returned a %T", starlark.List{}, l)
+	}
+	if l.Len() != 1 {
+		return nil, fmt.Errorf("expected a yaml document but found a yaml stream (documents separated by `---`). use %s instead to decode a yaml stream", decodeYAMLStreamN)
```
nicks: what happens if you call decode_yaml('') on an empty string? does it hit this error case?

internal/tiltfile/encoding/yaml.go
```diff
@@ -90,6 +134,28 @@ func encodeYAML(thread *starlark.Thread, fn *starlark.Builtin, args starlark.Tup
 	return tiltfile_io.NewBlob(ret, "encode_yaml"), nil
 }
 
+func encodeYAMLStream(thread *starlark.Thread, fn *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
```
nicks: conversely, should it be an error if you try to call encodeYAML on a list?

internal/tiltfile/encoding/extension.go
landism: mostly that I wanted to reference the read_yaml_stream and decode_yaml_stream function names from the errors when read_yaml or decode_yaml get streams instead of docs

internal/tiltfile/encoding/yaml.go
landism: nope, that's a good idea

internal/tiltfile/encoding/yaml.go
landism: good catch

internal/tiltfile/encoding/yaml.go
landism: nope - a list is a valid yaml document, e.g. `encode_yaml(["foo", "bar"]) == "- foo\n- bar\n"`

Issue #2970, opened 2020-02-17
Installation script should detect if there's already something named `tilt` in $PATH


jazzdan:
A [common problem](https://docs.tilt.dev/faq.html#q-when-i-run-tilt-version-i-see-template-engine-not-found-for-version-what-do-i-do) that folks run in to installing Tilt is that there's already something called Tilt installed in their `$PATH` (namely that Ruby gem)

Could we detect this case and warn/error in the installation script?

maiamcc:
@landism can we close this now?

PR #2974, opened 2020-02-18
install: check for other tilts on path

landism:
#2970


scripts/install.sh
```diff
@@ -35,7 +35,29 @@ else
...
 
-# TODO(nick): Add verification that Tilt installed successfully.
-
 set +x
-echo "Tilt installed! Run \`tilt up\` to start."
+
+
+VERSION="$(tilt version 2>&1 || true)"
+RUBY_TILT_PATTERN="template engine not found"
+TILT_DEV_PATTERN='^v[0-9]+\.[0-9]+\.[0-9]+(-dev)?, built [0-9]+-[0-9]+-[0-9]+$'
```
nicks: oh boy, this regexp seems very likely to break. can you add a ci job that verifies this works? maybe something that just runs the install script and checks the exit code?

landism: good idea!

PR #2977, opened 2020-02-18
cli: 'alpha tiltfile-result ' cli commmand to dump json of tiltfile load result

maiamcc:
a slap-dash first pass at a command to execute the Tiltfile and dump the result json.

requested both for prod pipelining and for Tiltfile testing/verification

note that we are NOT committing to keeping this command's API/the TiltfileLoadResult API consistent
(and if you think that's not 110% clear in the help text, pls suggest better language)


internal/cli/tiltfile_result.go
```diff
@@ -0,0 +1,87 @@
...
+
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+	tiltanalytics "github.com/windmilleng/tilt/internal/analytics"
+	"github.com/windmilleng/tilt/internal/tiltfile"
+	"github.com/windmilleng/tilt/pkg/logger"
+	"github.com/windmilleng/tilt/pkg/model"
+)
+
+const TiltfileErrExitCode = 5
```
maiamcc: picked this exit code arbitrarily as something that wasn't reserved for a specific meaning, but very open to suggestions on how i handle exit codes--the two cases I was trying to account for (a. something unexpected went wrong while trying to do the work, and b. there was a(n expected) error in the work) don't seem specifically addressed in e.g. [the guide nick linked](https://www.tldp.org/LDP/abs/html/exitcodes.html#EXITCODESREF).

internal/cli/tiltfile_result.go
```diff
@@ -0,0 +1,87 @@
...
+	ctx = tiltanalytics.WithAnalytics(ctx, a)
+
+	deps, err := wireTiltfileResult(ctx, a)
+	if err != nil {
+		failWithUnexpectedError(errors.Wrap(err, "wiring dependencies"))
+	}
+
+	tlr := deps.tfl.Load(ctx, fileName, model.NewUserConfigState(args))
+	if tlr.Error != nil {
+		// Some errors won't JSONify properly--instead of messing with that, print the error
```
landism: I was mildly confused by this comment - it says instead of jsonifying the error, we're just going to print it, but then it jsonifies it instead of printing it?

How about just printing the error directly to stderr and exiting, and not doing any json?

internal/cli/tiltfile_result.go
landism: it feels bikesheddy to ask "why 5?" but it also feels like just putting 5 here is making it really tempting to ask - maybe a "// arbitrary non-1 value chosen to allow callers to distinguish between tilt errors and tiltfile errors?"

internal/cli/tiltfile_result.go
```diff
@@ -0,0 +1,87 @@
...
+func newTiltfileResultDeps(tfl tiltfile.TiltfileLoader) tiltfileResultDeps {
+	return tiltfileResultDeps{
+		tfl: tfl,
+	}
+}
+
+func newTiltfileResultCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "tiltfile-result",
+		Short: "exec the tiltfile and print results as json",
```
landism: I'd probably also add a disclaimer here about the api changing

internal/cli/up.go
```diff
@@ -40,12 +40,12 @@ var webHost = DefaultWebHost
 var webDevPort = 0
 var noBrowser bool = false
 var logActionsFlag bool = false
+var fileName string = tiltfile.FileName
```
nicks: it would be better for command-specific flags to be on a command-specific struct, and not share each other's global state. it's pretty common to eventually want to move these commands to separate packages, cf kubectl

internal/cli/tiltfile_result.go
```diff
@@ -0,0 +1,88 @@
...
+		Run: tiltfileResultPrintJSON,
+	}
+
+	cmd.Flags().StringVar(&fileName, "file", tiltfile.FileName, "Path to Tiltfile")
+
+	return cmd
+}
+
+func tiltfileResultPrintJSON(cmd *cobra.Command, args []string) {
+	l := logger.NewLogger(logger.DebugLvl, os.Stdout)
```
nicks: can you elaborate a bit on why you didn't make this a normal tilt command that uses preCommand, like tilt up / tilt down? the way all these things interact is pretty subtle

internal/cli/tiltfile_result.go
```diff
@@ -0,0 +1,88 @@
...
+
+	tiltanalytics "github.com/windmilleng/tilt/internal/analytics"
+	"github.com/windmilleng/tilt/internal/tiltfile"
+	"github.com/windmilleng/tilt/pkg/logger"
+	"github.com/windmilleng/tilt/pkg/model"
+)
+
+const TiltfileErrExitCode = 5
+
+type tiltfileResultDeps struct {
```
nicks: it might be worth calling this cmdTiltfileResultDeps, for consistency with cmdUpDeps. it's a dependency of the command, not the result

internal/cli/tiltfile_result.go
```diff
@@ -0,0 +1,88 @@
...
+
+		os.Exit(TiltfileErrExitCode)
+	}
+
+	j, err := json.Marshal(tlr)
+	if err != nil {
+		failWithUnexpectedError(errors.Wrap(err, "marshaling JSON"))
+	}
+
+	err = dumpJSON(bytes.NewReader(j))
```
nicks: this....is probably not the function you want. 

this was written to take some JSON from the web API, and clean it up a bit

so what you're doing here is:
- encoding an object as json
- decoding it to a go object
- re-encoding it to json

you're probably should replace your json.Marshal calls with encodeJSON, then delete this code

internal/cli/tiltfile_result.go
nicks: for what it's worth, I don't think the code fulfills this contract. In particular, "Tiltfile loading returned an Error" doesn't mean "this is an expected error". It's just not something you can guarantee at this level of the API

landism: separately, I think this kind of comment is better to put in the code than on the PR

internal/cli/tiltfile_result.go
maiamcc: Oh I assumed `preCommand` has a bunch of `up`/`down`-specific stuff. Can do!

internal/cli/tiltfile_result.go
maiamcc: er that's fair, I didn't really mean expected vs unexpected, but I want the caller to be able to differentiate "something went wrong while trying to read the tiltfile" vs "got an error while executing the Tiltfile"

internal/cli/tiltfile_result.go
maiamcc: whoops yes!

internal/cli/tiltfile_result.go
maiamcc: tiltfile output goes to stdout and json || error goes to stderr, assuming caller will use the exit code to figure out what kind of error (if any) it is. good point about the json, i'll just print the error i guess

internal/cli/tiltfile_result.go
```diff
@@ -0,0 +1,75 @@
...
+	}
+
+	err = encodeJSON(tlr, os.Stderr)
+	if err != nil {
+		return wrapUnexpectedError(errors.Wrap(err, "encoding JSON"))
+	}
+	return nil
+}
+
+func wrapUnexpectedError(err error) error {
```
nicks: it seems weird that the error message is "evaluating Tiltfile", but it's only used in places where we're not evaluating the tiltfile?

internal/cli/tiltfile_result.go
```diff
@@ -0,0 +1,75 @@
...
+	tlr := deps.tfl.Load(ctx, c.fileName, model.NewUserConfigState(args))
+	if tlr.Error != nil {
+		// Some errors won't JSONify properly by default, so just print it
+		// to STDERR and use the exit code to indicate that it's an error
+		// from Tiltfile parsing.
+		fmt.Fprintln(os.Stderr, tlr.Error)
+		os.Exit(TiltfileErrExitCode)
+	}
+
+	err = encodeJSON(tlr, os.Stderr)
```
nicks: can you elaborate why we'd want this output to go to stderr? 

I expect most people would expect it to go to stdout, like it does for other unix command-line tools (and for our commands). That way you can pipe it to other tools, e.g.

```
tilt dump engine | jq ".TiltStartTime"
```

landism: +1

landism: similarly, to make this flow cleaner, I wonder if it'd make sense to, change tiltfile output:
1. don't print it by default
2. print it to stderr on failure
3. print it to stderr if some flag is set (though I couldn't come up with a name in 15 seconds, so who knows)

nicks: Ya, my first instinct is that any Tiltfile logs or warnings would go to stderr

internal/cli/tiltfile_result.go
maiamcc: mmm fair.

internal/cli/tiltfile_result.go
maiamcc: I can't remember from where but I had an idea of putting structured output into stderr? But yeah Matt's plan sgtm

internal/cli/tiltfile_result.go
```diff
@@ -0,0 +1,104 @@
...
+		Short: "Exec the Tiltfile and print results as JSON (note: the API is unstable and may change)",
+	}
+
+	cmd.Flags().StringVar(&c.fileName, "file", tiltfile.FileName, "Path to Tiltfile")
+
+	return cmd
+}
+
+func (c *tiltfileResultCmd) run(ctx context.Context, args []string) error {
+	// HACK(maia): we're overloading the -v|--verbose flags here, which isn't ideal,
```
nicks: What's the expected behavior here? I'm having trouble inferring what you're trying to accomplish.

TBH I expected this to be as simple as logger.WithLogger(ctx, logger.NewLogger(logLvl, os.Stderr)); I couldn't figure out what the extra options are for.

maiamcc: Oh, I'm just not printing Tiltfile logs except a. in verbose mode or b. on error. I don't feel strongly about this thi.

nicks: let's print the logs to stderr all the time. If someone complains that it should be an option, we can add the option then. simplicity is good!

maiamcc: ah whoops I missed this. how strongly do you feel about this? I think there's a strong case for "it's simpler for the user to have less output", and "only log stuff if requested or if something goes wrong" is a pretty common pattern, right?

mostly i'm not inclined to futz with this too much cuz it's alpha and i've already spent more time on it than i meant to, but 🤷‍♀

PR #2978, opened 2020-02-18
tiltfile: handle go vendoring

jazzdan:


Tiltfile
```diff
@@ -28,7 +28,10 @@ def go_lint(all_go_files):
   local_resource("go_lint", "make lint", deps=all_go_files)
 
 def get_all_ts_files(path):
-  return str(local('cd %s && find . -type f -name "*.ts*" | grep -v node_modules | grep -v __snapshots__' % path)).split("\n")
+  res = str(local('cd %s && find . -type f -name "*.ts*" | grep -v node_modules | grep -v __snapshots__' % path)).split("\n")
+  # TODO(dmiller): this thing always has a spare `""` hanging out at the end. Just `pop()` it off
+  res.pop()
```
landism: nit, since there's a TODO: probably a little cleaner to `.rstrip()` before the `.split`, which feels like it's common enough we should build it in

Tiltfile
```diff
@@ -40,10 +43,14 @@ def web_lint():
   ts_deps = get_all_ts_files("web")
   local_resource("web_lint", "cd web && yarn run check", deps=ts_deps, resource_deps=["yarn_install"])
 
+def go_vendor():
+  local_resource("go_vendor", "make vendor", deps=['go.sum', 'go.mod'])
```
landism: not really related to this PR, so feel free to merge anyway, but made me wonder - why filter out vendor in get_all_go_files? This seems handy:
1. I import something in a go file that's not in vendor
2. Tilt runs go build, which fails because it's not in go vendor, but maybe it automatically updates go.mod (not sure if it does this with -mod vendor)? or maybe I run go get to add it to go.mod
3. Tilt runs make vendor to update vendor
4. Tilt re-runs go build now that vendor has the library

jazzdan: Good point!

PR #2984, opened 2020-02-19
engine: otel span for cold start

jazzdan:


internal/engine/cold_start_tracker.go
```diff
@@ -0,0 +1,38 @@
+package engine
```
nicks: all new subscribers should go in a subpackage! might make sense to put this in the same package as the other trace stuff?

internal/engine/cold_start_tracker.go
```diff
@@ -0,0 +1,38 @@
...
+func (c *ColdStartTracker) OnChange(ctx context.Context, st store.RStore) {
+	if c.coldStartFinished {
+		return
+	}
+
+	state := st.RLockState()
+	defer st.RUnlockState()
+
+	if !state.InitialBuildsCompleted() && c.span == nil {
+		_, span := c.tracer.Start(ctx, "first_run")
```
nicks: fwiw, subscribers have an extra SetUp interface that might be helpful here

jazzdan:
ping @nicks @landism

landism:
It looks like you dropped "cold" from the file names but not the code?

jazzdan:
@landism whoops, thanks goland

internal/engine/telemetry/start_tracker.go
```diff
@@ -0,0 +1,38 @@
...
+
+	"go.opentelemetry.io/otel/api/trace"
+
+	"github.com/windmilleng/tilt/internal/store"
+)
+
+type StartTracker struct {
+	tracer            trace.Tracer
+	span              trace.Span
+	coldStartFinished bool
```
landism: one more!

PR #2987, opened 2020-02-20
web: add trace view

nicks:
Hello @hyu, @maiamcc,

Please review the following commits I made in branch nicks/buildview:

9c11fa394b63cf6c658cc41102c3642952d206d6 (2020-02-20 16:32:53 -0500)
web: add trace view

### DIFF
--- a/web/src/SecondaryNav.tsx
+++ b/web/src/SecondaryNav.tsx
@@ -6,8 +6,10 @@ import * as s from "./style-helpers"
 
 type NavProps = {
   logUrl: string
+  traceUrl: string | null
   alertsUrl: string
   facetsUrl: string | null
+  span: string
   resourceView: ResourceView
   numberOfAlerts: number
 }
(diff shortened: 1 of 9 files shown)


nicks:
here's what it looks like:

Screenshot from 2020-02-20 16-33-24

There's currently no way to get to this in the UI. You have to know the magic URL. 

It tries to visualize all the logs that were "caused" by a particular build. 

Future directions:
- A button to get to this UI
- Light affordances to express "historic" builds (with a clear start/end time) vs pending/ongoing builds
- breadcrumbs to navigate back to the full log, or to previous/future builds

nicks:
ping?

web/src/SecondaryNav.tsx
```diff
@@ -71,6 +73,36 @@ class SecondaryNav extends PureComponent<NavProps> {
     let logIsSelected = this.props.resourceView === ResourceView.Log
     let alertsIsSelected = this.props.resourceView === ResourceView.Alerts
     let facetsIsSelected = this.props.resourceView === ResourceView.Facets
+    let traceIsSelected = this.props.resourceView === ResourceView.Trace
+    let spanId = this.props.span
+
+    let traceItem = null
+    if (traceIsSelected && this.props.traceUrl) {
```
maiamcc: add a note here that this is to keep this tab hidden unless you know the url / `if traceIsSelected` should be removed when we make this discoverable, etc.

nicks: done

maiamcc:
Thanks for the link. FWIW that doc doesn't precisely answer my question as to the variable name "trace" in particular (the word "trace" doesn't actually appear in the doc?) but I don't feel strongly. just relevant info for you that that variable name isn't immediately legible, do with that what you will

PR #2990, opened 2020-02-21
hud: collapse uncategorized by default

landism:
### Problem

The "uncategorized" section in the HUD can be rather annoying:
1) Since there's generally no way to categorize its contents, it kinda makes you feel like Tilt just doesn't support what you're trying to do.
2) It ends up adding significantly to the amount of noise in the HUD, in a way that's not at all useful:
image
3) For new users, using a Tiltfile someone else wrote, it can be a source of confusion or look like it's misconfigured

IIRC, the reason it expands by default in the first place is it was part of the onboarding flow back in the olden days when Tilt created resources from docker_builds rather than from k8s workloads. The flow was that you'd load all your yaml, have a list of uncategorized k8s objects, and then as you added docker_builds, (most of) those objects would get moved into resources. It made sense to expand it and make it noisy, since (in our view at the time), it was a sign you hadn't yet finished onboarding / configuring all of your resources.

Since we now automatically create resources from k8s workloads, that flow doesn't make any sense at all, and I don't think there's a good reason to expand it by default.

### Solution

Don't expand it by default.


hyu:
Thanks, @landism! Your write-up is excellent, and I think this change is a definite improvement.

PR #2997, opened 2020-02-21
tiltfile: make read_file error on missing file if default unspecified

landism:
### Problem

The (`read_file docs`)[https://docs.tilt.dev/api.html#api.read_file] state:

> If the file_path does not exist and default is not None , default will be returned. In any other case, an error reading file_path will be a Tiltfile load error.

However, the actual implementation is that `default` has a default value of `""` rather than `None`, and unpacks the arg into a string, so it's not actually possible to get read_file to error on a missing file.

### Solution

change `read_file` to error if the file doesn't exist and default is unspecified

### Result

We can now get errors when `read_file` targets are missing!

This might break users that were relying on the existing behavior.
1. It's easy for them to get the existing behavior back by adding a `default=""` to their `read_file` call.
2. My gut is that the documented behavior is far more desirable and a better default.

### DIFF
--- a/internal/tiltfile/io/io.go
+++ b/internal/tiltfile/io/io.go
@@ -54,10 +54,21 @@ func (Extension) OnExec(t *starlark.Thread, tiltfilePath string) error {
 
 func readFile(thread *starlark.Thread, fn *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
 	var path starlark.Value
-	defaultReturn := ""
-	err := starkit.UnpackArgs(thread, fn.Name(), args, kwargs, "paths", &path, "default?", &defaultReturn)
+	// the type of default is Union[None, Str], so that we can distinguish unspecified from the empty string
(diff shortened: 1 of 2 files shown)


landism:
> It seems a liiitle weird that I as a user can set `default=None` and, when the file dne, I get an error instead of `None` -- maybe don't let the user set the default to None? But also eh, I don't really care, this is fine as is!

Yeah, that's not great. It's hard to simply disallow setting None intuitively, though. Given the API docs, the default value for the default value is None, so explicitly setting it to None should be a noop. Maybe "error on missing" should be a separate arg, except then it'd make everything clumsier because we'd want it to default to False if the default value is non-None, and True if it's None. Or, like, error_on_missing_if_default_none, to be precise, but that's just awful.

I don't have a good answer at the moment, and this at least makes it consistent with the docs, and makes it possible to have your Tiltfile fail if the file is missing, so going to merge this as an incremental improvement, if not ideal solution.

maiamcc:
👍

PR #2999, opened 2020-02-21
web: fix clicking on building resources

### LINKED ISSUE #2998: webui: unable to click on resources while they're building
repro:

try to click on a resource in the sidebar while tilt is building a docker image (you can add a `RUN sleep 20` to the dockerfile to make this easier)

expected:

it takes you to http://localhost:10350/r/myresource

observed:

it's not clickable, and the mouse cursor even remains an arrow instead of changing to the pointy finger indicating something is clickable

landism:
fixes #2998 

### Problem

Resources are not clickable in the webui while they're building.

This appears to be because their ::after element is stealing their clicks.

### Solution

Change the ::after element to not steal clicks. I don't know much about what is going on here or if there's a more elegant solution and just went with the first SO answer I found after diagnosing the problem.

### DIFF
--- a/web/src/Sidebar.tsx
+++ b/web/src/Sidebar.tsx
@@ -115,6 +115,7 @@ let SidebarItemStyle = styled.li`
   &.isBuilding::after {
     content: "";
     position: absolute;
+    pointer-events: none;
     width: 100%;
     top: 0;
     bottom: 0;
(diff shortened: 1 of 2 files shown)


tmc:
nit but having some regression test around this would be nice.

PR #3002, opened 2020-02-21
fix data race between hud and renderer

pleclech:
Hi,
Running with go --race, go complain there is a data race :

WARNING: DATA RACE
Write at 0x00c000694330 by goroutine 51:
  github.com/windmilleng/tilt/internal/hud.(*Renderer).SetUp()
     tilt/internal/hud/renderer.go:315 +0x3a7
  github.com/windmilleng/tilt/internal/hud.(*Hud).Run()
     tilt/internal/hud/hud.go:95 +0x1fb
  github.com/windmilleng/tilt/internal/cli.(*upCmd).run.func1()
     tilt/internal/cli/up.go:144 +0x10d
  golang.org/x/sync/errgroup.(*Group).Go.func1()
      go/pkg/mod/golang.org/x/sync@v0.0.0-20190911185100-cd5d95a43a6e/errgroup/errgroup.go:57 +0x78

Previous read at 0x00c000694330 by goroutine 56:
  github.com/windmilleng/tilt/internal/hud.(*Hud).refreshSelectedIndex()
     tilt/internal/hud/hud.go:338 +0x5d
  github.com/windmilleng/tilt/internal/hud.(*Hud).OnChange()
     tilt/internal/hud/hud.go:294 +0x3b3
  github.com/windmilleng/tilt/internal/store.(*subscriberEntry).notify()
     tilt/internal/store/subscriber.go:125 +0x151
  github.com/windmilleng/tilt/internal/store.(*subscriberList).NotifyAll.func1()
     tilt/internal/store/subscriber.go:105 +0x63
  github.com/windmilleng/tilt/internal/store.SafeGo.func1()
     tilt/internal/store/safe.go:28 +0x79

### DIFF
--- a/internal/hud/hud.go
+++ b/internal/hud/hud.go
@@ -92,9 +92,7 @@ func (h *Hud) Run(ctx context.Context, dispatch func(action store.Action), refre
 		h.mu.Unlock()
 	}()
 
-	h.mu.RLock()
 	screenEvents, err := h.r.SetUp()
-	h.mu.RUnlock()
 
 	if err != nil {
 		return errors.Wrap(err, "setting up screen")
--- a/internal/hud/hud.go
+++ b/internal/hud/hud.go
@@ -92,7 +92,10 @@ func (h *Hud) Run(ctx context.Context, dispatch func(action store.Action), refre
 		h.mu.Unlock()
 	}()
 
+	h.mu.RLock()
 	screenEvents, err := h.r.SetUp()
+	h.mu.RUnlock()
+
 	if err != nil {
 		return errors.Wrap(err, "setting up screen")
 	}
@@ -110,9 +113,8 @@ func (h *Hud) Run(ctx context.Context, dispatch func(action store.Action), refre
 			err := ctx.Err()
 			if err != context.Canceled {
 				return err
-			} else {
-				return nil
 			}
+			return nil
 		case e := <-screenEvents:
 			done := h.handleScreenEvent(ctx, dispatch, e)
 			if done {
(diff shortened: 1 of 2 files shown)


internal/hud/hud.go
```diff
@@ -92,7 +92,10 @@ func (h *Hud) Run(ctx context.Context, dispatch func(action store.Action), refre
 		h.mu.Unlock()
 	}()
 
+	h.mu.RLock()
```
nicks: Can you elaborate on what this lock is for? `h.r` doesn't change, so i don't think it needs a lock here, though it's probably harmless

pleclech: Hi @nicks,
I know that 'h.r' will normally not change after init, but as we read this in a go routine it's the only way i find to read safely this field (tho the read lock RLock) and keep the go data racer happy.

nicks: Ya, that makes sense, but I'm worried that this is papering over the real root cause of the bug.

When you repro this, are you running a particular test? Or running the `tilt up` tith the data race detection on?

pleclech: i run from tilt source : go run --race  ./cmd/tilt/main.go up --web-mode=prod --file=xyz/tiltdemo/Tiltfile 2>log.txt

internal/hud/renderer.go
```diff
@@ -319,6 +320,13 @@ func (r *Renderer) SetUp() (chan tcell.Event, error) {
 	return screenEvents, nil
 }
 
+func (r Renderer) RTY() rty.RTY {
```
nicks: ok, I see the problem. 

There's a bug in your RTY() function. This needs to be `(r *Renderer)` instead of `(r Renderer)`. Right now it's making a copy of Renderer when you call the function, which is not what you want.

once you fix that, you can remove the rendundant RLock() that we talked about above

pleclech: @nicks Oh yes sorry for that ! Thank you.

PR #3005, opened 2020-02-22
Add space between multiple PortForwardLinks

victorwuky:
Clubhouse story: https://app.clubhouse.io/windmill/story/4269/no-spacing-between-multiple-k8s-port-forwards

### Multiple
| Before / After | Screenshot |
| --- | --- |
| Before | multiple-before | 
| After | multiple-after |

### Single
| Before / After | Screenshot |
| --- | --- |
| Before | single-before | 
| After |  single-after |

### DIFF
--- a/web/src/ResourceInfo.tsx
+++ b/web/src/ResourceInfo.tsx
@@ -48,7 +48,10 @@ let PortForwardLabel = styled.span`
 
   ${s.mixinHideOnSmallScreen}
 `
-let PortForwardLink = styled.a``
+
+let PortForwardLink = styled.a`
+  margin-right: ${s.SizeUnit(0.25)};
+`
 
 let SnapshotButton = styled.button`
   border: 1px solid transparent;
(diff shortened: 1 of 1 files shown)


victorwuky:
@hyu : Does this work? It adds some space in the last portforwardlink at the right, not sure if that matters.

web/src/ResourceInfo.tsx
```diff
@@ -48,7 +48,10 @@ let PortForwardLabel = styled.span`
 
   ${s.mixinHideOnSmallScreen}
 `
-let PortForwardLink = styled.a``
+
+let PortForwardLink = styled.a`
```
hyu: This would add extra spacing to the right of every single Port Forward link, when what we really want is to add spacing between links when there is more than one. 

So, I would replace the line you added with this CSS adjacent sibling selector, so you end up with this:

```
let PortForwardLink = styled.a`
  & + &  {
    padding-left: ${s.SizeUnit(0.25)};
    border-left: 1px dotted ${s.Color.grayLight};
    margin-left: ${s.SizeUnit(0.25)};
  }
`
```

victorwuky: Cool! Didn't know you could put selectors right here. Thanks!

victorwuky: Looks great!

Screen Shot 2020-02-24 at 2 24 45 PM

PR #3006, opened 2020-02-22
Fix capitalization in terminal output

victorwuky:
Clubhouse story: https://app.clubhouse.io/windmill/story/4697/fix-capitalization-in-tui-output

### Before
before

### After
after

### DIFF
--- a/vendor/github.com/spf13/cobra/command.go
+++ b/vendor/github.com/spf13/cobra/command.go
@@ -967,7 +967,7 @@ func (c *Command) validateRequiredFlags() error {
 func (c *Command) InitDefaultHelpFlag() {
 	c.mergePersistentFlags()
 	if c.Flags().Lookup("help") == nil {
-		usage := "help for "
+		usage := "Help for "
 		if c.Name() == "" {
 			usage += "this command"
 		} else {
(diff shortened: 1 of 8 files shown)


victorwuky:
@maiamcc : Is this good? (I'll be sure to squash the commits before merging if everything else looks good.)

vendor/github.com/spf13/cobra/command.go
```diff
@@ -967,7 +967,7 @@ func (c *Command) validateRequiredFlags() error {
 func (c *Command) InitDefaultHelpFlag() {
 	c.mergePersistentFlags()
 	if c.Flags().Lookup("help") == nil {
-		usage := "help for "
+		usage := "Help for "
```
maiamcc: Don't edit files in `vendor/`, they're external packages we depend on , and edits here will cause weirdness if/when we pull down the latest version of this package.

vendor/github.com/windmilleng/wmclient/pkg/analytics/cli.go
```diff
@@ -51,7 +51,7 @@ const choiceFile = "analytics/user/choice.txt"
 func NewCommand() *cobra.Command {
 	analytics := &cobra.Command{
 		Use:   "analytics",
-		Short: "info and status about windmill analytics",
+		Short: "Info and status about windmill analytics",
 		RunE:  analyticsStatus,
 	}
```
maiamcc: See above re: `vendor/`

vendor/github.com/spf13/cobra/command.go
victorwuky: Fixed thanks!

vendor/github.com/windmilleng/wmclient/pkg/analytics/cli.go
victorwuky: Fixed thanks

PR #3013, opened 2020-02-25
scripts: don't sudo if $HOME/bin is in $PATH

### LINKED ISSUE #3004: sudo required for install script
Instead of `sudo` you chould create/use `~/.local/bin` and move the tilt exec there.  This is a standard pattern with some other tools.

jazzdan:
Fixes #3004

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -13,12 +13,11 @@ BREW=$(command -v brew)
 set -e
 
 function copy_binary() {
-  if [[ ":$PATH:" == *":$HOME/bin:"* ]]; then
-      mv tilt "$HOME/bin/tilt"
+  if [[ ":$PATH:" == *":$HOME/.local/bin:"* ]]; then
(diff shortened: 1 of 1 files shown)


scripts/install.sh
```diff
@@ -12,11 +12,22 @@ BREW=$(which brew)
 
 set -e
 
+function copy_binary() {
+  if [[ ":$PATH:" == *":$HOME/bin:"* ]]; then
```
nicks: i think i'd slightly prefer .local/bin because there's a formal spec around it?
https://www.freedesktop.org/software/systemd/man/file-hierarchy.html

scripts/install.sh
```diff
@@ -12,11 +12,22 @@ BREW=$(which brew)
...
+      mv tilt "$HOME/bin/tilt"
+  else
+      echo "Installing Tilt to /usr/local/bin which is write protected"
```
nicks: This feels a bit verbose

PR #3015, opened 2020-02-25
store: we should still live update unready containers

### LINKED ISSUE #3000: live_update erroneously depends on k8s readiness check
### Repro

1. Run something in a pod and make its readiness check fail.
2. Try to live update it

### Expected

Tilt performs the live update

### Observed

Tilt says it lost the pod so it can't live update, so it builds a new image and redeploys, killing the old pod.

### Analysis

When Tilt looks for a container to live update, it [ignores containers that are not ready](https://github.com/windmilleng/tilt/blob/c64bd7b64395587e8c7034e4f40e6d7a1a5aecdc/internal/store/build_result.go#L445).

However, if a container's readiness probe fails, it seems likely that it's still fine for live update. A readiness probe failing just means k8s shouldn't route traffic, not that the pod is irredeemable.

To more explicitly spell out a bad case here:
1. User performs a live update, which recompiles and restarts the process
2. While the process is starting up, the readiness probe is failing for 5 seconds
3. User performs another live update while the readiness probe is still failing
4. Since the container is not ready, tilt says there's no container to live update, and does a full image build + redeploy

I haven't looked into whether this is as simple as removing the c.Ready check for the purposes of live update, or if that leaves us open to syncing before the container is even syncable.

jazzdan:
To more explicitly spell out the bad case here:

1. User performs a live update, which recompiles and restarts the process
2. While the process is starting up, the readiness probe is failing for 5 seconds
3. User performs another live update while the readiness probe is still failing
4. Since the container is not ready, tilt says there's no container to live update, and does a full image build + redeploy

It seems like in practice this doesn't cause an issue with pods getting updated before their ready, in my manual testing.

Fixes #3000

### DIFF
--- a/internal/store/build_result.go
+++ b/internal/store/build_result.go
@@ -444,7 +444,7 @@ func RunningContainersForTargetForOnePod(iTarget model.ImageTarget, runtimeState
 		}
 		if c.ID == "" || c.Name == "" || !c.Running {
 			// If we're missing any relevant info for this container, OR if the
-			// container isn't ready, we can't update it in place.
+			// container isn't running, we can't update it in place.
 			// (Since we'll need to fully rebuild this image, we shouldn't bother
 			// in-place updating ANY containers on this pod -- they'll all
 			// be recreated when we image build. So don't return ANY ContainerInfos.)
--- a/internal/store/build_result.go
+++ b/internal/store/build_result.go
@@ -442,7 +442,7 @@ func RunningContainersForTargetForOnePod(iTarget model.ImageTarget, runtimeState
 		if c.ImageRef == nil || iTarget.Refs.ClusterRef().Name() != c.ImageRef.Name() {
 			continue
 		}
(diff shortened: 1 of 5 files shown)


internal/store/build_result.go
```diff
@@ -442,7 +442,7 @@ func RunningContainersForTargetForOnePod(iTarget model.ImageTarget, runtimeState
 		if c.ImageRef == nil || iTarget.Refs.ClusterRef().Name() != c.ImageRef.Name() {
 			continue
 		}
-		if c.ID == "" || c.Name == "" || !c.Ready {
+		if c.ID == "" || c.Name == "" {
```
nicks: hmmm... I think we want to check if the container is Running, not Ready?
https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#container-states

The reason this check was added is because there are a variety of cases where the container won't be live-updatable (e.g., the container crashed, or the image isn't pulling correctly)

PR #3018, opened 2020-02-26
engine: add which targets were updated to update span

jazzdan:


internal/engine/build_and_deployer.go
```diff
@@ -63,6 +63,10 @@ func (composite *CompositeBuildAndDeployer) BuildAndDeploy(ctx context.Context,
 	defer span.End()
 	var lastErr, lastUnexpectedErr error
 
+	for _, s := range specs {
+		span.SetAttributes(core.KeyValue{Key: core.Key(fmt.Sprintf("targetNames.%s", s.ID())), Value: core.Bool(true)})
```
nicks: reading the open telemetry spec, it seems like these attribute keys are supposed to be pre-determined constants
https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/data-semantic-conventions.md#span-conventions
right?

jazzdan: @nicks what gives you that idea, can you quote a part of that doc?

nicks: sure, mostly inferring this from the existing examples and conventions, like
https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/data-span-general.md
and
https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/data-http.md

I would be totally fine with this if we could find some examples with a convention for this. We could also ask in one of the opentelemetry community spaces

jazzdan: I'll point out that we're [already using dynamic keys](https://github.com/windmilleng/tilt/pull/3018/files#diff-7396551d13039e7bceb0b3a6cd301c13L75) for other attributes, but yeah I'll ask.

nicks: i think that case is a little different since the set of all possible build types is hard-coded, whereas this ID contains an image reference and is essentially unbounded

I went down this rabbit-hole because I was trying to figure out what the grammar is for attribute keys. all the existing examples seem to use "a.b.c". I'm a little nervous of whether downstream telemetry systems will blow up if you have random slashes, colons, etc in the names.

jazzdan: Ahh I see what you're saying.

I don't think we can predict/know all the different systems people will want to use this data with. If they need to clean up the data to import in to their system they can do that in their telemetry script.

nicks: right! the whole reason we want to follow the specification (both semantically and following their conventions) is to avoid this work, which is why I'm looking for more support for this approach in the spec/public docs

jazzdan: @dbentley do you have any thoughts?

dbentley: Why try to put them in the attribute key instead of the attribute value? (I fear I may be several levels below where you're at)

jazzdan: Because any given update could involve multiple resources. Though I guess it could be an array of values, is that what you're getting at?

dbentley: Yeah, or why Attributes instead of Events?

jazzdan: Hmm not sure what events are, seems like they are... specialized sub spans? Attributes with timestamps?

I'll just make this an array.

jazzdan: well... looks like arrays as attribute values were only [recently added](https://github.com/open-telemetry/opentelemetry-specification/blob/a600cd0ff6606eafa9588d89cba6f4e581ba53a9/CHANGELOG.md#v030-02-21-2020) to the spec and they haven't made their way to the go implementation yet. Sigh.

jazzdan: I'm just going to make this a static key and have the value be a comma separated string and call it a day.

PR #3026, opened 2020-02-27
k8swatch: fix a race condition in pod change events

nicks:
Hello @landism, @maiamcc,

Please review the following commits I made in branch nicks/ch4899/update:

75cbac9610e01c610b205d1b2c03faa296ef99dd (2020-02-27 17:41:14 -0500)
k8swatch: fix a race condition in pod change events
Sometimes pod events would show up in the wrong order due to goroutine scheduling

### DIFF
--- a/internal/engine/k8swatch/pod_watch.go
+++ b/internal/engine/k8swatch/pod_watch.go
@@ -179,6 +179,14 @@ func (w *PodWatcher) setupNewUIDs(ctx context.Context, st store.RStore, newUIDs 
 	}
 }
 
+func (w *PodWatcher) upsertPod(pod *v1.Pod) {
+	w.mu.Lock()
+	defer w.mu.Unlock()
+
+	uid := pod.UID
+	w.knownPods[uid] = pod
+}
+
 // Check to see if this pod corresponds to any of our manifests.
 //
 // Currently, we do this by comparing the pod UID and its owner UIDs against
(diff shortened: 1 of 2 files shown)


nicks:
i think this is what's causing some of the weird integration test failures from time to time

landism:
IIUC, this fixes the potential for dispatching pod events out of order, but it replaces out-of-order pod events with duplicates of the latest pod event? This is mostly fine for the purposes of ensuring that the engine state has the latest pod state, but seems bad when we care about intermediate state, e.g., maybe we skip logging one of the stages of pod rollout.

Would it be unreasonable to fix the race condition by dropping the `go` when calling `dispatchPodChange`? (or, if we don't want to block goroutines, having a chan of pod events that we dispatch in order)

nicks:
Ya, this is the same API guarantee that the informers make:
https://godoc.org/k8s.io/client-go/tools/cache#ResourceEventHandler

> OnUpdate is called when an object is modified. Note that oldObj is the
    last known state of the object-- it is possible that several changes
    were combined together, so you can't use this to see every single
    change. OnUpdate is also called when a re-list happens, and it will
    get called even if nothing changed. This is useful for periodically
    evaluating or syncing something.

in other words, the underlying informer APIs also dispatch duplicate events, and also expect that intermediate states will get collapsed together. So I don't think this is something we can prevent.

PR #3028, opened 2020-02-27
k8s: add metrics for auto-inferring registry

maiamcc:
curious how much use our node-annotation-with-registry-info is getting used.


nicks:
hmmm...we already do this? https://github.com/windmilleng/tilt/pull/2945

maiamcc:
oh whoops I missed that when it went by.

PR #3032, opened 2020-02-28
k8s: move all minikube calls out of wire injection. Fixes https://github.com/windmilleng/tilt/issues/3021

nicks:
Hello @landism, @maiamcc,

Please review the following commits I made in branch nicks/ch4928/minikube:

745a08f35ee6178b049675a0d012e772bcc07751 (2020-02-28 17:00:27 -0500)
k8s: move all minikube calls out of wire injection. Fixes https://github.com/windmilleng/tilt/issues/3021

### DIFF
--- a/internal/docker/env.go
+++ b/internal/docker/env.go
@@ -23,6 +23,10 @@ type Env struct {
 	// Minikube's docker client has a bug where it can't use buildkit. See:
 	// https://github.com/kubernetes/minikube/issues/4143
 	IsMinikube bool
+
+	// If the env failed to load for some reason, propagate that error
+	// so that we can report it when the user tries to do a docker_build.
+	Error error
 }
 
 // Serializes this back to environment variables for os.Environ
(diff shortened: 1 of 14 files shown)


internal/k8s/ip.go
```diff
@@ -24,11 +36,24 @@ func DetectNodeIP(ctx context.Context, env Env) (NodeIP, error) {
...
 		}
-		return "", errors.Wrap(err, "Could not read node IP from minikube")
+		return ""
 	}
 
-	return NodeIP(strings.TrimSpace(string(out))), nil
+	return NodeIP(strings.TrimSpace(string(out)))
+}
+
+func (a *nodeIPAsync) NodeIP(ctx context.Context) NodeIP {
```
maiamcc: ooh nice!

internal/docker/env.go
```diff
@@ -103,9 +107,9 @@ func ProvideClusterEnv(ctx context.Context, env k8s.Env, runtime container.Runti
 
 	resultEnv, err := overlayOSEnvVars(Env(result))
```
maiamcc: nbd but for slightly cleaner code, could have `overlayOSEnvVars` return an `Env` containing the error, if any (rather than returning both an Env and an error and creating the error-env here if err is non-nil)

nicks: ooh that's a good idea, and makes things simpler

PR #3040, opened 2020-03-02
rty: rendering errors shouldn't totally break Tilt. Fixes https://github.com/windmilleng/tilt/issues/3039

nicks:
Hello @landism, @maiamcc,

Please review the following commits I made in branch nicks/ch4973/crash:

f746c53c911814e2ab2f5a6da35d4dcf123079f7 (2020-03-02 16:49:56 -0500)
rty: rendering errors shouldn't totally break Tilt. Fixes https://github.com/windmilleng/tilt/issues/3039
Given that Tilt has a browser UI now, TUI rendering errors shouldn't be world-stopping.
For now, let's just throw them out silently and see what happens.
The test frameworks still check for errors.

### DIFF
--- a/internal/hud/hud.go
+++ b/internal/hud/hud.go
@@ -66,13 +66,13 @@ func NewDefaultHeadsUpDisplay(renderer *Renderer, webURL model.WebURL, analytics
 	}, nil
 }
 
-func (h *Hud) SetNarrationMessage(ctx context.Context, msg string) error {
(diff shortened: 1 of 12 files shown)


nicks:
Usually this happens when we've calculated a box wrong. So some text will just get cut off a bit.

If we had infinite eng time, we'd fix the box calculation, but the math can be fiddly. 

Ya, this approach is still very naive, but I think will end up being a much better user experience.

PR #3041, opened 2020-03-02
engine: print trigger reasons

nicks:
Hello @maiamcc, @hyu,

Please review the following commits I made in branch nicks/ch4965/reasons:

7deafb7d3913b1af33e4cb1db55bdf8982257ee6 (2020-03-02 18:14:00 -0500)
engine: print trigger reasons

### DIFF
--- a/pkg/model/build_reason.go
+++ b/pkg/model/build_reason.go
@@ -14,6 +14,10 @@ const (
 	BuildReasonFlagCrash
 
 	BuildReasonFlagInit
+
+	BuildReasonFlagTriggerWeb
+	BuildReasonFlagTriggerCLI
+	BuildReasonFlagTriggerUnknown
 )
 
 func (r BuildReason) With(flag BuildReason) BuildReason {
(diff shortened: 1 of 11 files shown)


nicks:
https://cloud.tilt.dev/snapshot/AdDN7OULEtY29dD7bSM=

Not totally sure what the right words should be here, trying to make it brief without being cryptic

pkg/model/build_reason.go
```diff
@@ -29,17 +33,23 @@ func (r BuildReason) IsCrashOnly() bool {
...
-	BuildReasonFlagChangedFiles: "Changed Files",
-	BuildReasonFlagConfig:       "Config Changed",
-	BuildReasonFlagCrash:        "Pod Crashed, Lost live_update Changes",
-	BuildReasonFlagInit:         "Initial Build",
+	BuildReasonFlagChangedFiles:   "Changed Files",
+	BuildReasonFlagConfig:         "Config Changed",
+	BuildReasonFlagCrash:          "Pod Crashed, Lost live_update Changes",
+	BuildReasonFlagInit:           "Initial Build",
+	BuildReasonFlagTriggerWeb:     "Button Pressed",
+	BuildReasonFlagTriggerCLI:     "Commandline Trigger",
```
maiamcc: maybe "CLI trigger"? (at minimum "command-line" should be hyphenated)

maiamcc: but also for this and "button pressed" it's really that we _forced_ an update (if it's releasing a pending update on a manual resource, that update shows up as "files changed"). Not sure I have good words here, but "forced update via web" / "forced update via CLI"?

nicks: changed it to CLI Trigger

It feels important that it leads with a noun, rather than a subjectless verb? "Forced Update" puts the emphasis on the wrong place, the part the user doesn't really care about

maiamcc: sure, I buy that.

nicks:
I left a TODO for the other thing (NeedsForceUpdate), I agree that it would be good to fix but don't think it belongs in this PR

PR #3043, opened 2020-03-03
engine: replace ForcedUpdate with IsOverrideTrigger

nicks:
Hello @maiamcc,

Please review the following commits I made in branch nicks/forceupdate:

1bda6e4628227671d6289b5a87ed34e1fc494f60 (2020-03-03 16:47:11 -0500)
engine: replace ForcedUpdate with IsOverrideTrigger

### DIFF
--- a/internal/engine/buildcontroller.go
+++ b/internal/engine/buildcontroller.go
@@ -63,7 +63,7 @@ func (c *BuildController) needsBuild(ctx context.Context, st store.RStore) (buil
 
 	buildReason := mt.NextBuildReason()
 	targets := buildTargets(manifest)
-	buildStateSet := buildStateSet(ctx, manifest, targets, ms)
+	buildStateSet := buildStateSet(ctx, manifest, targets, ms, buildReason)
 
 	return buildEntry{
 		name:          manifest.Name,
@@ -182,10 +182,9 @@ func buildTargets(manifest model.Manifest) []model.TargetSpec {
 }
 
 // Extract a set of build states from a manifest for BuildAndDeploy.
-func buildStateSet(ctx context.Context, manifest model.Manifest, specs []model.TargetSpec, ms *store.ManifestState) store.BuildStateSet {
+func buildStateSet(ctx context.Context, manifest model.Manifest, specs []model.TargetSpec,
+	ms *store.ManifestState, reason model.BuildReason) store.BuildStateSet {
 	result := store.BuildStateSet{}
-
-	anyFilesChangedSinceLastBuild := false
 
 	for _, spec := range specs {
 		id := spec.ID()
@@ -195,11 +194,8 @@ func buildStateSet(ctx context.Context, manifest model.Manifest, specs []model.T
 
 		status := ms.BuildStatus(id)
 		var filesChanged []string
-		for file, ts := range status.PendingFileChanges {
+		for file := range status.PendingFileChanges {
 			filesChanged = append(filesChanged, file)
-			if ms.LastBuild().Empty() || ts.After(ms.LastBuild().StartTime) {
-				anyFilesChangedSinceLastBuild = true
-			}
 		}
 		sort.Strings(filesChanged)
 
@@ -234,16 +230,13 @@ func buildStateSet(ctx context.Context, manifest model.Manifest, specs []model.T
 		result[id] = buildState
 	}
 
-	// If there are no files changed across the entire state set, then this is a force update.
-	// We want to do an image build of each image.
-	// TODO(maia): I think that instead of storing this on every build state, we can can figure
-	//  out that it's a force update when creating the BuildEntry (in `needsBuild`), store that
-	//  as the BuildReason, and pass the whole BuildEntry to the builder (so the builder can
-	//  know whether to skip in-place builds)
-	// if we're on a crash rebuild, then there won't have been any files changed for that reason
-	if !ms.NeedsRebuildFromCrash && !anyFilesChangedSinceLastBuild {
+	isLiveUpdateEligibleTrigger := reason.HasTrigger() &&
+		reason.Has(model.BuildReasonFlagChangedFiles) &&
+		manifest.TriggerMode != model.TriggerModeAuto
+	isImageBuildTrigger := reason.HasTrigger() && !isLiveUpdateEligibleTrigger
+	if isImageBuildTrigger {
 		for k, v := range result {
-			result[k] = v.WithNeedsForceUpdate(true)
+			result[k] = v.WithImageBuildTriggered(true)
 		}
 	}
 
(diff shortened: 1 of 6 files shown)


nicks:
This is a follow-up from the other PR; I tried to better express what this field is for and how it interacts with the different trigger modes

internal/engine/buildcontroller.go
```diff
@@ -234,16 +229,11 @@ func buildStateSet(ctx context.Context, manifest model.Manifest, specs []model.T
...
-	// We want to do an image build of each image.
-	// TODO(maia): I think that instead of storing this on every build state, we can can figure
-	//  out that it's a force update when creating the BuildEntry (in `needsBuild`), store that
-	//  as the BuildReason, and pass the whole BuildEntry to the builder (so the builder can
-	//  know whether to skip in-place builds)
-	// if we're on a crash rebuild, then there won't have been any files changed for that reason
-	if !ms.NeedsRebuildFromCrash && !anyFilesChangedSinceLastBuild {
+	// Check to see if this is a resource in auto mode with an overriding trigger.
+	// See comments on IsOverrideTrigger for more details on the intended flow.
+	if manifest.TriggerMode == model.TriggerModeAuto && ms.TriggerReason != 0 {
```
maiamcc: a manual resource may be forced too -- you'd want to check that it was triggered and had no pending file changes.

also fwiw, `if ms.TriggerReason != 0` seems (at least to me) a confusing way to check if the TriggerReason was unset. consider hiding this logic away in a `WasTriggered` (or something) helper func?

internal/engine/buildcontroller.go
```diff
@@ -234,16 +229,11 @@ func buildStateSet(ctx context.Context, manifest model.Manifest, specs []model.T
...
 		for k, v := range result {
-			result[k] = v.WithNeedsForceUpdate(true)
+			result[k] = v.WithIsOverrideTrigger(true)
```
maiamcc: it feels a little weird with the model to me to store this information on each individual target's state, doesn't it apply to the BuildStateSet as a whole? (I'm not sure it matters enough to change tho--but fwiw storing this info per-target was a DB hack that never got cleaned up)

internal/store/build_result.go
```diff
@@ -270,9 +270,18 @@ type BuildState struct {
...
-	// Whether there was a manual trigger
-	// TODO(nick): Replace this with a BuildReason
-	NeedsForceUpdate bool
+	// There are two kinds of triggers:
+	//
+	// 1) If a resource is in trigger_mode=TRIGGER_MODE_MANUAL, then the resource doesn't
+	//    build until there's an explicit trigger. Once we have an explicit trigger,
+	//    we can do a normal image build or normal live_update.
+	//
+	// 2) If a resource is in trigger_mode=TRIGGER_MODE_AUTO, then the resource auto-builds.
```
maiamcc: not quite--as stated above, you can still force a full build on a manual resource. This maybe belies the naming change behind this PR, let's talk in the high-level comments.

nicks:
I'm glad we broke this into a separate PR, since it sounds like it's a pretty controversial change! I'm a little bit confused about whether you're objecting to the naming or the behavior tho

nicks:
It might be better if I just withdraw this PR, and you can document/iterate on the behavior you'd like here, since I did this at your request, and it seems like you're pretty unhappy with the outcome

maiamcc:
both! The thing I was getting at in my comment on 

behavior: seems like this is a regression that will prevent us from forcing a full rebuild on a manual resource.
naming: not sure I understand the rationale behind the name change, and not sure how i'd message it to a user.

the thing I was getting at in my request is just that checking the triggerReason (now that we have it) seems more robust/accurate than checking `if !ms.NeedsRebuildFromCrash && !anyFilesChangedSinceLastBuild`. (Secondarily, it seems weird that we store this info on each individual build state, but thinking about it more, idk if it's worth it.)

sure, I'm happy to own making these changes.

nicks:
OK, I:
- added some tests for the behavior distinction we talked about
- changed the field to IsImageBuildTrigger, which seems like a better expression of what this does

Re: storing this info on each individual build state - that doesn't seem that weird to me. I feel like other graph-based build systems propagate this kind of state to all the nodes.

internal/engine/buildcontroller_test.go
```diff
@@ -554,35 +566,6 @@ func TestBuildControllerManualTriggerWithFileChanges(t *testing.T) {
 	}
 }
 
-func TestBuildControllerManualTriggerWithoutFileChangesForceUpdates(t *testing.T) {
```
nicks: i basically just merged this test into the table above

internal/store/build_result.go
```diff
@@ -270,9 +270,19 @@ type BuildState struct {
...
+	// There are three kinds of triggers:
+	//
+	// 1) If a resource is in trigger_mode=TRIGGER_MODE_AUTO, then the resource auto-builds.
+	//    Pressing the trigger will always do a full image build.
+	//
+	// 2) If a resource is in trigger_mode=TRIGGER_MODE_MANUAL and there are no pending changes,
+	//    then pressing the trigger will do a full image build.
+	//
+	// 3) If a resource is in trigger_mode=TRIGGER_MODE_MANUAL, and there are pending changes,
+	//    then pressing the trigger will do a live_update
```
maiamcc:
```suggestion
	//    then pressing the trigger will do a live_update (if one is configured; otherwise, will do an image build as normal)
```

internal/store/build_result.go
```diff
@@ -270,9 +270,19 @@ type BuildState struct {
...
+	//
+	// We want to capture only case (1) and (2).
```
maiamcc:
```suggestion
	// This field indicates case 1 || case 2 -- i.e. that we should skip live_update, and force an image build (even if there are no changed files)
```
or something (w/o explanation of what this bool causes down the road, the explanation is less useful imo)

internal/store/build_result.go
```diff
@@ -270,9 +270,19 @@ type BuildState struct {
...
+	IsImageBuildTrigger bool
```
maiamcc: nit: this reads weird, like the build state itself is a trigger? Consider `ImageBuildTriggered` or something like that.

internal/engine/buildcontroller_test.go
```diff
@@ -508,13 +508,18 @@ func TestBuildControllerManualTriggerBuildReasonInit(t *testing.T) {
...
 }
 
-func TestBuildControllerManualTriggerWithFileChanges(t *testing.T) {
+func TestBuildControllerImageBuildTrigger(t *testing.T) {
 	for _, tc := range []struct {
-		name        string
-		triggerMode model.TriggerMode
+		name               string
+		triggerMode        model.TriggerMode
+		filesChanged       []string
```
maiamcc: nit: i see no reason to specify which files if it's the same set of files every time, you might just say `filesChanged: true` 🤷‍♀  no big deal tho

nicks: done

internal/store/build_result.go
nicks: done

internal/store/build_result.go
nicks: done

PR #3044, opened 2020-03-03
k8s: report runtime differently on read failure

nicks:
Hello @landism, @maiamcc,

Please review the following commits I made in branch nicks/crio:

0d5cf9e5245bb4f8af1340ff47ee90d5f4daa419 (2020-03-03 17:55:16 -0500)
k8s: report runtime differently on read failure

### DIFF
--- a/internal/container/runtime.go
+++ b/internal/container/runtime.go
@@ -7,10 +7,11 @@ import "strings"
 type Runtime string
 
 const (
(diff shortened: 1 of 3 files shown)


nicks:
somebody asked during planning why we have so many unknown container runtimes; this tries to distinguish read failures from true unknowns

PR #3050, opened 2020-03-06
engine: if we fail to inject the synclet into a resource, try to recover gracefully

nicks:
Hello @landism, @maiamcc,

Please review the following commits I made in branch nicks/ch5085/inject:

3d58c41342ad79c9dff95b7103e68b2e6c62c283 (2020-03-05 21:22:30 -0500)
engine: if we fail to inject the synclet into a resource, try to recover gracefully
Fixes https://github.com/windmilleng/tilt/issues/2948

### DIFF
--- a/internal/engine/image_build_and_deployer.go
+++ b/internal/engine/image_build_and_deployer.go
@@ -229,7 +229,7 @@ func (ibd *ImageBuildAndDeployer) deploy(ctx context.Context, st store.RStore, p
 
 	ps.StartBuildStep(ctx, "Injecting images into Kubernetes YAML")
 
-	newK8sEntities, err := ibd.createEntitiesToDeploy(ctx, iTargetMap, kTarget, results, needsSynclet)
+	deploySet, err := ibd.createEntitiesToDeploy(ctx, iTargetMap, kTarget, results, needsSynclet)
 	if err != nil {
 		return nil, err
 	}
@@ -242,12 +242,11 @@ func (ibd *ImageBuildAndDeployer) deploy(ctx context.Context, st store.RStore, p
 		l.Infof("→ %s", displayName)
 	}
 
-	deployed, err := ibd.k8sClient.Upsert(ctx, newK8sEntities)
+	deployed, err := ibd.k8sClient.Upsert(ctx, deploySet.entities)
 	if err != nil {
 		return nil, err
 	}
 
-	// TODO(nick): Do something with this result
 	uids := []types.UID{}
 	podTemplateSpecHashes := []k8s.PodTemplateSpecHash{}
 	for _, entity := range deployed {
@@ -262,7 +261,7 @@ func (ibd *ImageBuildAndDeployer) deploy(ctx context.Context, st store.RStore, p
 		}
 		podTemplateSpecHashes = append(podTemplateSpecHashes, hs...)
 	}
-	results[kTarget.ID()] = store.NewK8sDeployResult(kTarget.ID(), uids, podTemplateSpecHashes, deployed)
+	results[kTarget.ID()] = store.NewK8sDeployResult(kTarget.ID(), uids, podTemplateSpecHashes, deployed, deploySet.HasEligibleSynclet())
 
 	return results, nil
 }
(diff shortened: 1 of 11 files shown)


internal/engine/image_build_and_deployer.go
```diff
@@ -346,10 +355,12 @@ func (ibd *ImageBuildAndDeployer) createEntitiesToDeploy(ctx context.Context,
 					var sidecarInjected bool
 					e, sidecarInjected, err = sidecar.InjectSyncletSidecar(e, injectedRefSelector, ibd.syncletContainer)
 					if err != nil {
-						return nil, err
+						return result, err
```
maiamcc: to be clear, we're gracefully handling the "we tried to inject the synclet but didn't" case, but not the "something went WRONG while injecting the synclet" case, right?

internal/engine/image_build_and_deployer.go
```diff
@@ -346,10 +355,12 @@ func (ibd *ImageBuildAndDeployer) createEntitiesToDeploy(ctx context.Context,
...
 					}
 					if !sidecarInjected {
-						return nil, fmt.Errorf("Could not inject synclet: %v", e)
+						result.syncletInjectionError = true
+					} else {
+						result.syncletInjected = true
 					}
 					injectedSynclet = true
```
maiamcc: seems like `sidecarInjected` and `injectedSynclet` variables are redundant?

internal/engine/image_build_and_deployer.go
```diff
@@ -346,10 +355,12 @@ func (ibd *ImageBuildAndDeployer) createEntitiesToDeploy(ctx context.Context,
 					var sidecarInjected bool
 					e, sidecarInjected, err = sidecar.InjectSyncletSidecar(e, injectedRefSelector, ibd.syncletContainer)
 					if err != nil {
-						return nil, err
+						return result, err
 					}
 					if !sidecarInjected {
-						return nil, fmt.Errorf("Could not inject synclet: %v", e)
+						result.syncletInjectionError = true
```
maiamcc: waiiiitasec, this name is confusing: this is the case where it didn't throw an ERROR, we just failed to inject it when we expected to. can you rename to clarify? maybe `failedToInject`?

internal/engine/image_build_and_deployer.go
```diff
@@ -273,16 +272,26 @@ func (ibd *ImageBuildAndDeployer) indentLogger(ctx context.Context) context.Cont
...
-	newK8sEntities := []k8s.K8sEntity{}
+	results store.BuildResultSet, needsSynclet bool) (k8sDeploySet, error) {
+	result := k8sDeploySet{}
 
 	// TODO(nick): The parsed YAML should probably be a part of the model?
 	// It doesn't make much sense to re-parse it and inject labels on every deploy.
 	entities, err := k8s.ParseYAMLFromString(k8sTarget.YAML)
 	if err != nil {
-		return nil, err
+		return result, err
```
maiamcc: what's the point of returning the result along with these errors?

internal/engine/live_update_build_and_deployer.go
```diff
@@ -264,7 +264,7 @@ func (lubad *LiveUpdateBuildAndDeployer) containerUpdaterForSpecs(specs []model.
 		return lubad.ecu
 	}
 
-	if shouldUseSynclet(lubad.updMode, lubad.env, lubad.runtime) {
+	if lubad.shouldUseSynclet(specs, stateSet) {
```
maiamcc: (ugh i can't comment on the line of code I actually care about but...)

re: ln 259, we might want to throw an error if updMode == Synclet but none of the k8s targets have an eligible synclet. Seems like we'd rather raise here where we can control the messaging better then just let something more obscure go wrong further down the pipe.

nicks: i think that's a good idea, but I also think that messaging that effectively would end up being extremely complex - there's no single codepath that has access to all the information you need

maiamcc: i think it's baaaasically right here. if you move the `containerUpdaterForSpecs` call down past where we know we have live update targets, then INSIDE the `containerUpdaterForSpecs` call you know:
1. you have live update targets
2. user explicitly specified updateMode=synclet
3. you tried and failed to inject the synclet into this k8s target

That's the info you need, right?

internal/engine/image_build_and_deployer.go
nicks: fixed! now it returns nil

internal/engine/image_build_and_deployer.go
nicks: ya, i renamed the fields a bit to be more specific. We're trying to handle the case where:
1) We injected an image successfully, but
2) We weren't able to find a place to inject the synclet sidecar

internal/engine/image_build_and_deployer.go
nicks: yep, it's now called injectedImageWithoutSynclet

internal/engine/image_build_and_deployer.go
nicks: part of the confusion here is that there's a loop, and one variable is for the outer loop and one is for the inner loop. I broke this into a separate function, to avoid variables both inside and outside the loop

internal/engine/live_update_build_and_deployer.go
nicks: you might have other k8s targets that want the synclet, and this only has access to one manifest, see discussion here:
https://github.com/windmilleng/tilt/issues/2948

PR #3052, opened 2020-03-06
web: fix the header box-shadow, which looks like it broke in the transition from scss -> styled-components

nicks:
Hello @hyu,

Please review the following commits I made in branch nicks/boxshadow:

6d4915f6b0645fccae792ccaf52ba26da961670b (2020-03-06 17:07:49 -0500)
web: fix the header box-shadow, which looks like it broke in the transition from scss -> styled-components

### DIFF
--- a/web/src/HUDLayout.tsx
+++ b/web/src/HUDLayout.tsx
@@ -72,7 +72,6 @@ let Header = styled.header`
   padding-right: ${s.Width.sidebar}px;
   height: ${s.Height.HUDheader}px;
   background-color: ${s.Color.grayDarkest};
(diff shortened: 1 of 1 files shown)


nicks:
Screenshot from 2020-03-06 17-07-14

nicks:
oh! that's fine. I noticed that the CSS was malformed, and thought this was a mistake. I updated this PR to remove the malformed CSS

nicks:
I always enjoy a good @hyu UX treatise! But it's also fine to say "I disagree, I intended to remove this"

PR #3076, opened 2020-03-14
hud: Add tilt web teaser

nicks:
Hello @hyu,

Please review the following commits I made in branch nicks/ch4314/hud:

127c76a0d239c0ba353f0c9c373408ac568965c0 (2020-03-13 20:21:55 -0400)
hud: Add tilt web teaser


nicks:
Screenshot from 2020-03-13 20-21-12

I was having trouble making this fit on small terminals, so I
- shorted the language a bit
- removed the error message in the statusbar (because i think that is an artifact from the current, better error UI)

internal/hud/renderer.go
```diff
@@ -181,12 +179,22 @@ func (r *Renderer) renderStatusBar(v view.View) rty.Component {
...
+}
+
+func (r *Renderer) renderStatusBar(v view.View) rty.Component {
+	l := rty.NewConcatLayout(rty.DirHor)
+	l.Add(rty.TextString(" "))
+	l.Add(r.renderStatusMessage(v))
+	l.Add(rty.TextString(" "))
+	l.AddDynamic(rty.NewFillerString(' '))
+
+	msg := " Terminal is limited. For more, open Tilt Web (enter) → "
```
victorwuky:
```suggestion
	msg := " Open Tilt in browser (enter) for full experience. Terminal is limited "	
```

@nicks : I don't think `Tilt Web` as a term is used anywhere and don't necessarily think we should use it, since it could confuse with Tilt Cloud. I think we should start with a positive call to action, before saying Terminal is limited. This is similar to how when you visit a web app with your phone, often there's a banner that says, "Download the iOS app for the full experience". So a positive reason (full experience) to use web instead of (or before) a negative reason (Terminal is limited).

Maybe then we don't even have to say "Terminal is limited", then?

nicks: i like the idea of focusing on the positive. unfortunately, that text is too long

nicks:
new text based on @victorwuky 's suggestion:
Screenshot from 2020-03-20 14-26-31

internal/hud/renderer.go
victorwuky: @nicks : I like your latest version since it assumes the browser is the default way to "explore".

Issue #3062, opened 2020-03-10
get minikube ip correctly with non-default-name clusters (was: Issue to detect local cluster on minikube)


DblK:
Hi,

I saw the [FAQ](https://docs.tilt.dev/faq.html#q-im-getting-push-errors-like-unauthorized-you-dont-have-the-needed-permissions-what-do-i-do) about the issue I encounter today about trying to push to a registry on a local cluster.

Everything was working fine with minikube locally until I changed the name of the project used by minikube to allow me to have two distinct clusters for dev purpose.
Launching minikube with `minikube -p anothercluster` let `tilt` considers this as a remote cluster instead of local one.

As I watch the source [code](https://github.com/windmilleng/tilt/blob/5c73343289128a11046358ec23135f2b6f6a53fa/internal/k8s/env.go#L66) I saw that we have few cases to find out if it's a local or a remote cluster.

Would it be possible to provide a command to tilt to manually specify a local cluster based on the name of the context?

This will solve my issue and allow user to run the local kubernetes cluster where and which the tool they want (maybe k3os on virtualbox or kvm).

Suggestions:
- `tilt set localenv my-context` ==> Set a specific k8s context to be a local one (ignoring push)
- `tilt get localenv`/`tilt ls localenv` => Get/List localenv

or in another way:
- `tilt local set my-context`
- `tilt local ps`/ `tilt local get`/ `tilt local ls`

Pick the ones you find the best ;)

maiamcc:
Thanks for opening an issue! The short-term workaround for you is to make sure your cluster name [has the prefix `minikube`](https://github.com/windmilleng/tilt/blob/5c73343289128a11046358ec23135f2b6f6a53fa/internal/k8s/env.go#L78), e.g. "minikube-anothercluster". Will this work for you for now?

As for being able to manually tell Tilt whether your cluster is local--agree, this would be a great feature! I think #3024 covers that request, do you agree? If so I'll close this issue in favor of that one, and track your request there.

DblK:
Thank for the fast answer! Renaming the cluster name did not the trick.
Tilt showed me:
```bash
Error: Refusing to run 'local' because minikube-xxx might be a production kube context.
If you're sure you want to continue add:
allow_k8s_contexts('minikube-xxx')
before this function call in your Tiltfile. Otherwise, switch k8s contexts and restart Tilt.
```
So I did add the line `allow_k8s_contexts('minikube-xxx')` as first line of my tiltfile, but still get the error.
[...]

PR #3087, opened 2020-03-17
k8s: calls to minikube use -p [clustername] to support minikube clusters with non-default name [ch5455]

maiamcc:
Hello @nicks, @landism,

Please review the following commits I made in branch maiamcc/minikube-node-ip:

c64c5b0b6a26272ca9ef56e813a8dfbe054641d8 (2020-03-17 15:11:04 -0400)
nodeIP uses minikube client, clustername

c5cc313d20e25d95cf0551ca4c9e89731c0d0423 (2020-03-17 15:10:37 -0400)
existing minikube client calls use -p and cluster name


maiamcc:
addresses #3062

internal/k8s/minikube_fake.go
```diff
@@ -0,0 +1,17 @@
+package k8s
+
+import "context"
+
+type FakeMinkube struct {
```
nicks: it seems weird to move this to the k8s package? what's the reasoning behind this change?

maiamcc: I ran into some dependency cycle issues because now the k8sClient needs a minikubeClient, but the minikubeClient needs (or at least wants) a `k8s.KubeContext` (ie the cluster name for `-p <clusterName>`).

This also felt kinda weird to me, would be happy for alternate suggestions.

nicks: ah, i see. OK. I wonder if we're going to have to move the k8s data types to their own package (like kubernetes itself does with v1.Pod, etc). I feel like there are a couple places where we run into this problem.

PR #3088, opened 2020-03-17
cli: only delete resources on `tilt down` that Tilt itself created

### LINKED ISSUE #2805: "tilt down" destroys a pre-existing namespace referenced by helm function
This affects Tiltfiles with the `helm` function. When you run "tilt down", Tilt will destroy the Kubernetes namespace referenced in the `helm` function, **even if that namespace existed in the Kubernetes cluster before running "tilt up" initially**.

**Repro steps:**
1. Create a Helm chart named "foo" containing any one resource:
   ```yaml
   apiVersion: v1
   kind: Pod
   metadata:
     name: foo
   spec:
     containers:
       - name: alpine
         image: alpine
     restartPolicy: Never
   ```

1. Create a Tilefile:
   ```python
   k8s_yaml(helm('foo', namespace='bar'))
   ```

1. Create the namespace manually:
   ```bash
   $ kubectl create namespace bar
   ```

1. Run "tilt up"
   ```bash
   $ tilt up
   
   ● (Tiltfile) ╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌
   HISTORY: EDITED FILES Tiltfile                
   ● foo ╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌
   HISTORY: FIRST BUILD                          
   K8S POD: foo                                  
   ▼ ● uncategorized ╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌
   (Kubernetes objects that don't match a group)
   bar:namespace
   ```

1. Run "tilt down"
   ```bash
   $ tilt down
   
   Beginning Tiltfile execution
   Running: [helm template /Users/asnyder/Documents/foo --generate-name --namespace bar]
   Successfully loaded Tiltfile (106.673875ms)
   Deleting via kubectl: Pod/foo
   
   Deleting via kubectl: Namespace/bar
   ```

**Expected behavior:** Tilt should not destroy a namespace that it did not itself create.

**Actual behavior:** Tilt will always destroy the namespace referenced in the `helm` function in the Tiltfile.

**Known Workaround:** It looks like the `helm()` function is prepending a Namespace resource to the output of `helm template`. That namespace can be stripped by modifying the Tilefile like so:

   ```python
   helm_template = str(helm('foo', namespace='bar'))
   helm_template = helm_template[helm_template.index('---'):]
   k8s_yaml(blob(helm_template))
   ```

jazzdan:
Fixes #2805

### DIFF
--- a/internal/cli/down.go
+++ b/internal/cli/down.go
@@ -9,6 +9,7 @@ import (
 
 	"github.com/windmilleng/tilt/internal/analytics"
 	"github.com/windmilleng/tilt/internal/engine"
+	"github.com/windmilleng/tilt/internal/k8s"
 	"github.com/windmilleng/tilt/internal/tiltfile"
 	"github.com/windmilleng/tilt/pkg/logger"
 	"github.com/windmilleng/tilt/pkg/model"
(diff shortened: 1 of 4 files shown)


internal/cli/down.go
```diff
@@ -73,8 +74,27 @@ func (c *downCmd) down(ctx context.Context, downDeps DownDeps, args []string) er
 		return errors.Wrap(err, "Parsing manifest YAML")
 	}
 
-	if len(entities) > 0 {
-		err = downDeps.kClient.Delete(ctx, entities)
+	var toDelete []k8s.K8sEntity
+	for _, e := range entities {
+		ref := e.ToObjectReference()
+		actual, err := downDeps.kClient.GetByReference(ctx, ref)
```
jazzdan: Is there a more efficient way to batch fetch resources?

nicks:
I don't think this fixes the attached issue?

When I tried running Tilt with the Helm example, it overwrites any existing namespace with the managed-by: Tilt label

jazzdan:
I didn't realize that we overwrite namespaces that already exist with the managed by tilt label so I think this change isn't useful. Closing.

PR #3090, opened 2020-03-18
ignore: also ignore .git from all target dependencies

jazzdan:


internal/ignore/path_matcher.go
```diff
@@ -54,6 +55,10 @@ func CreateFileChangeFilter(m IgnorableTarget) (model.PathMatcher, error) {
 	for _, r := range m.LocalRepos() {
 		matchers = append(matchers, git.NewRepoIgnoreTester(context.Background(), r.LocalPath))
 	}
+	// Also filter out any git directories in any dependencies
+	for _, p := range m.Dependencies() {
+		matchers = append(matchers, git.NewRepoIgnoreTester(context.Background(), p))
```
nicks: out of curiosity, why does this take a context?

internal/ignore/path_matcher.go
```diff
@@ -54,6 +55,10 @@ func CreateFileChangeFilter(m IgnorableTarget) (model.PathMatcher, error) {
 	for _, r := range m.LocalRepos() {
 		matchers = append(matchers, git.NewRepoIgnoreTester(context.Background(), r.LocalPath))
 	}
+	// Also filter out any git directories in any dependencies
+	for _, p := range m.Dependencies() {
```
nicks: wait, I'm confused -- if p has a .git directory, shouldn't it already be in LocalRepos()?

AFAIR, we've had a lot of bugs because of misunderstandings on the semantics of LocalRepos() and Dependencies(), where different codepaths make different assumptions about what they do/how they work. 

So I'm a bit nervous about this code assuming "all Dependencies() are LocalRepos". Does that mean all other codepaths should assume this too?

internal/ignore/path_matcher.go
jazzdan: No idea, it doesn't need to. Can change in a subsequent PR.

internal/ignore/path_matcher.go
jazzdan: This comment inspired me to find a much simpler solution, brb

jazzdan:
Closing in favor of #3091

PR #3091, opened 2020-03-18
tiltfile: custom build deps are repos for images

jazzdan:
Replaces #3090

### DIFF
--- a/internal/tiltfile/docker_test.go
+++ b/internal/tiltfile/docker_test.go
@@ -38,3 +38,27 @@ docker_build('gcr.io/fe', '.', live_update=[
 		},
 		m.ImageTargetAt(0).Dockerignores())
 }
+
+func TeseCustomBuldDepsAreLocalRepos(t *testing.T) {
+	f := newFixture(t)
+	defer f.TearDown()
+
+	f.yaml("fe.yaml", deployment("fe", image("gcr.io/fe")))
+	f.file("Dockerfile", `
+FROM alpine
+ADD . .
+`)
+	f.file("Tiltfile", `
+k8s_yaml('fe.yaml')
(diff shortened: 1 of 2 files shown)


internal/tiltfile/docker_test.go
```diff
@@ -38,3 +38,27 @@ docker_build('gcr.io/fe', '.', live_update=[
...
+`)
+	f.file(".dockerignore", "build")
+	f.file("src/index.html", "Hello world!")
+	f.file("src/.git/hi", "hi")
+
+	f.load()
+
+	m := f.assertNextManifest("fe")
+
+	assert.Contains(t, m.LocalPaths(), f.JoinPath("src", ".git"))
```
maiamcc: wait a sec, I don't see how this test is supposed to work -- i would expect `m.LocalPaths` to contain `src` but not `src/.git`

just tested this on master and indeed, it fails--once you rename it to `TESTCustomBuild...` so it actually gets run :P 

but moreover, imo it's not interesting to test that custom build deps are included in `LocalPaths` (that's literally the code you wrote)--I'd have more confidence from a higher level test that checked that, given `src` as a dep, we ignore changes to `src/.git` (which is the behavior we actually care about, and is implementation-agnostic)

jazzdan: I got back and forth a lot on this myself but my current philosophy is: we already have tests that tests that LocalRepos are ignored correctly (https://github.com/windmilleng/tilt/blob/master/internal/ignore/path_matcher_test.go), so writing another test that tests the thing that sets up Local Repos (this) with the thing that ignores Local Repos (path matcher) would be duplicative.

But yeah I need to make this test actually run and pass.

PR #3092, opened 2020-03-18
tiltfile: add os.realpath

### LINKED ISSUE #2098: Way to clean filepaths in Tiltfile
Requested by kavirajkanagaraj in #tilt:

> does anyone else miss helper function similar to Go’s filepath.Clean("//hello///how/are/you/") in Tiltfile? or python’s os.path.realpath()
> Its mainly combining different paths. We have multiple repositories for source code, helm charts and dockerfiles, so we just accept root of organisation and append different paths. The pain point is it may endup with multiple trailing slashes which causes problem particularly sending docker context. So good to have some helper function there!

It might be nice if there was some way to make it easier for external contributors to contribute to these functions.

See also #1980

jazzdan:
Fixes #2098

### DIFF
--- a/internal/tiltfile/os/os.go
+++ b/internal/tiltfile/os/os.go
@@ -2,6 +2,7 @@ package os
 
 import (
 	"os"
+	"path/filepath"
 	"strings"
 
 	"go.starlark.net/starlark"
@@ -23,6 +24,11 @@ func NewExtension() Extension {
 
 func (e Extension) OnStart(env *starkit.Environment) error {
 	err := env.AddBuiltin("os.getcwd", cwd)
+	if err != nil {
+		return err
+	}
+
+	err = env.AddBuiltin("os.realpath", realpath)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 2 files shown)


internal/tiltfile/os/os.go
```diff
@@ -27,6 +28,11 @@ func (e Extension) OnStart(env *starkit.Environment) error {
 		return err
 	}
 
+	err = env.AddBuiltin("os.realpath", realpath)
```
nicks: should this be os.path.realpath to mirror the python api?

internal/tiltfile/os/os.go
```diff
@@ -61,3 +67,19 @@ func cwd(t *starlark.Thread, fn *starlark.Builtin, args starlark.Tuple, kwargs [
 	dir := starkit.AbsWorkingDir(t)
 	return starlark.String(dir), nil
 }
+
+func realpath(t *starlark.Thread, fn *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
+	var path starlark.String
```
nicks: i would just make this a `string`, starlark will figure it out

internal/tiltfile/os/os.go
jazzdan: When I tried that I got an error saying "multi level modules are not currently supported".

Is that something we control or is that upstream?

nicks: good question! lemme poke at it

PR #3094, opened 2020-03-18
starkit: add support for nested modules

nicks:
Hello @jazzdan,

Please review the following commits I made in branch nicks/modules:

e80bc4035066f0d40fbe96e49e43d5b80112f92b (2020-03-18 16:50:57 -0400)
starkit: add support for nested modules


nicks:
(i don't have a strong preference for os.realpath vs os.path.realpath, but the infrastructure should at least support it)

jazzdan:
Yeah when you merge this in I'll change the other one to be multi level, to match what folks would expect from Python.

PR #3098, opened 2020-03-19
tiltfile: remove k8s resource assembly v1

jazzdan:
We don't have to merge this now, just was in the code and thought that I might as well queue up the PR for when we decide to delete it.

### DIFF
--- a/integration/imagetags/Tiltfile
+++ b/integration/imagetags/Tiltfile
@@ -1,6 +1,4 @@
 
-k8s_resource_assembly_version(2)
-
 include('../Tiltfile')
 
 repo = 'gcr.io/windmill-test-containers'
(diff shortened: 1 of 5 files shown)


jazzdan:
@nicks should we merge this now or wait for right before v1?

nicks:
actually, i think we should add a deprecation warning before we remove this. We still have it in some of our own demo repos, cf https://github.com/windmilleng/servantes/blob/master/Tiltfile

jazzdan:
@nicks actually it is already deprecated https://github.com/windmilleng/tilt/blob/master/internal/tiltfile/k8s.go#L886:L886

nicks:
@jazzdan Ya, that confused me too! But that's not the right warning. That warning shows up only if you're on version 1. It doesn't show up if you're on version 2. Since we're removing the function - _even for version 2 users_ - we need to have a deprecation warning for that

nicks:
another option would be to just remove the implementation, but make the warning something like "this function is now a no-op, you can remove it"

jazzdan:
Going to close this. Too many merge conflicts and we want to deprecate it first anyways.

Issue #3096, opened 2020-03-19
 K8s resources in uncategorized are not updated an change


elemental-lf:
I'm seeing this with `0.12.8` at the moment but the behaviour has been the same in older versions I think: Changes to K8s resources which are grouped under `uncategorized` are not reapplied automatically. I've been bitten be this multiple times when correcting permission issues in `ClusterRoles` during development. This also affects CRDs for which this would probably harder to fix as they are largely immutable after creation and recreating them would have some side-effects as it would delete all custom resources of that kind.

This is related to the discussion in https://github.com/windmilleng/tilt/issues/2989 I think.

(I still like to point out that Tilt is an awesome tool that has helped me immensely during development as it simplifies deployment and container image publishing, reduces turn-around times significantly and also visualizes all the related logs to quickly find a problem . I've looked at one competing solution in the meantime and I couldn't get it to run properly (which also might be my fault). It also felt more invasive than Tilt and made more changes to the Kubernetes cluster which I didn't like. Keep up the good work! Thank you!)

jazzdan:
Hi @elemental-lf, thanks for the kind words! We really appreciate it. :hugs: 

I wasn't able to reproduce this problem with this example repo: https://github.com/jazzdan/uncategorized-example

Is it possible you're telling Tilt about your YAML in a different way? Are you using a `local` to generate your YAML, or Helm/Kustomize or something?

Hope we can get this fixed because it sounds super annoying. Thanks for hanging in there and thanks for your help!

elemental-lf:
Hmm, thanks for the quick feedback, Dan! I've tried your example and it works for me, too. I originally suspected that my problem was simply the way Tilt was currently designed, but your example proofs me wrong. 

My `Tilefile` contains:
```python
helm_template = str(
    helm('charts/benji-k8s',
         namespace='rook-ceph',
         name='benji',
         values=['../../dual/dual/addons/values/global/benji.yaml', '../../dual/dual/addons/values/dev/benji.yaml']))
helm_template = helm_template[helm_template.index('---'):]
k8s_yaml(blob(helm_template))
```
The string splicing is a work-around for https://github.com/windmilleng/tilt/issues/2805. When I change something in the chart the changed file is detected and listed in the `Tilefile` section but no `kubectl apply ` under the `uncategorized` section is attempted. I tried removing the workaround for #2805 but it still doesn't work.

[...]

PR #3103, opened 2020-03-20
k8s: Add 5 second timeout to kubectl apply

jazzdan:
Related to #3096


nicks:
is 5 seconds long enough? my vague understanding is that if you are applying a lot of resources and have some expensive admission controllers, it might take longer. maybe 15 seconds to be safe?

nicks:
also, maybe not in this pr, but i wonder if we should use the kubectl --request-timeout flag

elemental-lf:
@nicks `kubectl --request-timeout` wouldn't help with #3096 as the problem is in `kubectl` itself as I just discovered. `--request-timeout` might be useful for other failure scenarios but the general timeout introduced in this PR protects against any `kubectl apply` hangs in the future.

PR #3108, opened 2020-03-24
all testdata should be treated as binary data

nicks:


nicks:
was getting weird Windows test failures because Windows was normalizing the line-endings in our golden files to CRLF

nicks:
(also changed to use assert.Equal over normal comparison, because it does a lot better surfacing spacing differences)

PR #3110, opened 2020-03-24
cloud: change whoami to POST

landism:
### Problem

/api/whoami is making a GET request with a body, which blows up on some proxies / caches

besides, we don't really want it to get cached anyway

### Solution

make it a POST

### DIFF
--- a/internal/cloud/cloud_username_manager.go
+++ b/internal/cloud/cloud_username_manager.go
@@ -91,7 +91,7 @@ func (c *CloudUsernameManager) CheckUsername(ctx context.Context, st store.RStor
 		return
 	}
 
-	req, err := http.NewRequest("GET", u.String(), body)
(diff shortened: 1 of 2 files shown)


internal/cloud/cloud_username_manager.go
```diff
@@ -91,7 +91,7 @@ func (c *CloudUsernameManager) CheckUsername(ctx context.Context, st store.RStor
 		return
 	}
 
-	req, err := http.NewRequest("GET", u.String(), body)
```
nicks: do we also need to update TFT to handle POSTs? Currently it's only setup to handle GET
https://github.com/windmilleng/tft/blob/master/snapshot_frontend/server/server.go#L106

landism: Yes, that's in a forthcoming PR. Sorry, I should have made that explicit in the description.

PR #3120, opened 2020-03-25
cloud: renames for recent tft work

landism:
1. team name -> team id
2. group engine state cloud status into a single struct
3. cloud username manager -> cloud status manager


internal/hud/webview/convert.go
```diff
@@ -158,9 +158,9 @@ func StateToProtoView(s store.EngineState, logCheckpoint logstore.Checkpoint) (*
 	for k, v := range s.Features {
 		ret.FeatureFlags[k] = v
 	}
-	ret.TiltCloudUsername = s.TiltCloudUsername
+	ret.TiltCloudUsername = s.CloudStatus.Username
 	ret.TiltCloudSchemeHost = cloudurl.URL(s.CloudAddress).String()
-	ret.TiltCloudTeamID = s.TeamName
+	ret.TiltCloudTeamID = s.TeamID
```
nicks: at least we got the protobuf names right!!

PR #3123, opened 2020-03-26
circleci: windows builds should use the same image as everyone else

nicks:


nicks:
i don't know why vscode is aggressively reorganizing the go.mod 🤷‍♂

PR #3141, opened 2020-03-27
cloud: refresh cloud status when state changes, and periodically

landism:


internal/cloud/cloud_status_manager.go
```diff
@@ -166,12 +193,10 @@ func (c *CloudStatusManager) OnChange(ctx context.Context, st store.RStore) {
...
 	// we never make a request post-(2), where the token was registered
 	// This is mitigated by - a) the window between (1) and (3) is small, and b) the user can just click refresh again
-	if time.Now().Before(sleepingAfterErrorUntil) ||
-		state.CloudStatus.Username != "" ||
-		state.CloudStatus.TokenKnownUnregistered ||
-		currentlyMakingRequest {
+	allowedToPerformLookup := !time.Now().Before(sleepingAfterErrorUntil) && !currentlyMakingRequest
+
+	if needsLookup && allowedToPerformLookup {
+		go c.CheckStatus(ctx, st, state.CloudAddress, requestKey, false)
```
nicks: fwiw, i've been moving more of the subscribers to do their work synchronously (not using goroutines at all), and just doing the parallelism more intelligently in store. it makes testing a bit easier.

internal/cloud/cloud_status_manager.go
```diff
@@ -10,28 +10,47 @@ import (
...
+	clock  clockwork.Clock
+
+	mu sync.Mutex
 
 	sleepingAfterErrorUntil time.Time
 	currentlyMakingRequest  bool
-	mu                      sync.Mutex
+
+	lastRequestKey       statusRequestKey
+	lastSuccessfulLookup time.Time
```
nicks: it seems a bit odd to have two different backoff mechanisms, one where the last time is stored and one where the expiry is stored.

internal/cloud/cloud_status_manager.go
landism: yeah, that sounds good to keep in mind

PR #3145, opened 2020-03-27
enging: move process exiting logic into exit.Controller

nicks:
Hello @maiamcc, @jazzdan,

Please review the following commits I made in branch nicks/exitcontrol:

7e6200481f6d56511a083f62c6e03de21b7c186d (2020-03-27 17:53:56 -0400)
enging: move process exiting logic into exit.Controller

8f244fca99878b79388b4df04aea134ca5652651 (2020-03-27 17:11:32 -0400)
store: Simplify subscription notification
In the old world, every Notify() would create a new goroutine.
The goroutine would block until the subscriber was done processing the last event,
then double-check if it still needed to process.
This made debugging difficult, because you could have 1000s of
goroutines blocked on a single slow subscriber.

In the new world, there are at most two goroutines in flight.

### DIFF
--- a/internal/engine/upper.go
+++ b/internal/engine/upper.go
@@ -7,11 +7,8 @@ import (
 	"path/filepath"
 	"time"
 
-	"github.com/windmilleng/tilt/internal/engine/local"
-
 	"github.com/davecgh/go-spew/spew"
 	"github.com/opentracing/opentracing-go"
-	"github.com/pkg/errors"
 	"github.com/windmilleng/wmclient/pkg/analytics"
 	v1 "k8s.io/api/core/v1"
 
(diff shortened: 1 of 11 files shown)


internal/engine/upper.go
```diff
@@ -130,9 +130,10 @@ func (u Upper) Init(ctx context.Context, action InitAction) error {
 }
 
 func upperReducerFn(ctx context.Context, state *store.EngineState, action store.Action) {
+	log.Printf("ACTIONX %T %+v", action, action)
```
jazzdan: :eyes:

nicks:
There's not much new logic in this PR. We basically have a bunch of logic that controls when Tilt exits. It's spread across lots of different places (the Store and the Reducer). But isn't particularly well-tested.

This tries to move it to a general controller that decides when Tilt exists.

The hope is that this sets us up well to have more command-line modes with different exit conditions (like https://github.com/windmilleng/tilt/issues/2794)

internal/engine/upper.go
nicks: oops :joy: :joy: :joy:

PR #3148, opened 2020-03-30
web: fix snapshot log highlight snippet

landism:
### Problem

When we take a snapshot with a highlighted snippet, Tilt sets the snippet text to "[object Object]", when it should be setting it to the selected text. This results in cloud.tilt.dev/snapshots showing this:
image

### Solution

[This line](https://github.com/windmilleng/tilt/blob/e6198fa9a41c5443dabcb700b04ad6ace538c878/web/src/LogPane.tsx#L208) was probably intended to be `sel.toString()`, and making that change gets us most of the way there. Unfortunately, now that the resource name prefixes are pulled out into separate HTML elements, simply using `sel.toString()` returns something like:
```
fortune
   line1
fortune
   line2
```

which looks really weird as a preview!

So I did something slightly more complicated, explained in the comments.

### DIFF
--- a/web/src/LogPane.test.tsx
+++ b/web/src/LogPane.test.tsx
@@ -1,9 +1,10 @@
 import React from "react"
 import ReactDOM from "react-dom"
-import LogPane from "./LogPane"
+import LogPane, { logText } from "./LogPane"
 import renderer from "react-test-renderer"
-import { mount } from "enzyme"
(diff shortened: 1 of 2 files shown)


landism:
actually, I'll extract the function that turns nodes -> string and write a test for that

jazzdan:
ooh I like that idea

PR #3149, opened 2020-03-30
circleci: update golangci-lint, fix problems

jazzdan:


internal/analytics/context.go
```diff
@@ -2,7 +2,9 @@ package analytics
 
 import "context"
 
-const analyticsContextKey = "Analytics"
+type contextKey struct{}
```
jazzdan: From [the docs for `context.WithValue`](https://golang.org/pkg/context/#WithValue):

>The provided key must be comparable and should not be of type string or any other built-in type to avoid collisions between packages using context.

PR #3150, opened 2020-03-30
starkit: CurrentExecPath is based on what's being loaded rather than the call stack

landism:
### Problem

When we execute code from a Tiltfile, we set the CWD equal to the path to the Tiltfile at the bottom of the call stack. This prevents us from defining extensions that, e.g., expose a function that declares a `docker_build`, or runs a `local`, with the caller's working directory.

### Solution

Set the CWD from the currently executing Tiltfile instead of from the callstack.
(laid out more explicitly in unit tests below)

### DIFF
--- a/internal/tiltfile/starkit/path.go
+++ b/internal/tiltfile/starkit/path.go
@@ -19,15 +19,11 @@ func AbsWorkingDir(t *starlark.Thread) string {
 	return filepath.Dir(CurrentExecPath(t))
 }
 
-// Path to the file at the bottom of the call stack.
+// Path to the file that's currently executing
 func CurrentExecPath(t *starlark.Thread) string {
-	depth := t.CallStackDepth()
-	for i := 0; i < depth; i++ {
-		filename := t.CallFrame(i).Pos.Filename()
-		if filename == "<builtin>" {
-			continue
-		}
-		return filename
+	ret := t.Local(execingTiltfileKey)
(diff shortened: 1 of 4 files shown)


internal/tiltfile/starkit/path.go
```diff
@@ -19,15 +19,11 @@ func AbsWorkingDir(t *starlark.Thread) string {
...
 func CurrentExecPath(t *starlark.Thread) string {
-	depth := t.CallStackDepth()
-	for i := 0; i < depth; i++ {
-		filename := t.CallFrame(i).Pos.Filename()
-		if filename == "<builtin>" {
-			continue
-		}
-		return filename
+	ret := t.Local(execingTiltfileKey)
+	if ret == nil {
```
nicks: i'm a little bit worried that we have deferred functions (like workloadToResourceFn) that aren't called from an active starlark thread, but maybe we should just change that API so everything has a real thread

landism: Good call. This breaks with workload_to_resource_function, but that also breaks in master, so not going to block this PR on it. I've filed https://app.clubhouse.io/windmill/story/5863/workload-to-resource-function-does-not-use-a-proper-starkit-thread-which-can-cause-internal-errors.

djcp:
oh lord this breaks so many things for us.

PR #3161, opened 2020-04-01
cli: Introduct `tilt ci` for waiting until pods are healthy

nicks:
Hello @jazzdan, @maiamcc,

Please review the following commits I made in branch nicks/ch5914/ci:

4b9e391747100f15ed52308ebe0627d19f2d018f (2020-04-01 11:38:30 -0400)
cli: Introduct `tilt ci` for waiting until pods are healthy

### DIFF
--- a/internal/cli/ci.go
+++ b/internal/cli/ci.go
@@ -1,0 +1,94 @@
+package cli
+
+import (
+	"context"
+	"fmt"
+	"log"
+	"time"
+
+	"github.com/fatih/color"
+	"github.com/spf13/cobra"
+
+	"github.com/windmilleng/tilt/internal/analytics"
+	"github.com/windmilleng/tilt/internal/cloud"
+	"github.com/windmilleng/tilt/internal/store"
+	"github.com/windmilleng/tilt/internal/tiltfile"
+	"github.com/windmilleng/tilt/pkg/logger"
+)
+
+type ciCmd struct {
+	fileName             string
+	outputSnapshotOnExit string
+}
+
+func (c *ciCmd) register() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:                   "ci [<tilt flags>] [-- <Tiltfile args>]",
+		DisableFlagsInUseLine: true,
+		Short:                 "Start Tilt in CI/batch mode with the given Tiltfile args",
+		Long: fmt.Sprintf(`
+Starts Tilt and runs resources defined in the Tiltfile.
+
+Exits with failure if any build fails or any server crashes.
+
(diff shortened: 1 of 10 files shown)


internal/cli/ci.go
```diff
@@ -0,0 +1,91 @@
...
+
+func (c *ciCmd) register() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:                   "ci [<tilt flags>] [-- <Tiltfile args>]",
+		DisableFlagsInUseLine: true,
+		Short:                 "Start Tilt in CI/batch mode with the given Tiltfile args",
+		Long: `
+Starts Tilt and runs services defined in the Tiltfile.
+
+Exits with failure if any build fails or any server crashes.
```
maiamcc, line 33: seems like the main difference between this and `--watch=false` (or whatever it's called now) is that this exits after pods are up and ready, and the latter exists after the final yaml apply is done. maybe call that out in the help text? (probably don't bother contrasting it to `--watch=false` but DO say when this command is "done".)

internal/cli/ci.go
```diff
@@ -0,0 +1,91 @@
...
+`,
+	}
+
+	addWebServerFlags(cmd)
+
+	cmd.Flags().StringVar(&updateModeFlag, "update-mode", string(buildcontrol.UpdateModeAuto),
```
maiamcc, line 39: do we expect this mode to run any updates after the initial deploy?

internal/cli/ci.go
nicks: For this particular documentation, I think we want to assume the user has never heard of --watch=false. Given that assumption, what do you hope to accomplish by introducing it here?

If we learn that `tilt ci` is a better fit for what most people need, then we don't want to encourage them to learn about --watch=false

internal/cli/ci.go
nicks: good point, removed

internal/cli/ci.go
nicks: oh, nm, i misread what you were saying, added some more text on exit conditions

victorwuky:
@nicks : Is this net new in v0.13.0? Seems like something we should blog about. I can start that.

nicks:
@victorwuky yup! I wanted to get it into a few trusted customer hands. They filed some bugs, which we fixed and are going to be in the next release. i can file a story about the blogpost

victorwuky:
Thanks @nicks. I created https://app.clubhouse.io/windmill/story/6051/blog-post-on-tilt-ci

PR #3164, opened 2020-04-01
snapshots: scroll to highlighted section on load

landism:
### Repro
Scroll up in the tilt log, highlight some lines, share a snapshot, open the snapshot.

### Expected
When opening the snapshot, the view scrolls to the highlighted lines

### Observed
The view scrolls to the bottom of the log

### Fix
It turns out that we're only scrolling to highlight on componentDidMount, except the highlight state is null at that point, so there's nothing to scroll to. So, let's scroll there on componentDidUpdate.
Also, when we call scrollIntoView, it was putting the top of the highlight at the top of the browser window, so the first few lines were hidden underneath the header (which also means if the highlight was only a couple lines, it'd be entirely hidden under the header!) - I moved it to the center of the screen, instead.

### DIFF
--- a/web/src/LogPane.test.tsx
+++ b/web/src/LogPane.test.tsx
@@ -441,12 +441,12 @@ it.each(["update", "mount"])(
         showManifestPrefix={false}
         handleSetHighlight={fakeHandleSetHighlight}
         handleClearHighlight={fakeHandleClearHighlight}
-        highlight={verb == "mount" ? highlight : null}
+        highlight={verb === "mount" ? highlight : null}
         isSnapshot={true}
       />
     )
 
-    if (verb == "update") {
+    if (verb === "update") {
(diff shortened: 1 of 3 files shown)


web/src/LogPane.test.tsx
```diff
@@ -423,35 +423,43 @@ it("renders highlighted lines", () => {
...
+      text: "foo\nbar",
+    }
+    const root = mount<LogPane>(
+      <LogPane
+        manifestName={""}
+        logLines={logLinesFromString(longLog)}
+        showManifestPrefix={false}
+        handleSetHighlight={fakeHandleSetHighlight}
+        handleClearHighlight={fakeHandleClearHighlight}
+        highlight={verb == "mount" ? highlight : null}
```
jazzdan, line 444: Triple equals!

PR #3166, opened 2020-04-01
web: warnings as errors

landism:
1. ensure js warnings cause `make check-js` to fail
2. fix any such warnings. these fell into a few camps:
   - `==` / `!=` instead of `===` / `!==`
   - unused variables / imports
   - using `.map` on a side-effecting func w/ no return, instead of `.forEach`

### DIFF
--- a/web/src/LogStore.ts
+++ b/web/src/LogStore.ts
@@ -321,7 +321,7 @@ class LogStore {
 
       if (
         candidate.firstLineIndex > startSpan.firstLineIndex &&
-        (nextBuildSpan === null ||
+        (!nextBuildSpan ||
           candidate.firstLineIndex < nextBuildSpan.firstLineIndex)
       ) {
         spans[key] = candidate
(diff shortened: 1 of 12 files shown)


web/src/LogStore.ts
```diff
@@ -315,13 +315,13 @@ class LogStore {
...
       let candidate = this.spans[key]
-      if (candidate.manifestName != startSpan.manifestName) {
+      if (candidate.manifestName !== startSpan.manifestName) {
         continue
       }
 
       if (
         candidate.firstLineIndex > startSpan.firstLineIndex &&
-        (nextBuildSpan == null ||
+        (nextBuildSpan === null ||
```
nicks, line 324: can we just make this !nextBuildSpan?

nextBuildSpan === null is almost never what you want, because that excludes undefined

jazzdan: Yeah I wish this lint excluded cases where one of the operands is `null`.

PR #3171, opened 2020-04-02
down: add option to delete namespaces

landism:


victorwuky:
@landism: I don’t see any docs for this feature. Can you please point me to it or add this to the docs so that users know that this feature exists.

nicks:
@victorwuky don't worry, the documentation in the command also got auto-generated to https://github.com/windmilleng/tilt.build/pull/512

victorwuky:
Thanks for pointing me to that @nicks . I'd like to link to the docs in release notes (and not commit), since the release notes should help a user learn about the feature and use it. See https://github.com/windmilleng/tilt/releases/tag/v0.13.1.

I'll be sure to update the release notes once the docs update is merged.

cc @maiamcc .

maiamcc:
@victorwuky release notes updated, good catch!

Issue #3125, opened 2020-03-26
Team owner promotes current Tilt version for team


victorwuky:
### Background
Tilt teams is a new feature for [Tilt Cloud](https://cloud.tilt.dev/) that allows you to create the concept of a "team" that can represent your entire software organization. The team creator is by default the team owner. You can associate a team with one (or even multiple) Tiltfiles using a given team id. When a user runs Tilt on their machine with a Tiltfile that is associated with that team id, they are automatically associated with that team.

### Description
From within Tilt Cloud, the team owner will already be able to see what recent version users have been using, on their team. This feature will further allow the team owner to control the "current version" for the team, essentially an encouraged version for users to upgrade to, but not beyond that. In the screenshot below, when Laura clicks `Promote to 12.5`, the current team version will become 12.5. Next time Matt runs Tilt, they'll get a notification nudging them to upgrade to 12.5.

Tilt Cloud as shown to Laura
![promote](https://user-images.githubusercontent.com/10860550/77667576-ca4bec80-6f58-11ea-888d-0c031c351238.png)

Tilt as shown to Matt
![upgrade](https://user-images.githubusercontent.com/10860550/77668152-6aa21100-6f59-11ea-9066-e8122c0749ee.png)

### Why
We want to afford the team owner (oftentimes the developer experience (DX) owner) an opportunity to manage the current team Tilt version for the team. Say a new version of Tilt is released, but the team owner (Laura in the above screenshot) wants to do some testing first. Laura gets an opportunity to do that, before essentially "rolling out" the version to the rest of her team.

### Caveats
The current team version is not a hard blocker that prevents users from installing certain versions of Tilt. It's a soft encouragement leveraging the in-app nudge.

victorwuky:
This is done. SEe https://docs.tilt.dev/teams.html#canary-tilt-releases-before-your-team-upgrades.

PR #3172, opened 2020-04-02
tiltfile: allow specifying a tilt version constraint

### LINKED ISSUE #3159: Tiltifle should be able to specify min/max Tilt version
As a devtools owner, I might want to declare my Tiltfile only compatible with particular versions of Tilt.

```
version_settings(min_version='0.10.0', max_version='0.11.0')
```

Tilt would error if its version string did not match what's in the Tiltfile, rather than failing cryptically.

We discussed this a bit when we first added version_settings in https://github.com/windmilleng/tilt/issues/2520. David at Unu has also asked L about this.

landism:
fixes #3159

uses the constraint syntax described [here](https://getcomposer.org/doc/articles/versions.md#writing-version-constraints)


internal/cli/down.go
```diff
@@ -19,7 +19,6 @@ import (
 
 type downCmd struct {
 	fileName         string
-	deleteNamespaces bool
```
nicks, line 22: ?

was this intentional?

internal/tiltfile/version/version.go
```diff
@@ -1,16 +1,24 @@
 package version
 
 import (
+	"fmt"
+
+	"github.com/mcuadros/go-version"
```
nicks, line 6: would it be reasonable to use blang/semver, since we already use it for docker? worried about having multiple competing semver parsers

nicks: also, https://github.com/mcuadros/go-version says that it's readonly/archived, which does not give me a lot of confidence that it's well-supported

landism: > would it be reasonable to use blang/semver, since we already use it for docker?

yep, I didn't spot that we had that when I was looking for existing version deps

internal/cli/down.go
landism: no

internal/tiltfile/tiltfile_state.go
```diff
@@ -163,7 +166,7 @@ func (s *tiltfileState) loadManifests(absFilename string, userConfigState model.
 		s.k8sContextExt,
 		dockerprune.NewExtension(),
 		analytics.NewExtension(),
-		version.NewExtension(),
+		version.NewExtension(s.tiltVersion.Version),
```
nicks, line 169: can we have `wire` dependency-inject this? that way we don't have to maintain the handwritten dependency-injection of Version. see how k8sContextExt works.

ideally, we should probably just have a wire provider that creates all the extensions.

internal/tiltfile/version/version.go
```diff
@@ -19,26 +27,50 @@ func (e Extension) NewState() interface{} {
...
+	state, err := GetState(m)
+	if err != nil {
+		return nil, errors.Wrap(err, "internal error reading version settings from starkit model")
+	}
+
+	checkUpdates := state.CheckUpdates
+	var constraint string
 	if err := starkit.UnpackArgs(thread, fn.Name(), args, kwargs,
-		"check_updates", &checkUpdates); err != nil {
+		"check_updates?", &checkUpdates,
```
nicks, line 49: i'm a little bit worried about every extension inventing a new convention for unpacking args into a data model. If there's a shortcoming in the API that we want to fix, lots of bespoke call patterns make it harder to fix.

nicks: i wonder if a pattern that might make sense is to do the UnpackArgs in SetState()?

e.g.,

```
err = starkit.SetState(thread, func(settings model.VersionSettings) (model.VersionSettings, err) {
  ...
  starkit.UnpackArgs(thread, fn.Name(), args, kwargs,
		"check_updates", &settings.CheckUpdates)
  ...
}
```

not totally sure, just spitballing here, this is ok for now

landism: yeah, that's a lot simpler, thanks

I also started a [convo](https://github.com/google/starlark-go/issues/267) on the starlark-go repo that I think will let us cheaply deal with types like `Union[bool|None]` that have been annoying in a bunch of cases.

internal/tiltfile/version/version.go
```diff
@@ -19,26 +27,37 @@ func (e Extension) NewState() interface{} {
...
+		ver, err := semver.Parse(e.tiltVersion)
+		if err != nil {
+			return nil, errors.Wrapf(err, "internal error parsing tilt version '%s'", e.tiltVersion)
+		}
+		rng, err := semver.ParseRange(constraint)
+		if err != nil {
+			return nil, errors.Wrapf(err, "error parsing version constraint")
+		}
+		if !rng(ver) {
+			return nil, fmt.Errorf("you are running Tilt version %s, which doesn't match the version constraint specified in your Tiltfile: '%s'", e.tiltVersion, constraint)
```
drubin, line 57: It would be nice to include a helpful error messaging like linking to the website saying howto upgrade.

Also explaining which version they should be running.

landism: > It would be nice to include a helpful error messaging like linking to the website saying howto upgrade.

Good call. I'll need to make sure we a page appropriate to link to.

> Also explaining which version they should be running.

Yeah, that would be nice, but I don't see a cheap way to do it with the information available, and it'd greatly increase the complexity of this feature to figure it out. We're currently logging the constraint. I'm inclined to wait to see how much use this gets / whether that's a problem before worrying about it. Also how this fits into #3125.

drubin: >Good call. I'll need to make sure we a page appropriate to link to. 
https://docs.tilt.dev/upgrade.html this is what I used into my plugin. https://github.com/windmilleng/tilt-extensions/pull/9 

>Yeah, that would be nice, but I don't see a cheap way to do it with the information available, and it'd greatly increase the complexity of this feature to figure it out. We're currently logging the constraint. 

Ah, I couldn't see a log message for this, so I assumed it wasn't there.

Was just a suggestion I don't think you should complicate it too much.

nicks:
is this PR waiting on anything?

landism:
> is this PR waiting on anything?

I was going to add a docs page with instructions for installing a specific version and link to that in the error message, but we can add that later.

PR #3176, opened 2020-04-03
Add a goreleaser Docker image

nicks:


nicks:
This builds the a Docker image for CI everytime we do a tilt release. Since you're exterminator next week, you get to be the guinea pig :). When you get a chance, please send me your docker username so i can give you access to the image repo.

maiamcc:
Ulllp, okay, let's try it! Docker username is `maiamcc`.

I don't see where in this PR we update the tag of the image that CI uses--what am I missing?

nicks:
I'm planning to use this in the example guides (cf https://github.com/windmilleng/tilt-example-python/blob/master/.circleci/config.yml)

maiamcc:
Neat! Though that image ref is hardcoded, right? Is the goal a ready supply of images that match the latest deploy that we can use as needed, or to dynamically update the image used in tests?

nicks:
good question!

1) The goreleaser template in this pr will update both tilt:latest and tilt:VERSION, so people can use whichever one they prefer.

2) When users choose between `:latest` and a pinned version, they're making the trade-off between "I want to try this out right now and not think too hard about what version I'm using" and "OK, now that I've played with this a bit and know what it is, I want to make updates predictable".

3) Practically, for our guides, we should probably use a pinned version, to make it easier to test if a Tilt upgrade broke the build (rather than introducing cache uncertainty)

maiamcc:
Cool, that all makes sense. My remaining question is:
> Practically, for our guides, we should probably use a pinned version, to make it easier to test if a Tilt upgrade broke the build (rather than introducing cache uncertainty)

When/how do you anticipate the version in our guides being updated? Do we want people to do that manually whenever we do a release? Whenever they make a change to the guide? Whenever we feel like it?

nicks:
I'll take ownership of updating them manually on major releases and/or when we're worried about something changing.

PR #3187, opened 2020-04-08
web: Refactor Sidebar component (prep for Account Menu)

hyu:
I changed the Sidebar component so I can add a SidebarAccount component as a direct child of HUD. Otherwise, I'd have to add it in this annoyingly nested hierarchy:

```
HUD > Sidebar > SidebarAccountMenu
```

...which also would have been limiting. It's nice when a child of **HUD.jsx** is a PureComponent or otherwise limits excessive rendering. This refactor sets me up to do this:

- `HUD > SidebarResources` - (PureComponent) To see the list of resources

- `HUD > SidebarAccountMenu` - (Functional Component with a React Hook) Account button that toggles a Menu.

## Reference Image
This shows what the Account Menu will look like. It's not implemented in this PR, but gives context for why this PR was created.

Screen Shot 2020-04-07 at 8 54 10 PM


web/src/HUD.test.tsx
```diff
@@ -73,25 +73,25 @@ it("renders resource", async () => {
   const hud = root.find(HUD)
   hud.setState({ view: oneResourceView() })
   expect(root.find(".Statusbar")).toHaveLength(1)
-  expect(root.find(".Sidebar")).toHaveLength(1)
+  expect(root.find("section.Sidebar")).toHaveLength(1)
```
hyu, line 76: When I convert a component from Sass to using Styled Components, I need to add an extra class (here, "Sidebar") which is purely so the test can pick up the component. Styled Components also generates extra divs wrapping the actual component, so it's necessary to narrow down the `find` using the HTML element (ex: `section`)

There are better [ways](https://www.google.com/search?client=firefox-b-1-d&q=jest+styled+components) to test Styled Components, but I didn't dig into that for this PR.

web/src/Statusbar.scss
```diff
@@ -87,7 +87,6 @@
 // Progress
 .Statusbar-progressPanel {
   width: $sidebar-width - $statusbar-tiltPanel-width;
-  max-width: $sidebar-maxWidth;
```
hyu, line 90: As part of this refactor, I converted Sass to Styled Components and thus reviewed the constants in **constants.scss** to see which ones I could de-dupe into **style-helpers.js**. I can across this one, `$sidebar-maxWidth` which doesn't really seem to be doing much. Both width and max-width are both pixel values, so AFAIK the browser always renders "width" anyway.

PR #3197, opened 2020-04-09
hud: make the hud colors consistent with the web colors

nicks:
Hello @maiamcc, @landism,

Please review the following commits I made in branch nicks/ch4775/status:

32a5cd94cde07f02e80e313cc810a9dd91537867 (2020-04-09 19:22:12 -0400)
hud: make the hud colors consistent with the web colors


internal/hud/resourceview.go
```diff
@@ -75,45 +75,43 @@ type statusDisplay struct {
...
-			return statusDisplay{color: cPending, spinner: true}
-		} else if res.CrashLog.Empty() {
-			return statusDisplay{color: cGood}
-		} else {
-			return statusDisplay{color: cBad}
-		}
-	}
-
-	if !res.CurrentBuild.Empty() && !res.CurrentBuild.Reason.IsCrashOnly() {
+// NOTE: This should be in-sync with combinedStatus in the web UI
```
maiamcc, line 78: is it worth it/is it easy to put this code in its own package and have both tui and webui import the same function, instead of keeping it in sync in two places?

nicks: They're implemented in different languages, and have different return values (since the return values are based on what's possible to display in each). so...not easy.

maiamcc: Oh whoops I didn't realize the other one was in JS. Okay cool 👍

PR #3198, opened 2020-04-10
re-enable buildkit on minikube v1.8.0+. Fixes https://github.com/windmilleng/tilt/issues/3192

nicks:
Hello @jazzdan, @maiamcc,

Please review the following commits I made in branch nicks/ch6163:

ee9f8629c328611ea1b06ddfcdc862ff3fe2c958 (2020-04-10 10:17:26 -0400)
re-enable buildkit on minikube v1.8.0+. Fixes https://github.com/windmilleng/tilt/issues/3192

### DIFF
--- a/internal/docker/env.go
+++ b/internal/docker/env.go
@@ -5,12 +5,22 @@ import (
 	"fmt"
 	"os"
 
+	"github.com/blang/semver"
 	"github.com/docker/cli/opts"
 	"github.com/pkg/errors"
 
 	"github.com/windmilleng/tilt/internal/container"
 	"github.com/windmilleng/tilt/internal/k8s"
+	"github.com/windmilleng/tilt/pkg/logger"
 )
+
+// We didn't pick minikube v1.8.0 for any particular reason, it's just what Nick
+// had on his machine and could verify with. It's hard to tell from the upstream
+// issue when this was fixed.
+//
+// We can move it earlier if someone asks for it, though minikube is pretty good
+// about nudging people to upgrade.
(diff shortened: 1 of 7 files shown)


nicks:
i didn't pick minikube v1.8.0 for any particular reason, it's just what I had on my machine and could verify with. (I couldn't figure out from the upstream ticket when this was fixed). We can move it earlier if someone asks for it, though minikube is pretty good about nudging people to upgrade

internal/k8s/minikube_fake.go
```diff
@@ -2,16 +2,21 @@ package k8s
 
 import "context"
 
-type FakeMinkube struct {
```
maiamcc, line 5: heh `Minkube` 😅

internal/docker/env.go
```diff
@@ -5,13 +5,17 @@ import (
...
+	"github.com/blang/semver"
 	"github.com/docker/cli/opts"
 	"github.com/pkg/errors"
 
 	"github.com/windmilleng/tilt/internal/container"
 	"github.com/windmilleng/tilt/internal/k8s"
+	"github.com/windmilleng/tilt/pkg/logger"
 )
 
+var minMinikubeVersionBuildkit = semver.MustParse("1.8.0")
```
maiamcc, line 17: the note in this PR about why v1.8.0 would be good to add here too, I think

nicks: good point, comment added!

internal/k8s/minikube_fake.go
nicks: i'm amazed at how consistent we were

PR #3200, opened 2020-04-10
dc: pipe configs to docker compose via stdin

maiamcc:
was playing around with this today, putting it up to run tests and see if it works


maiamcc:
hey @nicks / @landism -- I did a little work on "pass docker compose configs via stdin so we can modify them at will" during my spare time while exterminating.

 could you take a QUICK look at this (particularly `dockercompose/config.go` and `client.go`)? not looking for a full review (i can break this PR up for that) so much as a gut check for whether a. this is on the right track and b. this is worth finishing out?

if "yes" to both, I'll talk to victor about prioritizing this, because it shouldn't take too much longer and will unblock [docker_compose.entrypoint](https://app.clubhouse.io/windmill/story/6042/ensure-entrypoint-works-for-docker-compose), which will make [our restart_container replacement](https://app.clubhouse.io/windmill/epic/5410/replace-restart-container) much nicer (though it's not strictly necessary for the restart_container epic to proceed).

nicks:
I think there are two separate parts of this PR that are worth talking about

The first part is about taking docker-compose YAML on stdin. This change seems good and non-controversial. But I think this is going to introduce subtle bugs w/r/t handling current working directory correctly, see this document
https://docs.docker.com/compose/reference/overview/#command-options-overview-and-help
particularly the working directory default and how `-f -` affects that

The second part is about the data model for the docker-compose configs. Docker just announced that major efforts are underway to standardize the format. see
https://compose-spec.io/
and
https://github.com/compose-spec/compose-go
(the Go reference implementation)

so I'm a little bit nervous about investing a bunch of time into building our own data model, rather than doing something very hacky to do the replacement.

Thoughts?

maiamcc:
Cool, this is a good breakdown, thank you!

> The first part is about taking docker-compose YAML on stdin. This change seems good and non-controversial. But I think this is going to introduce subtle bugs w/r/t handling current working directory correctly

Yeah I hadn't considered the working dir issue 😬 but looks like we can work around that by passing `--project-directory`: 
```
Specify an alternate working directory
(default: the path of the Compose file)
```

> The second part is about the data model for the docker-compose configs. Docker just announced that major efforts are underway to standardize the format.

Yep, I totally agree with you about not investing a bunch of duplicative effort here. From the stuff I've found online, I don't have a clear sense of where their standardization efforts stand in re:
1. what compose file versions are compatible, and
2. whether the project is done yet / when it will be complete.
Do you have a better sense of this?

I DO feel like the solution in this PR is pretty quick and dirty though -- we're basically throwing everything into a map, and for the handful of fields we care about, we're getting/setting in the map itself (not bothering to split them out into struct fields etc). Do you feel like the implementation here is too much work to get behind, or sufficiently low-effort? If it seems like too much, do you have a hackier solution?

nicks:
maybe I'm misunderstanding - why do we need to change ServiceConfig at all?

To zoom out a bit - you specifically asked for my gut check on dockercompose/config.go.. I'm confused because it seems like these changes don't make sense in the short term (because we can pass in the YAML on stdin without them) and don't make sense in the long term (because they're duplicative w/ compose-spec). So I'm not clear what problem you're solving.

maiamcc:
OH yes you're totally right, I left out a bit of context. The eventual goal here is to be able to modify the docker-compose config that we apply, e.g. to inject an `entrypoint`.  (I have a vague recollection that we wanted to modify the config in other cases too?) So with that in mind, I guess the question then becomes:

is "ability to modify the dc.yaml we run with" a sufficiently compelling story for the short term that we would do SOME work on data model? And if so, how much work on the data model is "enough to get by" without being TOO time-intensive/duplicative?

I'm inclined to say that yes, ability to modify the yaml we run with is useful, if only just to bring `entrypoint` up to parity for k8s and dc; but curious about your thoughts.

nicks:
OK, that makes sense. Do you expect that we will need all the modification methods you added in this PR?

I guess I would have used the existing approach that we use with K8sTarget (i.e., store the YAML in the data model, and have narrowly-defined functions that parse it, inject things, and re-serialize). If we invent this new approach, I predict that there will be a strong temptation to invest a bunch of time into the data model (both by you, and by other engineers).

But ultimately, it's your call

maiamcc:
Mmhmm that makes sense. I guess I'm confused about what would be a more minimal data model, in your mind--I was trying to keep everything as unconstrained as possible by just storing everything as a big map of maps and only accessing the fields we actually need. were you imagining storing the yaml and just using string manipulation to change the fields we care about? or something else? if you have an idea in mind, would love a quick sketch.

(it's worth noting that the majority of the methods here existed in the previous code, where they also involved a small amount of yaml-parsing--now they just exist in a different form, but to me the old and new versions of say, getting the `BuildContext` seem about equal in complexity.)

nicks:
ya, i would have imagined storing the original yaml and the entrypoint in the EngineState, and doing the substitution at build-and-deploy-time, similar to how K8sTarget works.

maiamcc:
Cool, thanks @nicks for a bunch of useful insights. Closing this PR -- it was useful as an exploration, I'm not doing any further work on this at the moment. Future work documented in this epic: [Implement entrypoint support for Docker Compose](https://app.clubhouse.io/windmill/epic/6260/implement-entrypoint-support-for-docker-compose).

PR #3201, opened 2020-04-11
web: user icon toggles Tilt Cloud Account menu

hyu:
To make it easier to sign up for Tilt Cloud, this PR adds a user icon at the top right of the screen. You click it, and toggle a menu that encourages you to sign up for Tilt Cloud. If you're already signed in, you can see your username and Tilt instance's team info.

### Known Issues
- You want to close the dropdown menu by clicking on the background, but that doesn't actually work right now. Sorry! Another PR.
- No analytics yet
- Copy about Tilt Cloud is quite blah, copied from the outdated homepage.

--------------------------------------------------

### Not Signed In
2020-04-10 - Tilt Cloud account menu - not signed in

### Signed In
2020-04-10 - Tilt Cloud account menu - signed in


web/src/Button.ts
```diff
@@ -0,0 +1,25 @@
...
+  justify-content: center;
+  border-radius: ${SizeUnit(0.15)};
+  padding-left: ${SizeUnit(0.75)};
+  padding-right: ${SizeUnit(0.75)};
+  padding-top: ${SizeUnit(0.5)};
+  padding-bottom: ${SizeUnit(0.5)};
+  line-height: 1;
+  cursor: pointer;
+  max-width: ${SizeUnit(10)}; // Beyond which it looks less and less button-like
+
```
nicks, line 21: can we throw in a `transition: background-color 300ms ease`?

web/src/Button.ts
```diff
@@ -0,0 +1,25 @@
+import { css } from "styled-components"
+import { Font, FontSize, Color, SizeUnit } from "./style-helpers"
+
+export const Button = css`
```
nicks, line 4: it seems odd that Button is a CSS mixin and not a component. maybe ButtonMixin?

web/src/ButtonInput.tsx
```diff
@@ -0,0 +1,20 @@
...
+  border: 0;
+  width: 100%;
+`
+
+type ButtonLinkProps = {
+  value: string
+  type: string
+}
+
+function ButtonLink(props: ButtonLinkProps) {
```
nicks, line 16: I got really confused that this file is called ButtonInput but the class is called ButtonLink

web/src/SidebarAccount.tsx
```diff
@@ -0,0 +1,268 @@
...
+  padding-top: ${SizeUnit(0.15)};
+  padding-bottom: ${SizeUnit(0.15)};
+
+  &:hover ${SidebarAccountLabel} {
+    opacity: 1;
+  }
+`
+let SidebarAccountIcon = styled(AccountIcon)`
+  fill: ${Color.blue};
+  transition: color ${AnimDuration.default} linear;
```
nicks, line 55: did you intend `transition: fill` here?

web/src/SidebarAccount.tsx
```diff
@@ -0,0 +1,268 @@
...
+
+  &:hover {
```
nicks, line 57: i think you want this to be
${SidebarAccountButton}:hover &
to pin the hover effect to the button?

web/src/SidebarAccount.tsx
```diff
@@ -0,0 +1,268 @@
...
+  margin-left: ${SizeUnit(0.15)};
+  margin-right: ${SizeUnit(0.15)};
+  z-index: ${ZIndex.SidebarMenu};
+  transform: scaleY(0);
+  opacity: 0;
+  transform-origin: top center;
+  transition: transform ${AnimDuration.short} ease,
+    opacity ${AnimDuration.short} linear;
+
+  &.isVisible {
```
nicks, line 89: `is-visible` for consistency with the other names (e.g., is-signedIn)

web/src/SidebarAccount.tsx
```diff
@@ -0,0 +1,268 @@
...
+`
+
+type SidebarAccountProps = {
+  tiltCloudUsername: string | null
+  tiltCloudSchemeHost: string
+  tiltCloudTeamID: string | null
+}
+
+function notifyTiltOfRegistration() {
+  let url = `${window.location.protocol}//${window.location.host}/api/user_started_tilt_cloud_registration`
```
nicks, line 176: is there some reason you need to re-build the url yourself here? I would have expected `/api/user_started_tilt_cloud_registration` to be equivalent

web/src/SidebarAccount.tsx
```diff
@@ -0,0 +1,268 @@
...
+  fetch(url, {
+    method: "POST",
+    headers: {
```
nicks, line 179: as far as i can tell, we don't send or expect any json here?

web/src/ButtonLink.tsx
```diff
@@ -0,0 +1,24 @@
...
+import styled from "styled-components"
+import { Button } from "./Button"
+
+let ButtonLinkRoot = styled.a`
+  ${Button}
+`
+
+type ButtonLinkProps = {
+  href: string
+  label: string
```
nicks, line 11: usually it's more idiomatic to use props.children and allow any react element?

web/src/ButtonLink.tsx
```diff
@@ -0,0 +1,24 @@
...
+  target?: string
+  rel?: string
+}
+
+function ButtonLink(props: ButtonLinkProps) {
+  return (
+    <ButtonLinkRoot href={props.href} {...props}>
```
nicks, line 18: the {...props} here feels like a bug waiting to happen, can we just list out target and rel explicitly?

web/src/HUD.tsx
```diff
@@ -404,10 +405,18 @@ class HUD extends Component<HudProps, HudState> {
     let resources = (view && view.resources) || []
     let sidebarItems = resources.map(res => new SidebarItem(res))
     let isSidebarClosed = !!this.state.isSidebarClosed
+    let tiltCloudUsername = (view && view.tiltCloudUsername) || null
```
nicks, line 408: i think we generally do this now with `view?.tiltCloudUsername`

web/src/SidebarAccount.tsx
```diff
@@ -0,0 +1,268 @@
...
+    color: ${Color.text};
+  }
+  small {
+    font-size: ${FontSize.small};
+  }
+  &.is-signedIn {
+    margin-top: ${SizeUnit(0.3)};
+    text-align: right;
+  }
+  &.is-signedOut {
```
nicks, line 134: ?

web/src/SidebarAccount.tsx
```diff
@@ -0,0 +1,268 @@
...
+  display: block;
+  transition: opacity ${AnimDuration.default} ease;
+  transition-delay: ${AnimDuration.short};
+`
+let MenuContentTeamName = styled.strong`
+  transition: background-color ${AnimDuration.default} ease;
+  transition-delay: ${AnimDuration.short};
+`
+let MenuContentTeam = styled.p`
+  &:hover {
```
nicks, line 149: fwiw, i usually try to keep these together with the component, i.e.

let MenuContentTeam = styled.p``
let MenuContentTeamInTiltfile = styled.div`
  ${MenuContentTeam}:hover & {
    ...
  }
`
```

so that it's easier to see all the styling for a component in one place

File: web/src/SidebarAccount.tsx [hunk 14]
```diff
@@ -0,0 +1,268 @@
...
+          onClick={() => toggleAccountMenu(!accountMenuIsVisible)}
+        >
+          <SidebarAccountLabel>Your Tilt Cloud status</SidebarAccountLabel>
+          <SidebarAccountIcon />
+        </SidebarAccountButton>
+      </SidebarAccountHeader>
+      <SidebarAccountMenu className={accountMenuIsVisible ? "isVisible" : ""}>
+        <SidebarAccountMenuHeader>
+          <SidebarAccountMenuLogo></SidebarAccountMenuLogo>
+          {!props.tiltCloudUsername && (
```
nicks, line 201: i feel like we should pull the conditionals out?

having them inline like this makes this very hard to read.

e.g.,

```
let optionalLearnMore = null
if (!props.tiltCloudUsername) {
  optionalLearnMore = ...
}

File: web/src/assets/svg/logo-Tilt-Cloud.svg [hunk 15]
```diff
@@ -0,0 +1 @@
...52.7904 15.96c-1.176-.588-2.856-1.036-4.536-1.036-3.78 0-5.936 2.744-5.936 6.692 0 3.836 2.016 6.608 5.936 6.608 1.932 0 3.472-.644 4.704-1.568l-.364-.616h-.308c-1.064.784-2.464 1.4-4.06 1.4-3.388 0-5.04-2.632-5.04-5.88s1.708-5.852 5.124-5.852c1.484 0 2.912.42 4.2 1.064h.28zm14.0775 10.892-.364-.588h-.308c-.896.756-2.072 1.176-3.248 1.176-1.54 0-2.436-.952-2.436-2.688v-16.688h-5.32v.756h4.452v15.82c0 2.156 1.008 3.584 3.304 3.584 1.484 0 2.772-.504 3.92-1.372zm14.0476-5.292c0-3.892-2.184-6.636-5.572-6.636s-5.572 2.744-5.572 6.636 2.128 6.664 5.544 6.664 5.6-2.772 5.6-6.664zm-.868 0c0 3.248-1.736 5.88-4.704 5.88s-4.704-2.632-4.704-5.88 1.792-5.852 4.704-5.852 4.704 2.604 4.704 5.852zm5.952-6.44v8.484c0 3.08 1.624 4.62 4.256 4.62 2.212 0 3.808-1.092 4.9-2.156l.308 1.96h.56v-12.908h-.868v10.192c-1.176 1.008-2.8 2.128-4.816 2.128-2.212 0-3.472-1.344-3.472-3.808v-8.512zm24.8235-7.056v8.568c-.784-.756-2.128-1.708-4.116-1.708-3.276 0-5.6 2.576-5.6 6.776 0 3.78 1.876 6.524 5.404 6.524 2.016 0 3.64-1.204 4.312-2.184l.308 1.988h.56v-19.964zm-4.032 7.644c1.68 0 2.996.84 4.032 1.764v7.784c-.84 1.008-2.268 2.184-4.2 2.184-3.136 0-4.648-2.632-4.648-5.768 0-3.668 2.072-5.964 4.816-5.964z"/></svg>
```
nicks, line 1: it seems really weird to have an SVG with an embedded PNG, which means that the SVG won't behave like you'd typically expect, tho maybe this needs to be factored out between tilt + tilt-cloud

web/src/HUD.tsx
```diff
@@ -404,10 +405,18 @@ class HUD extends Component<HudProps, HudState> {
...
+    let tiltCloudSchemeHost = (view && view.tiltCloudSchemeHost) || ""
+    let tiltCloudTeamID = (view && view.tiltCloudTeamID) || null
     let sidebarRoute = (t: ResourceView, props: RouteComponentProps<any>) => {
       let name = props.match.params.name
       return (
         <Sidebar isClosed={isSidebarClosed} toggleSidebar={this.toggleSidebar}>
+          <SidebarAccount
```
nicks, line 415: do we need to do something to prevent this from showing up in snapshots?

web/src/assets/svg/logo-Tilt-Cloud.svg
hyu: Good catch! This was totally unintentional.

web/src/HUD.tsx
hyu: Oh right! This shouldn't appear on a Snapshot.

web/src/SidebarAccount.tsx
hyu: Nope, this was just a bit of mindless copying. Will fix.

web/src/SidebarAccount.tsx
hyu: I don't actually understand this code very well, and pretty much grabbed it wholesale from ShareSnapshotModal. I chatted briefly with @landism about it, and my rough understanding is that when you start the registration process, you're taken offsite. So this `onSubmit` function simultaneously sends an API call to Tilt Cloud (/api/whoami, I think?) to tell it to make a `long_get` request. This way, as soon as you complete the registration process, Tilt will receive your new registration info.

But, I've thought very little about how this particular `notifyTiltOfRegistration` function should work. Should the method or headers be different that what I have here?

hyu:
There is a small unresolved discussion in this PR, but for the sake of momentum I'm merging this now. If needed, I'll create another PR to improve the `notifyTiltOfRegistration` function.

PR #3212, opened 2020-04-15
up: add --default-tui arg

landism:
### Problem

In Tilt 1.0, we'll hide the HUD by default and make it opt-in via `--hud`. We want to develop the new, hidden-HUD UI prior to actually launching it.

### Solution

Add a new `--default-tui` argument which is hidden and defaults to true. This is basically a feature flag to enable the 1.0 behavior prior to release. (not using a Tiltfile feature flag since this decision needs to be made before we load the Tiltfile)

### DIFF
--- a/internal/cli/up.go
+++ b/internal/cli/up.go
@@ -44,9 +44,13 @@ var logActionsFlag bool = false
 type upCmd struct {
 	watch                bool
 	traceTags            string
-	hud                  bool
 	fileName             string
 	outputSnapshotOnExit string
+
+	defaultTUI bool
+	hud        bool
+	// whether hud was explicitly set or just got the default value
+	hudFlagExplicitlySet bool
 }
 
 func (c *upCmd) register() *cobra.Command {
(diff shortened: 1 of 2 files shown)


internal/cli/up.go
```diff
@@ -83,9 +87,28 @@ In that case, see https://tilt.dev/user_config.html and/or comments in your Tilt
 	cmd.Flags().BoolVar(&noBrowser, "no-browser", false, "If true, web UI will not open on startup.")
 	cmd.Flags().StringVar(&c.outputSnapshotOnExit, "output-snapshot-on-exit", "", "If specified, Tilt will dump a snapshot of its state to the specified path when it exits")
 
+	// this is to test the new behavior before enabling it in Tilt 1.0
+	// https://app.clubhouse.io/windmill/epic/5549/make-tui-hard-to-find-in-tilt-1-0
+	cmd.Flags().BoolVar(&c.defaultTUI, "default-tui", true, "If false, we'll hide the TUI by default")
```
nicks, line 92: i'm surprised this isn't called --default-hud or similar, given that we don't really use the word "tui" elsewhere in the code, tho I guess if this is going away it doesn't matter

landism: changed!

PR #3214, opened 2020-04-16
cli: Clarify exiting Tilt does not stop resources

victorwuky:


victorwuky:
@maiamcc : Per discussion today, I think this clarifies that tilt down doesn't impact local resources right?

internal/cli/down.go
```diff
@@ -36,8 +36,8 @@ Deletes resources specified in the Tiltfile
 
 Namespaces are not deleted by default. Use --delete-namespaces to change that.
 
-Local resources (https://docs.tilt.dev/local_resource.html) that are still running (with serve_cmd)
-are not impacted, and need to stopped separately.
+Local resources (https://docs.tilt.dev/local_resource.html) that are still running
```
maiamcc, line 39: Mmm I think this line is vestigial; since long-running local resources are theoretically NEVER left alive when tilt exits, so this line might be more confusing than helpful.

so there's a type of local resource that ISN'T long-lived, and just runs a single command. it might be useful to clarify:
> `tilt down` has no effect on local resources; let us know if this is functionality you need.

internal/cli/up.go
```diff
@@ -67,6 +67,9 @@ By default:
 
 This default behavior does not apply if the Tiltfile uses config.parse or config.set_enabled_resources.
 In that case, see https://tilt.dev/user_config.html and/or comments in your Tiltfile
+
+When you exit Tilt (using Ctrl+C), resources previously started continue running.
```
maiamcc, line 71: Not quite. Should read something like:

> When you exit Tilt (using Ctrl+C), Kubernetes resources and Docker Compose resources continue running; you can use `tilt down` to delete these resources. Any long-running local resources--i.e. those using serve_cmd--are terminated when you exit Tilt.

internal/cli/down.go
victorwuky: Okay, I got this incorrect. So I just reverted. I.e. this PR only changes the tilt up doc.

victorwuky:
Thanks for clarifying @maiamcc . Do you mind checking again?

PR #3218, opened 2020-04-16
dockerprune: add a setting to keep the most recent N builds of a tag. Fixes https://github.com/windmilleng/tilt/issues/3211

nicks:
Hello @maiamcc,

Please review the following commits I made in branch nicks/ch6284/prune:

ff8774c73af994c3f591f346ab75e203a94b4db1 (2020-04-16 19:19:35 -0400)
dockerprune: add a setting to keep the most recent N builds of a tag. Fixes https://github.com/windmilleng/tilt/issues/3211

### DIFF
--- a/internal/engine/dockerprune/docker_pruner.go
+++ b/internal/engine/dockerprune/docker_pruner.go
@@ -3,6 +3,7 @@ package dockerprune
 import (
 	"context"
 	"fmt"
+	"sort"
 	"strings"
 	"time"
 
@@ -91,7 +92,7 @@ func (dp *DockerPruner) OnChange(ctx context.Context, st store.RStore) {
 
 	// Prune as soon after startup as we can (waiting until we've built SOMETHING)
 	if dp.lastPruneTime.IsZero() && curBuildCount > 0 {
-		dp.PruneAndRecordState(ctx, settings.MaxAge, imgSelectors, curBuildCount)
+		dp.PruneAndRecordState(ctx, settings.MaxAge, settings.KeepRecent, imgSelectors, curBuildCount)
 		return
 	}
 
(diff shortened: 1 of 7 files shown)


internal/engine/dockerprune/docker_pruner.go
```diff
@@ -203,25 +202,85 @@ func (dp *DockerPruner) deleteOldImages(ctx context.Context, maxAge time.Duratio
...
-			toDelete[inspect.ID] = uint64(inspect.Size)
+			result = append(result, inspect)
+		}
+	}
+	return result
+}
+
+// Return all image objects that aren't in the N
+// most recently used for each tag.
+func (dp *DockerPruner) keepMostRecentInspects(ctx context.Context, inspects []types.ImageInspect, keepRecent int, selectors []container.RefSelector) []types.ImageInspect {
```
maiamcc, line 213: This function signature is confusing because the inspects you're "keeping" are actually the ones you're NOT returning. Maybe frame this is as `FilterOutNMostRecentInspects`?

internal/engine/dockerprune/docker_pruner.go
```diff
@@ -203,25 +202,85 @@ func (dp *DockerPruner) deleteOldImages(ctx context.Context, maxAge time.Duratio
...
+	// First, sort the images in order from most recent to least recent.
+	recentFirst := append([]types.ImageInspect{}, inspects...)
+	sort.SliceStable(recentFirst, func(i, j int) bool {
+		return recentFirst[i].Metadata.LastTagTime.After(recentFirst[j].Metadata.LastTagTime)
```
maiamcc, line 217: not a big deal (I know it wasn't in the original) but should we add a note why we're using `LastTagTime` instead of `CreatedAt`? Makes sense once i thought of it but wasn't immediately obvious.

internal/engine/dockerprune/docker_pruner.go
nicks: done

internal/engine/dockerprune/docker_pruner.go
nicks: done

maiamcc:
BTW did you update the docs for this?

nicks:
nope, it's not out yet!

PR #3219, opened 2020-04-17
Remove token to sign out of Tilt

victorwuky:
Screen Shot 2020-04-16 at 10 04 00 PM


victorwuky:
During epics meeting today, we talked about this. It's really annoying because I haven't been able to test these signed-out flows. And I frantically tried using incognito and installing/uninstalling new browsers which didn't work. 

@landism mentioned removing the token. I think I found the location and this works for me. cc @hyu as well.

victorwuky:
> is there a github issue anywhere for a way for people to sign out in the UX? I'm pretty confident people will want to do this

@nicks : I don't think so. I created one is there a github issue anywhere for a way for people to sign out in the UX? I'm pretty confident people will want to do this

PR #3247, opened 2020-04-28
dcwatch: If a container exits with status code 0, that's OK

nicks:
Hello @jazzdan, @maiamcc,

Please review the following commits I made in branch nicks/ch6586:

603edcd21ef20df4fb614d7c7c2a255133fa39c7 (2020-04-28 10:23:19 -0400)
dcwatch: If a container exits with status code 0, that's OK
Rebuild docker-compose status inference on top of ContainerState instead of
docker-compose events.  ContainerState has the information we need to fix this
problem, and is likely to be a more reliable API.

Fixes https://github.com/windmilleng/tilt/issues/3230


internal/dockercompose/state.go
```diff
@@ -66,28 +36,43 @@ type State struct {
...
-	runtimeStatus, ok := runtimeStatusMap[string(s.Status)]
-	if !ok {
+	if s.ContainerState.Error != "" || s.ContainerState.ExitCode != 0 {
 		return model.RuntimeStatusError
 	}
-	return runtimeStatus
+	if s.ContainerState.Running ||
+		// Status strings taken from comments on:
+		// https://godoc.org/github.com/docker/docker/api/types#ContainerState
+		s.ContainerState.Status == "running" ||
```
jazzdan, line 45: Should we make these strings constants?

nicks: good idea, done!

PR #3259, opened 2020-04-29
tiltfile: add a bunch of os.path functions, and fix bugs in os.path.realpath

nicks:
Hello @landism, @jazzdan,

Please review the following commits I made in branch nicks/ch6674/path:

308614ed38fc990021d2aa73a42b1f2d4a16399c (2020-04-29 18:17:27 -0400)
tiltfile: add a bunch of os.path functions, and fix bugs in os.path.realpath
Fixes https://github.com/windmilleng/tilt/issues/3173
Fixes https://github.com/windmilleng/tilt/issues/3086

### DIFF
--- a/internal/tiltfile/os/os.go
+++ b/internal/tiltfile/os/os.go
@@ -1,6 +1,7 @@
 package os
 
 import (
+	"fmt"
 	"os"
 	"path/filepath"
 	"runtime"
@@ -8,6 +9,7 @@ import (
 
 	"go.starlark.net/starlark"
 
+	"github.com/windmilleng/tilt/internal/tiltfile/io"
 	"github.com/windmilleng/tilt/internal/tiltfile/starkit"
 )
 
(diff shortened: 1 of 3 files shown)


internal/tiltfile/os/os.go
```diff
@@ -83,18 +104,60 @@ func cwd(t *starlark.Thread, fn *starlark.Builtin, args starlark.Tuple, kwargs [
...
+func basename(t *starlark.Thread, path string) (starlark.Value, error) {
+	return starlark.String(filepath.Base(path)), nil
+}
+
+func dirname(t *starlark.Thread, path string) (starlark.Value, error) {
+	return starlark.String(filepath.Dir(path)), nil
+}
+
+func exists(t *starlark.Thread, path string) (starlark.Value, error) {
+	_, err := os.Stat(starkit.AbsPath(t, path))
```
landism, line 136: should this watch the path?

nicks: I'm...honestly not sure. I could go either way on this. what do you think?

jazzdan: I think it should. Imagine this:

```python
if os.exists('path/to/repo'):
  include('path/to/repo/Tiltfile')
```

I think that sort of use case could be important for micorepo setups

landism: I think in general we want the Tiltfile to reexecute when any of its filedeps change.

I guess it'd be annoying if someone is doing something like:
```
if not os.exists('foo.txt'):
  # create foo.txt
```
since it'd execute twice. I don't have a good feeling for the use cases for this or whether this would be a common thing. Seems fine as-is and if someone wants the other behavior we can tell them to `watch_file`.

jazzdan: Asking people to manually watch_file sgtm too

nicks: ok, i made it watch by default, and we can see what happens!

PR #3268, opened 2020-04-30
web: tilt cloud signin link should open in a new window, not take you away from the web ui

maiamcc:
Hello @nicks, @hyu,

Please review the following commits I made in branch maiamcc/tilt-cloud-sign-in-link:

81b153e3796cce1838e9cb5b6f9e3975ec082df3 (2020-04-30 17:44:57 -0400)
web: tilt cloud signin link should open in a new window, not take you away from the web ui

### DIFF
--- a/web/src/SidebarAccount.tsx
+++ b/web/src/SidebarAccount.tsx
@@ -252,7 +252,8 @@ function SidebarAccount(props: SidebarAccountProps) {
       <MenuContentSignInLink>
         Or{" "}
         <strong>
-          <a href={props.tiltCloudSchemeHost} target="_blank">Sign In</a>
+          <a href={props.tiltCloudSchemeHost} target="_blank"
+             rel="noopener noreferrer nofollow">Sign In</a>
         </strong>{" "}
         to your account.
       </MenuContentSignInLink>
--- a/web/src/SidebarAccount.tsx
+++ b/web/src/SidebarAccount.tsx
@@ -252,7 +252,7 @@ function SidebarAccount(props: SidebarAccountProps) {
       <MenuContentSignInLink>
         Or{" "}
         <strong>
-          <a href={props.tiltCloudSchemeHost}>Sign In</a>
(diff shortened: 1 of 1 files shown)


web/src/SidebarAccount.tsx
```diff
@@ -252,7 +252,7 @@ function SidebarAccount(props: SidebarAccountProps) {
       <MenuContentSignInLink>
         Or{" "}
         <strong>
-          <a href={props.tiltCloudSchemeHost}>Sign In</a>
+          <a href={props.tiltCloudSchemeHost} target="_blank">Sign In</a>
```
nicks, line 255: at the risk of derailing this PR, i don't understand what this link does? It says "Sign In" but it's just a link to Tilt Cloud.

@hyu is it possible that this is just vestigial, and you meant to delete it when you added the real sign-in flow above?

maiamcc:
Ah yeah that's documented here: https://app.clubhouse.io/windmill/story/6734/web-ui-tilt-cloud-status-sign-in-link-doesn-t-do-anything

hyu:
The reason I put that extra "Sign In" link there is to allow our main CTA to be loudly exhorting you to sign up for Tilt Cloud, on the presumption that you don't yet have an account. But the extra "sign in" adds an escape hatch for people who already have an account, and just want to sign in quickly. 

At the time I added this UI, it felt out of scope to build a fast way to sign in. So I just hastily added a link to Tilt Cloud, where you can see another "Sign In" link. So yes, it's vestigial in that it's the laziest and worst possible way that you _could_ sign in. 😬

Ideally, it would be way more automated. Either what Maia wrote in the story: 
~~~
...navigate back to web UI, maybe refresh, and see "you are signed in" in the modal in the top right corner
~~~

Or...somehow sign in without visibly going to the Tilt Cloud UI at all?

As for this actual PR, it's definitely an improvement not to open Tilt Cloud in the same window as Tilt. That was a mistake!

PR #3272, opened 2020-05-01
cloud: check status even when snapshots are disabled

landism:


landism:
> LGTM but i wonder if we should provide another way to disable this check. Folks might get worried if Tilt is phoning home and they aren't using any cloud features.

how about if team id is empty and update checks are disabled?

landism:
> how about if team id is empty and update checks are disabled?

oh, except we still need the initial check to see if you're registered.

landism:
> LGTM but i wonder if we should provide another way to disable this check. Folks might get worried if Tilt is phoning home and they aren't using any cloud features.

Thanks for calling this out. We've discussed this in CH6731, CH6795, and decided to not block this merge on it.

PR #3300, opened 2020-05-07
tiltfile: add objects parameter to k8s_resource

jazzdan:


internal/tiltfile/k8s.go
```diff
@@ -66,6 +66,8 @@ type k8sResourceOptions struct {
 	triggerMode       triggerMode
 	tiltfilePosition  syntax.Position
 	resourceDeps      []string
+	// TODO(dmiller): this should be a different type once it has been parsed as a valid identifier
```
maiamcc, line 69: docstring for this field pls

jazzdan:
@maiamcc fwiw I plan on making workload optional last, after all the other changes are in.

internal/tiltfile/k8s.go
jazzdan: Hmm none of the other fields are documented, unless you count https://docs.tilt.dev/api.html#api.k8s_resource as documentation. Should this one be an exception?

dbentley:
When you say builtin func that makes a uid, I imagine something like 

```
k8s.oid(name`'foo')
k8s.oid(name='foo', kind='deployment')
k8s.oid(name`'foo', kind='deployment', namespace='karen-dev')
```

This is nice because you don't have to figure out which component goes where because it's self-documenting.

Am I understanding your point correctly?

Following up, is there some way you see that is a better UX for simple cases (but may lack some of the customizability)? I'd love to hear it fwiw.

maiamcc:
Whoops @dbentley this fell off my radar. Yeah, in re: a UID builtin, something like that is exactly what I was imagining :)

I really liked @jazzdan's original proposal of labels--many of your k8s objects will have them already, and they're easy to reason about/add to YAML, and easy to use to add groups of related entities all at once. Especially if you've got a bunch of different-but-related objects to add, specifying UIDs for all of them are going to be a pain. My hunch is that UIDs will be valuable for 20% cases but labels will suffice for 80% and be much easier to reason about.

Unrelated and just riffing: this is making me think that we should make YAML manipulation easier with some builtins or maybe extensions that go from a YAML dict --> collection of identifying info, probs name+namespace+group+version+kind. I imagine something like:
```
yamls = read_yaml('k8s.yaml')
for entity in yamls:
    info = get_info(entity)
    if info.name == "foo" && info.kind in ["service", "secret"]:
        do_thing(entity)
        # or even:
        add_labels(entity, {"app": "frontend", "stuff": "things"})
```

PR #3302, opened 2020-05-07
make: fix release

landism:


landism:
> Could you add error handling though? the script should barf if version == "". (also mayyybe if no token is found? that seems less likely, but fairly possible on a developer's machine.)

done, good call

> but most helpful thing would be to print the response body if we get a non-200 error code.

That's what `curl -S` does, but the response body wasn't useful enough. Added some more language for this case.

Issue #2608, opened 2019-11-27
feature request: load web assets from offline


nicks:
Currently, Tilt loads its web assets from a publicly readable google cloud storage bucket at
https://storage.googleapis.com/tilt-static-assets/

This is much easier for us to manage operationally. But it means that if you're offline, or have a firewall that's blocking this site for some reason, the Tilt web UI will fail to load.

One or two people in the #tilt channel have asked how hard it would be to make this work offline. I'm not totally sure, and haven't spent much time looking into using something like Packr to do this

nicks:
(Current workaround is to only use Tilt when offline, and unblock the above bucket in your firewall)

victorwuky:
A workaround is installing and running Tilt from source, and setting `web-mode=local` to serve local assets from your local system. See https://github.com/windmilleng/tilt/blob/master/CONTRIBUTING.md#web-ui

szucsitg:
We'd love this as well

victorwuky:
@szucsitg feel free to upvote (👍 ) the issue. We use to track interest.

CorentinWanteeed:
Hello ✋,

I am also interested about this feature. :)

Tchoupinax:
Hello ✋,

I am also interested about this feature. :)

nicks:
I think the Go team is working on a much more holistic solution to this problem https://github.com/golang/go/issues/35950, so I'm eagerly watching what happens there

samkon:
It can be all about DNS configuration you're using as I use google dns. Reset your DNS configurations and try it again.

PR #3304, opened 2020-05-07
no-check option for asset uploading, default for release

maiamcc:
Hello @nicks, @jazzdan,

Please review the following commits I made in branch upload-assets-messaging:

6d5a606431cdd593442c0f2435ef346a20221003 (2020-05-07 18:24:25 -0400)
no-check option for asset uploading, default for release

e6e313be7eb172c44a674491f535581dd7b07062 (2020-05-07 17:34:35 -0400)
scripts: upload-assets.py tells you what to do in case of conflict


maiamcc:
just running `gsutil cp` SHOULD overwrite the existing objects but I couldn't make that behavior happen so i went with this instead (`--no_check` doesn't guarantee overwriting, it just doesn't check). That might be unnecessary word mincing tho 🤷‍♀️

scripts/upload-assets.py
```diff
@@ -28,12 +28,17 @@
...
 url = dir_url + "index.html"
 print("Uploading to %s" % dir_url)
-status = subprocess.call([
-  "gsutil", "stat", "gs://tilt-static-assets/%s/index.html" % version
-])
-if status == 0:
-  print("Error: Files already exist: %s" % url)
-  sys.exit(1)
+
+if not args.no_check:
```
nicks, line 32: is there a risk this will get us into a weird state where two releases are on top of each other? do we need to delete the old folder if it exists?

scripts/release.sh
```diff
@@ -13,7 +13,7 @@ fi
 DIR=$(dirname "$0")
 cd "$DIR/.."
 
-./scripts/upload-assets.py latest
+./scripts/upload-assets.py latest --force
```
maiamcc, line 16: I'm... honestly not sure if we want this or if we want to allow users to pipe the `--force` arg thru via `make release` / have a separate invocation `upload-assets.py latest --clean` that JUST deletes the existing bucket and then user can run `make release` without the `--force` flag. thoughts?

maiamcc:
@nicks took another swing at this cuz it was bothering me this morning!

maiamcc:
@nicks @landism ping! (This is pretty different from the original version so wants re-review)

scripts/release.sh
landism: I can come up with some pros and cons but they seem minor enough and the impact of this decision small enough that it's probably not worth our time.

landism:
Alternate proposal, if you're interested:

change the asset server to always fetch assets by sha and take asset uploads out of the release process

landism (approved):

scripts/release.sh
landism: Actually, I'm in favor of the `--clean` option. As Nick has been saying, we should move the release process to CI, and the `--clean` option allows us to handle that error separately on our laptops.

nicks:
alternative proposal #2 - bake assets into the binary, cf https://go.googlesource.com/proposal/+/master/design/draft-embed.md, https://github.com/tilt-dev/tilt/issues/2608

nicks (approved):

nicks:
actually, nm, i guess embedded assets wouldn't fix the whole problem, since we'd still have to do asset releases for snapshots

scripts/release.sh
maiamcc: done!

PR #3308, opened 2020-05-08
Update upgrade docs as soon as possible

victorwuky:


victorwuky:
@maiamcc : I see https://github.com/windmilleng/tilt.build/pull/571/files hasn't been merged yet, but at least [one person in Slack](https://kubernetes.slack.com/archives/CESBL84MV/p1588944318467100) seems to have already seen 0.13.5, probably due to upgrade nudge, so I'm guessing they had to manually know to change 0.13.4 to 0.13.5 when copying pasting the upgrade command from https://docs.tilt.dev/upgrade.html.

Ideally https://docs.tilt.dev/upgrade.html can be updated automatically whenever there's a new release of windmilleng/tilt. Should we scope a story to do that soon-ish?

CONTRIBUTING.md
```diff
@@ -283,6 +283,10 @@ update the [install](https://github.com/windmilleng/tilt.build/tree/master/docs/
 the [default dev version](internal/cli/build.go),
 and the [installer version](scripts/install.sh).
 
+The [upgrade docs](https://github.com/windmilleng/tilt.build/blob/master/docs/upgrade.md)
```
maiamcc, line 286: Fwiw this looks like an additional checklist item, which it isn't -- maybe put the "asap-ness" clause in the above paragraph.

maiamcc:
Thinking more, maybe the most useful thing here would be specifying that Tilters don't need to wait on review for version bump PRs.

victorwuky:
> Thinking more, maybe the most useful thing here would be specifying that Tilters don't need to wait on review for version bump PRs.

@maiamcc : Sure. I'm always supportive of being more prescriptive if it is a very operationally focused. cc @nicks as well.

See latest update.

nicks:
sgtm

Issue #3307, opened 2020-05-08
Cannot sign in (via GitHub) in Tilt, when you already signed in to Tilt Cloud 


victorwuky:
See https://kubernetes.slack.com/archives/CESBL84MV/p1588944318467100

<img width="887" alt="Screen Shot 2020-05-08 at 9 39 48 AM" src="https://user-images.githubusercontent.com/10860550/81411112-e58d4880-910f-11ea-9921-298f5d9eb759.png">

maiamcc:
FWIW I don't think this is a bug so much as misleading copy--if we took out the "sign in" link and made the big blue button say "Sign Up / Connect Your Account",  I think everything would work as people expected.

andymartin-sch:
Ah, I totally agree. I actually hadn't tried clicking the "Sign Up via GitHub" button at all and didn't notice that it's already set up to do the register_token-ing.

maiamcc:
Have clarified the copy a bit, hope that helps! We're still working on improving this flow, but at the very least things shouldn't be _actively confusing_ now.

PR #3311, opened 2020-05-08
web: clearer copy on tilt cloud signin/link modal [ch6917]

maiamcc:
Hello @hyu, @vw, @nicks,

Please review the following commits I made in branch maiamcc/link-to-tc:

259316890e04682948c1c0dff41f62504ff04d54 (2020-05-08 16:21:30 -0400)
web: clearer copy on tilt cloud signin/link modal [ch6917]

addresses #3307

### DIFF
--- a/web/src/SidebarAccount.tsx
+++ b/web/src/SidebarAccount.tsx
@@ -246,7 +246,7 @@ function SidebarAccount(props: SidebarAccountProps) {
         <input name="token" type="hidden" value={cookies.get("Tilt-Token")} />
         <MenuContentButtonSignUp
           type="submit"
-          value="Link Local Tilt to Tilt Cloud"
+          value="Link Tilt to Tilt Cloud"
         />
       </form>
     </SidebarAccountMenuContent>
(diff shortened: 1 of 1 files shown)


web/src/SidebarAccount.tsx
```diff
@@ -247,21 +244,11 @@ function SidebarAccount(props: SidebarAccountProps) {
         onSubmit={notifyTiltOfRegistration}
       >
         <input name="token" type="hidden" value={cookies.get("Tilt-Token")} />
-        <MenuContentButtonSignUp type="submit" value="Sign Up via GitHub" />
+        <MenuContentButtonSignUp
+          type="submit"
+          value="Link Local Tilt to Tilt Cloud"
```
victorwuky, line 249:
```suggestion
          value="Link Tilt to Tilt Cloud"
```

Let’s not use “Local” since it’s not elsewhere in the product. And customers I think associate Tilt with the installed binary experience already. We have Git and GitHub (I know there are different orgs behind these products) so Tilt and Tilt Cloud seem to work well I think.

maiamcc: Sure! I agree with Han's comment in the ticket about sort of wanting a better/less confusing line of copy for this button, but I like that this is concise and it's 100% better than what we have on master today.

maiamcc:
Oh whoops I forgot to merge this 😅

PR #3322, opened 2020-05-11
tiltfile: implement objects parameter

jazzdan:


internal/tiltfile/tiltfile_state.go
```diff
@@ -616,6 +632,29 @@ func (s *tiltfileState) assembleK8sV2() error {
...
+					err = s.addEntityToResourceAndRemoveFromUnresourced(u, r)
+					if err != nil {
+						return err
+					}
+				} else if u, ok := fullNamesToEntities[o]; ok {
+					err = s.addEntityToResourceAndRemoveFromUnresourced(u, r)
+					if err != nil {
+						return err
+					}
+				} else if err != nil {
```
landism, line 650: IIUC, err is always nil here

internal/tiltfile/tiltfile_state.go
```diff
@@ -633,6 +672,20 @@ func (s *tiltfileState) assembleK8sV2() error {
...
+	for i, ur := range s.k8sUnresourced {
+		if ur == e {
+			// delete from unresourced
+			s.k8sUnresourced = append(s.k8sUnresourced[:i], s.k8sUnresourced[i+1:]...)
+			return nil
+
+		}
+	}
+
+	return fmt.Errorf("Unable to find %s in unresourced YAML", e.Name())
```
landism, line 686: if we hit this, it's a bug, right?
maybe worth specifying:
1. that it's an internal error
2. the name of the resource on which we encountered it

internal/tiltfile/tiltfile_state.go
```diff
@@ -616,6 +632,29 @@ func (s *tiltfileState) assembleK8sV2() error {
 				r.name = opts.newName
 				s.k8sByName[r.name] = r
 			}
+
+			// if the resource specifies additional objects, iterate through the unresourced k8s entities
```
landism, line 636: Are these cases ommitted or TODO?
```
# error cases
k8s_resource('foo', objects=['foo:deployment']) # can't include workloads as objects

# can't include an object twice
k8s_resource('foo', objects=['foo_secrets'])
k8s_resource('bar', objects=['foo_secrets'])

# If a name isn't unique among the resources Tilt is aware of that is also an error
k8s_resource('foo', objects=['foo']) # foo is both a deployment and a secret, for example
```

internal/tiltfile/tiltfile_state.go
```diff
@@ -602,6 +602,22 @@ func (s *tiltfileState) assembleK8sV2() error {
 		return err
 	}
 
+	// TODO(dmiller): support all three ways of specifying a name:
```
landism, line 605: IIUC, if we use something like [`k8sObjectSelector`](https://github.com/windmilleng/tilt/blob/ea3019f1cea5c2d6a03173707377c41db0ac47db/internal/tiltfile/tiltfile_state.go#L1210) (which maybe you don't want to use that specifically because ugh regexes), we don't really need these concepts or to consider these cases separately, and we can cover all the cases with:
1. For each object specified on a resource, construct a selector with as much data is provided.
2. If there is not exactly one k8s object that matches the selector, return an error.

internal/tiltfile/tiltfile_state.go
landism: We'll hit this if two resources try to include the same object, right? This seems like a confusing error to get in that case.

internal/tiltfile/tiltfile_state.go
```diff
@@ -602,6 +602,22 @@ func (s *tiltfileState) assembleK8sV2() error {
...
+	// 1. Short name (foo) (done)
```
landism, line 606: The CH story defines short name is the shortest unique fragment, so it should definitely be covered by the unique fragment case, right? And that also means we can't calculate short names by calling `UniqueNames` with just tha tone entity - we need to pass them all?

internal/tiltfile/tiltfile_state.go
jazzdan: Ah this is much better, thank you for pointing me to that

internal/tiltfile/tiltfile_state.go
```diff
@@ -616,6 +616,34 @@ func (s *tiltfileState) assembleK8sV2() error {
 				r.name = opts.newName
 				s.k8sByName[r.name] = r
 			}
+
+			selectors := make([]k8sObjectSelector, len(opts.objects))
+			for i, o := range opts.objects {
+				s, err := selectorFromString(o)
+				if err != nil {
+					return errors.Wrap(err, "Error making selector from string")
```
maiamcc, line 624: maybe include the string in the error message?

internal/tiltfile/tiltfile_state.go
```diff
@@ -633,6 +661,37 @@ func (s *tiltfileState) assembleK8sV2() error {
 	return nil
 }
 
+func selectorFromString(s string) (k8sObjectSelector, error) {
```
maiamcc, line 664: might be helpful to note the schema in a comment so ppl reading this code don't have to re-derive it

internal/tiltfile/tiltfile_state.go
```diff
@@ -616,6 +616,34 @@ func (s *tiltfileState) assembleK8sV2() error {
...
+			for _, selector := range selectors {
+				matches := []k8s.K8sEntity{}
+				for _, e := range s.k8sUnresourced {
+					if selector.matches(e) {
+						matches = append(matches, e)
+					}
+				}
+				numMatches := len(matches)
+				if numMatches == 0 {
+					return fmt.Errorf("Unable to find any unesourced matches for selector %+v", selector)
```
maiamcc, line 638: If this is going to be a user-facing error message, i think in the UI etc. we don't use the language "unresourced" -- either say "uncategorized" or (ideally) be more specific: "selectors passed to k8s_resource.objects must match exactly one k8s entity not already part of a Tilt resource" or something

internal/tiltfile/tiltfile_state.go
```diff
@@ -633,6 +661,37 @@ func (s *tiltfileState) assembleK8sV2() error {
...
+	r.entities = append(r.entities, e)
+	for i, ur := range s.k8sUnresourced {
+		if ur == e {
+			// delete from unresourced
+			s.k8sUnresourced = append(s.k8sUnresourced[:i], s.k8sUnresourced[i+1:]...)
+			return
+		}
+	}
+
+	panic("Unable to find entity in unresourced YAML after checking that it was there. This should never happen")
```
maiamcc, line 692: 😱

internal/tiltfile/tiltfile_test.go
```diff
@@ -4381,23 +4381,81 @@ allow_k8s_contexts("hello")
...
+	f.yaml("namespace.yaml", namespace("bar"))
+
+	f.file("Tiltfile", `
+docker_build('gcr.io/foo', 'foo')
+k8s_yaml('foo.yaml')
+k8s_yaml('secret.yaml')
+k8s_resource('foo', objects=['bar', 'bar:secret:default'])
+`)
+
+	f.loadErrString("Unable to find any unresourced matches for selector bar:secret:default. All YAML has already been resourced")
```
landism, line 4440: I think this error makes sense if someone has a mental model of resource assembly, but is kind of confusing when we present k8s_resource as declarative, not imperative. Also, "All YAML has already been resourced" seems hard to understand.

It'd take some extra bookkeeping, but I think the ideal error message here would be something like "selector %s on resource %s and selector %s on resource %s both match object %s. An object can only be claimed by one selector."

A downside of that is it's pretty special-cased to two selectors matching an object, and doesn't cover two other cases, and we'd need to handle those separately:
* no object exists at all that matches that selector
* the object that matches that selector was already automatically put into a resource (e.g., it's a workload, or a service that matches a workload)

I'd still kind of prefer to generate that error message, but if it's a pain to do all those cases, I think generating something like this would be relatively cheap and cover all cases:

`"Unable to find any uncategorized objects matching selector %s. Did find matching categorized objects %v"`
(where the format of %v is TBD. If there's only one, something like `"bar:secret:default, belonging to resource %s"`. If there are multiple, maybe it should be multiple lines?)

(which will hopefully be easier if we're matching against categorized objects, per previous comment)

internal/tiltfile/tiltfile_state.go
```diff
@@ -616,6 +616,41 @@ func (s *tiltfileState) assembleK8sV2() error {
...
+				}
+				selectors[i] = s
+			}
+
+			for i, selector := range selectors {
+				if len(s.k8sUnresourced) == 0 {
+					return fmt.Errorf("Unable to find any unresourced matches for selector %s. All YAML has already been resourced", opts.objects[i])
+				}
+				matches := []k8s.K8sEntity{}
+				for _, e := range s.k8sUnresourced {
```
landism, line 634: IIUC, we also need to match against resourced objects or else this will behave confusingly.

e.g.:

we have foo:namespace and foo:service

resource1 adds the object "foo:namespace". there's only one match, so this is fine.
resource2 adds the object "foo". because "foo:namespace" is no longer uncategorized, this succeeds.

1. Per the design, this should fail because "foo" is ambiguous.
2. The result depends on whether we evaluate resource1 or resource2 first (if we evaluate resource2 first, it'll fail because "foo" will still be ambiguous), which is theoretically nondeterministic.

Similarly, if we have a foo:deployment and foo:secret and then k8s_resource('foo', objects=['foo']), it'll work, because foo:deployment isn't unresourced.

It seems easier to describe if we make the checks independent of the current state of resourcing.

jazzdan: Yeah good call.

jazzdan:
@landism I'd like to propose that we add in the better error messaging in a subsequent PR (immediately subsequent). This one is getting a bit big and gnarly already but I think the major issues have been addressed.

internal/tiltfile/tiltfile_state.go
```diff
@@ -616,6 +616,49 @@ func (s *tiltfileState) assembleK8sV2() error {
...
+						namesOfAvailableUnresourcedYAML[i] = selectorStringFromK8sEntity(e)
+					}
+					return fmt.Errorf("Unable to find any unresourced matches for selector %s. Available unresourced YAML: %s", opts.objects[i], strings.Join(namesOfAvailableUnresourcedYAML, ", "))
+				}
+				if numMatches > 1 {
+					return fmt.Errorf("Found multiple (%d) matches for selector %+v: %v", numMatches, selector, matches)
+				}
+
+				e := matches[0]
+				s.addEntityToResourceAndRemoveFromUnresourced(e, r)
```
landism, line 660: If I'm reading this correctly, we'll end up panicking if the one match is from s.k8s rather than s.k8sUnresourced. I think tests for these cases would be useful:
* one resource specifies another resource's workload
* two resources specify the same object

(I think TestK8sResourceObjectsCantIncludeSameObjectTwice will also panic if you add another object to the unresourced yaml - it's only avoiding the panic because it's hitting the `if len(s.k8sUnresourced) == 0 {` check)

something like this might be more straightforward (or maybe it only feels more straightforward because pseudocode doesn't have to deal with go's verbosity?)
```
// figure out what resources claim what objects, validating constraints as we go
for r in resources:
  for sel in r.selectors:
    resourcedMatches = filter(resourcedObjects, sel)
    if len(resourcedMatches) > 0:
      error
    unresourcedMatches = filter(resourcedObjects, sel)
    if len(resourcedMatches) != 1:
      error
    if claims[unresourcedMatches[0]]:
      error // can print claims[unresourcedMatches[0]] and r+sel as the conflicting claims on the object
    claims[unresourcedMatches[0]] = {Resource: r, Selector: sel}

// at this point, we've matched all selectors and know we followed the rules, so just assign objects to resources
for obj, claim in claims:
  // assign obj to claim.Resource
```

maiamcc: FWIW I wrote up [this matrix of possibilities](https://docs.google.com/spreadsheets/d/1arxzEk_83LI7F4W8NFSL2mbmNiyho0xUrrT5TWeciOg/edit#gid=0): does that seem right?

jazzdan: FYI there's already a test case for two resources that specify the same workload, that's handled elsewhere.

I just added a test case for two resources specifying the same object.

jazzdan: > (I think TestK8sResourceObjectsCantIncludeSameObjectTwice will also panic if you add another object to the unresourced yaml - it's only avoiding the panic because it's hitting the if len(s.k8sUnresourced) == 0 { check)

Yep good call

internal/tiltfile/tiltfile_state.go
```diff
@@ -633,6 +686,54 @@ func (s *tiltfileState) assembleK8sV2() error {
 	return nil
 }
 
+// format is <name:required>:<kind:optional>:<namespace:optional>
+func selectorFromString(s string) (k8sObjectSelector, error) {
```
maiamcc, line 690: inconsistency: this format and the code below omit the "apiVersion" arg, while the docs and the error message in this func ("between 1 and 4 parts") indicate that `apiVersion` is an acceptable arg.

internal/tiltfile/tiltfile_state.go
```diff
@@ -616,12 +622,59 @@ func (s *tiltfileState) assembleK8sV2() error {
...
+				for _, r := range s.k8s {
+					resourcedMatches := filterEntitiesBySelector(r.entities, sel)
+					if len(resourcedMatches) > 0 {
+						return fmt.Errorf("object %s already belongs to resource %s", opts.objects[i], r.name)
+					}
+				}
+
+				unresourcedMatches := filterEntitiesBySelector(s.k8sUnresourced, sel)
+				if len(unresourcedMatches) != 1 {
+					errMsg := fmt.Sprintf("Found %d matches for %s in remaining YAML. Object must match exactly 1 entity", len(unresourcedMatches), opts.objects[i])
```
jazzdan, line 646: Maybe "selector is not unique, it matches each of these objects" but print full name

internal/tiltfile/tiltfile_state.go
```diff
@@ -616,12 +622,59 @@ func (s *tiltfileState) assembleK8sV2() error {
...
+				s, err := selectorFromString(o)
+				if err != nil {
+					return errors.Wrapf(err, "Error making selector from string '%s'", o)
+				}
+				selectors[i] = s
+			}
+
+			claims := make(map[k8s.K8sEntity]claimPair)
+			for i, sel := range selectors {
+				for _, r := range s.k8s {
```
maiamcc, line 637: hmmm nested for loop... could you map resource:[]entities all at once instead?

internal/tiltfile/tiltfile_state.go
maiamcc: (note you may have to change `selectorStringFromK8sEntity` as well)

internal/tiltfile/tiltfile_state.go
```diff
@@ -616,12 +622,59 @@ func (s *tiltfileState) assembleK8sV2() error {
...
+						}
+						errMsg = fmt.Sprintf("%s. Available YAML: %s", errMsg, strings.Join(namesOfAvailableUnresourcedYAML, ", "))
+					}
+
+					return errors.New(errMsg)
+				}
+
+				entity := unresourcedMatches[0]
+				if existingClaim, ok := claims[entity]; ok {
+					return fmt.Errorf("%s tried to claim %s but it was already claimed by %s via %s", opts.objects[i], entity.Name(), existingClaim.resource.name, existingClaim.selectorString)
```
jazzdan, line 662: This object (fullname) was passed to both k8s_resource foo (as X) and k8s_resource foo2 (as Y).

internal/tiltfile/tiltfile_state.go
```diff
@@ -1192,17 +1293,51 @@ const (
 
 // A selector matches an entity if all non-empty selector fields match the corresponding entity fields
 type k8sObjectSelector struct {
-	apiVersion *regexp.Regexp
-	kind       *regexp.Regexp
+	apiVersion       *regexp.Regexp
```
maiamcc, line 1296: this seems convoluted, would you consider something like https://github.com/windmilleng/tilt/blob/c2b2df88de3777eed5f1bb9f54b5c555707c8b42/internal/container/selector.go#L9 so you only have to store each bit of information once?

internal/tiltfile/tiltfile_state.go
```diff
@@ -633,6 +686,54 @@ func (s *tiltfileState) assembleK8sV2() error {
...
+	r.entities = append(r.entities, e)
+	for i, ur := range s.k8sUnresourced {
+		if ur == e {
+			// delete from unresourced
+			s.k8sUnresourced = append(s.k8sUnresourced[:i], s.k8sUnresourced[i+1:]...)
+			return
+		}
+	}
+
+	panic("Unable to find entity in unresourced YAML after checking that it was there. This should never happen")
```
maiamcc, line 734: IF this happens, it'd be useful to have some info about what the entity was

internal/tiltfile/tiltfile_state.go
```diff
@@ -591,6 +591,12 @@ func (s *tiltfileState) assembleK8sV1() error {
 
 }
 
+type claimPair struct {
```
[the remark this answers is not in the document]
jazzdan: `claimTuple`? :P

internal/tiltfile/tiltfile_state.go
maiamcc: leave a TODO about the eventual desired behavior?

maiamcc: (but agree that the error message as is is v confusing, i like your proposal)

internal/tiltfile/tiltfile_test.go
```diff
@@ -4381,25 +4381,317 @@ allow_k8s_contexts("hello")
...
+	f.setupFoo()
+	f.yaml("secret.yaml", secret("bar"))
+	f.yaml("anotherworkload.yaml", deployment("baz"))
+
+	f.file("Tiltfile", `
+docker_build('gcr.io/foo', 'foo')
+k8s_yaml('foo.yaml')
+k8s_yaml('anotherworkload.yaml')
 k8s_resource('foo', objects=['bar'])
+k8s_resource('baz', objects=['bar'])
```
maiamcc, line 4591: hm, the fact that this test doesn't trip the "already claimed" error makes me think that the "already claimed" error happens specifically when _the same k8s_resource call_ references the same k8s object via two different selectors, is that true?

maiamcc:
@jazzdan to what extent do you want a really thorough audit of "have you tested all the relevant edge cases"?

internal/tiltfile/tiltfile_test.go
jazzdan: Because we remove YAML from unresourced once we finish process one resource call, but not as we process one resource call, yes that's true.

internal/k8s/names.go
```diff
@@ -46,6 +46,33 @@ func UniqueNames(es []K8sEntity, minComponents int) []string {
 	return ret
 }
 
+func FragmentsToEntities(es []K8sEntity) map[string][]K8sEntity {
```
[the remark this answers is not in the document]
maiamcc: ...anyway this func is confusing enough that it 100% needs a docstring.
maybe:
> Maps all possible fragments to the k8s entitiy(/ies) they correspond to

I wonder if there's a clearer name but if there is uhhh, idk it 🤷‍♀️

internal/k8s/names.go
```diff
@@ -46,6 +46,33 @@ func UniqueNames(es []K8sEntity, minComponents int) []string {
...
+	ret := make(map[string][]K8sEntity, len(es))
+	// how many resources potentially map to a given name
+	counts := make(map[string]int)
+
+	// count how many entities want each potential name
```
maiamcc, line 54: hmmmmm do you use this map at all?

internal/k8s/names_test.go
```diff
@@ -71,3 +73,50 @@ func TestUniqueResourceNames(t *testing.T) {
...
+		t.Fatal(err)
+	}
+
+	actual := FragmentsToEntities(entities)
+
+	expected := map[string][]K8sEntity{
+		"devel-nick-lb-blorg-be":                 {entities[0]},
+		"devel-nick-lb-blorg-be:service":         {entities[0]},
+		"devel-nick-lb-blorg-be:service:default": {entities[0]},
+		// TODO(dmiller): this is wrong
```
maiamcc, line 89: lol i believe you , but if you merge this w/o fixing, more detail here for those of us who are not dan pls? :) and below as well

internal/tiltfile/tiltfile_state.go
```diff
@@ -591,10 +592,33 @@ func (s *tiltfileState) assembleK8sV1() error {
 
 }
 
-type claimPair struct {
-	resource       *k8sResource
-	selector       k8sObjectSelector
-	selectorString string
+func entitiesToFullName(entities []k8s.K8sEntity) (map[string]k8s.K8sEntity, error) {
```
maiamcc, line 596: This name to me implies a `map[k8sEntity]FullName`. You might try `mapFullNameToEntity` or `entitiesByFullName`

internal/tiltfile/tiltfile_state.go
```diff
@@ -591,10 +592,33 @@ func (s *tiltfileState) assembleK8sV1() error {
...
+
+	for _, e := range entities {
+		// TODO(dmiller): if it's already there error
+		ret[entityToFullName(e)] = e
+	}
+
+	return ret, nil
+}
+
+func entityToFullName(e k8s.K8sEntity) string {
```
maiamcc, line 611: i'm surprised that this code doesn't already exist in the k8s package, or at least in `tiltfile/k8s.go` (cf `k8sObjectID.String()`), doesn't necessarily need to change now but i'm a little antsy about similar logic being reimplemented in multiple different places.

internal/tiltfile/tiltfile_state.go
```diff
@@ -608,6 +632,32 @@ func (s *tiltfileState) assembleK8sV2() error {
...
+			if _, seen := seenEntities[entity]; !seen {
+				seenEntities[entity] = struct{}{}
+				uniqueFragments[name] = entity
+				uniqueFragmentNames = append(uniqueFragmentNames, name)
+			}
+		}
+
+	}
+
+	// TODO(dmiller): in all of these error messages use %q more judiciously
```
maiamcc: lulz 👍

internal/tiltfile/tiltfile_state.go
```diff
@@ -608,6 +632,32 @@ func (s *tiltfileState) assembleK8sV2() error {
 		return err
 	}
 
+	unresourcedFragmentsToEntities := k8s.FragmentsToEntities(s.k8sUnresourced)
```
maiamcc: `fragmentsToUnresourcedEntities` might be clearer?

internal/tiltfile/tiltfile_state.go
```diff
@@ -608,6 +632,32 @@ func (s *tiltfileState) assembleK8sV2() error {
...
+	unresourcedFragmentNames := []string{}
+	for name := range unresourcedFragmentsToEntities {
+		unresourcedFragmentNames = append(unresourcedFragmentNames, name)
+	}
+	sort.Sort(byLen(unresourcedFragmentNames))
+	uniqueFragments := map[string]k8s.K8sEntity{}
+	uniqueFragmentNames := []string{}
+	seenEntities := map[k8s.K8sEntity]interface{}{}
+
+	for _, name := range unresourcedFragmentNames {
```
maiamcc: ...wait why did you put all the fragments from the fragment -> entity map into an array to loop over that instead of just looping over the keys from the map?

internal/tiltfile/tiltfile_state.go
```diff
@@ -602,6 +637,40 @@ func (s *tiltfileState) assembleK8sV2() error {
...
 	}
 
+	resourcedEntities := []k8s.K8sEntity{}
+	for _, r := range s.k8sByName {
+		resourcedEntities = append(resourcedEntities, r.entities...)
+	}
+
+	allEntities := append(resourcedEntities, s.k8sUnresourced...)
+
+	fragmentsToEntities := k8s.FragmentsToEntities(allEntities)
```
maiamcc, line 647: as discussed, maybe delete this whole block since it's only used for one error message

internal/tiltfile/tiltfile_state.go
```diff
@@ -610,18 +679,56 @@ func (s *tiltfileState) assembleK8sV2() error {
...
+				if len(entities) > 1 {
+					matchingObjects := make([]string, len(entities))
+					for i, e := range entities {
+						matchingObjects[i] = fullNameFromK8sEntity(e)
+					}
+					return fmt.Errorf("%q is not a unique fragment. Objects that match %q are %s", o, o, strings.Join(matchingObjects, ", "))
+				}
+
+				entitiesToRemove := filterEntitiesBySelector(s.k8sUnresourced, selectors[i])
+				if len(entitiesToRemove) == 0 {
```
maiamcc, line 712: comment this case: "we've already taken these entities out of unresourced"

internal/tiltfile/tiltfile_state.go
```diff
@@ -610,18 +679,56 @@ func (s *tiltfileState) assembleK8sV2() error {
...
+					remainingUnresourcedFragments := make([]string, len(s.k8sUnresourced))
+					for i, entity := range s.k8sUnresourced {
+						remainingUnresourcedFragments[i] = fullNameFromK8sEntity(entity)
+					}
+					return fmt.Errorf("No object identified by the fragment %q could be found in remaining YAML. Valid remaining fragments are: %s", o, strings.Join(remainingUnresourcedFragments, ", "))
+				}
+				if len(entitiesToRemove) > 1 {
```
maiamcc, line 719: I thiiiink this will never happen (🤞 ) b/c we checked above if a selector refers to more than one entity -- maybe panic? at least mention in a comment that we don't expect this to ever happen

internal/tiltfile/tiltfile_state.go
maiamcc: ^ or else put a TODO

internal/tiltfile/tiltfile_state.go
jazzdan: Fullname is a concept that we made up. Since it isn't used in other places yet (but will in the future) let's leave it here. I am pretty confident that it will move somewhere else in a subsequent PR though.

internal/tiltfile/tiltfile_state.go
```diff
@@ -610,18 +625,57 @@ func (s *tiltfileState) assembleK8sV2() error {
...
+				entities, ok := fragmentsToEntities[o]
+				if !ok || len(entities) == 0 {
+					return fmt.Errorf("No object identified by the fragment %q could be found. Possible objects are: %s", o, sliceutils.QuotedStringList(fullNames))
+				}
+				if len(entities) > 1 {
+					matchingObjects := make([]string, len(entities))
+					for i, e := range entities {
+						matchingObjects[i] = fullNameFromK8sEntity(e)
+					}
+					return fmt.Errorf("%q is not a unique fragment. Objects that match %q are %s", o, o, sliceutils.QuotedStringList(matchingObjects))
```
maiamcc, line 654: `QuotedStringList(matchingObjects)` 🙌

internal/tiltfile/tiltfile_state.go
```diff
@@ -610,18 +625,57 @@ func (s *tiltfileState) assembleK8sV2() error {
...
+				if len(entitiesToRemove) == 0 {
+					// we've already taken these entities out of unresourced
+					remainingUnresourced := make([]string, len(s.k8sUnresourced))
+					for i, entity := range s.k8sUnresourced {
+						remainingUnresourced[i] = fullNameFromK8sEntity(entity)
+					}
+					return fmt.Errorf("No object identified by the fragment %q could be found in remaining YAML. Valid remaining fragments are: %s", o, sliceutils.QuotedStringList(remainingUnresourced))
+				}
+				if len(entitiesToRemove) > 1 {
+					panic(fmt.Sprintf("Fragment %q matches %d resources. Each object fragment must match exactly 1 resource. This should NOT be possible", o, len(entitiesToRemove)))
```
maiamcc, line 667:
```suggestion
					panic(fmt.Sprintf("Fragment %q matches %d resources. Each object fragment must match exactly 1 resource. This should NOT be possible at this point in the code, we should have already checked that this fragment was unique", o, len(entitiesToRemove)))
```

internal/tiltfile/tiltfile_state.go
```diff
@@ -1237,6 +1383,7 @@ func newK8sObjectSelector(apiVersion string, kind string, name string, namespace
 }
 
 func (k k8sObjectSelector) matches(e k8s.K8sEntity) bool {
+	log.Printf("Group: %s, Version: %s, GroupVersion: %+v, Kind: %s\n", e.GVK().Group, e.GVK().Version, e.GVK().GroupVersion(), e.GVK().Kind)
```
maiamcc, line 1386: should this be here?

jazzdan: nope, thx

PR #3323, opened 2020-05-11
k8s: log client timeout

landism:
### Problem

When we time out a k8s upsert, we log this:
```
Build Failed: kubectl apply:
stderr: : signal: killed
```

which doesn't indicate what killed it or why

### Solution

log this: `Killed kubectl. Hit timeout of 15s.`

### DIFF
--- a/internal/k8s/client_test.go
+++ b/internal/k8s/client_test.go
@@ -194,7 +194,6 @@ type fakeKubectlRunner struct {
 }
 
 func (f *fakeKubectlRunner) waitForDeadline(ctx context.Context) {
-	// ok not actually forever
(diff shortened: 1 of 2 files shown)


internal/k8s/client_test.go
```diff
@@ -161,37 +161,81 @@ func TestGetGroup(t *testing.T) {
...
+	pauseForever bool
+	stdout       string
+	stderr       string
+	err          error
 
 	calls []call
 }
 
+func (f *fakeKubectlRunner) waitForDeadline(ctx context.Context) {
+	// ok not actually forever
```
maiamcc, line 197: this comment is cute but also doesn't actually make sense because this function doesn't reference `pauseForever` anywhere?

PR #3329, opened 2020-05-12
tiltfile: add os.getenv, os.putenv, os.unsetenv. Fixes https://github.com/windmilleng/tilt/issues/2564

nicks:
Hello @jazzdan,

Please review the following commits I made in branch nicks/ch6958/putenv:

fcf4b19a168d47b8953667637b877ea1ac1e9fff (2020-05-12 17:01:51 -0400)
tiltfile: add os.getenv, os.putenv, os.unsetenv. Fixes https://github.com/windmilleng/tilt/issues/2564

### DIFF
--- a/internal/tiltfile/os/os.go
+++ b/internal/tiltfile/os/os.go
@@ -5,12 +5,12 @@ import (
 	"os"
 	"path/filepath"
 	"runtime"
-	"strings"
 
 	"go.starlark.net/starlark"
 
 	"github.com/windmilleng/tilt/internal/tiltfile/io"
 	"github.com/windmilleng/tilt/internal/tiltfile/starkit"
+	"github.com/windmilleng/tilt/internal/tiltfile/value"
 )
 
 // The starlark OS module.
@@ -56,11 +56,19 @@ func (e Extension) OnStart(env *starkit.Environment) error {
 		return err
 	}
 
-	environValue, err := environ()
+	err = env.AddValue("os.environ", Environ{})
 	if err != nil {
 		return err
 	}
-	err = env.AddValue("os.environ", environValue)
+	err = env.AddBuiltin("os.getenv", getenv)
+	if err != nil {
+		return err
+	}
+	err = env.AddBuiltin("os.putenv", putenv)
+	if err != nil {
+		return err
+	}
+	err = env.AddBuiltin("os.unsetenv", unsetenv)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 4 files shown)


nicks:
actually, this isn't quite right, one sec

nicks:
ok, fixed.

I am...not happy about how much code I needed to duplicate from starlark to get this to work right. The core problem is that we need to make sure updates to the dict are reflected in the environment. The code in environ_methods is basically just copied-and-pasted mechanically, where I just replaced *Dict with Environ (since they use the same API)

nicks:
(There's obviously a simpler implementation possible where *Dict supports different backing stores, but I'm not sure starlark would accept that upstream)

internal/tiltfile/os/os.go
```diff
@@ -77,18 +85,70 @@ func osName() string {
...
-		err := result.SetKey(starlark.String(pair[0]), starlark.String(pair[1]))
-		if err != nil {
-			return nil, err
-		}
+func getenv(t *starlark.Thread, fn *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
+	var varname starlark.Value
+	var defaultVal starlark.Value = starlark.None
+	err := starkit.UnpackArgs(t, fn.Name(), args, kwargs,
+		"varname", &varname,
+		"value?", &defaultVal,
```
jazzdan, line 93:
```suggestion
		"default?", &defaultVal,
```

internal/tiltfile/os/os.go
```diff
@@ -77,18 +85,70 @@ func osName() string {
...
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	varnameStr, ok := value.AsString(varname)
+	if !ok {
+		return nil, fmt.Errorf("varname must be a string, actual: %s", varname)
```
jazzdan, line 101: Does defaultVal also need to be enforced to be a string?

internal/tiltfile/os/os.go
nicks: ahhhh...good catch! this doc
https://docs.python.org/2/library/os.html
says it's value but this doc says it's default
https://docs.python.org/3/library/os.html
and when i tried it out in a real python interpreter, they all only accepted 'default' :rolledeyes:

internal/tiltfile/os/os.go
nicks: surprisingly no!

```
python -c "import os; print(os.getenv('x', {}))"
{}
```

internal/tiltfile/os/os.go
jazzdan: Weird. Windows weird.

PR #3339, opened 2020-05-14
k8s: Reset() the rest mapper if it fails to map an existing type. Fixes https://github.com/windmilleng/tilt/issues/3325

nicks:
Hello @maiamcc,

Please review the following commits I made in branch nicks/ch6999:

1d6f85d8a8440f7dbe3d4e0a4f20b67aac4887d7 (2020-05-14 10:52:37 -0400)
k8s: Reset() the rest mapper if it fails to map an existing type. Fixes https://github.com/windmilleng/tilt/issues/3325

### DIFF
--- a/internal/k8s/client.go
+++ b/internal/k8s/client.go
@@ -14,7 +14,6 @@ import (
 	"github.com/pkg/errors"
 	v1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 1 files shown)


nicks:
Yes! I tested it with `tilt ci`. I actually created a small integration test framework with tilt ci just to test this, but this specific example was slower than I wanted (i.e., it takes about 90 seconds to run) so i kicked it out of the PR

PR #3343, opened 2020-05-14
tiltfile/engine: k8s upsert timeout settable from tiltfile [ch6947]

maiamcc:
Hello @landism, @nicks,

Please review the following commits I made in branch maiamcc/configure-timeout:

e6860bad300cbca51f5660c827d8ba9350dddd63 (2020-05-14 16:18:00 -0400)
tiltfile/engine: expose update_settings.k8s_upsert_timeout_secs

0b9a9f192146471807ca21b024560ac520f33d2c (2020-05-13 18:15:53 -0400)
store: keep the whole UpdateSettings on EngineState, not just MaxParallelBuilds

### DIFF
--- a/internal/tiltfile/updatesettings/update_settings.go
+++ b/internal/tiltfile/updatesettings/update_settings.go
@@ -5,8 +5,9 @@ import (
 	"time"
 
 	"github.com/pkg/errors"
-	"github.com/tilt-dev/tilt/pkg/model"
 	"go.starlark.net/starlark"
+
+	"github.com/tilt-dev/tilt/pkg/model"
 
 	"github.com/tilt-dev/tilt/internal/tiltfile/starkit"
 )
--- a/internal/tiltfile/updatesettings/update_settings.go
+++ b/internal/tiltfile/updatesettings/update_settings.go
@@ -2,6 +2,7 @@ package updatesettings
 
 import (
 	"fmt"
+	"time"
 
 	"go.starlark.net/starlark"
 
@@ -27,9 +28,12 @@ func (e Extension) OnStart(env *starkit.Environment) error {
 }
 
 func (e Extension) updateSettings(thread *starlark.Thread, fn *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
-	var maxParallelUpdates int
+	maxParallelUpdates := model.DefaultMaxParallelUpdates
+	k8sUpsertTimeoutSecs := int(model.DefaultK8sUpsertTimeout / time.Second)
+
 	if err := starkit.UnpackArgs(thread, fn.Name(), args, kwargs,
-		"max_parallel_updates", &maxParallelUpdates); err != nil {
+		"max_parallel_updates?", &maxParallelUpdates,
+		"k8s_upsert_timeout_secs?", &k8sUpsertTimeoutSecs); err != nil {
 		return nil, err
 	}
 
@@ -37,9 +41,15 @@ func (e Extension) updateSettings(thread *starlark.Thread, fn *starlark.Builtin,
 		return nil, fmt.Errorf("max number of parallel updates must be >= 1(got: %d)",
 			maxParallelUpdates)
 	}
+	if k8sUpsertTimeoutSecs < 1 {
+		return nil, fmt.Errorf("minimum k8s upsert timeout is 1s; got %ds",
+			k8sUpsertTimeoutSecs)
+	}
 
 	err := starkit.SetState(thread, func(settings model.UpdateSettings) model.UpdateSettings {
+		// TODO(maia): ensure that this func can only be called once
 		settings = settings.WithMaxParallelUpdates(maxParallelUpdates)
+		settings = settings.WithK8sUpsertTimeout(time.Duration(k8sUpsertTimeoutSecs) * time.Second)
 		return settings
 	})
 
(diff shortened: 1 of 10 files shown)


internal/tiltfile/updatesettings/update_settings.go
```diff
@@ -27,19 +28,28 @@ func (e Extension) OnStart(env *starkit.Environment) error {
 }
 
 func (e Extension) updateSettings(thread *starlark.Thread, fn *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
-	var maxParallelUpdates int
+	maxParallelUpdates := model.DefaultMaxParallelUpdates
```
maiamcc, line 31: hm actually let me unpack these directly onto the struct, one sec

pkg/model/update_settings.go
```diff
@@ -23,6 +29,26 @@ func (us UpdateSettings) WithMaxParallelUpdates(n int) UpdateSettings {
 	return us
 }
 
+func (us UpdateSettings) K8sUpsertTimeout() time.Duration {
```
maiamcc, line 32: again, made getters/setters for private field to make sure we can never call upsert with a nonsense timeout. happy to hear that this is overkill and change

pkg/model/update_settings.go
```diff
@@ -1,9 +1,15 @@
 package model
 
-var DefaultMaxParallelUpdates = 3
+import "time"
+
+const (
+	DefaultMaxParallelUpdates = 3
+	DefaultK8sUpsertTimeout   = 15 * time.Second
```
nicks, line 7: should we double this to 30 seconds, since we know this is too short?

internal/tiltfile/updatesettings/update_settings.go
nicks, line 31: can you add a test for the case where someone calls update_settings twice?

```
update_settings(k8s_upsert_timeout_secs=60)
update_settings(max_parallel_updates=5)
```

I suspect this doesn't behave correctly.

internal/tiltfile/updatesettings/update_settings.go
```diff
@@ -27,19 +28,28 @@ func (e Extension) OnStart(env *starkit.Environment) error {
...
+	k8sUpsertTimeoutSecs := int(model.DefaultK8sUpsertTimeout / time.Second)
+
 	if err := starkit.UnpackArgs(thread, fn.Name(), args, kwargs,
-		"max_parallel_updates", &maxParallelUpdates); err != nil {
+		"max_parallel_updates?", &maxParallelUpdates,
+		"k8s_upsert_timeout_secs?", &k8sUpsertTimeoutSecs); err != nil {
```
nicks, line 36: i wonder if we're eventually going to end up with a time parser... i feel like there are a few of these where the units are odd and it feels very un-kubernetes-like (where you usually specify durations as '100ms', '10s', etc)

internal/tiltfile/updatesettings/update_settings.go
maiamcc: ha yeah i was trying to punt on that but you're right, it's too broken. i just made this function callable at max once.

maiamcc:
Yeahhhh I agree that it's not a great sign if a user is using this :-/ but having an escape hatch in the meantime seems like a good idea

internal/tiltfile/updatesettings/update_settings.go
```diff
@@ -26,20 +29,34 @@ func (e Extension) OnStart(env *starkit.Environment) error {
 	return env.AddBuiltin("update_settings", e.updateSettings)
 }
 
-func (e Extension) updateSettings(thread *starlark.Thread, fn *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
-	var maxParallelUpdates int
+func (e *Extension) updateSettings(thread *starlark.Thread, fn *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
+	if e.callPosition.IsValid() {
```
nicks, line 33: Hmm...I'm concerned about  a world where every extension invents a brand-new pattern with special restrictions, which makes it harder for the user to know what to expect. 

can we just agree on doing something like this
https://github.com/tilt-dev/tilt/blob/master/internal/tiltfile/version/version.go#L38
as a standard pattern?

If we want to introduce a new rule where you can only call a _settings() function once, we should do that for all the _settings() functions.

maiamcc: ah yeah I'd avoided doing that because it means I can't use the getter/setter pattern to protect from weird values getting in here -- but I can give it a shot

maiamcc:
yeah good point about arbitrarily enforcing some settings funcs callable only once.

But i remembered why I didn't unpack directly into the struct: b/c besides sacrificing the getter/setter security, I would have to make the `UpsertTimeout` field the number of seconds rather than the `time.Duration`, and it just gets a little awkwarder everywhere it's used.

wdyt think of this? (if this PR goes on much longer I'll probably just drop it for a bit to focus on v1 stuff)

internal/tiltfile/updatesettings/update_settings.go
```diff
@@ -26,26 +27,57 @@ func (e Extension) OnStart(env *starkit.Environment) error {
...
+		}
 		return settings
 	})
 
 	return starlark.None, err
 }
 
+func valueToInt(v starlark.Value) (val int, wasPassed bool, err error) {
+	switch x := v.(type) {
+	case nil:
```
nicks, line 71: might be good to include starlark.None here

maiamcc: hm you mean if it's nil OR starlark.None, assume it's not set?

PR #3348, opened 2020-05-15
k8s: print local registry as part of 'tilt doctor', and add support for new kind registry annotation

nicks:
Hello @maiamcc,

Please review the following commits I made in branch nicks/kind:

f1159e9ccd44f5eb0d91600eec063a4113bd214b (2020-05-15 17:11:58 -0400)
k8s: print local registry as part of 'tilt doctor', and add support for new kind registry annotation

### DIFF
--- a/internal/cli/doctor.go
+++ b/internal/cli/doctor.go
@@ -11,6 +11,7 @@ import (
 	wmanalytics "github.com/tilt-dev/wmclient/pkg/analytics"
 
 	"github.com/tilt-dev/tilt/internal/analytics"
+	"github.com/tilt-dev/tilt/pkg/logger"
 )
 
 type doctorCmd struct {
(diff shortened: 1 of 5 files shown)


internal/k8s/registry.go
```diff
@@ -17,6 +17,9 @@ import (
 const annotationRegistry = "tilt.dev/registry"
```
maiamcc, line 17: might comment this too, and put a little more detail in both comments: "Key for retrieving registry from node annotations if cluster created via Tilt's script <link> || Kind's script <link>"

maiamcc: also maybe call this `tiltAnnotationRegistry`?

internal/cli/doctor.go
```diff
@@ -118,6 +119,20 @@ func (c *doctorCmd) run(ctx context.Context, args []string) error {
 	kVersion, err := wireK8sVersion(ctx)
 	printField("Version", kVersion, err)
 
+	kClient, err := wireK8sClient(ctx)
+	if err == nil {
+		// blackhole any warnings
+		newCtx := logger.WithLogger(ctx, logger.NewDeferredLogger(ctx))
+		registry := kClient.LocalRegistry(newCtx)
+		if registry.Empty() {
+			printField("Local Registry", "none", nil)
```
maiamcc, line 128: not suuuper concerned about this, but it might confuse a user to see this if they'd, say, set `default_registry('localhost:5000')` in their Tiltfile. Consider calling the field "Auto-detected local registry"? But also, whatevs.

nicks: changed it to Cluster Local Registry, which i think will end up being the term of art for this

internal/k8s/registry.go
nicks: done

PR #3358, opened 2020-05-20
tiltfile: make workload parameter optional

jazzdan:


internal/tiltfile/k8s.go
```diff
@@ -382,43 +382,52 @@ func (s *tiltfileState) k8sResourceV2(thread *starlark.Thread, fn *starlark.Buil
 		return nil, err
 	}
 
+	resourceName := workload
+	needsToHaveObjects := false
 	if workload == "" {
-		return nil, fmt.Errorf("%s: workload must not be empty", fn.Name())
+		resourceName = newName
```
maiamcc, line 388: (no changes needed, just grumbling) i feel a little weird about the way this way of doing things splits up the re/naming logic--some of it is handled here if workload is empty, some happens down on ln 386, and i think it wouldn't be obvious to someone new coming into this codebase where name assignment happens. i'd rather see `k8sResourceOptions` store in an array, and in assembly we iterate over them all and name or rename them in the same place.

of course that involves a different strategy for determining whether you've called `k8s_resource` for a given resource name, and i get it if you don't wanna bother with that refactor right now. figured i'd mention if you feel moved to change, but this isn't a big deal.

internal/tiltfile/k8s.go
```diff
@@ -382,43 +382,52 @@ func (s *tiltfileState) k8sResourceV2(thread *starlark.Thread, fn *starlark.Buil
...
+		// If a resource doesn't specify a workload then it needs to have objects to be valid
+		needsToHaveObjects = true
```
maiamcc, line 390: nbd but why not check here whether they passed the objects param, rather than storing a bool and waiting til the bottom of the function?

internal/tiltfile/tiltfile_test.go
```diff
@@ -2866,7 +2866,7 @@ k8s_resource('foo', port_forwards=8000)
 	f.loadErrString("k8s_resource already called for foo")
 }
 
-func TestK8sResourceEmptyWorkloadSpecifier(t *testing.T) {
+func TestK8sResourceEmptyWorkloadSpecifierAndNoObjects(t *testing.T) {
```
maiamcc, line 2869: super nbd, if you feel inspired i might reorder the tests so that all the k8s_resource stuff is together instead of being split by workload_to_resource tests (there's a chunk up here and a chunk around ln4449) but 🤷

internal/tiltfile/k8s.go
```diff
@@ -382,43 +382,52 @@ func (s *tiltfileState) k8sResourceV2(thread *starlark.Thread, fn *starlark.Buil
 		return nil, err
 	}
 
+	resourceName := workload
+	needsToHaveObjects := false
 	if workload == "" {
```
maiamcc, line 387: uh oh, you don't require anywhere that if no `workload` passed, `new_name` must be passed. e.g. this test I wrote passes and it shouldn't:
```
func TestK8sResouceNonWorkloadRequiresNewName(t *testing.T) {
	f := newFixture(t)
	defer f.TearDown()

	f.yaml("secret.yaml", secret("bar"))
	f.yaml("namespace.yaml", namespace("baz"))

	f.file("Tiltfile", `
k8s_yaml('secret.yaml')
k8s_yaml('namespace.yaml')
k8s_resource(objects=['bar', 'baz:namespace:default'])
`)

	f.load()

	f.assertNextManifest("", k8sObject("bar", "Secret"), k8sObject("baz", "Namespace"))
	f.assertNoMoreManifests()
}
```

internal/tiltfile/tiltfile_state.go
```diff
@@ -623,6 +623,13 @@ func (s *tiltfileState) assembleK8sV2() error {
 	}
 
 	for workload, opts := range s.k8sResourceOptions {
+		if opts.nonWorkload {
```
maiamcc, line 626: this isn't a big deal at all but it feels kinda weird to me to make the resource, put it in the map, and immediately get it back out of the map. i might do something like
```
var r Resource
var ok bool
if opts.nonWorkload {
    r... := makeResource
    s.k8sByName[opts.newName] = r
} else if r, ok = s.k8sByName[workload]; !ok {
    return error("no resource found...)
}
    r.updateWithOpts(opts)
```

internal/tiltfile/tiltfile_state.go
```diff
@@ -623,6 +623,13 @@ func (s *tiltfileState) assembleK8sV2() error {
...
+			r, err := s.makeK8sResource(opts.newName)
+			if err != nil {
+				return err
+			}
+			s.k8sByName[opts.newName] = r
```
maiamcc, line 631: hmm what happens if you have an actual workload "foo" and a non-workload resource that you name "foo" -- will the latter overwrite the former in the `k8sByName` map?

internal/tiltfile/tiltfile_test.go
```diff
@@ -4798,6 +4798,48 @@ k8s_resource('foo', objects=['baz:namespace:qux'])
...
+k8s_resource(new_name='foo', objects=['bar', 'baz:namespace:default'])
+`)
+
+	f.load()
+
+	f.assertNextManifest("foo", k8sObject("bar", "Secret"), k8sObject("baz", "Namespace"))
+	f.assertNoMoreManifests()
+}
+
+func TestK8sWorkloadOnlyResourceWithAllTheOptions(t *testing.T) {
```
maiamcc, line 4820: do you have a comparable test for a non-workload resource with all the options? would be good to make sure that that codepath stays intact

internal/tiltfile/tiltfile_test.go
```diff
@@ -4798,6 +4798,48 @@ k8s_resource('foo', objects=['baz:namespace:qux'])
 	f.loadErrString("No object identified by the fragment \"baz:namespace:qux\" could be found. Possible objects are: \"foo:Deployment:default\", \"baz:Namespace:default\"")
 }
 
+func TestK8sResouceObjectsNonWorkloadOnly(t *testing.T) {
```
maiamcc, line 4801: also what's the meaning of the "Only" in this test name? I think I get that "workload only" means "no additional objects passed" but this usage doesn't make sense to me at first glance.

internal/tiltfile/tiltfile_state.go
maiamcc: oh nvm this is handled at ln 403. (This is why i wasn't jazzed about splitting up the re/naming logic 😅 ) still, there should definitely be a test for this case, and it still feels kind of brittle to leave that map overwrite technically possible in this code.

internal/tiltfile/k8s.go
jazzdan: It felt weird to me to parse the parameters out of order, but maybe that's not a big deal.

internal/tiltfile/k8s.go
jazzdan: Good catch! Thanks.

internal/tiltfile/tiltfile_state.go
jazzdan: Unfortunately without a larger refactor this will fail on non-workload-only-resources.

internal/tiltfile/tiltfile_test.go
jazzdan: It's *Non*WorkloadOnly, so only non-workload objects are passed

internal/tiltfile/k8s.go
```diff
@@ -382,43 +382,52 @@ func (s *tiltfileState) k8sResourceV2(thread *starlark.Thread, fn *starlark.Buil
...
 	}
 
-	// TODO(dmiller): this should be parsed as a valid identifier, and stored its own type
 	objects, err := value.SequenceToStringSlice(objectsVal)
 	if err != nil {
 		return nil, errors.Wrapf(err, "%s: resource_deps", fn.Name())
 	}
 
-	s.k8sResourceOptions[workload] = k8sResourceOptions{
+	position := thread.CallFrame(1).Pos
```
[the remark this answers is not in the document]
jazzdan: Fixed!

internal/tiltfile/tiltfile_state.go
maiamcc: aww :( womp. okay cool

internal/tiltfile/tiltfile_test.go
```diff
@@ -4798,6 +4783,100 @@ k8s_resource('foo', objects=['baz:namespace:qux'])
...
+	f.yaml("secret.yaml", secret("bar"))
+
+	f.file("Tiltfile", `
+k8s_yaml('foo.yaml')
+k8s_yaml('secret.yaml')
+k8s_resource('foo', new_name='bar')
+k8s_resource(new_name='bar', objects=['bar:secret'])
+`)
+
+	// NOTE(dmiller): because `range`ing over maps is unstable we don't which error we will encounter:
```
maiamcc, line 4873:
```suggestion
	// NOTE(dmiller): because `range`ing over maps is unstable we don't know which error we will encounter:
```
Ooh nice.

PR #3361, opened 2020-05-21
tiltfile: explicitly mark non-workload manifests as such so we can handle them differently in the UI

jazzdan:
…ifest

### DIFF
--- a/internal/tiltfile/tiltfile_state.go
+++ b/internal/tiltfile/tiltfile_state.go
@@ -628,6 +628,7 @@ func (s *tiltfileState) assembleK8sV2() error {
 			if err != nil {
 				return err
 			}
+			r.nonWorkload = true
 			s.k8sByName[opts.newName] = r
 		}
 		if r, ok := s.k8sByName[workload]; ok {
@@ -1039,7 +1040,7 @@ func (s *tiltfileState) translateK8s(resources []*k8sResource) ([]model.Manifest
 		}
 
 		k8sTarget, err := k8s.NewTarget(mn.TargetName(), r.entities, s.defaultedPortForwards(r.portForwards),
-			r.extraPodSelectors, r.dependencyIDs, r.imageRefMap)
+			r.extraPodSelectors, r.dependencyIDs, r.imageRefMap, r.nonWorkload)
 		if err != nil {
 			return nil, err
 		}
--- a/pkg/model/manifest.go
+++ b/pkg/model/manifest.go
@@ -113,10 +113,6 @@ func (m Manifest) IsK8s() bool {
 }
 
 func (m Manifest) IsNonWorkloadYAMLManifest() bool {
-	if m.Name == UnresourcedYAMLManifestName {
-		return true
-	}
-
 	k8sTarget, ok := m.deployTarget.(K8sTarget)
 	if !ok {
 		return false
--- a/pkg/model/manifest.go
+++ b/pkg/model/manifest.go
@@ -112,12 +112,11 @@ func (m Manifest) IsK8s() bool {
 	return ok
 }
 
-func (m Manifest) IsNonWorkloadYAMLManifest() bool {
-	k8sTarget, ok := m.deployTarget.(K8sTarget)
-	if !ok {
-		return false
+func (m Manifest) NonWorkloadManifest() bool {
+	if k8sTarget, ok := m.deployTarget.(K8sTarget); ok {
+		return k8sTarget.NonWorkload
 	}
-	return k8sTarget.NonWorkload
+	return false
 }
 
 func (m Manifest) DeployTarget() TargetSpec {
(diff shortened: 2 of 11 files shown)


jazzdan:
This makes the UI look better for non-workload-only manifests, aka what we used to call "YAML only" manifests.

Before:

before ui fix

After:

after ui fix

internal/store/runtime_state.go
```diff
@@ -111,7 +111,7 @@ func (s K8sRuntimeState) RuntimeStatusError() error {
 }
 
 func (s K8sRuntimeState) RuntimeStatus() model.RuntimeStatus {
-	if s.IsUnresourced {
+	if s.IsUnresourced || len(s.Pods) == 0 {
```
nicks, line 114: hmmm....this isn't right.

This means "we haven't seen any pods yet"

what you want is "we will never see any pods"

the distinction is important, because a lot of things will break if you report ready status prematurely (tilt ci, resource deps, etc)

jazzdan: @nicks gotcha, is there a built-in way to do that as of today?

nicks: that's what Unresourced is....i think you'll have to replace Unresourced with a new signifier for this?

nicks: i think you still need to remove this pod check (here and above)?

pkg/model/manifest.go
```diff
@@ -113,7 +115,7 @@ func (m Manifest) IsK8s() bool {
 }
 
 func (m Manifest) IsUnresourcedYAMLManifest() bool {
```
nicks, line 117: should we rename this to NonWorkloadManifest?

internal/tiltfile/tiltfile_state.go
```diff
@@ -1036,6 +1037,7 @@ func (s *tiltfileState) translateK8s(resources []*k8sResource) ([]model.Manifest
 			Name:                 mn,
 			TriggerMode:          tm,
 			ResourceDependencies: mds,
+			NonWorkload:          r.nonWorkload,
```
nicks, line 1040: can you add a unit test in internal/tiltfile that demonstrates how this is set? I'm not totally sure I understand the difference between "a manifest with no image targets" and "a manifest with no workloads"

jazzdan: Yeah it's possible that this is unnecessary, let me do some more digging.

pkg/model/manifest.go
```diff
@@ -112,8 +112,8 @@ func (m Manifest) IsK8s() bool {
 	return ok
 }
 
-func (m Manifest) IsUnresourcedYAMLManifest() bool {
-	return m.Name == UnresourcedYAMLManifestName
+func (m Manifest) IsNonWorkloadYAMLManifest() bool {
```
maiamcc, line 115: a. hannnng on, this would tell us that e.g. a local resource manifest or a docker-compose-no-image manifest was a non-workload manifest, that seems wrong. Also
b. you can have a workload manifest that we don't build images for! (i.e. a manifest that spins up pods running an image pulled from the internet.)
c. less importantly, "NonWorkloadYAMLManifest" seems redundant, couldn't you just say "NonWorkloadManifest"? (i.e. IF it's a k8s manifest and is non-workload, it'll necessarily be a yaml-only manifest)

jazzdan: Okay, I'm going to back to the strategy of setting a bit on the k8sTarget during Tiltfile execution before things get passed off to the engine, and then check that bit here.

internal/tiltfile/tiltfile_state.go
nicks: ping this comment; i think having a test for when we expect this to be true to be the important bit

pkg/model/manifest.go
maiamcc: yay i like this a lot better!

one nit: "NonWorkloadYAMLManifest" seems redundant, couldn't you just say "NonWorkloadManifest"? (i.e. IF it's a k8s manifest and is non-workload, it'll necessarily be a yaml-only manifest)

pkg/model/manifest.go
```diff
@@ -112,8 +112,12 @@ func (m Manifest) IsK8s() bool {
...
+	k8sTarget, ok := m.deployTarget.(K8sTarget)
```
maiamcc, line 116: nbd but can save some lines:
```
if k8sTarget, ok := m.deployTarget.(K8sTarget); ok {
    return k8sTarget.NonWorkload
}
return false
```

internal/tiltfile/tiltfile_state.go
jazzdan: Added some assertions to the tiltfile tests @nicks, thanks for pointing this out

PR #3362, opened 2020-05-21
engine: better handling for image builds in multiple manifests. Fixes https://github.com/windmilleng/tilt/issues/2992

nicks:
Hello @landism, @maiamcc,

Please review the following commits I made in branch nicks/ch6738:

c9950540c7fc011d2d348b89b250e99d1ba8215b (2020-05-21 11:07:44 -0400)
engine: better handling for image builds in multiple manifests. Fixes https://github.com/windmilleng/tilt/issues/2992

### DIFF
--- a/internal/engine/buildcontrol/build_control.go
+++ b/internal/engine/buildcontrol/build_control.go
@@ -92,6 +92,9 @@ func RemoveTargetsWithBuildingComponents(mts []*store.ManifestTarget) []*store.M
 	for _, mt := range mts {
 		if mt.State.IsBuilding() {
 			building[mt.Manifest.ID()] = true
+
+			// TODO(nick): This logic isn't quite right.  A manifest can re-use image
+			// results from another manifest's build.
 			for _, spec := range mt.Manifest.TargetSpecs() {
 				building[spec.ID()] = true
 			}
@@ -219,8 +222,9 @@ func IsBuildingLocalTarget(state store.EngineState) bool {
 }
 
 // Go through all the manifests, and check:
-// 1) all pending file changes, and
-// 2) all pending manifest changes
+// 1) all pending file changes
+// 2) all pending dependency changes (where an image has been rebuilt by another manifest), and
+// 3) all pending manifest changes
 // The earliest one is the one we want.
 //
 // If no targets are pending, return nil
--- a/internal/engine/buildcontroller.go
+++ b/internal/engine/buildcontroller.go
@@ -172,7 +172,12 @@ func buildStateSet(ctx context.Context, manifest model.Manifest, specs []model.T
 		}
 		sort.Strings(filesChanged)
 
-		buildState := store.NewBuildState(status.LastSuccessfulResult, filesChanged)
+		var depsChanged []model.TargetID
+		for dep := range status.PendingDependencyChanges {
+			depsChanged = append(depsChanged, dep)
+		}
+
+		buildState := store.NewBuildState(status.LastSuccessfulResult, filesChanged, depsChanged)
 
 		// Pass along the container when we can update containers in-place.
 		//
--- a/internal/engine/buildcontroller_test.go
+++ b/internal/engine/buildcontroller_test.go
@@ -1267,6 +1267,106 @@ func TestErrorHandlingWithMultipleBuilds(t *testing.T) {
 	f.assertAllBuildsConsumed()
 }
 
+func TestManifestsWithSameTwoImages(t *testing.T) {
+	f := newTestFixture(t)
+	m1, m2 := NewManifestsWithSameTwoImages(f)
+	f.Start([]model.Manifest{m1, m2})
+
+	f.waitForCompletedBuildCount(2)
+
+	call := f.nextCall("m1 build1")
+	assert.Equal(t, m1.K8sTarget(), call.k8s())
+
+	call = f.nextCall("m2 build1")
+	assert.Equal(t, m2.K8sTarget(), call.k8s())
+
+	aPath := f.JoinPath("common", "a.txt")
+	f.fsWatcher.Events <- watch.NewFileEvent(aPath)
+
+	f.waitForCompletedBuildCount(4)
+
+	// Make sure that both builds are triggered, and that they
+	// are triggered in a particular order.
+	call = f.nextCall("m1 build2")
+	assert.Equal(t, m1.K8sTarget(), call.k8s())
+
+	state := call.state[m1.ImageTargets[0].ID()]
+	assert.Equal(t, map[string]bool{aPath: true}, state.FilesChangedSet)
+
+	// Make sure that when the second build is trigged, we did the bookkeeping
+	// correctly around marking the first and second image built and only deploying
+	// the k8s resources.
+	call = f.nextCall("m2 build2")
+	assert.Equal(t, m2.K8sTarget(), call.k8s())
+
+	id := m2.ImageTargets[0].ID()
+	result := f.b.resultsByID[id]
+	assert.Equal(t, store.NewBuildState(result, nil, nil), call.state[id])
+
+	id = m2.ImageTargets[1].ID()
+	result = f.b.resultsByID[id]
+	assert.Equal(t, store.NewBuildState(result, nil, nil), call.state[id])
+
+	err := f.Stop()
+	assert.NoError(t, err)
+	f.assertAllBuildsConsumed()
+}
+
+func TestManifestsWithTwoCommonAncestors(t *testing.T) {
+	f := newTestFixture(t)
+	m1, m2 := NewManifestsWithTwoCommonAncestors(f)
+	f.Start([]model.Manifest{m1, m2})
+
+	f.waitForCompletedBuildCount(2)
+
+	call := f.nextCall("m1 build1")
+	assert.Equal(t, m1.K8sTarget(), call.k8s())
+
+	call = f.nextCall("m2 build1")
+	assert.Equal(t, m2.K8sTarget(), call.k8s())
+
+	aPath := f.JoinPath("base", "a.txt")
+	f.fsWatcher.Events <- watch.NewFileEvent(aPath)
+
+	f.waitForCompletedBuildCount(4)
+
+	// Make sure that both builds are triggered, and that they
+	// are triggered in a particular order.
+	call = f.nextCall("m1 build2")
+	assert.Equal(t, m1.K8sTarget(), call.k8s())
+
+	state := call.state[m1.ImageTargets[0].ID()]
+	assert.Equal(t, map[string]bool{aPath: true}, state.FilesChangedSet)
+
+	// Make sure that when the second build is triggered, we did the bookkeeping
+	// correctly around marking the first and second image built, and only
+	// rebuilding the third image and k8s deploy.
+	call = f.nextCall("m2 build2")
+	assert.Equal(t, m2.K8sTarget(), call.k8s())
+
+	id := m2.ImageTargets[0].ID()
+	result := f.b.resultsByID[id]
+	assert.Equal(t,
+		store.NewBuildState(result, nil, nil),
+		call.state[id])
+
+	id = m2.ImageTargets[1].ID()
+	result = f.b.resultsByID[id]
+	assert.Equal(t,
+		store.NewBuildState(result, nil, nil),
+		call.state[id])
+
+	id = m2.ImageTargets[2].ID()
+	result = f.b.resultsByID[id]
+	assert.Equal(t,
+		store.NewBuildState(result, nil, []model.TargetID{m2.ImageTargets[1].ID()}),
+		call.state[id])
+
+	err := f.Stop()
+	assert.NoError(t, err)
+	f.assertAllBuildsConsumed()
+}
+
 func (f *testFixture) waitUntilManifestBuilding(name model.ManifestName) {
 	msg := fmt.Sprintf("manifest %q is building", name)
 	f.WaitUntilManifestState(msg, name, func(ms store.ManifestState) bool {
--- a/internal/engine/upper.go
+++ b/internal/engine/upper.go
@@ -224,6 +224,94 @@ func handleBuildStarted(ctx context.Context, state *store.EngineState, action bu
 	removeFromTriggerQueue(state, mn)
 }
 
+func updateBuildStatus(status *store.BuildStatus, br model.BuildRecord, result store.BuildResult, isBuildSuccess bool) {
+	// Record the result
+	status.LastResult = result
+
+	// Remove pending file changes that were consumed by this build.
+	for file, modTime := range status.PendingFileChanges {
+		if store.BeforeOrEqual(modTime, br.StartTime) {
+			delete(status.PendingFileChanges, file)
+		}
+	}
+
+	// Remove pending dependency changes that were consumed by this build.
+	for id, modTime := range status.PendingDependencyChanges {
+		if store.BeforeOrEqual(modTime, br.StartTime) {
+			delete(status.PendingDependencyChanges, id)
+		}
+	}
+
+	// If the whole build succeeded, mark all of the images in the
+	// build as successful.
+	//
+	// This has a minor bug: if one of the images in a set failed, but another one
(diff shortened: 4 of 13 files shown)


nicks:
I think this isn't _too_ complex a solution, and sets us up pretty well to make things better incrementally. The most important bit of the PR is this comment:

```	
        // Stores the times that dependencies were marked dirty, so we can prioritize
	// the oldest one first.
	//
	// Long-term, we want to process all dependencies as a build graph rather than
	// a list of manifests. Specifically, we'll build one Target at a time.  One
	// the build completes, we'll look at all the targets that depend on it, and
	// mark PendingDependencyChanges to indicate that they need a rebuild.
	//
	// Short-term, we only use this for cases where two manifests share a common
	// image. This only handles cross-manifest dependencies.
	//
	// This approach allows us to start working on the bookkeeping and
	// dependency-tracking in the short-term, without having to switch over to a
	// full dependency graph in one swoop.
````

**Comment by nicks** [comment]
had a good convo with Maia offlist about this; I'm going to write some graph accessors to make it more clear what all the nested loops are doing

**Comment by nicks** [comment]
ok, i moved all the for loops in handleBuildResults into graph traversal accessors. ptal?

**Comment by landism** [comment]
hoo boy this is a big PR!

I'm still going through it, but while I do, some tests for these cases could be good:
1) resources r1 and r2 both depend on image i1 - user manually rebuilds r1 - does the image rebuild trigger a rebuild of r2?
2) resource r1 uses image i3, which is FROM image i2, which is FROM image i1. Is IsRecycled transitive? I'm still trying to understand this PR, so I'm not sure if this is the right way to test this, but maybe a construction here would be add resource r2 that uses image i1, rebuild r2 - do i2 and i3 get recycled when rebuilding r1? (also that's just an optimization so maybe not super important)

**Comment by nicks** [comment]
cool, let me send out two independent prs:
- a PR that isolates the TargetQueue changes we want
- a PR that isolates the types of graph structures that we want to be able to support, and adds tests for them
to make it clearer what's going on in this pr

**Comment by maiamcc** [comment]
FYI [here are notes i took inline](https://github.com/tilt-dev/tilt/tree/nicks/ch6738-scratch) while going through this PR with Matt -- some suggestions for how to clarify comments, and some questions/comments about implementation. If that's a helpful format, feel free to use, otherwise I can submit them all as PR comments when this is ready for re-review.

**Comment by nicks** [comment]
ok, i sent out 
- https://github.com/tilt-dev/tilt/pull/3372
- https://github.com/tilt-dev/tilt/pull/3374
for some of the cleanup tasks your comments inspired

after those PRs go in, the whole concept of recycling will magically go away, since it's really just working around ImageBuildAndDeployers return results for things they didn't build

**Comment by landism** [comment]
Is this waiting for review or changes?

**Comment by nicks** [comment]
waiting on review! haven't been pushing on it too hard since there's a lot of other things in the air

File: internal/engine/upper.go [hunk 1]
```diff
@@ -224,6 +224,94 @@ func handleBuildStarted(ctx context.Context, state *store.EngineState, action bu
...
+
+		updateBuildStatus(status, br, result, isBuildSuccess)
+	})
+
+	// Suppose we built manifestA, which contains imageA depending on imageCommon.
+	//
+	// We also have manifestB, which contains imageB depending on imageCommon.
+	//
+	// We need to mark imageB as dirty, because it was not built in the manifestA
+	// build but its dependency was built.
```
maiamcc, line 289: (idk if this comment is quite right but i DO think it's useful to mention the deploy-targets-depending-on-image-targets case--when this comment only talks about inter-image dependencies, it seems kinda special-case-y, but really this bit of code is much more generic than that!)
```suggestion
	// build but its dependency was built.
	
	// Note that this logic also applies to deploy targets depending on image targets. If we built manifestA, which contains imageX and deploy target k8sA, and manifestB contains imageX and deploy target k8sB, we need to mark target k8sB as dirty so that Tilt actually deploys the changes to imageX.
```

internal/engine/upper.go
```diff
@@ -224,6 +224,94 @@ func handleBuildStarted(ctx context.Context, state *store.EngineState, action bu
...
+	engineState.ForEachMutableBuildStatusDependingOn(results, func(id, reverseDepID model.TargetID, mn model.ManifestName, reverseDepStatus *store.BuildStatus) {
+		if mn == mt.Manifest.Name {
+			// We're only worried about manifests that we didn't build.
+			return
+		}
+
+		_, ok := results[reverseDepID]
+		if ok {
+			// The reverseDep is fresh! Skip it.
```
maiamcc, line 298:
```suggestion
			// The reverseDep is fresh (i.e. was just built)! Skip it.
```

internal/engine/upper.go
```diff
@@ -224,6 +224,94 @@ func handleBuildStarted(ctx context.Context, state *store.EngineState, action bu
...
+			return
+		}
+
+		_, depIsNew := results[id]
```
maiamcc, line 302: doesn't the iterator guarantee that we'll only ever get `id` if it's in `results`, making this check redundant?

internal/engine/upper.go
```diff
@@ -224,6 +224,94 @@ func handleBuildStarted(ctx context.Context, state *store.EngineState, action bu
...
+		if !depIsNew {
+			// This build didn't actually produce a new result. So no need to mark other
+			// manifests dirty.
+			return
+		}
+
+		// This target was not built, but one of its dependencies was
+		// built, so we have to mark it for rebuild.
```
maiamcc, line 310: (or otherwise clarify the object of the sentence, it's ambiguous :-/ )
```suggestion
		// built, so we have to mark the target for rebuild.
```

internal/engine/upper.go
```diff
@@ -224,6 +224,94 @@ func handleBuildStarted(ctx context.Context, state *store.EngineState, action bu
...
+		updateBuildStatus(status, br, result, isBuildSuccess)
+	})
+
+	// Suppose we built manifestA, which contains imageA depending on imageCommon.
+	//
+	// We also have manifestB, which contains imageB depending on imageCommon.
+	//
+	// We need to mark imageB as dirty, because it was not built in the manifestA
+	// build but its dependency was built.
+	engineState.ForEachMutableBuildStatusDependingOn(results, func(id, reverseDepID model.TargetID, mn model.ManifestName, reverseDepStatus *store.BuildStatus) {
```
maiamcc, line 290: the naming scheme `id` and `reverseDepID` had me tying myself in knots, is there a more straightforward naming scheme? (like say `parent` and `child`?) Alternately, could clarify in re: the example above--
> In terms of the above example: `id` refers to `imageCommon`, `reverseDepId` refers to `imageB`

internal/engine/buildcontrol/build_control.go
```diff
@@ -92,6 +92,9 @@ func RemoveTargetsWithBuildingComponents(mts []*store.ManifestTarget) []*store.M
 	for _, mt := range mts {
 		if mt.State.IsBuilding() {
 			building[mt.Manifest.ID()] = true
+
+			// TODO(nick): This logic isn't quite right.  A manifest can re-use image
```
maiamcc, line 96: But we still don't want them to build in parallel, or we might try to build the same image twice at the same time, right? Or am I missing something?

internal/engine/buildcontroller.go
```diff
@@ -165,7 +165,12 @@ func buildStateSet(ctx context.Context, manifest model.Manifest, specs []model.T
 
 	for _, spec := range specs {
 		id := spec.ID()
-		if id.Type != model.TargetTypeImage && id.Type != model.TargetTypeDockerCompose && id.Type != model.TargetTypeLocal {
+
+		// TODO(nick): There's no good reason for this 'if' statement, but lots of
```
maiamcc, line 169: 😞

internal/engine/buildcontroller_test.go
```diff
@@ -1266,6 +1266,92 @@ func TestErrorHandlingWithMultipleBuilds(t *testing.T) {
...
+
+	imageStateM1I0 := call.state[m1.ImageTargets[0].ID()]
+	assert.Equal(t, map[string]bool{aPath: true}, imageStateM1I0.FilesChangedSet)
+
+	call = f.nextCall("m2 build2")
+	assert.Equal(t, m2.K8sTarget(), call.k8s())
+
+	// Make sure that when the second build is trigged, we did the bookkeeping
+	// correctly around marking the first and second image built, and only rebuilding
+	// the third image.
```
maiamcc, line 1300: what's the third image in this case? I thought there were only two images across both manifests.

maiamcc: Ah wait i think this is copypasta from `TwoCommonAncestors`

I think you might want to bump this comment up above ln1295 and say something like:
> // Make sure that when the second build is trigged, we did the bookkeeping correctly around marking the first and second image built (and not rebuilding them), but that we DID redeploy the k8s target.

internal/engine/buildcontroller_test.go
```diff
@@ -1266,6 +1266,92 @@ func TestErrorHandlingWithMultipleBuilds(t *testing.T) {
...
+	idM2I0 := m2.ImageTargets[0].ID()
+	idM2I1 := m2.ImageTargets[1].ID()
+	assert.Equal(t, map[string]bool{}, call.state[idM2I0].FilesChangedSet)
```
maiamcc, line 1303: a. i'm surprised not to see an assertion about `state[idM2I0].DepsChanged` or `state[idM2I1].FilesChanged`, but also
b. maybe worth verifying that `call.state` contains `idM2I0` and `idM2I`--if for some reason it doesn't, these checks will still pass.
c. fwiw these names are hard to read, i have to squint at this code to be able to tell what's going on

internal/engine/buildcontroller_test.go
```diff
@@ -1266,6 +1266,92 @@ func TestErrorHandlingWithMultipleBuilds(t *testing.T) {
...
+	call = f.nextCall("m1 build2")
+	assert.Equal(t, m1.K8sTarget(), call.k8s())
+
+	imageStateM1I0 := call.state[m1.ImageTargets[0].ID()]
+	assert.Equal(t, map[string]bool{aPath: true}, imageStateM1I0.FilesChangedSet)
+
+	call = f.nextCall("m2 build2")
+	assert.Equal(t, m2.K8sTarget(), call.k8s())
+
+	// Make sure that when the second build is trigged, we did the bookkeeping
```
maiamcc, line 1340: again, consider bumping this comment up a few lines and mentioning that we care not only about the image bookkeeping, but also that we DID deploy the k8s target (tho this maybe isn't as big a part of this test as it is of `TestManifestsWithSameTwoImages`)

internal/engine/buildcontroller_test.go
```diff
@@ -1266,6 +1266,92 @@ func TestErrorHandlingWithMultipleBuilds(t *testing.T) {
...
+	// correctly around marking the first and second image built, and only rebuilding
+	// the third image.
+	idM2I0 := m2.ImageTargets[0].ID()
```
maiamcc, line 1343: again, these variable names are hard to parse

internal/engine/upper.go
```diff
@@ -224,6 +224,94 @@ func handleBuildStarted(ctx context.Context, state *store.EngineState, action bu
 	removeFromTriggerQueue(state, mn)
 }
 
+func updateBuildStatus(status *store.BuildStatus, br model.BuildRecord, result store.BuildResult, isBuildSuccess bool) {
+	// Record the last result
```
maiamcc, line 228:
```suggestion
	// Record the result
```
(otherwise it seems like you're talking about the result BEFORE the one you're currently processing)

internal/engine/upper.go
```diff
@@ -224,6 +224,94 @@ func handleBuildStarted(ctx context.Context, state *store.EngineState, action bu
...
+			delete(status.PendingDependencyChanges, id)
+		}
+	}
+
+	// If the whole build succeeded, mark all of the images in the
+	// build as successful.
+	//
+	// This has a minor bug: if one of the images in a set failed, but another one
+	// succeeded, this doesn't mark any of them as successful. We should fix this,
+	// though personally, I think LastSuccessfulResult has problematic semantics
```
maiamcc, line 250:
```suggestion
	// though personally, I (Nick) think LastSuccessfulResult has problematic semantics
```

internal/engine/upper.go
```diff
@@ -224,6 +224,94 @@ func handleBuildStarted(ctx context.Context, state *store.EngineState, action bu
...
+	isBuildSuccess := br.Error == nil
+
+	// Update all the build statuses for the current manifest.
+	for _, target := range mt.Manifest.TargetSpecs() {
+		status := mt.State.MutableBuildStatus(target.ID())
+		updateBuildStatus(status, br, results[target.ID()], isBuildSuccess)
+	}
+
+	// Update build statuses for duplicated image targets in other manifests.
+	// This ensures that those images targets aren't rebuilt.
```
maiamcc, line 270:
```suggestion
	// This ensures that those images targets aren't redundantly rebuilt.
```

pkg/model/build_reason.go
```diff
@@ -53,6 +61,7 @@ var translations = map[BuildReason]string{
 	BuildReasonFlagTriggerCLI:     "CLI Trigger",
 	BuildReasonFlagTriggerUnknown: "Unknown Trigger",
 	BuildReasonFlagTiltfileArgs:   "Tilt Args",
+	BuildReasonFlagChangedDep:     "New Image",
```
maiamcc, line 64: I'm not sure I as a user would understand what this meant... why not just "dependency updated" or something?

internal/engine/upper_test.go
```diff
@@ -2801,9 +2803,22 @@ func TestWatchManifestsWithCommonAncestor(t *testing.T) {
 	call = f.nextCall("m1 build2")
 	assert.Equal(t, m1.K8sTarget(), call.k8s())
 
+	imageStateM1I0 := call.state[m1.ImageTargets[0].ID()]
```
maiamcc, line 2806: this test seems of a piece with the tests above in `buildcontroller_test.go`, any reason it's here instead of there? nbd, just seems odd.

internal/engine/upper.go
nicks: if it makes you feel any better: the argument about what to name this relationship comes up on every build project.

"reverse deps" is kind of the term of art in this space, see:
https://docs.bazel.build/versions/master/query.html#rdeps
because even though it's awkward, it's at least unambiguous

internal/engine/upper.go
```diff
@@ -224,6 +224,94 @@ func handleBuildStarted(ctx context.Context, state *store.EngineState, action bu
...
+
+	// Suppose we built manifestA, which contains imageA depending on imageCommon.
+	//
+	// We also have manifestB, which contains imageB depending on imageCommon.
+	//
+	// We need to mark imageB as dirty, because it was not built in the manifestA
+	// build but its dependency was built.
+	engineState.ForEachMutableBuildStatusDependingOn(results, func(id, reverseDepID model.TargetID, mn model.ManifestName, reverseDepStatus *store.BuildStatus) {
+		if mn == mt.Manifest.Name {
+			// We're only worried about manifests that we didn't build.
```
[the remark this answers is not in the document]
nicks: done

internal/engine/upper.go
nicks: yep! this is vestigial

internal/engine/upper.go
nicks: done

internal/engine/upper_test.go
nicks: ya, probably. i think a lot of these need to be reorganized anyway

pkg/model/build_reason.go
nicks: i changed it to "image updated", since we don't really use the "dependency" language in other user-facing parts of tilt

internal/engine/upper.go
nicks: done

internal/engine/buildcontrol/build_control.go
maiamcc: ah yeah that makes sense!

PR #3363, opened 2020-05-21
tiltfile config: use pflag

### LINKED ISSUE #3360: Named args lost when specified after positional args
If a Tiltfile defines both a positional argument and a named argument, it seems like the named arg is lost if specified after the positional arg in the "tilt up" command.

For example, when running against the Tiltfile
```
config.define_string_list(
    "named",
    args = False,
)

config.define_string_list(
    "positional",
    args = True,
)

cfg = config.parse()

print(cfg.get("named", ["notfound"]))
```
The command:
`tilt up -- --named namedValue positionalValue`
Prints "["namedValue"]" which is expected.

But the command:
`tilt up -- positionalValue --named namedValue`
Causes "["notfound"]" to print, even though I'd expect both commands to have the same result.

landism:
Fixes #3360

### Problem

In a Tiltfile's `config.parse()`, as soon as it encounters any non-positional arguments, all remaining arguments are treated as non-positional

#### repro
in Tiltfile:
```
config.define_string('foo')
config.define_string_list('bar', args=True)
cfg = config.parse()
```
on cmd line:
`tilt up -- a --foo b`

#### expected
`cfg['foo'] == 'b'` and `cfg['bar'] == ['a']`

#### observed
`cfg['foo'] is undefined and `cfg['bar'] == ['a', '--foo', 'b']`

### Solution
Use [`pflag`](https://github.com/spf13/pflag) instead of golang's `flag`. Further discussion [here](https://github.com/tilt-dev/tilt/issues/3360#issuecomment-632236155).

### Result
1. tiltfile config parsing is consistent with the "main" tilt parsing (i.e., the cobra stuff that happens in `cmd` - cobra uses pflag), and most (AFAIK) unix tools.
2. this is a breaking change in that go's `flag.Parse()` accepts `-foo bar` *or* `--foo bar` to set `foo`, but pflag requires two dashes (i.e., `--foo`). I think this is fine: 1. if you run it with `-foo`, you get an error and usage output that says `--foo`, so it's a clear path to remedying 2. all our docs use two dashes, so this only breaks undocumented behavior.
3. This is a breaking change in that it's fixing the bug report and anyone who relies on the old behavior will be broken. This is more potentially controversial.

### DIFF
--- a/internal/tiltfile/config/config_def.go
+++ b/internal/tiltfile/config/config_def.go
@@ -2,12 +2,12 @@ package config
 
 import (
 	"bytes"
-	"flag"
 	"fmt"
 	"os"
 
 	jsoniter "github.com/json-iterator/go"
 	"github.com/pkg/errors"
+	flag "github.com/spf13/pflag"
 	"go.starlark.net/starlark"
 
 	"github.com/windmilleng/tilt/internal/tiltfile/starkit"
@@ -104,10 +104,16 @@ func (cd ConfigDef) parseArgs(args []string) (ret configMap, output string, err 
 			continue
 		}
 		fs.Var(ret[name], name, def.usage)
+		// for bools, make "--foo" equal to "--foo true"
+		if _, ok := ret[name].(*boolSetting); ok {
+			fs.Lookup(name).NoOptDefVal = "true"
+		}
 	}
 
 	err = fs.Parse(args)
 	if err != nil {
+		_, _ = fmt.Fprintf(w, "Error parsing tiltfile config args: %v\nUsage:\n", err)
+		fs.PrintDefaults()
 		return nil, w.String(), err
 	}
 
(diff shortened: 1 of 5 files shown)


internal/tiltfile/config/config_def.go
```diff
@@ -104,10 +104,16 @@ func (cd ConfigDef) parseArgs(args []string) (ret configMap, output string, err
...
+		// for bools, make "--foo" equal to "--foo true"
+		if _, ok := ret[name].(*boolSetting); ok {
+			fs.Lookup(name).NoOptDefVal = "true"
+		}
 	}
 
 	err = fs.Parse(args)
 	if err != nil {
+		_, _ = fmt.Fprintf(w, "Error parsing tiltfile config args: %v\nUsage:\n", err)
+		fs.PrintDefaults()
```
landism, line 116: pflag advertises itself as a drop-in replacement for flag, but it doesn't print usage on error by default (though it does if you set it to os.Exit or panic on error)

landism:
> fix bug report, which is probably backwards-incompatible

What is this referring to?

nicks:
re: "What is this referring to?"
https://github.com/tilt-dev/tilt/issues/3360

landism:
> #3360

👍

PR #3374, opened 2020-05-22
engine: build and deployers should return results for targets they built, and

nicks:
Hello @landism, @maiamcc,

Please review the following commits I made in branch nicks/noop:

a935df30053134965d80e454c480059308eb50b1 (2020-05-22 18:03:01 -0400)
engine: build and deployers should return results for targets they built, and not return results for targets they didn't build.
This should make https://github.com/tilt-dev/tilt/pull/3362
much simpler and easy to reason about.

Also adds test harnesses for different build graph configurations.

4fb88080da67e01cb8b18675d12965c1bf92ae8e (2020-05-22 16:40:23 -0400)
engine: remove last build state from the target queue build handlers

### DIFF
--- a/internal/engine/build_and_deployer.go
+++ b/internal/engine/build_and_deployer.go
@@ -18,11 +18,15 @@ import (
 
 type BuildAndDeployer interface {
 	// BuildAndDeploy builds and deployed the specified target specs.
-	//
-	// Returns a BuildResult that expresses the outputs(s) of the build.
+
+	// Returns a BuildResultSet containing output (build result and associated
+	// file changes) for each target built in this call. The BuildResultSet only
+	// contains results for newly built targets--if a target was clean and didn't
+	// need to be built, it doesn't appear in the result set.
 	//
 	// BuildResult can be used to construct a set of BuildStates, which contain
-	// the last successful builds of each target and the files changed since that build.
+	// the last successful builds of each target and the files changed since that
+	// build.
 	BuildAndDeploy(ctx context.Context, st store.RStore, specs []model.TargetSpec, currentState store.BuildStateSet) (store.BuildResultSet, error)
 }
 
--- a/internal/engine/buildcontrol/target_queue.go
+++ b/internal/engine/buildcontrol/target_queue.go
@@ -90,18 +90,37 @@ func NewImageTargetQueue(ctx context.Context, iTargets []model.ImageTarget, stat
 	}, nil
 }
 
-func (q *TargetQueue) Results() store.BuildResultSet {
-	return q.results
+// New results that were built with the current queue. Omits results
+// that were re-used previous builds.
+//
+// Returns results that the BuildAndDeploy contract expects.
+func (q *TargetQueue) NewResults() store.BuildResultSet {
+	newResults := store.BuildResultSet{}
+	for id, result := range q.results {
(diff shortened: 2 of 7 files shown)


nicks:
rebased!

internal/engine/buildcontrol/target_queue.go
```diff
@@ -90,8 +90,24 @@ func NewImageTargetQueue(ctx context.Context, iTargets []model.ImageTarget, stat
 	}, nil
 }
 
-func (q *TargetQueue) Results() store.BuildResultSet {
-	return q.results
+// New results that we built with a BuildHandler. Omits results
+// that were re-used previous builds.
```
maiamcc, line 94: I don't care HOW we built them (aren't all results built with a BuildHander?), I care WHEN we built them. "Results that were build in the current call (omits results...)" or something?

internal/engine/buildcontrol/target_queue.go
```diff
@@ -90,8 +90,24 @@ func NewImageTargetQueue(ctx context.Context, iTargets []model.ImageTarget, stat
...
+func (q *TargetQueue) NewResults() store.BuildResultSet {
+	newResults := store.BuildResultSet{}
+	for id, result := range q.results {
+		if q.IsDirty(id) {
```
maiamcc, line 98: I understand why this works but the verbiage is confusing ("IsDirty" makes me think it STILL needs a build??). Not a big deal to leave as-is but i wonder if there exists a comment that would clarify?
```suggestion
		if q.IsDirty(id) {  // i.e. WASDirty() -- if True, we built this target afresh
```

internal/engine/image_build_and_deployer.go
```diff
@@ -156,14 +156,20 @@ func (ibd *ImageBuildAndDeployer) BuildAndDeploy(ctx context.Context, st store.R
 		anyLiveUpdate = anyLiveUpdate || !iTarget.LiveUpdateInfo().Empty()
 		return store.NewImageBuildResult(iTarget.ID(), refs.LocalRef, refs.ClusterRef), nil
 	})
+
+	newResults := q.NewResults()
 	if err != nil {
-		return store.BuildResultSet{}, buildcontrol.WrapDontFallBackError(err)
+		return newResults, buildcontrol.WrapDontFallBackError(err)
```
maiamcc, line 162: hmmm so this results set ought to only contain results for targets we actually built this time, IF we built them successfully, right? (just checking my understanding)

internal/engine/build_and_deployer.go
```diff
@@ -23,6 +23,10 @@ type BuildAndDeployer interface {
 	//
```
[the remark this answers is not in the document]
nicks: done

internal/engine/buildcontrol/target_queue.go
nicks: changed it to a private method and renamed it isBuilding

internal/engine/image_build_and_deployer.go
nicks: yep

internal/engine/buildcontrol/target_queue.go
nicks: done

Issue #3370, opened 2020-05-22
helm 3 doesn't work without name= param


nicks:
Repro steps:
1) helm fetch stable/grafana --untar --untardir ./
2) Add a Tiltfile like so:

```
k8s_yaml(helm('./grafana'))
```

Expected result:
Tilt should successfully deploy grafana

Actual result:
Error message:
```
The ServiceAccount "RELEASE-NAME-grafana" is invalid: metadata.name: Invalid value: "RELEASE-NAME-grafana"
```

https://cloud.tilt.dev/snapshot/AfjDwOwL8Vp5I4zniYw=

I think the problem is that the helm() builtin uses `--generate-name`, but that doesn't actually work with `helm template`

nicks:
(the obvious work around is just to pass a specific name)

PR #3375, opened 2020-05-22
tiltfile: default helm release name to 'tilt'. Fixes https://github.com/tilt-dev/tilt/issues/3370

nicks:
Hello @jazzdan,

Please review the following commits I made in branch nicks/ch7120:

9019d5b21aff073142276147d14c51922bea91f6 (2020-05-22 18:54:50 -0400)
tiltfile: default helm release name to 'tilt'. Fixes https://github.com/tilt-dev/tilt/issues/3370

### DIFF
--- a/internal/tiltfile/files.go
+++ b/internal/tiltfile/files.go
@@ -177,17 +177,18 @@ func (s *tiltfileState) helm(thread *starlark.Thread, fn *starlark.Builtin, args
 
 	var cmd []string
 
+	if name == "" {
+		// Use 'chart' as the release name, so that the release name is stable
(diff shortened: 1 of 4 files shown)


nicks:
The reason this is important is because RELEASE-NAME is not a valid k8s object name, so lots of helm charts don't work if you don't specify this. `helm template --generate-name` doesn't do anything

nicks:
on second thought, I'm going to change the default to "chart". This seems to be what Helm does. I'm slightly worried that if we use "tilt-", people will think it's a tilt system thing

PR #3376, opened 2020-05-26
hud: store display names for k8s resources as well as YAML resources

jazzdan:
So that they can be displayed on the frontend


pkg/webview/view.proto
```diff
@@ -66,6 +66,8 @@ message K8sResourceInfo {
 
   // The span id for this pod's logs in the main logstore
   string span_id = 9;
+
+  repeated string k8s_display_names = 10;
```
nicks, line 70: consider removing the `k8s` part of this, since this is already inside a K8sResourceInfo

internal/hud/view/view.go
```diff
@@ -57,6 +57,7 @@ type K8sResourceInfo struct {
 	PodRestarts        int
 	SpanID             logstore.SpanID
 	RunStatus          model.RuntimeStatus
+	DislayNames        []string
```
nicks, line 60: DislayNames -> DisplayNames

pkg/webview/view.proto
nicks: later on in this file, you have a YAMLResourceInfo. is the ultimate goal to get rid of that?

jazzdan: YAMLResourceInfo and K8sResourceInfo are, sadly, two different things. Each resource is supposed to only have one a K8ResourceInfo, YAMLResourceInfo, LocalResourceInfo, etc

See here https://github.com/tilt-dev/tilt/blob/dmiller/ch6890/transfer-display-names-to-web-ui/internal/hud/view/view.go#L129:L129

PR #3377, opened 2020-05-26
tiltfile: add support for two docker_compose calls in the same tiltfile

nicks:
Hello @maiamcc, @jazzdan,

Please review the following commits I made in branch nicks/ch7144:

db0ea97f0b8d2d043073254ec64e5251435b19a2 (2020-05-26 17:42:18 -0400)
tiltfile: add support for two docker_compose calls in the same tiltfile


nicks:
someone complained about this on twitter. The error message isn't great (it doesn't tell you that you can do `docker_compose([file1, file2])`). But it seemed easy enough to fix

PR #3380, opened 2020-05-26
Add .kate-swp files to ignore pattern

### LINKED ISSUE #3378: Add kate swp files to ignore patterns
Kate (the default kde text editor) uses a variation of Vim's `.swp` files called `.*.kate-swp`. While it has the ability to change where the .kate-swp files are created, it's not on by default and actually a little tricky to find even if you are looking for it. I think for ootb experience, it might be nice to add `.*.kate-swp` to the ignore list in `tilt`.

I looked at the code for the editor and don't see anything like is mentioned in the `tilt` code about vim "incrementing" the extension (so, no `.kate.swo` files, etc).

iggy:
KDE's text editor, kate, uses a file similar to Vim's .swp files. Ignore
these files so we don't rebuild on every keypress.

Fixes #3378


victorwuky:
Hi @iggy . This is now available in the latest release: https://github.com/tilt-dev/tilt/releases/tag/v0.14.1

Issue #3359, opened 2020-05-20
Tilt becomes sluggish after log size grows very large


andymartin-sch:
Our developers will occasionally leave Tilt running for a long time and/or launch services in DEBUG mode and a lot of content will be logged. After a while, so much content is logged that Tilt becomes sluggish and other issues like https://github.com/tilt-dev/tilt/issues/3194 appear.

It would be nice if [k8s_resource](https://docs.tilt.dev/api.html#api.k8s_resource) was extended to allow developers to set a maximum number of log lines for Tilt to retain on a per-service basis (and/or if there was a global flag to set a default to be applied across all services).

This would potentially also resolve some of the drivers behind https://github.com/tilt-dev/tilt/issues/1885

nicks:
Hmmm...Tilt already puts a maximums on both the log store (how many logs are stored) and the rendering engine (how many logs are displayed). It also has a protocol for streaming logs incrementally. So I'm curious what part of this is falling down.

Do you have a sense of what part is sluggish? Is it the Tilt process itself? Is it the browser? If you close the browser window, does the issue go away? etc.

andymartin-sch:
> Do you have a sense of what part is sluggish? Is it the Tilt process itself? Is it the browser? If you close the browser window, does the issue go away? etc.

I think it's Tilt itself as `tilt dump webview` can take quite a while to complete.

The browser UI experience is that it can take a long time (~30s) for the Tilt UI to load, although switching between services after the initial loading is done seems to still be pretty snappy.

I repro'd things with a service that just logged a bunch of stuff and this is how `tilt dump webview` slowed down as the logs grew:
```
$ time tilt dump webview | wc
   3584   13264  225508

real    0m0.077s
user    0m0.064s
sys     0m0.019s

$ time tilt dump webview | wc
 810325 1631657 23574473
 
real    0m3.892s
user    0m1.584s
sys     0m0.425s

$ time tilt dump webview | wc
1614693 3243812 46831984

real    0m11.490s
user    0m4.653s
sys     0m6.437s

$ time tilt dump webview | wc
5628232 11271986 163124286

real    1m26.399s
user    0m20.864s
sys     1m9.538s
```

In the "real world" case that the engineer sent in, his webview was roughly as large as the final example above:
```
$ tilt dump webview | wc
 3342358 7107448 110867384
```

nicks:
[...]

PR #3395, opened 2020-06-01
hud/server: enable websocket compression

jazzdan:
May help with situations where Tilt is being run on a remote server and
the web UI is running on a local laptop resulting in the websocket
messages being sent over the Internet like in #3359.

Currently seeing if there is a measurable performance impact here, but I figure if it's a wash, then we might as well enable this. WDYT?


jazzdan:
Yeah I poked around in perf and didn't notice an appreciable increase in CPU, or a disproportionate amount of time being spend compressing data.

PR #3400, opened 2020-06-03
web: error modal when snapshot fails

jazzdan:
(Video is cropped a little strangely, sorry about that)

### DIFF
--- a/web/src/ErrorModal.tsx
+++ b/web/src/ErrorModal.tsx
@@ -1,0 +1,31 @@
+import React, { PureComponent } from "react"
+import Modal from "react-modal"
+import { ShowErrorModal } from "./types"
+import "./FatalErrorModal.scss"
+
+type props = {
+  error: string | undefined
+  showErrorModal: ShowErrorModal
+  handleClose: () => void
+}
+
+export default class ErrorModal extends PureComponent<props> {
+  render() {
+    let showModal =
+      Boolean(this.props.error) &&
+      (this.props.showErrorModal === ShowErrorModal.Default ||
+        this.props.showErrorModal === ShowErrorModal.Show)
+    return (
+      <Modal
+        isOpen={showModal}
+        className="FatalErrorModal"
+        onRequestClose={this.props.handleClose}
+      >
+        <h2 className="FatalErrorModal-title">Error</h2>
+        <div className="FatalErrorModal-pane">
+          <p>Tilt has encountered an error: {this.props.error}</p>
+        </div>
+      </Modal>
+    )
+  }
+}
--- a/web/src/ErrorModal.tsx
+++ b/web/src/ErrorModal.tsx
@@ -24,6 +24,30 @@ export default class ErrorModal extends PureComponent<props> {
         <h2 className="FatalErrorModal-title">Error</h2>
         <div className="FatalErrorModal-pane">
           <p>Tilt has encountered an error: {this.props.error}</p>
+          <p>
+            To get help try{" "}
+            <a href="https://github.com/tilt-dev/tilt/issues/new">
+              opening a GitHub issue
+            </a>{" "}
+            or{" "}
+            <a href="https://docs.tilt.dev/debug_faq.html#where-can-i-ask-questions">
+              contacting us on Slack
+            </a>
+            .
+          </p>
+        </div>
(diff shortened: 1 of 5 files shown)


victorwuky:
Thanks for creating this @jazzdan . This is much better than a silent error.

Can we make the error message more actionable for the user. Perhaps one of these:
- Ask them to copy and paste the error into a new Tilt issue so that we can investigate.
- Refer them to an existing Tilt issue to copy and paste the error. (Especially if we know what the error is, and we are planning to fix it in the future.)

Regarding the "unexpected character" error message. Can the user actually do anything with that information? If they can't, how about providing less information, such as "JSON.parse error". It seems that providing more detail that's not actionable would seem more frustrating to the user.

web/src/HUD.tsx
```diff
@@ -190,9 +194,25 @@ class HUD extends Component<HudProps, HudState> {
...
+            if (err instanceof Error) {
+              this.setAppState({
+                showSnapshotModal: false,
+                error: `Error sending snapshot: ${err.message}`,
+              })
+            } else {
+              this.setAppState({
+                showSnapshotModal: false,
+                error:
+                  "Error sending snapshot. Check JavaScript console for more details",
```
victorwuky, line 208: @jazzdan : I don't think we should be asking a user to check the browser console for more details. If it's important for the user to see the details, can't we surface that information inline in the error here? I don't think we should be asking the user to debug the problem for us.

jazzdan: Yeah I'm not an expert in JS error handling but it seems like sometimes this error value isn't very helpful. Maybe I should do a `try`/`catch` too?

jazzdan:
@victorwuky happy to add a message telling folks where to go for help, but I think being super smart about this (finding existing issues/parsing out the error and figuring out what kind of error it is) isn't worth it right now. I think that'd be a fair bit of work and shouldn't delay showing people the error.

victorwuky:
> @victorwuky happy to add a message telling folks where to go for help, but I think being super smart about this (finding existing issues/parsing out the error and figuring out what kind of error it is) isn't worth it right now. I think that'd be a fair bit of work and shouldn't delay showing people the error.

Sounds good @jazzdan. I think directing folks to https://github.com/tilt-dev/tilt/issues or #tilt in http://slack.k8s.io/ would be great. (Or even https://tilt.dev/contact, though I think that may be removed since it's not in https://tilt.dev/index2.) Whatever you think makes more sense.

web/src/HUD.tsx
victorwuky: @jazzdan : I was just trying to say I don't think we should show the words "Check JavaScript console for more details" to the user in their regular error flow. If that doesn't actually appear to the user, then it's fine. But if it does appear, I think we should remove it. 

I understand our users our technical and would certainly understand the message. But I don't think we should be exposing this in the main UX itself.

So simply:

```suggestion
                  "Error sending snapshot.",
```

Don't know what you mean by try/catch here. You'd have to ask someone else!

jazzdan:
Now it looks like this:

Screenshot from 2020-06-03 14-01-49

web/src/ErrorModal.tsx
```diff
@@ -0,0 +1,55 @@
...
+            <a href="https://github.com/tilt-dev/tilt/issues/new">
+              opening a GitHub issue
+            </a>{" "}
+            or{" "}
+            <a href="https://docs.tilt.dev/debug_faq.html#where-can-i-ask-questions">
+              contacting us on Slack
+            </a>
+            .
+          </p>
+        </div>
```
victorwuky, line 51: Copy and paste error @jazzdan ?

```suggestion
```

victorwuky:
Looks perfect @jazzdan ! Thanks! (Except for that small dupe problem.)

nicks:
re: the error message - ya, I think right now we expect the error message to always be non-actionable garbage. The problem is that the server does not report back errors in an intelligible way.

For now, can we just hard-code "Unknown error".

Then we can leave a TODO to decide on how the server should transmit error messages.

What I don't want to happen is that we write up a whole complicated decision tree that doesn't actually work, and then the next engineer who touches this code has to wonder why it doesn't work.

jazzdan:
@nicks done

PR #3401, opened 2020-06-03
feature: remove facets flag

jazzdan:


victorwuky:
@jazzdan: 

1. Is there any user-facing impact with this change? 
2. What caused this change? Is there a clubhouse story or a #tilt discussion somewhere?

Thanks.

jazzdan:
If you were setting the facets feature flag previously now that will be an error and you will have to delete that from your Tiltfile. Otherwise there's no user facing change @victorwuky

victorwuky:
Thanks @jazzdan

Issue #1928, opened 2019-07-27
Manual restart of pod from dashboard


majelbstoat:
Sometimes pods get into a weird state and the best thing to do is just blow it away and start from scratch. Currently, the only way I know how to do this is to re-run `tilt up`, but this restarts all pods. (I also have a standalone rolling-update script which works by patching an environment variable into the container spec, but it would be nice to do it all from the same place.)

I'd love to be able to do this from within the cli dashboard, perhaps by selecting the workload and pressing `r`?

landism:
thanks for filing! this is a thing we've been talking about and want to implement

maiamcc:
Closing this in favor of #1012.

maiamcc:
Oh whoops my mistake, this is a slightly different case than #1012 (which wants a build-and-deploy, whereas this ticket wants to restart a pod; e.g. if there are no code changes since the last deploy, then a build-and-deploy wouldn't affect anything). Re-opening.

andymartin-sch:
We've been rolling out our Tilt setup to more developers on our team, and it's come up a few times that it was surprising that the ["Force Update" functionality](https://blog.tilt.dev/2019/11/14/force-update.html) didn't restart the pod if no changes were made to the image.

Offhand, I can't think of any reason why someone who clicks the Force Update button would *not* be okay with the pod being restarting, but maybe there is one? Either way, I think having an easy way to force a restart of a pod would line up well with the "'Turn it off and on again" philosophy mentioned in the blog post.

Right now our fallback is to tell the user to delete the pod with `kubectl`, but we are SO CLOSE to being able to let "normal" app developers never use any command-line tools during their development workflows, so it would be nice if we could eliminate this use case of `kubectl` with a friendly button in Tilt.

Let me know if any other info from our use case would help!

rpocase:
Just adding another use case to this - I'm working on integrating istio into certain profiles and actually have istio redeployed/reconfigured as part of the Tiltfile. E.g., adding sidecar injection. My options now are to either `tilt down` or individually shutdown pods. Being able to selectively redeploy a pod from the interface would be a bit easier as I work through issues with each service individually.

maiamcc:
Thanks for the extra info, both of you! As I'm thinking about this more: are either of you using live update? Because restarting a pod wipes out any in-place changes we've made during live update, this feature is more technically complicated than I originally thought.

[...]

PR #3413, opened 2020-06-05
engine: forcing an update should also force a pod update. Fixes https://github.com/tilt-dev/tilt/issues/1928

nicks:
Hello @jazzdan, @maiamcc,

Please review the following commits I made in branch nicks/ch4742:

7cf255f5f557623434cba8c07803638c83ceeb7d (2020-06-05 17:35:16 -0400)
engine: forcing an update should also force a pod update. Fixes https://github.com/tilt-dev/tilt/issues/1928


nicks:
This is obviously a very simple solution to a pretty complex problem.

Hitting the "force update" button will replace the existing pods, even if they haven't changed. I think this is closer to what most people expect from this button, based on the feedback I've heard.

I still think there's value in more full-featured solutions, but this has been outstanding for so long, I'm hoping that this will fix 90% of the problems

PR #3428, opened 2020-06-08
web: set overflow-wrap: anywhere on log lines

nicks:
Hello @hyu, @jazzdan,

Please review the following commits I made in branch nicks/ch4197:

4695f40ea2c5514d93da4aecd6dd0c5d8412eeb8 (2020-06-08 18:01:28 -0400)
web: set overflow-wrap: anywhere on log lines


nicks:
Screenshot from 2020-06-08 18-02-22

i don't think there's any great solution here. also wonder if we should consider smaller font sizes

PR #3429, opened 2020-06-08
tiltfile: object identifiers are not case sensitive

### LINKED ISSUE #3423: k8s_resource object selector error message makes no sense
I'm trying to use the k8s_resource(objects=) syntax and getting this error message:

```
No object identified by the fragment "queue-proxy:Image:knative-serving" could be found. Possible objects are: ... "queue-proxy:Image:knative-serving", ...
```

but...that exact object is in the list? What's going on?

Here's the snapshot:
https://cloud.tilt.dev/snapshot/AcKL8-0LhBMFIMSeWiw=

cc @jazzdan do you have any ideas on this?

jazzdan:
Fixes #3423


internal/tiltfile/tiltfile_state.go
```diff
@@ -655,7 +655,7 @@ func (s *tiltfileState) assembleK8sV2() error {
 			}
 
 			for i, o := range opts.objects {
-				entities, ok := fragmentsToEntities[o]
```
nicks, line 658: i'm surprised they're not stored in objects as lowercase?

internal/tiltfile/tiltfile_state.go
```diff
@@ -1331,28 +1331,44 @@ type k8sObjectSelector struct {
...
 // If an arg is an empty string it will become an empty regex that matches all input
 // Otherwise the arg must match the input exactly
-func newExactK8sObjectSelector(apiVersion string, kind string, name string, namespace string) (k8sObjectSelector, error) {
-	ret, err := newK8sObjectSelector(
-		exactOrEmptyRegex(apiVersion),
-		exactOrEmptyRegex(kind),
-		exactOrEmptyRegex(name),
-		exactOrEmptyRegex(namespace),
-	)
+func newExactCaseInsensitiveK8sObjectSelector(apiVersion string, kind string, name string, namespace string) (k8sObjectSelector, error) {
```
nicks, line 1334: I feel like the nomenclature here needs to be cleaned up a bit, in particular:
- All functions are case insensitive now, but only one of them has "CaseInsensitive" in the name. That seems weird?
- The usage of the word "exact" here is not something i've seen before. I think python calls this "FullMatch"?

internal/tiltfile/tiltfile_state.go
```diff
@@ -1331,28 +1331,44 @@ type k8sObjectSelector struct {
...
-	ret.namespaceString = namespace
+	ret.namespace, err = regexp.Compile(exactOrEmptyRegex(namespace))
+	if err != nil {
+		return k8sObjectSelector{}, errors.Wrap(err, "error parsing namespace regexp")
+	}
 
 	return ret, nil
 }
 
+func makeCaseInsensitive(s string) string {
```
nicks, line 1361: i feel like half the problem here is that we're passing around a regexp around as a DSL string, rather than something more structured, and i wonder if this code would be easier to follow if we had something like

```
type re struct {
  pattern string
  ignoreCase bool
  fullMatch bool
}

func (r re) compile() { ... }
```

internal/tiltfile/tiltfile_state.go
jazzdan: ooh I like FullMatch. And I agree, I'll clean this stuff up.

internal/tiltfile/tiltfile_state.go
jazzdan: Yeahhh I tried something like this back when I was originally writing this a couple weeks ago. I'll give it another shot with fresh eyes.

internal/tiltfile/tiltfile_state.go
jazzdan: Hmm it might be weird to use a different string than the one the user entered in error messages though, even if we treat it as case insensitive. I think I want to preserve the original string.

landism:
Did you end up finding a source or some evidence for whether k8s allows two objects to have the same name with different casing? It's probably worth writing down the evidence/assumption, given we were unclear on it.

nicks:
Here are some references:
https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#api-conventions
https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names

As a general rule of thumb, Kubernetes tends to convert everything to lowercase, since they end up getting converted to case-insensitive HTTP endpoints

jazzdan:
Yeah that stuff is what I went off of.

I'm still a little unclear on kind names, but it seems that most everything has been lowercase in my testing.

Like in this part of the docs

>Resource collections should be all lowercase and plural, whereas kinds are CamelCase and singular. Group names must be lower case and be valid DNS subdomains.

It doesn't explicitly say that kind names are case insensitive, but I wasn't able to test this myself yet.

nicks:
Ya, I think the API docs are being super-precise in a way that's helpful if you understand the internals, but hard to parse if you do not.

There are two different things:
- kinds (which require camel case)
- resources (which are exposed as lowercase)
and there's a mapping from Kind -> resource that converts to lowercase

but to a user (e.g., someone who's just using kubectl and writing yaml), because there's a 1<>1 mapping, it's easy to read that as "oh, it's case-insensitive", when what's actually happening is that you're interacting with different things.

Does that make sense?

jazzdan:
Yeahhh that makes sense. Thanks @nicks!

jazzdan:
I updated some nomenclature but left the refactor as a TODO because, once again, I ran in to weird test failures. I think we're depending on some regex behavior that is unclear to me and super finicky.

Issue #3410, opened 2020-06-05
Provide more context for "Image not used in any deploy config"


jazzdan:
A common error that people encounter when using Tilt is "Image not used in any deploy config". It looks like this:

```
Image not used in any deploy config:
    ✕ 727006795293.ecr.us-east-1.amazonaws.com/foo
```
This can be really confusing because if you have a complicated set up it won't be obvious where `727006795293.ecr.us-east-1.amazonaws.com/foo` is coming. It might be being generated in your Tiltfile via a `local` and could also be interacting with [`default_registry`](https://docs.tilt.dev/personal_registry.html). In addition it won't be obvious where to look for image in your YAML. It's likely that the YAML is being templated if not outright generated. Is the image name an environment variable, computed somewhere else, in the Helm chart?

I think it would help if we listed the image namees that Tilt knows about alongside this error message? So like:

```
Image not used in any deploy config:
    ✕ 727006795293.ecr.us-east-1.amazonaws.com/foo
Images that Tilt has seen in all deploy configs:
    *  foo
```

But there are probably other/better ways to solve this too.

cc @jgoulah

nicks:
For reference, this is what the error message currently looks like:

```
Image not used in any deploy config:
    ✕ gcr.typo.io/foo
Did you mean…
    - gcr.io/foo
Skipping this image build
```

A common problem I've seen is that the user doesn't know what a "Deploy config" is, and doesn't know that they're supposed to be providing k8s configs. is that what's happening here?

jazzdan:
@nicks good point, but that's only true if we find any similar matches. I think this is a case where a misconfiguration is a causing `k8s_yaml` for `gcr.io/foo` to not get called at all.

jgoulah:
in this case it was that our yaml being passed to k8s_yaml() wasn't being captured correctly due to a bug in our stack, I think any additional output would be great though as this error has popped up for various reasons

victorwuky:
@jgoulah : This is now available in the latest Tilt release: https://github.com/tilt-dev/tilt/releases/tag/v0.14.3

PR #3430, opened 2020-06-09
tiltfile: improve the unmatched image warning message. Fixes https://github.com/tilt-dev/tilt/issues/3410

nicks:
Hello @jazzdan,

Please review the following commits I made in branch nicks/ch7422:

fc66199a3b962ef57b06727d805d5c35ef66023b (2020-06-09 12:28:48 -0400)
tiltfile: improve the unmatched image warning message. Fixes https://github.com/tilt-dev/tilt/issues/3410


internal/tiltfile/tiltfile_state.go
```diff
@@ -535,6 +541,35 @@ func (s *tiltfileState) assemble() (resourceSet, []k8s.K8sEntity, error) {
...
 }
 
+// Emit an error if there are unmatches images.
+//
+// There are 3 mistakes people commonly make if they
+// have unmatched images:
+// 1) They didn't include any Kubernetes or Docker Compose configs at all.
+// 2) They included Kubernetes configs, but they're custom resources
+//    and Tilt can't infer the image.
+// 3) They typo'd the image name, and need help finding the right name.
```
jazzdan, line 551: The one that @jgoulah seemed to be encountering is a slight variation on this, because of misconfiguration the YAML that should have consumed this was never included.

I don't think it's possible to account for this at this point in the code, I think this is more of a job for facets. Just wanted to note it here.

jgoulah: its similar to a typo but it also pops up when helm values don't render the image tag or if tht tag isn't what tilt expects

nicks: ya, i added some info to this to the comment. I have ideas on tooling to help with this, though it's slightly different than the facets approach. stay tuned!

PR #3437, opened 2020-06-10
engine: use regular logger when logging during process shutdown

jazzdan:
Otherwise nothing gets logged


jazzdan:
Closing due to #3443 being closer to the root cause.

PR #3440, opened 2020-06-11
Added Deprecate Property to Watch Flag (Issue  #3256) 

### LINKED ISSUE #3256: Remove the --watch flag for `tilt up`
`tilt up` has a flag `--watch`

Originally, --watch=false was the default. Tilt wouldn't watch any files or watch any Kubernetes resources.

We realized very early on that the default behavior should be `--watch=true`, because Tilt really should be watching your changes and telling you if your deployments have problems.

As we've added more features that only work if Tilt is watching Kubernetes (like resource_deps), the  behavior of --watch=false doesn't really make sense. It's mostly obsoleted by `tilt ci`  - see [docs](https://docs.tilt.dev/cli/tilt_ci.html). We should just remove the `--watch` flag.

abdullahzameek:
Fixes  #3256

I've added the deprecate property to the watch flag. The result of doing so is as follows: 

When "--watch" is used:
depracated-terminal

When there's no "--watch" included:
nowatch

### DIFF
--- a/internal/cli/up.go
+++ b/internal/cli/up.go
@@ -12,6 +12,9 @@ import (
 	"github.com/mattn/go-isatty"
 	"github.com/opentracing/opentracing-go"
 	"github.com/spf13/cobra"
+	"golang.org/x/sync/errgroup"
+	"k8s.io/klog"
+
 	"github.com/tilt-dev/tilt/internal/analytics"
 	"github.com/tilt-dev/tilt/internal/cloud"
 	engineanalytics "github.com/tilt-dev/tilt/internal/engine/analytics"
@@ -24,8 +27,6 @@ import (
 	"github.com/tilt-dev/tilt/pkg/logger"
 	"github.com/tilt-dev/tilt/pkg/model"
 	"github.com/tilt-dev/tilt/web"
-	"golang.org/x/sync/errgroup"
-	"k8s.io/klog"
 )
 
 const DefaultWebHost = "localhost"
--- a/internal/cli/up.go
+++ b/internal/cli/up.go
@@ -100,6 +100,8 @@ local resources--i.e. those using serve_cmd--are terminated when you exit Tilt.
 		c.hudFlagExplicitlySet = cmd.Flag("hud").Changed
 	}
 
+	cmd.Flag("watch").Deprecated = "it will be removed in future releases."
+
 	return cmd
 }
 
--- a/internal/cli/up.go
+++ b/internal/cli/up.go
@@ -51,6 +51,9 @@ type upCmd struct {
 	hud        bool
 	// whether hud was explicitly set or just got the default value
 	hudFlagExplicitlySet bool
+
+	//whether watch was explicitly set in the cmdline
+	watchFlagExplicitlySet bool
 }
 
 func (c *upCmd) register() *cobra.Command {
@@ -100,8 +103,13 @@ local resources--i.e. those using serve_cmd--are terminated when you exit Tilt.
 		c.hudFlagExplicitlySet = cmd.Flag("hud").Changed
 	}
 
+	//since the --watch flag isn't really useful and has been made obselete by tilt ci, print a deprecation warning.
 	cmd.Flag("watch").Deprecated = "it will be removed in future releases."
-
+	//if a deprecation message exists, then watch was explicitly set by the user, so mark the watchFlagExplicitlySet var as true for
+	//logging the message on the web UI and TUI
+	if len(cmd.Flag("watch").Deprecated) > 0 {
+		c.watchFlagExplicitlySet = true
+	}
 	return cmd
 }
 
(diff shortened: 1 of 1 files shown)


abdullahzameek:
> Nice! I didn't even know about this API.
> 
> Does it work ok with --hud=true?

@nicks So, for --hud=true, it just shows up like this: 
Screenshot from 2020-06-11 18-03-27

It doesn't show up in the actual hud logs

nicks:
oof, that's frustrating.

if you want to merge this pr as-is and iterate on it, that's fine.

if you look at how `tilt up` is implemented, it redirects logs here:
https://github.com/tilt-dev/tilt/blob/master/internal/cli/up.go#L135
so that the logs still work no matter what mode you're in. 

What if we checked if that flag was set below that line, and logged the message there?

abdullahzameek:
> oof, that's frustrating.
> 
> if you want to merge this pr as-is and iterate on it, that's fine.
> 
> if you look at how `tilt up` is implemented, it redirects logs here:
> https://github.com/tilt-dev/tilt/blob/master/internal/cli/up.go#L135
> so that the logs still work no matter what mode you're in.
> 
> What if we checked if that flag was set below that line, and logged the message there?

I'm reading through it, I'll try to fix it and then merge it all in one go!

abdullahzameek:
@nicks Managed to put the warning into the logs! 

Screenshot from 2020-06-11 23-09-14

If you view it from the TUI though, you see the message being printed twice - once called by flag.go file at https://github.com/tilt-dev/tilt/blob/master/vendor/github.com/spf13/pflag/flag.go#L485 and then again at when the logger output is called. So, if you do --hud=false, you'd see this: 
Screenshot from 2020-06-11 23-13-38

I'm not really sure how I can disable the function from the flag.go from printing the warning, other than by modifying the code from there itself, but that seems like a bad idea since its an external library...

internal/cli/up.go
```diff
@@ -100,6 +103,13 @@ local resources--i.e. those using serve_cmd--are terminated when you exit Tilt.
 		c.hudFlagExplicitlySet = cmd.Flag("hud").Changed
 	}
 
+	//since the --watch flag isn't really useful and has been made obselete by tilt ci, print a deprecation warning.
+	cmd.Flag("watch").Deprecated = "it will be removed in future releases."
+	//if a deprecation message exists, then watch was explicitly set by the user, so mark the watchFlagExplicitlySet var as true for
+	//logging the message on the web UI and TUI
+	if len(cmd.Flag("watch").Deprecated) > 0 {
```
nicks, line 110: i don't think this is doing what you expect...this will always be true, because you just set it above.

we probably shouldn't use the `Deprecated` field at all, because we don't want the flag library to do its own printing?

abdullahzameek: Ohh right, I assumed that the `Deprecated` property is only invoked when the user explicitly uses the flag in the command line, and that it would just be an empty string otherwise. Turns out that is not the case, my bad. 

Yeah, I think I'll look out for another way to do this because of the flag library printing. 

(Also, apologies for the late reply, it was getting late yesterday and I was a bit tired)

abdullahzameek: I think it's good now! I got too hung up on utilizing the `Deprecated` field (since I thought that was the best way to get about it) but turns out, it was just a matter of adding another line under the existing logic: 
`cmd.PreRun = func(cmd *cobra.Command, args []string) {
		c.hudFlagExplicitlySet = cmd.Flag("hud").Changed
		c.watchFlagExplicitlySet = cmd.Flag("watch").Changed
	}

`
Its printing the deprecation warning on the logs both on the TUI and the Web UI properly now (and doesn't print the message when the flag isn't present) 

Screenshot from 2020-06-12 10-16-47

Screenshot from 2020-06-12 10-17-14

internal/cli/up.go
```diff
@@ -98,6 +101,7 @@ local resources--i.e. those using serve_cmd--are terminated when you exit Tilt.
 
 	cmd.PreRun = func(cmd *cobra.Command, args []string) {
 		c.hudFlagExplicitlySet = cmd.Flag("hud").Changed
+		c.watchFlagExplicitlySet = cmd.Flag("watch").Changed
```
nicks, line 104: nice solution!

internal/cli/up.go
```diff
@@ -136,6 +140,11 @@ func (c *upCmd) run(ctx context.Context, args []string) error {
 
 	logOutput(fmt.Sprintf("Starting Tilt (%s)…", buildStamp()))
 
+	//if --watch was set, warn user about deprecation
+	if c.watchFlagExplicitlySet {
+		logOutput("Flag --watch has been deprecated, it will be removed in future releases.")
```
nicks, line 145: fyi, i think you can also use logger.Get(ctx).Warnf(...) if you want it to show up as a warning in the UI

abdullahzameek: Warnf() seems to be doing something interesting - it prints the warning fine on the TUI, but the keyword "warning" doesn't show up on the Web UI for some reason.

Screenshot from 2020-06-12 17-56-14

Screenshot from 2020-06-12 17-57-04

nicks: yep! it has the yellow UI bar in the web, and skips the prefix

PR #3453, opened 2020-06-12
web: fix fallback log styling when there are debug logs

nicks:
Hello @jazzdan, @hyu,

Please review the following commits I made in branch nicks/ch7547:

0dd5abbb56d7c63d899ad448c5523367b78000e8 (2020-06-12 14:42:30 -0400)
web: fix fallback log styling when there are debug logs

### DIFF
--- a/web/src/LogPaneLine.scss
+++ b/web/src/LogPaneLine.scss
@@ -20,11 +20,15 @@
   }
   &.is-buildEvent-fallback {
     background-color: $color-gray-darker;
-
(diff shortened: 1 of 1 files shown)


nicks:
Screenshot from 2020-06-12 14-41-59

still not great, but at least not totally broken

nicks:
we do some tricks with margin so that the fallback reason log (the stuff with the blue highlight) is flush against the build-reason bar

this tries to make sure we only do these tricks if we're close to the build-reason bar

PR #3461, opened 2020-06-15
tiltfile : Added Error when parameter to k8s_yaml() is None or '' 

abdullahzameek:
Added the error message by inspecting the value of the YAML entity from the Starlark value. If the value picked up by starlark is None or nil, it returns an empty array. Thus, checking if the array is empty should be sufficient(?) for checking the parameter to k8s_yaml()
Screenshot from 2020-06-16 13-52-55


abdullahzameek:
The current code can pick up when the value `None` is passed to `k8s_yaml()` and would respond with an appropriate error message. However, when it comes to `blob(' ')`, it got a bit complicated, because starlark returns a slice with an empty string. For the longest time, I did not realize that a slice with an empty string has a length of 1, which was why my code didn't catch the error. Added further logic to handle the empty string issue (it worked just fine on the blob(' ') case, but turns out that that logic causes all the helm tests to fail for some reason. I'm not really sure how helm works, so I'll probably come back to this issue later.

abdullahzameek:
So, right now, I've managed to make to tackle both the cases ( `None` and `blob('')`) and have written unit tests to verify that they work. However, when running the whole suite of tests, there were 6 tests that failed, and I noticed that they all had one thing in common - they dealt with invalid or empty YAML files in some form, which is what I believe the fix I wrote handles. So, I guess it's sort of expected that these would fail? 
Screenshot from 2020-06-17 22-22-51

CircleCI also says that one of the helm tests failed, but when I tried to reproduce it on my local machine, it passes. Any idea why this might be the case? 
Screenshot from 2020-06-17 22-27-10

internal/tiltfile/k8s.go
```diff
@@ -111,17 +111,27 @@ func (r k8sResource) refSelectorList() []string {
...
 		return nil, err
 	}
 
-	entities, err := s.yamlEntitiesFromSkylarkValueOrList(thread, yamlValue)
-	if err != nil {
-		return nil, err
+	value := starlarkValueOrSequenceToSlice(yamlValue)
+	if len(value) > 0 {
+		entities, err := s.yamlEntitiesFromSkylarkValueOrList(thread, yamlValue)
+		if len(entities) == 0 {
```
nicks, line 124: i think you just need to swap this `if` statement with the `if err != nil` below it.

A good rule of thumb in Go is to always handle the error first, then handle the other cases!

abdullahzameek: I will ink this in mind because I did just this and all the tests passed locally woohooo!

abdullahzameek:
@nicks Seems to be passing all the CI tests, awaiting your final review!

internal/tiltfile/k8s.go
```diff
@@ -111,17 +111,34 @@ func (r k8sResource) refSelectorList() []string {
...
+		val, _ := starlark.AsString(value[0])
+		entities, err := s.yamlEntitiesFromSkylarkValueOrList(thread, yamlValue)
+
+		if err != nil {
+			return nil, err
+		}
+
+		//the parameter blob('') results in an empty string
+		if len(entities) == 0 && val == "" {
+			return nil, fmt.Errorf("k8s_yaml: Empty or Invalid YAML Resource Detected")
```
nicks, line 135: i don't think you need the `k8s_yaml: ` prefix? the starlark infrastructure will already print the function name for you

internal/tiltfile/k8s.go
```diff
@@ -111,17 +111,34 @@ func (r k8sResource) refSelectorList() []string {
...
+		}
+		s.k8sUnresourced = append(s.k8sUnresourced, entities...)
+
+	} else {
+		return nil, fmt.Errorf("k8s_yaml: Empty or Invalid YAML Resource Detected")
```
nicks, line 140: ditto here

nicks:
hooray!

internal/tiltfile/k8s.go
abdullahzameek: Oh right, yeah it does : 
Screenshot from 2020-06-18 17-25-25

PR #3464, opened 2020-06-15
hud: add a prompt package for the new terminal mode

nicks:
Hello @landism, @jazzdan,

Please review the following commits I made in branch nicks/browser:

e503e1875abdb320492d5665b7541a807395db62 (2020-06-15 16:00:46 -0400)
hud: add a prompt package for the new terminal mode

### DIFF
--- a/internal/hud/prompt/prompt.go
+++ b/internal/hud/prompt/prompt.go
@@ -1,5 +1,142 @@
 package prompt
 
 import (
-	_ "github.com/mattn/go-tty"
+	"context"
+	"fmt"
+
+	tty "github.com/mattn/go-tty"
+	"github.com/pkg/browser"
+
+	"github.com/tilt-dev/tilt/internal/hud"
+	"github.com/tilt-dev/tilt/internal/store"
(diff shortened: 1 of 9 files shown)


nicks:
ack, i forgot what a pain dealing with TTY APIs is

internal/hud/prompt/prompt.go
```diff
@@ -1,5 +1,128 @@
...
+			r, err := t.ReadRune()
+			if err != nil {
+				st.Dispatch(store.ErrorAction{Error: err})
+				return
+			}
+			keyCh <- r
+		}
+	}()
+
+	for ctx.Err() == nil {
```
landism, line 111: IIUC, at the moment, this is just holding onto the subscriber's notify goroutine forever, which ~means its OnChange can only ever be called once (which then might surprise someone trying to edit this subscriber). It seems to work, but also seems different from how our other subscribers work, and potentially accidental (like, does this mostly make `p.printed` redundant?)

would it be better to put this for loop in a new goroutine? am I missing something?

PR #3473, opened 2020-06-16
engine: remove LastSuccessfulResult

nicks:
Hello @maiamcc,

Please review the following commits I made in branch nicks/lastresult:

2af94cc3d4a3b705c6fda5360a1afb2e0ecc4b3f (2020-06-16 18:49:40 -0400)
engine: remove LastSuccessfulResult
The fundamental problem is that, by definition, all concrete BuildResult objects
represent success. If an image build fails, we return a nil BuildResult.

The notion of "BuildResult" gets slippery when we talk about live update.
If the in-container RUN step failed, the live-updater returns a valid BuildResult
(containing the container id) AND an error. In some contexts, we want to
treat that as a success (because the container is still live and can get future
live udpates), but in other cases (like error reporting), we want to report
this as an error.

Better to just get rid of the concept of "SuccessfulResult" completely,
and be more precise about what we want.

### DIFF
--- a/internal/engine/upper_test.go
+++ b/internal/engine/upper_test.go
@@ -197,8 +197,16 @@ type fakeBuildAndDeployer struct {
 	nextDeployedUID           types.UID
 	nextPodTemplateSpecHashes []k8s.PodTemplateSpecHash
 
-	// Set this to simulate the build failing. Do not set this directly, use fixture.SetNextBuildFailure
-	nextBuildFailure error
+	// Set this to simulate a build with no results and an error.
+	// Do not set this directly, use fixture.SetNextBuildError
+	nextBuildError error
+
+	// Set this to simulate a live-update compile error
+	//
+	// This is slightly different than a compile error, because the containers are
+	// still running with the synced files. The build system returns a
+	// result with the container ID.
+	nextLiveUpdateCompileError error
 
 	buildLogOutput map[model.TargetID]string
 
@@ -250,6 +258,11 @@ func (b *fakeBuildAndDeployer) BuildAndDeploy(ctx context.Context, st store.RSto
 	}
 
 	defer func() {
+		b.mu.Unlock()
+
+		// block until we know we're supposed to resolve this build
+		b.waitUntilBuildCompleted(ctx, buildKey)
+
 		// don't update b.calls until the end, to ensure appropriate actions have been dispatched first
 		select {
 		case b.calls <- call:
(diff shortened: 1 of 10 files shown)


internal/engine/upper_test.go
```diff
@@ -198,9 +198,18 @@ type fakeBuildAndDeployer struct {
...
 	nextPodTemplateSpecHashes []k8s.PodTemplateSpecHash
 
-	// Set this to simulate the build failing. Do not set this directly, use fixture.SetNextBuildFailure
+	// Set this to simulate the build failing.
+	// Do not set this directly, use fixture.SetNextBuildFailure
 	nextBuildFailure error
 
+	// Set this to simulate a live-update user build failure.
+	//
+	// This is slightly different than a normal build failure,
```
maiamcc, line 207: ...wait, doesn't this belie [the comment from above](https://github.com/tilt-dev/tilt/pull/3473/files#diff-febc1ffa49840bcca6121206c409ab75R231) of "if this build failed, [lastResult] will be set to nil"? How do these interact? if there any way to tell just from the BuildState whether the last LiveUpdate succeeded or failed? (I suppose the follow-up question being, is there anywhere we NEED to know that?)

nicks: did you read the comment in the PR description?

I would honestly prefer we get rid of the whole nomenclature around "success" and "failure" in our build system. Different BuildAndDeployers have different opinions about what they mean in practice. But I think removing that nomenclature is a much bigger and more complex change.

nicks: (Bazel ran into this problem hard - they split it into "SUCCESS" and "a whole bunch of different non-success statuses")
https://github.com/bazelbuild/bazel/blob/master/src/main/java/com/google/devtools/build/lib/actions/SpawnResult.java#L36

nicks: (i particularly :heart: EXECUTION_FAILED vs EXECUTION_FAILED_CATASTROPHICALLY)

maiamcc:
Whoops sorry @nicks, you should tag someone else into this, I don't think I'll get to it before EOD and then I'm on vacation.

internal/engine/upper_test.go
maiamcc: hmm rereading the PR comment, I think the "if a build failed... set to nil" comment needs to be much more explicit about the new definition of "failed build" / that it only applies to image build results.

nicks:
no worries, i'll assign it to someone else! ya, i think a lot of the comments here in this code need to be improved, because they don't reflect how the code currently works (which to be fair, predates this pr)

nicks:
ok, changed a bunch of names to better reflect what this code does. @jazzdan can you take a look?

PR #3476, opened 2020-06-18
tiltfile: support auto_init on k8s_resource

landism:
I tried to write a k8s Job today and wished I had this


victorwuky:
@nicks: Should https://docs.tilt.dev/api.html#api.k8s_resource be updated here?

PR #3477, opened 2020-06-18
k8s: improve jsonpath api to allow both extraction and injection

nicks:
Hello @landism, @maiamcc,

Please review the following commits I made in branch nicks/jsonpath:

3426e7f199012fdab1267f1f9c7c7579a55ae86d (2020-06-18 15:31:42 -0400)
k8s: improve jsonpath api to allow both extraction and injection

### DIFF
--- a/internal/k8s/jsonpath/value.go
+++ b/internal/k8s/jsonpath/value.go
@@ -1,0 +1,42 @@
+package jsonpath
+
+import "reflect"
+
+// A wrapper around reflect.Value that allows callers
+// to call Set() on a Value retrieved from inside a map index.
+//
+// In the normal reflect package, the value in a map index
+// is not addressable, because the internal map implementation
+// might have to rearrange the map storage.
+//
+// In this implementation, we keep an extra reference to the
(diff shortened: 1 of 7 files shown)


nicks:
sorry, this turned out to be a bigger pain in the ass than I expected...in go, map field values are not addressable, so i had to introduce a new struct to do this

internal/k8s/jsonpath/value.go
```diff
@@ -0,0 +1,34 @@
+package jsonpath
+
+import "reflect"
+
+// A hack over reflect.Value that allows maps keys to be settable.
```
landism, line 5: I first read this as "lets you change the keys in a map" (like, change `{"foo": true}` to `{"bar": true}`), which I don't think is what it's doing - would "allows map values to be settable" or "allows values indexed by maps' keys to be settable" work?

nicks: good point, clarified the language here a bit

nicks:
good point, i missed the tests! copied them over on a separate PR

Issue #3478, opened 2020-06-18
auto_init=False makes tilt ci hang


damoon:
Using auto_init=False for local_resource makes tilt ci hang.

damoon:
i believe tasks with disabled auto start should be ignored during "tilt ci"

maiamcc:
Good catch, ty!

maiamcc:
This should be fixed in [the latest release](https://github.com/tilt-dev/tilt/releases/tag/v0.15.0) -- let us know if you run into any trouble!

damoon:
work for me :D
thanks for the fix

PR #3485, opened 2020-06-19
ci: ci mode shouldnt wait for resources with auto_init=false [ch7838]

maiamcc:
addresses #3478

Not sure that an integration test is the best way to verify this behavior--
@nicks are there better structures already in the code for testing ci mode?


internal/engine/exit/controller.go
```diff
@@ -55,6 +55,11 @@ func (c *Controller) shouldExit(store store.RStore) Action {
 		// If any of the resources are in error, exit.
 		allOK := true
 		for _, mt := range state.ManifestTargets {
+			// don't wait on resources requiring manual trigger for initial build
```
nicks, line 58: it might make sense to have a unit test for this

nicks: (see the tests under controller_test.go)

maiamcc:
Ah I didn't realize there was such solid unit testing for this code. In that case, I'd rather leave it as just a unit test and not add another integration test (tho if either of you are particularly into the idea of there being an integration test @nicks @landism I can put it back!)

PR #3495, opened 2020-06-22
prompt: emit a warning that the hud key shortcut is going away

nicks:
Hello @landism, @victorwuky,

Please review the following commits I made in branch nicks/ch6282/hide:

860cd5464891516001f43c6b75ef1c328e9bbd92 (2020-06-22 13:19:02 -0400)
prompt: emit a warning that the hud key shortcut is going away

### DIFF
--- a/internal/hud/prompt/prompt.go
+++ b/internal/hud/prompt/prompt.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"strings"
 	"sync"
+	"time"
 
 	"github.com/fatih/color"
 	tty "github.com/mattn/go-tty"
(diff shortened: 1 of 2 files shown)


internal/hud/prompt/prompt.go
```diff
@@ -203,8 +209,16 @@ func (p *TerminalPrompt) OnChange(ctx context.Context, st store.RStore) {
 					msg.stopCh <- true
 
 				case 'h':
+					warningMsg := "We're removing the HUD key shortcut from the prompt.\n" +
+						"In future releases, open the HUD with: tilt up --hud=true"
+					_, _ = fmt.Fprintf(p.stdout, "%s\n", warningMsg)
+
+					time.Sleep(p.hudDelay)
```
landism, line 216: A 6-second delay seems like it'd be really annoying! Would it be cheap to do something like "press h again to launch the HUD"?

Issue #3460, opened 2020-06-15
image injection is unaware of image_json_path


nicks:
e.g., create an instance of a CRD like this:
```
apiVersion: openfaas.tilt.dev/v1alpha1
kind: Function
metadata:
  name: addone
  namespace: openfaas-fn
spec:
  image: addone
```

and define a docker_build for `addone`

specify `k8s_kind('Function', image_json_path="{.spec.image}")` in your Tiltfile

The Tilt code for injecting into instances of CRDs just replaces the image name [anywhere it finds it](https://github.com/windmilleng/tilt/blob/5d5545054425dfb4f00e8f9a61bd039680062904/internal/k8s/image.go#L168), even though the Tiltfile says where the image is.

This results in .metadata.name being replaced with something like addone:tilt-12323435456, which makes it illegal. It also opens us up to incorrectly replacing other fields where the bug would not be so immediately apparent.

We should propagate the image_json_path to the image injector and only replace things at that path.

originally filed by @landism , rediscovered today when people were experimenting with knative in the tilt channel

Issue #3427, opened 2020-06-08
CRD image injection when repo and tag are expressed differently


nicks:
Most CRDs expect an image field where an image is injected. Tilt has an API for injecting the image:

```
  image: frontend:latest
```

But sometimes, Helm charts express the image repo and the image tag separately

```
image:
  repo: frontend
  tag: latest
```

Unfortunately, Tilt's API for discovering images in CRDs assumes one field (see https://docs.tilt.dev/api.html#api.k8s_kind and image_json_path)

We'd probably need something like
image_repo_json_path, image_tag_json_path

(requested by bkpandey in the tilt channel)

pandeybk:
Thank you @nicks, adding helmrelease manifest example for reference, https://github.com/fluxcd/flux-get-started/blob/master/releases/ghost.yaml#L18-L19

nicks:
I think the tricky part of this is handling it in the case where the "tag" field is omitted in the yaml, but tilt needs to inject it.

Currently thinking about an api like

```
k8s_kind(kind, image_object={'json_path': ..., 'repo_field': 'repo', 'tag_field': 'tag'}
```

the api is a bit awkward, but i think will unblock you without leading to too many weird api edge cases.

nicks:
@pandeybk Tilt v0.15.2 supports the above syntax! let us know if you have any issues with it

PR #3509, opened 2020-06-24
k8s: factor out an interface for extracting/injecting images

nicks:
Hello @jazzdan,

Please review the following commits I made in branch nicks/ch4395:

738d64af209d34a823fe1c6bd095ab0bbc8c565b (2020-06-24 19:47:38 -0400)
k8s: factor out an interface for extracting/injecting images
This is moving towards solutions for:
- https://github.com/tilt-dev/tilt/issues/3427
- https://github.com/tilt-dev/tilt/issues/3460

It ended up being kind of a yak shave, since we need to be able to attach
the object selectors to K8sTarget so that we know which entities to
inject into.

### DIFF
--- a/internal/k8s/entity.go
+++ b/internal/k8s/entity.go
@@ -334,8 +334,8 @@ func Filter(entities []K8sEntity, test func(e K8sEntity) (bool, error)) (passing
 	return passing, rest, nil
 }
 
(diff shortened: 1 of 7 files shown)


nicks:
oops, this is the wrong branch

PR #3526, opened 2020-06-29
tiltfile: add an API for adding image object locators. Fixes https://github.com/tilt-dev/tilt/issues/3427

nicks:
Hello @jazzdan, @landism,

Please review the following commits I made in branch nicks/ch8121:

fa7f254ee849816641a0a0e072822e94498e89da (2020-06-29 15:40:29 -0400)
tiltfile: add an API for adding image object locators. Fixes https://github.com/tilt-dev/tilt/issues/3427

### DIFF
--- a/internal/tiltfile/k8s/locators.go
+++ b/internal/tiltfile/k8s/locators.go
@@ -1,0 +1,129 @@
+package k8s
+
+import (
+	"fmt"
+
+	"go.starlark.net/starlark"
+
+	"github.com/pkg/errors"
+
+	"github.com/tilt-dev/tilt/internal/k8s"
+	"github.com/tilt-dev/tilt/internal/tiltfile/value"
+)
+
+// Deserializing locators from starlark values.
+type JSONPathImageLocatorListSpec struct {
+	Specs []JSONPathImageLocatorSpec
+}
+
+func (s JSONPathImageLocatorListSpec) IsEmpty() bool {
+	return len(s.Specs) == 0
+}
+
+func (s *JSONPathImageLocatorListSpec) Unpack(v starlark.Value) error {
+	list := value.ValueOrSequenceToSlice(v)
+	for _, item := range list {
+		spec := JSONPathImageLocatorSpec{}
+		err := spec.Unpack(item)
+		if err != nil {
+			return err
+		}
+		s.Specs = append(s.Specs, spec)
+	}
+	return nil
+}
+
+func (s JSONPathImageLocatorListSpec) ToImageLocators(selector k8s.ObjectSelector) ([]k8s.ImageLocator, error) {
+	result := []k8s.ImageLocator{}
+	for _, spec := range s.Specs {
+		locator, err := spec.ToImageLocator(selector)
+		if err != nil {
+			return nil, err
+		}
+		result = append(result, locator)
+	}
+	return result, nil
+}
+
+type JSONPathImageLocatorSpec struct {
+	jsonPath string
+}
+
+func (s *JSONPathImageLocatorSpec) badValueErr(v starlark.Value) error {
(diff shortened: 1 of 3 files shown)


nicks:
(also this was a fun opportunity to show off the new starlark Unpack() function for unpacking types)

internal/tiltfile/k8s/locators.go
```diff
@@ -0,0 +1,117 @@
...
+		return s.badValueErr(v)
+	}
+	tagField, found, err := m.Get(starlark.String("tag_field"))
+	if err != nil || !found {
+		return s.badValueErr(v)
+	}
+
+	s.jsonPath, ok = starlark.AsString(jsonPath)
+	if !ok {
+		return s.badValueErr(v)
```
landism, line 102: could you add a second arg to `badValueErr` with a more specific error? (e.g., "json_path must be a String. Found %T")

1. it'd annoying to play a guessing game as to which field is misspelled / which value is bad
2. the existing error message doesn't mention value types, so it wouldn't help with these errors anyway

internal/tiltfile/k8s/locators.go
```diff
@@ -0,0 +1,117 @@
...
+
+func (s *JSONPathImageLocatorSpec) badValueErr(v starlark.Value) error {
+	return fmt.Errorf("Expected map of the form {'json_path': ..., 'repo_field': ..., 'tag_field': ...}. Actual: %s", v)
+}
+
+func (s *JSONPathImageLocatorSpec) Unpack(v starlark.Value) error {
+	var ok bool
+	s.jsonPath, ok = starlark.AsString(v)
+	if !ok {
+		return fmt.Errorf("Expected string, got: %s", v)
```
landism, line 58: if we're expressing the expected value as a type, it seems like we should use %T for the observed value (plus it'd be annoying if `v`'s `.String()` didn't make it obviously not-a-string, resulting in a message like "Expected string, got: foobar")

internal/tiltfile/k8s/locators.go
```diff
@@ -0,0 +1,117 @@
...
+	}
+	repoField, found, err := m.Get(starlark.String("repo_field"))
+	if err != nil || !found {
+		return s.badValueErr(v)
+	}
+	tagField, found, err := m.Get(starlark.String("tag_field"))
+	if err != nil || !found {
+		return s.badValueErr(v)
+	}
+
```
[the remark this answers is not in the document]
nicks: done

internal/tiltfile/k8s/locators.go
nicks: done

internal/tiltfile/k8s/locators.go
```diff
@@ -0,0 +1,117 @@
...
+	repoField string
+	tagField  string
+}
+
+func (s JSONPathImageObjectLocatorSpec) IsEmpty() bool {
+	return s == JSONPathImageObjectLocatorSpec{}
+}
+
+func (s *JSONPathImageObjectLocatorSpec) badValueErr(v starlark.Value) error {
+	return fmt.Errorf("Expected map of the form {'json_path': ..., 'repo_field': ..., 'tag_field': ...}. Actual: %s", v)
```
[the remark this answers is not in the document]
nicks: ya, that's fair. TBH, I'm not particularly proud of this API, it's more of a hack to unblock people who want to do this.

long-term, i think we should probably push for an annotation on the Custom Resource Definition or Custom Resource itself to declare where the images are, tho this would require more collaboration with people in the kubernetes community to build consensus on it

internal/tiltfile/k8s/locators.go
nicks: done (though used the starlark Type() instead of the Go type

PR #3531, opened 2020-06-30
tiltfile: add config.tilt_subcommand

landism:
### Problem

Users often want to do different things depending on whether Tilt is upping or downing.

### Solution

Add a `config.tilt_subcommand` builtin that a Tiltfile author can use to tell what Tilt is currently running. e.g., if the user runs `tilt up` or `tilt down`, this variable will have the value `"up"` or `"down"`.


nicks:
some high-level thoughts (mostly name bikeshedding)
- The name "tilt" seems like a weird name for this module. Surprised you didn't put it under 'config', which seems closer to the class of problems users want to solve with this. Or maybe 'cli'
- In `tilt alpha tiltfile-result`, do you expect subcommand to be alpha or tiltfile-result?

landism:
> The name "tilt" seems like a weird name for this module. Surprised you didn't put it under 'config', which seems closer to the class of problems users want to solve with this. Or maybe 'cli'

Yeah, I don't love it as a module name, and mostly picked it because I felt like "tilt.sub_command" would read the best in the Tiltfile, without thinking about "tilt" as a module itself. I'll move it to 'config'

> * In `tilt alpha tiltfile-result`, do you expect subcommand to be alpha or tiltfile-result?

c) "alpha tiltfile-result"! I actually wrote a test for this and apparently failed to push after adding it. I figure it's an edge case and it's better to give the user complete information and let them hack it apart if they need to.

landism:
I also had to change up some error handling in the cli module so that I could run tests without getting an os.Exit

internal/tiltfile/config/config.go
```diff
@@ -25,12 +25,20 @@ type Settings struct {
...
 type Extension struct {
 	UserConfigState model.UserConfigState
+	TiltSubcommand  TiltSubcommand
 }
 
-func NewExtension(userConfigState model.UserConfigState) *Extension {
-	return &Extension{UserConfigState: userConfigState}
+type ExtensionProvider func(userConfigState model.UserConfigState) *Extension
+
+func NewExtensionProvider(tiltSubcommand TiltSubcommand) ExtensionProvider {
```
nicks, line 38: this is ok for now, but this pattern seems like a red flag that this particular extension isn't designed well.

I think the problem is that most extensions have a life-cycle that matches the lifetime of the whole tilt app, but this extensions lifetime is scoped to a single tiltfile execution. We probably need a better way to inject state into the `OnStart` hook

landism: yeah, it was already kind of weird that we're putting userConfigState in the extension and then propagating it to the extension state in NewState, and it'd be nice to avoid that as well

anyway, it turns out it was pretty easy to just initialize the field on the constructor right before execution, which still might be a bit ugly but I think is at least 1) more centralized ugliness and 2) closer to where we want to be in the future

PR #3540, opened 2020-07-02
web: get rid of blue box around tilt cloud status

### LINKED ISSUE #3420: Nit: Weird highlighting of `Your Tilt Cloud status` text in Tilt UI
Very minor issue, but there is some weird highlighting behavior around the text in the upper-right hand corner of the Tilt UI with the text `Your Tilt Cloud status`.

One example:
pic

I think this is a longstanding thing, not related to any recent Tilt release.

landism:
fixes #3420

(only happens after clicking on it)

old:
image

new:
image


landism:
It turns out this is a bad solution for accessibility reasons.

PR #3546, opened 2020-07-02
tilt: re-add config.tilt_subcommand

landism:
Three commits:
405df05 - straight-up revert of the previous revert, probably uninteresting to review given you've reviewed #3551
2944f1f - fix the bug that motivated the previous revert
4c83331 - found better ways to do existing things with the cobra API (using `rootCmd.PersistentPreRun` means we also capture subcommands that are added via `rootCmd.AddCommand(...)` instead of `addCommand(rootCmd, ...)`)


nicks:
Can you elaborate a bit more on why we need to change the command line parsing and error-handling to implement this feature? I think I'm missing something.

Mutating the Command config in RunE seems...like it's not how this library is intended to be used

nicks:
also: have you considered an approach where we pass the subcommand to wireCmdUp, like how we inject analytics? That might be a lot simpler, because you wouldn't have to play games with global scope and command-parsing order-of-operations

landism:
> Can you elaborate a bit more on why we need to change the command line parsing and error-handling to implement this feature? I think I'm missing something.
> 
> Mutating the Command config in RunE seems...like it's not how this library is intended to be used

It's not necessary. I added a test that runs `tilt alpha tiltfile-result`, and it turns out that [calls `os.Exit`](https://github.com/tilt-dev/tilt/blob/914a83d0a9cedc1a1bcf0b9f8bf99bfaecdf60c9/internal/cli/tiltfile_result.go#L89), which results in go test failing without telling you why or even what test was running.

Other options considered:
1. simply making the test create a valid Tiltfile, so that `tilt alpha tiltfile-result` passes and doesn't call `os.Exit`, bypassing the problem
2. keep the `os.Exit` in there and do something like [this](https://stackoverflow.com/a/33404435)
3. Turn SilenceUsage on all the time (https://github.com/spf13/cobra/issues/340#issuecomment-243790200) and write our own logic to figure out whether to print Usage.

I'd like to have more tests on the cobra code and figured getting rid of the `os.Exit`s was paying down some tech debt, but (1) is probably the right answer for this PR. Trying to make our cobra usage more testable should be separate.

landism:
> also: have you considered an approach where we pass the subcommand to wireCmdUp, like how we inject analytics? That might be a lot simpler, because you wouldn't have to play games with global scope and command-parsing order-of-operations

yeah, that's good

nicks:
ya, Cobra doesn't have great separation between its internals and its API, which can make it sometimes a bit befuddling to figure out what the API contract is.

PR #3557, opened 2020-07-07
tiltfile : added warning when there's duplicate yaml resources

abdullahzameek:
Hello @nicks , @maiamcc,

Please review the following commits I made in branch abdullahzameek/1852-yaml-duplicate:
8df8ea351cb443cd55bbc120acbc83c707a215d9  (17:49:26 2020 +0400)  - tiltfile: added warning when there's multiple yaml definitions (solved merge conflict) 
998c79ab6a6bc4b58edc9eab9bbe7b1abdb2861b (17:35:30 2020 +0400) -  tiltfile: fixed unit test for testing duplicate resources

### DIFF
--- a/internal/tiltfile/tiltfile_state.go
+++ b/internal/tiltfile/tiltfile_state.go
@@ -1129,9 +1129,11 @@ func (s *tiltfileState) translateK8s(resources []*k8sResource) ([]model.Manifest
 	if err != nil {
 		return nil, err
 	}
+
 	for _, manifest := range result {
 		s.warnDuplicateYamlEntities(manifest)
 	}
+
 	return result, nil
 }
 
@@ -1399,6 +1401,7 @@ func (s *tiltfileState) translateDC(dc dcResourceSet) ([]model.Manifest, error) 
 	for _, manifest := range result {
 		s.warnDuplicateYamlEntities(manifest)
 	}
+
 	return result, nil
 }
 
--- a/internal/tiltfile/tiltfile_state.go
+++ b/internal/tiltfile/tiltfile_state.go
@@ -254,7 +254,7 @@ to your Tiltfile. Otherwise, switch k8s contexts and restart Tilt.`, kubeContext
 		if err != nil {
 			return nil, starkit.Model{}, err
 		}
-		s.warnDuplicateYamlResources(yamlManifest)
+
 		manifests = append(manifests, yamlManifest)
 	}
 
@@ -263,6 +263,9 @@ to your Tiltfile. Otherwise, switch k8s contexts and restart Tilt.`, kubeContext
 		return nil, starkit.Model{}, err
 	}
 
+	for _, manifest := range manifests {
+		s.warnDuplicateYamlEntities(manifest)
+	}
 	return manifests, result, nil
 }
 
@@ -1126,7 +1129,9 @@ func (s *tiltfileState) translateK8s(resources []*k8sResource) ([]model.Manifest
 	if err != nil {
 		return nil, err
 	}
-
+	for _, manifest := range result {
+		s.warnDuplicateYamlEntities(manifest)
+	}
 	return result, nil
 }
 
@@ -1391,6 +1396,9 @@ func (s *tiltfileState) translateDC(dc dcResourceSet) ([]model.Manifest, error) 
 		s.postExecReadFiles = sliceutils.AppendWithoutDupes(s.postExecReadFiles, dc.configPaths...)
 	}
 
+	for _, manifest := range result {
+		s.warnDuplicateYamlEntities(manifest)
+	}
 	return result, nil
 }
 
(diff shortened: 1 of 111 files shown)


abdullahzameek:
@maiamcc made changes as requested!

pkg/model/manifest.go
```diff
@@ -45,6 +45,10 @@ type Manifest struct {
 	ResourceDependencies []ManifestName
 }
 
+func (m Manifest) GetDeployTarget() TargetSpec {
```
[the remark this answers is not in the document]
abdullahzameek: Removed this in 0aba336e88adca6c4e916b08a20654bb2cb96019

internal/tiltfile/tiltfile_state.go
```diff
@@ -1478,6 +1479,28 @@ func (s *tiltfileState) translateLocal() ([]model.Manifest, error) {
...
 	return result, nil
 }
+func (s *tiltfileState) warnDuplicateYamlResources(m model.Manifest) {
+	var tempName []string
+	deployTarget := m.K8sTarget()
+	displayNames := deployTarget.DisplayNames
+
+	duplicates := make(map[string]int)
+
+	// depends on the fact that the duplicate resource has formatting of doggos:service:default:core:n where n is nth
```
maiamcc, line 1489: good comment! I'd add a little more context: adding an index onto the end was a stopgap by us to keep Tilt from breaking when someone had the same entity twice in the same resource. We use it here to determine whether an entity appears more than once in this manifest. In future we may decide to make this an error instead (in which case we'd want to do this check closer to manifest assembly).

(The goal of a comment like this is to give the next dev to touch this code the tools they need to make decisions--so you're documenting not just WHAT you did, but WHY you did it, alternatives considered, etc.)

internal/tiltfile/tiltfile_state.go
```diff
@@ -1478,6 +1479,28 @@ func (s *tiltfileState) translateLocal() ([]model.Manifest, error) {
...
+	//occurrence of that resource.
+	for _, name := range displayNames {
+		tempName = strings.Split(name, ":")
+		if len(tempName) > 4 { //if it is > 4 that means there's an n
+			duplicates[tempName[0]]++ // we only care about the name of the resource
+		}
+	}
+
+	for resource, numOccurences := range duplicates {
```
maiamcc, line 1498: (unimportant comment and you can ignore it if you've got too much to do :) ) you're only adding something to the `duplicates` map if it has more than 4 components, i.e. if it has an `n` at the end--so when you add something to that map, you already know that it appears more than once in this manifest. I believe that makes the `numOccurences > 1` check redundant. 

given that, if you wanted, you could just make `duplicates` a `map[string]bool` (this is the idiomatic go way of doing a set` and just `for entity := range duplicates { s.logger.Warnf(...) }`

internal/tiltfile/tiltfile_state.go
```diff
@@ -1478,6 +1479,28 @@ func (s *tiltfileState) translateLocal() ([]model.Manifest, error) {
...
+		if numOccurences > 1 {
+			s.logger.Warnf("The following YAML Resource has been duplicated: " + resource)
```
maiamcc, line 1500: A. as said in a comment above, this should be "entity" and not "resource"
B. think about whether this gives the user enough information to debug. Does it say what resource the duplicated entities belong to? Why is this a problem, how should the user fix it?
C. (minor) I'm not sure you want one warning for each duplicated entity; consider instead [something like this](https://github.com/tilt-dev/tilt/blob/3f7711b77d4885b5e1ee2026b56f9edec9a0c8ca/internal/tiltfile/live_update.go#L25) where you list all the offenders in a single warning

internal/tiltfile/tiltfile_state.go
```diff
@@ -254,6 +254,7 @@ to your Tiltfile. Otherwise, switch k8s contexts and restart Tilt.`, kubeContext
 		if err != nil {
 			return nil, starkit.Model{}, err
 		}
+		s.warnDuplicateYamlResources(yamlManifest)
```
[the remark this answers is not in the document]
maiamcc: This is the one bit of this code that actually needs changing to work as expected imo -- my other comments are just comment/wording, don't be overwhelmed!

abdullahzameek:
Hello @maiamcc, 

Please find the requested changes in the following commits: 
c5490e943f528af23986cb53cabf5d33b1344904 (Thu Jul 9 20:45:57 2020 +0400) - tiltfile: added warning to all manifests
357224e5155ea2ae03b57f621b117c9247a2ca59 (Mon Jul 13 16:41:54 2020 +0400) -   fixed assertion in unit test

internal/tiltfile/tiltfile_state.go
```diff
@@ -1476,9 +1488,37 @@ func (s *tiltfileState) translateLocal() ([]model.Manifest, error) {
...
+	//error about the duplicated resource closer to the actual manifest assembly.
+	for _, name := range displayNames {
+		tempName := strings.Split(name, ":")
+		if len(tempName) > 4 { //this means there's an n which was appended at the manifest assembly because of multiple same entities present
+			duplicates[strings.Join(tempName[:4], ":")]++
+		}
+	}
+
+	for entity := range duplicates {
+		s.logger.Warnf("The following YAML Entity has been duplicated: " + entity)
```
maiamcc, line 1518: Not sure how helpful this error message will be to the user--what about something like:
```
"Resource %s defines k8s entity %s more than once. Only one definition will be applied to your cluster; to avoid unexpected behavior, please delete the redundant YAML."
```
or??? 🤷

abdullahzameek:
@maiamcc made all the changes, hopefully this is good to go now!

internal/tiltfile/tiltfile_state.go
```diff
@@ -1479,6 +1483,35 @@ func (s *tiltfileState) translateLocal() ([]model.Manifest, error) {
...
+		}
+	}
+
+	//get the names of all the duplicates
+	var duplicateNames []string
+	for key := range duplicates {
+		duplicateNames = append(duplicateNames, key)
+	}
+	if len(duplicateNames) != 0 {
+		s.logger.Warnf("Resource %s contains multiple specifications of k8s entity(s): %s. Only one can be applied to the cluster; to ensure expected behavior, remove the duplicate specifications",
```
maiamcc, line 1510: Whoops I realize the error message reads a little funny when there are multiple duplicates detected. wdyt of something like this?

```suggestion
		s.logger.Warnf("Resource %s contains multiple specifications of k8s entity(s): %s. Only one specification per entity can be applied to the cluster; to ensure expected behavior, remove the duplicate specifications",
```

internal/tiltfile/tiltfile_state.go
```diff
@@ -1479,6 +1483,35 @@ func (s *tiltfileState) translateLocal() ([]model.Manifest, error) {
...
+	for _, name := range displayNames {
+		tempName := strings.Split(name, ":")
+		if len(tempName) > 4 { //this means there's an n which was appended at the manifest assembly because of multiple same entities present
+			duplicates[strings.Join(tempName[:4], ":")]++
+		}
+	}
+
+	//get the names of all the duplicates
+	var duplicateNames []string
+	for key := range duplicates {
```
maiamcc, line 1506: heh there's no reason for `duplicates` to be a map any more, you could cut out this step by just storing duplicate names in a slice from the start :)

abdullahzameek: Ooooh I tried that, but then before I append a particular element to the slice, I'd need to check if its already there right? And since there isn't an easy way to check that like in Python (`if elem in list:`), I thought it'll probably be almost the same (in terms of computational time) to just keep the dictionary aha

internal/tiltfile/tiltfile_state.go
abdullahzameek: Oh yeah thats fair, so if there's doggos and cats, it'll seem like you can have either only doggos or only cats. I'll fix that!

internal/tiltfile/tiltfile_state.go
maiamcc: ......you're totally right, I forgot how computers worked 😅  ignore this comment!

abdullahzameek: No worries! I really wish Go had an `in` operator...would make slice stuff a looooot easier haha

abdullahzameek:
merged into master!

PR #3560, opened 2020-07-07
analytics: tag tilt.up.running w/ subcommand

landism:
### Problem

The `tilt.up.running` event also gets emitted by `tilt ci`, and includes no information to distinguish `tilt ci` from `tilt up`.

### Solution

Add a tag for the subcommand (e.g., "up", "ci") on `tilt.up.running` events.

### Notes

* We should rename `tilt.up.running` as well, but that makes this change riskier (since it gives us an inconsistency to account for in our analytics queries) and can be decoupled from this change.
* Ideally we have a separate mechanism to filter out runs in CI from human runs and this isn't the only thing we rely on, but:
  * that's always going to be a heuristic situation and more signal is better
  * this might be nice to have for other reasons
* Perhaps it'd be better add this as a global tag rather than just on this one count, but that's probably a couple hours of work instead of a few minutes.

### DIFF
--- a/internal/cli/down.go
+++ b/internal/cli/down.go
@@ -12,7 +12,6 @@ import (
 	"github.com/tilt-dev/tilt/internal/analytics"
 	"github.com/tilt-dev/tilt/internal/engine"
 	"github.com/tilt-dev/tilt/internal/k8s"
-	"github.com/tilt-dev/tilt/internal/tiltfile/config"
 	"github.com/tilt-dev/tilt/pkg/logger"
 	"github.com/tilt-dev/tilt/pkg/model"
 )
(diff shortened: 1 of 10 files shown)


maiamcc:
CC @victorwuky: when providing metrics to customers derived from `tilt.up.running`, can filter on `subcommand: up` to make sure we're just giving them metrics related to `tilt up` calls (and/or can use the `subcommand` field to see if anyone is running `tilt ci`, etc.)

maiamcc:
thanks for this @landism !!

victorwuky:
Thanks @maiamcc @landism . I'll be sure to try this out once we have a new release.

PR #3561, opened 2020-07-07
tiltfile: resources with extra pod selectors or images should be considered workloads

nicks:
Hello @jazzdan,

Please review the following commits I made in branch nicks/nonworkload:

993df2b62ed73023b453e2d30d3980b359a8a301 (2020-07-07 12:30:59 -0400)
tiltfile: resources with extra pod selectors or images should be considered workloads

### DIFF
--- a/internal/tiltfile/k8s.go
+++ b/internal/tiltfile/k8s.go
@@ -55,7 +55,7 @@ type k8sResource struct {
 
 	resourceDeps []string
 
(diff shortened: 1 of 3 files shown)


nicks:
I wonder if we should change the terminology here to "Podless" or something, since that what it means in practice (We don't track pods for nonworkloads)

PR #3563, opened 2020-07-07
engine: updateMode=synclet only if explicitly set with flag [8240]

maiamcc:
Hello @landism, @nicks,

Please review the following commits I made in branch maiamcc/rm-default-synclet:

dee0445636115c5ccea6139a21ed2afd3c5cd8b5 (2020-07-07 18:16:45 -0400)
fix other tests, rm redundant ones

da404d3ff71a9969b57d47c7da4faec34aed94b7 (2020-07-07 17:14:54 -0400)
fix liveupdate tests that expected synclet

12a71c2d55e9e4ec4a3646c4d158008890129d4f (2020-07-07 12:19:47 -0400)
only use synclet with explicit flag (needs tests)

### DIFF
--- a/internal/engine/build_and_deployer_test.go
+++ b/internal/engine/build_and_deployer_test.go
@@ -80,8 +80,66 @@ func TestGKEDeploy(t *testing.T) {
 		t.Errorf("Expected yaml to contain %q. Actual:\n%s", expectedYaml, f.k8s.Yaml)
 	}
 
+	if strings.Contains(f.k8s.Yaml, sidecar.DefaultSyncletImageName) {
+		t.Errorf("Should not deploy synclet without explicitly specifying updateMode=Synclet: %s", f.k8s.Yaml)
+	}
+}
+
+func TestGKEDeployWithSynclet(t *testing.T) {
+	f := newBDFixtureWithUpdateMode(t, k8s.EnvGKE, container.RuntimeDocker, buildcontrol.UpdateModeSynclet)
+	defer f.TearDown()
+
+	manifest := NewSanchoLiveUpdateManifest(f)
+	targets := buildTargets(manifest)
+	_, err := f.bd.BuildAndDeploy(f.ctx, f.st, targets, store.BuildStateSet{})
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if f.docker.BuildCount != 1 {
+		t.Errorf("Expected 1 docker build, actual: %d", f.docker.BuildCount)
+	}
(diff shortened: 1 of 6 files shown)


internal/engine/build_and_deployer_test.go
```diff
@@ -100,233 +100,121 @@ func TestDockerForMacDeploy(t *testing.T) {
...
-	assert.Equal(t, 0, f.sCli.UpdateContainerCount)
-	assert.Equal(t, 0, f.docker.BuildCount)
-	assert.Equal(t, 0, f.docker.PushCount)
-	f.assertK8sUpsertCalled(true)
+	if strings.Contains(f.k8s.Yaml, sidecar.DefaultSyncletImageName) {
+		t.Errorf("Should not deploy the synclet if manifest has no LiveUpdate instructions: %s", f.k8s.Yaml)
+	}
 }
 
-func TestContainerBuildLocal(t *testing.T) {
```
maiamcc, line 162: Deleted a bunch of tests that are covered more concisely in `build_and_deployer_live_update_test.go`

internal/engine/build_and_deployer_test.go
```diff
@@ -728,16 +579,14 @@ func TestDockerComposeLiveUpdate(t *testing.T) {
...
-	}
+	assert.Empty(t, f.k8s.Yaml, "expect no k8s YAML for DockerCompose resource")
+	assert.Empty(t, 0, f.k8s.ExecCalls,
+		"Expected no k8s Exec calls, actual: %d", f.k8s.ExecCalls)
 	f.assertContainerRestarts(1)
 }
 
 func TestReturnLastUnexpectedError(t *testing.T) {
-	f := newBDFixture(t, k8s.EnvGKE, container.RuntimeDocker)
+	f := newBDFixture(t, k8s.EnvDockerDesktop, container.RuntimeDocker)
```
maiamcc, line 589: We should probably fix the SanchoLiveUpdate manifest (i.e. the one we use by default in most of these tests) to NOT have a restart_container step in it, since that's been deprecated on the user side, and having that step in the manifest means we can't test it against UpdateModeKubectlExec (need to use a different update mode or a different test manifest). [Tracked that work here](https://app.clubhouse.io/windmill/story/8279/not-blocking-rm-restart-container-from-our-default-test-manifests) but for now it's easier to just change the k8s cluster so we get an update mode that allows `restart_container()` -- we'll deal with the rest if/when we actually deprecate that codepath.

Issue #3468, opened 2020-06-16
parallelize deployments with a common image


nicks:
Currently, if N Kubernetes deployments depend on the same image (or transitively on the same base image), the build controller won't try to parallelize them at all.

This avoids race conditions where the same image gets built multiple times.

But it should be able to
- Do the first deployment
- Detect that the image can be re-used in the other N-1 deployments
- Parallelize the other N-1 deployments with the re-used image

PR #3566, opened 2020-07-08
engine: parallize builds with common re-usable images. Fixes https://github.com/tilt-dev/tilt/issues/3468

nicks:
Hello @maiamcc, @landism,

Please review the following commits I made in branch nicks/ch7744:

75a55f0ae520130718dc631413ab292bd0025475 (2020-07-08 14:47:03 -0400)
engine: parallize builds with common re-usable images. Fixes https://github.com/tilt-dev/tilt/issues/3468

### DIFF
--- a/internal/engine/buildcontrol/build_control.go
+++ b/internal/engine/buildcontrol/build_control.go
@@ -86,6 +86,43 @@ func isWaitingOnDependencies(state store.EngineState, mt *store.ManifestTarget) 
 	return false
 }
 
+// Check to see if this is an ImageTarget where the built image
+// can be potentially reused.
+//
+// Note that this is a quick heuristic check for making parallelization decisions.
+//
+// The "correct" decision about whether an image can be re-used is more complex
+// and expensive, and includes:
+//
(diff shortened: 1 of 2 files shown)


internal/engine/buildcontrol/build_control.go
```diff
@@ -86,27 +86,71 @@ func isWaitingOnDependencies(state store.EngineState, mt *store.ManifestTarget)
...
+					continue
+				}
+
 				building[spec.ID()] = true
 			}
 		}
 	}
 
-	isBuilding := func(m model.Manifest) bool {
+	isBuilding := func(mt *store.ManifestTarget) bool {
```
landism, line 143: would a better name for this be "hasBuildingComponents"? I was confused that we have this isBuilding that's a func of ManifestTarget -> bool and also have mt.State.IsBuilding() -> bool, but they have different meanings

if yes and you wanna change it in this PR, that's great, or I can do it in a follow-up PR.

nicks: i like that! done

PR #3572, opened 2020-07-09
tiltfile: StringOrStringList: require Union[string,Union[List,Tuple]] instead of Union[string,Iterable]

### LINKED ISSUE #3570: helm.set accepts a map but then discards the values
### Repro

`helm('./mychart', set={'a': 'b'})`

### Observed

Tilt runs `["helm" "template" "chart" "/private/tmp/helmset/foo" "--set" "a"]`

### Expected

The docs [specify](https://docs.tilt.dev/api.html#api.helm) `set`'s type as `Union [ str , List [ str ]]`.
Either of these behaviors would be preferable:
* generate an error because I passed the wrong type to `set`
* accept the dict and format it to what helm expects for me

### Notes

This is due to a bug in [`value.AsStringOrStringList`](https://github.com/tilt-dev/tilt/blob/d8cca417e025feed882c0a5ea18a901dc501525f/internal/tiltfile/value/string.go#L20): it only checks that it's an Iterable of strings, and a starlark Map implements Iterable by returning its keys.

These args also use `value.AsStringOrStringList` and potentially suffer from this:
* `docker_build`'s `ssh`, `secret`, and `tags`
* `helm`'s `valueFiles`

Hence, it probably makes sense to fix the bug and then consider the arg type change separately.
(nb: while we're here: that functions' interface just returns true/false and leaves error handling up to the caller, which can result in unfortunate error messages like `Error: Argument 'set' must be string or list of strings. Actual: *starlark.List` when there's a non-string value inside the list)

landism:
fixes #3570

also improves the error message when one of the list elements is the wrong type

My main outstanding concern is if there are non-List Iterables other than map that we still want to accept. I did test that generators still get represented as lists.

### DIFF
--- a/internal/tiltfile/value/string.go
+++ b/internal/tiltfile/value/string.go
@@ -1,6 +1,10 @@
 package value
 
-import "go.starlark.net/starlark"
+import (
+	"fmt"
+
+	"go.starlark.net/starlark"
+)
 
 type ImplicitStringer interface {
 	ImplicitString() string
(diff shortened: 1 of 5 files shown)


internal/tiltfile/value/string.go
```diff
@@ -17,31 +21,31 @@ func AsString(x starlark.Value) (string, bool) {
 
 // Unpack an argument that can either be expressed as
 // a string or as a list of strings.
-func AsStringOrStringList(x starlark.Value) ([]string, bool) {
+func AsStringOrStringList(x starlark.Value) ([]string, error) {
```
nicks, line 24: bonus points if you convert this to use the Unpack() interface, so that more of the error messaging will be handled by starlark internally

internal/tiltfile/value/string.go
```diff
@@ -17,31 +21,31 @@ func AsString(x starlark.Value) (string, bool) {
...
-		var item starlark.Value
-		for iter.Next(&item) {
-			s, ok := AsString(item)
-			if !ok {
-				return nil, false
-			}
-			result = append(result, s)
-		}
-		return result, true
+	list, ok := x.(*starlark.List)
```
nicks, line 34: is it also important to handle tuples or nah?

internal/tiltfile/value/string.go
landism: oy, of course!

internal/tiltfile/value/string.go
landism: I could go either way. I'm not sure whether allowing them or not is more pythonic (or starlarkic?). Python allows `for` over tuples, so I've added them (though it also allows `for` over dicts, so that's not necessarily much of an argument - at least they won't fail weirdly in this case like dicts do)

Issue #3562, opened 2020-07-07
tilt should surface image pulling events


nicks:
If a pod is taking a long time to start due to a long image pull, the Tilt UX will just hang for a long time.

It would be nice if Tilt would expose the Pulling/Pulled event for images in the log stream

PR #3577, opened 2020-07-10
print image pull/pulling events

nicks:
Hello @maiamcc, @hyu,

Please review the following commits I made in branch nicks/ch8352:

91c7e1bfe2b6e594334b97c5212a0198aade9980 (2020-07-10 16:13:28 -0400)
print image pull/pulling events
here's an example of how this looks in practice:
https://cloud.tilt.dev/snapshot/AYakxvAL8dRwcxxqjC4=

Fixes https://github.com/tilt-dev/tilt/issues/3562

### DIFF
--- a/internal/engine/upper_test.go
+++ b/internal/engine/upper_test.go
@@ -2449,34 +2449,6 @@ func TestK8sEventNotLoggedIfNoManifestForUID(t *testing.T) {
 		"should not log event message b/c it doesn't have a UID -> Manifest mapping")
 }
 
-func TestK8sEventDoNotLogNormalEvents(t *testing.T) {
-	f := newTestFixture(t)
-	defer f.TearDown()
-
-	name := model.ManifestName("fe")
-	manifest := f.newManifest(string(name))
-
-	f.Start([]model.Manifest{manifest})
-	f.waitForCompletedBuildCount(1)
-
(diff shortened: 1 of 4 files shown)


nicks:
The attached snapshot displays something i'm finding very frustrating when using Tilt - sometimes it takes a while to pull the image to the node, and we don't do a good job displaying that this is what Kubernetes is waiting on. This display is still pretty awful, but at least gives info on what image is being pulled.

internal/engine/upper_test.go
```diff
@@ -2449,34 +2449,6 @@ func TestK8sEventNotLoggedIfNoManifestForUID(t *testing.T) {
 		"should not log event message b/c it doesn't have a UID -> Manifest mapping")
 }
 
-func TestK8sEventDoNotLogNormalEvents(t *testing.T) {
```
maiamcc, line 2452: wait, doesn't this test still apply? do you have test coverage for:
1. don't log "normal" events
2. log if !normal and if `pull` or `pulling` (but presumably not otherwise)?

nicks: ya, mostly this was testing code that didn't matter anymore, and really needed to be testing the subscriber code. added a new test!

PR #3589, opened 2020-07-15
hud/server: tilt logs poc [ch8431]

maiamcc:
Hello @nicks, @jazzdan,

MANY todo's (mostly tracked in the code, and I'm trying to track in clubhouse as well)
but here's a POC of `tilt logs`. currently it streams ALL the logs from the Tilt instance
running on port 10350 via websocket.

does this seem like the right track? am i making silly mistakes with code org / websocket stuff?

eventual plan is to re-org the code such that the Go websocket reader lives where it currently does,
and View handlers can live in their own packages. (As stated in a comment, I figured it'd be
worth it to make a generic Go object that can consume `View` info off the websocket, b/c
presumably we'll want other cmds that read state from a running Tilt, e.g. `tilt status`.)

this command isn't visible to users so it can be landed and tinkered with freely.


internal/hud/server/websocket_reader.go
```diff
@@ -0,0 +1,174 @@
...
+					log.Println("🚨 sending response:", err)
+				}
+			}
+
+		}
+	}()
+
+	for {
+		select {
+		case <-done:
```
jazzdan, line 155: Is there a situation in which the above goroutine returns and it isn't an error? If so, should we return a non-zero exit code here?

internal/hud/server/websocket_reader.go
```diff
@@ -0,0 +1,174 @@
+package server
+
+import "net/url"
+
+// A thinger to connect to the HUD server websocket, for CLI commands that
+// want to read state from a running Tilt (e.g. `tilt logs`).
+// TODO(maia): figure out if this should live here, or elsewhere (I assume
```
jazzdan, line 7: My rule of thumb when it comes to such things is to wait for the second time I need it. We're more likely to have a better idea of what it should look like then, and I wouldn't want to hold things up in the meantime.

internal/cli/logs.go
```diff
@@ -0,0 +1,52 @@
...
+	return cmd
+}
+
+func (c *logsCmd) run(ctx context.Context, args []string) error {
+	a := analytics.Get(ctx)
+
+	a.Incr("cmd.logs", nil)
+	defer a.Flush(time.Second)
+
+	span, ctx := opentracing.StartSpanFromContext(ctx, "Up")
```
jazzdan, line 41: Should we be creating spans here?

Should it be called up?

internal/hud/server/websocket_reader.go
```diff
@@ -0,0 +1,174 @@
...
+	if fromCheckpoint < ls.checkpoint {
+		// The server is re-sending some logs we already have, so slice them off.
+		deleteCount := ls.checkpoint - fromCheckpoint
+		segments = segments[deleteCount:]
+	}
+
+	// TODO(maia): filter for the resources that we care about (`tilt logs resourceA resourceC`)
+	//   --> and if there's only one resource, don't prefix logs with resource name?
+	for _, seg := range segments {
+		// TODO(maia): secrets???
```
nicks, line 79: you don't need to worry about secrets here. the central logstore has secrets so that it can retroactively scrub logs, but you've already printed them out so the proverbial horse has left the barn

internal/hud/server/websocket_reader.go
```diff
@@ -0,0 +1,174 @@
...
+			if toCheckpoint > 0 {
+				// If server is using the incremental logs protocol, ack the
+				// message so the next time the websocket sends data, it only
+				// sends logs from here on forward
+				resp := proto_webview.AckWebsocketRequest{
+					ToCheckpoint:  toCheckpoint,
+					TiltStartTime: v.TiltStartTime,
+				}
+				marshaller := jsonpb.Marshaler{OrigName: false, EmitDefaults: true}
+				respJson, err := marshaller.MarshalToString(&resp)
```
nicks, line 139: idiomatically you'd just plug the io streams together like
writer := c.NextWriter(TextMessage)
defer writer.Close()
marshaller.Marshal(writer, msg)

to avoid converting back and forth from bytes to strings (which puts a lot of stress on the garbage collector)

internal/hud/server/websocket_reader.go
```diff
@@ -0,0 +1,174 @@
...
+	ls.printer.Print(ls.logstore.ContinuingLines(ls.checkpoint))
+
+	if toCheckpoint > ls.checkpoint {
+		ls.checkpoint = toCheckpoint
+	}
+
+	return nil
+}
+
+func (wsr *WebsocketReader) Listen() {
```
nicks, line 92: would probably pass in a context here and handle the interrupt that way

internal/hud/server/websocket_reader.go
```diff
@@ -0,0 +1,174 @@
...
+	}
+	defer c.Close()
+
+	done := make(chan struct{})
+
+	go func() {
+		// TODO(maia): make sure this closes okay 😅
+		defer close(done)
+		for {
+			_, data, err := c.ReadMessage()
```
nicks, line 112: idiomatically you'd usually do c.NextReader(), and unmarshal directly off the reader (for some reasons as using NextWriter below)

internal/hud/server/websocket_reader.go
```diff
@@ -0,0 +1,174 @@
...
+			if err != nil {
+				log.Println("🚨 error reading:", err)
```
nicks, line 114: another advantage of NextReader() - an error here would just mean the connection closed, and you don't need to worry about that

internal/hud/server/websocket_reader.go
```diff
@@ -0,0 +1,174 @@
...
+				return
+			}
+
+			v := proto_webview.View{}
```
nicks, line 118: might break out this into a method to more cleanly stop handle the error cases, stop, and message them

internal/cli/logs.go
maiamcc: 😬 maybe and probably not

internal/hud/server/websocket_reader.go
maiamcc: fair nough -- it would be MORE work now to re-merge the types lol, but I changed the note up top to reflect this!

internal/hud/server/websocket_reader.go
```diff
@@ -0,0 +1,176 @@
...
+			return nil
+		case <-ctx.Done():
+			err := ctx.Err()
+			if err != context.Canceled {
+				return err
+			}
+
+			// Cleanly close the connection by sending a close message and then
+			// waiting (with timeout) for the server to close the connection.
+			err = c.WriteMessage(websocket.CloseMessage, websocket.FormatCloseMessage(websocket.CloseNormalClosure, ""))
```
maiamcc, line 165: @nicks can you check me here?
1. in general, is this procedure for closing the connection right?
2. should i be writing a `CloseMessage` even if it's not a `context.Canceled` error?

nicks: i assume you just copied from this?
https://github.com/gorilla/websocket/blob/5ed622c449da6d44c3c8329331ff47a9e5844f71/examples/echo/client.go#L70

but ya, this seems fine to me, but you probably know more about this than i do at this point!

internal/hud/server/websocket_reader.go
```diff
@@ -0,0 +1,194 @@
...
+	}()
+
+	for {
+		select {
+		case <-done:
+			return nil
+		case <-ctx.Done():
+			err := ctx.Err()
+			if err != context.Canceled {
+				return err
```
maiamcc, line 140: @nicks i still don't know whether I should try to call wsr.conn.Close() here?

nicks: honestly, your guess is as good as mine

maiamcc: 😬 👍

Issue #3598, opened 2020-07-20
Use keg_only build if user has tilt gem installed


StefanoDeVuono:
A [known issue](https://docs.tilt.dev/faq.html#q-when-i-run-tilt-version-i-see-template-engine-not-found-for-version-what-do-i-do) with home-brew installations of tilt is that they can conflict with the `tilt` gem. We can fix this by checking if this gem is installed and doing a keg_only brew install

PR #3599, opened 2020-07-20
use keg_only install for homebrew if tilt gem is installed

StefanoDeVuono:
#3598 

A [known issue](https://docs.tilt.dev/faq.html#q-when-i-run-tilt-version-i-see-template-engine-not-found-for-version-what-do-i-do) with home-brew installations of tilt is that they can conflict with the `tilt` gem. We can fix this by checking if this gem is installed and doing a keg_only brew install

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -47,6 +47,10 @@ brews:
   description: "Tilt powers multi-service developments for teams that deploy to Kubernetes."
   test: |
     system "#{bin}/tilt version"
+  custom_block: |
+    if  Gem::Specification::find_all_by_name('tilt').any?
+      keg_only :param, 'because you you requested it'
+    end 
 dockers:
 - image_templates:
     - "tiltdev/tilt"


.goreleaser.yml
```diff
@@ -47,6 +47,10 @@ brews:
   description: "Tilt powers multi-service developments for teams that deploy to Kubernetes."
   test: |
     system "#{bin}/tilt version"
+  custom_block: |
+    if  Gem::Specification::find_all_by_name('tilt').any?
```
nicks, line 51: I'm not sure I understand this stanza. can you point us to documentation to better understand how it works?

nicks: poking around a bit, it seems like keg_only is really for cases like openssl, where end-users know they have to use the absolute path to the openssl installation:
https://docs.brew.sh/Formula-Cookbook#check-for-dependencies

This use-case seems closer to `conflicts_with`: https://docs.brew.sh/Formula-Cookbook#specifying-conflicts-with-other-formulae

StefanoDeVuono: I'm open to both `keg_only` and `conflicts_with`.

The cookbook mentions that "sometimes there’s hard conflict between formulae, and it can’t be avoided or circumvented with keg_only" and "as a general rule, conflicts_with should be a last-resort option. It’s a fairly blunt instrument." It looks like `conflicts_with` is more escalatory than `keg_only` which is why I chose it.

What would be ideal is having a flag we could pass to `brew install`, like `brew install tilt --keg-only` or `brew install tilt --conflicts-with`. What are your thoughts?

nicks: haha, ya

The problem is that, AFAICT, keg_only doesn't fix the conflict between the two tilt packages. 

I agree that the language in the documentation just warns you away from conficts_with, without making it super-clear why you'd choose one vs the other

Question: how do _you_ invoke tilt when you have the ruby tilt on your path?

One possibility is to have a second homebrew package that installs it under a different name (tlt? tilt-dev?). goreleaser makes it pretty easy to have multiple homebrew packages

StefanoDeVuono: Oh `keg_only` _definitely_ fixes the problem. It prevents a link to `/usr/local/bin/tilt` which is where some ruby environments have installed the tilt gem. This then allows users to create a link from opt or the Cellar if they wish.

The project I'm working on (if it were just for my environment, I'd figure out another way of installing tilt--the dev package, not the gem) uses `tlt`. I edited the tilt formula to make it keg_only and then created a symlink from `/usr/local/opt/tilt/bin/tilt` to `/usr/local/bin/tlt`.

Using `tlt` to `tilt-dev` sounds great too! (We're already effectively using `tlt` for all our commands.)

maiamcc:
@StefanoDeVuono I'm closing this PR as it's been inactive for a while, but if you want to take another swing at this given @nicks's comments, feel free to re-open or to submit a new PR!

PR #3600, opened 2020-07-21
tiltfile: added error when duplicate yaml is detected

abdullahzameek:
Hi @maiamcc @jazzdan, 

Please review the following commits I made in branch abdullahzameek/3588-duplicate-yaml:

0b8e13f234d1c34c1eeeb3129f2512978c523b77 (Tue Jul 21 17:27:42 2020 +0400)
tiltfile: added error when duplicate yaml is detected

### DIFF
--- a/internal/k8s/names.go
+++ b/internal/k8s/names.go
@@ -5,8 +5,14 @@ import (
 	"strings"
 )
 
+const FmtduplicateYamlDetectedError = "Duplicate YAML Entity(s): [%s] has been detected across one or more resources.  Only one specification per entity can be applied to the cluster; to ensure expected behavior, remove the duplicate specifications."
+
+func DuplicateYamlDetectedError(duplicatedYaml string) string {
+	return fmt.Sprintf(FmtduplicateYamlDetectedError, duplicatedYaml)
+}
+
 // Calculates names for workloads by using the shortest uniquely matching identifiers
-func UniqueNames(es []K8sEntity, minComponents int) []string {
+func UniqueNames(es []K8sEntity, minComponents int) ([]string, error) {
 	ret := make([]string, len(es))
 	// how many resources potentially map to a given name
 	counts := make(map[string]int)
@@ -28,21 +34,13 @@ func UniqueNames(es []K8sEntity, minComponents int) []string {
 			}
 		}
 		if ret[i] == "" {
-			// If we hit this case, this means we have two resources with the same
-			// name/kind/namespace/group This usually means the user is trying to
-			// deploy the same resource twice. Kubernetes will not treat these as
-			// unique.
-			//
-			// We should surface a warning or error about this somewhere else that has
-			// more context on how to fix it.
-			// https://github.com/tilt-dev/tilt/issues/1852
-			//
-			// But for now, append the index to the name to make it unique
-			ret[i] = fmt.Sprintf("%s:%d", names[len(names)-1], i)
+			// Previously, we appended an index to the duplicate yaml entity, but as of July 21st 2020, we return an error
+			//if we've detected a duplicate yaml entity.
+			return nil, fmt.Errorf(DuplicateYamlDetectedError(names[len(names)-1]))
 		}
 	}
 
-	return ret
+	return ret, nil
 }
 
 // FragmentsToEntities maps all possible fragments (e.g. foo, foo:secret, foo:secret:default) to the k8s entity or entities that they correspond to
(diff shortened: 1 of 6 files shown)


internal/k8s/names.go
```diff
@@ -5,8 +5,14 @@ import (
 	"strings"
 )
 
+const FmtduplicateYamlDetectedError = "Duplicate YAML Entity(s): [%s] has been detected across one or more resources.  Only one specification per entity can be applied to the cluster; to ensure expected behavior, remove the duplicate specifications."
```
jazzdan, line 8: Does this need to be exported? (It's exported because it begins with an uppercase letter)

abdullahzameek: Yep! I'm using that `const` and function in `tiltfile_test.go` in `loadErrString()`

internal/tiltfile/tiltfile_test.go
```diff
@@ -3836,57 +3837,61 @@ k8s_yaml('foo.yaml')
...
-	f.assertWarnings(duplicateWarningStr)
+	labelsFoo := map[string]string{"foo": "bar"}
+	f.yaml("all.yaml",
+		deployment("foo-deployment", image("gcr.io/foo1"), namespace("ns1"), withLabels(labelsFoo)),
+		service("foo-service", withLabels(labelsFoo)),
+		service("foo-service", withLabels(labelsFoo)))
+	f.file("Tiltfile", `
+k8s_yaml('all.yaml')
+`)
+	f.loadErrString(k8s.DuplicateYamlDetectedError("foo-service:service:default:core"))
```
abdullahzameek, line 3880: @jazzdan This seemed to be a good way to get access to that const....should I re-declare it in this file, or is the current method okay?

internal/k8s/names.go
maiamcc, line 8: This is a very silly wording correction but the current implementation will only ever error about one duplicate at a time, so you don't need `entity(s)`, just `entity` is fine
```suggestion
const FmtduplicateYamlDetectedError = "Duplicate YAML Entity: "%s" has been detected across one or more resources.  Only one specification per entity can be applied to the cluster; to ensure expected behavior, remove the duplicate specifications."
```

internal/tiltfile/tiltfile_state.go
```diff
@@ -654,7 +651,14 @@ func (s *tiltfileState) assembleK8sV1() error {
 }
 
 func (s *tiltfileState) assembleK8sV2() error {
-	err := s.assembleK8sByWorkload()
+	//Note: We're using UniqueNames() over here just to check for duplicate resources, hence
```
maiamcc, line 654:
```suggestion
	//Note: We're using UniqueNames() over here just to check for duplicate YAML entities, hence
```

internal/tiltfile/tiltfile_test.go
jazzdan: Ah! I missed this. Yeah, makes sense! My question was a genuine one, not a gotcha :sweat_smile:

internal/k8s/names.go
maiamcc: hmm, I don't see `FmtduplicateYamlDetectedError` used anywhere but this file; this const can be private but accessed by a public function `DuplicateYamlDetectedError` elsewhere!

Small naming sidenote that doesn't super matter: I think you want `fmtDuplicate...` (capital D), and in general acronyms/initialisms like `YAML` should be [consistently cased](https://github.com/golang/go/wiki/CodeReviewComments#initialisms) -- so this func should be `fmtDuplicateYAMLDetectedError`

internal/tiltfile/tiltfile_state.go
abdullahzameek: Aaaaa I keep making this mistake of referring to YAML entities as resources haha

internal/k8s/names.go
abdullahzameek: Ohhh right, I thought I wouldn't be able to access a private const, even through a public function. Will correct that and the casing!

abdullahzameek:
@landism, 

@maiamcc and I discussed that since this PR has code that changes the way duplicate YAML entities are handled entirely from the previous approach (appending a numerical index to the end of the entity name), this might be a breaking change in this week's release?

internal/k8s/names.go
abdullahzameek: Corrected this, will merge into master now!

abdullahzameek:
Merged into master!

jazzdan:
Wohoo! Thanks @abdullahzameek

PR #3604, opened 2020-07-22
build: store Dockerfile in random path, dockerignore it

jazzdan:
To match what Docker does here https://github.com/docker/cli/blob/bece8cc41c248050e1c70ea73ce0097bc5a8e11e/cli/command/image/build/context.go#L397

### DIFF
--- a/internal/build/docker_builder.go
+++ b/internal/build/docker_builder.go
@@ -12,6 +12,7 @@ import (
 	"github.com/docker/distribution/reference"
 	"github.com/docker/docker/client"
 	"github.com/docker/docker/pkg/jsonmessage"
+	"github.com/docker/docker/pkg/stringid"
 	controlapi "github.com/moby/buildkit/api/services/control"
 	"github.com/opencontainers/go-digest"
 	opentracing "github.com/opentracing/opentracing-go"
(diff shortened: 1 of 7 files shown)


jazzdan:
Hmm this integration test failure might be legit, I'll look in to it tomorrow.

jazzdan:
Ah I see what's happening.

Every time we build an image now the Dockerfile is named something else. This results in a different image hash, which causes the `TestIdempotent` line to fail because the pod spec changes causing a new pod to be deployed. Two deploys with no changes to the source file system are now no longer idempotent.

Is there a way around this? Tagging in @nicks since I think you know the most about this.

landism:
> Is there a way around this?

IIUC the only reason to make it random is to reduce the chances it conflicts with a file that actually exists in the context. How about using something non-random that'd suffice for same, e.g. `Dockerfile.tilt-$manifestname`? If you wanna get super careful you could check if the chosen filename exists and increment a suffix until you find one that doesn't.

landism:
Another line of questioning might be - if the docker cli is doing this, how is it avoiding changing image digests on each build? (at least, I'm pretty sure it avoids that)

jazzdan:
@landism I asked in Docker slack, got this answer:

>It's all handled on the client.
>As I recall this is a stanky hack for content trust.

Not seeing any other references to that file pattern in the Docker code though https://sourcegraph.com/search?q=repo:%22%5Egithub%5C%5C.com/docker/*%22+%22.dockerfile.%22&patternType=literal

jazzdan:
I don't think this is worth digging in to any further. It doesn't seem to provide that much value and it comes at the cost of significant overhead. I'm going to close this for now until we decide to investigate further.

maiamcc:
womp womp thanks for trying :/

Issue #3584, opened 2020-07-14
determining the tag tilt deploys with


nicks:
Suppose you have a custom_build.

Tilt will set the environment variable EXPECTED_REF. When the custom_build script finishes, Tilt looks in the docker image store to find the image. Then it retags the image with a content-based tag and deploys it.

For Tilt, it's important to deploy a content-based tag. The architecture assumes that the tag changes when the content of the image changes, and stays the same when the content doesn't change. KNative uses a similar strategy to make similar guarantees.

But there are cases where the custom_build script needs access to the content-based tag:
1) When another script in the build pipeline needs to use the same tag
2) When the container you're running needs to use the tag to deploy workers

We currently have a solution to (2) but it's not well-documented. You need to add an env var to your container:

```
env:
  - name: IMAGE_BUSYBOX
    value: busybox
```

and then in your Tiltfile

```
custom_build('busybox', ..., match_in_env_vars=True)
```

to tell Tilt to inject the image reference (with the right tag) into the env variable

We don't have a good solution to (1) right now, though you could imagine a simple command like

```
tilt dump image-deploy-ref MY_IMAGE
```

that the custom_build script can run after it finishes building the image, so that tilt can communicate back what it will tag the image with (you need to build the image first to determine what the tag will be)

nicks:
(part of this might be resolved with a better airflow example -- the env variable is similar to how i'm running airflow pipelines with the KubernetesPodOperator)

nicks:
this is released in tilt v0.17.0

PR #3607, opened 2020-07-22
cli: add 'tilt dump image-deploy-ref', for determining the deploy tag of an image

nicks:
Hello @maiamcc, @landism,

Please review the following commits I made in branch nicks/ch8524:

d77706bcae5c133327ff1e21e067f9b59b9105b3 (2020-07-22 16:58:12 -0400)
cli: add 'tilt dump image-deploy-ref', for determining the deploy tag of an image
helps with more complex custom_build commands as described in https://github.com/tilt-dev/tilt/issues/3584

### DIFF
--- a/internal/build/docker_builder.go
+++ b/internal/build/docker_builder.go
@@ -37,6 +37,7 @@ type dockerImageBuilder struct {
 
 type DockerBuilder interface {
 	BuildImage(ctx context.Context, ps *PipelineState, refs container.RefSet, db model.DockerBuild, filter model.PathMatcher) (container.TaggedRefs, error)
+	DumpImageDeployRef(ctx context.Context, ref string) (reference.NamedTagged, error)
 	PushImage(ctx context.Context, name reference.NamedTagged) error
 	TagRefs(ctx context.Context, refs container.RefSet, dig digest.Digest) (container.TaggedRefs, error)
 	ImageExists(ctx context.Context, ref reference.NamedTagged) (bool, error)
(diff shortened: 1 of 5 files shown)


internal/cli/dump.go
```diff
@@ -134,6 +139,54 @@ func (c *dumpCliDocsCmd) run(cmd *cobra.Command, args []string) {
 	}
 }
 
+func newDumpImageDeployRefCmd() *cobra.Command {
+	return &cobra.Command{
+		Use:   "image-deploy-ref REF",
+		Short: "Determine the deployed name and tag of a custom-built image",
```
maiamcc, line 145: idk if this is actually clearer, take it or leave it 🤷 
```suggestion
		Short: "Determine the name and tag with which Tilt will deploy the given custom-built image",
```
(and change in `Long` description as well)

internal/build/docker_builder.go
```diff
@@ -68,6 +69,31 @@ func (d *dockerImageBuilder) BuildImage(ctx context.Context, ps *PipelineState,
 	return d.buildFromDf(ctx, ps, db, paths, filter, refs)
 }
 
+func (d *dockerImageBuilder) DumpImageDeployRef(ctx context.Context, ref string) (reference.NamedTagged, error) {
+	refParsed, err := container.ParseNamed(ref)
+	if err != nil {
+		return nil, errors.Wrap(err, "DumpImageDeployRef")
+	}
+
+	data, _, err := d.dCli.ImageInspectWithRaw(ctx, ref)
```
maiamcc, line 78: Have a vague instinct that as much of this logic as possible should live in a helper function that's called by both `dump-image...` and the `CustomBuilder`, to guard against the logic going out of sync. How hard would that be?

maiamcc:
Also are you going to update the Custom Build docs to reflect this strategy?

internal/cli/dump.go
nicks: done!

nicks:
yep, the attached issue outlines what the new docs will look like. it's a bit of an advanced feature (and i'm concerned that more people will use it incorrectly than use it correctly). But we're pairing with a partner team and watching closely how they use it

internal/build/docker_builder.go
maiamcc: whoops yes. cool!

PR #3611, opened 2020-07-23
Track installs via install.sh

abdullahzameek:
Hi @nicks, 

Please review the following commits I made in branch abdullahzameek/install-script-analytics:

3de56189c61a68762ecb0c79dbb8e4e1cf2d31b5 (Thu Jul 23 20:13:37 2020 +0400) - cli: added tilt verify-install to install script
d583d03f639e63dabf68b2dc2e25919921cf17af  (Thu Jul 23 18:14:25 2020 +0400) -   cli: created verify install option for tilt

### DIFF
--- a/internal/cli/verify_install.go
+++ b/internal/cli/verify_install.go
@@ -1,0 +1,24 @@
+package cli
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/spf13/cobra"
+)
+
+type verifyInstallCmd struct {
+}
+
+func (c *verifyInstallCmd) register() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "verify-install",
+		Short: "Verifies Tilt Installation",
+	}
+	return cmd
+}
+
+func (c *verifyInstallCmd) run(ctx context.Context, args []string) error {
+	fmt.Println("wohooo this cmd works>")
+	return nil
+}
--- a/internal/cli/verify_install.go
+++ b/internal/cli/verify_install.go
@@ -6,7 +6,6 @@ import (
 
 	"github.com/spf13/cobra"
 	"github.com/tilt-dev/tilt/internal/analytics"
-	engineanalytics "github.com/tilt-dev/tilt/internal/engine/analytics"
 )
 
 type verifyInstallCmd struct {
@@ -22,13 +21,7 @@ func (c *verifyInstallCmd) register() *cobra.Command {
 
 func (c *verifyInstallCmd) run(ctx context.Context, args []string) error {
 	a := analytics.Get(ctx)
-	machineID, _ := a.GlobalTag("machine")
-
-	cmdVerifyInstallTags := engineanalytics.CmdTags(map[string]string{
-		"machine": machineID,
-	})
-	a.Incr("cmd.verifyInstall", cmdVerifyInstallTags.AsMap())
-
+	a.Incr("cmd.verifyInstall", nil)
 	defer a.Flush(time.Second)
 
 	return nil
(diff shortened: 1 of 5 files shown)


abdullahzameek:
Screenshot from 2020-07-23 20-18-21

Can confirm that the data is being sent to wmstats (checked through jupyter)

internal/cli/verify_install.go
```diff
@@ -0,0 +1,35 @@
...
+	cmd := &cobra.Command{
+		Use:   "verify-install",
+		Short: "Verifies Tilt Installation",
+	}
+	return cmd
+}
+
+func (c *verifyInstallCmd) run(ctx context.Context, args []string) error {
+	a := analytics.Get(ctx)
+	machineID, _ := a.GlobalTag("machine")
```
nicks, line 25: you can remove this, it's unnecessary

internal/cli/verify_install.go
```diff
@@ -0,0 +1,35 @@
...
+
+	cmdVerifyInstallTags := engineanalytics.CmdTags(map[string]string{
```
nicks, line 27: this too

scripts/install.sh
```diff
@@ -78,4 +78,4 @@ if [[ -z $NO_INSTALL ]]; then
 fi
 
 version_check
-
+tilt verify-install
```
nicks, line 81: can you also add this 
- here: https://github.com/tilt-dev/tilt/blob/master/scripts/install.ps1#L49 (for windows)
- here: https://github.com/tilt-dev/tilt/blob/master/.goreleaser.yml#L49 (for homebrew)

internal/cli/verify_install.go
abdullahzameek: I'm sort of unclear on this...won't we need lines 24-25 to get the "machine" tag from analytics?

nicks: nope! it's all handled for you by the framework. If you look at other calls throughout tilt, none of them do this, and they all have the machine tag attached.

abdullahzameek: Hmmm I'm still sort of confused....
Wouldn't we need `cmdVerifyInstallTags` to do `a.Incr()`?

nicks:
Nope, you can pass nil

abdullahzameek:
@nicks Removed the unnecessary code. and added the call to the Windows and Homebrew scripts!

scripts/install.ps1
```diff
@@ -46,4 +47,5 @@ Remove-Item -Force -Recurse -Path "$extractDir"
 Write-Output "Tilt installed!"
 Write-Output "Run '$dest up' to start."
 Write-Output "Or add $binDir to your PATH"
+tilt verify-install
```
nicks, line 50: this probably needs to be $dest, since, as the comment above notes, this assumes tilt is on your path

(though tbh it might be easier for me to just make this change myself so i can test it on windows if you want to back this out)

scripts/install.ps1
```diff
@@ -9,6 +9,7 @@ $zip = "tilt-" + $version + ".zip"
 $extractDir = "tilt-" + $version
 $binDir = "$HOME\bin"
 $dest = "$binDir\tilt.exe"
+$verifyInstall = "tilt verify-install"
```
nicks, line 12: dead code

scripts/install.ps1
abdullahzameek: Ohhh right...
Yeah, that would actually be great if you could change it on your end and test it (since i don't have access to a Windows machine 😅)

scripts/install.ps1
abdullahzameek: Oops I put that in there thinking I'd be using it with `Invoke-Command` or something, will remove!

nicks:
Wait I made a mistake

We can't merge the change to install.sh until we've done a release

PR #3614, opened 2020-07-24
Recognise Krucible clusters as development clusters

BenElgar:
Hi,

I was trying to use Tilt with Krucible, a platform for creating Kubernetes clusters specifically for development, and having to explicitly whitelist the clusters is a bit of a nuisance. This PR adds support for recognising Krucible Kubernetes clusters as development clusters. I also took the liberty of renaming the function to better reflect the intent.

Krucible: https://usekrucible.com/

Let me know what you think!
Cheers,
Ben


BenElgar:
Hi @nicks! Thank you!

At the moment we don't support any particular in-cluster registry, though users can install one. That being said, it is planned and will hopefully make the integration with tilt easier.

PR #3618, opened 2020-07-27
small windows installer fix

nicks:


jazzdan:
>Return terminates execution of the current function and passes control to the statement immediately after the function call. Exit terminates the current execution session altogether. It also closes the console window and may or may not close ISE depending on what direction the wind is facing.

o.0

nicks:
i knooooooow

PR #3630, opened 2020-07-29
cli: changed hud to legacy

abdullahzameek:


internal/cli/up.go
```diff
@@ -83,7 +83,7 @@ local resources--i.e. those using serve_cmd--are terminated when you exit Tilt.
 	cmd.Flags().StringVar(&updateModeFlag, "update-mode", string(buildcontrol.UpdateModeAuto),
 		fmt.Sprintf("Control the strategy Tilt uses for updating instances. Possible values: %v", buildcontrol.AllUpdateModes))
 	cmd.Flags().StringVar(&c.traceTags, "traceTags", "", "tags to add to spans for easy querying, of the form: key1=val1,key2=val2")
-	cmd.Flags().BoolVar(&c.hud, "hud", true, "If true, tilt will open in HUD mode.")
+	cmd.Flags().BoolVar(&c.legacy, "legacy", true, "If true, tilt will open in Legacy mode.")
```
nicks, line 86: I played around with this and this behavior doesn't feel right to me. see my comment here:
https://app.clubhouse.io/windmill/epic/7733/rename-hud-option-for-tilt-1-0-to-emphasize-legacy-status-of-hud#activity-8636

nicks:
Also, as a general rule of thumb, we CANNOT have breaking changes with no valid intermediate state. i.e., whenever you want to replace an old flag with a new flag you need to have

version 1: old flag exists, new flag does not exist
version 2: old flag and new flag both exist
version 3: new flag exits, old flag remove

internal/cli/up.go
abdullahzameek: Will follow up on that!

abdullahzameek:
> Also, as a general rule of thumb, we CANNOT have breaking changes with no valid intermediate state. i.e., whenever you want to replace an old flag with a new flag you need to have
> 
> version 1: old flag exists, new flag does not exist
> version 2: old flag and new flag both exist
> version 3: new flag exits, old flag remove

Yeah, come to think of it, it would be weird if `"--hud"` disappeared overnight ahaha. Will revert back to the original and come up with an intermediate solution.

PR #3634, opened 2020-07-29
Add support for .dockerignore specified by non-default Dockerfile name

### LINKED ISSUE #3231: Add support for Dockerfile.dockerignore files
From the tilt channel:
> Looks like tilt doesn’t support docker_build from Dockerfile.foo with related Dockerfile.foo.dockerignore. Can anybody confirm this ?

I haven't been able to find official specification of this feature, only a brief reference to it in release notes and in random github comments
https://github.com/moby/moby/issues/12886#issuecomment-480575928
and here
https://github.com/docker/engine/pull/215

It's supposed to be implemented on the buildkit side, though we need to dig in a bit about why it's not working out of the box

salim-runsafe:
Fixes #3231 

I'm not super confident in how this impacts docker-compose, but all tests pass.

### DIFF
--- a/internal/tiltfile/docker_compose.go
+++ b/internal/tiltfile/docker_compose.go
@@ -301,7 +301,7 @@ func (s *tiltfileState) dcServiceToManifest(service *dcService, dcSet dcResource
 	paths = append(paths, dcInfo.LocalPaths()...)
 	paths = append(paths, filepath.Dir(dcSet.tiltfilePath))
 
-	dcInfo = dcInfo.WithDockerignores(s.dockerignoresFromPathsAndContextFilters(paths, []string{}, []string{}))
+	dcInfo = dcInfo.WithDockerignores(s.dockerignoresFromPathsAndContextFilters(paths, []string{}, []string{}, ""))
 
 	localPaths := []string{dcSet.tiltfilePath}
 	for _, p := range paths {
--- a/internal/tiltfile/docker_compose.go
+++ b/internal/tiltfile/docker_compose.go
@@ -301,7 +301,7 @@ func (s *tiltfileState) dcServiceToManifest(service *dcService, dcSet dcResource
 	paths = append(paths, dcInfo.LocalPaths()...)
 	paths = append(paths, filepath.Dir(dcSet.tiltfilePath))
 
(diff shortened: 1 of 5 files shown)


internal/tiltfile/docker_compose.go
```diff
@@ -301,7 +301,7 @@ func (s *tiltfileState) dcServiceToManifest(service *dcService, dcSet dcResource
 	paths = append(paths, dcInfo.LocalPaths()...)
 	paths = append(paths, filepath.Dir(dcSet.tiltfilePath))
 
-	dcInfo = dcInfo.WithDockerignores(s.dockerignoresFromPathsAndContextFilters(paths, []string{}, []string{}))
```
nicks, line 304: i think you need to pass `service.DfPath` here

salim-runsafe: That's exactly what I've been trying to figure out for the last half hour. I wish I'd seen your comment sooner!

internal/tiltfile/tiltfile_docker_compose_test.go
```diff
@@ -157,6 +157,39 @@ services:
...
+	f.load("baz")
+	configPath := f.TempDirFixture.JoinPath("docker-compose.yml")
+	f.assertDcManifest("baz",
+		dcConfigPath([]string{configPath}),
+		dcYAMLRaw(dcYAML),
+		dcDfRaw(simpleDockerfile),
+		// TODO(maia): assert m.tiltFilename
+	)
+
+	expectedConfFiles := []string{"Tiltfile", ".tiltignore", "docker-compose.yml", "baz/alternate-Dockerfile", "baz/alternate-Dockerfile.dockerignore"}
```
salim-runsafe, line 188: @nicks can you confirm that this should not expect `.dockerignore` in the root directory?

nicks: yep!

salim-runsafe:
Thanks for your help with this @nicks. Do you know when the release will be that incorporates this change?

nicks:
i suspect we'll do a release today or tomorrow

nicks:
(but no promises, depends on some other vetting and requests)

PR #3635, opened 2020-07-29
logs: wire deps [ch8434]

maiamcc:
Hello @nicks, @jazzdan,

Please review the following commits I made in branch maiamcc/logs-wire-deps:

ac4c504d0f5e7f488c182fecfd50776f0190eaae (2020-07-29 15:10:08 -0400)
logs: wire deps [ch8434]

### DIFF
--- a/internal/cli/logs.go
+++ b/internal/cli/logs.go
@@ -5,7 +5,6 @@ import (
 	"log"
 	"time"
 
-	"github.com/opentracing/opentracing-go"
 	"github.com/spf13/cobra"
 
 	"github.com/tilt-dev/tilt/internal/hud/server"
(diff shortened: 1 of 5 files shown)


internal/hud/server/logs_reader.go
```diff
@@ -94,17 +90,22 @@ func (ls *LogStreamer) Handle(v proto_webview.View) error {
 
 	return nil
 }
+func StreamLogs(ctx context.Context, host model.WebHost, port model.WebPort, printer *hud.IncrementalPrinter) error {
```
nicks, line 93: model.WebURL might be a better choice here. it will behave correctly if the user specifies 0.0.0.0 (there's a subtle difference between the host we're listening on and the host we're connecting to)

internal/cli/logs.go
```diff
@@ -39,13 +38,14 @@ func (c *logsCmd) run(ctx context.Context, args []string) error {
...
 	}
 
-	reader := server.ProvideWebsockerReader()
-	return reader.Listen(ctx)
+	logDeps, err := wireLogsDeps(ctx, a, "logs")
+	if err != nil {
+		return err
+	}
+
+	return server.StreamLogs(ctx, logDeps.host, logDeps.port, logDeps.printer)
```
nicks, line 50: add the addConnectServerFlags() call to this file while you're here?

internal/hud/server/logs_reader.go
maiamcc: ugh it SHOULD be model.WebURL except that encodes the schema too :( i might play with it and split it out into `model.BaseWebURL` and `model.WebURL`

internal/cli/logs.go
maiamcc: I've got that in a different ticket!

internal/hud/server/logs_reader.go
maiamcc: oh wait duh I can just manipulate the `url.URL` to change the schema doyyy.

Issue #3619, opened 2020-07-27
building custom postgres version never becomes ready


nicks:
Repro steps

1) Use kubedb as here:
https://github.com/tilt-dev/tilt-example-frameworks/tree/master/kubedb-postgres

2) Build a custom version of postgres by using k8s_kind to inject images into PostgresVersion
k8s_kind('PostgresVersion', image_json_path='{.spec.db.image}')

3) Add a resource_dep on the PostgresVersion

Result:
PostgresVersion doesn't create any pods, so the resource never becomes ready

I think we need some sort of `k8s_resource` api to identify that a resource doesn't create pods. Right now we assume that anything with an image has pods, which is not right.

PR #3640, opened 2020-07-30
tiltfile: add an api for changing the readiness mode of a resource

nicks:
Hello @jazzdan, @maiamcc,

Please review the following commits I made in branch nicks/ch8647:

95ef863dd3726c566748fc7340dbfb6d6030270c (2020-07-30 10:01:37 -0400)
tiltfile: add an api for changing the readiness mode of a resource
Helps resolve https://github.com/tilt-dev/tilt/issues/3619

### DIFF
--- a/internal/tiltfile/tiltfile_state.go
+++ b/internal/tiltfile/tiltfile_state.go
@@ -693,6 +693,7 @@ func (s *tiltfileState) assembleK8sV2() error {
 		}
 		if r, ok := s.k8sByName[workload]; ok {
 			r.extraPodSelectors = opts.extraPodSelectors
+			r.podReadinessMode = opts.podReadinessMode
 			r.portForwards = opts.portForwards
 			r.triggerMode = opts.triggerMode
 			r.autoInit = opts.autoInit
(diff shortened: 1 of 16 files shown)


internal/tiltfile/tiltfile_state.go
```diff
@@ -1106,13 +1107,23 @@ func (s *tiltfileState) translateK8s(resources []*k8sResource) ([]model.Manifest
...
+		if podReadinessMode == model.PodReadinessNone {
+			// Auto-infer the readiness mode
+			podReadinessMode = model.PodReadinessWait
+
+			// If the resource was
+			// 1) manually grouped (i.e., we didn't find any images in it)
+			// 2) doesn't have pod selectors, and
+			// 3) doesn't depend on images
+			// assume that it will never create pods.
+			if r.manuallyGrouped && len(r.extraPodSelectors) == 0 && len(r.dependencyIDs) == 0 {
```
jazzdan, line 1120: This still feels overloaded to me, but definitely better than what came before it.

nicks: ya...let me think about this a bit more before merging immediately

nicks: spitballing a bit here, i think whenever we define a new CRD, we need to know:

- how to find the images in it
- how to find any pods it deploys (if they can't be found by owner references)
- if it should not expect pods at all (e.g., PostgresVersion)
- if it should wait for the pods to be ready before building the next resource (e.g., servers)
- if it should wait for the pods to be complete before building the next resource (e.g., jobs)

and it's complicated a bit by the fact that there are both normal CRDs where the image shows up in the same place each time, and more meta CRDs (like HelmRelease) where it might appear in different places

i think we should at least have a way to have per-kind pod readiness configs (iin addition to a per-resource pod readiness config)

jazzdan: ooh yeah per kind would be interesting to explore.

nicks:
ya, i'm going to merge this just to make sure i don't forget, and will try to also add a way to configure this by kind

PR #3655, opened 2020-08-03
tilt logs: add port/resource flags, filter by resource [ch8432 ch8433]

maiamcc:
Hello @nicks, @jazzdan,

Please review the following commits I made in branch maiamcc/filter-resources:

857bdf022b6d19dc5f413bccb760be26297bb410 (2020-08-03 13:03:31 -0400)
tilt logs: add port/resource flags, filter by resource [ch8432 ch8433]

Code review reminders, by giving a LGTM you attest that:

* Commits are adequately tested
* Code is easy to understand and conforms to style guides
* Incomplete code is marked with TODOs
* Code is suitably instrumented with logging and metrics

PR #3664, opened 2020-08-04
logstore: can pass options to logStore.ContinuingLines

maiamcc:
this unblocks #3655 as per [nick's comment](https://github.com/tilt-dev/tilt/pull/3655#discussion_r464553690)

### DIFF
--- a/pkg/model/logstore/logstore.go
+++ b/pkg/model/logstore/logstore.go
@@ -395,6 +395,13 @@ func (s *LogStore) IsLastSegmentUncompleted() bool {
 }
 
 func (s *LogStore) ContinuingLines(checkpoint Checkpoint) []LogLine {
+	return s.ContinuingLinesWithOptions(checkpoint, nil, true)
+}
+
+// ContinuingLines, but configurable.
+// `manifestNames` is a list of manifest names to filter on (i.e. only return logs for
+//     those manifests). If this list is empty, return logs for all manifests.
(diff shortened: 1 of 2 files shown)


pkg/model/logstore/logstore.go
```diff
@@ -395,6 +395,13 @@ func (s *LogStore) IsLastSegmentUncompleted() bool {
 }
 
 func (s *LogStore) ContinuingLines(checkpoint Checkpoint) []LogLine {
+	return s.ContinuingLinesWithOptions(checkpoint, nil, true)
+}
+
+// ContinuingLines, but configurable.
+// `manifestNames` is a list of manifest names to filter on (i.e. only return logs for
+//     those manifests). If this list is empty, return logs for all manifests.
+func (s *LogStore) ContinuingLinesWithOptions(checkpoint Checkpoint, manifestNames []model.ManifestName, showPrefix bool) []LogLine {
```
nicks, line 404: fwiw, i would probably make an options struct for any optional arguments

pkg/model/logstore/logstore.go
```diff
@@ -422,9 +429,13 @@ func (s *LogStore) ContinuingLines(checkpoint Checkpoint) []LogLine {
 	}
 	tempLogStore.recomputeDerivedValues()
 
+	spanOptions := tempLogStore.spans
```
nicks, line 432: i think you need to tweak the isSameSpanContinuation/isChangingSpanContinuation logic above, to only apply if you're printing the span at all

maiamcc: @nicks okay I thiiiink I fixed this, can you double check my logic + test cases?

Issue #3665, opened 2020-08-04
docker-compose builds not building in parallel


nicks:
Repro steps:
1) checkout this repo
https://github.com/tilt-dev/express-redis-docker
2) use this docker-compose.yml
```
version: "3.3"
services:
  frontend1:
    build:
      context: .
  frontend2:
    build:
      context: .
```

3) tilt up

Expected:
tilt builds the 2 frontends in parallel

Actual:
tilt says it's running the 2 frontends in parallel, but builds them in serial

Notes: The easy workaround for this is to put the docker builds in the Tiltfile instead of in the compose file. This bug is a consequence of:
1) docker-compose is not thread safe, and will fail in all sorts of weird ways if you have two docker-compose up sessions running concurrently (see: https://github.com/tilt-dev/tilt/issues/2817), but 
2) tilt's execution model means it needs to invoke frontend1 and frontend2 as separate processes

One possible solution to this is to do try to be more intelligent about how we interact with the docker-compose CLI

I also have a long-term dream to make the docker-compose execution model more like the k8s execution model - infer the image builds, inject them into the yaml, then apply the yaml. but this is a much more complex change, but might be more tractable now that there's a spec https://github.com/compose-spec/compose-go

nicks:
here's a snapshot of what this looks like in practice:
https://cloud.tilt.dev/snapshot/AdiGz_ILfncPxdTp098=

PR #3678, opened 2020-08-07
dockercompose: allow docker-compose build to run in parallel, even though docker-compose up cannot

nicks:
Hello @maiamcc, @jazzdan,

Please review the following commits I made in branch nicks/ch8819:

b7e50dc5ebc8cb7d5b5475c2c5c999d3e3d7b1c0 (2020-08-07 11:29:40 -0400)
dockercompose: allow docker-compose build to run in parallel, even though docker-compose up cannot
Fixes https://github.com/tilt-dev/tilt/issues/3665

### DIFF
--- a/internal/dockercompose/client.go
+++ b/internal/dockercompose/client.go
@@ -44,34 +44,49 @@ func NewDockerComposeClient(env docker.LocalEnv) DockerComposeClient {
 }
 
 func (c *cmdDCClient) Up(ctx context.Context, configPaths []string, serviceName model.TargetName, shouldBuild bool, stdout, stderr io.Writer) error {
-	// docker-compose up is not thread-safe, because network operations are non-atomic. See:
-	// https://github.com/tilt-dev/tilt/issues/2817
-	c.mu.Lock()
-	defer c.mu.Unlock()
-
-	var args []string
+	var genArgs []string
 	if logger.Get(ctx).Level().ShouldDisplay(logger.VerboseLvl) {
-		args = []string{"--verbose"}
+		genArgs = []string{"--verbose"}
 	}
 
 	for _, config := range configPaths {
-		args = append(args, "-f", config)
+		genArgs = append(genArgs, "-f", config)
 	}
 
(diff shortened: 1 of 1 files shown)


nicks:
wahhhh

maiamcc:
> wahhhh

...Oh dear. You okay there, @nicks?

nicks:
Just my response to taking DM off the reviewer list

internal/dockercompose/client.go
```diff
@@ -44,34 +44,46 @@ func NewDockerComposeClient(env docker.LocalEnv) DockerComposeClient {
...
+		cmd.Stderr = stderr
+		err := cmd.Run()
+		if err != nil {
+			return FormatError(cmd, nil, err)
+		}
+	}
+
+	// docker-compose up is not thread-safe, because network operations are non-atomic. See:
+	// https://github.com/tilt-dev/tilt/issues/2817
+	// TODO(nick): It might make sense to use a CondVar so that we can log a message
```
maiamcc, line 70: (maybe worth adding a note that `docker-compose build` (as called above) CAN run in parallel, so we don't need to lock the mutex for it?)

maiamcc:
RIP dan miller :(

internal/dockercompose/client.go
```diff
@@ -44,34 +44,46 @@ func NewDockerComposeClient(env docker.LocalEnv) DockerComposeClient {
...
+	// when we're waiting on another build...
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	runArgs := append([]string{}, genArgs...)
+	runArgs = append(runArgs, "up", "--no-deps", "-d")
```
maiamcc, line 75: (I assume `docker-compose up` is smart enough to not build the image if it was just built with `docker-compose build`?)

nicks: ya, i believe it's pretty smart about that, but i can set --no-build to emphasize this point

internal/dockercompose/client.go
nicks: done

PR #3682, opened 2020-08-10
engine: record update stats (with more tags) for ALL BaDs

maiamcc:
(If we really don't want to make changes to internal analytics right now
I can rescind this)

was talking to @victorwuky about how to get events for user updates,
realized what we have in place right now is a. kludgy and b. incompletely
instrumented--wdyt of this instead?

### DIFF
--- a/internal/engine/build_and_deployer.go
+++ b/internal/engine/build_and_deployer.go
@@ -4,12 +4,9 @@ import (
 	"context"
 	"fmt"
 	"strings"
-	"time"
 
 	"go.opentelemetry.io/otel/api/core"
 	"go.opentelemetry.io/otel/api/trace"
-
-	"github.com/tilt-dev/tilt/internal/analytics"
 
 	"github.com/tilt-dev/tilt/internal/container"
 	"github.com/tilt-dev/tilt/internal/engine/buildcontrol"
@@ -135,18 +132,4 @@ func DefaultBuildOrder(lubad *LiveUpdateBuildAndDeployer, ibad *ImageBuildAndDep
 	}
 
 	return BuildOrder{lubad, dcbad, ibad, ltbad}
-}
-
-func sendBuildCompleteTiming(ctx context.Context, startTime time.Time, updateType string, err error) {
-	hasErr := fmt.Sprintf("%t", err != nil)
-
-	analytics.Get(ctx).Timer("update", time.Since(startTime), map[string]string{
-		"type":     updateType,
-		"hasError": hasErr,
-	})
-
-	// legacy format ("build.image" etc.)
-	analytics.Get(ctx).Timer(fmt.Sprintf("build.%s", updateType), time.Since(startTime), map[string]string{
-		"hasError": hasErr,
-	})
 }
--- a/internal/engine/build_and_deployer.go
+++ b/internal/engine/build_and_deployer.go
@@ -4,9 +4,12 @@ import (
 	"context"
 	"fmt"
 	"strings"
+	"time"
 
 	"go.opentelemetry.io/otel/api/core"
 	"go.opentelemetry.io/otel/api/trace"
+
+	"github.com/tilt-dev/tilt/internal/analytics"
 
 	"github.com/tilt-dev/tilt/internal/container"
 	"github.com/tilt-dev/tilt/internal/engine/buildcontrol"
@@ -132,4 +135,18 @@ func DefaultBuildOrder(lubad *LiveUpdateBuildAndDeployer, ibad *ImageBuildAndDep
 	}
 
 	return BuildOrder{lubad, dcbad, ibad, ltbad}
+}
+
+func sendBuildCompleteTiming(ctx context.Context, startTime time.Time, updateType string, err error) {
+	hasErr := fmt.Sprintf("%t", err != nil)
+
+	analytics.Get(ctx).Timer("update", time.Since(startTime), map[string]string{
+		"type":     updateType,
+		"hasError": hasErr,
+	})
+
+	// legacy format ("build.image" etc.)
+	analytics.Get(ctx).Timer(fmt.Sprintf("build.%s", updateType), time.Since(startTime), map[string]string{
+		"hasError": hasErr,
+	})
 }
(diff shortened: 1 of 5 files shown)


internal/engine/image_build_and_deployer.go
```diff
@@ -115,8 +115,10 @@ func (ibd *ImageBuildAndDeployer) BuildAndDeploy(ctx context.Context, st store.R
 
 	startTime := time.Now()
 	defer func() {
-		// todo: image tag?
-		ibd.analytics.Timer("build.image", time.Since(startTime), nil)
+		ibd.analytics.Timer("update", time.Since(startTime), map[string]string{
```
nicks, line 118: are we concerned about breaking the existing graphs?

maiamcc: mm... fair question. I assumed we weren't, but if we are, I can certainly leave the old Timer line in there as well

nicks: as a general rule of thumb, with data work, the teams I've worked with have had a strong bias towards following existing convention. 

Changing the names of things creates a lot of work for any future person that has to analyze the data, or a data engineer that has to clean up the data.

victorwuky: Yeah I would advise against changing names of things. We have legacy data (which we could potentially backfill), but also old versions of Tilt in the wild that are sending data using the legacy format. I think a safer approach is adding _new_ data, and leave old things alone.

maiamcc: I like the new nicer data because it nests EVERYTHING under `update` so you can query for all `update` events OR check the type of update by filtering on tags (instead of looking for `name == build.image || name == build.container || ...`). But that's a fair point about changing the schema. Here's a version that logs data in both formats.

internal/engine/build_and_deployer.go
```diff
@@ -133,3 +136,17 @@ func DefaultBuildOrder(lubad *LiveUpdateBuildAndDeployer, ibad *ImageBuildAndDep
 
 	return BuildOrder{lubad, dcbad, ibad, ltbad}
 }
+
+func sendBuildCompleteTiming(ctx context.Context, startTime time.Time, updateType string, err error) {
+	hasErr := fmt.Sprintf("%t", err != nil)
+
+	analytics.Get(ctx).Timer("update", time.Since(startTime), map[string]string{
```
nicks, line 143: Sorry I think you're misunderstanding my objection to this change. Let me try again, zooming out a bit to give you a bit of context

Most metrics frameworks have strong semantic conventions around what how metrics should be named and tagged. They have long docs on this. They don't do this because metrics people like writing long docs! They do it because it helps a ton when we need to do long-term analysis to have agreed upon conventions. 

We don't want the convention to be "whatever is easiest for the next query we want to write", since the format has big long-term consequences

What I'm asking for here is not "let's add more formats for continuity". I'm asking you to do a a small amount of upfront work to explain:

- why is this new format better?
- what existing conventions are you basing it on?
- if you're introducing a new convention, have you considered the trade-offs?

Here's a typical doc in this space:
https://github.com/opentracing/specification/blob/master/specification.md#start-a-new-span
(Though by no means the only doc)

If possible, I would rather keep the same "build.[type]" convention for now. I think the general "one metric name per function" is a good one. And as we move towards opencensus metrics, we should consider more thoroughly what our tag and name architecture should look like

internal/engine/image_build_and_deployer.go
victorwuky: Thanks @maiamcc to responding to my feedback. But sorry I waded into here when I shouldn't have. I'll just say that my desired high-level requirement is to make it as easy and simple as possible to write queries and that they don't break with new changes. I understand that from time to time there are other tradeoffs that make it so this isn't always possible. I'm just happy to be able to run queries here in the first place! So whatever you end up doing, as long as you let me know soon-ish afterward of changes (and that we can keep the [docs](https://github.com/tilt-dev/tilt-warehouse/blob/master/SCHEMA.md#tiltbuildimage) updated), that's good enough for me!

:homer-bushes-gif:

internal/engine/build_and_deployer.go
maiamcc: I thought I spoke to my reasoning behind this change above:
> I like [this new way of reporting the metric] because it nests EVERYTHING under `update` so you can query for all `update` events OR check the type of update by filtering on tags (instead of looking for `name == build.image || name == build.container || ...`).
To give more context, I had a talk with Victor yesterday in which we wanted to know whether ANY builds had been done (not caring about specific types) and it was doable-but-kind-of-a-pain to pull metrics for all the specific builds by name, plus it'd be be annoying to keep queries up to date if we add new build types.

Without knowing the specifics of how we retrieve data from the DB, I would ASSUME that this way of storing data (as an "update" event with a "type" tag) gives all the functionality of the previous system with additional ease of looking at all update events at once. (Though there may be e.g. performance nuances of filtering by tags that i'm missing.)

But sounds like either you find that reasoning unpersuasive or incomplete, or that the burden of proof for changing this is higher than I imagined, which is fine. I can just stick to the existing schema for now and we can discuss this more later/if the problems I anticipate keep being annoying.

nicks: ya, this might be where we're talking past each other - "this was easiest for the query i was writing right now" isn't really the level of analysis I was looking for, and we'll leave it at that.

internal/engine/local_target_build_and_deployer.go
```diff
@@ -30,6 +32,14 @@ func (bd *LocalTargetBuildAndDeployer) BuildAndDeploy(ctx context.Context, st st
 			"LocalTargetBuildAndDeployer requires exactly one LocalTarget (got %d)", len(targets))
 	}
 
+	startTime := time.Now()
+	defer func() {
+		analytics.Get(ctx).Timer("build.local", time.Since(startTime), map[string]string{
+			"type":     "local",
```
nicks, line 38: remove 'type'?

internal/engine/build_and_deployer.go
maiamcc: ouch, I don't think that summarizes what I was saying. it sounds like you think I'm ignoring your request/criteria, which I'm not, I'm just not clear on what you're looking for or what would be a sufficiently compelling argument.

PR #3684, opened 2020-08-11
Remove no-browser mode

victorwuky:
Since already have `hud`, `legacy`, and `stream` modes.

### DIFF
--- a/internal/hud/server/controller.go
+++ b/internal/hud/server/controller.go
@@ -8,7 +8,6 @@ import (
 	"sync/atomic"
 	"time"
 
-	"github.com/pkg/browser"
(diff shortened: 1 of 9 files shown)


internal/hud/server/controller.go
```diff
@@ -52,7 +50,7 @@ func (s *HeadsUpServerController) isWebsocketConnected() bool {
 }
 
 func (s *HeadsUpServerController) maybeOpenBrowser(st store.RStore) {
-	if s.webURL.Empty() || s.webLoadDone || (bool)(s.noBrowser) {
```
nicks, line 55: this isn't right...noBrowser is always true now, so you need to remove this whole function

PR #3693, opened 2020-08-12
tiltfile: add a force_parallel option for local_resource

nicks:
Hello @landism,

Please review the following commits I made in branch nicks/ch8725:

12499b4450137ed63d0cb2e925174f74633140b7 (2020-08-12 17:45:22 -0400)
tiltfile: add a force_parallel option for local_resource

### DIFF
--- a/internal/engine/buildcontrol/build_control.go
+++ b/internal/engine/buildcontrol/build_control.go
@@ -19,17 +19,16 @@ func NextTargetToBuild(state store.EngineState) *store.ManifestTarget {
 		return nil
 	}
 
-	// Local targets aren't parallelizable with any other kind of target.
-	// So if we're already building a local target, bail immediately.
-	if IsBuildingLocalTarget(state) {
+	// If we're already building an unparallelizable local target, bail immediately.
+	if IsBuildingUnparallelizableLocalTarget(state) {
 		return nil
 	}
 
 	targets := state.Targets()
 	if IsBuildingAnything(state) {
-		// Local targets aren't parallelizable with any other kind of target.
-		// So if we're already building any targets, remove all the local ones.
-		targets = RemoveLocalTargets(targets)
+		// If we're building a target already, remove anything that's not parallelizable
+		// with what's currently building.
+		targets = RemoveUnparallelizableLocalTargets(targets)
 	}
 
 	// Uncategorized YAML might contain namespaces or volumes that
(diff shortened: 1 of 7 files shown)


internal/tiltfile/local_resource.go
```diff
@@ -45,6 +47,7 @@ func (s *tiltfileState) localResource(thread *starlark.Thread, fn *starlark.Buil
 		"serve_cmd?", &serveCmdVal,
 		"cmd_bat?", &updateCmdBatVal,
 		"serve_cmd_bat?", &serveCmdBatVal,
+		"force_parallel?", &forceParallel,
```
landism, line 50: `allow_parallel` would feel a bit clearer to me, but not a blocker

pkg/model/local_target.go
```diff
@@ -15,6 +15,10 @@ type LocalTarget struct {
 	ignores   []Dockerignore
 
 	repos []LocalGitRepo
+
+	// Indicates that we should force this to run in parallel with other local
```
landism, line 19:
```suggestion
	// Indicates that we should allow this to run in parallel with other
```

1. maybe this highlights why "force_parallel" feels weird to me - what does that mean if it's the only target eligible to build?
2. the "other resources" are not limited to local

pkg/model/local_target.go
```diff
@@ -15,6 +15,10 @@ type LocalTarget struct {
...
+	// resources, even if it's not safe to do so.
```
landism, line 20:
```suggestion
	// resources (by default, this is presumed unsafe and is not allowed).
```

just to make this better communicate that it's not "we know it's not safe", but "we don't know it is safe"

internal/engine/buildcontrol/build_control.go
```diff
@@ -21,7 +21,7 @@ func NextTargetToBuild(state store.EngineState) *store.ManifestTarget {
 
 	// Local targets aren't parallelizable with any other kind of target.
 	// So if we're already building a local target, bail immediately.
-	if IsBuildingLocalTarget(state) {
+	if IsBuildingUnparallelizableLocalTarget(state) {
```
landism, line 24: unless I've misunderstood the intent of this PR, we also need to change the `RemoveLocalTargets` bit a few lines down

internal/tiltfile/local_resource.go
nicks: done!

pkg/model/local_target.go
nicks: done

pkg/model/local_target.go
nicks: done

internal/engine/buildcontrol/build_control.go
nicks: oooh good catch! fixed and added a test

internal/engine/buildcontrol/build_control.go
```diff
@@ -21,15 +21,15 @@ func NextTargetToBuild(state store.EngineState) *store.ManifestTarget {
...
 		return nil
 	}
 
 	targets := state.Targets()
 	if IsBuildingAnything(state) {
-		// Local targets aren't parallelizable with any other kind of target.
+		// Some local targets aren't parallelizable with any other kind of target.
 		// So if we're already building any targets, remove all the local ones.
```
landism, line 31:
```suggestion
		// Some local targets aren't parallelizable with any other kind of target.
```
just since "all the local ones" is no longer accurate, a better wording doesn't come immediately to mind, and the function name does a pretty good job of communicating this anyway, might as well delete this second line

nicks: reworded a bit

PR #3694, opened 2020-08-12
cli: add subcommand tag to all metrics we send [ch8887]

maiamcc:
So we can filter out `tilt ci` results for all relevant metrics
(currently, only `tilt.up.running` is tagged with `subcommand`)

cc @victorwuky


maiamcc:
...wow that took a long time for tests to go green 😅

Issue #3692, opened 2020-08-12
tilt logs: don't necessarily tail


maiamcc:
Some users will want to use `tilt logs` to _stream_ logs, others will want a snapshot of the logs at that point in time. There should be optionality around this.

PR #3697, opened 2020-08-12
tilt logs: add --follow flag, by default don't follow [ch8923]

maiamcc:
Resolves issue #3692.

I'm really on the fence about
a. the `--follow` flag naming (trying to imitate `tail`, but also `-f` exists on `tilt up` and does a different thing
b. whether the default should be following or not following

Will gladly accept roll with any (even weak) opinions here.

### DIFF
--- a/internal/hud/server/logs_reader.go
+++ b/internal/hud/server/logs_reader.go
@@ -28,20 +28,22 @@ type WebsocketReader struct {
 	conn         WebsocketConn
 	marshaller   jsonpb.Marshaler
 	unmarshaller jsonpb.Unmarshaler
+	keepAlive    bool // whether to keep listening on websocket, or close after first message
 	handler      ViewHandler
 }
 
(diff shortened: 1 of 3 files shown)


internal/hud/server/logs_reader.go
```diff
@@ -28,20 +28,22 @@ type WebsocketReader struct {
 	conn         WebsocketConn
 	marshaller   jsonpb.Marshaler
 	unmarshaller jsonpb.Unmarshaler
+	keepAlive    bool // whether to keep listening on websocket, or close after first message
```
nicks, line 31: fwiw, i would probably not call this keepAlive, because that's a pretty common term of art in this space related to the mechanism for keeping connections open, and you'll tend to see it a lot in networking code

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Keep-Alive
https://pkg.go.dev/google.golang.org/grpc@v1.31.0?tab=doc#WithKeepaliveParams

maiamcc: ah yep -- I remembered that this term was already A Thing but now how _specific_ a thing it was.

PR #3698, opened 2020-08-12
tiltfile: add a stringable type to make it easier to unpack strings

nicks:
Hello @maiamcc, @landism,

Please review the following commits I made in branch nicks/stringable:

0422a04e761f78a4880c43997f1b2db003517a69 (2020-08-12 19:35:49 -0400)
tiltfile: add a stringable type to make it easier to unpack strings


nicks:
@landism ya, possibly! one thing I went back and forth on is if we should have a string Unpacker that trims the whitespace/newlines off blobs, or matches them against some validity regexp, or if that would be confusing

PR #3703, opened 2020-08-14
release: Use a Docker container to do cross-compilation of the Tilt binaries

nicks:
Hello @landism,

Please review the following commits I made in branch nicks/crosscompile:

616450ce92852a72165965b8eb42696ece508e19 (2020-08-14 14:39:48 -0400)
release: Use a Docker container to do cross-compilation of the Tilt binaries


scripts/release.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
```
landism, line 1: CONTRIBUTING.md points at `make release`, which I think should now run this script instead of goreleaser

nicks: oops! i meant to change `make release` to point to this script. updated the release documentation!

PR #3709, opened 2020-08-17
Revert "web: report log highlights to analytics (#3564)"

nicks:
Hello @landism, @victorwuky,

Please review the following commits I made in branch nicks/selection:

d40e89ccc21a6bc3d1a35bf2f70a33f884d1044b (2020-08-17 11:46:20 -0400)
Revert "web: report log highlights to analytics (#3564)"
It sounds like we have what we need from this, and don't need
to log this anymore, right?

This reverts commit 022b45ca268da13c50d5dbc2c25755b6649e4331.

### DIFF
--- a/web/src/HUD.test.tsx
+++ b/web/src/HUD.test.tsx
@@ -10,9 +10,8 @@ import {
   oneResourceNoAlerts,
 } from "./testdata"
 import { createMemoryHistory } from "history"
-import { SnapshotHighlight, SocketState } from "./types"
+import { SocketState } from "./types"
 import ReactModal from "react-modal"
-import { Count, memoryIncr } from "./analytics"
 
 ReactModal.setAppElement(document.body)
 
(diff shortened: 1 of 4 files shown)


victorwuky:
@nicks : Is there a reason we want to get rid of this, and say, keep other things, like clicks of `Logs` and `Alerts` tabs?

nicks:
sure!

There's value in seeing how often people click those tabs over time, because this may reflect changes in how people understand Tilt.

There's less value in seeing changes to selection behavior over time, because we don't have any reason to believe that your selection behavior reveals any attitude about Tilt the product.

Are you worried that tracking clicks on tabs is overbroad? or do you think there's reason to think selection behavior will change?

victorwuky:
> Are you worried that tracking clicks on tabs is overbroad? 

@nicks : No I don't think so. I'd rather keep as many things as we can.

> or do you think there's reason to think selection behavior will change?

I don't know. And isn't that the point of analytics? To gather the information and observe it over time, as things might change. For example, perhaps we decide to create a feature that surfaces the create snapshot CTA upon highlighting text. I think it would be great to keep this then right?

Overall, I don't mind if we get rid of this specific one though. I can raise this point again for discussion if there's one I feel more strongly about later.

nicks:
sgtm

Issue #3582, opened 2020-07-14
how should tilt behave when you switch git branches?


nicks:
If you have Tilt running, and you switch git branches, the experience isn't great. Tilt will pick up a ton of file changes and re-deploy all the services. This is almost never what you want. What would be a better experience?

Possible solutions:

- Tilt could pause and pop up a prompt in the web UI to confirm that you've switched branches. You would have to explicitly confirm that you want to re-deploy everything

- Some teams we work with use feature branches to work on different subsets of services. For example, branch A might be for service X + Y, and branch B might be for service Y + Z. 

If you switch, maybe we check if you have Tilt configs, and we ask you if you want to set new tilt args? (https://docs.tilt.dev/tiltfile_config.html)

Or maybe Tilt could remember the tilt args associated with a branch.

Other Questions:
- Is there ever a case where you'd want service X from branch A and service Z from branch B? Or would this behavior be too confusing / hard to reason about? Maybe you want all your services to come from the same branch

joseph-zhong:
Thanks Nick for opening this issue, this is an interesting case and deserves some more attention

I'm thinking that the way Tilt should handle git branching could be analogous for a developer working on multiple branches with two clones of their src repo: simultaneously able to work on both long-living deployment A or deployment B, without having to teardown/rebuild before switching, which is especially helpful if those steps are expensive.

That said, I'm thinking having Tilt "pausing" with a prompt before re-deploying seems like a flexible solution that could work. 

I'm thinking that redeployment is almost never actually what users want if it's much more expensive than git-branch switching, and I'm wondering if it'd be a good idea to re-deploy only when strictly necessary.

If we assume that users are already using git, would it be possible to associate branch names or commit SHAs to a particular Tilt image build?

E.g. Say a user switches to branch-Z without first running `tilt down`, would it be possible for tilt to "gracefully reconnect" to a pre-existing branch-Z deployment if the branch name or git SHA associated with image name hasn't changed since the last deployment?
Tilt could then decide to manually re-deploy only when there isn't a pre-existing deployment or if there had been changes

nicks:
I'm going to start by adding a little pause mode to the engine, just to make this a little less painful when it happens, and have it ask for confirmation before restarting auto-builds.

The other part of this (e.g., having it reconnect automatically when you switch branches) I'm going to save for https://github.com/tilt-dev/tilt/issues/3690, which i think gets at it a bit better

nicks:
[...]

PR #3712, opened 2020-08-18
fswatch: create a subscriber that watches for git branch changes

nicks:
Hello @maiamcc, @landism,

Please review the following commits I made in branch nicks/gitwatcher:

e0eaa503600b589859e310e0b9cc88be9855e701 (2020-08-18 16:36:02 -0400)
fswatch: create a subscriber that watches for git branch changes
Allows us to detect and try different UI treatments when the user
changes their git branch, as described in https://github.com/tilt-dev/tilt/issues/3582

d40e89ccc21a6bc3d1a35bf2f70a33f884d1044b (2020-08-17 11:46:20 -0400)
Revert "web: report log highlights to analytics (#3564)"
It sounds like we have what we need from this, and don't need
to log this anymore, right?

This reverts commit 022b45ca268da13c50d5dbc2c25755b6649e4331.


maiamcc:
1. can you split the revert out into its own commit?
2. happy to review, but I'm curious: is this "switching branch behavior" part of a prioritized workstream? just a fast thing you think you can knock off on your own? something else?

nicks:
re: (1) - oops, sorry, i just messed up my git branches
re: (2) - ya, not a prioritized epic, just something i was hoping to knock out quick. one of the teams i was working with say their devs complain about it a lot (see issue description)

nicks:
(and i think i could be convinced that we should have a more well-defined epic for this, with clearer success criteria, even if it's kind of a mini-epic)

internal/engine/fswatch/gitmanager.go
```diff
@@ -0,0 +1,138 @@
...
+
+	for _, repo := range setup {
+		m.repoWatches[repo] = repoNotifyCancel{repo: repo}
+
+		watcher, err := m.fsWatcherMaker(
+			[]string{gitHeadPath(repo)},
+			watch.EmptyMatcher{},
+			logger.Get(ctx))
+		if err != nil {
+			logger.Get(ctx).Debugf("Error watching git branches: %s", repo.LocalPath)
```
maiamcc, line 86:
```suggestion
			logger.Get(ctx).Debugf("Error making watcher for git branches: %s", repo.LocalPath)
```

internal/engine/fswatch/gitmanager.go
```diff
@@ -0,0 +1,138 @@
...
+	return setup, teardown
+}
+
+func (m *GitManager) OnChange(ctx context.Context, st store.RStore) {
+	setup, teardown := m.diff(ctx, st)
+
+	for _, repo := range teardown {
+		p, ok := m.repoWatches[repo]
+		if !ok || p.notify == nil {
+			continue
```
maiamcc, line 71: not a big deal, but if this the repo exists in `m.repoWatches` w/o being successfully started, this means we'll never REMOVE it from `m.repoWatches` and it'll hit this teardown loop every time (or else if we ever WANT to watch this repo again, we won't attempt to START the watch again b/c it's already in `repoWatches`), right?

internal/engine/fswatch/gitmanager.go
```diff
@@ -0,0 +1,138 @@
...
+		_ = p.notify.Close()
+		p.cancel()
+		delete(m.repoWatches, repo)
+	}
+
+	for _, repo := range setup {
+		m.repoWatches[repo] = repoNotifyCancel{repo: repo}
+
+		watcher, err := m.fsWatcherMaker(
+			[]string{gitHeadPath(repo)},
```
maiamcc, line 82: I think this merits a comment about what's going on ("a filewatcher watching for changes to the file that sets the git head. Changes to this file represent..." etc.)

internal/engine/fswatch/gitmanager.go
```diff
@@ -0,0 +1,138 @@
...
+}
+
+func (m *GitManager) dispatchGitBranchStatus(st store.RStore, repo model.LocalGitRepo) {
+	b, err := ioutil.ReadFile(gitHeadPath(repo))
+	if err != nil {
+		// ignore errors reading the file
+		return
+	}
+
+	st.Dispatch(GitBranchStatusAction{
```
maiamcc, line 129: this fires every time there's a change to where HEAD is pointing, not JUST when the user changes branches, right? (presumably the "are we on a different branch now" logic happens wherever this event is handled?)

1. whence the "status" in the struct name? (esp confusing given that `git status` is its own operation). but maybe more importantly
2. consider giving a more generic name like `GitHeadChangeAction`, if the branch-change-check purely happens in the _handling_ of this event

internal/engine/upper.go
```diff
@@ -516,6 +518,10 @@ func handleFSEvent(
 	}
 }
 
+func handleGitBranchStatus(state *store.EngineState, bsa fswatch.GitBranchStatusAction) {
+	// TODO(nick): Do something with this data.
```
maiamcc, line 522: do you have a plan for how to handle it if we get "aah you changed the branch" signals from all 10 of our resources because they all reside in the same git repo?

internal/engine/fswatch/gitmanager_test.go
```diff
@@ -0,0 +1,87 @@
...
+	"github.com/tilt-dev/tilt/internal/watch"
+
+	"github.com/tilt-dev/tilt/internal/k8s/testyaml"
+	"github.com/tilt-dev/tilt/internal/store"
+	"github.com/tilt-dev/tilt/internal/testutils/manifestbuilder"
+	"github.com/tilt-dev/tilt/internal/testutils/tempdir"
+	"github.com/tilt-dev/tilt/pkg/model"
+)
+
+func TestGitManager(t *testing.T) {
```
maiamcc, line 20: no huge deal but maybe worth testing that an error reading the .git/HEAD file (see ln 126 above) doesn't break anything

nicks: it doesn't seem worthwhile to mock out the file system just to test this

maiamcc: Fair enough

internal/engine/upper.go
maiamcc: ah whoops, i misread a bit of the code, never mind!

PR #3723, opened 2020-08-24
tiltfile/cli: builtin timings mode for tiltfile-results

maiamcc:
this is an hour's worth of work to let someone look at the time it takes
to call various Tiltfile builtins. e.g. for the Tiltfile in the Tilt repo
(which runs linters etc.), here are all builtin calls taking longer that 500ms:
```
$ tilt alpha tiltfile-result --builtin-timings --dur-threshold 500ms
- local("cd . && find . -type f -name \"*.go\"") took 783.932424ms
- local("go list ./...") took 3.562675092s
- local("cd web && find . -type f -name \"*.ts*\" | grep -v node_modules | grep -v __snapshots__") took 609.758001ms
```
and longer than 1s:
```
$ tilt alpha tiltfile-result --builtin-timings --dur-threshold 1s
- local("go list ./...") took 3.377749034s
```

at minimum I think it's safe for Tom to install and run this off a branch;
happy to hear it shouldn't be committed (or that the data I'm gathering
should be stored elsewhere etc.)

### DIFF
--- a/internal/cli/tiltfile_result.go
+++ b/internal/cli/tiltfile_result.go
@@ -104,7 +104,7 @@ func (c *tiltfileResultCmd) run(ctx context.Context, args []string) error {
 	// Instead of printing result JSON, print Builtin Timings instead
 	if c.builtinTimings {
 		if len(tlr.BuiltinCalls) == 0 {
-			return fmt.Errorf("oh no, got no builtin calls :(")
+			return fmt.Errorf("executed Tiltfile, but recorded no Builtin calls")
 		}
 		for _, call := range tlr.BuiltinCalls {
 			if call.Dur < c.durThreshold {
--- a/internal/cli/tiltfile_result.go
+++ b/internal/cli/tiltfile_result.go
@@ -4,6 +4,8 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"regexp"
+	"time"
 
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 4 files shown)


internal/cli/tiltfile_result.go
```diff
@@ -42,17 +50,19 @@ func newTiltfileResultCmd() *tiltfileResultCmd {
...
+Exit code 0: successful Tiltfile evaluation (data printed to stdout)
 Exit code 1: some failure in setup, printing results, etc. (any logs printed to stderr)
 Exit code 5: error when evaluating the Tiltfile, such as syntax error, illegal Tiltfile operation, etc. (any logs printed to stderr)
 
 Run with -v | --verbose to print Tiltfile execution logs on stderr, regardless of whether there was an error.`,
 	}
 
 	addTiltfileFlag(cmd, &c.fileName)
+	cmd.Flags().BoolVarP(&c.builtinTimings, "builtin-timings", "b", false, "If true, print timing data for Tiltfile builtin calls instead of Tiltfile result JSON")
+	cmd.Flags().StringVar(&c.durThreshold, "dur-threshold", "", "Only compatible with Builtin Timings mode. Should be a Go duration string. If passed, only print information about builtin calls lasting this duration and longer.")
```
nicks, line 65: there's actually a builtin duration flag that will do this for you!
https://golang.org/pkg/flag/#DurationVar
https://pkg.go.dev/github.com/spf13/pflag?tab=doc#DurationVar

nicks:
love it!

nicks:
also -- fwiw -- i think it would be reasonable to do a release with this. Passing around manually built binaries is always a bit janky and error-prune (due to js compilation)

internal/cli/tiltfile_result.go
maiamcc: 🙌 🎉

Issue #3324, opened 2020-05-11
tools for debugging the docker build context


nicks:
A common complaint about Tilt is that people have trouble figuring out what files are in their Docker context, or how those files map into the container.

For example, 
- recognizing that Tilt is ignoring .git (https://github.com/windmilleng/tilt/issues/2169)
- You added a `.dockerignore` or `.tiltignore` or `only=` or `ignore=`, but forgot
- You want to know (in advance) how changes to a docker_build invocation would change the docker context

A simple first step might be to have a command like:

```
tilt dump context fe
```

that dumps all the files in the docker build context for image `fe`, and creates a tarball you can look through

A more advanced step might be to have a command like:

```
tilt why path/to/file
```

(similar to `yarn why`) that can tell you to how Tilt applies rules for including or excluding a file.

An even more advanced step might be having a web UI over these two commands

PR #3724, opened 2020-08-24
webview: remove paths watched, which are never used

nicks:
Hello @maiamcc,

Please review the following commits I made in branch nicks/pathswatched:

c1c569fa567e84bd8c7dc034b37fc855b6963c34 (2020-08-24 18:53:39 -0400)
webview: remove paths watched, which are never used


internal/hud/webview/convert.go
```diff
@@ -34,18 +34,18 @@ func StateToProtoView(s store.EngineState, logCheckpoint logstore.Checkpoint) (*
 		ms := mt.State
 
 		var absWatchDirs []string
-		var absWatchPaths []string
-		for _, p := range mt.Manifest.LocalPaths() {
+		for i, p := range mt.Manifest.LocalPaths() {
+			if i > 50 {
```
nicks, line 38: fwiw, i think should replace this with some notion of a base dir for custom builds -- it doesn't make much sense for custom builds with 4000 file inputs

nicks:
Yes! I'm glad you asked!

I think our original thesis was that showing this data to you would help you figure out when your watches aren't working correctly.

In practice, we learned that it just pushes a lot of information in your face that users didn't know what to do with. The problems people have using watches involve much more investigation of how the different systems overlap

After we removed the watch paths from the UI, this guide was written in response to that need: https://docs.tilt.dev/file_changes.html

I do think there have been ideas floating around of a more investigation-based debugger, like 'yarn why', described here: https://github.com/tilt-dev/tilt/issues/3324

PR #3725, opened 2020-08-25
tiltfile: fix a bug in decode_yaml_stream when there are empty entries

nicks:
Hello @landism,

Please review the following commits I made in branch nicks/decodeyaml:

d20281da983a31e9a0cddbbba8778fef25282c1a (2020-08-25 12:49:28 -0400)
tiltfile: fix a bug in decode_yaml_stream when there are empty entries


nicks:
before this change, the entry would be duplicated in the stream

Issue #3719, opened 2020-08-21
Tilt up appears to delete previously created resources


parak:
Given a set of resources that are only managed when tilt up runs (managed via a tilt_subcommand variable), when tilt down runs those resources remain as expected. For example, this is used when a PVC needs to be retained between tilt runs.

Subsequently after the tilt down, when tilt up runs again, these resources seem to get deleted then recreated. The aforementioned PVC gets an age value indicating that it was created on that tilt up run. Looking at the verbose (klog=8) logs, the very first thing logged in the uncategorized section is a list of all the resources that have been retained from a prior tilt up, and shortly thereafter kubectl logs that they are all being deleted. This also deletes other resources, such as secrets, configmaps, serviceaccounts, clusterroles, etc.

The main log message of deletions starts with 'kubectl stdout:' and gives a list of all the resources that have been deleted, mirroring the list logged to uncategorized.

Thanks!

nicks:
Hmmm...I was not able to reproduce this issue. Here's the sandbox I'm playing around in:
https://github.com/tilt-dev/tilt-issue-3719/blob/master/Tiltfile

Does that capture what you're doing, or am i missing something? When I tilt up in that sandbox, it preserves both the configmap and the pvc

landism:
> Looking at the verbose (klog=8) logs, the very first thing logged in the uncategorized section is a list of all the resources that have been retained from a prior tilt up, and shortly thereafter kubectl logs that they are all being deleted.

My hypothesis for what is happening:
1. Tilt applies all mutable k8s entities as a single `kubectl apply` batch.
2. If even one of those entities fails to apply with an error necessitating a delete/create, the whole batch will be deleted/created. ([source](https://github.com/tilt-dev/tilt/blob/7871a0e94a7c018b7d2e16aacaa00f4f4440b3bf/internal/k8s/client.go#L294-L293))
3. One such error is when the annotations are too long ([source](https://github.com/tilt-dev/tilt/blob/7871a0e94a7c018b7d2e16aacaa00f4f4440b3bf/internal/k8s/client.go#L344-L343)), which often happens with ConfigMaps, which is one of the types mentioned in the OP.

If your log shows an error along the lines of "metadata.annotations: Too long: must have at most X bytes" when applying the ConfigMap, that'd confirm this hypothesis. If you have a ConfigMap that's potentially bigger than 256k, that'd be evidence for this hypothesis.

[...]

PR #3739, opened 2020-08-31
k8s: apply resources individually. Fixes https://github.com/tilt-dev/tilt/issues/3719

nicks:
Hello @landism,

Please review the following commits I made in branch nicks/ch9071:

060d81ae8fb5348bb66d709c97f53aa045438d85 (2020-08-31 19:39:08 -0400)
k8s: apply resources individually. Fixes https://github.com/tilt-dev/tilt/issues/3719


nicks:
Ya, let's just move forward with this and see if people complain.

This whole system (subshelling out to kubectl) was intended to be a quick hack until we could revisit it. We really should move to doing this in-process like Helm did here:
https://github.com/helm/helm/blob/master/pkg/kube/client.go#L415
to skip the overhead of re-authenticating with the cluster

PR #3742, opened 2020-09-01
model: changes to deps don't invalidate build

maiamcc:
this addresses the problem i was seeing with my [POC for async deps loading](https://github.com/tilt-dev/kubernetes/blob/maiamcc/async-poc/Tiltfile)
where a change to a resource's Deps would cause the resource to rebuild.
With this PR, a change in CustomBuild.Deps or LocalResource.Deps alone
does not invalidate a build and cause the resource to rebuild.

### DIFF
--- a/internal/sliceutils/sliceutils.go
+++ b/internal/sliceutils/sliceutils.go
@@ -52,6 +52,18 @@ func StringSliceEquals(a, b []string) bool {
 	return true
 }
 
+func StringSliceElementsMatch(a, b []string) bool {
+	if len(a) != len(b) {
+		return false
+	}
+
+	aCopy := append([]string{}, a...)
+	sort.Strings(aCopy)
+	bCopy := append([]string{}, b...)
+	sort.Strings(bCopy)
+	return StringSliceEquals(aCopy, bCopy)
+}
+
 // StringSliceStartsWith returns true if slice A starts with the given elem.
 func StringSliceStartsWith(a []string, elem string) bool {
 	if len(a) == 0 {
--- a/internal/sliceutils/sliceutils.go
+++ b/internal/sliceutils/sliceutils.go
@@ -52,18 +52,6 @@ func StringSliceEquals(a, b []string) bool {
 	return true
 }
 
-func StringSliceElementsMatch(a, b []string) bool {
-	if len(a) != len(b) {
-		return false
-	}
-
-	aCopy := append([]string{}, a...)
-	sort.Strings(aCopy)
-	bCopy := append([]string{}, b...)
-	sort.Strings(bCopy)
-	return StringSliceEquals(aCopy, bCopy)
-}
-
 // StringSliceStartsWith returns true if slice A starts with the given elem.
 func StringSliceStartsWith(a []string, elem string) bool {
 	if len(a) == 0 {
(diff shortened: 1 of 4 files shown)


pkg/model/manifest.go
```diff
@@ -260,9 +260,11 @@ func (m1 Manifest) fieldGroupsEqual(m2 Manifest) (primitivesEq, dockerEq, k8sEq,
 	lt2 := m2.LocalTarget()
 	localEq = DeepEqual(lt1, lt2)
 
-	depsEq = DeepEqual(m1.ResourceDependencies, m2.ResourceDependencies)
+	depsEq = sliceutils.StringSliceElementsMatch(m1.LocalPaths(), m2.LocalPaths())
```
nicks, line 263: I don't understand this check? 

LocalPaths() are derived from the other data models, so e.g., if dockerEq is false, than depsEq is false. Right?

pkg/model/manifest.go
```diff
@@ -460,6 +462,8 @@ var localTargetAllowUnexported = cmp.AllowUnexported(LocalTarget{})
 var selectorAllowUnexported = cmp.AllowUnexported(container.RefSelector{})
 var refSetAllowUnexported = cmp.AllowUnexported(container.RefSet{})
 var registryAllowUnexported = cmp.AllowUnexported(container.Registry{})
+var ignoreCustomBuildDepsField = cmpopts.IgnoreFields(CustomBuild{}, "Deps")
```
nicks, line 465: I'm a little bit nervous about putting these options in a generic DeepEqual function. it seems very likely to bite someone and lead to other bugs.

can we pass these cmpopts in as an option from ChangesInvalidateBuild?

pkg/model/manifest.go
nicks: relatedly, I feel like we should be moving Manifest towards using the DeepEqual infrastructure, rather than adding more manually curated checks

pkg/model/manifest.go
nicks: (i would also be fine with changing the name of the function and all its callers to express that it's only checking for build invalidation)

pkg/model/manifest.go
maiamcc: in `DeepEquals` (which as you point out is a v questionable name) i ignore a handful of fields (`CustomBuild.Deps` and `LocalTarget.Deps` that are ONLY relevant to `LocalPaths()` and nothing else; if one of those fields changes, we want it to NOT invalidate the build (but we still say the manifests are unequal), hence splitting out the `depsEq` (so if all that's different btwn m1 and m2 is `CustomBuild.Deps`, then `dockerEq = true` ("everything but the deps") and `depsEq = false`).

For something like `DockerBuild.BuildPath`, it affects `LocalPaths` but also other stuff (it's used in the image build), so a change in that field DOES mean that `dockerEq = false`. does that make sense/sound reasonable?

nicks: hmmmm....i think i see your point, but...it's going to make this pretty complicated to reason about.

TBH, I'm pretty sure this is all moot, i sent you this PR:
https://github.com/tilt-dev/tilt/pull/3745
which removes the Equal() method. Then that will make a bunch of this complexity disappear, since the return values will be unused

maiamcc:
@nicks it was easier to add your changes (#3745) to mine than vice versa--does this look right?

internal/sliceutils/sliceutils.go
```diff
@@ -52,6 +52,18 @@ func StringSliceEquals(a, b []string) bool {
 	return true
 }
 
+func StringSliceElementsMatch(a, b []string) bool {
```
nicks, line 55: this is dead now

nicks (approved):

internal/sliceutils/sliceutils.go
maiamcc: Good catch ty

Issue #3740, opened 2020-09-01
Does ignore in local_resource support glob pattern?


DatXN:
Hi all, 
I follow this example `https://docs.tilt.dev/example_csharp.html` and use local resource

This is the project structure:

> **Working dir**
> 
> ```
> .
> ├── devops
> │   ├── k8s
> │   │	  ├── Tiltfile2
> │   │	  └── out
> ├── project1
> │   ├── bin
> │   ├── obj
> │   └── ...
> ├── project2
> │   ├── bin
> │   ├── obj
> │   └── ...
> ├── ...
> ├── Tiltfile1
> ├── .dockerignore
> └── ...
> ```

This is the value of my Tiltfile2:

```
local_resource(
    'publish_local_svc_order_api',
    'dotnet publish ... -c Release -o out',
    deps=[
        '../../../'
        ],
    ignore=[
        './../../**/obj',
        '../../../**/bin'
    ]
)

```

This is the value of my Tiltfile1:
```
local_resource(
    'publish_local_svc_order_api',
    'dotnet publish ... -c Release -o out',
    deps=[
        '.'
        ],
    ignore=[
        '**/obj',
        '**/bin'
    ]
)

```
However, when the local build is running the file in **/bin and **/obj is changed and causing the local_resource run into a loop.

So my questions are:
1. **Does ignore support glob pattern? or is this a bug?**

2. I know .dockerignore doesn't allow ignore with pattern coming from a child folder (   './../../**/obj' ) but can Titl.dev support this? It will very helpful if I can put the Tiltfile inside a child folder (for organization purpose).

3. Or do we have the feature to set the "working dir/context" of Tiltfile execution instead of always where the Tiltfile is located.

Thanks, DatXN.

victorwuky:
@nicks : Made this an `enhancement` issue, as a feature request. Let me know if that's incorrect.

nicks:
Ya, to answer your question -
1) local_resource uses the same interpreter as dockerignore, so it doesn't support **. Most build systems, AFAIK, don't support this, or put strong limits on how they support it, because double-star globs can lead to some extreme performance characteristics if you're not careful. So I think it's not likely we'll be able to support this, but am happy to leave the issue open.
2) I'm actually surprised that the `..` doesn't work! I think this is a bug. Let me poke at this a bit and try to get more detail.
[...]

PR #3743, opened 2020-09-01
dockerignore: convert ignore patterns to absolute paths

nicks:
Hello @landism,

Please review the following commits I made in branch nicks/ignore:

e19c0452dddf3a20acd7d3d296c46d944d781964 (2020-09-01 18:07:57 -0400)
dockerignore: convert ignore patterns to absolute paths
In most places in Tilt, we try to use absolute paths everywhere.
So this makes things more consistent with the rest of Tilt, and lets us be
a bit more flexible in how we handle subdirs and parent dirs in ignores.

Fixes https://github.com/tilt-dev/tilt/issues/3740

### DIFF
--- a/internal/watch/notify_test.go
+++ b/internal/watch/notify_test.go
@@ -421,9 +421,6 @@ func TestWatchNonexistentDirectory(t *testing.T) {
 	f := newNotifyFixture(t)
 	defer f.tearDown()
 
-	ignore, _ := dockerignore.NewDockerPatternMatcher(f.paths[0], []string{"./"})
-	f.setIgnore(ignore)
-
 	root := f.JoinPath("root")
 	err := os.Mkdir(root, 0777)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


internal/watch/notify_test.go
```diff
@@ -421,9 +421,6 @@ func TestWatchNonexistentDirectory(t *testing.T) {
 	f := newNotifyFixture(t)
 	defer f.tearDown()
 
-	ignore, _ := dockerignore.NewDockerPatternMatcher(f.paths[0], []string{"./"})
```
nicks, line 424: i'm not totally sure what these were here for, but they're totally bogus -- they prevent the fsync() in this test from working

nicks: oh haha - now i remember, i think they're working around a bug in windows where it gets change events for the directory...let me see what i can do about that...

nicks: fixed! addressed it the same way i fixed it for macos

landism (approved):
oy, this and similar issues have almost tempted me to go through the codebase swapping out `string`s for a `Path` type

nicks:
YESSSSSSSSS

PR #3747, opened 2020-09-02
model: better support for no-build no-push custom_builds

nicks:
Hello @maiamcc,

Please review the following commits I made in branch nicks/no-push:

881e7b7c79314b3284eda2102c8169fe1ad418cc (2020-09-02 17:50:04 -0400)
model: better support for no-build no-push custom_builds

### DIFF
--- a/internal/tiltfile/tiltfile_test.go
+++ b/internal/tiltfile/tiltfile_test.go
@@ -4029,6 +4029,24 @@ k8s_yaml('foo.yaml')
 		f.assertNextManifest("foo").ImageTargets[0].OverrideArgs)
 }
 
+// See comments on ImageTarget#MaybeIgnoreRegistry()
+func TestCustomBuildSkipsLocalDockerAndTagPassedIgnoresLocalRegistry(t *testing.T) {
+	f := newFixture(t)
+	defer f.TearDown()
+
+	f.dockerfile("Dockerfile")
+	f.yaml("foo.yaml", deployment("foo", image("gcr.io/foo")))
+	f.file("Tiltfile", `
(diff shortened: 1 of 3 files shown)


nicks:
this is in response to the discussion here
https://github.com/tilt-dev/tilt-extensions/pull/78
and here
https://github.com/tilt-dev/tilt-extensions/pull/77

i understand this change is pretty weird. But I think it might be the kind of weird that cancels out existing weird, so that we're net less weird.

internal/tiltfile/tiltfile_test.go
```diff
@@ -4029,6 +4029,24 @@ k8s_yaml('foo.yaml')
 		f.assertNextManifest("foo").ImageTargets[0].OverrideArgs)
 }
 
+// See comments on ImageTarget#MaybeIgnoreRegistry()
+func TestCustomBuildNoBuildNoPush(t *testing.T) {
```
maiamcc, line 4033: not apparent what you're testing here b/c "no build no push" is a kinda wibbly translation of "skips local docker and doesn't use expected ref", and also since it's not immediately obvious what the behavior SHOULD be, I'd love it in the test name. How about:
`TestCustomBuildSkipsLocalDockerAndTagPassedIgnoresLocalRegistry` or something? Wordier but much easier to tell what's going on.

maiamcc (approved):
I am deeply afraid, but as far as being deeply afraid, this lgtm

internal/tiltfile/tiltfile_test.go
nicks: done!

someday we'll come up with a custom_build v2!!!

PR #3749, opened 2020-09-03
engine: report metrics to opentelemetry.tilt.dev, and give an example of how to report tilt cli commands

nicks:
Hello @landism, @maiamcc,

Please review the following commits I made in branch nicks/ch9214:

dd45c783105f570298785534290df91675ffdb44 (2020-09-03 12:21:54 -0400)
engine: report metrics to opentelemetry.tilt.dev, and give an example of how to report tilt cli commands

### DIFF
--- a/internal/engine/metrics/controller.go
+++ b/internal/engine/metrics/controller.go
@@ -1,0 +1,78 @@
+package metrics
+
+import (
+	"context"
+	"crypto/tls"
+
+	"contrib.go.opencensus.io/exporter/ocagent"
+	"go.opencensus.io/stats/view"
+	"google.golang.org/grpc/credentials"
+
+	"github.com/tilt-dev/tilt/internal/store"
+	"github.com/tilt-dev/tilt/pkg/logger"
+	"github.com/tilt-dev/tilt/pkg/model"
+)
+
+type Controller struct {
+	exporter *DeferredExporter
+	metrics  model.MetricsSettings
+}
+
+func NewController(exporter *DeferredExporter) *Controller {
+	return &Controller{
+		exporter: exporter,
+	}
+}
+
+func (c *Controller) newMetricsSettings(rStore store.RStore) model.MetricsSettings {
+	state := rStore.RLockState()
+	defer rStore.RUnlockState()
+	return state.MetricsSettings
+}
+
+func (c *Controller) OnChange(ctx context.Context, rStore store.RStore) {
+	newMetricsSettings := c.newMetricsSettings(rStore)
+	oldMetricsSettings := c.metrics
+	if newMetricsSettings == oldMetricsSettings {
+		return
+	}
+
+	c.metrics = newMetricsSettings
+	view.SetReportingPeriod(newMetricsSettings.ReportingPeriod)
+
+	if oldMetricsSettings.Enabled && !newMetricsSettings.Enabled {
+		// shutdown the old metrics
+		err := c.exporter.SetRemote(nil)
+		if err != nil {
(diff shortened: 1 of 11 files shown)


nicks:
Most of this should be pretty straightforward boilerplate:
1) Wrapping the opencensus code so that metrics that come in before tiltfile execution eventually get reported
2) standard store/Subscriber boilerplate that reacts to changes to MetricsSettings

The interesting bits are about the opencensus views. The state of the art for reporting metrics these days is pretty consistent across libraries I looked at (prometheus and statusd and opencensus and opentelemetry and openmetrics all look pretty much the same).

They ask you to "pre-index" how your metrics are going to look, and send a heartbeat to the server with a summary of the metrics as they're collected.

I think the idea is that the metrics are pre-indexed a bit, so that if a server request fails, we can re-send them without the risk of double-counting them

internal/engine/metrics/controller.go
```diff
@@ -0,0 +1,78 @@
...
+	metrics  model.MetricsSettings
+}
+
+func NewController(exporter *DeferredExporter) *Controller {
+	return &Controller{
+		exporter: exporter,
+	}
+}
+
+func (c *Controller) diff(rStore store.RStore) model.MetricsSettings {
```
landism, line 27: should rename or pull the diff logic up into this method

internal/engine/metrics/controller.go
```diff
@@ -0,0 +1,78 @@
...
+	state := rStore.RLockState()
+	defer rStore.RUnlockState()
+	return state.MetricsSettings
+}
+
+func (c *Controller) OnChange(ctx context.Context, rStore store.RStore) {
+	newMetrics := c.diff(rStore)
```
landism, line 34: this name made me think it more directly contained to actual metrics / values - how about "settings" or "metricsSettings"?

internal/engine/metrics/controller.go
```diff
@@ -0,0 +1,78 @@
...
+	oldMetrics := c.metrics
+	if newMetrics == oldMetrics {
+		return
+	}
+
+	c.metrics = newMetrics
+	view.SetReportingPeriod(newMetrics.ReportingPeriod)
+
+	if !newMetrics.Enabled {
```
landism, line 43:
```suggestion
	if oldMetrics.Enabled && !newMetrics.Enabled {
```

nit: so that we don't log "shutting down" when changing settings other than "enabled", or in case someone adds some non-idempotent code to this block thinking it's the "shut down" block that fails when it wasn't running in the first place

internal/engine/metrics/controller_test.go
```diff
@@ -0,0 +1,77 @@
...
+
+	// Verify that changing the metrics settings creates a new remote exporter.
+	ms.Insecure = true
+	f.st.SetState(store.EngineState{
+		MetricsSettings: ms,
+	})
+	f.mc.OnChange(f.ctx, f.st)
+	assert.NotSame(t, remote, f.exp.remote)
+
+	// Verify that disablingthe metrics settings nulls out the remote exporter.
```
landism, line 43:
```suggestion
	// Verify that disabling the metrics settings nulls out the remote exporter.
```

landism (approved):

internal/engine/metrics/controller.go
nicks: done!

internal/engine/metrics/controller.go
nicks: done!

internal/engine/metrics/controller_test.go
nicks: done!

PR #3750, opened 2020-09-03
tiltfile: add map[string]string unpacker

landism:
Having to access the value via `.Map` is a little clumsy, but I'm not sure there's a way to avoid that.

### DIFF
--- a/internal/tiltfile/value/map.go
+++ b/internal/tiltfile/value/map.go
@@ -6,14 +6,12 @@ import (
 	"go.starlark.net/starlark"
 )
 
-type StringStringMap struct {
-	Map map[string]string
-}
+type StringStringMap map[string]string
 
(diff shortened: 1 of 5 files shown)


internal/tiltfile/value/map.go
```diff
@@ -0,0 +1,39 @@
+package value
+
+import (
+	"fmt"
+
+	"go.starlark.net/starlark"
+)
+
+type StringStringMap struct {
```
nicks, line 9: have you considered doing this as 

type StringStringMap map[string]string

?

i think we only need the wrapper struct if we want to treat the "unset" value differently

nicks (approved):

internal/tiltfile/value/map.go
landism: oh, duh

PR #3757, opened 2020-09-04
metrics: only send metrics when logged in. Attach token and other env info

nicks:
Hello @landism,

Please review the following commits I made in branch nicks/metrics:

d907da2425af9d72fcdaf375cbb2b94ef2bc06d6 (2020-09-04 16:30:16 -0400)
metrics: only send metrics when logged in. Attach token and other env info

### DIFF
--- a/internal/engine/metrics/controller.go
+++ b/internal/engine/metrics/controller.go
@@ -3,44 +3,68 @@ package metrics
 import (
 	"context"
 	"crypto/tls"
+	"runtime"
 
 	"contrib.go.opencensus.io/exporter/ocagent"
+	"go.opencensus.io/resource"
 	"go.opencensus.io/stats/view"
 	"google.golang.org/grpc/credentials"
 
+	"github.com/tilt-dev/tilt/internal/git"
 	"github.com/tilt-dev/tilt/internal/store"
+	"github.com/tilt-dev/tilt/internal/token"
 	"github.com/tilt-dev/tilt/pkg/logger"
 	"github.com/tilt-dev/tilt/pkg/model"
 )
 
-type Controller struct {
-	exporter *DeferredExporter
-	metrics  model.MetricsSettings
+type MetricsState struct {
+	settings model.MetricsSettings
+	token    token.Token
+	username string
+	teamID   string
 }
 
-func NewController(exporter *DeferredExporter) *Controller {
+func (s MetricsState) Enabled() bool {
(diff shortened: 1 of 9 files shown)


internal/engine/metrics/controller.go
```diff
@@ -76,3 +101,20 @@ func (c *Controller) OnChange(ctx context.Context, rStore store.RStore) {
...
+		return &resource.Resource{
+			Type: "tilt.dev/tilt",
+			Labels: map[string]string{
+				"os":         runtime.GOOS,
+				"version":    c.tiltBuild.AnalyticsVersion(),
+				"git_origin": c.gitRemote.String(),
+				"team":       state.teamID,
+
+				// We'll inject the username server-side from the token.
+				"token": state.token.String(),
```
landism, line 117: I'm mildly nervous about sending the token as a label:
1. if a user changes the metrics address to a different opentelemetry server for testing purposes, they've now persisted credentials
2. Does our opentelemetry server persist these labels? It'd be nice to not need to worry about secrets being persisted.

A few things that might be worth considering:
(cheap): only include the token if the host is in .tilt.dev (only mitigates (1) above, not (2)).
(a bit less cheap, and maybe weird): hash the token with the address, so it's only valid for authing to that specific address, and have the server side do the same (we'd need to also include the username so we know what token(s) to hash)
(seems nicer but probably a lot more work): could we do send the token via an exporter / grpc option so that it doesn't risk getting persisted as a label? e.g., make a simple proxy endpoint that gets the token from a header, performs auth, and forwards data on to the actual opentelemetry server. That might be more complexity than you want to take on for this, though.

landism (approved):

internal/engine/metrics/controller.go
nicks: Yep! this is a good concern!

My plan is to create an opentelemetry processor that 
1) pulls out the token
2) checks it against the database
3) injects the actual user id
4) throws away the token

opentelemetry has a pretty good plugin system for doing this kind of thing, here's a plugin for adding/removing random labels
https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/master/processor/metricstransformprocessor

i originally was going to implement it as a proxy. But the proxy would need to be smart enough to unpack the GRPC request and verify that the token in the header matches the user id in the message itself. which would probably be fine, but it seems less friction to plug into the existing processor framework.

Does that make sense?

landism: SGTM

PR #3758, opened 2020-09-08
Keyboard shortcut to open snapshot modal

victorwuky:
If you type <kbd>s</kbd>, the snapshot modal is opened.


victorwuky:
@nicks : What do you think about this change?

(I also don't know how to properly install node packages. So I'd need a pointer there if this is an otherwise good change.)

nicks:
deferring to @hyu on this. 

I haven't thought through what our keyboard shortcut strategy should be. i personally would not expect a web app to bind bare letters, and would consider this very surprising in any other web app.

victorwuky:
@nicks : One character keyboard shortcuts I think are common. Gmail, Clubhouse, and GitHub and probably many others all bind <kbd>/</kbd> to the search bar.

PR #3763, opened 2020-09-10
handle wildcard globs correctly. Fixes https://github.com/tilt-dev/tilt/issues/3740

nicks:
Hello @landism,

Please review the following commits I made in branch nicks/ch9348:

8c5ba1858bbb8aa97216702cc1be28eb0ec1f501 (2020-09-10 16:43:51 -0400)
handle wildcard globs correctly. Fixes https://github.com/tilt-dev/tilt/issues/3740


nicks:
Most of this diff is vendoring, it's really just pulling in https://github.com/tilt-dev/dockerignore/pull/4

internal/dockerignore/ignore.go
```diff
@@ -7,14 +7,14 @@ import (
 	"strings"
 
 	"github.com/docker/docker/builder/dockerignore"
-	"github.com/docker/docker/pkg/fileutils"
+	tiltDockerignore "github.com/tilt-dev/dockerignore"
```
landism, line 10: How would you feel about calling this something like tilt-fileutils or tilt-docker-fileutils? I was a bit thrown that this wasn't replacing github.com/docker/docker/builder/dockerignore.

landism (approved):

internal/dockerignore/ignore.go
nicks: ya, i think you're right that we expect this to replace builder/dockerignore! But I want to go in the opposite direction! :)

move code from builder/dockerignore to tilt-dev/dockerignore, and add apis for making it easier to log where the dockerignore patterns came from

The 'fileutils' name is vestigial from when this was for things like CopyFile, but that's not really what we want it for
https://pkg.go.dev/github.com/docker/docker/pkg/fileutils?tab=doc

Issue #3760, opened 2020-09-08
short-term fix for custom_build caching semantics


nicks:
All Tilt image builds (docker_build and custom_build) use content-based immutable tags. In other words,
- Tilt builds the image
- Tilt checks the contents of the image to compute a digest
- Tilt injects the digest into the kubernetes yaml. Kubernetes is smart enough not to redeploy the image if the digest hasn't changed.

For details on this, see:
https://docs.tilt.dev/custom_build.html#why-tilt-uses-immutable-tags

There are two ways this can fall down.

The first is that if the custom_build builds remotely, Tilt doesn't have any way to compute the digest of the output image. This only affects builds that use custom_build(skips_local_docker=True)

The second way is if the custom_build has its own caching mechanism, not based on content hashing. For example if it wants to tag the deployment in some way, and always reconnect to the deployment if the tag matches what it expects.

I can think of a few different options to fix this, based on what stage we want to do the caching:

1) Compute a digest of inputs, as suggested here: https://github.com/tilt-dev/tilt/issues/3690. This puts all the implementation on the Tilt side, but weird things might happen if the user has specified deps wrong.

2) Provide a way for custom_build to specify the digest of the output image. Then Tilt could compare the digest with what's currently deployed, and skip deployment if the digest hasn't changed.

3) Provide a way for custom_build to do its own caching check. i.e., tilt passes in the existing digest/deployment, and the custom_build says "yes, i want to reuse that one!" This might run either pre-image-build or post-image-build

Note that these options aren't redundant - it would be reasonable to implement all of them. (3) is probably the biggest footgun -- a buggy user script could really corrupt your dev environment in weird and difficult-to-diagnose ways. The exact semantics and API might be hard to get right - i'm not totally sure if pre-image-build or post-image-build is right. (2) is probably the safest, but I think it would be hard for users to implement well -- it would be more as a building block for other remote builders.

nicks:
I feel like this is super painful for some teams when it happens. We'd prefer a short-term fix that we can iterate on rather than trying to solve for everyone the first time

victorwuky:
@joseph-zhong @calvinytong: Do you mind reviewing this and seeing if it makes sense?

calvinytong:
This looks reasonable to me. I think 2 would be enough to unblock us. 

[...]

PR #3766, opened 2020-09-11
builder implementation of outputs_image_ref_to for custom build scripts that want to compute their own tag

nicks:
Hello @landism,

Please review the following commits I made in branch nicks/custombuilder:

82d91d839043c16f298908009fc5df0333bd820b (2020-09-11 15:28:27 -0400)
builder implementation of outputs_image_ref_to for custom build scripts that want to compute their own tag

### DIFF
--- a/internal/build/custom_builder_test.go
+++ b/internal/build/custom_builder_test.go
@@ -2,6 +2,7 @@ package build
 
 import (
 	"context"
+	"fmt"
 	"path/filepath"
 	"runtime"
 	"testing"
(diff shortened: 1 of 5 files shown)


nicks:
this is the build-engine side of https://github.com/tilt-dev/tilt/issues/3760

internal/build/custom_builder_test.go
```diff
@@ -140,6 +141,73 @@ func TestCustomBuilderExecsRelativeToTiltfile(t *testing.T) {
...
+
+	cb := model.CustomBuild{
+		WorkDir:           f.tdf.Path(),
+		Command:           model.ToHostCmd("echo 999 > ref.txt"),
+		OutputsImageRefTo: f.tdf.JoinPath("ref.txt"),
+	}
+	_, err := f.cb.Build(f.ctx, refSetFromString("gcr.io/foo/bar"), cb)
+	require.NotNil(t, err)
+	assert.Contains(t, err.Error(),
+		fmt.Sprintf("Output image ref in file %s was invalid: could not parse ref docker.io/library/999", f.tdf.JoinPath("ref.txt")))
```
landism, line 191: nit: maybe ParseNamedTagged should return the original string in the error. If I got this error I'd think it pretty fishy that it's telling me it can't parse the value in my file but not showing me the actual value in the file. Probably worth `%q`ing, too, to get a better error when the file is empty.

landism (approved):

internal/build/custom_builder_test.go
nicks: good point! done

PR #3775, opened 2020-09-14
tiltfile: allow escaping of selector strings

### LINKED ISSUE #3761: Tilt kubernetes object selector should allow escaping for ':' in resource name
Currently it is not possible to reference a resource using the tilt-specific kubernetes object selector if that resource has a name with a ':' in it.

For example, if using the helm_remote extension to deploy cert-manager, it is not possible to group a subset of the created roles and clusterroles created as part of a k8s_resource.

Example of the created roles/clusterroles that exhibit this behavior:
```
     → cert-manager-cainjector:leaderelection:role
     → cert-manager:leaderelection:role
     → cert-manager-webhook:dynamic-serving:role
     → cert-manager-cainjector:leaderelection:rolebinding
     → cert-manager:leaderelection:rolebinding
     → cert-manager-webhook:dynamic-serving:rolebinding
```

landism:
fixes #3761

I wanted to use some existing library for splitting w/ escapes, but I failed to find one.

### DIFF
--- a/internal/sliceutils/sliceutils.go
+++ b/internal/sliceutils/sliceutils.go
@@ -77,3 +77,73 @@ func AppendWithoutDupes(a []string, b ...string) []string {
 
 	return ret
 }
+
+type EscapeSplitOptions struct {
+	Delimiter  rune
+	EscapeChar rune
+}
+
+func NewEscapeSplitOptions() EscapeSplitOptions {
+	return EscapeSplitOptions{
+		Delimiter:  ':',
+		EscapeChar: '\\',
+	}
+}
+
+func UnescapeAndSplit(s string, opts EscapeSplitOptions) ([]string, error) {
+	var parts []string
+	escapeNextChar := false
+	cur := ""
+	for i, r := range s {
+		if escapeNextChar {
+			if r == opts.Delimiter || r == opts.EscapeChar {
+				cur += string(r)
+			} else {
+				// grab the 6 chars around the invalid char to make it easier to find the offending string from the error
+				snippetStart := i - 3
+				if snippetStart < 0 {
+					snippetStart = 0
+				}
+				snippetEnd := i + 3
+				if snippetEnd > len(s) {
+					snippetEnd = len(s)
+				}
+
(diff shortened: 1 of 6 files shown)


internal/sliceutils/sliceutils.go
```diff
@@ -77,3 +77,73 @@ func AppendWithoutDupes(a []string, b ...string) []string {
...
+				snippetStart := i - 3
+				if snippetStart < 0 {
+					snippetStart = 0
+				}
+				snippetEnd := i + 3
+				if snippetEnd > len(s) {
+					snippetEnd = len(s)
+				}
+
+				return nil, fmt.Errorf("invalid escape sequence '%c%c' in '%s'", opts.EscapeChar, r, s[snippetStart:snippetEnd])
```
nicks, line 112: question: this syntax already allows regexps between the colons right? doesn't that mean you can't escape things for the regexp anymore?

nicks (approved):
I'm fine with moving forward on this in the short-term, but I have long-term concerns that this doubles down on an bespoke object selector syntax that no one else uses and doesn't give us any kind of special advantage. 

I wish we had just used a thin starlark layer over ObjectReference (https://pkg.go.dev/k8s.io/api/core/v1?tab=doc#ObjectReference), like objects=[objectReference(name='foo')]

or even better, LabelSelector (https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1?tab=doc#LabelSelector)

internal/sliceutils/sliceutils.go
landism: Nope - the `ObjectSelector` type allows regexes, but the function that converts this syntax into an `ObjectSelector` [calls `regexp.QuoteMeta`](https://github.com/tilt-dev/tilt/blob/816e9a6a9d150f8b543c7bb2112a2c558caff0cd/internal/k8s/object_selector.go#L100) on all of its components.

Other places that use the `ObjectSelector` type and do allow regexes (also highlighting the need for cleaning this up): `k8s_kind`, `k8s_image_json_path`, `filter_yaml`

landism:
> I have long-term concerns that this doubles down on an bespoke object selector syntax that no one else uses and doesn't give us any kind of special advantage.

Yeah, this seems like a good conversation to have at some point.

Issue #3482, opened 2020-06-19
Extension fetcher should grab the whole directory, not just the Tiltfile


maiamcc:
As seen in https://github.com/tilt-dev/tilt-extensions/pull/35, sometimes an extension relies on more files than just the Tiltfile (e.g. supplementary scripts). It would be useful if the Tilt extension fetcher grabbed the whole directory of an extension, not just the Tiltfile. (Though: would there be security risks here?)

nicks:
fixed in v0.17.5

PR #3779, opened 2020-09-15
tiltextension: download the whole directory instead of just the tiltfile. Fixes https://github.com/tilt-dev/tilt/issues/3482

nicks:
Hello @landism,

Please review the following commits I made in branch nicks/ch7839:

e31bc008099ec4cff3e7833c2cc8705ad9d7c1a2 (2020-09-15 10:34:53 -0400)
tiltextension: download the whole directory instead of just the tiltfile. Fixes https://github.com/tilt-dev/tilt/issues/3482


internal/tiltfile/tiltextension/github_fetcher.go
```diff
@@ -4,63 +4,60 @@ import (
...
+	dir, err := f.dlr.Download(path.Join("github.com/tilt-dev/tilt-extensions", moduleName))
 	if err != nil {
-		return ModuleContents{}, err
+		return ModuleContents{}, fmt.Errorf("Fetching tilt-extensions: %v", err)
 	}
 
 	return ModuleContents{
 		Name:              moduleName,
-		TiltfileContents:  string(body),
-		GitCommitHash:     masterSHA,
```
landism, line 63: DM was persisting this to make it easier to deal with versioning in the future, though I'm now unable to recall or find a writeup of what his plan was for it. Do you have confidence it's unimportant, or is it worth thinking about a bit / re-adding?

landism (approved):
LGTM aside from the GitCommitHash question

internal/tiltfile/tiltextension/github_fetcher.go
nicks: go-get has a notion of versioning.  I think it's based on reading all the repo tags, looking at the ones that look like semver strings, and trying to make some intelligent decisions about the latest upgrade-able one.

It doesn't even let you get commit hashes, because it needs to be able to download from systems that don't have commit hashes.

I have real complaints about the way go-get does versioning. But I'd rather inherit those problems than trying to implement our own versioning from scratch.

Will leave a comment on this in the code

Issue #3507, opened 2020-06-24
add a Tiltfile accessor for the root working directory


nicks:
Currently, all `os.path` functions resolve their paths relative to the working directory of the Tiltfile being executed.

This makes it easier to write custom commands that execute in the caller's directory, as discussed here: https://github.com/tilt-dev/tilt/pull/3150

It would also be good to have an API for accessing the root path where tilt_modules and other "global" caches could be stored.

For an example of this use-case, see: https://github.com/tilt-dev/tilt-extensions/pull/40

PR #3784, opened 2020-09-17
tiltfile: add config.tiltfile_path, config.tiltfile_dir. Fixes https://github.com/tilt-dev/tilt/issues/3507

nicks:
Hello @landism,

Please review the following commits I made in branch nicks/ch9402:

4f0bef1c6d13effb52db129dc422b92c07220796 (2020-09-17 09:58:21 -0400)
tiltfile: add config.tiltfile_path, config.tiltfile_dir. Fixes https://github.com/tilt-dev/tilt/issues/3507

### DIFF
--- a/internal/tiltfile/config/config.go
+++ b/internal/tiltfile/config/config.go
@@ -81,6 +81,19 @@ func (e *Extension) OnStart(env *starkit.Environment) error {
 	}
 
 	err := env.AddValue("config.tilt_subcommand", starlark.String(e.TiltSubcommand))
+	if err != nil {
+		return err
+	}
+
+	err = env.AddValue("config.main_path", starlark.String(env.StartPath()))
+	if err != nil {
+		return err
+	}
+
+	err = env.AddValue("config.main_dir", starlark.String(filepath.Dir(env.StartPath())))
+	if err != nil {
+		return err
+	}
 
 	return err
 }
--- a/internal/tiltfile/config/config_test.go
+++ b/internal/tiltfile/config/config_test.go
@@ -459,6 +459,44 @@ print(config.tilt_subcommand)
 	require.Equal(t, "foo\n", f.PrintOutput())
 }
 
+func TestTiltfilePath(t *testing.T) {
+	f := NewFixture(t, model.UserConfigState{}, "foo")
+	defer f.TearDown()
+
+	f.File("foo/Tiltfile", `
+print(config.main_path)
+`)
+	f.File("Tiltfile", `
+include('./foo/Tiltfile')
+print(config.main_path)
+`)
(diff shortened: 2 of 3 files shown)


internal/tiltfile/config/config_test.go
```diff
@@ -459,6 +459,34 @@ print(config.tilt_subcommand)
...
+print(config.tiltfile_dir)
+`)
+
+	_, err := f.ExecFile("Tiltfile")
+	require.NoError(t, err)
+
+	require.Equal(t, f.Path()+"\n", f.PrintOutput())
+}
+
 func NewFixture(tb testing.TB, userConfigState model.UserConfigState, tiltSubcommand model.TiltSubcommand) *starkit.Fixture {
```
landism, line 490: it'd be good to have tests that check the value of `config.tiltfile_path`/dir from inside a `load`ed Tiltfile

internal/tiltfile/config/config.go
```diff
@@ -81,6 +81,19 @@ func (e *Extension) OnStart(env *starkit.Environment) error {
 	}
 
 	err := env.AddValue("config.tilt_subcommand", starlark.String(e.TiltSubcommand))
+	if err != nil {
+		return err
+	}
+
+	err = env.AddValue("config.tiltfile_path", starlark.String(env.StartPath()))
```
landism, line 88: I think it'd be good to communicate via the name whether this is the root or current tiltfile path. `config.entry_tiltfile_path`? `config.entrypoint_tiltfile_path`?

landism (approved):

internal/tiltfile/config/config_test.go
nicks: what about entry_path / entry_dir? or maybe main_path / main_dir ?

was trying to find analogues to this in python but was not coming up with good ones.

internal/tiltfile/config/config.go
nicks: what about entry_path / entry_dir? or maybe main_path / main_dir ?

was trying to find analogues to this in python but was not coming up with good ones.

landism: I think python would be `__main__.__file__`. We've already got `__file__`, but not sure we want to go down the road of exposing `__main__` as a thing.

entry_path or main_path both seem fine to me

nicks: changed it to main_path. ya, agree that __main__ might be a road we eventually go down, but not sure this is the right time.

internal/tiltfile/config/config_test.go
nicks: done

PR #3801, opened 2020-09-25
tiltfile/helm: throw error when we can't parse helm version, more robust version check if version output is dirty [ch9459]

### LINKED ISSUE #3788: Can't get Helm version for Helm v3.3.2 and v3.3.3 (was: --name option not recognized)
when we try to use a tiltfile like

`k8s_yaml(helm('./project/charts/p1', name='project1', namespace='local-project', values=['./values_local.yaml']))`

Error we get:

`--name option not recognized`

Using tilt v0.17.5 and helm v3.3.2

maiamcc:
Hello @nicks,

Please review the following commits I made in branch maiamcc/helm-version-check:

aa38f01a8ec2a855eaf9e0d334288fae12705e5c (2020-09-25 14:01:01 -0400)
tiltfile/helm: throw error when we can't parse helm version, more robust version check if version output is dirty [ch9459]

Fixes #3788 (tho the helm behavior in question is fixed as of v3.3.4 -- see https://github.com/helm/helm/issues/8776)

### DIFF
--- a/internal/tiltfile/helm_test.go
+++ b/internal/tiltfile/helm_test.go
@@ -112,7 +112,8 @@ const exampleHelmV3_2VersionOutput = `v3.2.4`
 // see https://github.com/tilt-dev/tilt/issues/3788
 const exampleHelmV3_3VersionOutput = `WARNING: Kubernetes configuration file is group-readable. This is insecure. Location: /Users/someone/.kube/config
 WARNING: Kubernetes configuration file is world-readable. This is insecure. Location: /Users/someone/.kube/config
-v3.3.3+g55e3ca0`
+v3.3.3+g55e3ca0
+`
 
 func TestParseHelmV2Version(t *testing.T) {
 	expected := helmV2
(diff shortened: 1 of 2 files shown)


internal/tiltfile/helm_test.go
```diff
@@ -109,39 +109,40 @@ const exampleHelmV3_0VersionOutput = `v3.0.0`
 const exampleHelmV3_1VersionOutput = `v3.1.0`
 const exampleHelmV3_2VersionOutput = `v3.2.4`
 
+// see https://github.com/tilt-dev/tilt/issues/3788
+const exampleHelmV3_3VersionOutput = `WARNING: Kubernetes configuration file is group-readable. This is insecure. Location: /Users/someone/.kube/config
+WARNING: Kubernetes configuration file is world-readable. This is insecure. Location: /Users/someone/.kube/config
+v3.3.3+g55e3ca0`
```
nicks, line 115: can you also add a trailing newline to the test case to make sure it still works? since that's how the real output looks

nicks (approved):

PR #3803, opened 2020-09-25
engine: add image build duration metrics

nicks:
Hello @maiamcc,

Please review the following commits I made in branch nicks/latency:

4853ebc8d2ec38528f5e6e54949abccb27f65457 (2020-09-25 16:41:31 -0400)
engine: add image build duration metrics
This is loosely adapted from the opencensus guide on how to group and organize
metrics for consumption by metrics systems.

### DIFF
--- a/internal/engine/image_builder.go
+++ b/internal/engine/image_builder.go
@@ -3,15 +3,55 @@ package engine
 import (
 	"context"
 	"fmt"
+	"time"
 
 	"github.com/docker/distribution/reference"
+	"go.opencensus.io/stats"
+	"go.opencensus.io/stats/view"
+	"go.opencensus.io/tag"
 
 	"github.com/tilt-dev/tilt/internal/build"
 	"github.com/tilt-dev/tilt/internal/container"
 	"github.com/tilt-dev/tilt/internal/engine/buildcontrol"
 	"github.com/tilt-dev/tilt/internal/ignore"
+	"github.com/tilt-dev/tilt/pkg/logger"
 	"github.com/tilt-dev/tilt/pkg/model"
 )
+
+// Metric and label names must match the following rules:
+// https://prometheus.io/docs/concepts/data_model/#metric-names-and-labels
+var KeyImageRef = tag.MustNewKey("image_ref")
+
+var KeyBuildError = tag.MustNewKey("build_error")
+
+// Loosely adapted from how opencensus does HTTP aggregations:
+// https://github.com/census-instrumentation/opencensus-specs/blob/master/stats/HTTP.md#http-stats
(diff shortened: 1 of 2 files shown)


internal/engine/image_builder.go
```diff
@@ -3,16 +3,55 @@ package engine
...
+
+var ImageBuildDurationView = &view.View{
+	Name:        "image_build_duration_dist",
+	Measure:     ImageBuildDuration,
+	Aggregation: ImageBuildDurationDistribution,
+	Description: "Image build time, by image ref",
+	TagKeys:     []tag.Key{KeyImageRef, KeyBuildError},
+}
+
+var ImageBuildCount = &view.View{
```
maiamcc, line 47: can't you just get `count` info by counting the number of histogram entries?

maiamcc (commented):
This looks right enough but I'm not sure I understand the schema--is there a doc/guide/proposal somewhere? (I see that some of this went by in #3749 which I missed, whoops.)

nicks:
This doc is a pretty good overview of how opencensus thinks about aggregations
https://github.com/census-instrumentation/opencensus-specs/blob/master/stats/HTTP.md#http-stats
https://github.com/census-instrumentation/opencensus-specs/blob/master/stats/DataAggregation.md

the PR also has Go-specific links

There are other systems for specifying aggregation schemas (prometheus client, dogstats), but they all seem to be converging towards similar design choices

internal/engine/image_builder.go
nicks: this is a good question! I'm not 100% sure. i found docs that suggested you could compute counts from distributions, but also every example i saw showed them doing both.

My educated guess is that there are certain backends that don't support going from distributions -> counts

internal/engine/image_builder.go
```diff
@@ -43,6 +83,25 @@ func (icb *imageBuilder) CanReuseRef(ctx context.Context, iTarget model.ImageTar
 func (icb *imageBuilder) Build(ctx context.Context, iTarget model.ImageTarget,
 	ps *build.PipelineState) (refs container.TaggedRefs, err error) {
 	userFacingRefName := container.FamiliarString(iTarget.Refs.ConfigurationRef)
+	startTime := time.Now()
+	ctx, err = tag.New(ctx, tag.Upsert(KeyImageRef, userFacingRefName))
+	if err != nil {
+		return container.TaggedRefs{}, err
+	}
+
+	defer func() {
```
maiamcc, line 92: we're probably going to want helper functions for a lot of this stuff soon to make it easy to drop into various functions, but this is good for now

maiamcc (approved):

PR #3812, opened 2020-09-30
hud/web: can attach link_text to endpoints when sending to web ui, endpoints render with link_text if provided [ch9607]

maiamcc:
Hello @nicks, @yu,

here's the frontend work to display endpoints with user-specified link text
(so a user can label their endpoints e.g. "debugger" and "app")

For more info on what this functionality might look like to a user/in
the Tiltfile, see https://github.com/tilt-dev/tilt.build/pull/761

(Nick, if I end up renaming `link_text` as per your comments in the docs PR,
I'll update here)


pkg/webview/view.proto
```diff
@@ -119,7 +124,7 @@ message Resource {
   repeated string pending_build_edits = 9;
   google.protobuf.Timestamp pending_build_since = 10;
   bool has_pending_changes = 11;
-  repeated string endpoints = 12;
+  repeated Link endpoints = 12;
```
nicks, line 127: as a general rule of thumb, never, ever change the type of a protobuf field. treat it like a database schema

in this particular case, this is persisted to a TFT database, so this change would break all existing snapshots. but the reason we have that general rule is so that there's a common understanding of what changes are OK, and you don't have to think too hard about versioning

instead, create a new field with the type you want

maiamcc: whoooop yeah I was just seeing that this broke snapshots. will redo!

maiamcc: is there a convention for how to name the new field vis a vis the old one, or anything?

nicks: i would probably call it `links` (after the type) or `endpoints2` (to communicate that it replaces `endpoints`)

pkg/webview/view.proto
```diff
@@ -119,7 +124,8 @@ message Resource {
   repeated string pending_build_edits = 9;
   google.protobuf.Timestamp pending_build_since = 10;
   bool has_pending_changes = 11;
-  repeated string endpoints = 12;
+  repeated string DEPRECATED_endpoints = 12;
+  repeated Link endpointLinks = 28;
```
nicks, line 128: protobuf fields should be slug cased (endpoint_links)

pkg/webview/view.proto
```diff
@@ -119,7 +124,8 @@ message Resource {
   repeated string pending_build_edits = 9;
   google.protobuf.Timestamp pending_build_since = 10;
   bool has_pending_changes = 11;
-  repeated string endpoints = 12;
+  repeated string DEPRECATED_endpoints = 12;
```
nicks, line 127: let's make this:

```
  reserved "endpoints";
  reserved 12;
```

(there are a couple examples above)

because we use the protobuf-to-json gateway, it's also not safe to change the names of fields

web/src/ResourceInfo.tsx
```diff
@@ -114,19 +116,20 @@ class ResourceInfo extends PureComponent<HUDHeaderProps> {
...
+          <ResourceLink
+            href={ep.url}
             target="_blank"
             rel="noopener noreferrer"
-            key={ep}
+            key={ep.url}
           >
-            {ep}
-          </PortForwardLink>
+            {ep.name ? ep.name : ep.url}
```
nicks, line 131: idiomatically you'd usually do this as

```
ep.name || ep.url
```

pkg/webview/view.proto
maiamcc: whoop ty

web/src/ResourceInfo.tsx
maiamcc: 😳 👍

nicks (approved):

PR #3816, opened 2020-10-01
hud/server: print more request info on test failure

maiamcc:
this came up in another PR, figured I'd add more info for debugging test failures


maiamcc:
Whoops nvm this info is already in logs, I thought it wasn't.

Issue #3792, opened 2020-09-22
api request throttling from other namespaces


nicks:
```
I0920 12:16:47.896538   10105 request.go:621] Throttling request took 1.199331443s, request: GET:https://localhost:36261/apis/batch/v1/namespaces/namespace2/jobs/job-id
I0920 12:16:58.096509   10105 request.go:621] Throttling request took 11.398590634s, request: GET:https://localhost:36261/apis/apps/v1/namespaces/namespace3/replicasets/rs-id
```

I suspect what's going on here is that if you have a big cluster, our code for watching events and determining where they came from is being overzealous about fetching objects for all namespaces, tho i need to verify this

Issue #2846, opened 2020-01-25
Fatal Authorization Error: Tilt Watches Pods and Services In Wrong Namespace


armsnyder:
We use a shared k8s cluster on our team for development, and we've found that users are getting `403 Forbidden` errors in Tilt after running `tilt up`.

>FATAL ERROR
Tilt has encountered a fatal error: Error watching pods. Are you connected to kubernetes? : WatchPods: makeInformer: unknown, Reason: Forbidden, Code: 403
Once you fix this issue you'll need to restart Tilt. In the meantime feel free to browse through the UI.

Our API audit logs show that Tilt is making k8s API requests to services and pods at the cluster-level and at the `default` namespace-level. Developers do not have access to these resources; their access is limited to certain namespaces only.

This is a problem for any cluster with users who have permissions to a restricted set of namespaces, like in clusters managed by Rancher2, for example.

Our Tiltfiles do not include any cluster-level or `default` namespace pods or services, so it is not clear why Tilt is making these requests nor why they should cause a fatal error for users.

**Repro steps:**
1. Create a namespace "foo" in k8s
1. Create a user in k8s
1. Give the user permissions to all verbs and all resources in the "foo" namespace
1. Create a KUBECONFIG file with the user's login credentials
1. Create a minimal Tiltfile which creates a k8s resource
   ```python
   k8s_yaml(blob('''
   apiVersion: v1
   kind: ServiceAccount
   metadata:
     name: foo
     namespace: foo
   '''))
   ```
1. Run `tilt up`
1. See an error in the console
   >FATAL ERROR
Tilt has encountered a fatal error: Error watching pods. Are you connected to kubernetes? : WatchPods: makeInformer: unknown, Reason: Forbidden, Code: 403
Once you fix this issue you'll need to restart Tilt. In the meantime feel free to browse through the UI.

Here are the API audit logs from Tilt:

```json
{
    "requestURI": "/api/v1/pods?labelSelector=app.kubernetes.io%2Fmanaged-by%3Dtilt&watch=true",
    "verb": "watch",
    "userAgent": "tilt/v0.0.0 (darwin/amd64) kubernetes/$Format",
    "objectRef": {
        "resource": "pods",
        "apiVersion": "v1"
    },
    "responseStatus": {
        "metadata": {},
        "status": "Failure",
        "reason": "Forbidden",
        "code": 403
    },
    "requestReceivedTimestamp": "2020-01-25T06:01:14.612754Z"
}
{
[...]

PR #3818, opened 2020-10-01
k8swatch: only watch pods in namespaces we've deployed to. fixes https://github.com/tilt-dev/tilt/issues/2846

nicks:
Hello @landism,

Please review the following commits I made in branch nicks/ch9648:

553494746ff7b15e9a686fb9e4c5db9b7b687155 (2020-10-01 17:20:07 -0400)
k8swatch: only watch pods in namespaces we've deployed to. fixes https://github.com/tilt-dev/tilt/issues/2846
This is a big change!

Before this change, Tilt watched all pods with the "managed-by: tilt" label and
extra pod selector labels. Then it traces owner references (or uses
extra_pod_selectors) to determine which resource they belong to.

After this change, Tilt watches all pods in namespaces you deploy to, and traces
owner references (or uses extra_pod_selectors) to determine which resource they
belong to.

I think that for most users (where they're using a local cluster, or a remote
cluster where each user gets their own namespace), this will be a big
improvement. The watches will be more efficient, get throttled less, and require
less configuration (since there will now be many cases where Tilt doesn't need
extra_pod_selectors).

But I could also imagine edge cases where this will be worse (e.g., a big
namespace with lots of non-Tilt stuff).

45be8d2a04cff5d475b1534fcba50fe8dbd0a992 (2020-10-01 16:19:36 -0400)
k8swatch: only watch events and services in namespaces we've deployed to. fixes https://github.com/tilt-dev/tilt/issues/3792

### DIFF
--- a/internal/engine/k8swatch/event_watch_manager.go
+++ b/internal/engine/k8swatch/event_watch_manager.go
@@ -95,7 +95,7 @@ func (m *EventWatchManager) OnChange(ctx context.Context, st store.RStore) {
 func (m *EventWatchManager) setupWatch(ctx context.Context, st store.RStore, ns k8s.Namespace, tiltStartTime time.Time) {
 	ch, err := m.kClient.WatchEvents(ctx, ns)
 	if err != nil {
-		err = errors.Wrap(err, "Error watching k8s events\n")
+		err = errors.Wrapf(err, "Error watching events. Are you connected to kubernetes?\nTry running `kubectl get events -n %q`", ns)
 		st.Dispatch(store.NewErrorAction(err))
 		return
 	}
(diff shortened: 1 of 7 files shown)


landism:
Still looking at code, but a couple of questions from the description:

> But I could also imagine edge cases where this will be worse (e.g., a big
namespace with lots of non-Tilt stuff).

1. Could we avoid this by watching the intersection of pods with appropriate labels and pods in namespaces Tilt has deployed to? Or is it an effort / complexity thing?
2. Can a CRD deployed to one namespace cause a pod to be created in another? Would this change cause Tilt to miss that pod?

nicks:
re: (1) ya, this is what i originally implemented. It ends up getting very complicated and difficult to reason about. fwiw, the kubernetes ecosystem seems to be moving towards using owner references consistently, so it's possible that our label selectors will go away entirely

re: (2) in theory, yes. in practice, no. We have several examples of users surprised that Tilt is watching namespaces that they're not deploying to. I don't think I've seen other tools do this. We can revisit this with some sort of watch_namespace primitive if it becomes an issue.

landism (approved):
LGTM

PR #3823, opened 2020-10-02
tiltfile: port_forward constructor accepts 'name' param [ch9608]

maiamcc:
Ha, we already had a `port_forwards` constructor method, who knew??

BTW I verified in [Grafana](https://cloud.tilt.dev/grafana/d/V29UlQIMz/tilt-global-built-ins-usage?orgId=1) that ~no one is currently using that func/it's not documented anywhere, so I feel okay about changing the signature.

### DIFF
--- a/internal/tiltfile/k8s.go
+++ b/internal/tiltfile/k8s.go
@@ -852,15 +852,19 @@ func convertPortForwards(val starlark.Value) ([]model.PortForward, error) {
 }
 
 func (s *tiltfileState) portForward(thread *starlark.Thread, fn *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
-	var local int
-	var container int
+	var local, container int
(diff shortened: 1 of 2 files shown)


internal/tiltfile/k8s.go
```diff
@@ -871,6 +875,9 @@ type portForward struct {
 var _ starlark.Value = portForward{}
 
 func (f portForward) String() string {
+	if f.Name != "" {
```
nicks, line 878: it seems really weird to invent a new syntax here, how about:

fmt.Sprintf("port_forward(local_port=%d, container_port=%d, name=%q)", f.LocalPort, f.ContainerPort, f.Name)

nicks (approved):

internal/tiltfile/k8s.go
maiamcc: fairrrr!

PR #3826, opened 2020-10-02
Add j/k keyboard shortcuts

nicks:
Hello @hyu, @maiamcc,

Please review the following commits I made in branch nicks/ch9619:

50328ffffe3e5c2e20fb13f8e5655b80c4404108 (2020-10-02 19:12:44 -0400)
Add j/k keyboard shortcuts

0ecb7ec011c38ec8163c8e12507caa1f2b84b843 (2020-10-02 18:33:55 -0400)
web: split SidebarItem into its own file

### DIFF
--- a/web/src/SidebarKeyboardShortcuts.test.tsx
+++ b/web/src/SidebarKeyboardShortcuts.test.tsx
@@ -1,0 +1,66 @@
+import { MemoryRouter } from "react-router"
+import React from "react"
+import PathBuilder from "./PathBuilder"
+import { createMemoryHistory } from "history"
+import { mount } from "enzyme"
+import { twoResourceView } from "./testdata"
+import SidebarItem from "./SidebarItem"
+import SidebarKeyboardShortcuts from "./SidebarKeyboardShortcuts"
+
+const fakeHistory = createMemoryHistory()
+const pathBuilder = new PathBuilder("localhost", "/")
+const shortcuts = (items: SidebarItem[], selected: string) => {
+  let root = mount(
+    <MemoryRouter initialEntries={["/"]}>
+      <SidebarKeyboardShortcuts
+        items={items}
+        selected={selected}
+        history={fakeHistory}
+        pathBuilder={pathBuilder}
+      />
+    </MemoryRouter>
+  )
+  return root
+    .find(SidebarKeyboardShortcuts)
+    .instance() as SidebarKeyboardShortcuts
+}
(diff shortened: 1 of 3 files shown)


web/src/SidebarKeyboardShortcuts.tsx
```diff
@@ -0,0 +1,58 @@
...
+      case "j":
+      case "k":
+        // An array of sidebar items, plus one at the beginning for 'All'
+        let names = [""].concat(items.map(item => item.name))
+        let index = names.indexOf(selected)
+        let dir = e.key === "j" ? 1 : -1
+        let targetIndex = (index + dir + names.length) % names.length
+        let name = names[targetIndex]
+        let path = name ? pathBuilder.path(`/r/${name}`) : pathBuilder.path("/")
+        history.push(path)
```
maiamcc, line 49: is this in use right now? (presumably it's to let you go "back" to the previously selected resource eventually, right?)

web/src/SidebarKeyboardShortcuts.test.tsx
```diff
@@ -0,0 +1,58 @@
...
+        pathBuilder={pathBuilder}
+      />
+    </MemoryRouter>
+  )
+  return root
+    .find(SidebarKeyboardShortcuts)
+    .instance() as SidebarKeyboardShortcuts
+}
+
+it("navigates forwards", () => {
```
maiamcc, line 28: nice! worth a test that e.g. cmd+J/cmd+K don't have an effect? (i don't feel strongly)

maiamcc (approved):
🙌 ⌨️

hyu:
I really just came here to say that being able to j/k is verrrry satisfying. ☺️

nicks:
@hyu ya, TBH, I'm very surprised how big a dopamine hit this is. possibly because it's VERY fast, which makes Tilt feel faster / more responsive

nicks:
i'm going to change the semantics so that it doesn't wrap, which i think is more of what i expect from other tools

web/src/SidebarKeyboardShortcuts.tsx
nicks: ya, this is how we do all navigations right now. (This is how `Link` element works under the hood, see https://github.com/ReactTraining/react-router/blob/master/packages/react-router-dom/modules/Link.js#L103)

PR #3827, opened 2020-10-03
k8s: add down policy annotation

### LINKED ISSUE #3755: deleting storageclasses on tilt down
A user asked:
> Is there a recommended way to not delete certain helm resources on a tilt down, such as a storageclass?

StorageClass is a cluster-wide resource for configuring a type of storage.

To me, it's similar to a namespace, which we had a similar discussion about:
https://github.com/tilt-dev/tilt/issues/3160
https://github.com/tilt-dev/tilt/issues/2805

Should we treat StorageClasses the same way, and only delete them if you pass a special flag?

I really admire the Helm solution to this:
https://helm.sh/docs/topics/charts_hooks/

where you can add annotations to a resource like:

```
    "helm.sh/hook": post-install
    "helm.sh/hook-weight": "-5"
    "helm.sh/hook-delete-policy": hook-succeeded
```

to specify what helm should do with a resource at different stages of the lifecycle. So you could imagine a Tilt annotation like:

```
  "tilt.dev/down-policy": keep
```

mKeRix:
Inspired by the issue linked at the bottom of this PR I added support for keeping k8s entities based on an annotation. This mirrors the behavior also found in other k8s-adjacent tools like Helm.

For example, you could define a persistent volume to be kept, so that users can `down` everything when done but keep the database contents saved. To achieve the same behavior right now you can use an if based on `tilt_subcommand`, but that doesn't play well with e.g. the `helm_remote` extension.

The annotation is called `tilt.dev/down-policy` and supports the value `keep` (everything else is ignored). YAML usage example:

```yaml
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: mysql-pvc
  annotations:
    tilt.dev/down-policy: keep
spec:
  accessModes:
    - ReadWriteOnce
  resources:
    requests:
      storage: 20Gi
```

Fixes #3755.

### DIFF
--- a/internal/cli/down.go
+++ b/internal/cli/down.go
@@ -82,6 +82,14 @@ func (c *downCmd) down(ctx context.Context, downDeps DownDeps, args []string) er
 		return errors.Wrap(err, "Parsing manifest YAML")
 	}
 
+	entities, _, err = k8s.Filter(entities, func(e k8s.K8sEntity) (b bool, err error) {
+		downPolicy, exists := e.Annotations()["tilt.dev/down-policy"]
+		return !exists || downPolicy != "keep", nil
+	})
+	if err != nil {
+		return errors.Wrap(err, "Filtering entities by down policy")
+	}
+
 	if !c.deleteNamespaces {
 		var namespaces []k8s.K8sEntity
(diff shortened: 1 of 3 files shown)


victorwuky:
@nicks would you be able to review this? Thanks.

nicks (approved):
nice! LGTM

PR #3828, opened 2020-10-03
k8s: reverse entity order when tearing down

mKeRix:
In #3739 the kubectl behavior within Tilt was changed from applying/deleting everything at once to going step by step. This however has some side effects: if a resource that is in use by another is deleted first, `tilt down` will hang and never finish. This is especially annoying for things like a PVC, which requires the deployment that uses it to be deleted first.

## Simple Example for Reproducing

To  reproduce this locally you can try out this simple example, taken and slightly modified from one of our actual Tiltfiles:

**Tiltfile**
```
load('ext://namespace', 'namespace_create', 'namespace_inject')
load('ext://helm_remote', 'helm_remote')

namespace_create('mongo')
helm_remote('mongodb', 'https://charts.bitnami.com/bitnami', 'bitnami', namespace='mongo', version='9.2.1', values="helm/mongo-values.yaml")
k8s_resource('mongodb', port_forwards=27017, objects=['mongodb:persistentvolumeclaim'])
```

**helm/mongo-values.yaml**
```yaml
architecture: standalone
auth:
  enabled: false
```

**.tiltignore**
```
.helm/
```

## Possible Workarounds

In its current state you can circumvent this issue by not adding the PVC to the `objects` of the resource, as this will cause them to be deleted last along with the other uncategorized entities. For us this is not a real solution though, since we would like to use the redeploy button in the UI to reset the DB to a pre-configured state - this requires the PVC to be re-created as well.

## Solution

We can inverse the order for deleting entities to ensure that dependent items are deleted in correct order. Since `tilt up` is also sequential we know that spinning it up in one direction has worked already, so tearing it down will work in the exact opposite direction.

I have verified that this PR works for both scenarios: the PVC included in the resource objects and letting the PVC stay uncategorized.


victorwuky:
@nicks would you please be able to review this? Thanks.

mKeRix:
I just realized that this needs to be applied during redeploys of single resources as well - will add that to the PR later.

internal/cli/down.go
```diff
@@ -82,6 +82,11 @@ func (c *downCmd) down(ctx context.Context, downDeps DownDeps, args []string) er
 		return errors.Wrap(err, "Parsing manifest YAML")
 	}
 
+	// reverse order for tearing down
+	for i, j := 0, len(entities)-1; i < j; i, j = i+1, j-1 {
```
nicks, line 86: I like the spirit of this idea! 

But I don't think this will do what you expect. 

I don't think the order of the manifests is guaranteed to be sorted to begin with, so reversing that order isn't quite right

I think your better bet is reverse-sort all the entities with this function: https://github.com/tilt-dev/tilt/blob/master/internal/k8s/entity.go#L73

mKeRix: That was just the pointer I needed! I updated the PR, it looks much cleaner now imo and should also lead to more stable results. I verified it again with both scenarios as described in the original post.

nicks (approved):

nicks:
thanks @mKeRix !

PR #3833, opened 2020-10-06
tiltfile: can attach links to local resource [ch9609]

maiamcc:
Hello @nicks, @landism,

Please review the following commits I made in branch maiamcc/local-resource-link-tiltfile:

047d4bab2e1f4d9cf8d4fe26a15b5cdfefc30822 (2020-10-06 12:33:42 -0400)
clean up

01f30a2aef18a2dd96a6f22a988d74776b2e2845 (2020-10-05 18:35:42 -0400)
implement, tests green

b3eef229e1c9f3f8b4ddc1ab7df1e7289d43bb92 (2020-10-05 12:56:25 -0400)
red test


internal/tiltfile/links.go
```diff
@@ -0,0 +1,93 @@
...
+
+import (
+	"fmt"
+
+	"go.starlark.net/starlark"
+
+	"github.com/tilt-dev/tilt/pkg/model"
+)
+
+func convertLinks(val starlark.Value) ([]model.Link, error) {
```
nicks, line 11: can you make this implement the Unpack interface? here's a good example
https://github.com/tilt-dev/tilt/blob/master/internal/tiltfile/value/duration.go#L21

internal/tiltfile/links.go
```diff
@@ -0,0 +1,93 @@
...
+				return nil, fmt.Errorf("`link` arg %v includes element %v which must be a string or a link; is a %T", val, v, v)
+			}
+		}
+		return result, nil
+	default:
+		return nil, fmt.Errorf("`link` value must be an string, a link, or a sequence of those; is a %T", val)
+	}
+}
+
+func (s *tiltfileState) link(thread *starlark.Thread, fn *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
```
nicks, line 50: can you put this in an extension? i don't have strong opinions about which extension it is. just would like to keep the tiltfileState kitchen sink from growing

internal/tiltfile/local_resource.go
```diff
@@ -87,6 +90,11 @@ func (s *tiltfileState) localResource(thread *starlark.Thread, fn *starlark.Buil
 		return nil, fmt.Errorf("local_resource must have a cmd and/or a serve_cmd, but both were empty")
 	}
 
+	links, err := convertLinks(linksVal)
```
nicks, line 93: once this is an Unpacker, this will go away (since starlark will be able to correctly identify which argument failed to parse)

internal/tiltfile/links.go
```diff
@@ -0,0 +1,93 @@
...
+
+	return link{
+		model.Link{
+			URL:  url,
+			Name: name,
+		},
+	}, nil
+}
+
+type link struct {
```
nicks, line 67: should we embed a Starlark Struct for this?
https://pkg.go.dev/go.starlark.net/starlarkstruct#Struct

e.g.,

```
type link struct {
  Struct
  
  model.Link
}
```

Struct comes out of the box with a bunch of semantics that make sense for this value, rather than hand-implementing Truth(), Hash(), etc yourself.

maiamcc: ooh neat, I was just cribbing from `portForward` that sgtm

internal/tiltfile/links.go
maiamcc: I mean theoretically yes? but it feels weird to me to hide what is theoretically a first class Tiltfile object's constructor in an extension. I feel like we should be encouraging everyone to uses local_resource.links to name their links for self-documenting-ness, so the less friction we have there, the better. wdyt?

nicks: i think you might be misunderstanding my comment. local_resource should eventually go in an extension too. how the code is organized shouldn't change the syntax.

maiamcc: oh whoops yeah -- I thought you meant [an extension](https://github.com/tilt-dev/tilt-extensions)

internal/tiltfile/links.go
maiamcc: hm do we have an example of this somewhere in the code? (I can't see one.) i'm running into some trouble with pointer receivers and i'm going to leave this for now.

nicks: sgtm, i can poke at it later

maiamcc:
@nicks @landism ready for re-review

nicks (approved):

PR #3834, opened 2020-10-06
tiltfile: add http:// to local resource links if no scheme given

maiamcc:
not even convinced that we want to assume url scheme vs. requiring that the
user give us the fully qualified URL? there also might be more URL validation
we want to do sometime, but this was the easiest/cheapest to do right now.


internal/tiltfile/links.go
```diff
@@ -89,5 +96,28 @@ func (l link) Hash() (uint32, error) {
...
+		// returns an error in very few cases, but it's better than nothing.
+		return "", err
+	}
+
+	if u.Scheme == "" {
+		// If the given URL string doesn't have a scheme, assume it's http
+		u.Scheme = "http"
+	}
+
+	return u.String(), nil
```
nicks, line 122: i wonder if our data model should be a URL rather than a string...

nicks (approved):

PR #3843, opened 2020-10-08
tiltfile: make a link object a struct

nicks:
Hello @maiamcc, @landism,

Please review the following commits I made in branch nicks/struct:

eee98cf3b6efe0603035287c5fdbfb9716958cf6 (2020-10-08 16:37:01 -0400)
tiltfile: make a link object a struct

### DIFF
--- a/internal/tiltfile/links/links.go
+++ b/internal/tiltfile/links/links.go
@@ -6,6 +6,7 @@ import (
 
 	"github.com/pkg/errors"
 	"go.starlark.net/starlark"
+	"go.starlark.net/starlarkstruct"
 
 	"github.com/tilt-dev/tilt/pkg/model"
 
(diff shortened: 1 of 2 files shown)


internal/tiltfile/links/links.go
```diff
@@ -117,6 +101,10 @@ func (e Extension) link(thread *starlark.Thread, fn *starlark.Builtin, args star
 	}
 
 	return Link{
+		Struct: starlarkstruct.FromStringDict(starlarkstruct.Default, starlark.StringDict{
```
landism, line 104:
```suggestion
		Struct: starlarkstruct.FromStringDict(starlark.String("link"), starlark.StringDict{
```

I don't know why the code calls the first arg "constructor", but it looks like it acts as the type. It shows up in `.String()` and is compared when checking if two structs are equal or can be added (apparently structs support addition!?)

internal/tiltfile/links/links.go
```diff
@@ -117,6 +101,10 @@ func (e Extension) link(thread *starlark.Thread, fn *starlark.Builtin, args star
...
+			"url":  starlark.String(withScheme),
+			"name": starlark.String(name),
+		}),
 		Link: model.Link{
```
landism, line 108: It feels a bit awkward to have this data in here twice - maybe it'd be less awkward to remove this field and have `LinkList.Unpack` convert from `links.Link` to `model.Link`?

landism (approved):

internal/tiltfile/links/links.go
nicks: done

internal/tiltfile/links/links.go
nicks: i played around with this and the conversion logic from a generic Struct to a model.Link would be complicated/error-prone (i.e., the struct interface isn't really made for this kind of conversion), so i'm not sure that'd be less awkward

maiamcc (commented):
nice, ty for this!

Issue #3845, opened 2020-10-09
throttling request regression


nicks:
still seeing some issues. From a report:

I1008 20:56:54.054073 3613134 request.go:645] Throttling request took 1.19957263s, request: GET:https://localhost:56463/apis/batch/v1/namespaces/ns/jobs/job1
I1008 20:57:04.254081 3613134 request.go:645] Throttling request took 11.399111973s, request: GET:https://localhost:56463/apis/batch/v1/namespaces/ns/jobs/job2
I1008 20:57:14.254106 3613134 request.go:645] Throttling request took 11.853686772s, request: GET:https://localhost:56463/apis/apps/v1/namespaces/ns/deployments/deployment3

> Took about 60 seconds from Tilt startup to start building (versus 0.17.0 which is 12 seconds)

concerned that the big change between v0.17.0 and now is the upgrade to the kubernetes client...if there was a major regression in the client, we'll need to come up with some repro steps

nicks:
i can reproduce the throttling if i have, say, 100 pods initially in my namespace. we have to fetch all the pod metadata to check if they belong to the deployment we're deploying

one fix might be to turn off rate-limiting for the metadata fetch...the metadata fetch shouldn't be that expensive anyway...but i'm not sure what other side effects this might have

another fix might be to use a metadata informer instead of fetching resources individually https://pkg.go.dev/k8s.io/client-go/metadata/metadatainformer

PR #3851, opened 2020-10-12
k8s: use the metadata client to fetch metadata, instead of full objects

nicks:
Hello @landism, @maiamcc,

Please review the following commits I made in branch nicks/metadata:

f773fe2f12ad9ccbcb8eddf3ba9c1dbf3e3dbb50 (2020-10-12 16:47:16 -0400)
k8s: use the metadata client to fetch metadata, instead of full objects
I was pulling in the metadata client to help with https://github.com/tilt-dev/tilt/issues/3845,
and this seemed like a good optimization to do while I was here

### DIFF
--- a/internal/k8s/client.go
+++ b/internal/k8s/client.go
@@ -24,6 +24,7 @@ import (
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/kubernetes"
 	apiv1 "k8s.io/client-go/kubernetes/typed/core/v1"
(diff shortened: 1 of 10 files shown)


nicks:
see notes on https://pkg.go.dev/k8s.io/client-go/metadata/#Client, seems like this got introduced in Kubernetes 1.14, and the client will emulate pre-1.14 behavior

landism (approved):
nice!

PR #3854, opened 2020-10-13
k8s: pre-populate metadata cache. fixes https://github.com/tilt-dev/tilt/issues/3845

nicks:
Hello @landism,

Please review the following commits I made in branch nicks/ch9742:

4a1579abe2e82a6216a8de43c7c232c5277fe2d0 (2020-10-12 20:01:29 -0400)
k8s: pre-populate metadata cache. fixes https://github.com/tilt-dev/tilt/issues/3845


nicks:
if the PR looks kind of intimidating, it's only really ~100 lines of code and 1,000 lines of tests, most of it very repetitive

internal/k8s/owner_fetcher.go
```diff
@@ -44,20 +47,67 @@ func (t ObjectRefTree) String() string {
...
+	fetch, ok := v.resourceFetches[rns]
+	if !ok {
+		fetch = &sync.Once{}
+		v.resourceFetches[rns] = fetch
+	}
+	return fetch
+}
+
+// As an optimization, we batch fetch all the ObjectMetas of a resource type
+// on startup.
```
landism, line 90: Maybe nitty, but this isn't at startup, is it? IIUC, it's the first time we check the objectmeta of any instance of a type, which I think is, like, the first time event watcher or pod watcher get updates? In either case, this part of the code is sufficiently removed from whether it happens at startup that the comment is hard to verify / keep up to date.

landism (approved):
yay batches

do you think it's also worth getting rid of the last uses of K8sClient.dynamic and replacing it with K8sClient.meta, or has this PR taken care of all the throttling issues?

internal/k8s/owner_fetcher.go
nicks: good point! I reworded this

nicks:
re: dynamic - i think we still need it for cases when we need the full resource data (pods, events, etc)

PR #3857, opened 2020-10-13
tiltfile/model: can attach arbitrary links to k8s resource [ch8238]

maiamcc:
This was originally out of scope for the Endpoints epic, but turns out
one of our customers wants this specifically and it was fast, so!

### DIFF
--- a/internal/store/engine_state.go
+++ b/internal/store/engine_state.go
@@ -614,16 +614,34 @@ func (ms *ManifestState) HasPendingChangesBeforeOrEqual(highWaterMark time.Time)
 var _ model.TargetStatus = &ManifestState{}
 
 func ManifestTargetEndpoints(mt *ManifestTarget) (endpoints []model.Link) {
+	// TODO(maia): we should probably let users define the order of their endpoints;
+	//  at minimum, arbitrary links and port fwd URLs should stay together (currently
+	//  if a k8s resource has both links and port fwds, the whole list of endpoints
+	//  will be alphabetized). However, we don't expect lots of people to specify
+	//  links and port fwds on the same resource, so this is fine for now.
 	defer func() {
 		sort.Sort(model.ByURL(endpoints))
 	}()
 
-	// If the user specified port-forwards in the Tiltfile, we
-	// assume that's what they want to see in the UI
-	portForwards := mt.Manifest.K8sTarget().PortForwards
-	if len(portForwards) > 0 {
-		for _, pf := range portForwards {
-			endpoints = append(endpoints, pf.ToLink())
+	if mt.Manifest.IsK8s() {
+		k8sTarg := mt.Manifest.K8sTarget()
+		endpoints = append(endpoints, k8sTarg.Links...)
+
+		// If the user specified port-forwards in the Tiltfile, we
+		// assume that's what they want to see in the UI (so it
+		// takes precedence over any load balancer URLs
+		portForwards := k8sTarg.PortForwards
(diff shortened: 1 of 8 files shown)


landism:
> one of our customers wants this specifically

just asking since I could see these being easily mixed up, they want arbitrary links, or they want to link to arbitrary paths on the port forward?

I could imagine the most common case being, e.g., my k8s resource runs a web server on port 9090 and I want to link to localhost:9090/login. It seems a lot nicer to be able to specify `port_forwards=[port_forward(port=9090, link=link('login', '/login'))]` or whatever than `port_forwards='9090', links=[link('login', 'http://localhost:9090/login')]`. This might also give control over order in a more useful way.

maiamcc:
lol, in fact, both. for user request for this, [see slack thread](https://tilt-dev.slack.com/archives/C013WM9QH3K/p1602604015010600)

for adding paths to port forwards (https://github.com/tilt-dev/tilt/pull/3858), [request is here](https://tilt-dev.slack.com/archives/CUC1SLYUF/p1600298955006300)

internal/store/engine_state.go
```diff
@@ -614,16 +614,34 @@ func (ms *ManifestState) HasPendingChangesBeforeOrEqual(highWaterMark time.Time)
 var _ model.TargetStatus = &ManifestState{}
 
 func ManifestTargetEndpoints(mt *ManifestTarget) (endpoints []model.Link) {
+	// TODO(maia): we should probably let users define the order of their endpoints;
+	//  at minimum, arbitrary links and port fwd URLs should stay together (currently
+	//  if a k8s resource has both links and port fwds, the whole list of endpoints
+	//  will be alphabetized). However, we don't expect lots of people to specify
+	//  links and port fwds on the same resource, so this is fine for now.
```
landism, line 621: > However, we don't expect lots of people to specify links and port fwds on the same resource, so this is fine for now.

Hmm, I have not talked to any users about this, but I'd expect by far the most common use case to be to point at paths *on* the port forward, and for it to be rare to use links without port forwards.

I suspect a nicer default would be to use this order:
1. arbitrary links, in order specified in the tiltfile
2. port forwards

(1) gives the user mostly total control, and if they really want to move a port forward link up, they can specify it also as an arbitrary link.

If most people are not specifying links and port forwards on the same resource, it seems like it could be annoying of us to take their lovingly ordered links and force them into sorted order.

pkg/model/k8s_target.go
```diff
@@ -61,6 +61,10 @@ type K8sTarget struct {
 	// (right now, Tiltfile and Engine have two different ways of finding a
 	// given image in a k8s entity.
 	refInjectCounts map[string]int
+
+	// (optional) one+ links assoc'd with this resource (to be displayed in UIs,
```
landism, line 65: petty nit
```suggestion
	// zero+ links assoc'd with this resource (to be displayed in UIs,
```

internal/store/engine_state.go
maiamcc: > Hmm, I have not talked to any users about this, but I'd expect by far the most common use case to be to point at paths on the port forward, and for it to be rare to use links without port forwards.

That's a use case for the `path` arg introduced in #3858 -- imho you should use that instead of specifying port forward 8000 and link "localhost:8000/v1/ui". (Tho you're probs right that we'll get ppl doing that anyway.)

I agree that the ordering is dumb tho--let me tackle that in a separate PR?

landism (approved):

PR #3858, opened 2020-10-13
tiltfile/model: can attach a path to a PortForward; endpoint will be linked in UI as '$HOST:$PORT/$PATH'

maiamcc:
again, was originally out of scope in my epic, but was back in the code
and realized this wouldn't take long.

this comes from a customer request:
> we have some services that return 404s at / but serve up things at urls like /v1/ui/


pkg/model/manifest.go
```diff
@@ -491,7 +496,11 @@ func (pf PortForward) ToLink() Link {
...
 		host = "localhost"
 	}
-	url := fmt.Sprintf("http://%s:%d/", host, pf.LocalPort)
+	path := pf.Path
+	if strings.HasPrefix(path, "/") {
+		path = pf.Path[1:]
+	}
+	url := fmt.Sprintf("http://%s:%d/%s", host, pf.LocalPort, path)
 
 	// We panic on error here because we provide the URL format ourselves,
```
landism, line 505: This is no longer true! With this change, a user could put a bad string in their Tiltfile and cause Tilt to panic.

It might be simpler to make `pf.Path` a `URL` and assert `!url.IsAbs` at parse time.

internal/tiltfile/k8s.go
```diff
@@ -861,18 +861,19 @@ func convertPortForwards(val starlark.Value) ([]model.PortForward, error) {
...
-	var name string
+	var name, path string
 
 	// TODO: can specify host (see `stringToPortForward` for host validation logic)
 	if err := s.unpackArgs(fn.Name(), args, kwargs,
 		"local_port", &local,
 		"container_port?", &container,
-		"name?", &name); err != nil {
+		"name?", &name,
+		"path?", &path); err != nil {
```
landism, line 871: Hm, I was surprised we can only have one link, but in a few minutes I haven't come up with a good way to fit multiple links into this API. Bluh.

maiamcc: you mean like there's no good way to tell Tilt to display links to `localhost:800/foo` and `localhost:8000/bar` just with the port forward syntax? Yeah :-/ i think the workaround for that (which I hadn't considered but that you brought up in #3857 is
```
k8s_resource(...,
  port_forwards=port_forward(8000, path='foo'),
  links('localhost:8000/bar')
)
```

hopefully we won't have too many folks wanting that?

landism: hm, reading that, it's not clear what `path` means - what do you think of `link_path`?

landism: > hopefully we won't have too many folks wanting that?

I've definitely wanted this on a lot of services I've worked on. Maybe a good answer is to make non-absolute `links` relative to the port forward.

maiamcc: boooo yah you're right. will do.

pkg/model/manifest.go
maiamcc: ughhhh you're right and i hate it.

are you suggesting IsAbs to check if it has a leading slash or to check that someone didn't pass in a whole-ass url with scheme etc?

internal/tiltfile/k8s.go
maiamcc: FYI i called it `link_path` in the tiltfile and left it as `path` internally but i'm happy for you to tell me that that's dumb

nicks (approved):

PR #3862, opened 2020-10-14
web: add the shortcut dialog to the sidebar

nicks:
Hello @hyu, @maiamcc,

Please review the following commits I made in branch nicks/ch9620/dialog:

f84be98c693dc1691acf63f194d34c6d1eeeec17 (2020-10-14 16:17:00 -0400)
web: add the shortcut dialog to the sidebar


nicks:
Screenshot from 2020-10-14 16-17-28

i also tried to reconcile the differences between the two dialogs

web/src/SidebarAccount.tsx
```diff
@@ -252,30 +186,125 @@ function SidebarAccount(props: SidebarAccountProps) {
...
+      e.preventDefault()
+    }
+  }
+
+  render() {
+    return <span></span>
+  }
+}
+
+function SidebarAccount(props: SidebarAccountProps) {
```
maiamcc, line 227: nbd but might consider renaming "SidebarAccount" as it's no longer just account info. "SidebarInfo"?

web/src/SidebarAccount.tsx
```diff
@@ -252,30 +186,125 @@ function SidebarAccount(props: SidebarAccountProps) {
...
+  const [shortcutsDialogOpen, setShortcutsDialogOpen] = useState(false)
+  const [accountMenuOpen, setAccountMenuOpen] = useState(false)
+
+  let toggleAccountMenu = () => setAccountMenuOpen(!accountMenuOpen)
+  let toggleShortcutsDialog = () => setShortcutsDialogOpen(!shortcutsDialogOpen)
+
+  let optionalLearnMore = null
```
maiamcc, line 234: maybe optionalTiltCloudLearnMore / optionalTCLearnMore? (Again, FFTI)

maiamcc (approved):
nice!

web/src/SidebarAccount.tsx
nicks: the diff is misleading - this isn't new code, this was all here before. i'd be open to a PR to rename things a bit, though not sure we should get too hung up on names right now if this is all changing anyway.

web/src/SidebarAccount.tsx
nicks: ditto

PR #3865, opened 2020-10-15
web: reload resource on (enter)

nicks:
Hello @hyu, @maiamcc,

Please review the following commits I made in branch nicks/ch9618:

fb736e1f84b4d41a776bb366b9773c94fed61ee4 (2020-10-15 15:58:06 -0400)
web: reload resource on (enter)

### DIFF
--- a/web/src/ShortcutsDialog.tsx
+++ b/web/src/ShortcutsDialog.tsx
@@ -40,6 +40,7 @@ export default function ShortcutsDialog(props: props) {
   return (
     <FloatDialog title="Keyboard Shortcuts" {...props}>
       <Shortcut keys="j, k" label="Navigate Resource" />
+      <Shortcut keys="(enter)" label="Trigger rebuild for a resource" />
       <Shortcut keys="Shift+1, 2..." label="Open Endpoint" />
       <Shortcut keys="1" label="View Logs" />
       <Shortcut keys="2" label="View Alerts" />
--- a/web/src/SidebarKeyboardShortcuts.test.tsx
+++ b/web/src/SidebarKeyboardShortcuts.test.tsx
@@ -10,11 +10,14 @@ import { fireEvent } from "@testing-library/dom"
 var fakeHistory: any
 const pathBuilder = new PathBuilder("localhost", "/")
 let component: any
+let triggered: any = false
 const shortcuts = (items: SidebarItem[], selected: string) => {
   let CaptureHistory = () => {
     fakeHistory = useHistory()
     return <span />
   }
+  triggered = false
+
   component = mount(
     <MemoryRouter initialEntries={["/init"]}>
       <CaptureHistory />
@@ -22,6 +25,9 @@ const shortcuts = (items: SidebarItem[], selected: string) => {
         items={items}
         selected={selected}
         pathBuilder={pathBuilder}
+        onTrigger={() => {
+          triggered = true
+        }}
       />
     </MemoryRouter>
   )
@@ -60,4 +66,12 @@ it("navigates backwards no wrap", () => {
   let sks = shortcuts(items, "")
   fireEvent.keyDown(document.body, { key: "k" })
   expect(fakeHistory.location.pathname).toEqual("/init")
+})
+
+it("triggers update", () => {
+  let items = twoResourceView().resources.map(res => new SidebarItem(res))
+  let sks = shortcuts(items, "")
+  expect(triggered).toEqual(false)
+  fireEvent.keyDown(document.body, { key: "Enter" })
+  expect(triggered).toEqual(true)
 })
--- a/web/src/SidebarKeyboardShortcuts.tsx
+++ b/web/src/SidebarKeyboardShortcuts.tsx
@@ -9,6 +9,7 @@ type Props = {
   selected: string
   pathBuilder: PathBuilder
   history: History
+  onTrigger: () => void
 }
 
 /**
@@ -54,6 +55,27 @@ class SidebarKeyboardShortcuts extends Component<Props> {
         history.push(path)
         e.preventDefault()
         break
+
+      case "Enter":
+        for (
+          let current: any = e.target;
+          current;
+          current = current.parentNode
+        ) {
+          let tagName = current.tagName
+          if (
+            tagName === "INPUT" ||
+            tagName === "TEXTAREA" ||
+            tagName === "A" ||
+            tagName === "BUTTON"
+          ) {
+            return
+          }
+        }
+
+        this.props.onTrigger()
+        e.preventDefault()
+        break
     }
   }
 
@@ -66,6 +88,7 @@ type PublicProps = {
   items: SidebarItem[]
   selected: string
   pathBuilder: PathBuilder
+  onTrigger: () => void
 }
 
 export default function(props: PublicProps) {
--- a/web/src/SidebarResources.tsx
+++ b/web/src/SidebarResources.tsx
@@ -20,6 +20,7 @@ import {
   Width,
 } from "./style-helpers"
 import SidebarItem from "./SidebarItem"
+import SidebarKeyboardShortcuts from "./SidebarKeyboardShortcuts"
 
 // Styles:
 const barberpole = keyframes`
@@ -80,13 +81,6 @@ let SidebarItemStyle = styled.li`
     animation: ${barberpole} 8s linear infinite;
   }
 `
-let SidebarItemLink = styled(Link)`
-  display: flex;
-  align-items: stretch;
-  text-decoration: none;
-  // To truncate long names, root element needs an explicit width (i.e., not flex: 1)
-  width: calc(100% - ${Width.sidebarTriggerButton}px);
-`
 
 let SidebarItemAll = styled(SidebarItemStyle)`
   text-transform: uppercase;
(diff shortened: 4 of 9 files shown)


maiamcc:
So I pulled this branch and took it for a spin and noticed some weirdness ([see video](https://drive.google.com/file/d/12vg2iNSxHNAcnT1rGGBB7JjmcM-W_5B7/view?usp=sharing):
* navigating to a normal resource with j/k and hitting enter works as expected, yay
* navigating to Tiltfile or All sends a request to trigger manifest "(tiltfile)" / manifest `null`, which fails. Not the worst thing but maybe we don't even want to send that request?
* navigating to a resource by clicking on it and hitting enter seems to do nothing?
* and then if you navigate AWAY from that resource with j/k and hit enter, it doesn't trigger anything and jumps you back to the resource you had clicked on?

nicks:
Ya, that all sounds like the intended behavior as specified!

I raised this concern with @SurbhiSGupta on https://app.clubhouse.io/windmill/story/9618/trigger-rebuild-shortcut (i.e., because [enter] is already the keyboard shortcut for link navigation in browsers, having it serve as both will lead to some odd/ambiguous behavior), but we're willing to try this out and see how it goes!

maiamcc:
But there are two cases where hitting enter does NOT trigger the selected resource. How is that the specified behavior?

nicks:
ya, in those two cases, the focus is on a link element, so accessibility specs and toolings kind of insist that [enter] needs to execute that link

SurbhiSGupta:
This is behaving a bit different than what I had in mind - hitting j or k selects the resource a or b - the preview on left pane changes accordingly - just as when you select by mouse click - hitting enter should ideally reload that selected resource.

Replacing enter with spacebar is an option but how will it resolve this issue?

SurbhiSGupta:
> ya, in those two cases, the focus is on a link element, so accessibility specs and toolings kind of insist that [enter] needs to execute that link

Please clarify what this means Nick - by hitting j / k user is scrolling through the resource list - why do we need a 'focus' state?

nicks:
ya, hitting j/k and then hitting [enter] should reload

the behavior maia's describing - clicking on it and hitting enter - shouldn't reload, because the a11y spec says it shouldn't (and don't really want to go down the rabbit hole of fighting all the things that are trying to enforce the a11y spec)

how about this -- when you're currently on a resource, clicking that resource in the sidebar ALWAYS reloads? I think that would be consistent with the a11y spec and be closer to the interaction I think Surbhi's looking for.

SurbhiSGupta:
Yes I agree Nick - 

> when you're currently on a resource, clicking that resource in the sidebar ALWAYS reloads

I will additionally speak with Maia to make sure this interaction is intuitive and get back to you.

PS: Copying our conversation from slack here for ref:

(Nick Santos)
_i wonder if we should make it 'r' or (spacebar) or smth_

(Surbhi Gupta)
_This is behaving a bit different than what I had in mind - hitting j or k selects the resource a or b - the preview on left pane changes accordingly - just as when you select by mouse click - hitting enter should ideally reload that selected resource_

(Nick Santos)
_ya, but if you click on a link, then the link has focus, so hitting (enter) means that we have to navigate to that link. here's the specification: https://www.w3.org/TR/wai-aria-practices/#link
we can try to fight it, but I'm not super excited about fighting with the ecosystem of browsers and screenreaders and extensions that try to prevent us from fighting it
w3.orgw3.org
WAI-ARIA Authoring Practices 1.1
This document provides readers with an understanding of how to use WAI-ARIA 1.1 [WAI-ARIA] to create accessible rich internet applications. It describes considerations that might not be evident to most authors from the WAI-ARIA specification alone and recommends approaches to make widgets, navigation, and behaviors accessible using WAI-ARIA roles, states, and properties. This document is directed primarily to Web application developers, but the guidance is also useful for user agent and assistive technology developers._

(Surbhi Gupta)
__I_ agree I would not contradict w3, but I think the discrepancy is in focus vs select mode - why are they both separate?
My understanding is that even in the mouse world, hovering on a resource will not give you logs on that resource - only clicking on a resource will do that.__

nicks:
ok, i updated the PR with the new behavior - clicking a selected resource rebuilds that resource, same as [enter]

web/src/SidebarKeyboardShortcuts.test.tsx
```diff
@@ -61,3 +67,11 @@ it("navigates backwards no wrap", () => {
   fireEvent.keyDown(document.body, { key: "k" })
   expect(fakeHistory.location.pathname).toEqual("/init")
 })
+
+it("triggers update", () => {
+  let items = twoResourceView().resources.map(res => new SidebarItem(res))
+  let sks = shortcuts(items, "")
+  expect(triggered).toEqual(false)
+  fireEvent.keyDown(document.body, { key: "Enter" })
+  expect(triggered).toEqual(true)
```
maiamcc, line 76: does this test assert WHICH element got triggered? that seems important

maiamcc: also, worth a test for the new behavior (that clicking on a selected element triggers update -- and maybe also that clicking on an unselected element selects it but does not trigger update)?

web/src/SidebarResources.tsx
```diff
@@ -129,6 +123,65 @@ let SidebarItemTimeAgo = styled.span`
...
+function SidebarItemLink(props: {
+  title: string
+  onTrigger?: () => void
+  onNavLink: string
+  isSelected: boolean
+  children: any
+}) {
+  if (props.isSelected && props.onTrigger) {
+    return (
+      <SidebarItemLinkButton onClick={props.onTrigger} title={props.title}>
```
maiamcc, line 173: when hovering over a selected sidebar item, i do NOT get the pointing-hand cursor, that seems wrong if they're clickable now?

web/src/SidebarResources.tsx
```diff
@@ -129,6 +123,65 @@ let SidebarItemTimeAgo = styled.span`
 type Resource = Proto.webviewResource
 type Build = Proto.webviewBuildRecord
 
+export const triggerUpdate = (name: string): void => {
+  let url = `/api/trigger`
+
```
maiamcc, line 128: I wonder if it's worth a human-readable error message (even just in the console) if you send a "trigger" command for the tiltfile or for all resources (which ends up being a trigger command for NO resources). (The former is less important cuz I hope to have that implemented sometime next week tho, so 🤷‍♀️ )

maiamcc (approved):
Cool, this behavior seems right now!

web/src/ShortcutsDialog.tsx
```diff
@@ -40,6 +40,7 @@ export default function ShortcutsDialog(props: props) {
   return (
     <FloatDialog title="Keyboard Shortcuts" {...props}>
       <Shortcut keys="j, k" label="Navigate Resource" />
+      <Shortcut keys="(enter)" label="Trigger rebuild for a resource" />
```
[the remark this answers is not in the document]
nicks: the mocks say "a resource", and i'm trying to keep to the mocks unless there's a good reason not to

web/src/SidebarKeyboardShortcuts.tsx
```diff
@@ -54,6 +55,27 @@ class SidebarKeyboardShortcuts extends Component<Props> {
         history.push(path)
         e.preventDefault()
         break
+
+      case "Enter":
+        for (
```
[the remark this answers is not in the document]
nicks: https://developer.mozilla.org/en-US/docs/Web/API/Event/target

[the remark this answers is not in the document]
maiamcc: Ah I was unclear, I was asking how a keyboard event could have a target -- but it's just the thing in focus when the key is hit, right?

web/src/SidebarResources.tsx
nicks: fixed!

web/src/SidebarResources.tsx
nicks: eh, you can already see this in the console via the network logging

hyu (commented):
So, I was discussing this with @SurbhiSGupta earlier today, and following the discussion here on GitHub, and have just tried out this PR. With apologies for adding another cook to this discussion...

I feel that the compromise solution (click selected resource = trigger rebuild) is confusing. And I think it's better to choose another keyboard shortcut (like Surbhi's earlier proposal of `r`) which is less fraught than `enter`. (@nicks, thanks for your good points about accessibility; I was chagrined that I hadn't considered the centrality of `enter` in keyboard navigation earlier in the shortcut discussion. 😬)

Main reason I think this proposed UI is weird: The Trigger Rebuild button is a visually distinct button that is intentionally separate in functionality from its parent, the resource item. In this PR, I have the impression if even if I am not clicking the Trigger Rebuild button, it still triggers a rebuild. 

We're teaching users that clicking on a resource item makes it the active view. To change that behavior on a second click feels "off" for what I would expect from this kind of sidebar navigation UI.

(And I have another point that's more minor and future-proofing, but: we've been talking about a possible version of Tilt where double-clicking on a resource item opens it in a new tab. Not to say that we should indefinitely reserve "double-clicking" for precisely this behavior, but...I think this is an example of a double-click behavior that has more precedent in Sidebar navigation. e.g., VSCode)

nicks:
:tongue: 

i knew this would be contentious - but that's good, since we're getting a lot of signal that "restart" is an operation that people care a lot about. let's talk about this more on monday.

nicks:
going to withdraw this pr and put up a new one

nicks:
https://github.com/tilt-dev/tilt/pull/3869

PR #3876, opened 2020-10-23
web: make resources pinnable

landism:
pin

I broke some refactors into separate commits to make it easier to review.

Pins do not currently persist. That's the next thing to do.


nicks (commented):
can you add a unit test that exercises the pinned resource stuff?

web/src/SidebarResources.tsx
```diff
@@ -154,123 +174,148 @@ type SidebarProps = {
   selected: string
   resourceView: ResourceView
   pathBuilder: PathBuilder
+  initialPinnedItems?: NonNullable<Array<string>>
 }
 
-class SidebarResources extends PureComponent<SidebarProps> {
```
nicks, line 159: hmmmm...why did you remove the pure component part? AFAIR, we did this because we were seeing issues with lots of redundant rerenders and perf problems, are those no longer issues?

landism:
> can you add a unit test that exercises the pinned resource stuff?

done!

web/src/SidebarResources.tsx
landism: Thanks for highlighting this! I was unaware of the history. I converted it to a functional component so that I could use `useContext`, and I erroneously assumed functional components were treated as pure.

I've reverted this change but when I tried to verify that we were still benefitting from PureComponent and that it still worked correctly with ContextProvider, things got complicated. It looks like PureComponent has never done much for us here. I filed https://app.clubhouse.io/windmill/story/9949/web-purecomponent-optimizations-seem-to-not-be-working with what I found.

I've tested this in the storybook and verified the components using the context successfully rerender without their ancestor PureComponent rendering, so I think this will be fine if/when we fix the PureComponent stuff.

nicks (approved):
oops, sorry, didn't see the updates! lgtm

PR #3877, opened 2020-10-23
web: show triggr button for tiltfile [ch9864]

maiamcc:
Hello @nicks, @hyu,

here's the frontend change for triggering tiltfile;
won't land until backend code is in, but had it done
so figured I'd send for review


nicks (approved):
woooo

PR #3878, opened 2020-10-23
engine / server: can trigger update for tiltfile [ch9863]

maiamcc:
Hello @nicks, @landism,

Please review the following commits I made in branch maiamcc/trigger-tiltfile:

aa0e3e876c94fd38e254b934e84f686bf3bc038a (2020-10-23 12:41:35 -0400)
config controller tests

532ac55ed90a266194d623898825c232a55dd3a4 (2020-10-22 17:27:52 -0400)
upper tests

f3c7e8004b3b8fa78a3474d796a3852e49ba515c (2020-10-20 17:55:10 -0400)
server tests

39a730ef5d81f55dc2de8be7a5f577bca51dd700 (2020-10-20 17:39:49 -0400)
engine / server: can trigger update for tiltfile [ch9863]


internal/store/engine_state.go
```diff
@@ -82,6 +82,7 @@ type EngineState struct {
 	WatchSettings model.WatchSettings
 
 	PendingConfigFileChanges map[string]time.Time
+	PendingTiltfileTrigger   TriggerTiltfileAction
```
nicks, line 85: Hmmmm...TiltfileState should have a BuildReason on it already. have you considered using that field?

we've been generally trying to unify these data structures, rather than inventing new ones for Tiltfile

internal/store/actions.go
```diff
@@ -161,3 +161,14 @@ type PanicAction struct {
 }
 
 func (PanicAction) Action() {}
+
+type TriggerTiltfileAction struct {
```
nicks, line 165: ditto here, i'm not totally sure why we need a new action vs using the existing action, given how the codebase has been moving towards unifying these

internal/store/engine_state.go
maiamcc: ah i didn't realize that field existed, ty.

maiamcc:
updated to use existing code as much as possible!

internal/engine/configs/configs_controller.go
```diff
@@ -92,6 +93,10 @@ func (cc *ConfigsController) needsBuild(ctx context.Context, st store.RStore) (b
 		reason = reason.With(model.BuildReasonFlagTiltfileArgs)
 	}
 
+	if state.PendingTiltfileTrigger.After(lastStartTime) {
```
nicks, line 96: this needs to use AfterOrEqual, like other places in the build engine. the test failure on windows are legit; they're happening because the windows system clock isn't monotonic in the way the unix clock is

internal/store/engine_state.go
```diff
@@ -82,6 +82,7 @@ type EngineState struct {
 	WatchSettings model.WatchSettings
 
 	PendingConfigFileChanges map[string]time.Time
+	PendingTiltfileTrigger   time.Time
```
nicks, line 85: can you comment on why the tiltfile needs a trigger time, but other resources don't need a trigger time?

AFAICT, it seems like one trigger system uses a queue to control dupes, and the other trigger system uses time to control dupes. Not sure it's a good idea to use both, because then any future person editing this code has to learn the gotchas/subtleties of both approaches (as my comment above illustrates)

internal/engine/upper.go
```diff
@@ -451,6 +450,14 @@ func handleBuildCompleted(ctx context.Context, engineState *store.EngineState, c
 }
 
 func appendToTriggerQueue(state *store.EngineState, mn model.ManifestName, reason model.BuildReason) {
+	if mn == model.TiltfileManifestName {
+		// the Tiltfile doesn't get queued like other manifests; instead we mark it as
+		// "pending" in the engine state and it gets picked up by the configs controller
+		state.PendingTiltfileTrigger = time.Now()
```
nicks, line 456: also, as a general rule, we shouldn't use time.Now() in reducers. Actions get queued up, and there's nothing really guaranteeing that time.Now() inside the reducer corresponds well to when the trigger happened. 

Here's a doc explaining this - https://redux.js.org/basics/reducers#handling-actions (the distinction comes up in any reactive framework that needs functions to be pure)

internal/store/engine_state.go
maiamcc: I did it this way cuz the mechanisms for tiltfile updates and other manifest updates are so different--like it felt weird to me to use the queue for tiltfile builds when the tiltfile's position in the queue didn't affect ordering at all vis a vis everything else in the queue? but let me take a look and see how awkward it'll be in real life.

maiamcc:
Okay this is ready!

nicks (approved):

internal/engine/configs/configs_controller_test.go
```diff
@@ -73,11 +73,49 @@ func TestConfigsControllerDockerNotConnectedButNotRequired(t *testing.T) {
...
+	f := newCCFixture(t)
+	defer f.TearDown()
+
+	f.addManifest("fe")
+	bar := manifestbuilder.New(f, "bar").WithK8sYAML(testyaml.SanchoYAML).Build()
+	f.setTiltfileHasBuilt()
+
+	state := f.st.LockMutableStateForTesting()
+	state.AppendToTriggerQueue(model.TiltfileManifestName, model.BuildReasonFlagTriggerWeb)
+	state.PendingConfigFileChanges["somefile.txt"] = time.Now()
```
nicks, line 108: If you want this test to work as written, you probably don't want to use time.Now() here, due to the monotonic clocks issue i mentioned earlier. simplest solution is to use `time.Now().Add(- time.Second)` here instead

internal/engine/configs/configs_controller_test.go
```diff
@@ -73,11 +73,49 @@ func TestConfigsControllerDockerNotConnectedButNotRequired(t *testing.T) {
...
+	f := newCCFixture(t)
+	defer f.TearDown()
+
+	f.addManifest("fe")
+	bar := manifestbuilder.New(f, "bar").WithK8sYAML(testyaml.SanchoYAML).Build()
+	f.setTiltfileHasBuilt()
+
+	state := f.st.LockMutableStateForTesting()
+	state.AppendToTriggerQueue(model.TiltfileManifestName, model.BuildReasonFlagTriggerWeb)
+	state.PendingConfigFileChanges["somefile.txt"] = time.Now().Add(time.Second)
```
nicks, line 108: this needs to be negative one second (in the past) not + 1 second (in the future)

nicks: actually nm maybe i'm confused about what this is testing...

maiamcc: yeah this wasn't a great way to write this test but i'm testing that the build reflects trigger AND another reason -- in this case that there were pending config changes. I'm adding a second to the timestamp here to make EXTRA SURE that this pending filechange is timestamped after the last tiltfile load (set in line 104). But let me see if there's not a better way to do this

maiamcc: ...honestly I'm not even sure this + "TestBuildReasonTrigger" are a useful set of tests. will simplify.

PR #3886, opened 2020-10-27
web: persist pinned sidebar items in localStorage

landism:


web/src/SidebarPin.tsx
```diff
@@ -43,9 +49,15 @@ export const sidebarPinContext = React.createContext<SidebarPinContext>({
 export function SidebarPinContextProvider(
   props: PropsWithChildren<{ initialValue?: NonNullable<Array<string>> }>
 ) {
+  let lsc = useContext(localStorageContext)
+
   const [pinnedResources, setPinnedResources] = useState<
     NonNullable<Array<string>>
-  >(props.initialValue ?? [])
+  >(props.initialValue ?? lsc.get<Array<string>>("pinned-resources") ?? [])
```
nicks, line 56: hmmm...i don't understand why we have two sources of truth for the initial pinned resources?

also, i think this should use a lazy initial state, as in 
https://reactjs.org/docs/hooks-reference.html#lazy-initial-state
to avoid unnecessarily reading from the localstore

web/src/LocalStorage.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+// Provides access localStorage, but namespaced by `tiltfileKey`
+// Also handles serialization and typing.
+export function LocalStorageContextProvider(
+  props: React.PropsWithChildren<{ tiltfileKey?: string }>
+) {
+  let context: LocalStorageContext
+
+  if (!props.tiltfileKey) {
+    // if there's no tiltfile key (presumably because we're initializing), just make this all a noop
```
nicks, line 25: it might be be better to error out here.

HUD short-circuits to a load screen if it's initializing https://github.com/tilt-dev/tilt/blob/master/web/src/HUD.tsx#L246 so you don't need to worry about this case, and i'm not even sure this would work correctly

web/src/SidebarPin.tsx
landism: > hmmm...i don't understand why we have two sources of truth for the initial pinned resources?

`props.initialValue` is only set in tests / stories and is used to give tests / stories a way to set the initial pinned items without having to worry about storage location / format. I'll see if I can find a good way to do this via localStorageContext.

> i think this should use a lazy initial state

good call

nicks: ya, i'd be fine with it if this was called initialValueForTesting or smth

web/src/LocalStorage.tsx
landism: or even better, make the arg non-optional!

nicks (approved):

Issue #3433, opened 2020-06-09
include() should be able to load symbols


nicks:
We currently have two mechanisms for loading Tiltfiles

```load()``` - which can load symbols from other Tiltfiles, but requires static strings
```include()``` - which can only execute other Tiltfiles (not load symbols), but allows dynamic strings

Unfortunately, this breaks down if you want to load symbols from a dynamically-determined file, e.g.,

```
for dir in get_list_of_dirs():
  include(dir + '/Tiltfile', 'some_data')
```

which we've seen in a few places. For example, cluster API does something like this for providers, and we've heard another user do this with helm values.

KOGI:
Another upvote for this. Currently trying and failing to do something like this:
`load('{root}/some/child/dir/foobar.tilt'.format(root=config.main_dir), 'my_symbol')`

nicks:
This is now implemented as `load_dynamic`: https://docs.tilt.dev/api.html#api.load_dynamic

PR #3887, opened 2020-10-27
tiltfile: implement load_dynamic.

nicks:
Hello @landism,

Please review the following commits I made in branch nicks/include:

ecb9e44b3c954247483ad57252abd7dadee0479c (2020-10-27 14:54:06 -0400)
tiltfile: implement load_dynamic.
Intended as a replacement for include(), but with a more self-descriptive name
and better semantics for loading symbols.

Fixes https://github.com/tilt-dev/tilt/issues/3433

### DIFF
--- a/internal/tiltfile/include/include.go
+++ b/internal/tiltfile/include/include.go
@@ -11,6 +11,9 @@ import (
 // The main difference is that include() doesn't bind any arguments into the
 // global scope, whereas load() forces you to bind at least one argument into the global
 // scope (i.e., you can't load() a Tilfile for its side-effects).
+//
+// Users should generally be discouraged from using this function. They should use
+// load_dynamic instead, which has a more self-descriptive name and can load symbols.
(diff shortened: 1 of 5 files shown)


internal/tiltfile/include/include.go
```diff
@@ -11,6 +11,9 @@ import (
 // The main difference is that include() doesn't bind any arguments into the
 // global scope, whereas load() forces you to bind at least one argument into the global
 // scope (i.e., you can't load() a Tilfile for its side-effects).
+//
+// Users should generally be discouraged from using this function. They should use
```
landism, line 15: is it worth deprecating `include`?

landism (approved):
neat!

internal/tiltfile/include/include.go
nicks: probably eventually, tho i think it's good practice to leave it around for a little while before formally deprecating

Issue #3883, opened 2020-10-27
Tilt installer doesn't work on Alpine Linux


jpetazzo:
On Alpine, `$OSTYPE` is `linux-musl`, and the installer script only looks for `linux-gnu` and `darwin`, so the install script fails:

```
/ # curl -fsSL https://raw.githubusercontent.com/tilt-dev/tilt/master/scripts/install.sh | bash -x
+ VERSION=0.17.9
++ command -v brew
+ BREW=
+ set -e
+ [[ -z '' ]]
+ install_tilt
+ [[ linux-musl == \l\i\n\u\x\-\g\n\u ]]
+ [[ linux-musl == \d\a\r\w\i\n* ]]
+ set +x
The Tilt installer does not work for your platform: 
Please file an issue at https://github.com/tilt-dev/tilt/issues/new
```

Since the `tilt` binary is currently statically linked, it works fine on Alpine anyway:
```
bash-5.0# file tilt 
tilt: ELF 64-bit LSB executable, x86-64, version 1 (SYSV), statically linked, Go BuildID=nMePJCSaJbBKq9Y44r6d/c-EfWPM0UtRXslRgxayq/-ZifGfxUWg3pEmCeRrf0/Y9a5MW6aMwBnN23Sjrn3, stripped
bash-5.0# ./tilt version
v0.17.9, built 2020-10-19
```

Would it make sense to add a check for `linux-musl` so that the installer works on Alpine?
Or do you foresee that you might want to switch to dynamic linking in the future, which would break Alpine?

Would it make sense to add a link to the manual install process in the output of the install script?

I'm happy to send PRs either way that you'll deem useful.

Thank you! ♥

jpetazzo:
![Why not both?](https://user-images.githubusercontent.com/171481/97374228-25abd780-18b8-11eb-8334-dbd70cbda5de.png)

nicks:
closing this since it sounds like it works now!

PR #3888, opened 2020-10-27
Allow install script to work on Alpine

jpetazzo:
On Alpine Linux, the bash variable `$OSTYPE` is `linux-musl`
instead of `linux-gnu`. But since the `tilt` binary is
statically linked, it will work anyway, so we can let the
script do its thing on Alpine.

“I'd just like to interject for a moment. What you're referring
to as Linux, is in fact, GNU/Linux, or–
— Well, actually, I'm running Alpine Linux on this system, so–
— I think you mean musl/Linux. It's only called Alpine Linux when
  it comes from the Alps region of France.”

See #3883.

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -26,7 +26,7 @@ function copy_binary() {
 }
 
 function install_tilt() {
-  if [[ "$OSTYPE" == "linux-gnu" ]]; then
+  if [[ "$OSTYPE" == "linux"* ]]; then
       if [[ "$BREW" != "" ]]; then
           set -x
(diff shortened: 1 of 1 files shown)


maiamcc (approved):
>“I'd just like to interject for a moment. What you're referring
to as Linux, is in fact, GNU/Linux, or–
> — Well, actually, I'm running Alpine Linux on this system, so–
> — I think you mean musl/Linux. It's only called Alpine Linux when
it comes from the Alps region of France.”

This is art.

PR #3889, opened 2020-10-27
Improve install script error message for unsupported platforms

jpetazzo:
- Correctly display $OSTYPE instead of $OS
- Add link to the documentation page

See #3883.


maiamcc (approved):
chef's kiss carl

PR #3893, opened 2020-10-29
buildcontrol: don't hold back live updates if the container is crashlooping

nicks:
Hello @landism,

Please review the following commits I made in branch nicks/restarts:

08e4f97ac73ff87dbe0a2595c5b9a3c0e1cf3c35 (2020-10-28 23:10:14 -0400)
buildcontrol: don't hold back live updates if the container is crashlooping

### DIFF
--- a/internal/engine/buildcontrol/build_control.go
+++ b/internal/engine/buildcontrol/build_control.go
@@ -407,6 +407,17 @@ func IsLiveUpdateTargetWaitingOnDeploy(state store.EngineState, mt *store.Manife
 			if len(cInfos) != 0 {
 				return false
 			}
+
+			// If the container in this pod is in a crash loop, then don't hold back
+			// updates until the deploy finishes -- this is a pretty good signal
+			// that it might not become healthy.
+			pod := mt.State.K8sRuntimeState().MostRecentPod()
+			for _, c := range pod.Containers {
+				if c.Restarts > 0 {
+					return false
+				}
+			}
+
 		} else if mt.Manifest.IsDC() {
 			cInfos := store.RunningContainersForDC(mt.State.DCRuntimeState())
 			if len(cInfos) != 0 {
--- a/internal/engine/buildcontrol/build_control_test.go
+++ b/internal/engine/buildcontrol/build_control_test.go
@@ -263,6 +263,15 @@ func TestHoldForDeploy(t *testing.T) {
 
 	sancho.State.K8sRuntimeState().Pods["pod-1"] = successPod("pod-1", sanchoImage.Refs.ClusterRef())
 	f.assertNextTargetToBuild("sancho")
+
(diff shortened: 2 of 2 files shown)


internal/engine/buildcontrol/build_control.go
```diff
@@ -407,6 +407,18 @@ func IsLiveUpdateTargetWaitingOnDeploy(state store.EngineState, mt *store.Manife
...
 				return false
 			}
+
+			// If the container in this pod is in a crash loop, Tilt won't be able to
+			// live update. In theory, we could wait around to see if it comes out of
+			// the crash loop and becomes live-updatable again, but we don't want to
+			// hold back the image build on this hope.
+			pod := mt.State.K8sRuntimeState().MostRecentPod()
+			for _, c := range pod.Containers {
+				if c.Restarts > 0 {
```
maiamcc, line 417: these may be quibbles/rare enough cases that it's still worth it to keep the code as-is, but:
1. seems like you're not differentiating "has crashed at least once" from "in crash loop". if container crashed from a runtime error and the container is NOT in a crash loop, we can still probably live update -- is it worth it to try and account for this case?
2. if a container other than the one(s) you're trying to update is crash-looping, should that stop us from updating a healthy container(s)?

nicks: @maiamcc i think you might be misinterpreting what this codepath is doing.

This is codepath is trying to determine if we should pause builds because the pod is still deploying the first time

So for (1), the distinction between a pod that crashed at some point in the past and a pod that's crashing now isn't meaningful, because in both cases the pod has already deployed.

And for (2), you've actually got the 'if' reversed. Returning false here doesn't stop the build, it lets the image build go forward.

landism: FWIW, I was confused by this, too. If the distinction between a crash loop and a previous crash isn't meaningful, then this PR title and code are over-specifying by saying "in a crash loop" instead of "has crashed" - a reasonable reader (or, if not a reasonable reader, at least me) would think it likely the intent was to apply only to a crash loop and wonder if the code not matching intent was a bug.

also re (1), IIUC if the pod crashed at some point in the past but is now healthy, this function will return on line 408 above, because RunningContainers will be non-empty (which is itself a bug - we should be checking the status of the containers running the images being built, rather than just accepting any container being running)

nicks: Yep, exactly...the way the code is written right now, you wouldn't even get here if the container was healthy. If you moved this check to the top of this function, the check would still be correct

landism (approved):

internal/engine/buildcontrol/build_control.go
nicks: added more tests for the other cases, hopefully they make things clearer

PR #3899, opened 2020-11-02
Update issue templates

nicks:


maiamcc (approved):
ha I didn't realize templates were just files. Ty for adding!

PR #3910, opened 2020-11-10
web: analytics track number of pinned resources on startup

landism:
### Problem

Prior to this PR, we only record analytics when a user takes the action of pinning or unpinning a resource. It's possible that most common pinning workflow will be to pin the resources one most cares about and then never touch the pins again. This would mean that in order to evaluate current pin usage, we have to go back through all analytics events looking for when the user pinned items.

### Solution

Also record the number of pinned items when the UI is loaded.

### DIFF
--- a/web/src/SidebarPin.tsx
+++ b/web/src/SidebarPin.tsx
@@ -52,18 +52,21 @@ export function SidebarPinContextProvider(
 ) {
   let lsc = useContext(localStorageContext)
 
-  function getInitialCount(): string[] {
-    let ret = lsc.get<Array<string>>("pinned-resources") ?? []
+  const [pinnedResources, setPinnedResources] = useState<Array<string>>(
+    () =>
+      props.initialValueForTesting ??
+      lsc.get<Array<string>>("pinned-resources") ??
+      []
+  )
+
+  useEffect(() => {
     incr("ui.web.pin", {
-      pinCount: ret.length.toString(),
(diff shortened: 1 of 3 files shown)


web/src/SidebarPin.tsx
```diff
@@ -52,11 +52,17 @@ export function SidebarPinContextProvider(
 ) {
   let lsc = useContext(localStorageContext)
 
+  function getInitialCount(): string[] {
+    let ret = lsc.get<Array<string>>("pinned-resources") ?? []
+    incr("ui.web.pin", {
```
nicks, line 57: i'm not sure you can guarantee this will only be called once? i think it get called whenever a new component is created...maybe a global?

landism: hm, I was unable to get it to be called multiple times, and the [docs](https://reactjs.org/docs/hooks-reference.html#lazy-initial-state) say:
> If the initial state is the result of an expensive computation, you may provide a function instead, which will be executed only on the initial render

I've switched it to a `useEffect` with zero dependencies, which I have greater confidence won't be re-called. I'd like to verify it with a test, but since I haven't been able to come up with any conditions that would have caused multiple calls in the first place, that'd be tricky!

I'm not inclined to spend too much more time digging into this since even if we can't guarantee there's only one call, the damage is almost certainly minimal.

nicks (approved):

PR #3914, opened 2020-11-10
custom_build should allow 0 deps. Also, clean up path unpacking

nicks:
Hello @landism,

Please review the following commits I made in branch nicks/ch10018:

04f72bc58ba682b3bf7c118425d0ccf50997693a (2020-11-10 17:09:28 -0500)
custom_build should allow 0 deps. Also, clean up path unpacking

### DIFF
--- a/internal/tiltfile/value/path.go
+++ b/internal/tiltfile/value/path.go
@@ -1,6 +1,8 @@
 package value
 
 import (
+	"fmt"
+
 	"go.starlark.net/starlark"
 )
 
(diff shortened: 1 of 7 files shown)


internal/tiltfile/value/path.go
```diff
@@ -24,3 +26,53 @@ func (p *LocalPath) Unpack(v starlark.Value) error {
...
+	}
+
+	defer iter.Done()
+
+	values := []string{}
+	var item starlark.Value
+	for iter.Next(&item) {
+		str, err := ValueToAbsPath(p.t, item)
+		if err != nil {
+			return fmt.Errorf("unpacking list item #%d: %v", len(values), err)
```
landism, line 72:
```suggestion
			return fmt.Errorf("unpacking list item at index %d: %v", len(values), err)
```

landism: maybe just me,  but I think "index" is more unambiguously 0-based - without other information, I'd bias towards interpreting "#%d" as 1-based

And/or putting the value of `item.String()` in the error message would make clear which element is problematic. (though that might be better done inside of `ValueToAbsPath`)

internal/tiltfile/value/path.go
```diff
@@ -24,3 +26,53 @@ func (p *LocalPath) Unpack(v starlark.Value) error {
...
+	}
+
+	var iter starlark.Iterator
+	switch x := v.(type) {
+	case *starlark.List:
+		iter = x.Iterate()
+	case starlark.Tuple:
+		iter = x.Iterate()
+	default:
+		return fmt.Errorf("value should be a string or List or Tuple of strings, but is of type %s", v.Type())
```
landism, line 62: "string or" should only be included if `p.AllowsSingleString`

Alternatively, I'd be in favor of deleting that option and always behaving as if it's true. AFAIK the only reason we are inconsistent on whether a single string is allowed is a lack of attention paid to consistency (and the fact that we previously had this logic duplicated everywhere and no centralized place to standardize the behavior). Maybe I'm missing something or maybe changing the existing API is more scope than you want in this PR, though.

landism (approved):

internal/tiltfile/value/path.go
nicks: i agree!

internal/tiltfile/value/path.go
nicks: ok, i buy that argument!

PR #3920, opened 2020-11-12
ci-based release process

nicks:
Hello @nicks,

Please review the following commits I made in branch nicks/release:

cb0773e739972bf73c3c3ba8bec52786b755de2e (2020-11-12 15:50:50 -0500)
ci-based release process

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -150,6 +150,22 @@ jobs:
       - run: ./configs/storybook/deploy.sh
       - slack/notify-on-failure:
           only_for_branches: master
+          
+  release:
+    docker:
+      - image: gcr.io/windmill-public-containers/tilt-releaser
+    steps:
+      - setup_remote_docker:
+          version: 19.03.12
+      - checkout
+      - gcp-cli/install
+      - gcp-cli/initialize
+      - run:
+          name: Setup Docker Credentials
+          command: gcloud auth configure-docker
+      - run: ./scripts/release-ci.sh
+      - slack/status:
+          mentions: "nick"
 
 workflows:
   version: 2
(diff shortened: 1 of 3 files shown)


nicks:
OK! This is a revamp of the release process, based on the release process for ctlptl. Now you just create a tag v0.x.y and push it to git and circleci will run the release for you. GitOps!

nicks:
going to merge this so i can do a release, happy to address comments on a follow-up PR!

.circleci/config.yml
```diff
@@ -150,14 +150,32 @@ jobs:
...
+          version: 19.03.12
+      - checkout
+      - gcp-cli/install
+      - gcp-cli/initialize
+      - run:
+          name: Setup Docker Credentials
+          command: gcloud auth configure-docker
+      - run: ./scripts/release-ci.sh
+      - slack/status:
+          mentions: "nick"
```
maiamcc, line 168: is this still supposed to be here?

maiamcc (commented):
🙌 🥳

.circleci/config.yml
nicks: yep! also happy to mention others

scripts/release-ci.sh
```diff
@@ -0,0 +1,39 @@
...
+echo "$DOCKER_TOKEN" | docker login --username "$DOCKER_USERNAME" --password-stdin
+
+mkdir -p ~/.tilt-dev
+echo "$TILT_CLOUD_TOKEN" > ~/.tilt-dev/token
+
+git fetch --tags
+./scripts/upload-assets.py latest
+goreleaser --rm-dist --skip-publish --snapshot
+
+VERSION=$(git describe --abbrev=0 --tags)
```
landism, line 35: If you want, I think `$CIRCLE_TAG` instead of `latest` on 32 and `$VERSION` elsewhere would be a little more direct and leave less room for things to get weird (like, if a config error/change causes this job to run on a branch instead of a tag, the current code would use the latest tag rather than failing), but I don't think it's a big deal

landism (commented):
woooo!!

PR #3933, opened 2020-11-18
Change Homebrew URLs to https

blaet:
Goreleaser uses https by default [for Homebrew](https://goreleaser.com/customization/homebrew/). 
This is just a change to conform to that standard, and provide a tiny bit of extra security (and anonymity) in the process.


nicks (approved):
LGTM! good catch!

Issue #3943, opened 2020-11-20
applying too many resources takes too long


nicks:
We've had reports that applying lots of configmaps takes a long time (like 60seconds)

Part of the problem is that Tilt converts these into serial arguments to `tilt kubectl`

There's probably a few different ways we can optimize this a lot:
- batching objects of the same type together
- using the kubernetes client directly, so we don't need to re-auth every time

don't think it's hard to optimize this, just haven't had an excuse to do it.

nicks:
i had to back out the change :sob:

nicks:
rolled forward! and fixed in v0.17.13

PR #3945, opened 2020-11-23
k8s: use the Helm kube client instead of shelling out to a separate process to apply changes

nicks:
Hello @landism,

Please review the following commits I made in branch nicks/ch10270:

52dba84ebf72495ae31fde5a848f79aae6b9ecb4 (2020-11-23 16:13:28 -0500)
k8s: use the Helm kube client instead of shelling out to a separate process to apply changes
fixes https://github.com/tilt-dev/tilt/issues/3943

### DIFF
--- a/internal/k8s/client.go
+++ b/internal/k8s/client.go
@@ -9,10 +9,9 @@ import (
 	"strings"
 	"time"
 
-	_ "helm.sh/helm/v3/pkg/kube"
-
 	"github.com/pkg/browser"
 	"github.com/pkg/errors"
+	"helm.sh/helm/v3/pkg/kube"
 	v1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
@@ -22,6 +21,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/version"
 	"k8s.io/apimachinery/pkg/watch"
+	"k8s.io/cli-runtime/pkg/resource"
 	"k8s.io/client-go/discovery"
 	"k8s.io/client-go/discovery/cached/memory"
 	"k8s.io/client-go/dynamic"
@@ -127,8 +127,15 @@ type Client interface {
 }
 
 type RESTMapper interface {
-	RESTMapping(gk schema.GroupKind, versions ...string) (*meta.RESTMapping, error)
+	meta.RESTMapper
 	Reset()
+}
+
+type HelmKubeClient interface {
+	Update(original, target kube.ResourceList, force bool) (*kube.Result, error)
+	Delete(existing kube.ResourceList) (*kube.Result, []error)
+	Create(l kube.ResourceList) (*kube.Result, error)
+	Build(r io.Reader, validate bool) (kube.ResourceList, error)
 }
 
 type K8sClient struct {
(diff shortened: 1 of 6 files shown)


nicks:
I feel like we've been talking about doing this for a while, and finally got the time to do it. Tested with a couple 100K ConfigMaps, and this makes the apply way faster (like from 4-5 seconds to 0.5 seconds)

internal/k8s/client.go
```diff
@@ -314,42 +339,141 @@ func (k K8sClient) Upsert(ctx context.Context, entities []K8sEntity, timeout tim
...
 		if !shouldTryReplace {
-			return nil, errors.Wrapf(err, "kubectl apply:\nstderr: %s", stderr)
+			return nil, err
 		}
 
-		// NOTE(maia): we don't use `kubecutl replace --force`, because we want to ensure that all
+		// NOTE(maia): we don't use `kubectl replace --force`, because we want to ensure that all
 		// dependant pods get deleted rather than orphaned. We WANT these pods to be deleted
 		// and recreated so they have all the new labels, etc. of their controlling k8s entity.
 		logger.Get(ctx).Infof("Applying %s failed. Retrying with 'kubectl delete && create': %s", entity.Name(), reason)
```
landism, line 458:
```suggestion
		logger.Get(ctx).Infof("Updating %s failed. Attempting to delete + recreate': %s", entity.Name(), reason)
```

internal/k8s/client.go
```diff
@@ -314,42 +339,141 @@ func (k K8sClient) Upsert(ctx context.Context, entities []K8sEntity, timeout tim
...
-		// --ignore-not-found because, e.g., if we fell back due to large metadata.annotations, the object might not exist
-		_, stderr, err = k.actOnEntity(ctx, []string{"delete", "--ignore-not-found=true"}, entity)
-		if err != nil {
-			return nil, errors.Wrapf(err, "kubectl delete (as part of delete && create):\nstderr: %s", stderr)
+		if len(existing) > 0 {
```
landism, line 459: Doesn't hurt, but just to check my understanding - IIUC we can only get here if `existing == 0`. Have I missed a case or is this just defensive programming?

internal/k8s/client.go
```diff
@@ -314,42 +339,141 @@ func (k K8sClient) Upsert(ctx context.Context, entities []K8sEntity, timeout tim
...
 // an immutible field error, attempts to `replace --force` it.
 func (k K8sClient) applyEntityAndMaybeForce(ctx context.Context, entity K8sEntity) ([]K8sEntity, error) {
-	stdout, stderr, err := k.actOnEntity(ctx, []string{"apply", "-o", "yaml"}, entity)
+	existing, resources, err := k.prepareUpdate(ctx, []K8sEntity{entity})
 	if err != nil {
-		reason, shouldTryReplace := maybeShouldTryReplaceReason(stderr)
+		return nil, errors.Wrap(err, "kubectl apply")
+	}
 
+	result, err := k.helmKubeClient.Update(existing, resources, false)
```
landism, line 448: bummer that the helm update implementation does have separate errors per object, but then combines them instead of returning an `[]error` like `Delete`

internal/k8s/client.go
landism: or maybe we want to use "kubectl delete", etc, metaphorically?

landism (approved):

internal/k8s/client.go
nicks: mostly defensive -- once we rip out the kubectl runner, i expect this is going to get refactored a lot (since the current code flow doesn't really make sense if there's no kubectl runner)

internal/k8s/client.go
nicks: i knowwwwwwww, maybe we could convince them to accept a patch that returned multiple errors, or a multierror

internal/k8s/client.go
nicks: ya, i went back and forth on this -- i think this speaks to the general debate about how much to hide vs expose kubernetes mechanics and when

PR #3958, opened 2020-12-01
k8swatch: fix a bug where a manifest is removed then re-added, but we never find the pod

nicks:
Hello @maiamcc, @landism,

Please review the following commits I made in branch nicks/watchbug:

828404bab78263eb3b97349db61cc642083a64b7 (2020-12-01 10:27:13 -0500)
k8swatch: fix a bug where a manifest is removed then re-added, but we never find the pod

### DIFF
--- a/internal/engine/k8swatch/watcher.go
+++ b/internal/engine/k8swatch/watcher.go
@@ -86,6 +86,14 @@ func (ks *watcherKnownState) createTaskList(state store.EngineState) watcherTask
 		}
 	}
 
+	// If we're no longer deploying a manifest, delete it from the known deployed UIDs.
+	// This ensures that if it shows up again, we process it correctly.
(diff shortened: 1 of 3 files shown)


internal/engine/k8swatch/watcher.go
```diff
@@ -86,6 +86,12 @@ func (ks *watcherKnownState) createTaskList(state store.EngineState) watcherTask
 		}
 	}
 
+	for uid := range ks.knownDeployedUIDs {
```
maiamcc, line 89: comment with why we're doing this?

maiamcc (approved):
🙌

internal/engine/k8swatch/watcher.go
nicks: done

PR #3976, opened 2020-12-08
web: adjust the size of the all sidebar item

nicks:
Hello @hyu,

Please review the following commits I made in branch nicks/ch10423:

be44abd80b18e1cf7de4ff342bbf6d2b2d66ffbd (2020-12-08 16:00:00 -0500)
web: adjust the size of the all sidebar item

### DIFF
--- a/web/src/SidebarItemView.stories.tsx
+++ b/web/src/SidebarItemView.stories.tsx
@@ -2,7 +2,11 @@ import React from "react"
 import { MemoryRouter } from "react-router"
 import PathBuilder from "./PathBuilder"
 import SidebarItem from "./SidebarItem"
-import { SidebarItemAll, SidebarItemView } from "./SidebarResources"
+import {
+  SidebarItemAll,
+  SidebarItemView,
+  SidebarItemViewProps,
+} from "./SidebarResources"
 import { oneResourceNoAlerts } from "./testdata"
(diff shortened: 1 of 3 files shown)


nicks:
cc @SurbhiSGupta - I also made it monospace instead of sans-serif to match the mocks

Screenshot from 2020-12-08 15-59-38

SurbhiSGupta:
@nicks monospace lgtm

hyu:
@nicks — The height adjustment is good, but I don't see any need to change the font for "ALL" to monospace. 

@SurbhiSGupta, is it intentional to use monospace for ALL instead of the same sans font that we use for the rest of the resources? When I saw that in the mock, I assumed it was just a remnant of the old design. 

We should use the same sans font for all the resource names. To visually explain, here's a side-by-side comparison, below. I think the design on the left makes it more obvious that the "ALL" item is a selectable item, vs the design on the right where it might look a bit like a heading.

Screen Shot 2020-12-11 at 12 21 00 PM

hyu (approved):
Made an argument for sans vs monospace font, but it's not blocking.

nicks:
don't worry, we have a ticket to do a design QA pass post-launch to go through a lot of these little things with Surbhi, rather than stopping to have a big discussion on them at each step of the way. I don't have a strong opinion on this either way and it's easy to change later.

Issue #3882, opened 2020-10-26
Click to copy pod ID in Web UI


maiamcc:
From a user in Slack:
> it would be handy to have a copy button next to the [pod] name at the top of the Tilt UI when a service is selected.

Click-to-copy Pod ID in the web UI is a cheap way we can make it easier for users to interact with/script against their pods (this makes an incremental bit of progress towards what #2060 is trying to achieve).

maiamcc:
There's lots of discussion to be done on how/whether to signal copy-ability with iconography etc., but I think that even without icons, if you click the pod ID and it gets copied and you see a tooltip saying "Pod ID copied!" that would be strictly better than the current behavior. On these grounds I'm marking this "good first issue" (even though there may be more involved design work later). @hyu @SurbhiSGupta lmk if you disagree!

nicks:
fixed in https://github.com/tilt-dev/tilt/pull/3982

PR #3982, opened 2020-12-09
Add copy button next to resource's name in ResourceInfo

salim-runsafe:
This PR addresses #3882 by putting a copy icon next to the resource's name in the ResourceInfo component. When the icon is clicked it copies the resource's name to the clipboard and flips the icon to a checkmark for 1.5 seconds.

There is one known bug where if the user switches to another resource during the time where the checkmark is visible, it will remain for the duration of the 1.5 seconds.


salim-runsafe:
This is how it appears.

image

nicks (approved):
wooo this is awesome! LGTM

PR #3987, opened 2020-12-11
Add "None" (Completed) state for Sidebar Items

hyu:
Screen Shot 2020-12-10 at 7 58 20 PM


web/src/SidebarIcon.tsx
```diff
@@ -75,20 +76,32 @@ let SidebarIconRoot = styled.div`
...
+    }
   }
 `
 
 export default class SidebarIcon extends PureComponent<SidebarIconProps> {
   render() {
     let icon = <span>&nbsp;</span>
     if (this.props.status === ResourceStatus.Warning) {
-      icon = <WarningSvg fill="#fff" height="9px" width="9px" />
+      icon = <WarningSvg fill={Color.white} width="10px" height="10px" />
```
hyu, line 100: FYI: I removed the height and width when we're not adjusting the original size specified in the SVG file.

web/src/style-helpers.ts
```diff
@@ -13,6 +13,7 @@ export enum Color {
 
   offWhite = "#eef1f1",
   grayLightest = "#93a1a1", // Solarized base1 (darkest content tone)
+  grayLighter = "#2D4D55",
```
hyu, line 16: We're woefully limited in grays specified in style-helpers *vs* grays in use in Figma. I've worked around this in the past by using `Color.white` or `Color.black` with alpha transparency, but feels _mildly_ computationally expensive when used in UI elements that can proliferate. Also, this alpha method generates random grays outside our official color scale.

FWIW: I'm irritated at my past self for coming up with a Color naming scheme that isn't very friendly to expansion! Next time I have a chance, I'm going to use a [number-based color scale](https://dev.to/ynab/a-semantic-color-system-the-theory-hk7)!

nicks: ya, after we throw away the old UI i want to go back and align these with the color scale we have in figma

nicks (approved):

PR #3994, opened 2020-12-14
scripts: install script calls brew upgrade if tilt is already installed via brew

maiamcc:
(as always, my bash is slapdash, so lmk what i missed)


maiamcc:
This bit @SurbhiSGupta the other day!

nicks (approved):

PR #4001, opened 2020-12-15
scripts: update the install script to warn if brew tilt conflicts with another tilt

nicks:
Hello @maiamcc,

Please review the following commits I made in branch nicks/issue4000:

6ef0aeca21229a4b2cc1b40f3a5acd9f5c931df9 (2020-12-15 18:01:22 -0500)
scripts: update the install script to warn if brew tilt conflicts with another tilt

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -35,6 +35,16 @@ function brew_install_or_upgrade() {
     # fresh install
     brew install tilt-dev/tap/tilt
   fi
+
+  set +x
+  location=$(command -v tilt)
+  brew_root=$(brew --prefix)
+  if [[ "$location" != "$brew_root"* ]]; then
+      echo "ERROR"
+      echo "Brew installed Tilt as: $brew_root/bin/tilt"
+      echo "But you have another Tilt binary at: $location"
+      exit 1
+  fi
 }
 
 function install_tilt() {
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -39,10 +39,15 @@ function brew_install_or_upgrade() {
   set +x
   location=$(command -v tilt)
   brew_root=$(brew --prefix)
+  brew_tilt="$brew_root/bin/tilt"
   if [[ "$location" != "$brew_root"* ]]; then
-      echo "ERROR"
-      echo "Brew installed Tilt as: $brew_root/bin/tilt"
-      echo "But you have another Tilt binary at: $location"
+      echo "Warning: you have a conflicting binary at: $location"
+      echo "  Brew installed Tilt as: $brew_tilt"
+      echo ""
+      echo "  If you want to use the Tilt you just installed, you can:"
+      echo "  1) Remove the other binary: rm $location"
+      echo "  2) Adjust your PATH to put Brew first: export PATH=\"$brew_root/bin:\$PATH\""
+      echo "  3) Alias tilt: alias tilt=$brew_tilt"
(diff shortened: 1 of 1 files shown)


scripts/install.sh
```diff
@@ -35,6 +35,16 @@ function brew_install_or_upgrade() {
...
     brew install tilt-dev/tap/tilt
   fi
+
+  set +x
+  location=$(command -v tilt)
+  brew_root=$(brew --prefix)
+  if [[ "$location" != "$brew_root"* ]]; then
+      echo "ERROR"
+      echo "Brew installed Tilt as: $brew_root/bin/tilt"
+      echo "But you have another Tilt binary at: $location"
```
nicks, line 45: wondering if it makes sense to suggest they run `rm $location`, or if that would be too aggressive

nicks:
@maiamcc ping?

scripts/install.sh
```diff
@@ -35,6 +35,16 @@ function brew_install_or_upgrade() {
     # fresh install
     brew install tilt-dev/tap/tilt
   fi
+
+  set +x
+  location=$(command -v tilt)
```
maiamcc, line 40: ack wait no, this will complain if they have the ruby gem installed:
```
$ command -v tilt
/Users/maia/.rvm/gems/ruby-2.5.1/bin/tilt
```
(like it might not be a bad idea to warn "you're going to need to alias OUR tilt" or something, but the error message as it is will be misleading at best.)

maiamcc (changes requested):

scripts/install.sh
nicks: hmmmm...can you elaborate on why that's an issue?

Currently this prints:
But you have another Tilt binary at: $location

which seems accurate?

maiamcc: ah i thought this was to prevent you from running an out-of-date version of OUR tilt.

if this is just meant to detect when `$ tilt` will run the wrong thing, the check is fine, but:
* definitely don't suggest rm, b/c it might be a dep of some ruby project somewhere
* don't call it an "ERROR" because this is an expected case
* I know I'd read the current copy as "you have another version of Tilt-by-windmill installed" (people often don't know the "tilt" gem is even a thing), maybe worth clarifying
* add next steps. so maybe:
    ```
    WARNING: you have another Tilt binary at $location
    (This could be an old version of Tilt, the Ruby gem of the same name, etc.)
    To run the latest Tilt, put the following in your shell config:
    ```
and then suggest they put the brew directory in front of their path, or add a tiltdev alias, or whatever.

maiamcc (approved):

scripts/install.sh
nicks: here's what it looks like now:

```
Warning: you have a conflicting binary at: /home/nick/go/bin/tilt
  Brew installed Tilt as: /home/linuxbrew/.linuxbrew/bin/tilt

  If you want to use the Tilt you just installed, you can:
  1) Remove the other binary: rm /home/nick/go/bin/tilt
  2) Adjust your PATH to put Brew first: export PATH="/home/linuxbrew/.linuxbrew/bin:$PATH"
  3) Alias tilt: alias tilt=/home/linuxbrew/.linuxbrew/bin/tilt
```

maiamcc: lgtm!

maiamcc:
👍  ship it

PR #4004, opened 2020-12-16
Sidebar Item Layout and Visual Tweaks

hyu:
- Adjusted spacing between sections and resource items
- Made pin a touch smaller so it doesn't compete with status icons, and so we can reduce sidebar gutter width
- Made it more satisfying to click the pin (subtle scale animation)
- "Favorites" → "Pinned" sidebar section header

Screen Shot 2020-12-15 at 8 19 21 PM


web/src/SidebarPin.tsx
```diff
@@ -6,37 +6,52 @@ import React, {
...
+    transform: scale(1.2);
+  }
+`
+
+let UnpinnedPinIcon = styled(PinResourceFilledSvg)`
+  transition: fill ${AnimDuration.default} linear,
+    transform ${AnimDuration.short} ease, opacity ${AnimDuration.short} linear;
+  opacity: 0;
+
+  ${SidebarItemRoot}:hover & {
```
nicks, line 40: this is not new in this PR, but it feels bad that UnpinnedPinIcon depends on SidebarItemRoot - it feels like a weird dependency to have, and have been thinking about a better way to do this

nicks (approved):

PR #4005, opened 2020-12-16
web: add the overview item popover

nicks:
Hello @hyu,

Please review the following commits I made in branch nicks/overviewitempopover:

bbeee3335a915bd71298eed88c7454a942924593 (2020-12-16 18:07:57 -0500)
web: add the overview item popover

Code review reminders, by giving a LGTM you attest that:

* Commits are adequately tested
* Code is easy to understand and conforms to style guides
* Incomplete code is marked with TODOs
* Code is suitably instrumented with logging and metrics

PR #4006, opened 2020-12-16
web: attach the overview item popover

nicks:
Hello @hyu,

Please review the following commits I made in branch nicks/overviewitempopoverattach:

41cc649000064b2874442abd262c9a84ded4a58a (2020-12-16 18:42:21 -0500)
web: attach the overview item popover

a6b96c6c92d363cda96586fb8e20a00933bd9af9 (2020-12-16 18:10:55 -0500)
web: add the overview item popover

### DIFF
--- a/web/src/OverviewItemView.tsx
+++ b/web/src/OverviewItemView.tsx
@@ -1,3 +1,5 @@
+import Popover from "@material-ui/core/Popover"
+import { makeStyles } from "@material-ui/core/styles"
 import React, { useState } from "react"
 import { Link } from "react-router-dom"
 import TimeAgo from "react-timeago"
@@ -157,7 +159,6 @@ export let OverviewItemBox = styled.div`
 let OverviewItemRuntimeBox = styled.div`
   display: flex;
   align-items: top;
-  border-bottom: 1px solid ${Color.grayLighter};
   transition: border-color ${AnimDuration.default} linear;
 `
 
@@ -177,6 +178,7 @@ let OverviewItemBuildBox = styled.div`
   display: flex;
   align-items: center;
   flex-shrink: 1;
+  border-top: 1px solid ${Color.grayLighter};
 `
 
 let OverviewItemText = styled.div`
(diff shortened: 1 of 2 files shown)


nicks:
Screenshot from 2020-12-16 18-43-04

This builds on https://github.com/tilt-dev/tilt/pull/4005

the interaction isn't exactly what we want, but it's good enough to move forward with for now i think

hyu (approved):
This works! But I had a different transition animation in mind: when you click, the card grows in height but stays anchored in place, and doesn't affect the layout of other cards. 

The benefit: it won't look like there are two instances of the `BuildBox` (one in the default card, and another in the popover). It'll better communicate the physical metaphor that the card is this object that reveals extra details when you click. Also, it'll fix some of the subtle animation issues like the popover not exactly aligning (see screenshot below).

But, I don't mind merging this popover for now since it's still quite functional!

(My rough idea of how to do the above: the BuildBox is absolutely positioned, and anchored relative to the RuntimeBox. The "BuildBox" has the port-forward etc links with height 0 sitting above it. When you click, we add a class that transitions the section to height auto.)

### Subtle alignment issues
Seems like the appearance of the popover removes the scrollbar in Firefox, foiling perfect popover alignment.

nicks:
good questions!

re: the firefox layout issues.  Ya this is...weird. I'm not sure yet what's causing them. I'm pretty sure they're a Firefox bug though - if I click "Inspect Element", the bug goes away (!). And it's not reproducible in the main UI. Some of the scrollbar jankiness is just an artifact of the storybook setup, which we may have to tweak a bit.

re: the transition animation: ya, that would be a valid approach too! 

My rule of thumb for this kind of UI development is to start with the keyboard/mouse interactions I want, try to find an existing UI component that most closely aligns with the semantic interactions, and then go back and hack the visuals to look right. I think that will be easier than starting with the right visuals, then trying to fix up the interaction. But that approach would certainly be doable.

In this case, the Popover component we're using here is pretty good -
https://material-ui.com/api/popover/
it defaults to a Grow transition, but I think we need to come back and replace it with a Collapse transition:
https://material-ui.com/components/transitions/#transitioncomponent-prop

Issue #4016, opened 2020-12-18
tiltfile warnings


nicks:
Many of our users use the Tiltfile to do environment sanity checks -- e.g., if they're on an old version of a dependency.

They've expressed the ability to raise a non-blocking warning/alert that's surfaced more prominently in the UI than the standard print()

PR #4017, opened 2020-12-18
tiltfile: add warn() function. Fixes https://github.com/tilt-dev/tilt/issues/4016

nicks:
Hello @maiamcc,

Please review the following commits I made in branch nicks/warning:

3c7ea53a76d29fd0b5249767aa4f373b19d84903 (2020-12-18 16:04:30 -0500)
tiltfile: add warn() function. Fixes https://github.com/tilt-dev/tilt/issues/4016


internal/tiltfile/print/print.go
```diff
@@ -0,0 +1,56 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	ctx, err := starkit.ContextFromThread(thread)
+	if err != nil {
+		return nil, err
+	}
+
+	logger.Get(ctx).Warnf("%s", msg)
```
maiamcc, line 53: I assume the fact of using a `Warnf` here means that this message will show up as a Warning in the alerts tab in the UI, and is not just another log line, right?

Is it worth verifying that in the tests? (Idk if there's an easy way in `print_test.go`, but there's definitely an existing way to verify warnings in `tiltfile_test.go`.)

maiamcc (approved):

internal/tiltfile/print/print.go
nicks: i changed the test to only look at warning logs

here's how tiltfile_test.go warning assertion works under the hood:
https://github.com/tilt-dev/tilt/blob/master/internal/tiltfile/tiltfile_test.go#L5476
it's not that much different than what's here...

yes, right now this will show up in the alerts tab AND be another log line, though we've been talking about alternative UX in the new overview

PR #4024, opened 2020-12-21
web: resource status bar

nicks:
Hello @hyu,

Please review the following commits I made in branch nicks/ch10494:

14e38dc575f5ecc2df8fa37e29adb4e8688862e8 (2020-12-21 11:06:16 -0500)
web: resource status bar

4004c1ad3da29fb55042de80e9e45b270a74e892 (2020-12-21 11:01:30 -0500)
web: replace react-modal with Popover in dialogs
With react-modal, we had to manually position the dialogs
to anchor them to the trigger buttons.

With the new system, the Popover library handles the anchor.

This will make it much easier to use the dialogs in different UIs


nicks:
Screenshot from 2020-12-21 11-06-04

web/src/OverviewResourceBar.tsx
```diff
@@ -1,7 +1,16 @@
-import React from "react"
+import React, { Component, useRef, useState } from "react"
 import styled from "styled-components"
+import { AccountMenuContent, AccountMenuHeader } from "./AccountMenu"
+import { incr } from "./analytics"
+import { ReactComponent as AccountIcon } from "./assets/svg/account.svg"
+import { ReactComponent as HelpIcon } from "./assets/svg/help.svg"
+import { ReactComponent as StatusSquareIcon } from "./assets/svg/status-square.svg"
```
hyu, line 7: SVG is fine, but for a simple design like this it can also just be a div with fixed width / height / border radius in CSS.

web/src/OverviewResourceBar.tsx
```diff
@@ -46,16 +55,259 @@ let ResourceStatus = styled.div`
   padding: 4px 16px;
 `
 
+type ResourceBarStatusProps = {
+  view: Proto.webviewView
+}
+
+// If there are more than 20 resources,
```
hyu, line 62: Oh! You detailed an edge case that Surbhi and I hadn't actually considered in the design itself. I think of these as 1:1 with the resource list below, so each block always refers to a single specific resource. 

(After playing with this PR, I even had the thought that we could show a tooltip with the name of the resource on hover, so you could quickly see which resources were in an error state from this compact UI.)

I like what you did here for now, since it won't break on contact for users with tons of resources. But I made a note to discuss with Surbhi what the ideal design should be for 20+ resources!

hyu (approved):

web/src/OverviewResourceBar.tsx
nicks: that's fair! changed it to pure html

web/src/OverviewResourceBar.tsx
nicks: :+1:

PR #4027, opened 2020-12-21
Tests MVP: a much less hacky way of specifying tests

maiamcc:
I don't know if we want to merge this into master or just leave it in a mega branch but: here's a tiltfile function `test` that just wraps local resource to create a test resource -- again, just a local resource with an `isTest` flag on it. cc @ellenkorbes

### DIFF
--- a/internal/tiltfile/local_resource.go
+++ b/internal/tiltfile/local_resource.go
@@ -55,8 +55,11 @@ func (s *tiltfileState) localResource(thread *starlark.Thread, fn *starlark.Buil
 		isTest = true
 
 		// TODO: implement timeout
+		//   (Maybe all local resources should accept a timeout, not just tests?)
 	}
 
+	// TODO: using this func for both `local_resource()` and `test()`, but in future
+	//   we should probably unpack args separately
 	if err := s.unpackArgs(fn.Name(), args, kwargs,
 		"name", &name,
 		"cmd?", &updateCmdVal,
@@ -83,7 +86,7 @@ func (s *tiltfileState) localResource(thread *starlark.Thread, fn *starlark.Buil
 	}
 	tags, err := value.SequenceToStringSlice(tagsVal)
 	if err != nil {
-		return nil, errors.Wrapf(err, "%s: tags", fn.Name())
+		return nil, errors.Wrapf(err, "%s: resource_deps", fn.Name())
 	}
 
 	ignores, err := parseValuesToStrings(ignoresVal, "ignore")
--- a/internal/tiltfile/local_resource.go
+++ b/internal/tiltfile/local_resource.go
@@ -28,6 +28,10 @@ type localResource struct {
 	ignores       []string
 	allowParallel bool
 	links         []model.Link
+
+	// for use in testing mvp
+	tags   []string
+	isTest bool
 }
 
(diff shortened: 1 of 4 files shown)


internal/tiltfile/local_resource.go
```diff
@@ -66,6 +81,10 @@ func (s *tiltfileState) localResource(thread *starlark.Thread, fn *starlark.Buil
 	if err != nil {
 		return nil, errors.Wrapf(err, "%s: resource_deps", fn.Name())
 	}
+	tags, err := value.SequenceToStringSlice(tagsVal)
+	if err != nil {
+		return nil, errors.Wrapf(err, "%s: resource_deps", fn.Name())
```
landism, line 86:
```suggestion
		return nil, errors.Wrapf(err, "%s: tags", fn.Name())
```

internal/tiltfile/local_resource.go
```diff
@@ -37,12 +41,22 @@ func (s *tiltfileState) localResource(thread *starlark.Thread, fn *starlark.Buil
...
 
-	var resourceDepsVal starlark.Sequence
+	var resourceDepsVal, tagsVal starlark.Sequence
 	var ignoresVal starlark.Value
 	var allowParallel bool
 	var links links.LinkList
 	autoInit := true
 
+	var isTest bool
+	if fn.Name() == testN {
```
landism, line 51: I'm tempted to suggest separate functions to take care of everything through unpackArgs, and then a common helper function that deals with all that, so that we're not exposing all args to both functions (e.g. `test` shouldn't accept `serve_cmd` and `local_resource` shouldn't accept `tags`). Maybe it's annoying to list all those args out so many times though.

internal/tiltfile/local_resource.go
```diff
@@ -37,12 +41,22 @@ func (s *tiltfileState) localResource(thread *starlark.Thread, fn *starlark.Buil
...
+		// If we're initializing a test, by default parallelism is on
+		allowParallel = true
+
+		isTest = true
+
+		// TODO: implement timeout
```
landism, line 57: I'm in favor of moving this TODO outside of the `if` statement and allowing timeouts on local_resource cmds, if it doesn't make this PR annoying. I was actually wishing I had this last week.

internal/tiltfile/tiltfile_state.go
```diff
@@ -304,6 +304,7 @@ const (
 
 	// local resource functions
 	localResourceN = "local_resource"
+	testN          = "test" // test is just a fork of local resource
```
landism, line 307: maybe move the comment to line 482 and add "controlled, by `fn.Name()`" or so, to help the reader understand the duplicated "localResource" isn't just bad copy/paste

landism (approved):
LGTM as-is but here are some comments if you want them

ellenkorbes:
@maiamcc IMO we can merge this, just keep in mind we'll probably change it soon, and *please* don't delete the other branch (:

internal/tiltfile/local_resource.go
maiamcc: ha yeah I was talking to L about this: "args like `tags` and `timeout` are only relevant to `test()`... though honestly it would be great if you could set them on `local_resource()` as well!"

internal/tiltfile/local_resource.go
maiamcc: wheeeee it's hack o'clock.

yes i agree, if we leave this code in long-term we should change this, but right now #yolo. i'll add a TODO.

PR #4035, opened 2020-12-23
ui: aggregate data about tests in the sidebar [ch10575]

maiamcc:
ugly as sin and doesn't have terribly useful data, but I think this is
enough to hint at the sort of thing we might more usefully display here
when we flesh this feature out.

I _think_ this is safe to land because it only shows up when someone
calls `test` in the Tiltfile, but lmk if we should be more careful
about guarding this from users until it's ready to go.

i didn't worry about styling or getting it to be pretty, responsive, etc. cuz i
don't even know if we'll be using this layout or the new UI layout to show
the mvp to people. We (by we I probably mean @hyu) can go back and
style later if we want to make this look nice in the current sidebar.

Screen Shot 2020-12-23 at 3 58 39 PM

This is what happens when there's not enough vertical space in the sidebar :/
Screen Shot 2020-12-23 at 3 58 46 PM

cc @ellenkorbes


web/src/SidebarItem.tsx
```diff
@@ -42,6 +43,8 @@ class SidebarItem {
 
     this.name = res.name ?? ""
     this.isTiltfile = !!res.isTiltfile
+    this.isTest =
+      (res.localResourceInfo && !!res.localResourceInfo.isTest) || false
```
nicks, line 47: slightly simpler as `!!res.localResourceInfo?.isTest`

nicks (approved):
LGTM, ship it!

web/src/SidebarItem.tsx
maiamcc: nice. ugh i swear i will NEVER understand js.

nicks: oh, this is a new programming language nerd thing! typescript only added it in 2020

https://en.wikipedia.org/wiki/Safe_navigation_operator

i think it started out in Groovy/C# and has taken over a ton of programming languages over the last decade

PR #4051, opened 2021-01-06
web: add an api for filtering build logs and run logs

nicks:
Hello @hyu,

Please review the following commits I made in branch nicks/pairing2:

7a194d2026a4ccf344f4b27c876bca6e5a4221ab (2021-01-06 16:56:04 -0500)
web: add an api for filtering build logs and run logs


nicks:
we should probably factor the fixtures out a bit once we understand what we need, but this seems like a good start

hyu (approved):
Thanks for the pairing on this! ✨

PR #4055, opened 2021-01-07
server: streamline hud server tests

maiamcc:
it'll be helpful with my upcoming PR, and also refactoring/prettifying tests is soothing


landism (approved):
image
🙌

PR #4057, opened 2021-01-07
server: api endpoint for setting trigger mode [ch10576]

maiamcc:
Hello @landism,

Please review the following commits I made in branch maiamcc/override-trigger-mode:

ea43d65b088d8635febe50c88f8ecd958d835c2f (2021-01-07 17:50:02 -0500)
validate trigger mode, double check request in engine

17518058ff8ebd4f0b82c8bf54d3dfa585e9b9d1 (2021-01-07 17:49:58 -0500)
missing return and rename tests

3851de8b81e4e3c8946516c7b96a0fae59b78d03 (2021-01-07 17:49:56 -0500)
more specific implementation

2c7cbfb2507d39c7a870925170301e0a95fbdfa1 (2021-01-07 17:49:52 -0500)
server: first pass at override endpoint

### DIFF
--- a/internal/engine/upper.go
+++ b/internal/engine/upper.go
@@ -182,7 +182,7 @@ func upperReducerFn(ctx context.Context, state *store.EngineState, action store.
 	case metrics.MetricsDashboardAction:
 		handleMetricsDashboardAction(state, action)
 	case server.OverrideTriggerModeAction:
-		handleManifestOverrideAction(state, action)
+		handleOverrideTriggerModeAction(ctx, state, action)
 	default:
 		state.FatalError = fmt.Errorf("unrecognized action: %T", action)
 	}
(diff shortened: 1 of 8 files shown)


maiamcc:
(Note that this is a shitty first pass at this behavior, and in particular, any changes to Trigger Mode made via this endpoint will be wiped away by the next Tiltfile load.)

internal/engine/upper.go
```diff
@@ -881,3 +883,28 @@ func handleMetricsModeAction(state *store.EngineState, action metrics.MetricsMod
...
+	action server.OverrideTriggerModeAction) {
+	// TODO(maia): in this implementation, overrides do NOT persist across Tiltfile loads
+	//   (i.e. the next Tiltfile load will wipe out the override we just put in place).
+	//   If we want to keep this functionality, the next step is to store the set of overrides
+	//   on the engine state, and whenever we load the manifest from the Tiltfile, apply
+	//   any necessary overrides.
+
+	// We validate trigger mode when we receive a request, so this should never happen
+	if !model.ValidTriggerMode(action.TriggerMode) {
+		logger.Get(ctx).Errorf("Error overriding trigger mode: invalid trigger mode %d", action.TriggerMode)
```
landism, line 897: I think it's good policy that "should never happen" error messages explicitly include "internal error" so that if a user *does* hit it, they can skip debugging and go straight to reporting.

internal/engine/upper.go
```diff
@@ -881,3 +883,28 @@ func handleMetricsModeAction(state *store.EngineState, action metrics.MetricsMod
...
+		return
+	}
+
+	for _, mName := range action.ManifestNames {
+		mt, ok := state.ManifestTargets[mName]
+		if !ok {
+			// We validate manifest names when we receive a request, so this should never happen
+			logger.Get(ctx).Errorf("Error overriding trigger mode: no such manifest %q", mName)
```
landism, line 905: ditto

pkg/model/override.go
```diff
@@ -0,0 +1,9 @@
+package model
+
+// Manifest properties specified by the user in the UI; these override any properties
+// specified in the Tiltfile  (but do not persist over multiple runs).
+// A `nil` property indicates that it wasn't specified by the user (to differentiate
+// from the user specifying the zero value).
+type Overrides struct {
```
landism, line 7: should this be in a future PR?

landism (approved):

pkg/model/override.go
maiamcc: 😅 👍

Issue #4041, opened 2021-01-04
sidebar resource status shows as "unknown" before its initial build


landism:
### Steps to Reproduce
I observed this in the Tilt Cloud project as part of normal use, but an easy way to observe it is to make a tiltfile with a bunch of resources:
```
for i in range(1,50):
  local_resource('cmd' + str(i), 'echo ' + str(i))
```
start up Tilt and immediately scroll to the bottom of the sidebar (might require a prod build, or for the dev assets to already have been initialized in your browser)

### Current Behavior
The status of any resources that have not yet run is listed as "Unknown"
![image](https://user-images.githubusercontent.com/7453991/103582032-f1ddc700-4e91-11eb-9999-276361d6dd86.png)

### Expected Behavior
For Tilt to know the status of its resources!
Maybe it should just initialize resource status to "pending"?

#### `tilt doctor` Output
```
$ tilt doctor
Tilt: v0.18.0-dev, built 2020-12-14
System: darwin-amd64
```
(built from a4426386)

#### About Your Use Case
Not urgent or important functionality-wise, but feels like a broken first impression.

Issue #4009, opened 2020-12-17
serve/update status doesn't work right for serve-only resources


nicks:
Create a local resource with a serve cmd but no update cmd, e.g.,

```
local_resource(
  'busybox',
  serve_cmd='busybox httpd -f -p 8080',
  links=['localhost:8080'])
```

The status looks like this:

![Screenshot from 2020-12-17 12-09-06](https://user-images.githubusercontent.com/278641/102519583-ac7f6400-4060-11eb-8923-58fa8d28007c.png)

which doesn't really make sense? It didn't execute anything, why would it take 0.3s?

PR #4058, opened 2021-01-07
web: compute update status on the server side instead of the client side.

nicks:
Hello @maiamcc, @landism,

Please review the following commits I made in branch nicks/ch10662:

c46c3c3fcb81fe414a2bcba662c04239bffb92fa (2021-01-07 18:00:50 -0500)
web: compute update status on the server side instead of the client side.
fixes https://github.com/tilt-dev/tilt/issues/4041
fixes https://github.com/tilt-dev/tilt/issues/4009


internal/store/engine_state.go
```diff
@@ -667,6 +667,32 @@ func (ms *ManifestState) HasPendingChangesBeforeOrEqual(highWaterMark time.Time)
 	return ok, earliest
 }
 
+func (ms *ManifestState) UpdateStatus(triggerMode model.TriggerMode) model.UpdateStatus {
```
nicks, line 670: the whole thrust of this PR is just taking this 'if' branch from the client and moving it to the server, and improving it a bit with stuff the server knows (like whether a local resource has an update command)

web/src/status.tsx
```diff
@@ -1,33 +1,25 @@
...
+  } else if (
+    res.updateStatus == UpdateStatus.NotApplicable ||
+    res.updateStatus == UpdateStatus.None
+  ) {
     return ResourceStatus.None
+  } else if (buildWarnings(res).length > 0) {
+    return ResourceStatus.Warning
+  } else if (res.updateStatus == UpdateStatus.Ok) {
+    return ResourceStatus.Healthy
+  } else if (res.updateStatus == UpdateStatus.Error) {
```
maiamcc, line 19: seems like we should show error above warning, doesn't this want to be above ln 15?

maiamcc (approved):
💯

web/src/status.tsx
nicks: oops, good catch! fixed

PR #4059, opened 2021-01-08
web: clean up design for tabs UI

hyu:
Screen Shot 2021-01-07 at 7 54 54 PM

Interaction:

Note: This PR didn't introduce it, but there's a weird effect where trying to close the only remaining tab doesn't actually remove the tab. (It does take you to Overview, though.)

### DIFF
--- a/web/src/OverviewResourcePane.tsx
+++ b/web/src/OverviewResourcePane.tsx
@@ -25,7 +25,8 @@ let OverviewResourcePaneRoot = styled.div`
 let Main = styled.div`
   display: flex;
   width: 100%;
-  flex: 1 1 auto;
+  // In Safari, flex-basis "auto" squishes OverviewTabBar + OverviewResourceBar
+  flex: 1 1 100%;
   overflow: hidden;
 `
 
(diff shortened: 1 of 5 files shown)


web/src/OverviewResourcePane.tsx
```diff
@@ -25,7 +25,8 @@ let OverviewResourcePaneRoot = styled.div`
 let Main = styled.div`
   display: flex;
   width: 100%;
-  flex: 1 1 auto;
+  // In Safari, flex-basis "auto" squishes OverviewTabBar + OverviewResourceBar
+  flex: 1 1 100%;
```
hyu, line 29: I did this to fix an issue in my version of Safari (kinda outdated...13.1.2). I'm not 100% sure what the bug is, but this appears to fix the layout issue without causing adverse effects on my other browsers.

web/src/OverviewTabBar.tsx
```diff
@@ -2,29 +2,110 @@ import React, { useEffect, useState } from "react"
...
+
+let CloseButton = styled.button`
+  cursor: pointer;
+  background-color: transparent;
+  border: 0 none;
+  padding-left: ${SizeUnit(0.25)};
+  padding-right: ${SizeUnit(0.25)};
+
+  svg {
+    fill: none;
```
nicks, line 102: i'm surprised you used fills, rather than, say, transitioning the opacity, but this is ok too

nicks (approved):

Issue #3844, opened 2020-10-09
upgrade warning when current version is newer than team version


nicks:
![image (4)](https://user-images.githubusercontent.com/278641/95585963-f79c4b80-0a0d-11eb-8f41-d8a957859e35.png)

see attached image. this doesn't look right!

PR #4066, opened 2021-01-11
web: add an update dialog. Also fixes https://github.com/tilt-dev/tilt/issues/3844 while I'm here

nicks:
Hello @hyu,

Please review the following commits I made in branch nicks/ch10670:

04f3b55ca482f3b6d71e22136e05f845efbe460a (2021-01-11 16:30:20 -0500)
web: add an update dialog. Also fixes https://github.com/tilt-dev/tilt/issues/3844 while I'm here


nicks:
Screenshot from 2021-01-11 16-29-34

here's what it looks like now

i know we want to change the styling on the Dialog, but it would be a lot easier to punt on that until after we've deleted the old UX (since it's a pretty big change to the core components)

hyu (approved):

Issue #3795, opened 2020-09-22
tilt should prefer a local registry over an in-cluster builder


nicks:
Currently, if a cluster offers BOTH a local registry and an in-cluster builder, Tilt will prefer the in-cluster registry

We should change this to prefer the local registry

This is not urgent. You have to very deliberately set minikube up this way to fall into this case (our example scripts encourage you to use containerd, which does not have this problem)

PR #4069, opened 2021-01-12
docker: if local registry is present, prefer that to in-cluster docker

maiamcc:
addresses #3795

Also added names to all the test cases to make them easier to find/reason about, but lmk if i'm misrepresenting what any of the existing tests are asserting

### DIFF
--- a/internal/build/test_utils.go
+++ b/internal/build/test_utils.go
@@ -50,7 +50,7 @@ func newDockerBuildFixture(t testing.TB) *dockerBuildFixture {
 	ctx, _, _ := testutils.CtxAndAnalyticsForTest()
 	env := k8s.EnvGKE
 
-	dEnv := docker.ProvideClusterEnv(ctx, env, wmcontainer.RuntimeDocker, k8s.FakeMinikube{}, wmcontainer.Registry{})
+	dEnv := docker.ProvideClusterEnv(ctx, env, wmcontainer.RuntimeDocker, k8s.FakeMinikube{})
 	dCli := docker.NewDockerClient(ctx, docker.Env(dEnv))
 	_, ok := dCli.(*docker.Cli)
 	// If it wasn't an actual Docker client, it's an exploding client
--- a/internal/build/test_utils.go
+++ b/internal/build/test_utils.go
@@ -50,7 +50,7 @@ func newDockerBuildFixture(t testing.TB) *dockerBuildFixture {
 	ctx, _, _ := testutils.CtxAndAnalyticsForTest()
 	env := k8s.EnvGKE
 
-	dEnv := docker.ProvideClusterEnv(ctx, env, wmcontainer.RuntimeDocker, k8s.FakeMinikube{})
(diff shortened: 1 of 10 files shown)


nicks:
Hmmm....from a data modelling perspective, it seems weird for the existence of a registry inside of a cluster to change the cluster Docker connection.

I don't think this is going to work right in the case where the user has Minikube installed but doesn't have Docker installed? (I've talked to a few users who do this, especially on Linux)

Have you considered an approach were you switch to the local Docker connection, like we do for Docker Compose? e.g., https://github.com/tilt-dev/tilt/blob/master/internal/docker/switch.go#L37

maiamcc:
> Hmmm....from a data modelling perspective, it seems weird for the existence of a registry inside of a cluster to change the cluster Docker connection.

sure, happy to redo this--i was imagining it more like "not bothering to collect information on the cluster docker connection" but that's a fair point

> I don't think this is going to work right in the case where the user has Minikube installed but doesn't have a local Docker installed? (I've talked to a few users who do this, especially on Linux)

huh i didn't know there were cases where the local client was invalid. In that case, even IF there's a local registry configured*, we'd want to use the cluster client, right? (though they necessarily wouldn't get the benefit of this local registry with docker, b/c we're using the cluster client.) So presumably in `switchCli`, we'd check, what,
```
func (c *switchCli) client() Client {
	c.mu.Lock()
	defer c.mu.Unlock()
	if c.orc == model.OrchestratorK8s && (!c.localRegExists || c.localCli.Invalid()) {
		return c.clusterCli
	}
	return c.localCli
}
```
?

do you know off the top of your head what the LocalClient looks like if you don't have docker installed/how I would check that it's valid?

nicks:
you'll get an exploding client and you can use CheckConnected()
https://github.com/tilt-dev/tilt/blob/master/internal/docker/exploding.go#L29

maiamcc:
1. minikube with local registry:
    ```
    Tilt: v0.18.3-dev, built 2021-01-19
    System: darwin-amd64
    ---
    Docker
    - Host: [default]
    - Version: 1.41
    - Builder: 2
    ---
    Kubernetes
    - Env: minikube
    ...
    ```
2. minikube without local registry:
    ```
    Tilt: v0.18.3-dev, built 2021-01-19
    System: darwin-amd64
    ---
    Docker (cluster)
    - Host: tcp://192.168.64.7:2376
    - Version: 1.41
    - Builder: 2
    ---
    Docker (local)
    - Host: [default]
    - Version: 1.41
    - Builder: 2
    ---
    Kubernetes
    - Env: minikube
    ...
    ```

I expect this to work but I'm not easily able to start up minikube+local registry w/o docker (i assume i'd need to use the hyperkit/virtualbox drivers, and then manually put on the "here's the local registry" annotations or configmap or whatever in the cluster?) -- @nicks is this easy behavior for you to verify?

nicks:
oof, i tested it. doesn't work. it's using the registry in some codepaths and the in-cluster builder in other codepaths, then exploding.

I think you also need to update all the places that use UsesLocalDockerRegistry() with similar logic.

maiamcc:
Closing this because it's mega stale, but if anyone wants to pick it up ever, my blessings on you 🙇‍♀️

PR #4072, opened 2021-01-13
tiltfile: add env support for local + local_resource

### LINKED ISSUE #3767: [feature request] Support setting environment in `local()` and `local_resource()`
I've got to execute some local binary with complex environment which should be generated dynamically.

At the moment I do something like this:
```python
def with_env(cmd, env):
    variables = " ".join(["{k}='{v}'".format(k=k, v=v) for (k, v) in env.items()])
    return 'env {variables} sh -c "set -x; {cmd}"'.format(variables=variables, cmd=cmd)

env_common = read_yaml('./env-common.yaml')
env_secret = read_yaml('./env.yaml')
env_combined = dict(env_common, **env_secret)

cmd_raw = "printenv" # this is just for test purposes, I'm using a proper binary in my case
cmd = with_env(cmd_raw, env_combined)
local_resource('my_cmd', cmd)
```

What I would like to do is to avoid this monstrosity and do the following:
```python
env_common = read_yaml('./env-common.yaml')
env_secret = read_yaml('./env.yaml')

cmd = "printenv" # this is just for test purposes, I'm using a proper binary in my case
local_resource('my_cmd', cmd, env=dict(env_common, **env_secret))
```

milas:
New, optional argument(s) for `local` and `local_resource` to pass
environment variables to the invoked command. In all cases, the
argument is a `dict` and must have both string keys and values:
no type conversion will take place.

For `local`, there is a single `env` argument.
For `local_resource`, there is `env`, which is used for the update
command (`cmd`) and `serve_env`, which is used for the serve command
(`serve_cmd`).

Any environment variables that are passed will take precedence over
the parent process' (i.e. `tilt`) environment. However, it's possible
to read from the environment in `Tiltfile`, so in the event that a
default is desired, a practical approach is to combine with `os.getenv`:

```python
local_resource('mycmd', './mycmd', env={'KEY': os.getenv('KEY', 'default')})
```

Closes #3767.

### DIFF
--- a/internal/engine/local/execer.go
+++ b/internal/engine/local/execer.go
@@ -222,4 +222,3 @@ func (e *processExecer) killProcess(ctx context.Context, c *exec.Cmd, processExi
 		return
 	}
 }
-
--- a/internal/engine/local/execer.go
+++ b/internal/engine/local/execer.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"fmt"
 	"io"
-	"os"
 	"os/exec"
 	"sync"
 	"syscall"
@@ -13,6 +12,7 @@ import (
 
 	"github.com/stretchr/testify/require"
 
+	"github.com/tilt-dev/tilt/internal/localexec"
 	"github.com/tilt-dev/tilt/pkg/logger"
 	"github.com/tilt-dev/tilt/pkg/model"
 	"github.com/tilt-dev/tilt/pkg/procutil"
@@ -145,7 +145,7 @@ func (e *processExecer) processRun(ctx context.Context, cmd model.Cmd, w io.Writ
 	defer close(statusCh)
 
 	logger.Get(ctx).Infof("Running serve cmd: %s", cmd.String())
-	c := ExecCmd(cmd, logger.Get(ctx))
+	c := localexec.ExecCmd(cmd, logger.Get(ctx))
 
 	c.SysProcAttr = &syscall.SysProcAttr{}
 	procutil.SetOptNewProcessGroup(c.SysProcAttr)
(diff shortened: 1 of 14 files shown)


internal/tiltfile/files.go
```diff
@@ -10,11 +10,13 @@ import (
 	"path/filepath"
 	"strings"
 
+	"github.com/tilt-dev/tilt/internal/engine/local"
```
nicks, line 13: our dependency graph is going to get weird if the Tiltfile execution controller depends on the Local Resource execution controller

Maybe we need a third package for these kinds of command APIs that both packages can depend on? or just put the new ExecLocalCmd code in pkg/model?

internal/tiltfile/local_resource.go
```diff
@@ -63,12 +64,14 @@ func (s *tiltfileState) localResource(thread *starlark.Thread, fn *starlark.Buil
 	if err := s.unpackArgs(fn.Name(), args, kwargs,
 		"name", &name,
 		"cmd?", &updateCmdVal,
+		"env?", &updateEnv,
```
nicks, line 67: i think these need to get added to the end of the arg list, so that these don't break people using positional args)

the starlark folks have been receptive to having functions that only accept named args (not positional args), but that's for a future compatibility-breaking release! https://github.com/google/starlark-go/issues/249

nicks (commented):
nice!

internal/tiltfile/files.go
milas: Absolutely, I was conflicted about where to put this and meant to revisit. I think a new package is probably warranted - I'm hesitant to add to `pkg/model` since it's an internal construct for execution and shouldn't really be part of public API surface

milas: Moved to `internal/localexec` in cdeef139b0b444b7088f9da26caa6a7dac9257e7

internal/tiltfile/local_resource.go
milas: D'oh - you're absolutely right; moved them to where they should be in f20723045a7eeb7e590529ce2ae887882946fbb5

internal/engine/local/execer.go
```diff
@@ -223,3 +222,34 @@ func (e *processExecer) killProcess(ctx context.Context, c *exec.Cmd, processExi
...
+	c := exec.CommandContext(ctx, cmd.Argv[0], cmd.Argv[1:]...)
+	populateExecCmd(c, cmd, logger.Get(ctx))
+	return c
+}
+
+func populateExecCmd(c *exec.Cmd, cmd model.Cmd, l logger.Logger) {
+	c.Dir = cmd.Dir
+	// env from command definition takes precedence over parent env (exec.Cmd takes last in case of dupes)
+	execEnv := os.Environ()
+	execEnv = append(execEnv, cmd.Env...)
```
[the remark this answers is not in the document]
milas: Agreed - swapped it around in 98f6b8b157b28c6942af9dc58c542a9d190f646e

nicks (approved):
LGTM!

PR #4076, opened 2021-01-13
ui: tests appear in own section of overview

maiamcc:
if this is the right way to do this, i'll add a test and re-request review

totally stumped on good names for the sections besides "non-test resources" and "tests" -- because tests ARE resources. Unless we want to change what that word means? sigh. thoughts welcome.


maiamcc:
Screen Shot 2021-01-13 at 3 56 35 PM

nicks (approved):

PR #4079, opened 2021-01-14
web: render correctly when the resource is not found

nicks:
Hello @hyu,

Please review the following commits I made in branch nicks/notfound:

ede2b63a7116ab778fae2cafd848d00afda50304 (2021-01-14 12:30:32 -0500)
web: render correctly when the resource is not found

### DIFF
--- a/web/src/OverviewPane.tsx
+++ b/web/src/OverviewPane.tsx
@@ -70,7 +70,7 @@ export default function OverviewPane(props: OverviewPaneProps) {
   ) : null
   return (
     <OverviewPaneRoot>
-      <OverviewTabBar />
+      <OverviewTabBar selectedTab={""} />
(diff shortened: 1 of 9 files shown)


nicks:
Screenshot from 2021-01-14 12-26-59

hyu (approved):
Definitely an improvement to how we handle "resource not found," but as I'm looking at this more closely, it's weird to see a persisting tab for a non-existent resource. Can't think of a better UX off the top of my head, though! Will mull on this.

(Somewhat contrived) repro: 
- Update URL to change resource name


Screen Shot 2021-01-14 at 7 25 37 PM

nicks:
ahhhh good point! I changed it so that OverviewTab takes an explicit selectedTab parameter, rather than inferring it from the url. which seems like the right design anyway.

PR #4083, opened 2021-01-15
ui: tests for test section of overview pane

maiamcc:
Hello @nicks, @hyu,

Please review the following commits I made in branch maiamcc/test-section-tests:

5c6b5eda71a15ca774cf2445b201cdaa0b03b3b0 (2021-01-15 12:29:33 -0500)
ui: tests for test section of overview pane

### DIFF
--- a/web/src/OverviewPane.tsx
+++ b/web/src/OverviewPane.tsx
@@ -40,31 +40,40 @@ let ServicesContainer = styled.div`
   overflow: auto;
 `
 
-function PinnedServicesDivider() {
-  return (
-    <ServicesDividerRoot>
-      <GridDividerPinSvg style={{ marginLeft: "28px" }} />
-      <ServicesLabel>Pinned Resources</ServicesLabel>
-    </ServicesDividerRoot>
-  )
+function PinnedResources(items: OverviewItem[]) {
+  return items?.length ? (
+    <div className={"resources-container pinned"}>
+      <ServicesDividerRoot className={"ServicesDivider pinned"}>
+        <GridDividerPinSvg style={{ marginLeft: "28px" }} />
+        <ServicesLabel>Pinned Resources</ServicesLabel>
+      </ServicesDividerRoot>
+      <OverviewGrid items={items} />
+    </div>
+  ) : null
 }
 
-function AllResourcesDivider() {
-  return (
-    <ServicesDividerRoot>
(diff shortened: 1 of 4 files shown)


maiamcc:
Oh wait let me make the error reporting a little better

maiamcc:
...never mind, there's no clean way to do this, this level of verbosity in test output is fine--a user can go back through the stack trace to find where things went wrong if they need to :-/

web/src/OverviewPane.tsx
```diff
@@ -40,31 +40,40 @@ let ServicesContainer = styled.div`
...
 `
 
-function PinnedServicesDivider() {
-  return (
-    <ServicesDividerRoot>
-      <GridDividerPinSvg style={{ marginLeft: "28px" }} />
-      <ServicesLabel>Pinned Resources</ServicesLabel>
-    </ServicesDividerRoot>
-  )
+function PinnedResources(items: OverviewItem[]) {
```
nicks, line 43: can you make these React components that return React.Fragments?

web/src/OverviewPane.tsx
```diff
@@ -40,31 +40,40 @@ let ServicesContainer = styled.div`
...
+  return items?.length ? (
+    <div className={"resources-container pinned"}>
```
nicks, line 45: Let's get rid of all the CSS class names in this PR. None of them are used for styling. Once this is a React.Fragment, you should be able to use the react component itself as a selector.

maiamcc:
thanks for the tip -- this should be better!

web/src/OverviewPane.tsx
```diff
@@ -40,31 +40,44 @@ let ServicesContainer = styled.div`
...
-  return (
-    <ServicesDividerRoot>
-      <GridDividerAllSvg style={{ marginLeft: "28px" }} />
-      <ServicesLabel>All Resources</ServicesLabel>
-    </ServicesDividerRoot>
-  )
+export function PinnedResources(props: ResourceProps) {
+  return props.items?.length ? (
+    <React.Fragment>
+      <ServicesDividerRoot className={"ServicesDivider pinned"}>
```
nicks, line 50: you should be able to get rid of these classes too now?

nicks (approved):

web/src/OverviewPane.tsx
maiamcc: whoops good catch

PR #4084, opened 2021-01-15
engine: add initial touch points for local readiness probe

milas:
Currently a noop in that the readiness probe status is simply
initialized to successful and then never touched. In the future,
this will continue to be the behavior for local resources without
a readiness probe defined, while local resources with a readiness
probe will start off in a failing state until the process is up &
probe passes. (This aligns with the behavior of readiness probes
in K8s.)

### DIFF
--- a/internal/store/runtime_state.go
+++ b/internal/store/runtime_state.go
@@ -39,16 +39,31 @@ type LocalRuntimeState struct {
 	HasSucceededAtLeastOnce bool
 	PID                     int
 	SpanID                  model.LogSpanID
+
+	readinessProbeSuccessful bool
 }
 
 func (LocalRuntimeState) RuntimeState() {}
 
 var _ RuntimeState = LocalRuntimeState{}
 
+func NewLocalRuntimeState(m model.Manifest) LocalRuntimeState {
+	// TODO(milas): make initial readiness probe state conditional based on existence of probe
+	state := LocalRuntimeState{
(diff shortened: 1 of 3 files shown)


internal/store/runtime_state.go
```diff
@@ -39,16 +39,31 @@ type LocalRuntimeState struct {
 	HasSucceededAtLeastOnce bool
 	PID                     int
 	SpanID                  model.LogSpanID
+
+	readinessProbeSuccessful bool
```
milas, line 43: Internally within K8s, probes actually return a `Result` (see [kubernetes/pkg/probe/probe.go](https://github.com/kubernetes/kubernetes/blob/a9e2e18cc6737072fef2bce838bcf508a8e708f3/pkg/probe/probe.go#L19-L31)).

Valid results are `Success` / `Warning` / `Failure` / `Unknown`. `Warning` is logically the same as `Success` but means there's some debugging stuff -- it's only used in K8s for HTTP probe when there's an excessive number of redirects triggered. `Unknown` isn't actually used - things w/o probes are set to `Success` immediately and those w/ probes are initialized to `Failure` until probe runs + succeeds.

I think it's reasonable to simplify to a bool in our case. (The actual `Prober` iface can still return a result that gets coerced into a bool.)

nicks (approved):

nicks:
you might need to sync and re-push (there was a weird merge at HEAD that broke js tests)

PR #4087, opened 2021-01-15
web: incremental line rendering

nicks:
Hello @maiamcc, @hyu,

Please review the following commits I made in branch nicks/ch10779:

6d58e787fb576756bbe31262eec89f74ec6bec9a (2021-01-15 17:14:27 -0500)
web: incremental line rendering
navigating to a pane with a lot of logs was still feeling a bit sluggish

This performs well even when you get up to 1,000,000+ lines,
so should hold up for a while.

### DIFF
--- a/web/src/OverviewLogPane.tsx
+++ b/web/src/OverviewLogPane.tsx
@@ -72,6 +72,7 @@ function newLineEl(
     classes.push("is-buildEvent-fallback")
   }
   let span = document.createElement("span")
+  span.setAttribute("data-sl-index", String(line.storedLineIndex))
   span.classList.add(...classes)
 
   if (showManifestPrefix) {
@@ -100,6 +101,46 @@ function newLineEl(
   return span
 }
 
+// An index of lines such that lets us find:
+// - The next line
+// - The previous line
+// - The line by stored line index.
+type LineHashListEntry = {
+  prev?: LineHashListEntry | null
+  next?: LineHashListEntry | null
+  line: LogLine
+  el?: Element
+}
+
+let CursorEntry = {} as LineHashListEntry
+
+class LineHashList {
+  private last: LineHashListEntry | null = null
+  private byStoredLineIndex: { [key: number]: LineHashListEntry } = {}
+
+  lookup(line: LogLine): LineHashListEntry | null {
+    return this.byStoredLineIndex[line.storedLineIndex]
+  }
+
+  append(line: LogLine) {
+    let existing = this.byStoredLineIndex[line.storedLineIndex]
+    if (existing) {
+      existing.line = line
+    } else {
+      let last = this.last
+      let newEntry = { prev: last, line: line }
+      this.byStoredLineIndex[line.storedLineIndex] = newEntry
+      if (last) {
+        last.next = newEntry
+      }
+      this.last = newEntry
+    }
+  }
+}
+
+// The number of lines to render at a time.
+export const renderWindow = 250
+
 // React is not a great system for rendering logs.
 // React has to build a virtual DOM, diffs the virtual DOM, and does
 // spot updates of the actual DOM.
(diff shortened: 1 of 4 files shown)


web/src/OverviewLogPane.tsx
```diff
@@ -320,27 +372,150 @@ export class OverviewLogComponent extends Component<OverviewLogComponentProps> {
...
+      })
+    } else {
+      // If this is the first render, put the lines in the backward buffer, so
+      // that the last lines get rendered first.
+      lines.forEach((line) => {
+        this.backwardBuffer.push(line)
+      })
+    }
+
+    if (this.renderBufferRafId) return
```
maiamcc, line 392: what's going on here? can you explain at minimum to me, but maybe also via comment for the next reader?

web/src/OverviewLogPane.tsx
```diff
@@ -320,27 +372,150 @@ export class OverviewLogComponent extends Component<OverviewLogComponentProps> {
...
+    this.renderBufferRafId = this.props.raf.requestAnimationFrame(
+      this.renderBuffer
+    )
+  }
+
+  // We have two render buffers:
+  // - a buffer of newer logs that we haven't rendered yet.
+  // - a buffer of older logs that we haven't rendered yet.
+  // First, process the newer logs.
+  // If we're out of new logs to render, go back through the old logs.
```
maiamcc, line 402: understanding check: a single call of this function renders at most n (where n = renderWindow) lines of forward buffer and n lines of backward buffer, and then calls itself recursively if there's more stuff to render?

web/src/OverviewLogPane.tsx
```diff
@@ -320,27 +372,150 @@ export class OverviewLogComponent extends Component<OverviewLogComponentProps> {
...
+
+    // Render the lines in the forward buffer first.
+    let forwardLines = this.forwardBuffer.slice(0, renderWindow)
+    this.forwardBuffer = this.forwardBuffer.slice(renderWindow)
+    for (let i = 0; i < forwardLines.length; i++) {
+      let line = forwardLines[i]
+      let rendered = this.renderLineHelper(line)
+      if (!rendered) {
+        // This must be updating the progress of something that hasn't been
+        // rendered yet.
```
maiamcc, line 425: so we're dropping this line in the backward buffer, despite it still being a "new" log line that needs rendering, on the assumption that by the time we get around to processing the backward buffer, we may have rendered the line that this line updates?

web/src/OverviewLogPane.tsx
```diff
@@ -320,27 +372,150 @@ export class OverviewLogComponent extends Component<OverviewLogComponentProps> {
...
+
+    let root = this.rootRef.current
+    let existingLineEl = entry.el
+    if (existingLineEl) {
+      root.replaceChild(lineEl, existingLineEl)
+      entry.el = lineEl
+      return true
+    }
+
+    // If the only child is the cursor, then this is the first line we're trying
```
maiamcc, line 497: "first" as in... the first log line of the entire tilt run? the thing we want to put on the screen first time-wise, meaning the bottom-most line? the thing we we want to put at the very top of the screen?

maiamcc (approved):
I thiiink this looks good -- had trouble wrapping my head around some bits of logic, left comments where that happened because a. i'm curious but also b. you may want to put comments there for the next person to read this code, if my comprehension/not is representative.

web/src/OverviewLogPane.tsx
nicks: Schedule a render job if there's not one already scheduled.

web/src/OverviewLogPane.tsx
nicks: yep! it's yielding the CPU and then scheduling another render job

web/src/OverviewLogPane.tsx
nicks: ya

web/src/OverviewLogPane.tsx
nicks: Added this comment:

    // Our rendering algorithm inserts the lines incrementally by placing
    // each line next to an existing line.
    //
    // In the base case (where no lines have been rendered yet), there's only
    // one possible choice (before the cursor).

there are cases where it won't be the bottom-most line! But it doesn't really matter, because there's only one place to insert the line.

nicks: ooh, i thought of a way to do this that has fewer edge cases, so i deleted this bit

PR #4091, opened 2021-01-19
tilt: bump kustomize, handle patches in kustomization.yaml

### LINKED ISSUE #4081: Kustomize fails via builtin but works via local/shell
### Expected Behavior

If `kustomize build .` works, then using `kustomize('.')` should also work.

### Current Behavior

My `kustomization.yaml` contains `patches:` with a path to a file. When I include `kustomize('.')` and run `tilt up` I get an error:

```
  /home/my/development/tilt-issue/Tiltfile:2:27: in <toplevel>
  <builtin>: in kustomize
Error: resolving deps: yaml: unmarshal errors:
  line 5: cannot unmarshal !!map into patch.StrategicMerge
```

I looked at Tilt's code and I cannot fault it but I cannot reproduce this error in the shell either.

I confirmed Tilt calls `kustomize` rather than `kubectl kustomize` because that gives a different error due to kubectl embedding kustomize v2.

### Steps to Reproduce

1. Given these files:

    <details><summary><code>kustomization.yaml</code></summary>

    ```yml
    resources:
    - deployment.yaml

    patches:
    - path: hostAliases.yaml
        target:
        kind: Deployment
    ```
    </details>

    <details><summary><code>deployment.yaml</code></summary>

    ```yml
    apiVersion: apps/v1
    kind: Deployment
    metadata:
    name: http
    spec:
    selector:
        matchLabels:
        app: http
    template:
        metadata:
        labels:
            app: http
        spec:
        containers:
            - name: server
            image: nginx:1.18.0
    ```
    </details>

    <details><summary><code>hostAliases.yaml</code></summary>

    ```yml
    apiVersion: apps/v1
    kind: Deployment
    metadata:
    name: .*
    spec:
    template:
        spec:
        hostAliases:
            - ip: "172.17.0.1"
            hostnames:
                - "something.outside.minikube"
    ```
    </details>

    <details open><summary><code>Tiltfile</code></summary>

    ```python
    kustomize_yaml = kustomize('.')
    k8s_yaml(kustomize_yaml)
    ```
    </details>
2. Run `tilt up` to see the error.
3. It will, however, work if you call kustomize via `local`:

    ```python
    kustomize_yaml = local('kustomize build .')
    k8s_yaml(kustomize_yaml)
    ```

#### `tilt doctor` Output

```
$ tilt doctor
Tilt: v0.18.4, built 2021-01-13
System: linux-amd64
---
Docker (cluster)
- Host: tcp://192.168.49.2:2376
- Version: 1.40
- Builder: 2
---
Docker (local)
- Host: [default]
- Version: 1.40
- Builder: 2
---
Kubernetes
- Env: minikube
- Context: minikube
- Cluster Name: minikube
- Namespace: default
- Container Runtime: docker
- Version: v1.18.3
---

---
Analytics Settings
- User Mode: opt-out
- Machine: 1fd472e961fcce630f3368d17e45955e
- Repo: RDaN2EqCh6NObMupaA2w+A==
```

#### `kustomize` version
```
$ kustomize version
{Version:kustomize/v3.9.1 GitCommit:7439f1809e5ccd4677ed52be7f98f2ad75122a93 BuildDate:2020-12-29T19:25:35Z GoOs:linux GoArch:amd64}
```

#### About Your Use Case

This was originally spotted by my colleague and then I replicated this on my machine. Both are running Ubuntu and minikube. In this case we wanted to patch some `hostAliases` so we could call services running locally outside of minikube, but applying patches this way has many use cases. Also, it's the only way to apply something to every Deployment (or matching a regex).

So while there is a workaround available, I just don't understand why is this happening.

landism:
fixes #4081

### Problem

When using the `kustomize` Tiltfile builtin with a kustomization.yaml that uses [patches](https://kubectl.docs.kubernetes.io/references/kustomize/patches/), Tilt spits out
```
Error: resolving deps: yaml: unmarshal errors:
  line 5: cannot unmarshal !!map into patch.StrategicMerge
```

### Solution

campbellr pretty much fully diagnosed this in the bug's thread - "patches" existed with one meaning, then was deprecated, then was re-introduced. Tilt's kustomize code needs to be bumped to allow the "patches" field.

(also, not part of the observed behavior, but patch files should be added as deps)


web/src/testdata.tsx
```diff
@@ -273,36 +273,6 @@ function oneResourceUnrecognizedError(): Resource {
   return resource
 }
 
-function oneResourceTest(): Resource {
```
nicks, line 276: it looks like this is a bad merge?

landism: yep, thanks

nicks (approved):

nicks (approved):

nicks:
the kustomize vendoring situation is such a mess; I've filed a few bugs about it but it seems like a political morass

PR #4097, opened 2021-01-20
web: implement tab nav semantics

nicks:
Hello @landism, @hyu,

Please review the following commits I made in branch nicks/ch10596:

4bd61e65e76d70a1337dff62d23a6e7ff76e68e3 (2021-01-20 13:32:32 -0500)
web: implement tab nav semantics

4318645843a0f0000416bb4eca4ebef9ec7ddac0 (2021-01-19 20:38:47 -0500)
web: pull up all the Tab logic into a separate React context
We're moving towards a model where clicking on the Sidebar
can have complex interactions with the Tab bar, so we need
a better way to centralize state between them.

This PR doesn't change any of the semantics. It just moves state around.


nicks:
(once all the refactoring in the previous PR is done, changing the semantics themselves gets much simpler)

web/src/TabNav.tsx
```diff
@@ -139,56 +143,83 @@ export function OverviewNavProvider(
       return
     }
 
-    setTabs((prevState) => {
-      if (prevState.includes(name)) {
-        return prevState
-      }
+    if (tabs.includes(name) && name == selectedTab) {
```
landism, line 146: my understanding is that it's safer to call `setState` with `(prevState: S) => S` rather than just `S` (and this is what the old code does)

The new code is using the closed value of `tabs` from the start of the function, which might always have the value we want, but I'm not certain, so I've tended to err on the side of updating the previous value at the time of the callback invocation.

Also, this: https://medium.com/ableneo/react-setstate-does-not-immediately-update-the-state-84dbd26f67d5

Maybe this is all fine as-is and will always do what we want. Maybe at some point we end up with an annoying-to-diagnose bug as a result from it, or maybe that's impossible. I haven't really spent time trying to convince myself it's safe to rely on the closed value and would be happy to learn it's fine.

landism: we can also just leave this as-is and trust that if it is actually a problem, bug reports will be soon enough that we'll remember this as a potential issue

landism (approved):

landism:
image
sorry about that, typo'd the wrong issue into my PR, and apparently it's not possible to delete these cross-references

nicks:
lol, no worries

hyu (approved):

web/src/TabNav.tsx
nicks: (left a comment about this)

PR #4101, opened 2021-01-21
web: split SidebarPinButton into a standalone component

nicks:
Hello @hyu,

Please review the following commits I made in branch nicks/pinbutton:

30af9cad1dfa1f899e9bc828cb07e657e7121069 (2021-01-21 14:10:08 -0500)
web: split SidebarPinButton into a standalone component


web/src/SidebarItemView.tsx
```diff
@@ -265,7 +273,7 @@ export default function SidebarItemView(props: SidebarItemViewProps) {
   return (
     <SidebarItemRoot
       key={item.name}
-      className={`${isSelectedClass} ${isBuildingClass}`}
+      className={`u-showPinOnHover ${isSelectedClass} ${isBuildingClass}`}
```
nicks, line 276: I think the big change worth talking about is this line. 

We need a way to propagate hover behavior from an ancestor to its descendent - i.e., "if the mouse hovers on the ancestor, do something with the descendant.

In the old convention, the only way to do this would be to create a circular dependency between the ancestor and the descendant component.

This PR tries to make the component more reusable, by adding a magic "u-showPinOnHover" class name.

I could imagine many other conventions to achieve the same effect, but think there's value in agreeing to a common convention for this.

what do you think?

hyu: I think this is a good solution. 

I like the idea of using this convention sparingly for situations just like this, to break a circular dependency.

(In support of this convention: I see that there's only a handful of other `.u-` prefix classes in the codebase, which aren't really used widely. Of the 3 we have — `.u-hide`, `.u-flexRow`, `.u-flexColumn` — only one is in use, for the old Snapshot modal. We don't actually have a CSS system with composable classes, so I like the idea of reserving these ".u-" utility classes only for the use case above. )

hyu (approved):

web/src/SidebarItemView.tsx
nicks: the u- convention comes from this old post! http://nicolasgallagher.com/about-html-semantics-front-end-architecture/

i'm not totally using it correctly here, but i'm hoping using some convention (any convention) will discourage people from adding new css classes unthoughtfully (since in this codebase, we're only using CSS classes for a very narrow set of reasons)

PR #4108, opened 2021-01-22
web: infer log filters from the url

nicks:
Hello @hyu,

Please review the following commits I made in branch nicks/filters:

99c25d98c45d019d29f0361f272e888f4281eb05 (2021-01-22 15:47:34 -0500)
web: infer log filters from the url


web/src/filters.ts
```diff
@@ -0,0 +1,63 @@
+// Types and parsing logic for log filters.
```
hyu, line 1: Can this file be logfilters.ts? I'm working on sidebar resource type filters, so the disambiguation helps.

hyu (approved):

web/src/filters.ts
nicks: changed to logfilters.ts!

PR #4111, opened 2021-01-22
web: add buttons for warnings and errors

nicks:
Hello @hyu,

Please review the following commits I made in branch nicks/ch10829:

a684ac249148e4db422745ff40fb6db3f741f767 (2021-01-22 17:35:34 -0500)
web: add buttons for warnings and errors

4055d566875b21ee1e150b098cdd9b0b11a83f82 (2021-01-22 16:09:07 -0500)
web: infer log filters from the url

### DIFF
--- a/web/src/OverviewActionBar.stories.tsx
+++ b/web/src/OverviewActionBar.stories.tsx
@@ -1,5 +1,7 @@
+import { createMemoryHistory } from "history"
 import React from "react"
-import { MemoryRouter } from "react-router"
+import { Router } from "react-router"
+import { FilterLevel, FilterSource, useFilterSet } from "./logfilters"
 import OverviewActionBar from "./OverviewActionBar"
 import { SidebarPinMemoryProvider } from "./SidebarPin"
 import { oneResource } from "./testdata"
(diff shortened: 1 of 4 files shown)


nicks:
here's what it looks like in context:

Screenshot from 2021-01-22 17-39-24
Screenshot from 2021-01-22 17-39-07

i think it needs some additional controls (e.g., the ability to navigate to the error in context), but this is a good start

hyu:
These buttons for log level need to co-exist with our UI for setting Filter Source. What do you think about using a segmented control for log levels, since one of these three options must be selected at any given time? (Vs Log Source, which has two options that are independent toggles.)

(I also updated this design in Figma)

Screen Shot 2021-01-22 at 6 41 32 PM


That said, this is so exciting to filter by log level!

hyu (approved):

nicks:
oh ho! They're going to become pills, but they're going to be pills on the other axis!

I borrowed some ideas from chrome devtools (which is solving a similar problem - logs at different levels from different sources). There's going to be a dropdown that lists the different sources and gives a bit more hierarchy (for example, I want to display the number of errors in the button itself). stay tuned!

hyu:
Ahhh, I understand now! Sorry, I didn't fully understand the design you were going after. I see what you mean, now!

PR #4115, opened 2021-01-25
web: add alerts section to overview

landism:
image


nicks:
hmmmm...this is the first I've heard of this feature and seems like it will make it hard to test the new grid view? It seems bad to have cards that shift around while you're trying to interact with them. am i missing something?

landism:
> It seems bad to have cards that shift around while you're trying to interact with them. am i missing something?

Good question! I've started a convo in the CH story.

landism:
after discussion, we're going to punt this and instead make this a sort option on the sidebar

PR #4116, opened 2021-01-25
web: change Sidebar tab semantics

nicks:
Hello @hyu, @landism,

Please review the following commits I made in branch nicks/ch10949:

0415af82f97e65b92e8933b47337f9b472c3fdd4 (2021-01-25 11:50:45 -0500)
web: change Sidebar tab semantics
now, we never change the number of tabs open if the selected resource is already open

### DIFF
--- a/web/src/TabNav.tsx
+++ b/web/src/TabNav.tsx
@@ -86,26 +86,19 @@ export function LegacyNavProvider(
 
 // New Overview semantics:
 //
-// 1. When you single click a resource on the left sidebar,
+// Any resource supports two navigation operations: "activate" and
+// "activate-new-tab".  The exact input bindings are user-agent OS-specific, but
+// without loss of generality, treat them as "click" and "ctrl/command-click"
+//
+// 1) When you activate or activate-new-tab a resource on the left sidebar that's already open,
 //    it changes the current tab to the new resource.
 //
-// 2. When you ctrl-click a resource on the left sidebar,
-//    it opens a new tab, and brings the new tab into focus.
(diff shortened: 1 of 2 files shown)


web/src/TabNav.tsx
```diff
@@ -86,26 +86,15 @@ export function LegacyNavProvider(
...
 //
-// 1. When you single click a resource on the left sidebar,
+// 1) When you click or ctrl-click a resource on the left sidebar that's already open,
 //    it changes the current tab to the new resource.
 //
-// 2. When you ctrl-click a resource on the left sidebar,
-//    it opens a new tab, and brings the new tab into focus.
+// 2) Otherwise, when you select a resource on the left sidebar,
+//    a) click opens it in the current tab,
+//    b) ctrl-click opens it in a new tab on the immediate right of current tab
```
hyu, line 94: Not part of this PR, but I noticed during review:

For me on Firefox, the Mac OS context menu still has precedence. I assume the expected behavior is that Tilt should hijack?

(Maybe I'm missing something. You're able to override the context menu on Mac?)

hyu (approved):

web/src/TabNav.tsx
nicks: yep, on macos it's the command key

fwiw, most of the standards docs on this tend to be super vague and hand-wavey for exactly the reason you describe - this tends to be dictated by a complicated set of overlapping OS-level preferences and browser-level preferences, so i changed this comment to be more hand-wavey

landism (approved):
so many fewer cases!

PR #4118, opened 2021-01-25
web: update j/k keyboard shortcuts to match tab navigation

nicks:
Hello @hyu, @landism,

Please review the following commits I made in branch nicks/ch10950:

b846d99e687d82143cac3ef42846370d69812d98 (2021-01-25 12:23:51 -0500)
web: update j/k keyboard shortcuts to match tab navigation

### DIFF
--- a/web/src/HUD.tsx
+++ b/web/src/HUD.tsx
@@ -413,10 +413,7 @@ export default class HUD extends Component<HudProps, HudState> {
               <Route
                 path={this.path("/r/:name/overview")}
                 render={(props: RouteComponentProps<any>) => (
-                  <OverviewResourcePane
-                    name={props.match.params?.name || ""}
-                    view={this.state.view}
-                  />
+                  <OverviewResourcePane view={this.state.view} />
                 )}
               />
               <Route render={() => <OverviewPane view={this.state.view} />} />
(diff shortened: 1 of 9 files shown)


hyu:
So, this did not behave as I expected. I thought using j/k would change the current tab, and not open any new tabs. So, basically, keyboard navigation works exactly the same as clicking things on the sidebar. But the behavior differs. Here's a video I made with mouse nav vs keyboard nav:

nicks:
this is basically a dataflow problem because under some conditions, react schedules the update differently. so i had to make a pretty big change to the dataflow to fix this

hyu (approved):

milas:
FYI I still see the behavior from Han's GIF. I think the issue is that the [effect hook to create a tab-if-not exists](https://github.com/tilt-dev/tilt/blob/565ab5a8b4bfab51b91a847650b444958c74d79d/web/src/TabNav.tsx#L139-L143) causes a race condition. It could probably be fixed by making sure that only happens on mount, but not sure if there are other circumstances where you could end up with a tab that's not in the list 🤔

nicks:
fwiw, i think we might end up getting rid of the tabs anyway, since the
current design thinking is that they're redundant with pinning, see:
https://app.clubhouse.io/windmill/story/11126/finalize-the-layout-for-overview

PR #4119, opened 2021-01-25
ui / resource view: filter sidebar for test/resource/both [ch10872]

maiamcc:
ty to @hyu for pairing on this with me!

this still needs:
- [x] tests
- [x] don't display filtering options if no tests present
- [ ] styling

but I wanted to confirm that this is a reasonable track, since a) I've never written stateful react stuff before, and b) I don't know how if doing it this way instead of with more broadly accessible state (like we do with pins etc.) is a bad move or violates some style guideline or something. (fwiw Han and I tried that, but it was a PITA b/c we'd have to change `SidebarResources` into a function instead of a class in order to use react hooks in it, and didn't seem worth it because we DON'T need this filter state to be broadly accessible, the only thing that cares about it is the sidebar.)

next step here after finishing the above TODOs is to hook up this state to local storage somehow so it can persist when you navigate away, but I figured we could do that later


web/src/OverviewResourceSidebarTypeFilter.tsx
```diff
@@ -0,0 +1,53 @@
+import React from "react"
+import styled from "styled-components"
+import { ResourceFilters } from "./types"
+
+const SidebarResourceTypeFilterStyle = styled.div`
```
nicks, line 5: instead of using css to hide these, can we just not include the element at all?

tho i guess this is all going away soon anyway.

web/src/types.ts
```diff
@@ -140,3 +140,8 @@ export enum ResourceName {
   tiltfile = "(Tiltfile)",
   all = "(all)",
 }
+
+export type ResourceFilters = {
+  showServices: boolean // TODO: namessss
```
nicks, line 145: ?

nicks (approved):

web/src/HUD.tsx
```diff
@@ -576,6 +576,7 @@ export default class HUD extends Component<HudProps, HudState> {
       let name = props.match.params.name
       return (
         <Sidebar isClosed={isSidebarClosed} toggleSidebar={this.toggleSidebar}>
+          {/* sidebarResourceFilterProviderBlah */}
```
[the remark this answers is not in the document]
maiamcc: whoops, left over from something else

web/src/OverviewResourceSidebarTypeFilter.tsx
maiamcc: Yes def can do it with a conditional--tho what do you mean "this is all going away soon"?

web/src/types.ts
maiamcc: lol whoops

maiamcc:
added some tests (with some more that I'll add in a subsequent PR but didn't want to block on this any longer)!

web/src/OverviewSidebarOptions.tsx
```diff
@@ -0,0 +1,42 @@
...
+  display: flex;
+`
+
+type SidebarResourceTypeFilterProps = {
+  curState: SidebarOptions
+  toggleShowResources: () => void
+  toggleShowTests: () => void
+}
+
+export function SidebarOptionsSetter(
```
nicks, line 15: our standard naming convention on this would be something like

export default function OverviewSidebarOptions(props: OverviewSidebarOptionsProps) {
  return <OverviewSidebarOptionsRoot>...</OverviewSidebarOptionsRoot>
}

for stuff like this, a good rule of thumb is to have a consistent naming convention even at the expense of an accurate name

I count at least 4 conflicting names in this file, which seems not ideal for a single component

nicks (approved):

web/src/OverviewSidebarOptions.tsx
maiamcc: ah whoops at least some of that is stuff i missed during a rename 😅  good catch

PR #4120, opened 2021-01-25
web: logo wordmark selected state

nicks:
Hello @hyu,

Please review the following commits I made in branch nicks/ch10952:

0878e70748790c0829ee54a9532632cc00c23ffd (2021-01-25 17:07:12 -0500)
web: logo wordmark selected state


nicks:
Screenshot from 2021-01-25 17-07-40
Screenshot from 2021-01-25 17-07-30

I created an svg that uses the "fillStd" class so that you can independently color the right components.

hyu:
Visually this works, but this PR introduced an unsightly jump when navigating between Overview and Tab!




Looks like it's not present on master:

hyu (approved):

nicks:
@hyu I was not able to reproduce the jump you're seeing...and I don't see anything in this PR that I would expect to change positioning. is it possible that you have a local change in your client causing this issue?

nicks:
i think what you're seeing is a firefox flexbox bug that happens on overflow, and you have different data when you test on master. in  any case, i fixed that too while here.

PR #4122, opened 2021-01-26
web: fix card layouts on firefox

nicks:
Hello @hyu,

Please review the following commits I made in branch nicks/ch10982:

f5d18e56b8f4dcc37b08b231e20807ba011a34e8 (2021-01-25 20:03:00 -0500)
web: fix card layouts on firefox
I suspect there's some deeper bug here where setting height: 100%
on a flex child isn't working correctly in Firefox...but I think
setting height: 100% on a flex child isn't a great practice anyway.


hyu (approved):
Yayyy no more weird layouts on Firefox!

PR #4126, opened 2021-01-26
web: centralize all tabs dataflow into OverviewTabNavProvider

nicks:
Hello @hyu, @landism,

Please review the following commits I made in branch nicks/ch10999:

950758da3b62996bec2cdd816985a6146a7b3863 (2021-01-26 15:47:02 -0500)
web: centralize all tabs dataflow into OverviewTabNavProvider
This is what I should have done in the first place. It makes everything
much simpler and ensures we're not affected as much by React scheduler
conditions.

### DIFF
--- a/web/src/OverviewResourcePane.tsx
+++ b/web/src/OverviewResourcePane.tsx
@@ -32,7 +32,7 @@ let Main = styled.div`
 export default function OverviewResourcePane(props: OverviewResourcePaneProps) {
   let nav = useTabNav()
   let resources = props.view?.resources || []
-  let name = nav.candidateTab || nav.selectedTab || ""
+  let name = nav.invalidTab || nav.selectedTab || ""
   let r: Proto.webviewResource | undefined
   let all = name === "" || name === ResourceName.all
   if (!all) {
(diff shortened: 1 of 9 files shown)


web/src/OverviewResourcePane.tsx
```diff
@@ -32,7 +32,7 @@ let Main = styled.div`
 export default function OverviewResourcePane(props: OverviewResourcePaneProps) {
   let nav = useTabNav()
   let resources = props.view?.resources || []
-  let name = nav.candidateTab || nav.selectedTab || ""
+  let name = nav.invalidTab || nav.selectedTab || ""
```
hyu, line 35: Checking my understanding: So, it's possible that you have an invalid tab (e.g., resource no longer exists). We figure this out by checking the name provided by the URL against a list of current resources. Then, if it's invalid, we do stuff like show "resource not found" or whatnot.

hyu (approved):
Nice! Tested this out and it feels a lot better than earlier today.

web/src/OverviewResourcePane.tsx
nicks: yup! exactly

PR #4127, opened 2021-01-26
web: Revamp how the OverviewItemView details are rendered, to improve the animation and interaction

nicks:
Hello @hyu,

Please review the following commits I made in branch nicks/ch10958:

57847ef15fbdb74bad4d1e6f27f326209cfc3eb7 (2021-01-26 16:48:44 -0500)
web: Revamp how the OverviewItemView details are rendered, to improve the animation and interaction

d0d5321dc0e042394ca6b370aa940074c57ef068 (2021-01-26 16:05:56 -0500)
engine: handle readiness probes for local resources (#4102)
Local controller creates a probe worker when starting a serve command
if a probe spec is defined.

The worker will emit readiness probe status update actions that will
result in updates to the local runtime state, which internally tracks
the current passing state of the readiness probe. (Local resources
without a readiness probe will consider themselves to always have a
passing probe, but this is never exposed.)

Some adjustments to the way that "ready OR succeeded" is
tracked/computed align this closer to the K8s runtime state and try
to reduce the amount of logic going on inside of the state object.
This ensures that dependent resources won't start until the probe
is passing (if defined).
3664df6f7ff3ba8606cdf222f1508d3a10156965 (2021-01-26 12:41:41 -0500)
web: make the 'View Details' button bold

### DIFF
--- a/web/src/OverviewItemView.tsx
+++ b/web/src/OverviewItemView.tsx
@@ -1,7 +1,7 @@
 import Collapse from "@material-ui/core/Collapse"
 import Popover from "@material-ui/core/Popover"
 import { makeStyles } from "@material-ui/core/styles"
-import React, { useState } from "react"
+import React, { useEffect, useRef, useState } from "react"
 import { Link } from "react-router-dom"
 import TimeAgo from "react-timeago"
 import styled, { css, keyframes } from "styled-components"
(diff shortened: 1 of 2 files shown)


web/src/OverviewItemView.tsx
```diff
@@ -484,15 +532,27 @@ export function OverviewItemDetails(props: OverviewItemDetailsProps) {
...
+  let ref = useRef(null as any)
+
+  // Simulate an 'expansion' animation by modifying the height post-render.
+  useEffect(() => {
+    if (ref.current && height) {
+      ref.current.style.height = ref.current.firstChild.scrollHeight + "px"
+    }
+  }, [height])
+
+  let widthStyle = width ? width + "px" : "330px"
```
hyu, line 546: Minor question: Since this default width is referenced in multiple places within this React component, do you think we should have a var or const somewhere in here that declares the 330px? To make it slightly harder to change the value in one place but leave incorrect calculations elsewhere?

Previously, I stored these values in style-helper, but I do see the sense in confining vars only used within a single component in that same component.

hyu (approved):
Nice! Subtle but definite improvement for the animation.

web/src/OverviewItemView.tsx
nicks: this is a good question!

The long-ish answer is that there are two different components here:
(1) OverviewItemView, which has a dynamic width based on the current screen size and the number of other cards
(2) OverviewItemDetail, which must be adjusted dynamically to match OverviewItemView

but since we want to view both of these components in storybooks (outside the context of the stuff that's determining their size), we use 330px in a bunch of places as a nice 'test' fixture size, but not an authoritative value that you'd want to keep in-sync.

But maybe the answer is to move test values into test files, so they don't accidentally get used as real values.

hyu: Ahhh, I see. Yeah, in that case it would have helped me to understand that 330px was really just for the test fixture. Upon a quick read, I made the incorrect assumption that it was a "real" default for the actual component. So, I vote for a convention of keeping test values in test files.

PR #4130, opened 2021-01-27
integration: add coverage for local readiness probes

milas:
~There's some general integration tests improvements here (cc @nicks @maiamcc) to copy the test resource directory (i.e. the dir with `Tiltfile` for the test) to a temporary directory that gets cleaned up at the end of the test. The local resource test was already creating files and manually cleaning up - the readiness probe additions added more to that; I figured rather than continuing to manually clean up (and potentially risk messing with local working repo), this made sense.~
**EDIT:** Just kidding - some of the integration tests are structured in such a way that this isn't quite as straightforward (i.e. it'll need to touch a bunch of the fixtures) so reverted it and added explicit cleanup for new files in this test. 

Main focus is expanding the `local_resource` integration test to include a readiness probe. (See #4112 for `Tiltfile` syntax for readiness probes and #4102 for engine support.)

There's also a small improvement to the readiness probe log output so that it's prepended with `[readiness probe]` to disambiguate from resource app logs. (They're logged at `VerboseLvl` and only when the state _changes_, not on *every* probe run, so they shouldn't be too spammy, but they could be extremely confusing to a user with no context. Longer-term, we might want to consider a better separation here.)

### DIFF
--- a/integration/fixture_test.go
+++ b/integration/fixture_test.go
@@ -45,9 +45,63 @@ type fixture struct {
 	tiltArgs      []string
 }
 
-func newFixture(t *testing.T, dir string) *fixture {
-	dir = filepath.Join(packageDir, dir)
-	err := os.Chdir(dir)
+func copyDir(src string, dest string) error {
+	absSrc, err := filepath.Abs(src)
+	if err != nil {
+		return fmt.Errorf("could not get absolute path for src dir: %v", err)
+	}
+	absDest, err := filepath.Abs(dest)
+	if err != nil {
+		return fmt.Errorf("could not get absolute path for dest dir: %v", err)
+	}
+	return filepath.Walk(absSrc, func(srcPath string, info os.FileInfo, err error) error {
+		if err != nil {
+			return err
+		}
+		if srcPath == absSrc {
+			return nil
+		}
+		if !strings.HasPrefix(srcPath, absSrc) {
+			return fmt.Errorf("invalid src path: %s", srcPath)
+		}
+		destPath := strings.Replace(srcPath, absSrc, absDest, 1)
+		if info.IsDir() {
+			return os.Mkdir(destPath, info.Mode())
+		} else {
+			srcData, err := ioutil.ReadFile(srcPath)
+			if err != nil {
+				return fmt.Errorf("failed to read src file %q: %v", srcPath, err)
+			}
+			if err := ioutil.WriteFile(destPath, srcData, info.Mode()); err != nil {
+				return fmt.Errorf("failed to write dest file %q: %v", destPath, err)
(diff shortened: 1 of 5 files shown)


integration/fixture_test.go
```diff
@@ -80,11 +134,14 @@ func (f *fixture) testDirPath(s string) string {
 }
 
 func (f *fixture) installTilt() {
+	f.t.Helper()
```
landism, line 83: TIL

there are so many places the lack of this has annoyed me!

integration/local_resource_test.go
```diff
@@ -1,34 +1,52 @@
...
 	require.NoError(t, f.logs.WaitUntilContains("hello! foo #1", 5*time.Second))
+
+	// write a sentinel file for the probe to find and change its result
+	require.NoError(t, ioutil.WriteFile(f.testDirPath("probe-success"), nil, 0777))
+	require.NoError(t, f.logs.WaitUntilContains("[readiness probe] fake probe success message", 5*time.Second))
+
+	// wait for second resource to start and then ensure that the order in the logs is as expected
+	require.NoError(t, f.logs.WaitUntilContains(barServeLogMessage, 5*time.Second))
+	curLogs := f.logs.String()
+	assert.Greater(t, strings.Index(curLogs, barServeLogMessage), strings.Index(curLogs, readinessProbeSuccessMessage),
```
landism, line 42: IIUC, `readinessProbeSuccessMessage`'s `"`s don't actually appear in the logs, so `strings.Index` here returns `-1`, so the index of `barServeLogMessage` will always be greater (presuming it actually appears in the log)

It'd probably be safest to get the indexes separately and assert they're not -1, but just using `readinessProbeSuccessMessage` on line 37 would suffice (to catch the issue, and then you'd need to remove the quotes from the const).

landism (approved):

landism:
> Just kidding - some of the integration tests are structured in such a way that this isn't quite as straightforward

Aw, I was excited for this

integration/fixture_test.go
milas: You can see where I've broken a test based on this getting added 😶

milas:
I threw the test tempdir refactor into a branch - didn't want to abandon it entirely but also didn't want to pollute this PR with a million random fixture changes 😂

integration/local_resource_test.go
milas: Whoops - that was the whole reason I'd made a `const` with the log message so that I could assert they both existed before doing index magic but did it very sloppily 😅

milas: Fixed in fd7cd65d

Issue #3434, opened 2020-06-10
Attempt to randomize port on startup if default is unavailable


rpocase:
When switching between tilt projects that contain tilt wrappers (e.g. `make tilt`) I'll often hit issues with port allocation issues. My options tend to be:
* stop one project (which could be costly to start back up) 
* fall back to running tilt directly (could need to run several setup tasks directly) 
* assign tilt ports in each project (or at least allow setting them via ENV_VAR)

I'd like for tilt to have a (optional?) method to randomize port allocation if the default port is in use. 

An alternative I've thought about is a tilt daemon/manager that is project context aware. Tilt could host its web-ui on the same port but provide subpaths for components running under different Tiltfile contexts (e.g. localhost:10350/project1, localhost:10350/project2).

nicks:
fwiw, i've been hitting this a lot myself recently. I too go back and forth if I actually want to have two separate tilt instances to keep track of, or rather if I want them to "join" together

The first one has a simpler mental model, but the second one seems more useful (particularly if we're able to crack the nut of sidebar organization when you have a lot of services)

PR #4131, opened 2021-01-27
integration: use a random port for each tilt up invocation

milas:
This just makes things easier if you've got Tilt running in the
background on your machine already so that you can still run an
integration test.

### DIFF
--- a/integration/crash_test.go
+++ b/integration/crash_test.go
@@ -4,22 +4,27 @@ package integration
 
 import (
 	"bytes"
+	"fmt"
 	"testing"
 	"time"
 
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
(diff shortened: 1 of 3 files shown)


victorwuky:
Thanks @milas for putting this up. Here's an issue: https://github.com/tilt-dev/tilt/issues/3434.

nicks (approved):

milas:
@victorwuky This is _just_ for integration tests so they don't fail when running locally if you happen to have Tilt open in the background, i.e. the test logic is explicitly passing a `--port` arg with a (sorta) random value now, so it won't address that issue. The broader enhancement(s) to actually simplifying using Tilt across multiple projects simultaneously remain valid.

victorwuky:
Thanks for explaining @milas !

PR #4136, opened 2021-01-28
engine: fix local runtime state for jobs

### LINKED ISSUE #4133: `tilt ci` doesn't terminate once all containers are ready
This feels like a variant of issue #3262 in the sense that `tilt ci` appears to hang indefinitely after a workload has been deployed to a Kubernetes cluster.

### Expected Behavior

`tilt ci` will exit once the `kubebuilder` operator from the reproduce below is deployed and ready in a local dev Kubernetes cluster.

### Current Behavior

`tilt ci` doesn't exit once all of the containers in the operator pod are ready.

### Steps to Reproduce
1. Provision a local kubernetes cluster (`kind`, `minikube`, `k3s`, it shouldn't matter).  Make sure your `~/.kube/config` points at that cluster.
2. Run the following in an empty directory:
```bash
cat <<EOF > Tiltfile
load('ext://kubebuilder', 'kubebuilder')
kubebuilder("domain.tilt.dev", "tiltgroup", "v1", "CustomKind")
EOF
tilt ci
```
3. Observe that the `tilt ci` command doesn't terminate even after all of the containers in the `controller-manager` pod are ready.

#### `tilt doctor` Output

```bash
$ tilt doctor

Tilt: v0.18.5, built 2021-01-22
System: darwin-amd64
---
Docker
- Host: [default]
- Version: 1.41
- Builder: 2
---
Kubernetes
- Env: kind-0.6+
- Context: kind-kind
- Cluster Name: kind-kind
- Namespace: default
- Version: v1.20.2
---

---
Analytics Settings
- User Mode: opt-out
- Machine: df1a2bcef66ccf958d6affd19a2cd1c9
- Repo: 
```

#### About Your Use Case

The `tilt ci` command provides an ideal way of performing smoke tests on Kubernetes deployment specs & container images.  We have CI builds that worked using the `tilt ci` incantation with `tilt 0.18.3` but newer builds of `tilt` won't work on our local machines.

This issue prevents developers from using the `tilt ci` command locally to replicate the steps running in our CI builds.

The attached [`tilt-dump.json.txt`](https://github.com/tilt-dev/tilt/files/5883134/tilt-dump.json.txt) was generated from the lingering `tilt ci` incantation after deploying the operator to a local kind cluster running on my laptop.

milas:
This was regressed by #4082 - this commit reverts the changes made
there and refactors readiness probe logic to work with it.

Closes #4133.


internal/engine/local/controller.go
```diff
@@ -215,17 +215,31 @@ func processStatuses(
...
-		if !stillHasSameProcNum() {
+		if !stillHasSameProcNum() || sm.status == Unknown {
 			continue
 		}
-		procState := sm.status.ToProcessState()
-		if procState != "" {
+
+		var runtimeStatus model.RuntimeStatus
+		if sm.status == Done || sm.status == Error {
+			runtimeStatus = model.RuntimeStatusError
```
landism, line 229: this seem wrong for jobs - `sm.Status == Done` doesn't indicate an error, right?

internal/engine/upper.go
```diff
@@ -786,7 +800,16 @@ func handleLocalServeReadinessProbeAction(ctx context.Context, state *store.Engi
 	}
 
 	lrs := ms.LocalRuntimeState()
-	lrs.SetReadinessProbeState(action.Ready)
+	lrs.Ready = action.Ready
+	lrs.LastReadyOrSucceededTime = time.Now()
```
landism, line 804: this should only happen if `Ready` is `true`, right?

internal/engine/local/controller.go
milas: This was an unintentional change in behavior, will remove it.

For context, previously [`ToRuntimeStatus()`](https://github.com/tilt-dev/tilt/blob/v0.18.3/internal/engine/local/controller.go#L241-L259) would just return an empty str for `Done`, so no update would get emitted, so we can continue to do the same thing here.

internal/engine/upper.go
milas: Fixed in 4eee419

internal/engine/local/controller.go
milas: Restored parity with previous behavior in 737a56b

landism (approved):

PR #4139, opened 2021-01-28
web: log tiltfile errors with the right error level

nicks:
Hello @landism,

Please review the following commits I made in branch nicks/ch11044:

b3587be01ec274bddaffef5ef96e87f45a70045f (2021-01-28 15:24:34 -0500)
web: log tiltfile errors with the right error level
also, simplify configs controller tests while i'm here

### DIFF
--- a/internal/tiltfile/tiltfile.go
+++ b/internal/tiltfile/tiltfile.go
@@ -222,7 +222,9 @@ func (tfl tiltfileLoader) Load(ctx context.Context, filename string, userConfigS
 	tlr.UpdateSettings = us
 
 	duration := time.Since(start)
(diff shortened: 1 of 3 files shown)


internal/tiltfile/tiltfile.go
```diff
@@ -222,7 +222,9 @@ func (tfl tiltfileLoader) Load(ctx context.Context, filename string, userConfigS
 	tlr.UpdateSettings = us
 
 	duration := time.Since(start)
-	s.logger.Infof("Successfully loaded Tiltfile (%s)", duration)
+	if tlr.Error == nil {
```
landism, line 225: Haha, this has annoyed me for so long. It never occurred to me that it was actually as simple as this.

landism (approved):

PR #4147, opened 2021-02-01
web: Toggle Trigger Mode

hyu:
There's some CSS cleanup and fine-tuning that I'd like to do:
- adjust spacing in Resource view (currently covering timeago)
- maybe use flexbox layout for the buttons, instead of absolute positioning...?

But this PR also includes implementation, and I've been sitting on this long enough and just want to push it out to get some momentum.

Screen Shot 2021-02-01 at 12 49 33 PM


web/src/OverviewItemView.tsx
```diff
@@ -244,6 +254,24 @@ export function triggerUpdate(name: string, action: string) {
   })
 }
 
+export function toggleTriggerMode(name: string, mode: TriggerMode) {
+  incr("ui.web.toggleTriggerMode", { to_mode: mode.toString() })
+
+  let url = `//${window.location.host}/api/override/trigger_mode`
```
nicks, line 260: fwiw, i don't think the window.location.host is necessary here, you can use a '/api/overide' url

web/src/OverviewItemView.tsx
```diff
@@ -244,6 +254,24 @@ export function triggerUpdate(name: string, action: string) {
   })
 }
 
+export function toggleTriggerMode(name: string, mode: TriggerMode) {
+  incr("ui.web.toggleTriggerMode", { to_mode: mode.toString() })
```
nicks, line 258: to_mode -> toMode for consistency?

web/src/OverviewItemView.tsx
```diff
@@ -244,6 +254,24 @@ export function triggerUpdate(name: string, action: string) {
...
+
+  let url = `//${window.location.host}/api/override/trigger_mode`
+
+  fetch(url, {
+    method: "post",
+    body: JSON.stringify({
+      manifest_names: [name],
```
nicks, line 265: i thought all our json serialization stuff defaulted to camelCase, as in this file:
https://github.com/tilt-dev/tilt/blob/master/web/src/view.d.ts

i worry this is going to really bite us later if half our code uses camelCase and half our code uses slug case, depending on who wrote it that day.

web/src/SidebarPinButton.tsx
```diff
@@ -30,8 +30,7 @@ let UnpinnedPinIcon = styled(PinResourceFilledSvg)`
   opacity: 0;
 
   .u-showPinOnHover:hover &,
-  ${PinButton}:focus &,
-  ${PinButton}.u-persistShow & {
```
nicks, line 34: why did the persistShow stuff go away? it seems weird that a PR about trigger mode has a bunch of changes to pinning

web/src/OverviewItemView.tsx
```diff
@@ -164,6 +167,13 @@ export let OverviewItemBox = styled.div`
     animation: ${barberpole} 8s linear infinite;
   }
 `
+let OverviewItemActions = styled.div`
```
nicks, line 170: are you sure this does the layout correctly when the resource name is long?

usually mixing absolute positioning with flexbox layouts defeats the point of using flexbox, because now the flex doesn't work correctly.

web/src/SidebarItemView.tsx
```diff
@@ -179,6 +185,14 @@ let SidebarItemTimeAgo = styled.span`
   align-items: center;
 `
 
+let SidebarItemActions = styled.div`
+  position: absolute;
```
nicks, line 189: ditto here re: using absolute positioning in a flexbox

web/src/TriggerModeToggle.tsx
```diff
@@ -0,0 +1,94 @@
...
+  .fillStd {
+    fill: ${Color.blue};
+    transition: fill ${AnimDuration.short} linear;
+  }
+  .strokeStd {
+    stroke: ${Color.blue};
+    opacity: ${ColorAlpha.almostOpaque};
+    transition: stroke ${AnimDuration.short} linear;
+  }
+  .autoSvgToggleOn {
```
nicks, line 28: maybe call this autoSvg-toggleOn, to signify that this is the "toggleOn" component of the auto svg?

web/src/TriggerModeToggle.tsx
```diff
@@ -0,0 +1,94 @@
...
+    e.preventDefault()
+
+    // stopPropagation prevents the overview card from opening.
+    e.stopPropagation()
+
+    props.onModeToggle(desiredMode)
+  }
+
+  return (
+    <TriggerModeToggleStyle
```
nicks, line 84: should these be called TriggerModeToggleRoot to match the existing convention?

web/src/SidebarPinButton.tsx
hyu: Ah yes, sorry about this! I didn't do this carefully enough.

Based on designs that Surbhi and I were working on, I ended up moving the pin position to make room for the new Auto button (it won't fit in the Sidebar otherwise, since pin is also on the bottom right). So I was playing around with the classes to try to understand what `u-persistShow` does, but I didn't properly clean up after myself.

Can you explain what `u-persistShow` does, actually? I had thought it was necessary because buttons on the BuildBox at the bottom sort of gets covered up by the Popover without this class. But now I'm not so sure if that's what it does.

nicks: i think there were a bunch of iterations where it showed up differently in different contexts. to be clear, i'm not opposed to doing a cleanup pass on the pinning stuff, i was mainly surprised to see it tacked on to an unrelated PR.

web/src/TriggerModeToggle.tsx
hyu: Good naming suggestion, thanks!

web/src/TriggerModeToggle.tsx
hyu: Oh right, we had used the "Style" suffix at one point. I'll try to remember the "Root" convention.

web/src/OverviewItemView.tsx
hyu: Thanks for calling this out. I'm updating this to consistently use camelCase.

web/src/SidebarPinButton.tsx
hyu: Gotcha! I reverted the `persistShow` changes but kept the change where I moved the button.

web/src/OverviewItemView.tsx
```diff
@@ -234,7 +244,7 @@ export function triggerUpdate(name: string, action: string) {
   fetch(url, {
     method: "post",
     body: JSON.stringify({
-      manifest_names: [name],
+      manifestNames: [name],
```
nicks, line 247: is there a server side you have to update?

web/src/TriggerModeToggle.tsx
```diff
@@ -0,0 +1,94 @@
...
+  .fillStd {
+    fill: ${Color.blue};
+    transition: fill ${AnimDuration.short} linear;
+  }
+  .strokeStd {
+    stroke: ${Color.blue};
+    opacity: ${ColorAlpha.almostOpaque};
+    transition: stroke ${AnimDuration.short} linear;
+  }
+  .autoSvg-ToggleOn {
```
nicks, line 28: autoSvg-ToggleOn -> autoSvg-toggleOn

web/src/TriggerModeToggle.tsx
```diff
@@ -0,0 +1,94 @@
...
+    .fillStd {
+      fill: ${Color.grayLight};
+    }
+    .strokeStd {
+      stroke: ${Color.grayLight};
+    }
+    .autoSvg-ToggleOn {
+      fill: none;
+    }
+    .autoSvgToggleOff {
```
nicks, line 42: autoSvgToggleOff -> autoSvg-toggleOff

nicks (approved):

web/src/OverviewItemView.tsx
hyu: Yes, I agree that flexbox layout is better. Fixed!

web/src/OverviewItemView.tsx
hyu: So, I started making these case style changes, but since I'm primarily focused on the front-end of this PR, I was a bit careless with making the changes consistently to the server. I also noticed that there are other places we should use camelCase, which are outside the scope of this PR (e.g., `actionPayload` struct).

I decided to revert the changes and leave it to Maia to review in a separate PR when she comes back.

nicks: sgtm

web/src/SidebarItemView.tsx
hyu: Fixing this too!

web/src/OverviewItemView.tsx
maiamcc: @nicks [all the hud api endpoints use snake case](https://github.com/tilt-dev/tilt/blob/57dbb09cac112a251bbb2547fc085d9358333579/internal/hud/server/server.go#L48), so at least we're consistent within api endpoints and within typescript stuff, so I feel fine leaving this for now. (there might be a case to be made for converting the api endpoints to camelCase but that seems like a conversation separate from this epic.)

PR #4150, opened 2021-02-01
web: ensure snapshots are under 4MB upload limit

### LINKED ISSUE #3687: snapshot creation errors if it's larger than 4mb
If you try to send a snapshot larger than 4mb, you hit the GRPC message size limit and get an error

milas:
Truncate logs included in the snapshot upload to avoid being rejected
by the server for > 4MB size.

The logs are truncated to 2MB as there is a substantial amount of
overhead from JSON as well as the rest of the actual UI state that
needs to fit in there as well.

Closes #3687.

### DIFF
--- a/web/src/HUD.tsx
+++ b/web/src/HUD.tsx
@@ -56,6 +56,10 @@ type HudProps = {
   interfaceVersion: InterfaceVersion
 }
 
+// Snapshot logs are capped to 2MB (max upload size is 4MB; this ensures between the rest of state and JSON overhead
+// that the snapshot should still fit)
+const maxSnapshotLogSize = 2 * 1000 * 1000;
+
 // The Main HUD view, as specified in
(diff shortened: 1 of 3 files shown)


web/src/HUD.tsx
```diff
@@ -56,6 +56,10 @@ type HudProps = {
   interfaceVersion: InterfaceVersion
 }
 
+// Snapshot logs are capped to 2MB (max upload size is 4MB; this ensures between the rest of state and JSON overhead
+// that the snapshot should still fit)
+const maxSnapshotLogSize = 2 * 1000 * 1000;
```
nicks, line 61: can we make it more like 1MB? metadata overhead can be pretty high!

nicks (approved):

web/src/HUD.tsx
milas: Yep, lowered in 0c708297!

PR #4157, opened 2021-02-04
web: persist sidebar options in localStorage

landism:
fwiw all the indenting in the tests is just for:
```
afterEach(() => {
    localStorage.clear()
  })
```

### DIFF
--- a/web/src/OverviewPane.test.tsx
+++ b/web/src/OverviewPane.test.tsx
@@ -1,7 +1,7 @@
 import { mount, ReactWrapper } from "enzyme"
 import React from "react"
 import { MemoryRouter } from "react-router"
-import { makeKey, tiltfileKeyContext } from "./LocalStorage"
+import { accessorsForTesting, tiltfileKeyContext } from "./LocalStorage"
 import OverviewItemView from "./OverviewItemView"
 import OverviewPane, {
   AllResources,
@@ -27,6 +27,10 @@ function assertContainerWithResources(
   }
 }
 
+const pinnedResourcesAccessor = accessorsForTesting<string[]>(
+  "pinned-resources"
+)
+
 it("renders all resources if no pinned and no tests", () => {
   const root = mount(
     <MemoryRouter initialEntries={["/"]}>{TwoResources()}</MemoryRouter>
@@ -38,10 +42,7 @@ it("renders all resources if no pinned and no tests", () => {
 })
 
 it("renders pinned resources", () => {
-  localStorage.setItem(
-    makeKey("test", "pinned-resources"),
-    JSON.stringify(["snack"])
-  )
+  pinnedResourcesAccessor.set(["snack"])
 
   const root = mount(
     <MemoryRouter initialEntries={["/"]}>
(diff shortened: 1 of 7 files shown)


maiamcc (approved):
🙌 amazing! Can you also add a test that local storage does the thing?

web/src/OverviewPane.test.tsx
```diff
@@ -38,10 +42,7 @@ it("renders all resources if no pinned and no tests", () => {
 })
 
 it("renders pinned resources", () => {
-  localStorage.setItem(
-    makeKey("test", "pinned-resources"),
-    JSON.stringify(["snack"])
-  )
+  pinnedResourcesAccessor.set(["snack"])
```
maiamcc, line 45: 😍

web/src/SidebarResources.test.tsx
```diff
@@ -132,8 +135,108 @@ describe("SidebarResources", () => {
...
-    expect(localStorage.getItem(makeKey("test", "pinned-resources"))).toEqual(
-      JSON.stringify(["vigoda"])
-    )
+    expect(pinnedItemsAccessor.get()).toEqual(["vigoda"])
   })
+
+  const loadCases: [string, SidebarOptions, string[]][] = [
+    [
+      "showResources",
+      { showResources: false, showTests: true, alertsOnTop: true },
```
maiamcc, line 144: I believe the default value of `alertsOnTop` is false -- not a big thing but if you're trying to test one property at a time getting loaded from localStorage and being different from the default, you should flip all the `alertsOnTop` values.

web/src/SidebarResources.test.tsx
```diff
@@ -132,8 +135,108 @@ describe("SidebarResources", () => {
...
+        options.showTests,
+        options.alertsOnTop
+      )
+    }
+  )
+
+  const saveCases: [string, SidebarOptions][] = [
+    [
+      "showResources",
+      { showResources: false, showTests: true, alertsOnTop: true },
```
maiamcc, line 195: see note above re: default value of `alertsOnTop`

web/src/SidebarResources.test.tsx
```diff
@@ -132,8 +135,108 @@ describe("SidebarResources", () => {
...
+      )
+
+      root.find("input#resources").simulate("change", {
+        target: { checked: expectedOptions.showResources },
+      })
+      root
+        .find("input#tests")
+        .simulate("change", { target: { checked: expectedOptions.showTests } })
+      if (
+        root.find(AlertsOnTopToggle).hasClass("is-enabled") !=
```
maiamcc, line 232: this should probably be `simulate("change"...)` to match the others? (maybe `simulate("click")` is sufficient for saving things to storage but not for triggering the onChange handler? eh I don't know. You can also leave this as-is and I can fix it if need be when I go back and write AlertsOnTop tests.)

maiamcc (approved):
HERO 🥇

web/src/SidebarResources.test.tsx
landism: The others are `CheckBox`es and define `onChange` handlers, while `AlertsOnTopToggle` is a `Button` and defines an `onClick` handler. It would be nice if they were more consistent, but I think that's a code change rather than a test change, and more than I want to take on in this PR.

PR #4164, opened 2021-02-04
tiltfile: Adds links to dc_resources

ebenoist:
Adds `links()` support to `dc_resource` for parity with k8s and local. I tried following the original local pr to add links, but I'm sure I've missed a thing or two.

### DIFF
--- a/internal/tiltfile/tiltfile_test.go
+++ b/internal/tiltfile/tiltfile_test.go
@@ -47,6 +47,7 @@ import (
 
 type localResourceLinks []model.Link
 type k8sResourceLinks []model.Link
+type dcResourceLinks []model.Link
 
 const simpleDockerfile = "FROM golang:1.10"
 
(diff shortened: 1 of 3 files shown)


internal/tiltfile/tiltfile_test.go
```diff
@@ -814,6 +815,30 @@ k8s_resource('foo', links=EXPR)
 				db(image("gcr.io/foo")),
 				deployment("foo"))
 		})
+
+		t.Run("dc-"+c.name, func(t *testing.T) {
+			f := newFixture(t)
+			defer f.TearDown()
+
+			f.setupFoo()
```
nicks, line 823: this sets up a kubernetes file. i think you need a docker_compose file? something like

```
	f.file("docker-compose.yml", `version: '3'
services:
  foo:
    image: gcr.io/foo
`)
```

and then drop

```
docker_compose('docker-compose.yml')
```

next to dc_resource

nicks (commented):
nice addition! but i think the tests are failing? added a comment below

internal/tiltfile/tiltfile_test.go
ebenoist: Will do! Thanks for taking a look

ebenoist: Got the tests appropriately setup (and more importantly passing).

nicks (approved):
wooo this is great thank you. LGTM!

I think we'll also need to modify this function:
https://github.com/tilt-dev/tilt/blob/master/internal/store/engine_state.go#L698
to get the links into the UI (which merges the implied ports and the explicit ports), but that's ok as a separate PR

ebenoist:
> wooo this is great thank you. LGTM!
> 
> I think we'll also need to modify this function:
> https://github.com/tilt-dev/tilt/blob/master/internal/store/engine_state.go#L698
> to get the links into the UI (which merges the implied ports and the explicit ports), but that's ok as a separate PR

Added this here! https://github.com/tilt-dev/tilt/pull/4175

PR #4170, opened 2021-02-05
web: clear logs functionality

### LINKED ISSUE #1885: Feature request: Clear logs?
Can we have a way to clear logs? I'm looking and can't find a way :)

I've been using the WebUI and would love a way to clear logs there; perhaps adding a "Clear logs" button at the top would do the trick? Or maybe also add a keyboard-command "Ctrl-K" to clear like we do in the Terminal?

For example

I tried reverting back to using the Tilt Terminal View to view/clear the logs, but no luck... Ctrl-K/L kills in the AsciiUI!

Cheers

milas:
New button in the resource bar to clear logs. On the all view,
this will clear logs for all resources. On an individual resource,
it will just clear that resource's logs.

If a resource currently has error/warnings, the latest build or
runtime span respectively will be retained. This is to avoid having
a bunch of alerts that just show empty text on click.

### TODO
- [x] Hide button in snapshots

Closes #1885.

### DIFF
--- a/web/src/ClearLogs.tsx
+++ b/web/src/ClearLogs.tsx
@@ -1,8 +1,5 @@
 import React from "react"
 import styled from "styled-components"
-import { Alert } from "./alerts"
-import { FilterSource } from "./logfilters"
-import { isBuildSpanId, isTiltfileSpanId } from "./logs"
 import { useLogStore } from "./LogStore"
 import { Color, FontSize } from "./style-helpers"
 
@@ -18,48 +15,19 @@ const ClearLogsButton = styled.a`
 
 export interface ClearLogsProps {
   resource?: Proto.webviewResource
-  alerts?: Alert[]
 }
 
-const ClearLogs: React.FC<ClearLogsProps> = ({ resource, alerts }) => {
+const ClearLogs: React.FC<ClearLogsProps> = ({ resource }) => {
   const logStore = useLogStore()
   const label = resource?.name ? "Clear Logs" : "Clear All Logs"
 
   const clearLogs = () => {
     let spans: { [key: string]: Proto.webviewLogSpan }
-    let relevantAlerts: Alert[]
     if (resource) {
       const manifestName = resource.name ?? ""
       spans = logStore.spansForManifest(manifestName)
-      relevantAlerts = (alerts || []).filter(
-        (alert) => alert.resourceName == manifestName
-      )
     } else {
       spans = logStore.allSpans()
-      relevantAlerts = alerts || []
-    }
-
-    // don't delete the latest span that corresponds to an alert
-    for (const alert of relevantAlerts) {
-      let spanIdsForResource = logStore.getOrderedSpansIdsForManifest(
-        alert.resourceName
-      )
-      if (alert.source == FilterSource.build) {
-        spanIdsForResource = spanIdsForResource.filter(
-          (spanId) => isBuildSpanId(spanId) || isTiltfileSpanId(spanId)
-        )
-      } else if (alert.source == FilterSource.runtime) {
-        spanIdsForResource = spanIdsForResource.filter(
-          (spanId) => !isBuildSpanId(spanId) && !isTiltfileSpanId(spanId)
-        )
-      } else {
-        // unsupported source, so don't attempt to prevent truncation
-        continue
-      }
-      if (spanIdsForResource.length !== 0) {
-        const latestSpanId = spanIdsForResource[spanIdsForResource.length - 1]
-        delete spans[latestSpanId]
-      }
     }
 
     logStore.removeSpans(Object.keys(spans))
--- a/web/src/ClearLogs.tsx
+++ b/web/src/ClearLogs.tsx
@@ -1,12 +1,22 @@
 import React from "react"
 import styled from "styled-components"
 import { useLogStore } from "./LogStore"
-import { Color, FontSize } from "./style-helpers"
+import {
+  AnimDuration,
+  Color,
+  FontSize,
+  mixinResetButtonStyle,
+} from "./style-helpers"
 
-const ClearLogsButton = styled.a`
+const ClearLogsButton = styled.button`
+  ${mixinResetButtonStyle}
   margin-left: auto;
-  cursor: pointer;
   font-size: ${FontSize.small};
+  color: ${Color.white};
+
+  & {
+    transition: color ${AnimDuration.default} ease;
+  }
 
   &:hover {
     color: ${Color.blue};
--- a/web/src/ClearLogs.tsx
+++ b/web/src/ClearLogs.tsx
@@ -13,10 +13,7 @@ const ClearLogsButton = styled.button`
   margin-left: auto;
   font-size: ${FontSize.small};
   color: ${Color.white};
-
-  & {
-    transition: color ${AnimDuration.default} ease;
-  }
+  transition: color ${AnimDuration.default} ease;
 
   &:hover {
     color: ${Color.blue};
--- a/web/src/ClearLogs.tsx
+++ b/web/src/ClearLogs.tsx
@@ -1,0 +1,71 @@
+import React from "react"
+import styled from "styled-components"
+import { Alert } from "./alerts"
+import { FilterSource } from "./logfilters"
+import { isBuildSpanId, isTiltfileSpanId } from "./logs"
+import { useLogStore } from "./LogStore"
+import { Color, FontSize } from "./style-helpers"
+
+const ClearLogsButton = styled.a`
+  margin-left: auto;
+  cursor: pointer;
+  font-size: ${FontSize.small};
+
+  &:hover {
+    color: ${Color.blue};
+  }
+`
+
+export interface ClearLogsProps {
+  resource?: Proto.webviewResource
+  alerts?: Alert[]
+}
+
+const ClearLogs: React.FC<ClearLogsProps> = ({ resource, alerts }) => {
+  const logStore = useLogStore()
+  const label = resource?.name ? "Clear Logs" : "Clear All Logs"
+
+  const clearLogs = () => {
+    let spans: { [key: string]: Proto.webviewLogSpan }
+    let relevantAlerts: Alert[]
+    if (resource) {
+      const manifestName = resource.name ?? ""
+      spans = logStore.spansForManifest(manifestName)
+      relevantAlerts = (alerts || []).filter(
+        (alert) => alert.resourceName == manifestName
+      )
+    } else {
+      spans = logStore.allSpans()
+      relevantAlerts = alerts || []
+    }
+
+    // don't delete the latest span that corresponds to an alert
+    for (const alert of relevantAlerts) {
+      let spanIdsForResource = logStore.getOrderedSpansIdsForManifest(
+        alert.resourceName
+      )
+      if (alert.source == FilterSource.build) {
+        spanIdsForResource = spanIdsForResource.filter(
+          (spanId) => isBuildSpanId(spanId) || isTiltfileSpanId(spanId)
+        )
+      } else if (alert.source == FilterSource.runtime) {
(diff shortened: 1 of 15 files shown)


web/src/ClearLogs.tsx
```diff
@@ -0,0 +1,71 @@
...
+      spans = logStore.spansForManifest(manifestName)
+      relevantAlerts = (alerts || []).filter(
+        (alert) => alert.resourceName == manifestName
+      )
+    } else {
+      spans = logStore.allSpans()
+      relevantAlerts = alerts || []
+    }
+
+    // don't delete the latest span that corresponds to an alert
```
nicks, line 42: you've introduced a new concept here of a "latest" span. I don't think this concept makes sense. 

You adapted it from ordered build spans, which was already sus. But at least ordered build spans corresponds to something real in the underlying engine - a manifest can only have one build at a time. But I don't think this is true for arbitrary spans for things like pod logs and events

might make sense to take a step back and think about what we're trying to do and the right way to model it?

nicks: the special-casing of tiltfile span ids is a pretty big red flag to me that we may need to talk more about the model of what you're trying to do here

but i'm not 100% clear on why we need to keep the last alert at all?

milas: The intent here was to avoid getting into a state where the alerts bar + log filters are showing errors/warnings but there's no longer any content.

Example:
Screen Shot 2021-02-08 at 1 14 50 PM

(That's after hitting "Clear Logs" in the current state of this branch.)

The runtime case is definitely weirder given that it only applies to K8s and as shown here, it's unlikely the log output is useful to diagnose the problem anyway.

Arguably, this is a pre-existing issue as alerts & log output are decoupled as-is -- alerts actually have a message but afaict it's not shown anywhere, so it's already possible to have the log filter suggesting there are errors/warnings but none actually exist in the log output.

milas: I'll simplify to removing _all_ spans for manifest rather than trying to keep around the related error context. I do think the way alerts are presented / interact with log filters is fairly unintuitive when using the UI and clearing logs can/might make that slightly worse, but at least it's an intentional action so hopefully not hugely surprising

nicks: ah, ok. so if I understand this correctly - you're trying to going through the alerts, and trying to preserve the logs that keep those alerts on screen?

what if we did the opposite? go through the logs that we're clearing, and add a flag to indicate "ignore alerts in span X before time Y"?

milas: Okay, take a look at latest changes. Since the alert logic already has span info, it was easier to filter out alerts that are for a span that doesn't exist in the log store rather than doing additional book keeping by timestamp.

web/src/OverviewActionBar.tsx
```diff
@@ -475,6 +478,7 @@ export default function OverviewActionBar(props: OverviewActionBarProps) {
           filterSet={props.filterSet}
           alerts={alerts}
         />
+        {isSnapshot || <ClearLogs resource={resource} />}
```
nicks, line 481: will this make the text "true" appear in snapshots? not sure that's what you want

web/src/OverviewLogPane.tsx
```diff
@@ -212,12 +216,33 @@ export class OverviewLogComponent extends Component<OverviewLogComponentProps> {
...
   }
 
-  onLogUpdate() {
+  onLogUpdate(e: LogUpdateEvent) {
     if (!this.rootRef.current || !this.cursorRef.current) {
       return
     }
 
-    this.readLogsFromLogStore()
+    let isAffectedByUpdate = false
```
nicks, line 224: hmmmm...this seems like a brittle optimization. is this really needed? AFAIR, the "append" event didn't need this, and the "truncate" event seems infrequent enough that it's not worth the effort.

web/src/LogStore.ts
```diff
@@ -59,7 +59,17 @@ class StoredLine {
...
 
-type callback = () => void
+export enum LogUpdateAction {
+  append,
+  truncate,
+}
+
+export interface LogUpdateEvent {
+  action: LogUpdateAction
+  spans: { [key: string]: Proto.webviewLogSpan }
```
nicks, line 69: what problem does this field solve? this seems like it's adding extra bookkeeping but not providing any real benefit.

web/src/logs.ts
```diff
@@ -50,3 +50,7 @@ export function sourcePrefix(n: string) {
 export function isBuildSpanId(spanId: string): boolean {
   return spanId.indexOf("build:") !== -1
 }
+
+export function isTiltfileSpanId(spanId: string): boolean {
```
nicks, line 54: i think this is dead code now

web/src/OverviewResourcePane.tsx
```diff
@@ -46,12 +48,37 @@ export default function OverviewResourcePane(props: OverviewResourcePaneProps) {
...
+    if (r) {
+      alerts = combinedAlerts(r, logStore)
+    } else if (all) {
+      resources.forEach((r) => alerts.push(...combinedAlerts(r, logStore)))
+    }
+    return alerts
   }
+  const [alerts, setAlerts] = useState<Alert[]>([])
+  // update alerts whenever the selected resource changes
+  useEffect(() => setAlerts(buildAlerts), [r])
```
nicks, line 62: I'm pretty confident this isn't going to work like you expect. You really don't want circular dependencies between hooks like this.

instead of making alerts a hook, maybe just create a dummy state variable?

```
const [trancateCount, setTruncateCount] = useState(0)

let alerts = ...
```

it's a little janky, but it should force the alerts to regenerate correctly each time the truncate count increments, which is closer to the data flow you want.

nicks (commented):
only blocking comment is the weird use of hooks

web/src/ClearLogs.tsx
```diff
@@ -0,0 +1,39 @@
+import React from "react"
+import styled from "styled-components"
+import { useLogStore } from "./LogStore"
+import { Color, FontSize } from "./style-helpers"
+
+const ClearLogsButton = styled.a`
```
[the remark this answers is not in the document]
milas: Ah `mixinResetButtonStyle` is probably what I was missing - I was fighting the button styles a lot and gave up. Thanks for the pointer.

web/src/OverviewResourcePane.tsx
milas: Hmm...not sure I totally understand how that would help here.

You're suggesting something like this?
```javascript
  const [truncateCount, setTruncateCount] = useState<number>(0)

  // add a listener to rebuild alerts whenever a truncation event occurs
  useEffect(() => {
    const rebuildAlertsOnLogClear = (e: LogUpdateEvent) => {
      if (e.action === LogUpdateAction.truncate) {
        setTruncateCount(truncateCount + 1)
      }
    }

    logStore.addUpdateListener(rebuildAlertsOnLogClear)
    return () => logStore.removeUpdateListener(rebuildAlertsOnLogClear)
  }, [])

  let alerts: Alert[] = []
  if (r) {
    alerts = combinedAlerts(r, logStore)
  } else if (all) {
    resources.forEach((r) => alerts.push(...combinedAlerts(r, logStore)))
  }
```
That won't actually force a re-render, unfortunately.

Because alerts need to react to changes in `LogStore`, they need to be in state so React can see the change and re-render. But as a result, that means that I need to handle props changes.

The "subscription" hook for `LogStore` has `[]` for deps, so it'll just subscribe on mount and unsubscribe on unmount.
The hook here is for props changes, so has `[r]` for deps, so it'll only run when the resource changes.

I believe this is correct use of hooks - setting state is generally safe as long as correctly guarded with dependencies to prevent an infinite re-render loop, and pretty common for subscription cases.

That said, one thought would be to modify the `LogStore` subscriptions (which might address some of the other feedback) so that you can pass a resource/manifest name on subscribe and only get callbacks if that one changes. Then, a single effect would suffice:
```javascript
  // add a listener to rebuild alerts whenever resource changes or a truncation event occurs
  useEffect(() => {
    const rebuildAlertsOnLogClear = (e: LogUpdateEvent) => {
      if (e.action === LogUpdateAction.truncate) {
        setAlerts(buildAlerts)
      }
    }
    
    setAlerts(buildAlerts)

    logStore.addUpdateListener(rebuildAlertsOnLogClear, r?.name)
    return () => logStore.removeUpdateListener(rebuildAlertsOnLogClear)
  }, [r])
```

nicks: > That won't actually force a re-render, unfortunately.

you made a mistake - you'll need to create a useEffect() dependency on truncateCount, because it reads truncateCount. it should work after that.

> Because alerts need to react to changes in LogStore, they need to be in state so React can see the change and re-render.

Yep, we agree this is the core issue. But you're not handling dependencies correctly. Here's a repro case you can try yourself:

1) add `cmd='exit 1` to the storybook resource in web/storybook.tiltfile
2) Run tilt up
3) Navigate to http://localhost:10350/r/(all)/overview
4) Click 'clear logs'
5) Click the reload button on the storybook item

Expected result:
The error count should be '1'

Actual result:
The error count is '0'

We can talk about how you could handle this correctly, but i think the approach i described (where the data is unidirectional) is going to be simpler to reason about.

web/src/OverviewLogPane.tsx
milas: You're right - it's unnecessary. I think at one point I had something else funky going on and thought I needed this. Got rid of it and dramatically simplified the logic here

web/src/LogStore.ts
milas: It was for in conjunction with conditionally triggering an update by the receiver but is unnecessary in the end (see other comment) - got rid of it and associated bookkeeping.

web/src/OverviewResourcePane.tsx
milas: D'oh - ok, everything works well with that approach (when I use dependencies properly 😝). Changes are in b8376cb, please unresolve if I did something else whacky!

web/src/logs.ts
milas: Away it goes!

web/src/OverviewActionBar.tsx
milas: No, this is a goofy but common abuse of JS logical operators -> https://reactjs.org/docs/conditional-rendering.html#inline-if-with-logical--operator - `false || expression` evaluates to `expression`, not a `boolean` type 😬

(JSX explicitly ignores `false` from an expression within element content to enable this use case)

If there's a different pattern used throughout Tilt UI, I can swap. Is `isSnapshot ? null : <ClearLogs ... />` preferable?

web/src/LogStore.ts
```diff
@@ -293,12 +368,20 @@ class LogStore {
     return this.logHelper(spans, 0).lines
   }
 
+  allSpans(): { [key: string]: LogSpan } {
+    const result: { [key: string]: LogSpan } = {}
+    for (let spanId in this.spans) {
+      result[spanId] = { ...this.spans[spanId] }
```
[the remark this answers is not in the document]
milas: Removed the clone - it wasn't strictly necessary, I just did it to avoid callers unintentionally mutating internal `LogStore` state. Not as much of a concern now that there's way less logic with the result of this now.

web/src/ClearLogs.tsx
```diff
@@ -0,0 +1,49 @@
...
+} from "./style-helpers"
+
+const ClearLogsButton = styled.button`
+  ${mixinResetButtonStyle}
+  margin-left: auto;
+  font-size: ${FontSize.small};
+  color: ${Color.white};
+
+  & {
+    transition: color ${AnimDuration.default} ease;
```
nicks, line 18: you can move this with the other properties... "&" doesn't really make sense as a selector, since it's just a self-reference

nicks (approved):

PR #4172, opened 2021-02-05
web: fix vertical overflow for build event fallback lines

milas:
Honestly not sure why this only impacts build event fallback.
I can't see any regression by removing this across other log
line types including long wrapping ones etc.

These messages are only visible with `--debug` anyway, so not
a big deal. There was always a slight truncation but it's much
more severe with the reduced font size.

### Before
Screen Shot 2021-02-05 at 3 49 47 PM

### After
Screen Shot 2021-02-05 at 3 56 52 PM

### DIFF
--- a/web/src/LogPaneLine.scss
+++ b/web/src/LogPaneLine.scss
@@ -4,7 +4,6 @@
 
 .LogPaneLine {
   display: flex;
-  position: relative;
   font-size: $font-size-smallest;  
 
   &.is-highlighted {


nicks:
i'm pretty sure this will break the stuff that's absolutely positioned relative to the line? cf https://github.com/tilt-dev/tilt/blob/master/web/src/LogPaneLine.scss#L38

milas:
You're absolutely right - I didn't have anything in that state at the time...back to the drawing board 😓

PR #4181, opened 2021-02-08
web: Add divider lines between sections in Overview

victorwuky:
https://app.clubhouse.io/windmill/story/11181/section-divider-line

# Without tests

Screen Shot 2021-02-08 at 7 12 47 PM

# With tests

Screen Shot 2021-02-08 at 7 12 12 PM

### DIFF
--- a/web/src/OverviewGrid.tsx
+++ b/web/src/OverviewGrid.tsx
@@ -19,10 +19,10 @@ let OverviewGridRoot = styled.ul`
   box-sizing: border-box;
   &:nth-child(even) {
     border-bottom: 1px dotted ${Color.grayLight};
-  }
+  };
   &:last-child {
     border-bottom-width: 0;
-  }
+  };
 `
 
 export default function OverviewGrid(props: OverviewGridProps) {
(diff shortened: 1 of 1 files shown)


victorwuky:
@SurbhiSGupta : What color should the divider line be? See https://github.com/tilt-dev/tilt/blob/master/web/src/constants.scss.

web/src/OverviewGrid.tsx
```diff
@@ -16,6 +16,12 @@ let OverviewGridRoot = styled.ul`
   padding-left: ${SizeUnit(0.25)};
   position: relative;
   box-sizing: border-box;
+  &:nth-child(even) {
```
nicks, line 19: i think you can remove the nth-child(even) selector?

the grid is always preceded by a divider, so this is currently a no-op (and makes the code more brittle)

nicks (approved):

web/src/OverviewGrid.tsx
victorwuky: Ah yes thanks @nicks . Fixing.

PR #4182, opened 2021-02-09
web: show 'alerts on top' toggle even when there are no tests

landism:
image

I'm not sure if it's poor form to simply set `visibility: hidden` rather than omit the checkboxes entirely - it's certainly vastly easier than changing a bunch of css to keep the 'alerts on top' toggle in place when the filter checkboxes are absent.

### DIFF
--- a/web/src/OverviewSidebarOptions.test.tsx
+++ b/web/src/OverviewSidebarOptions.test.tsx
@@ -103,7 +103,7 @@ describe("overview sidebar options", () => {
     assertSidebarItemsAndOptions(root, ["(Tiltfile)"], false, false, false)
   })
 
-  it("doesn't show SidebarOptionSetter if no tests present", () => {
+  it("doesn't show filter options if no tests present", () => {
     let items = [tiltfileResource(), oneResource()].map(
       (r) => new SidebarItem(r)
     )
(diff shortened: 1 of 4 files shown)


web/src/OverviewSidebarOptions.test.tsx
```diff
@@ -123,8 +124,8 @@ describe("overview sidebar options", () => {
     let sidebar = root.find(SidebarResources)
     expect(sidebar).toHaveLength(1)
 
-    let optSetter = sidebar.find(OverviewSidebarOptions)
-    expect(optSetter).toHaveLength(0)
+    let filters = sidebar.find(FilterOptionList)
```
maiamcc, line 127: nit: this test should be called "it doesn't show filter options if no tests present"

web/src/SidebarResources.tsx
```diff
@@ -156,9 +156,14 @@ export class SidebarResources extends React.Component<SidebarProps> {
             />
           </SidebarListSection>
           <PinnedItems {...this.props} />
-          {testsPresent && (
-            <OverviewSidebarOptions options={options} setOptions={setOptions} /> // TODO: if this vanishes because no tests present, reset it to show everything
-          )}
+          {
+            // TODO: if this vanishes because no tests present, reset it to show everything}
```
maiamcc, line 160: tbh i'd just take this comment out, we have it tracked in clubhouse

maiamcc (approved):
> I'm not sure if it's poor form to simply set `visibility: hidden` rather than omit the checkboxes entirely
haaa that is 100% a question for @hyu, but the code logic looks fine and i'd feel good about merging this and letting Han go back over the styling after if needful, cuz imho it looks good enough(tm).

PR #4191, opened 2021-02-11
web: Fix Sidebar item layout bugs

hyu:
Long resource names no longer break the layout!

| Before | After |
|---|---|
| Screen Shot 2021-02-04 at 5 52 12 PM | Screen Shot 2021-02-11 at 6 43 50 PM |

&nbsp;
Also, the pin position looks more intentional:

| Before | After |
|---|---|
|Screen Shot 2021-02-11 at 6 47 48 PM | Screen Shot 2021-02-11 at 6 48 06 PM|

### DIFF
--- a/web/src/OverviewItemView.tsx
+++ b/web/src/OverviewItemView.tsx
@@ -188,7 +188,7 @@ let RuntimeBoxStack = styled.div`
   flex-grow: 1;
   flex-shrink: 1;
   // To truncate long resource names…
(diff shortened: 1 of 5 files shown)


maiamcc:
Will look in a bit. To confirm @hyu , does the right margin look reasonable when the Auto toggle is there, too?

nicks (approved):

Issue #3909, opened 2020-11-10
one chatty resource can push out the logs of other resources


nicks:
### Expected Behavior
I should still be able to look at Tiltfile logs, even when one resource is super chatty

### Current Behavior
One chatty resource can push out the logs of other resources

#### About Your Use Case
Discovered this while trying to reproduce https://github.com/tilt-dev/tilt/issues/3908.

The problem is that we use a single max bytes limit for all logs, rather than truncating particular resources

PR #4202, opened 2021-02-16
logstore: truncate chatty resources before non-chatty resources

nicks:
Hello @landism, @milas,

Please review the following commits I made in branch nicks/ch11384:

0bdb3cee474bbe2b9b3dd9f2ccb977fe6bd12684 (2021-02-15 19:14:30 -0500)
logstore: truncate chatty resources before non-chatty resources
This is a simple algorithm to give priority to truncating resources
with lots of logs.

Fixes https://github.com/tilt-dev/tilt/issues/3909

### DIFF
--- a/pkg/model/logstore/logstore.go
+++ b/pkg/model/logstore/logstore.go
@@ -771,20 +771,66 @@ func (s *LogStore) ensureMaxLength() {
 		return
 	}
 
-	// Figure out where we have to truncate.
-	bytesSpent := 0
-	truncationIndex := -1
+	// First, count the number of bytes in each manifest.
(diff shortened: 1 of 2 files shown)


milas (approved):
Nice! Want to make the same(-ish) changes to `LogStore.ts` next? 😜

nicks:
wait, i changed my mind on this...i think we should track bytes per manifest instead of per span. one moment...

nicks:
updated! the code is almost exactly the same, it just uses a different key

milas (approved):

nicks:
and yep, this should be easy to adapt to the typescript too

PR #4207, opened 2021-02-16
tiltfile: add os.path.relpath

maiamcc:
this would have been useful for the [Tests In Tilt code snippets](https://docs.tilt.dev/tests_in_tilt.html#go-tests-by-package) i was writing for docs, so I implemented it

### DIFF
--- a/internal/tiltfile/os/os.go
+++ b/internal/tiltfile/os/os.go
@@ -169,10 +169,13 @@ func abspath(t *starlark.Thread, path string) (starlark.Value, error) {
 }
 
 func relpath(thread *starlark.Thread, fn *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
-	var basepath, targpath string
+	var targpath string
+	basepath := starkit.AbsWorkingDir(thread) // by default, relative to CWD
+
+	// Matches syntax of Python's os.path.relpath()
 	if err := starkit.UnpackArgs(thread, fn.Name(), args, kwargs,
-		"basepath", &basepath,
-		"targpath", &targpath); err != nil {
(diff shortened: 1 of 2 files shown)


nicks (approved):

internal/tiltfile/os/os.go
```diff
@@ -164,6 +168,17 @@ func abspath(t *starlark.Thread, path string) (starlark.Value, error) {
 	return starlark.String(starkit.AbsPath(t, path)), nil
 }
 
+func relpath(thread *starlark.Thread, fn *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
+	var basepath, targpath string
+	if err := starkit.UnpackArgs(thread, fn.Name(), args, kwargs,
+		"basepath", &basepath,
```
landism, line 174: might wanna swap the arg order to allow compatibility with python

https://docs.python.org/3/library/os.path.html#os.path.relpath

landism (approved):

internal/tiltfile/os/os.go
landism: and it looks like defaulting the basepath to cwd would even be good for the docs!

maiamcc: oh ugh you're right -- i did it this way to mirror the Go func but this makes much more sense 👍

victorwuky:
@maiamcc : Should we add an entry to https://docs.tilt.dev/api.html for this change?

maiamcc:
Yep, was waiting for the release to go out, just merged https://github.com/tilt-dev/tilt.build/pull/824

victorwuky:
Thanks @maiamcc

PR #4211, opened 2021-02-16
server: Merge an apiserver into the tilt webserver

nicks:
Hello @milas, @ellenkorbes,

Please review the following commits I made in branch nicks/ch11337:

161b3396b96a273f7d0b54c0409bda561ac16057 (2021-02-16 16:06:00 -0500)
server: Merge an apiserver into the tilt webserver
To test:
Run: make install
Run: tilt up -f web/storybook.tiltfile
Run: kubectl --kubeconfig=./scripts/kubeconfig get manifests

b17f57b4de19efba92c6fbe69b037472985f6623 (2021-02-16 16:05:04 -0500)
Autogenerated by kubernetes/code-generator
scripts/update-codegen.sh

caed76d5acbd3ae30daa4f9e698c00fed33bac92 (2021-02-16 16:04:29 -0500)
vendor: code generation

6eff3653906337c997bf0ca3982d9f3364dc425c (2021-02-16 14:21:55 -0500)
Autogenerated by apiserver-boot
apiserver-boot create group version resource --group core --version v1alpha1 --kind Manifest


nicks:
Most of this PR you've seen before. The big change is merging the API server and the web server in this commit:
https://github.com/tilt-dev/tilt/pull/4211/commits/8f90b8b4fe145149b55408a0f5ae5ee8b2a77918

rather than having them listen on different ports

internal/hud/server/controller.go
```diff
@@ -54,26 +62,67 @@ func (s *HeadsUpServerController) OnChange(ctx context.Context, st store.RStore)
...
-	go func() {
-		<-ctx.Done()
-		_ = httpServer.Shutdown(context.Background())
-	}()
+	server, err := config.Complete().New()
+	if err != nil {
+		return err
+	}
+
+	server.GenericAPIServer.AddPostStartHookOrDie("start-tilt-server-informers", func(context genericapiserver.PostStartHookContext) error {
```
milas, line 94: Not hugely important (I'm assuming this can't practically fail) but could use `AddPostStartHook` and propagate error to avoid surprise panic

milas (approved):

internal/hud/server/controller.go
nicks: ya, you're right on both counts - it really only fails if you've registered a conflicting hook, and it also seems safer to use the error-returning function. done!

Issue #4206, opened 2021-02-16
improve error messaging around live_update permission denied


nicks:
Currently, it's possible to shoot yourself in the foot and create a container that's impossible to live_update.

Here's an example of a user who had a security context in their Kubernetes YAML that made the file-system readonly, so that live-updates didn't work:

https://github.com/tilt-dev/tilt/issues/4200

Tilt should detect "permission denied" errors, and give pointers to possible fixes

PR #4213, opened 2021-02-16
containerupdate: improved error messaging when the file system isn't writable

nicks:
Hello @maiamcc, @landism,

Please review the following commits I made in branch nicks/ch11396:

63fa3751e92f147ee790ed4b289951ff45823a66 (2021-02-16 18:20:57 -0500)
containerupdate: improved error messaging when the file system isn't writable
Fixes https://github.com/tilt-dev/tilt/issues/4206

### DIFF
--- a/internal/containerupdate/exec_updater.go
+++ b/internal/containerupdate/exec_updater.go
@@ -1,6 +1,7 @@
 package containerupdate
 
 import (
+	"bytes"
 	"context"
 	"fmt"
 	"io"
@@ -28,7 +29,7 @@ func (cu *ExecUpdater) UpdateContainer(ctx context.Context, cInfo store.Containe
 	if !hotReload {
 		return fmt.Errorf("ExecUpdater does not support `restart_container()` step. If you ran Tilt " +
 			"with `--updateMode=exec`, omit this flag. If you are using a non-Docker container runtime, " +
-			"see https://github.com/tilt-dev/rerun-process-wrapper for a workaround")
+			"see https://github.com/tilt-dev/tilt-extensions/tree/master/restart_process for a workaround")
 	}
 
 	l := logger.Get(ctx)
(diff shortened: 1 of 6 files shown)


nicks:
Here's what the error looks like in context:
Screenshot from 2021-02-16 18-20-39

internal/containerupdate/exec_updater.go
```diff
@@ -64,3 +69,15 @@ func (cu *ExecUpdater) UpdateContainer(ctx context.Context, cInfo store.Containe
 
 	return nil
 }
+
+func handleK8sExecError(out *bytes.Buffer, err error) error {
+	msg := strings.ToLower(fmt.Sprintf("%s\n%s", out.String(), err.Error()))
+	if strings.Contains(msg, "permission denied") || strings.Contains(msg, "cannot open") {
+		return fmt.Errorf("denied access to the container filesystem. Please check:\n"+
```
landism, line 76: Not a blocker and this is a big improvement over status quo, so feel free to ignore.

I found the sample output a bit confusing:
1) The error doesn't show what was interpreted as "denied access" - I found it confusing to read "denied access to the container filesystem" and then the evidence presented is "original error: command terminated with exit code 2"
2) I really expected the text following "with unexpected error:" to be the error. Can we move the error to the start of this or change the colon to a period?

internal/containerupdate/exec_updater.go
```diff
@@ -64,3 +69,15 @@ func (cu *ExecUpdater) UpdateContainer(ctx context.Context, cInfo store.Containe
...
+			"  1) That the container image has a writable file system\n"+
+			"  2) That the container image default user has write access to the file system\n"+
```
landism, line 78: The output in the screenshot appears to be double-spaced, but that doesn't look to be the intent here. Not a blocker, just a bit eyebrow-raising.

landism (approved):

internal/containerupdate/exec_updater.go
nicks: ya, i believe that's a feature of the fallback ui, i'm not entirely sure what the thinking behind this is /cc @hyu

internal/containerupdate/exec_updater.go
nicks: ya, i think this is more of a complaint about the way we assemble live-update errors - it's trying to combine errors from a lot of different sources and make sense of them, while also trying not to hide anything from the user. not sure what the right answer is here.

nicks: here's what it looks like now

Screenshot from 2021-02-17 12-04-11

internal/containerupdate/exec_updater.go
hyu: This design is from a while ago, but I _think_ the intent was to make it very obvious that fallback was happening, and perhaps there were also some bad assumptions about what fallbacks would look like (e.g., not multi-line).

Re-evaluating this now, I think we should just make the fallback message have the same spacing as everything else.

nicks: adjusting the spacing a bit more, here's what it looks like now:

Screenshot from 2021-02-18 09-57-05

PR #4221, opened 2021-02-18
web: refine Toggle Mode button design

hyu:
- The Toggle Mode button now only appears on hover, making the cards appear less cluttered by default
- Manual now looks highlighted, while Auto looks more neutral. (Previously, it was the other way around.) This reflects the current reality that auto updating is default, unless you explicitly set a resource to trigger mode manual. 
- The Trigger button turns blue (matching the Trigger Mode toggle) to indicate it's in Manual trigger mode.

Screen Shot 2021-02-18 at 1 01 57 PM


web/src/OverviewItemView.tsx
```diff
@@ -190,42 +194,28 @@ let RuntimeBoxStack = styled.div`
...
-  align-items: stretch;
-  flex-shrink: 1;
-  border-top: 1px solid ${Color.grayLighter};
-`
-
-let OverviewItemText = styled.div`
+let OverviewItemType = styled.div`
   display: flex;
   align-items: center;
-  white-space: nowrap;
```
hyu, line 214: When I moved the Trigger + Trigger Mode buttons in the card layout, I checked text truncation again to make sure it was working properly. Along the way, I did a little refactoring of the styles related to various forms of text on the card (Resource Type, Resource Name, Build status).

I noticed some old CSS rules that made it look as though Type + Build text would be truncated. But it didn't actually work. But it kind of doesn't matter anyway because we control that text, and there's no option that's so long that it requires truncation. At least, I don't think so. (I didn't test all the variants extensively, I admit)

Anyway, since truncation wasn't working anyway, I made it more obvious that Resource Name is the only thing that's actively truncated, and tested it to make sure it works.

maiamcc (approved):
> Manual now looks highlighted, while Auto looks more neutral. (Previously, it was the other way around.) This reflects the current reality that auto updating is default, unless you explicitly set a resource to trigger mode manual.

i'm a liiiittle skeptical of how this will feel in practice, but willing to give it a shot! let's just make sure to pay attn to user feedback on this.

> The Toggle Mode button now only appears on hover, making the cards appear less cluttered by default

likewise, also skeptical-but-not-too-skeptical (user can no longer tell at a glance what tests they do and don't expect to run automatically) but let's try it and see

PR #4222, opened 2021-02-18
web: fix broken nav for resources with URL unsafe names

### LINKED ISSUE #4026: When a resource name has a slash in it, clicking on it in the webui sidebar doesn't work
### Steps to Reproduce
Make a resource with a slash in the name, e.g.:
```
local_resource('a/b', 'echo hi')
```
`tilt up`, open the web ui, and click on the "a/b" resource in the sidebar.

### Expected Behavior
Tilt shows me the logs for the "a/b" resource.

### Current Behavior
Tilt shows me "No match for /r/a/b"

This is presumably a matter of uri-encoding the resource name in the sidebar link, and the link should be to `/r/a%2Fb` instead of `/r/a/b/`

milas:
This just URI encodes the name - you'll get some ugly URLs
if your resources have crazy names, but c'est la vie.

For example:
	`foo` -> `/r/foo/overview`
	`a/b/c` -> `/r/a%2Fb%2Fc/overview`
	`🎉` -> `/r/🎉/overview`
	`alert("test")` -> `/r/alert("test")/overview`

Screen Shot 2021-02-18 at 1 17 10 PM

Closes #4026.

### DIFF
--- a/web/src/HUD.tsx
+++ b/web/src/HUD.tsx
@@ -576,7 +576,7 @@ export default class HUD extends Component<HudProps, HudState> {
     let tiltCloudTeamName = view?.tiltCloudTeamName || null
     let isSnapshot = this.pathBuilder.isSnapshot()
     let sidebarRoute = (t: ResourceView, props: RouteComponentProps<any>) => {
-      let name = props.match.params.name
+      const name = decodeURIComponent(props.match.params?.name ?? "")
       return (
         <Sidebar isClosed={isSidebarClosed} toggleSidebar={this.toggleSidebar}>
           <SidebarAccount
@@ -630,7 +630,7 @@ export default class HUD extends Component<HudProps, HudState> {
     let isSnapshot = this.pathBuilder.isSnapshot()
 
     let traceRoute = (props: RouteComponentProps<any>) => {
-      let name = props.match.params?.name ?? ""
+      const name = decodeURIComponent(props.match.params?.name ?? "")
       let span = props.match.params?.span ?? ""
 
       let r = resources.find((r) => r.name === name)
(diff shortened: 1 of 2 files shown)


milas:
FYI The new UI is actually much better with how it handles nav, so there's actually only 2x places in `TabNav.jsx` (one on open, one on nav match) that need to be updated. All the others are just for the old UI, which is why I didn't bother to try to make logic more reusable there.

Restricting character set as proposed in #4026 is still a good idea IMO, but probably better to URI encode regardless.

landism:
> All the others are just for the old UI, which is why I didn't bother to try to make logic more reusable there.

Ugh, yeah, it's frustrating how hard it is to make sure this is solved consistently.
I'm not entirely sure what's obsolete in the new UI (like is /r/$name/trace accessible?) but these seem like new UI?
https://github.com/tilt-dev/tilt/blob/16041acfe27080ccd5c1815334cd7b035cab4144/web/src/OverviewTabBar.tsx#L126
https://github.com/tilt-dev/tilt/blob/791aee9c8930d86fbd86795f4fd530a96e7503e5/web/src/OverviewLogPane.tsx#L574

milas:
Ah, good catch on those other places. Okay, I think this probably _is_ best handled by just forcing URL safe names for resources in the first place - so much for a quick and easy bugfix 😛 

(Only reason I started to poke at it in JS was potential concern over security issues, but everything is properly escaped in actual rendered HTML, it's just a matcher issue with the router)

landism:
theoretically react router should be automatically decoding matches, e.g.: https://github.com/ReactTraining/react-router/commit/7279d22276814d3d6ad2c799a9400e704e33b4b6

I spent half an hour and couldn't figure out why it wasn't happening for us

landism:
> I spent half an hour and couldn't figure out why it wasn't happening for us

Ah, it turns out they later undid this behavior after users with un-encoded params filed bugs because their params were getting incorrectly decoded.

PR #4226, opened 2021-02-19
tiltfile: send analytics for extensions [ch1142]

maiamcc:
this feels kind of weird but is the best implementation i could think of that matches the generic-ness of `LoadInterceptors` (i.e. if in the future we have other different kinds of `LoadInterceptors` that want to send data back to analytics, we can pretty easily)

alternately i could
- wire the `analytics` object through to the Extension itself and record the info at the point at which we parse the extension path, which seems awful, or
- store the relevant info on the Extension and type check the `LoadInterceptors` to see if any are a `tiltextensions.Extension`, and THEN pull the data off, which seems counter to the pretty generic way `LoadInterceptors` are build right now (of course I don't precisely understand WHY the current generic-ness, so... ¯\_(ツ)_/¯ )
- ??

### DIFF
--- a/internal/tiltfile/starkit/load_interceptor.go
+++ b/internal/tiltfile/starkit/load_interceptor.go
@@ -11,4 +11,7 @@ type LoadInterceptor interface {
 	// Ensure the content is present in the path returned
 	// Returns "" if this interceptor doesn't act on this path
 	LocalPath(t *starlark.Thread, path string) (string, error)
+
+	// Optionally, a LoadInterceptor can record some info to send to analytics later
+	AnalyticsInfo() []AnalyticsInfo
 }
--- a/internal/tiltfile/tiltextension/extension.go
+++ b/internal/tiltfile/tiltextension/extension.go
@@ -7,6 +7,7 @@ package tiltextension
 import (
 	"context"
 	"os"
+	"path/filepath"
 	"strings"
 
 	"go.starlark.net/starlark"
@@ -20,6 +21,11 @@ type Extension struct {
 	info    starkit.ExtensionsAnalyticsInfo
 }
 
+type Fetcher interface {
+	Fetch(ctx context.Context, moduleName string) (ModuleContents, error)
+	CleanUp() error
+}
+
 func NewExtension(fetcher Fetcher, store Store) *Extension {
 	return &Extension{
 		fetcher: fetcher,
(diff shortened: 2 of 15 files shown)


internal/tiltfile/starkit/load_interceptor.go
```diff
@@ -11,4 +11,7 @@ type LoadInterceptor interface {
 	// Ensure the content is present in the path returned
 	// Returns "" if this interceptor doesn't act on this path
 	LocalPath(t *starlark.Thread, path string) (string, error)
+
+	// Optionally, a LoadInterceptor can record some info to send to analytics later
+	AnalyticsInfo() []AnalyticsInfo
```
nicks, line 16: ya, this feels like a lot of layers of indirection

have you considered using the existing data model pattern for collecting this data?
https://github.com/tilt-dev/tilt/blob/master/internal/tiltfile/starkit/model.go

we have pretty good conventions for adding things to that data model and pulling it off at the end of execution in ways that prevent common mistakes.

nicks (commented):
left a few high level comments!

internal/tiltfile/tiltfile_state.go
```diff
@@ -205,7 +202,7 @@ func (s *tiltfileState) loadManifests(absFilename string, userConfigState model.
 		shlex.NewExtension(),
 		watch.NewExtension(),
 		loaddynamic.NewExtension(),
-		tiltextension.NewExtension(fetcher, tiltextension.NewLocalStore(filepath.Dir(absFilename))),
```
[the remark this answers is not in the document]
maiamcc: ah, I just made it so i could dependency inject a fake version into unit tests, that's a super fair point and we should probably not do that.

internal/tiltfile/tiltextension/extension.go
```diff
@@ -66,8 +85,15 @@ func (e *Extension) LocalPath(t *starlark.Thread, arg string) (string, error) {
 		_ = e.fetcher.CleanUp()
 	}()
 
+	// Record that we loaded this extension
+	e.info.ExtensionsLoaded[moduleName] = true
```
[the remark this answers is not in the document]
maiamcc: whoops! put the line in the wrong place.

internal/tiltfile/starkit/load_interceptor.go
maiamcc: i honestly completely forgot about that part of tiltfile world. is https://github.com/tilt-dev/tilt/pull/4229 more like it? (def feels much better to me)

nicks: :+1:

PR #4229, opened 2021-02-19
tiltfile: send analytics on extensions loaded (take 2)

maiamcc:
Hello @nicks, @landism,

here's a much cleaner version of #4226, tests tk but wanted to put something up before the weekend and get eyes on it.

### DIFF
--- a/internal/tiltfile/tiltextension/extension.go
+++ b/internal/tiltfile/tiltextension/extension.go
@@ -28,12 +28,12 @@ func NewExtension(fetcher Fetcher, store Store) *Extension {
 	}
 }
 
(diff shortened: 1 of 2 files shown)


internal/tiltfile/tiltextension/extension.go
```diff
@@ -26,6 +28,16 @@ func NewExtension(fetcher Fetcher, store Store) *Extension {
 	}
 }
 
+type ExtState struct {
```
nicks, line 31: maybe just "State", since it will usually be prefixed by the package name?

nicks (approved):

PR #4246, opened 2021-02-25
server: add controller manager

milas:
Currently there are no controllers managed, but this provides the
initialization logic and some `wire` basics.

We will probably want to examine how to best handle DI for the
actual controllers given how they interact with the manager for
initialization.


internal/controllers/manager.go
```diff
@@ -0,0 +1,84 @@
...
+type TiltServerControllerManager struct {
+	config *rest.Config
+	cancel context.CancelFunc
+}
+
+var _ store.SetUpper = &TiltServerControllerManager{}
+var _ store.Subscriber = &TiltServerControllerManager{}
+var _ store.TearDowner = &TiltServerControllerManager{}
+
+func ProvideRESTConfig(apiserverHost model.WebHost, apiserverPort model.WebPort) *rest.Config {
```
milas, line 29: This can/should be replaced by just injecting `server.APIServerConfig` into the controller manager

milas:
This is failing integration tests because the API server hasn't finished booting by the time the controller manager starts and it evidently doesn't retry.

I think the best course of action is to wait for #4244 to be merged, and inject an API client here and poll (up to a timeout) until it's ready? Thoughts, @nicks? Is there a better way to know the apiserver is up + healthy?

nicks:
re: "This is failing integration tests because the API server hasn't finished booting by the time the controller manager starts and it evidently doesn't retry."

If that was true, then how would the tests in #4244 pass?

milas:
It was getting back junk data, but it looks like the issue might actually have been that I accidentally rebased out the special handling of `/api` - once added back from merging in `master` with the other changes and swapping to `server.APIServerConfig`, things are looking better locally (will see how CI does now)
https://github.com/tilt-dev/tilt/blob/f2ca16ba5e2b3abc228e8325fed4323112422887/internal/hud/server/controller.go#L101

nicks:
i think this PR will help https://github.com/tilt-dev/tilt/pull/4248 because it will do everything in memory if it doesn't have a port

internal/engine/upper.go
```diff
@@ -184,6 +185,8 @@ func upperReducerFn(ctx context.Context, state *store.EngineState, action store.
 		handleMetricsDashboardAction(state, action)
 	case server.OverrideTriggerModeAction:
 		handleOverrideTriggerModeAction(ctx, state, action)
+	case apiserver.SyncAction:
```
nicks, line 188: is this still used?

internal/controllers/manager.go
```diff
@@ -0,0 +1,84 @@
...
+
+	// TODO(milas): we should provide a logr.Logger facade for our logger rather than using zap
+	w := logger.Get(ctx).Writer(logger.DebugLvl)
+	ctrl.SetLogger(zap.New(zap.UseDevMode(true), zap.WriteTo(w)))
+
+	utilruntime.Must(corev1alpha1.AddToScheme(scheme))
+
+	mgr, err := ctrl.NewManager(m.config, ctrl.Options{
+		Scheme:           scheme,
+		Port:             9443,
```
nicks, line 60: this might be a dumb question, but what happens if we set Port to 0? not totally clear why the manager needs a server

nicks: in particular, will hard-coding a port here break tilt if you try to run two tilt instances at the same time? (this is a common thing people do)

milas: I was just looking through that this morning to understand (and see if it's necessary/possible to share the listener with the HUD/apiserver).

It's only used for admission webhooks, and thankfully the webhook server is lazily started only when a webhook is registered with the manager. I'll omit the definition, and we can re-add in the future if we end up using webhooks.

(If we _do_ end up using it, it unfortunately just defaults to `9443` if unspecified/`0`, so we'd need to add options for it.)

nicks (approved):

internal/engine/upper.go
milas: Ah, nope - was left over from POC + ensuring import cycle insanity was resolved. Removed!

milas:
FYI merged latest from `master` and dropped the "don't actually initialize if no config" logic and everything looks good using the in-memory connection 🎉

internal/controllers/manager.go
milas: I changed this to `-1` with an explanation so that in the event someone adds a webhook it'll error out during initialization and we can deal with adding a second port + flag/env vars etc at that point.

internal/controllers/manager.go
milas: Made this change in 81f95d52c1a61c5221aa4cd35a5ec6d77996a12b and got rid of `ProvideRESTConfig()`

nicks (approved):

Issue #3993, opened 2020-12-14
k8s_resource.objects param doesn't let you distinguish btwn two k8s entities that differ only by group


maiamcc:
### Expected Behavior
`k8s_resource.objects` param should let you specify any k8s entity that Tilt knows about

### Current Behavior
In the case where Tilt has two k8s entities that differ only by group, e.g.
* `foobar:pod:my-ns:apiextensions.k8s.io:0`
* `foobar:pod:my-ns:apiextensions.k8s.io:1`

There is no way to disambiguate them via `k8s_resource.objects` (since that param. only accepts selectors of the form `name:kind:namespace`--otherwise you get the error "Too many parts in selector. Selectors must contain between 1 and 3 parts (colon separated)"). If you pass `foobar:pod:my-ns`, you get:
> "foobar:pod:my-ns" is not a unique fragment. Objects that match "foobar:pod:my-ns" are "foobar:pod:my-ns", "foobar:pod:my-ns"

### Steps to Reproduce
1. Tiltfile:
    ```
    load('ext://helm_remote', 'helm_remote')
    
    REPO_URL = 'https://containous.github.io/traefik-helm-chart'
    OBJ = 'tlsoptions.traefik.containo.us:CustomResourceDefinition:default'
    helm_remote('traefik', repo_name='traefik',
                repo_url=REPO_URL,
                namespace='default', version='8.9.0')
    k8s_resource('traefik', port_forwards='8000',
            objects=[OBJ]
    )
    ````
2. `tilt up`
3. See error ([snapshot](https://cloud.tilt.dev/snapshot/AdDdvP0LZHzNSEX_08Y=)):
    > "tlsoptions.traefik.containo.us:CustomResourceDefinition:default" is not a unique fragment. Objects that match "tlsoptions.traefik.containo.us:CustomResourceDefinition:default" are "tlsoptions.traefik.containo.us:CustomResourceDefinition:default", "tlsoptions.traefik.containo.us:CustomResourceDefinition:default"

#### About Your Use Case
[Reported by Isaac Wilder in Slack](https://kubernetes.slack.com/archives/CESBL84MV/p1607956584122500). Minor issue/not blocking.

nicks:
re: "only by group"...those two entities look like the same group? i.e., the group is apiextensions.k8s.io, and they're just two instances of the same object.

wondering if this is the same as https://github.com/tilt-dev/tilt/issues/3727

maiamcc:
[...]

PR #4249, opened 2021-02-25
k8s: treat namespace:  = namespace: default when determining if k8s entities are duplicates

maiamcc:
was taking a second look at #3993 and @landism pointed out that duplicate entities shouldn't have been allowed in the first place, so I went digging and found this bug.

### DIFF
--- a/internal/k8s/entity.go
+++ b/internal/k8s/entity.go
@@ -89,7 +89,7 @@ func (e K8sEntity) ToObjectReference() v1.ObjectReference {
 		Kind:       kind,
 		APIVersion: apiVersion,
 		Name:       meta.GetName(),
-		Namespace:  meta.GetNamespace(),
+		Namespace:  e.Namespace().String(), // so that we normalize "" --> "default"
 		UID:        meta.GetUID(),
 	}
 }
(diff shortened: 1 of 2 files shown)


internal/k8s/entity.go
```diff
@@ -89,7 +89,7 @@ func (e K8sEntity) ToObjectReference() v1.ObjectReference {
 		Kind:       kind,
 		APIVersion: apiVersion,
 		Name:       meta.GetName(),
-		Namespace:  meta.GetNamespace(),
```
nicks, line 92: hmmm...i don't think this is accurate?

empty namespace != "default", it's equal to the kube context namespace...

maiamcc: augh, so the comment is wrong, i meant it normalized to _the default_ rather than "default", but I foolishly thought that `k8s.DefaultNamespace` was the default for your kubecontext whoops. If we don't have a good way to a. figure out and b. access the default ns for your context, then ugh. (and unless there's a really easy answer, I suspect this isn't worth worrying about right now.)

nicks: ya, the DefaultNamespace was a mistake and is something i've been meaning to remove :eyes:

maiamcc:
We don't have a good concept of "default namespace for the current kubecontext" (the best we can do is the namespace "default"), closing this because it's more trouble than it's worth to do properly.

PR #4254, opened 2021-02-26
apis: fork probe data model

nicks:
Hello @milas,

Please review the following commits I made in branch nicks/probe:

e34d60fcd444c8e817daeb02366f574dfb782c11 (2021-02-26 13:19:19 -0500)
apis: fork probe data model


nicks:
The only difference between this and the upstream data model is that our Probes can't have named ports, only int ports

milas (approved):

PR #4257, opened 2021-03-01
controllers: add a cmd controller

nicks:
Hello @milas,

Please review the following commits I made in branch nicks/cmdscaffold:

e8ef063338ccd1d0e4fa9d6857c6a14d6e1e8cb4 (2021-03-01 11:22:57 -0500)
controllers: add a cmd controller

### DIFF
--- a/internal/controllers/core/cmd/wire.go
+++ b/internal/controllers/core/cmd/wire.go
@@ -1,0 +1,7 @@
+package cmd
+
+import "github.com/google/wire"
+
+var WireSet = wire.NewSet(
+	NewController,
+)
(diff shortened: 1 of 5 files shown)


internal/controllers/core/cmd/wire.go
```diff
@@ -0,0 +1,7 @@
+package cmd
+
+import "github.com/google/wire"
+
+var WireSet = wire.NewSet(
```
milas, line 5: I don't think this actually being used currently, but I do like this pattern since make the top level wire cleaner:
```go
var controllerSet = wire.NewSet(
    filewatch.WireSet,
    cmd.WireSet,
    ...
)
```
(I'm assuming controllers will accumulate various sub-dependencies beyond their own ctor)

milas (approved):

internal/controllers/core/cmd/wire.go
nicks: oops, fixed. yep, there will be other deps

PR #4262, opened 2021-03-01
server: FileWatch apiserver controller

milas:
This mostly leaves the existing file watch code untouched aside from
some light refactoring to share more code during the transition.
For now, however, the API server file watches are independent and
not currently auto-generated from manifests nor do they trigger
dispatches back to the Tilt store.

In its current state, this will work as a POC for other objects in
the API server model to watch a `FileWatch` and react to e.g. rebuild
on a file change.

Error handling probably needs some improvement; in particular there's
a big question of how to handle FS watch errors - today, Tilt largely
considers them fatal and dispatches a store error action. The API
server model is experimenting with exposing errors on the status for
the `FileWatch` object.

Next steps are to start splitting apart the Tilt store to accommodate a
cleaner separation between file watches + build targets and integrate
the controller with it.


milas:
I'm still working on tests but wanted to put this up as a functional checkpoint to help with other controller development.

nicks:
ya, i'm also trying to figure out how to structure the tests. surprisingly the real controller tests use etcd!

pkg/clientset/tiltapi/wire.go
```diff
@@ -0,0 +1,19 @@
...
+
+import (
+	"fmt"
+
+	"k8s.io/client-go/rest"
+
+	"github.com/tilt-dev/tilt/pkg/model"
+)
+
+func ProvideRESTConfig(apiserverHost model.WebHost, apiserverPort model.WebPort) *rest.Config {
```
nicks, line 11: have you seen https://github.com/tilt-dev/tilt/blob/master/internal/hud/server/apiserver.go#L78 ? it also works for cases where there's an in-memory connection

pkg/clientset/tiltapi/wire.go
```diff
@@ -0,0 +1,19 @@
+package tiltapi
```
nicks, line 1: i'm a little nervous about mixing non-generated code with generated packages

pkg/apis/core/v1alpha1/filewatch_types.go
```diff
@@ -53,6 +53,12 @@ type FileWatchList struct {
 
 // FileWatchSpec defines the desired state of FileWatch
 type FileWatchSpec struct {
+	RootPath string `json:"rootPath"`
+
+	Paths []string `json:"matchPatterns,omitempty"`
+
+	// +optional
+	IgnorePatterns []string `json:"ignorePatterns,omitempty"`
```
nicks, line 61: i feel like we often need to model sets of ignore patterns with different roots? e.g., a .dockerignore and a tiltfile-generated ignores. is your plan to normalize them to the same root path?

internal/controllers/core/filewatch/controller.go
```diff
@@ -15,31 +15,93 @@ package filewatch
...
+		// monitor is removed upon deletion
+		if !controllerutil.ContainsFinalizer(&fileWatchApiObj, fsFileWatchFinalizer) {
+			controllerutil.AddFinalizer(&fileWatchApiObj, fsFileWatchFinalizer)
+			if err := r.Update(ctx, &fileWatchApiObj); err != nil {
+				return ctrl.Result{}, err
+			}
+		}
+	} else {
+		if controllerutil.ContainsFinalizer(&fileWatchApiObj, fsFileWatchFinalizer) {
+			if err := r.WatchManager.StopWatch(fileWatchApiObj.Name); err != nil {
```
nicks, line 67: i wonder if finalizers are overkill for what we're doing here...they're designed for external resources that persist even if the controller server reboots, but that's obviously not the case with this particular controller.

internal/controllers/core/filewatch/controller.go
```diff
@@ -15,31 +15,93 @@ package filewatch
...
 }
 
 // +kubebuilder:rbac:groups=,resources=filewatches,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=,resources=filewatches/status,verbs=get;update;patch
 // +kubebuilder:rbac:groups=,resources=filewatches/finalizers,verbs=update
 
-func (r *Controller) Reconcile(_ context.Context, _ ctrl.Request) (ctrl.Result, error) {
-	// this is currently a no-op stub
+func (r *Controller) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	log := logger.Get(ctx).WithFields(logger.Fields{"apiserver_entity": req.NamespacedName.String()})
```
nicks, line 48: fwiw, I've been putting the manifest name as a label on the object, and using that to construct a logger that goes through the normal tilt logging flow...don't want to reinvent everything at once...

internal/controllers/core/filewatch/watch_manager.go
```diff
@@ -0,0 +1,237 @@
...
+		notifier: notifier,
+		cancel:   cancel,
+	}
+
+	go m.notifyLoop(watchCtx, name, notifier)
+
+	// the old watch is cleaned up AFTER starting the new one to avoid missing events
+	// (this might mean duplicates are received, which is an acceptable trade-off)
+	if oldWatch.cancel != nil {
+		oldWatch.cancel()
```
nicks, line 121: it seems odd to me that this implementation looks different from StopWatch?

nicks (commented):
some high-level questions

pkg/apis/core/v1alpha1/filewatch_types.go
```diff
@@ -99,6 +105,15 @@ func (in *FileWatchList) GetListMeta() *metav1.ListMeta {
 
 // FileWatchStatus defines the observed state of FileWatch
 type FileWatchStatus struct {
+	LastEventTime *metav1.Time `json:"lastEventTime,omitempty"`
+	SeenFiles     []string     `json:"seenFiles,omitempty"`
+	ErrorMessage  string       `json:"errorMessage"`
+	// FileEvents []FileEvent `json:"fileEvents,omitempty"`
+}
+
+type FileEvent struct {
```
nicks, line 114: is this used?

pkg/clientset/tiltapi/wire.go
milas: Ah yep, this is vestigial - this code has gone through one too many rebases 😝

pkg/clientset/tiltapi/wire.go
milas: Agreed - thankfully this can actually just ago away entirely now in this case (see previous comment)

internal/controllers/core/filewatch/controller.go
milas: That's fair, without a finalizer, it could be handled when the object isn't found earlier in `Reconcile()`, especially since we intentionally _don't_ want to block actual deletion even if something goes awry with the file watcher.

pkg/apis/core/v1alpha1/filewatch_types.go
milas: I originally had the impression that the base path was always the same, perhaps something like this makes more sense?

```go
type PatternSet struct {
   BasePath string
   Patterns []string
}

type FileWatchSpec struct {
   Watches []PatternSet
   Ignores []PatternSet
}
```

Then the controller could create a set of absolute path patterns from the watches for the actual FS watcher logic and a `CompositeMatcher` from the various ignores?

Taking things up one level - I _think_ it makes sense for each `FileWatch` to be self-contained, but I'm still not 100% sure what/how the flow for propagating global ignores should be. (cc @ellenkorbes)

internal/controllers/core/filewatch/watch_manager.go
milas: Cleaned this up - new logic is consistent and a bit paranoid to balance not leaking resources but also avoid any weird races

pkg/apis/core/v1alpha1/filewatch_types.go
milas: No - I started down this path and then removed it to favor data model simplicity, but I think it _might_ end up being necessary so that consumers watching can get "files changed since" semantics. Deleted it for now, we can reintroduce if the need arises once the controller is being used

pkg/apis/core/v1alpha1/filewatch_types.go
nicks: ya, that should work

nicks: fwiw, most build systems don't support arbitrary Watch/Ignore patterns, because it can lead to pathological performance characteristics if you're not careful. 

Most things in Tilt let you watch a directory (recursively) and then set up patterns to ignore (with some constraints on those patterns), but even this has some (intentional) bugs to avoid perf regressions, see https://github.com/moby/moby/issues/30018

milas: Okay, maybe this which avoids the terminology ambiguity and some repetition for simplistic cases:
```go
type WatchSet struct {
   BasePath string
   Paths []string
   IgnorePatterns []string
}

type FileWatchSpec {
   WatchSets []WatchSet
}
```

```yaml
spec:
  watchSets:
    - basePath: "/a"
      paths: ["./b", "c"]
      ignores: ["c/d"]
    - basePath: "/a/b"
      ignores: ["zzz"]
```

Same logic applies - create absolute paths for all the watched paths and a `CompositeMatcher` from the various ignores

milas: New version uses this, I like it so far? Will update GDoc

internal/controllers/core/filewatch/controller.go
milas: Got rid of the finalizer along with dramatically simplifying implementation

milas:
Closing in lieu of #4277 - some of this code will return in the future though

PR #4265, opened 2021-03-02
test: add gomega async assertion

nicks:
Hello @maiamcc, @landism,

Please review the following commits I made in branch nicks/wait:

2366c1f18fe01a198e6fcd5bd5d91c0723538e78 (2021-03-02 15:48:20 -0500)
test: add gomega async assertion


nicks:
a lot of the Kubernetes sample code uses this assertion framework. it's a bit richer than testify and has support for async matchers, so i played around with it a bit. wdyt?
http://onsi.github.io/gomega/

nicks:
actually, i convinced myself that we shouldn't do this, and should just use testify

PR #4268, opened 2021-03-03
tiltfile: change tag key for extensions metric so our analytics system imports it correctly [ch11508]

maiamcc:
@nicks is this all that's needed? (I would verify but idk how to force a data migration job--would be happy to learn if you have time to show me.)


nicks (approved):

nicks:
you could probably verify with https://github.com/tilt-dev/tilt/blob/master/integration/analytics_test.go (which uses a real stats endpoint) but i wouldn't worry about it

PR #4271, opened 2021-03-03
web: rm ResourceInfo

landism:
there were used by the old web ui and are now dead code

(which I only discovered after being very confused that my changes weren't taking effect :()


nicks (approved):
wooooo

PR #4267, opened 2021-03-03
local: demonstrate how to integrate the new controller with tilt

nicks:
Hello @milas, @ellenkorbes,

Please review the following commits I made in branch nicks/localcommands6:

6a678fe673bde96367b40fb4a1b89829dcf7e85d (2021-03-03 11:28:34 -0500)
local: demonstrate how to integrate the new controller with tilt

821ede97e96a6ea3c3649f84f61f2b7096276e6f (2021-03-03 11:27:02 -0500)
cmd: add a local command controller

f384e12d1a2de7f0646c93842141bdbc018dba1d (2021-03-03 11:23:19 -0500)
server: use unbuffered memory connections for testing, which seem to be more robust

Code review reminders, by giving a LGTM you attest that:

* Commits are adequately tested
* Code is easy to understand and conforms to style guides
* Incomplete code is marked with TODOs
* Code is suitably instrumented with logging and metrics

PR #4272, opened 2021-03-03
engine: Write Cmd objects to the EngineState

nicks:
Hello @milas, @ellenkorbes,

Please review the following commits I made in branch nicks/action-time:

8f6df154b2785d368e945a1b97aedcc22c3bcab1 (2021-03-03 18:34:08 -0500)
engine: Write Cmd objects to the EngineState
This is alternative proposal for how to move to the new API,
where we write the objects to the engine state first,
then convert to a controller second.

This also contains some standardization w/r/t how we do logging
in this new system.

7c142e686ed75cbddfe345c422e15b703bdbee18 (2021-03-03 17:02:23 -0500)
apis: add reason to CmdStatus, remove manifest

### DIFF
--- a/internal/engine/local/actions.go
+++ b/internal/engine/local/actions.go
@@ -1,21 +1,21 @@
 package local
 
-import (
-	"github.com/tilt-dev/tilt/pkg/model"
-)
-
-type LocalServeStatusAction struct {
-	ManifestName model.ManifestName
-	Status       model.RuntimeStatus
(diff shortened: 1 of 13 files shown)


nicks:
This is an alternative proposal to https://github.com/tilt-dev/tilt/pull/4267 on how to do the migration, where i think it will be a lot simpler to see what's going on and parallelize the conversion

milas (approved):

PR #4276, opened 2021-03-04
store: add a mechanism to summarize changes

nicks:
Hello @maiamcc, @landism, @milas,

Please review the following commits I made in branch nicks/summarize:

322e26319790773e3b7badb6726669dc9e77ccaf (2021-03-04 18:20:51 -0500)
store: add a mechanism to summarize changes
The goal of this is to help us move more towards a system like
Kubernetes (https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.8.2/pkg/reconcile#Request)
where the system tells us which object is changing, rather
than writing lots of code to figure it out on our own.

### DIFF
--- a/internal/cloud/cloud_status_manager.go
+++ b/internal/cloud/cloud_status_manager.go
@@ -170,7 +170,7 @@ func (c *CloudStatusManager) needsLookup(requestKey statusRequestKey) bool {
 		requestKey != c.lastRequestKey
 }
 
-func (c *CloudStatusManager) OnChange(ctx context.Context, st store.RStore) {
+func (c *CloudStatusManager) OnChange(ctx context.Context, st store.RStore, _ store.ChangeSummary) {
 	state := st.RLockState()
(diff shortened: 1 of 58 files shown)


milas (approved):
This looks fine to me, but I do wonder if given the goal is
> the system tells us which object is changing

Would it be easier to have something like a `TypedAction` interface that has `Types() []schema.GroupVersionKind` (could also restrict to 1x type) and then just pass the list of kinds to `OnChange` (which would be `nil` for any "legacy" events that don't implement the iface)? (Alternatively, could also have `TypedSubscriber` interface which also provides a list of kinds it cares about so that the notifier could handle the filtering.)

nicks:
Ya, that might be where this object ends up - with a list of types. But I'm pretty sure it's going to need to be more complex than that. For example, 

- actions can be batched together, and there are cases where you'll need to know if there are ANY legacy actions
- the Reconcile() function tends to expect the name of the object that changed
-

PR #4277, opened 2021-03-05
filewatch: use apiserver FileWatch model

milas:
This follows the "action-first" approach to use the apiserver model
for `FileWatch` and dispatch simple create/update/delete actions.

I didn't touch the git manager, so some things that are no longer
needed for normal file watching but it relies on were moved over
there.

This is actually pretty similar to the previous implementation since
file watch isn't really its own resource/concept in the current design.
It's a bit incremental of the incremental - the next step would be to
remove the diffing logic and instead _just_ react to changes to the
`FileWatchSpec`s within the watch manager (which would then just
the authority on `FileWatchStatus` and instead push the onus for
actually creating/updating **specs** to someone else.


internal/engine/fswatch/watchmanager.go
```diff
@@ -237,65 +162,143 @@ func (w *WatchManager) OnChange(ctx context.Context, st store.RStore) {
...
+			}
+			// since we're holding a reference to the old one, we just replace it in the map and clean up
+			// after the new watcher is successfully running to avoid missing any events
+			if err := w.addOrReplace(ctx, st, name, spec); err != nil {
+				logger.Get(ctx).Debugf("Error updating watch for %q: %v", name.String(), err)
+				continue
+			}
+			w.cleanupWatch(ctx, tw)
+
+			fw := state.FileWatches[name]
```
nicks, line 200: ah, i hit this problem too...

the approach to this that i've been taking is that, when a subscriber sends a Create/Update, it should keep a local map of the last create/updated object, and then use that map to make create/update idempotent

I think the problem with doing it this way is that you're going to run into weird circular data flow issues where this code is reading and writing the same data....

internal/engine/fswatch/watchmanager.go
```diff
@@ -237,65 +162,143 @@ func (w *WatchManager) OnChange(ctx context.Context, st store.RStore) {
...
+			} else {
+				fw = fw.DeepCopy()
+				spec.DeepCopyInto(&fw.Spec)
+			}
+
+			// we always have the latest status, so just unconditionally overwrite whatever is in the object
+			tw.status.DeepCopyInto(&fw.Status)
+
+			st.Dispatch(NewFileWatchUpdateAction(fw))
+		} else {
```
nicks, line 219: i think that if you make create/update idempotent, you might be able to merge these two branches

internal/engine/fswatch/watchmanager.go
```diff
@@ -237,65 +162,143 @@ func (w *WatchManager) OnChange(ctx context.Context, st store.RStore) {
...
+					ObjectMeta: metav1.ObjectMeta{Namespace: name.Namespace, Name: name.Name},
+					Spec:       *spec.DeepCopy(),
+				}
+			} else {
+				fw = fw.DeepCopy()
+				spec.DeepCopyInto(&fw.Spec)
+			}
+
+			// we always have the latest status, so just unconditionally overwrite whatever is in the object
+			tw.status.DeepCopyInto(&fw.Status)
```
nicks, line 216: should we reset the status when the spec changes? (I honestly don't know, this is something I struggled with too)

pkg/apis/core/v1alpha1/filewatch_types.go
```diff
@@ -86,9 +94,31 @@ func (in *FileWatch) IsStorageVersion() bool {
 	return true
 }
 
-func (in *FileWatch) Validate(ctx context.Context) field.ErrorList {
-	// TODO(user): Modify it, adding your API validation here.
-	return nil
+func (in *FileWatch) Validate(_ context.Context) field.ErrorList {
```
nicks, line 97: ooh, love this

internal/engine/upper.go
```diff
@@ -486,32 +495,66 @@ func handleStartProfilingAction(state *store.EngineState) {
...
-	event fswatch.TargetFilesChangedAction) {
+func handleFileWatchCreateEvent(_ context.Context, state *store.EngineState, action fswatch.FileWatchCreateAction) {
+	name := types.NamespacedName{Namespace: action.FileWatch.GetNamespace(), Name: action.FileWatch.GetName()}
+	state.FileWatches[name] = action.FileWatch
+}
 
-	if event.TargetID.Type == model.TargetTypeConfigs {
-		for _, f := range event.Files {
-			state.PendingConfigFileChanges[f] = event.Time
+func handleFileWatchUpdateEvent(_ context.Context, state *store.EngineState, action fswatch.FileWatchUpdateAction) {
```
nicks, line 503: fwiw, i ended up moving the Handle functions into the subpackage. which made them easier to use in tests...

nicks (commented):
whew! Ya, i'm glad we tried to do these in parallel, since i'm starting to see the common themes (rather than wondering if something was particiular to Cmd)

internal/engine/fswatch/watchmanager.go
milas: Ha! This is iteration 3, but I should probably just go back to the original. First, I had totally duplicated copies of the full `FileWatch` both places as you're suggesting. A second (short-lived) version was to rely entirely on the store to avoid duplicating here, which obviously did not pan out. I agree this final approach is problematic - I was trying to get closer to the K8s model somewhat (i.e. pretend that it's level-triggered).

I'll flip back to that - I think the next big experiment that needs to be done is shifting responsibility for the creation of the `FileWatch` to the resource to move towards our desired model, at which point, hopefully this becomes much more like a real reconciler, at which point it should just _read_ specs to trigger FS watchers, and let the FS watchers do status updates in the background. At that point, shouldn't be any circular weirdness 😵

internal/engine/fswatch/watchmanager.go
milas: Ah, good call - I had gone that route in the other PR but for some reason didn't do the same here. Just pushed some changes that make this dramatically more straightforward 😛

internal/engine/fswatch/watchmanager.go
milas: I've gone back and forth as well (I think the other PR resets it 😬) - I'm guessing it's best to standardize on resetting for consistency?

Here, while there could be value in still being able to see the previous changesets, it's also actually misleading because if you add a new file to the watch paths, it of course won't get retroactively added to those events.

I think one thing that `FileWatchStatus` is missing is probably more actual _status_ about the filesystem watcher. So perhaps `WatchStartTime` which gets updated whenever the filesystem watcher is started (or "updated"); that gives consumers another value to use against their own internal watermark, e.g. `if (myLastUpdated.Before(fwStatus.LastEventTime) || myLastUpdated.Before(fwStatus.WatchStartTime)`; that would allow e.g. easily triggering a rebuild when the watch patterns are changed (if desired behavior).

pkg/apis/core/v1alpha1/filewatch_types.go
milas: Was super happy that it just worked out of the box!

internal/engine/fswatch/watchmanager.go
nicks: ya, i had a similar experience - it's simpler if the controller that does create/delete is separate from the controller that does UpdateStatus. i think that having the ChangeSummary stuff to tell the controller what changed will also help with this

internal/engine/upper.go
```diff
@@ -486,32 +495,66 @@ func handleStartProfilingAction(state *store.EngineState) {
...
+func processFileWatchStatus(state *store.EngineState, name types.NamespacedName, status *filewatches.FileWatchStatus) {
+	if status.Error != "" || len(status.FileEvents) == 0 {
+		return
+	}
+
+	// since the store is called on EVERY update, can always just look at the last event
+	latestEvent := status.FileEvents[len(status.FileEvents)-1]
+
+	// TODO(milas): should probably centralize this logic and ensure `:` can't exist in resource names
+	targetParts := strings.SplitN(name.Name, ":", 2)
```
nicks, line 531: i wonder if these should be annotations on the FileWatch to know who should get triggered, at least to help with migration (rather than relying on the Name)

nicks (approved):

pkg/apis/core/v1alpha1/filewatch_types.go
```diff
@@ -99,6 +129,14 @@ func (in *FileWatchList) GetListMeta() *metav1.ListMeta {
 
 // FileWatchStatus defines the observed state of FileWatch
 type FileWatchStatus struct {
+	LastEventTime *metav1.Time `json:"lastEventTime,omitempty"`
```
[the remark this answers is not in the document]
milas: Added in 7d317d6 + 7982480c4

internal/engine/upper.go
milas: Switched to a label (to mirror manifest name - I think this is correct since its immutable? but can easily switch to annotation) in 6c4ed98

PR #4285, opened 2021-03-08
engine: print container restart

nicks:
Hello @maiamcc, @hyu,

Please review the following commits I made in branch nicks/ch11536:

1495a42b9030d26836717997631940ba399a804b (2021-03-08 12:47:30 -0500)
engine: print container restart

### DIFF
--- a/internal/engine/pod.go
+++ b/internal/engine/pod.go
@@ -54,8 +54,26 @@ func handlePodChangeAction(ctx context.Context, state *store.EngineState, action
 
 	prunePods(ms)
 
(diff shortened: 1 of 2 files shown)


nicks:
I'm not super tied to this exact wording. here's a picture of how this looks in the UI:

Screenshot from 2021-03-05 19-53-00

maiamcc (approved):
Nice! My only concern is that this messaging might get confusing in cases where we need to do a crash rebuild, but definitely better than the current state!

PR #4287, opened 2021-03-08
test: add extension tests to our integration test suite

nicks:
Hello @landism,

Please review the following commits I made in branch nicks/extensiong:

8b52b9164442dcdb8b0f4883a13b34870419f759 (2021-03-08 16:08:54 -0500)
test: add extension tests to our integration test suite

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -77,6 +77,10 @@ integration-kind:
 	KUBECONFIG="$(kind get kubeconfig-path --name="integration")" go test -mod vendor -p $(GO_PARALLEL_JOBS) -tags 'integration' -timeout 700s ./integration -count 1
 	kind delete cluster --name=integration
 
+# Run the extension integration tests against the current kubecontext
+test-extensions:
+	scripts/test-extensions.sh
+
 dev-js:
 	cd web && yarn install && yarn run start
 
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -77,6 +77,11 @@ integration-kind:
 	KUBECONFIG="$(kind get kubeconfig-path --name="integration")" go test -mod vendor -p $(GO_PARALLEL_JOBS) -tags 'integration' -timeout 700s ./integration -count 1
 	kind delete cluster --name=integration
 
+# Run the extension integration tests, assuming the extensions
```
landism, line 80: I'm not sure how GOPATH is involved. It looks like this is running them from /tmp?

landism (approved):

Makefile
nicks: ooh, good catch, fixed

nicks:
i needed to update this a bit so i could run the dev binary in prod web-mode, PTAL?

scripts/test-extensions.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/sh
+#
+# Checkout extensions to a temporary directory and run tilt against them.
+
+set -ex
+
+DIR=$(mktemp -d)
+git clone git@github.com:tilt-dev/tilt-extensions "$DIR"
```
milas, line 8: FYI You might want to do `https://github.com/tilt-dev/tilt-extensions` so it'd work without an SSH key, but not a big deal

nicks: good point!

PR #4289, opened 2021-03-09
engine: separate out Cmd creation from Cmd running

nicks:
Hello @milas, @ellenkorbes,

Please review the following commits I made in branch nicks/action-time3:

3d6b951e5e32f24fa33d098774252e07595b21e6 (2021-03-08 19:09:18 -0500)
engine: separate out Cmd creation from Cmd running
ServerController is getting closer to what typical "user-authored" controller will look like.
It doesn't manage any work directly - it simply creates new Cmd objects that do the real work.


ellenkorbes:
i don't understand all the code but it seems to do what we want.

i do worry that CmdServer/servercontroller introduce a lot of duplication and i wonder if there's some way to go around that.

ellenkorbes (approved):

internal/engine/local/servercontroller.go
```diff
@@ -0,0 +1,197 @@
...
+				v1alpha1.LabelManifest: name,
+			},
+		},
+		Spec: cmdSpec,
+	}
+	c.createdCmds[name] = cmd
+	st.Dispatch(CmdCreateAction{Cmd: cmd})
+}
+
+type CmdServer struct {
```
milas, line 176: Will these become API server types in the future?

milas (approved):

internal/engine/local/servercontroller.go
nicks: ya, i think so - but still figuring out what the Status field should look like

PR #4299, opened 2021-03-10
filewatch: ensure absolute paths are always used

milas:
There was no bug here - the relative paths _would_ get resolved
by the watchers deeper in the code.

However, the FileWatch API design is pushing to avoid implicit
relative paths based on Tilt's CWD; this makes the logic simpler
a lot of places, ensures that it's self-contained, and enables
future possibilities around daemon mode/multiple Tiltfiles (where
there is no longer "one" CWD).

The specs with relative paths being generated from manifests would
get rejected by the API server, so this is pre-empting that by
making sure they're valid and will migrate cleanly. A temporary
hack to explicitly call the validate logic also exists.

Any errors just get dispatched as warnings to the manifest logs
currently, as there's not a great way to do more robust error-handling
given how logic/responsibility is split; once things move into the API
server, especially as more controllers become responsible for creating
their own FileWatch objects (rather than relying on implicit logic to
generate them), it will be easier to handle errors further upstream,
e.g. put a resource into an error state upfront.

### DIFF
--- a/internal/engine/fswatch/watchmanager.go
+++ b/internal/engine/fswatch/watchmanager.go
@@ -19,6 +19,7 @@ import (
 	filewatches "github.com/tilt-dev/tilt/pkg/apis/core/v1alpha1"
 	"github.com/tilt-dev/tilt/pkg/logger"
 	"github.com/tilt-dev/tilt/pkg/model"
+	"github.com/tilt-dev/tilt/pkg/model/logstore"
 )
 
 // When we see a file change, wait this long to see if any other files have changed, and bundle all changes together.
@@ -39,6 +40,10 @@ var ConfigsTargetID = model.TargetID{
 	Name: "singleton",
 }
 
+type dispatcher interface {
+	Dispatch(action store.Action)
+}
+
 // If you modify this interface, you might also need to update the watchRulesMatch function below.
 type WatchableTarget interface {
 	ignore.IgnorableTarget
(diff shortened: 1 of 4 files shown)


internal/engine/fswatch/watchmanager.go
```diff
@@ -48,9 +53,94 @@ type WatchableTarget interface {
...
+	absPath, err := filepath.Abs(path)
+	if err != nil {
+		return "", fmt.Errorf("could not determine absolute path for %q: %v", path, err)
+	}
+	return absPath, nil
+}
+
+func specForTarget(ctx context.Context, mn model.ManifestName, t WatchableTarget, globalIgnores []model.Dockerignore, dispatcher dispatcher) filewatches.FileWatchSpec {
+	logValidationError := func(err error) {
+		dispatcher.Dispatch(store.NewLogAction(
```
nicks, line 68: hmmm...what prevents this from going into an infinite loop and logging the error on every change?

nicks:
high-level question: for demos like in L's comment here (https://app.clubhouse.io/windmill/story/11532/document-how-to-build-entr), I thought we DID want people to be able to put relative paths in the FileWatch YAML? Otherwise it seems like it'd be hard to use the API if the YAML was non-portable to other machines.

I was vaguely aware of a mechanism for "normalizing" fields (similar to how it normalizes image names and other default fields) that we might be able to use here, but am not super familiar with the details

internal/engine/fswatch/watchmanager.go
milas: Gahh, very valid point. Will restructure to prevent spamming on every store change to be only each time spec changes.

milas:
Hmm...interesting! Okay, we could relax that restriction on the validation front. (We could also always offer a "base path" annotation to use instead of cwd in the future for resolving relative paths if useful, e.g. multiple Tiltfiles.)

I was putting that restriction in place because it makes things easier to reason about, particularly as FileWatch gets further removed from manifest/target.

milas:
There's some normalization during `Tiltfile` parsing on the starlark side: https://github.com/tilt-dev/tilt/blob/master/internal/tiltfile/value/value.go#L36-L48

This might be a case where most things actually already are "as desired" but tests are writing impossible data directly into state.

milas:
Closing this as we _want_ to allow relative paths. Will update validation in the opposite direction in a follow-up PR

PR #4304, opened 2021-03-11
k8s: inject pull policies into crds

### LINKED ISSUE #4291: Overriding imagePullPolicy in CRDs with tilt-managed images
I've presented this as a bug/oversight but maybe it was never implemented? Sorry if this is actually a feature request in disguise.

### Expected Behavior
When I use `k8s_kind` on a CRD with images, I want to specify `imagePullPolicy: Always` in my manifests and have Tilt override it when it manages the build so that I can both

1. always pull the latest image if I don't need to work on that image
2. have Tilt manage the build and use my local code changes.

That is, it should work the same way as images being built for normal TIlt resources (eg. pods from deployments).

### Current Behavior
Tilt is not overriding the `imagePullPolicy` of my CRDs and is failing with

```
ImagePullBackOff: Back-off pulling image "foo.bar/baz/helloworld:tilt-4a9902d1db2b9157"
```

even when I am specifying a `docker_build` for it. It looks like tilt knows enough to override the image name, but not enough to override the imagePullPolicy, which is what makes me think this is an oversight/bug.

### Steps to Reproduce

1. Use `k8s_kind` with an `image_json_path` to a CRD that has images. For example, https://github.com/tilt-dev/tilt/issues/3045#issuecomment-785155248
2. Add a `docker_build` for the image
3. `tilt up` and for this use case, trigger an argo workflow with this image

#### `tilt doctor` Output

```
Tilt: v0.18.9, built 2021-02-12
System: darwin-amd64
---
Docker (cluster)
- Host: tcp://192.168.64.3:2376
- Version: 1.41
- Builder: 2
---
Docker (local)
- Host: Error: Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?
---
Kubernetes
- Env: minikube
- Context: minikube
- Cluster Name: minikube
- Namespace: local
- Container Runtime: docker
- Version: v1.20.2
---

```

#### About Your Use Case

Our use case is being able to use Tilt to manage both our "services" (normal deployments that create pods) and our "steps" (argo workflow steps) in the same way, without having to resort to workarounds or remembering to manually pull.

landism:
Fixes #4291

### Problem

For standard k8s objects, Tilt will inject an image pull policy. For CRDs, it will not.
It's relatively common for CRDs to use the normal k8s container spec, with "image" and "imagePullPolicy" fields. Tilt provides a way for users to configure how to inject the image ref, but leave users to their own devices when it comes to imagePullPolicy.

### Solution

If a CRD "image" field has an "imagePullPolicy" neighbor, inject the pull policy.

Note: the image object locator (where you specify, e.g., a repo and ref field) just ignores the pull policy. An obvious answer here is to add an option to specify the pull policy field. My weak gut instinct is it's not worth the effort to worry about that until someone complains. I'm happy to do that as well if someone has a different gut instinct.

### DIFF
--- a/internal/k8s/jsonpath/value.go
+++ b/internal/k8s/jsonpath/value.go
@@ -42,9 +42,13 @@ func (v *Value) Set(newV reflect.Value) {
 }
 
 func (v *Value) Sibling(name string) (val Value, ok bool) {
+	if !v.parentMap.IsValid() {
+		return Value{}, false
+	}
+
 	key := reflect.ValueOf(name)
 	sib := v.parentMap.MapIndex(key)
-	if sib == (reflect.Value{}) {
+	if !sib.IsValid() {
 		return Value{}, false
 	}
 
--- a/internal/k8s/jsonpath/value.go
+++ b/internal/k8s/jsonpath/value.go
@@ -40,3 +40,17 @@ func (v *Value) Set(newV reflect.Value) {
 	}
 	v.Value.Set(newV)
 }
+
+func (v *Value) Sibling(name string) (val Value, ok bool) {
+	key := reflect.ValueOf(name)
+	sib := v.parentMap.MapIndex(key)
+	if sib == (reflect.Value{}) {
+		return Value{}, false
+	}
+
+	return Value{
+		Value:        sib,
+		parentMap:    v.parentMap,
+		parentMapKey: key,
+	}, true
+}
--- a/internal/k8s/locator.go
+++ b/internal/k8s/locator.go
@@ -126,7 +126,7 @@ func (l *JSONPathImageLocator) Inject(e K8sEntity, selector container.RefSelecto
 			val.Set(reflect.ValueOf(container.FamiliarString(injectRef)))
 			modified = true
 
-			if pullPolicyVal, ok := val.Sibling(imagePullPolicyKey); ok {
+			if pullPolicyVal, ok := val.Sibling(imagePullPolicyKey); ok && pullPolicyVal.CanSet() {
 				pullPolicyVal.Set(reflect.ValueOf(string(pullPolicy)))
 			}
 		}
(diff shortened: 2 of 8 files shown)


internal/k8s/jsonpath/value.go
```diff
@@ -40,3 +40,17 @@ func (v *Value) Set(newV reflect.Value) {
 	}
 	v.Value.Set(newV)
 }
+
+func (v *Value) Sibling(name string) (val Value, ok bool) {
+	key := reflect.ValueOf(name)
+	sib := v.parentMap.MapIndex(key)
+	if sib == (reflect.Value{}) {
```
nicks, line 47: i think `sib.IsZero()` would be more idiomatic

internal/k8s/locator.go
```diff
@@ -121,6 +125,10 @@ func (l *JSONPathImageLocator) Inject(e K8sEntity, selector container.RefSelecto
 		if selector.Matches(ref) {
 			val.Set(reflect.ValueOf(container.FamiliarString(injectRef)))
 			modified = true
+
+			if pullPolicyVal, ok := val.Sibling(imagePullPolicyKey); ok {
```
nicks, line 129: can you also check pullPolicyVal.CanSet()? always get nervous about Value.Set panicking

nicks (approved):
nice!

internal/k8s/locator.go
landism: ha, did you see the CI failure?

nicks: lol, nope i did not! :star2:

internal/k8s/jsonpath/value.go
landism: what a great error message: `panic: reflect: call of reflect.Value.IsZero on zero Value`

It looks like:
1) `Value.IsZero` returns true if it's got the value's zero value, and panics otherwise.
2) `MapIndex` returns an untyped zero value
3) `Value.IsValid` is what we want for this case

PR #4313, opened 2021-03-12
apis: use MicroTime instead of Time

nicks:
Hello @milas,

Please review the following commits I made in branch nicks/microtime:

96be94cbef484cdac528fcccdd9bb8bc77fcd663 (2021-03-12 15:19:22 -0500)
apis: use MicroTime instead of Time
Currently, metav1.Time is only stored with second-level granularity,
which is probably not sufficient for this API.

### DIFF
--- a/internal/controllers/core/cmd/controller.go
+++ b/internal/controllers/core/cmd/controller.go
@@ -7,7 +7,6 @@ import (
 	"io"
 	"strings"
 	"sync"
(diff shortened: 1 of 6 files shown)


nicks:
for context: https://github.com/kubernetes/kubernetes/issues/36752

milas (approved):

PR #4315, opened 2021-03-12
engine: restart cmds on file change

nicks:
Hello @milas,

Please review the following commits I made in branch nicks/trigger:

3c0f41a1e6ef13b7a937bd812aeebe103281c1a2 (2021-03-12 17:03:40 -0500)
engine: restart cmds on file change

96be94cbef484cdac528fcccdd9bb8bc77fcd663 (2021-03-12 15:19:22 -0500)
apis: use MicroTime instead of Time
Currently, metav1.Time is only stored with second-level granularity,
which is probably not sufficient for this API.


nicks:
this ended up being more straightforward than i expected. 

I tweaked @ellenkorbes 's idea of having a TiltSpec object into the RestartOn spec -- i think it's a good thing in general for "restartable" resources to have to explicitly opt in to being restartable, because there are lots of things in the ecosystem that might not be restartable

milas (approved):

PR #4318, opened 2021-03-15
cli: fix json errors on engine dump

### LINKED ISSUE #4316: dump engine stopped working after upgrading to 0.18.12
### Expected Behavior
`tilt dump engine` should produce dump of internal state. We use this json to query status of the services.

### Current Behavior

```
 ~/Downloads/tilt.0.18.12.mac.x86_64/tilt dump engine
dump engine: Could not decode: invalid character ':' looking for beginning of object key string
```

however it works just fine when I move back to `0.18.11`

```
➜  ~/Downloads/tilt.0.18.11.mac.x86_64/tilt dump engine
{
  "AnalyticsEnvOpt": 0,
  "AnalyticsNudgeSurfaced": false,
  "AnalyticsTiltfileOpt": 0,
  "AnalyticsUserOpt": 2,
  "CloudAddress": "cloud.tilt.dev",
  "CloudStatus": {
    "TeamName": "",
    "TokenKnownUnregistered": false,
    "Username": "denyska",
    "WaitingForStatusPostRegistration": false
  },
  "CompletedBuildCount": 0,
  "ConfigFiles": [
    "/Users/denyska/Projects/myrepo/Tiltfile"
  ],
  "CurrentlyBuilding": {},
  "DockerPruneSettings": {
    "Enabled": false,
    "Interval": 0,
    "KeepRecent": 0,
    "MaxAge": 0,
    "NumBuilds": 0
  },
  "EngineMode": {
....
```

### Steps to Reproduce

1.
2.
3.
4.

#### `tilt doctor` Output
```
~/Downloads/tilt.0.18.12.mac.x86_64/tilt  doctor
Tilt: v0.18.12, built 2021-03-09
System: darwin-amd64
---
Docker (cluster)
- Host: tcp://192.168.64.48:2376
- Version: 1.41
- Builder: 2
---
Docker (local)
- Host: [default]
- Version: 1.41
- Builder: 2
---
Kubernetes
- Env: minikube
- Context: minikube
- Cluster Name: minikube
- Namespace: default
- Container Runtime: docker
- Version: v1.20.2
---

...
```

milas:
~~Add a custom encoder for `types.NamespacedName`.~~

Hide new apiserver types from the engine dump output. These types
should only be used by querying the apiserver rather than relying
on this debugging command.

(The actual error is due to JSON serialization of
`types.NamespacedName`; an earlier version of this change made
that serialize correctly but has been reverted since we don't
want to serialize it at all.)

Closes #4316.

### DIFF
--- a/internal/store/engine_state.go
+++ b/internal/store/engine_state.go
@@ -121,8 +121,8 @@ type EngineState struct {
 
 	// API-server-based data models. Stored in EngineState
 	// to assist in migration.
-	Cmds        map[string]*Cmd
-	FileWatches map[types.NamespacedName]*filewatches.FileWatch
+	Cmds        map[string]*Cmd                                 `json:"-"`
+	FileWatches map[types.NamespacedName]*filewatches.FileWatch `json:"-"`
 }
 
 type CloudStatus struct {


nicks (approved):
I'm fine with this...

but if we're going to encode NamespacedName as just the Name part, I'm not sure if it makes sense to use NamespacedName in the EngineState to begin with...

nicks:
also - is there a test that would have caught this?

milas:
`types.NamespacedName::String()` gives `$namespace/$name`, so it'll include the namespace. I just only did an empty check for name because a stray namespace without a name doesn't make much sense (this is _kinda_ similar to how we handle `TargetID`; if _either_ is empty, it'll return an empty string).

Re: tests, at the moment, we only optimistically try to use it to dump state when we think a test has already failed for debugging purposes that I saw. I was going to start by adding calls to the integration tests (where practical) to ensure it passes on a variety of different scenarios but wanted to get a fix up first

nicks (approved):

PR #4324, opened 2021-03-16
tests: fix engine dump on test failure

milas:
Randomized port assignment wasn't playing nice with this. I
simplified the port assignment so that it's part of the main Tilt
driver fixture so that all calls will use it consistently. In the
event that a test does pass `--port` as an arg, it won't include
`TILT_PORT` on the env for that call to avoid precedence issues.

Now, when `tilt dump engine` is called, it'll have the right port
via `TILT_PORT` in env, so that the engine state can be dumped.

This also (minimally) simplifies a couple tests that needed to capture
the port and explicitly pass it back for subsequent calls, either to
intentionally simulate a port conflict or to interact with the running
engine.

### DIFF
--- a/integration/tilt.go
+++ b/integration/tilt.go
@@ -8,19 +8,31 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
-	"strconv"
 	"strings"
 	"sync"
 	"syscall"
+	"testing"
 	"time"
+
+	"github.com/stretchr/testify/require"
 )
 
 type TiltDriver struct {
+	Port    int
 	Environ map[string]string
 }
 
-func NewTiltDriver() *TiltDriver {
+func NewTiltDriver(t testing.TB) *TiltDriver {
+	l, err := net.Listen("tcp", "")
+	require.NoError(t, err, "Could not get a free port")
+	port := l.Addr().(*net.TCPAddr).Port
+	require.NoError(t, l.Close(), "Could not get a free port")
+	return NewTiltDriverWithExplicitPort(port)
+}
+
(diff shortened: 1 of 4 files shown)


integration/tilt.go
```diff
@@ -37,6 +49,16 @@ func (d *TiltDriver) cmd(args []string, out io.Writer) *exec.Cmd {
...
 	}
+	hasPortArg := false
+	for _, arg := range args {
+		if strings.HasPrefix("--port=", arg) {
+			hasPortArg = true
+			break
+		}
+	}
+	if !hasPortArg && d.Port > 0 {
+		cmd.Env = append(cmd.Env, fmt.Sprintf("TILT_PORT=%d", d.Port))
```
nicks, line 60: mostly out of curiosity, is there a reason you're using the env variable instead of the port flag here? would rather be consistent within the test, since it seems likely this will bite someone later.

nicks (approved):

integration/tilt.go
milas: The goal here was to have some sane default behavior but get out of the way if you explicitly want to do something with ports.

Since `--port` isn't a global flag, it could make some Tilt commands fail, so this was easier than adding it explicitly to each wrapper, and so the check to omit it if `--port` was defined was to make sure there is an escape hatch, but I can make it a bit more explicit, e.g. have `AutomaticPortSelection bool` on the fixture and don't touch it if false?

milas: ^ kinda did something similar, made it more explicit and throw hard errors if you're in "automatic port" mode and try to specify either `--port` or `TILT_PORT` so there should be no surprises

PR #4325, opened 2021-03-16
tests: use local registry for integration tests

milas:
Add `kube-public` read-only access for the `tilt-integration` user
so that it can find the `ConfigMap` for registry information.

I was debugging some integration tests today and realized the images
weren't ending up in my local registry, and I wanted to examine them.
This gives full read-only access to `kube-public` which I think is fine,
but we could probably scope this down to `ConfigMap` and even more
specifically the `local-registry-hosting` resource name.

Hopefully this also makes some of the integration tests slightly faster?

### DIFF
--- a/integration/access.yaml
+++ b/integration/access.yaml
@@ -1,6 +1,6 @@
 # Grant access to everything in the tilt-integration namespace
 kind: Role
-apiVersion: rbac.authorization.k8s.io/v1beta1
+apiVersion: rbac.authorization.k8s.io/v1
 metadata:
   name: tilt-integration-user-full-access
   namespace: tilt-integration
@@ -15,7 +15,7 @@ rules:
   verbs: ["*"]
 ---
 kind: RoleBinding
-apiVersion: rbac.authorization.k8s.io/v1beta1
+apiVersion: rbac.authorization.k8s.io/v1
 metadata:
   name: tilt-integration-user-view
   namespace: tilt-integration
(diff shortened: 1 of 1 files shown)


nicks (approved):
Hmmmm...i think this is fine, but i think you might be misdiagnosing the problem?

The integration tests already appear to be pushing to the local registry...

If the integration tests weren't using the local registry, then they wouldn't pass at all AFAIK, since they don't have credentials to push anywhere else.

milas:
It was using `kind load` - here's a random test failure (logs only available on failures): https://app.circleci.com/pipelines/github/tilt-dev/tilt/6544/workflows/1ddbe322-21ba-47a8-837b-2b484c23627c/jobs/58219/tests#failed-test-0

```
WARNING: You are running Kind without a local image registry.
Tilt can use the local registry to speed up builds.
Instructions: https://github.com/tilt-dev/kind-local
...
live-update-… │ STEP 4/5 — Pushing gcr.io/windmill-test-containers/integration/live_update_after_crash_rebuild:tilt-4983c78f9011268f
live-update-… │      Loading image to KIND
live-update-… │      Image: "gcr.io/windmill-test-containers/integration/live_update_after_crash_rebuild:tilt-4983c78f9011268f" with ID "sha256:4983c78f9011268f1a92118c6560e2a53b0e824669f0701e8d616250de98abd1" not yet present on node "kind-control-plane", loading...
live-update-… │ 
```

Just meant it was bypassing local registry, so wasn't a huge deal, but it's a bit of a pain when you want to actually inspect the images it's building with Docker CLI easily

nicks:
ah, nice find!

PR #4332, opened 2021-03-17
filewatch: convert store subscriber to reconciler

milas:
This makes the apiserver the canonical location for the FileWatch
objects.


internal/controllers/core/filewatch/watcher.go
```diff
@@ -0,0 +1,86 @@
...
+func (w *watcher) recordEvent(ctx context.Context, client ctrlclient.Client, st store.RStore, fsEvents []watch.FileEvent) error {
+	now := metav1.NowMicro()
+	w.mu.Lock()
+	defer w.mu.Unlock()
+	event := filewatches.FileEvent{Time: *now.DeepCopy()}
+	for _, fsEvent := range fsEvents {
+		event.SeenFiles = append(event.SeenFiles, fsEvent.Path())
+	}
+	if len(event.SeenFiles) != 0 {
+		logger.Get(ctx).Debugf("File event for %q: %v", w.name.String(), event.SeenFiles)
```
nicks, line 58: did you intend to leave this in?

internal/controllers/fake/fixture.go
```diff
@@ -0,0 +1,123 @@
...
+}
+
+// object just bridges together a couple of different representations of runtime.Object.
+// Scaffolded/code-generated types should meet this by default.
+type object interface {
+	ctrlclient.Object
+	resource.Object
+}
+
+type ControllerFixture struct {
```
nicks, line 36: ooh i like this!

pkg/apis/core/v1alpha1/filewatch_types.go
```diff
@@ -119,11 +119,13 @@ func (in *FileWatchList) GetListMeta() *metav1.ListMeta {
 
 // FileWatchStatus defines the observed state of FileWatch
 type FileWatchStatus struct {
+	// MonitorStartTime is the timestamp of when filesystem monitor was started.
+	MonitorStartTime metav1.MicroTime `json:"monitorStartTime,omitempty"`
 	// LastEventTime is the timestamp of the most recent file event. It is nil if no events have been seen yet.
```
nicks, line 124: nil -> zero? (now that this isn't a pointer?)

nicks (approved):
LGTM!

didn't read this super closely since it seems like a lot of the code is simply moved

pkg/apis/core/v1alpha1/filewatch_types.go
milas: Ah, yep! FWIW the reason I changed these was from what I've seen, apiserver expects anything w/o `omitempty` to be **required** (even if it's a pointer type - assumption I guess is you're using pointer for perf reasons rather than nullability reasons?) so it doesn't really get us anything other than more annoying ergonomics on the Go side, and since we are required to use `omitempty`, the JSON side is the same either way

milas:
> didn't read this super closely since it seems like a lot of the code is simply moved

Yeah, it was unfortunately not a particularly elegant lift-and-shit, sorry about the nightmare of a diff 😦 

But you're correct - logic within the actual watching portions remained the same, and reconciler was 80% similar to previous incarnation, just cleaner due to not being a store subscriber 🎉

PR #4337, opened 2021-03-17
cli: updog accepts configs as input

nicks:
Hello @milas, @maiamcc,

Please review the following commits I made in branch nicks/ch11531:

8ebac2d998e128c63f5d596e682676639fd89771 (2021-03-17 17:32:21 -0400)
cli: updog accepts configs as input
Most of the input-decoding code here is loosely adapted from ctlptl and
not super great. We should probably clean it up at some point and move
it into a common package.

kubectl's input parsing code is much more complicated because it needs to
discover types from the server before parsing anything

### DIFF
--- a/internal/cli/updog.go
+++ b/internal/cli/updog.go
@@ -2,19 +2,24 @@ package cli
 
 import (
 	"context"
+	"fmt"
 	"log"
+	"os"
+	"strings"
 	"time"
 
 	"github.com/spf13/cobra"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/cli-runtime/pkg/genericclioptions"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	ctrlclient "sigs.k8s.io/controller-runtime/pkg/client"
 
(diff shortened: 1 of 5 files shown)


internal/cli/updog.go
```diff
@@ -135,10 +151,16 @@ func provideUpdogSubscriber(objects []client.Object, client client.Client) *updo
 
 func (s *updogSubscriber) SetUp(ctx context.Context, _ store.RStore) error {
 	for _, obj := range s.objects {
+		// Create() modifies its object mysteriously, so copy it first
```
milas, line 154: Ugh this confused me greatly too at one point - there's weirdly code to restore GVK on `Update()` but not `Create()` and an ominous note on the restore function that it could hopefully go away at some point after deps are upgraded to 1.16, but I think it's past that point already and presumably still an issue 😢 

https://github.com/kubernetes-sigs/controller-runtime/blob/b5065bd85190e92864522fcc85aa4f6a3cce4f82/pkg/client/client.go#L155

milas (approved):

internal/cli/updog.go
nicks: haha, vincepri has filed some tilt issues, maybe we can trade a tilt fix for fixing that TODO in controller-runtime :joy:

PR #4350, opened 2021-03-19
server: use logger facade instead of Zap

milas:
This removes dependencies on `zapr` (the logr facade for Zap) and
makes logging a bit more consistent.


nicks (approved):
LGTM!

facade is a good word

milas:
Okay, since these are obnoxiously spammy - here's a thought for an alternate approach: use [klogr](https://github.com/kubernetes/klog/tree/master/klogr) and redirect them all to `klog.v(5)` or something so they're with all the other K8s logs?

nicks:
ya, that seems reasonable to me

milas:
Closing in favor of #4357 which will redirect these to `klog` instead so their output can be controlled consistently along with `apiserver` & friends

PR #4354, opened 2021-03-22
web: add zoom in/out for logs

### LINKED ISSUE #4210: Zoom into logs for effectively larger font size

Font size is too small for some folks. There should be a way to zoom.

### Current Behavior

Just one font size

### Why Do You Want This?

**Additional context**

milas:
Add buttons next to "Clear logs" to adjust the font size for logs.

At a high-level, a CSS custom property (variable) is used to set
a % font size for log lines (the absolute font size is set on the
container, and the default property value is 100%).

This provides an easy and performant way for JS to adjust the log
font size, particularly since the actual log CSS is pure CSS from
an SCSS stylesheet, not using `styled-components` in React.

The scale value is read from/written to local storage under a global
(NOT Tiltfile-namespaced) key so that it's persistent and applies to
all Tilt sessions. As a bonus, that means it live synchronizes between
tabs which is kinda nifty.

Closes #4210.

### DIFF
--- a/web/src/LogActions.test.tsx
+++ b/web/src/LogActions.test.tsx
@@ -1,0 +1,61 @@
+import { mount } from "enzyme"
+import React from "react"
+import {
+  LogFontSizeScaleCSSProperty,
+  LogFontSizeScaleLocalStorageKey,
+  LogFontSizeScaleMinimumPercentage,
+  LogsFontSize,
+} from "./LogActions"
+
+describe("LogsFontSize", () => {
+  const cleanup = () => {
+    localStorage.clear()
+    document.documentElement.style.removeProperty("--log-font-scale")
+  }
+
+  beforeEach(() => {
+    cleanup()
+    // CSS won't be loaded in test context, so just explicitly set it
+    document.documentElement.style.setProperty("--log-font-scale", "100%")
+  })
+  afterEach(cleanup)
+
+  const getCSSValue = () =>
+    getComputedStyle(document.documentElement).getPropertyValue(
+      LogFontSizeScaleCSSProperty
+    )
+  // react-storage-hooks JSON (de)serializes transparently,
+  // need to do the same when directly manipulating local storage
+  const getLocalStorageValue = () =>
+    JSON.parse(localStorage.getItem(LogFontSizeScaleLocalStorageKey) || "")
+  const setLocalStorageValue = (val: string) =>
+    localStorage.setItem(LogFontSizeScaleLocalStorageKey, JSON.stringify(val))
+
+  it("restores persisted font scale on load", () => {
+    setLocalStorageValue("360%")
+    mount(<LogsFontSize />)
+    expect(getCSSValue()).toEqual("360%")
+  })
+
+  it("decreases font scale", () => {
+    const root = mount(<LogsFontSize />)
+    root.find("LogFontSizeDecreaseButton").simulate("click")
+    expect(getCSSValue()).toEqual("95%")
+    expect(getLocalStorageValue()).toEqual(`95%`) // JSON serialized
+  })
+
+  it("has a minimum font scale", () => {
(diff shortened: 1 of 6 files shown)


web/src/LogActions.tsx
```diff
@@ -0,0 +1,120 @@
...
+  // (i.e. log zoom applies across all Tiltfiles)
+  const [logFontScale, setLogFontSize] = useStorageState<string>(
+    localStorage,
+    LogFontSizeScaleLocalStorageKey,
+    () =>
+      document.documentElement.style.getPropertyValue(
+        LogFontSizeScaleCSSProperty
+      )
+  )
+  useEffect(() => {
```
nicks, line 65: if logFontScale is invalid when we read it, it seems like we'd want to fix that in the local variable here, rather than letting all the downstream functions do their own validation?

web/src/LogActions.tsx
```diff
@@ -0,0 +1,120 @@
...
+`
+
+const FontSizeControls = styled.div`
+  color: ${Color.gray6};
+  vertical-align: middle;
+  display: flex;
+  flex-wrap: nowrap;
+`
+
+const FontSizeControlsDivider = styled.div`
```
nicks, line 26: it feels weird to do a divider like this...usually i see plain html content, or using ":after" on a button element, but not ":after" on a divider element

web/src/LogActions.test.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+  it("restores persisted font scale on load", () => {
+    setLocalStorageValue("360%")
+    mount(<LogsFontSize />)
+    expect(getCSSValue()).toEqual("360%")
+  })
+
+  it("decreases font scale", () => {
+    const root = mount(<LogsFontSize />)
+    root.find("LogFontSizeDecreaseButton").simulate("click")
```
nicks, line 42: i think the idiomatic way to do this is to import the styled component itself, and do a .find(Component) (rather than passing around strings)

web/src/LogActions.tsx
```diff
@@ -0,0 +1,120 @@
...
+  font-size: ${FontSize.default};
+  &:after {
+    content: "|";
+  }
+`
+
+const FontSizeButton = styled.button`
+  ${mixinResetButtonStyle}
+  color: ${Color.gray6};
```
nicks, line 35: would it make sense to use the transition css from the clear log button?

web/src/LogActions.tsx
```diff
@@ -0,0 +1,120 @@
...
+
+  padding: 0 4px;
+
+  &:after {
```
nicks, line 39: ditto here - seems weird to put this in ':after' selector rather than plain html content

nicks (approved):

web/src/LogActions.test.tsx
milas: For whatever reason this would NOT work yesterday but yarn is happy now - fixed in ee8fdff68f7a1b32e6bc64bebf9fa55ead759dfe

web/src/LogActions.tsx
milas: Made more standard in e86d044c5df84e78539232797a92bd5578dfd32a

web/src/LogActions.tsx
milas: Good call, added some transition in e86d044c5df84e78539232797a92bd5578dfd32a

web/src/LogActions.tsx
milas: Tweaked in b4c82591e5b107070a1112cd8ad7d8bc3057441b - will reset to `100%` here if invalid, so there's no need for extra validation when stepping.

(FWIW None of this _should_ be needed - I'm just extra paranoid because I've firsthand experienced lots of icky behaviors with localStorage, particularly when apps change "schema" of data in keys and havoc ensues)

maiamcc:
an hero

PR #4355, opened 2021-03-22
local: kill servers when their manifest is removed

nicks:
Hello @milas, @landism,

Please review the following commits I made in branch nicks/orphans:

2f49289dd575d4ea77bcd78f1e27df0920c5be6d (2021-03-22 16:22:21 -0400)
local: kill servers when their manifest is removed


nicks:
(i think this broke when i was converting things over to the new system)

milas (approved):
Terrifying test name, but changes look good 👍

PR #4357, opened 2021-03-23
server: redirect all controller-runtime logs to klog

milas:
Unlike most other K8s libraries/projects, controller-runtime uses
a logr abstract interface internally. Unfortunately, it doesn't
_really_ make use of levels internally very much, so redirecting
all its logs to the Tilt debug log adds a bunch of noise (see #4350).

As an alternative approach, we can use `klogr` which is a logr impl
using klog under the hood and redirect it to a higher verbosity
klog instance. Given that apiserver internal logs use klog already,
this gives a single, consistent way to get internal logging across
these components. In the future, we might want to re-introduce a
Tilt-specific facade for these and have more granular handling by
each component to avoid noisy libraries from flooding the output,
but this is a good interim step.


internal/controllers/manager.go
```diff
@@ -51,9 +50,11 @@ func (m *TiltServerControllerManager) GetClient() ctrlclient.Client {
 func (m *TiltServerControllerManager) SetUp(ctx context.Context, st store.RStore) error {
 	ctx, m.cancel = context.WithCancel(ctx)
 
-	// TODO(milas): we should provide a logr.Logger facade for our logger rather than using zap
-	w := logger.Get(ctx).Writer(logger.DebugLvl)
-	ctrl.SetLogger(zap.New(zap.UseDevMode(true), zap.WriteTo(w)))
+	// controller-runtime internals don't really make use of verbosity levels, so in lieu of a better
```
maiamcc, line 53: of a better what 👀

maiamcc (approved):
🙌  ty!

internal/controllers/manager.go
milas: 🤦 One day I will re-wrap a comment without losing a

PR #4363, opened 2021-03-24
k8s: remove ability to watch all namespaces

nicks:
Hello @landism, @maiamcc,

Please review the following commits I made in branch nicks/namespace:

6ed82b7cbea44d6b648804a7ed7b2196d049cdc8 (2021-03-24 18:16:01 -0400)
k8s: remove ability to watch all namespaces

### DIFF
--- a/internal/engine/k8swatch/pod_watch.go
+++ b/internal/engine/k8swatch/pod_watch.go
@@ -118,7 +118,7 @@ func (w *PodWatcher) OnChange(ctx context.Context, st store.RStore, _ store.Chan
 }
 
 func (w *PodWatcher) setupWatch(ctx context.Context, st store.RStore, ns k8s.Namespace) {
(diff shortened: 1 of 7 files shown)


landism:
Just to be sure I understand:
1. It's already the case that we're never actually calling this with an empty namespace, and the namespace change is just cleaning up code and, absent bugs, has no theoretical user impact.
2. This PR is also removing the ability to watch by label, which could potentially have performance implications if a user has a namespace that has some tilt-managed resources and also a lot of non-tilt-managed resources, but we don't have strong evidence that's relevant and we can worry about perf problems if they manifest.
That right?

landism (approved):

nicks:
ya, all the callers were calling it with a real namespace (see: https://github.com/tilt-dev/tilt/issues/2846)

the only caller that was using the watch-by-label functionality was the service watcher, which really isn't perf sensitive anyway.

PR #4366, opened 2021-03-25
tiltrun: add data model

milas:
The `TiltRun` type provides an aggregation of resources for a given
Tiltfile.

The goal is to fulfill the needs of the exit controller, which is
used in CI mode to either signal an exit with an error if a failure
occurs (e.g. failed build or runtime error) or clean exit if all
resources build + come up cleanly.


pkg/apis/core/v1alpha1/tiltrun_types.go
```diff
@@ -0,0 +1,274 @@
...
+	metav1.TypeMeta `json:",inline"`
+	metav1.ListMeta `json:"metadata,omitempty"`
+
+	Items []TiltRun `json:"items"`
+}
+
+// TiltRunSpec defines the desired state of TiltRun
+type TiltRunSpec struct {
+	// TiltfilePath is the path to the Tiltfile for the run. It cannot be empty.
+	TiltfilePath string `json:"tiltfilePath"`
```
milas, line 58: I could see this being optional in the future so that you could theoretically not have a `Tiltfile` at all and purely interact with API -> create `TiltRun` and then (via API) create resources attached to that.

pkg/apis/core/v1alpha1/tiltrun_types.go
```diff
@@ -0,0 +1,274 @@
...
+	// ExitCondition defines the criteria for Tilt to exit.
+	ExitCondition ExitCondition `json:"exitCondition"`
```
milas, line 60: I'm wondering if going towards "Tilt as a daemon" I should change this verbiage to something like `DoneCondition` as it might not _necessarily_ trigger an "exit" in the traditional sense.

pkg/apis/core/v1alpha1/tiltrun_types.go
```diff
@@ -0,0 +1,274 @@
...
+
+// TiltRunStatus defines the observed state of TiltRun
+type TiltRunStatus struct {
+	// PID is the process identifier for this instance of Tilt.
+	PID int64 `json:"pid"`
+	// StartTime is when the Tilt engine was first started and began processing resources.
+	StartTime metav1.MicroTime `json:"startTime"`
+	// Tiltfile is the current state of the Tilt engine including details about the last time the Tiltfile was
+	// processed and any build errors (e.g. syntax errors) encountered.
+	Tiltfile TiltfileState `json:"tiltfile"`
```
milas, line 151: A couple things that are kind of intertwined here:
* I kinda stuck with the way engine handles it where `Tiltfile` is special/top-level and not in the "user resources", but I dunno if that actually makes sense or if this should just be a known key in `Resources` slice?
* Going with the previous comment about not necessary _needing_ a `Tiltfile` in the future, this would then need to be `nil`-able and kind of awkward, which further leads me to thinking it should just be an entry in `Resources`

milas: One "strange" thing about mixing `Tiltfile` w/ the rest of the resources is that it's the exception to the "everything has runtime but not necessarily a build" as it's purely a build.

@nicks mentioned recently possibly getting rid of "build" as a concept (at this level), and I'm becoming more inclined to give that a shot. We could split resources (which is desirable from a DAG perspective anyway I think), so you would have `foo:build` and `foo:serve` resources.

pkg/apis/core/v1alpha1/tiltrun_types.go
```diff
@@ -0,0 +1,274 @@
...
+	// RuntimeStatusRunning indicates that the resource is currently executing.
+	RuntimeStatusRunning RuntimeStatus = "running"
+	// RuntimeStatusSucceeded indicates that the resource ran to completion successfully.
+	//
+	// This is only used for a RuntimeType of "job"; "server" resources never exit during normal operation.
+	RuntimeStatusSucceeded RuntimeStatus = "succeeded"
+	// RuntimeStatusFailed indicates that the resource encountered an error during execution.
+	RuntimeStatusFailed RuntimeStatus = "failed"
+	// RuntimeStatusInactive indicates that the resource has been requested to not execute at this time.
+	RuntimeStatusInactive RuntimeStatus = "inactive"
```
milas, line 233: This covers things like `auto_init=False` today and would map well with being able to pause/stop resources from UI in the near-ish future.

That said, it's kind of funky and we could probably get rid of it by splitting resources into build/runtime sub-resources (see my other comment) because then disabled _runtime_ resources would just be "filtered out" and not exist in the list to begin with. (Of course, it's often useful to know "this thing exists but is disabled"... 🤔)

Personally, I don't hate it, but flagging in case anyone finds it unconscionable or has a more elegant solution up their sleeves

pkg/apis/core/v1alpha1/tiltrun_types.go
nicks: Speaking to some of the high-level questions, I think it's reasonable to assume:
- The apiserver will be a persistent process that has a separate lifecycle than the tilt up or tilt ci process
- The Tiltfile is one of many ways to load configuration into that apiserver

And there might be extensions that load configuration differently (like the `updog` command currently)

pkg/apis/core/v1alpha1/tiltrun_types.go
nicks: ExitCondition seems ok to me. This is the Exit condition for _some_ process (for shell scripting), but we can be hand-wavey about which process that is for now

pkg/apis/core/v1alpha1/tiltrun_types.go
```diff
@@ -0,0 +1,274 @@
...
+	ExitCondition ExitCondition `json:"exitCondition"`
+}
+
+type ExitCondition string
+
+const (
+	// ExitConditionManual cedes control to the user and will not exit based on resource status.
+	// This is used by `tilt up`.
+	ExitConditionManual ExitCondition = "manual"
+	// ExitConditionFirstFailure terminates upon the first encountered build or runtime failure.
```
nicks, line 69: i'm not sure I would call this FirstFailure, since it will also exit if all servers are ready and all jobs are complete?

pkg/apis/core/v1alpha1/tiltrun_types.go
nicks: ya, i would be in favor of making the Tiltfile a generic resource, rather than something special (we initially made it special, but over time we've made it more like the other resources)

maybe it helps to think of Tiltfile as a Job that registers other resources on success?

pkg/apis/core/v1alpha1/tiltrun_types.go
```diff
@@ -0,0 +1,274 @@
...
+	// If nil, the resource does not perform builds (for example, a local resource without a serve_cmd).
+	Build *BuildState `json:"build,omitempty"`
+	// Runtime provides information about the current execution of the resource.
+	Runtime RuntimeState `json:"runtime,omitempty"`
+}
+
+// BuildState includes details about a currently pending build, currently active build, and (last) terminated build.
+type BuildState struct {
+	// Pending gives details about the currently enqueued build (if any).
+	Pending *PendingBuild `json:"pending,omitempty"`
```
nicks, line 184: i would probably use the convention `BuildStatePending`, to match the kubernetes naming convention 

https://pkg.go.dev/k8s.io/api/core/v1#ContainerStateWaiting

pkg/apis/core/v1alpha1/tiltrun_types.go
nicks: maybe 'disabled' , since we use the words enable/disable resources elsewhere in the tiltfile api? but ya, it's a good call that we need this for resources that aren't expected to run in CI mode

pkg/apis/core/v1alpha1/register.go
```diff
@@ -54,6 +55,7 @@ func AllResourceObjects() []resource.Object {
 }
 func AllResourceLists() []runtime.Object {
 	return []runtime.Object{
+		&TiltRun{},
```
nicks, line 58: i think this should be TiltRunList?

nicks (approved):
LGTM

some minor naming questions

pkg/apis/core/v1alpha1/tiltrun_types.go
milas: That's a great point 🤔 

One possible change would be just make this a mouthful but more descriptive (`first-failure-or-all-succeeded`).

I briefly thought about splitting this into `FailureCondition` + `SuccessCondition` but you can come up with weird combinations (e.g. success of "all succeeded" but failure of "never" is probably _not_ what you want).

I'm now wondering if this could be something like `[]ExitCondition ExitConditions` (where the list is OR'd, i.e. first to hit) and then have a corresponding status field of `TriggeredExitCondition`. For `tilt ci`, you'd have `spec.exitConditions: ["first-failure", "all-successful"]`.

That said, it still introduces awkwardness with `manual` (or maybe more accurately that should be called `user-signal`) - it's _always_ a valid exit condition, so in the `tilt up` case, should the exit condition list just be nil/empty?

nicks: i honestly think ExitConditionCI, since it has CI-like semantics, and "CI" is the term of art people are going to be searching for when they want this

PR #4379, opened 2021-03-29
web: add resource name filter

landism:
Notes:
* The changes with shortcut keys and "isEditable" to disable shortcut keys while you're typing in an <input> or <textarea>. Without this change, if you typed 's' in the filter box, the snapshot modal would pop up.
* I don't like the name `preparePersistedState` and would welcome suggestions. TL;DR. The problem that led to its creation: if you ran Tilt with a previous version, you'd have a `SidebarOptions` persisted w/ no `resourceNameFilter` field. The web UI would happily deserialize it into a `SidebarOptions` and `resourceNameFilter` would be `undefined` and things would blow up. The two main alternatives considered:
  * make `resourceNameFilter` an optional field. This is annoying because then everywhere that accesses it has to worry about this problem
  * rename to something like `maybeUpgradePersistedState`


web/src/LocalStorage.tsx
```diff
@@ -34,20 +34,35 @@ export function accessorsForTesting<S>(name: string) {
 }
 
 // Like `useState`, but backed by localStorage and namespaced by the tiltfileKey
-export function usePersistentState<S>(name: string, defaultValue: S) {
+// prepareSavedState: transforms any state read from storage - allows, e.g., filling in default values for
```
maiamcc, line 37: hmm, maybe `normalizeSavedState`? `backfillSavedState`? `validateSavedState`? i don't love any of these (including yours) but don't hate it, the comment is enough to make sure I know what it does

web/src/OverviewSidebarOptions.test.tsx
```diff
@@ -199,6 +203,53 @@ describe("overview sidebar options", () => {
...
+      defaultOptions.testsOnly,
+      defaultOptions.alertsOnTop,
+      "B p"
+    )
+  })
+
+  it("says no matches found", () => {
+    sidebarOptionsAccessor.set({
+      ...defaultOptions,
+      resourceNameFilter: "asdfawfwef",
```
maiamcc, line 233: what a good search term

web/src/SidebarResources.tsx
```diff
@@ -94,6 +106,18 @@ function sortByHasAlerts(itemA: SidebarItem, itemB: SidebarItem): number {
   return Number(hasAlerts(itemB)) - Number(hasAlerts(itemA))
 }
 
+function matchesResourceName(item: SidebarItem, filter: string): boolean {
+  filter = filter.trim()
+  // this is functionally redundant but probably an important enough case to make its own thing
```
maiamcc, line 111: 🙃  is there a test case for this? (i.e. simulate change to "") should there be?

web/src/SidebarResources.tsx
```diff
@@ -150,6 +178,14 @@ export class SidebarResources extends React.Component<SidebarProps> {
         ? "isOverview"
         : ""
 
+    if (listItems.length === 0 && options.resourceNameFilter) {
```
maiamcc, line 181: hmm arguably if you click "tests only" and there are no tests, you should get this message then too? though i'm not actually sure we offer you the "tests only" button in that case... (you don't actually need to change this, just thinking aloud)

maiamcc (approved):
i'll leave the css/styling for @hyu but this lgtm!

web/src/SidebarResources.tsx
landism: yeah, changed the second condition to `this.props.items.length() !== 0` - it seems like in general if there were originally items and we filtered them all out, it's good to say no matches

web/src/OverviewSidebarOptions.test.tsx
landism: carefully crafted to ensure it did not match any of the test data

web/src/LocalStorage.tsx
landism: I think I like `maybeUpgradeSavedState`

web/src/SidebarResources.tsx
landism: ehh, there are a bunch of tests that have an empty filter, and I'm not sufficiently skeptical of material-ui's textfield to think it's worth going out of our way to test the state-change

web/src/OverviewSidebarOptions.test.tsx
maiamcc: bespoke, artisanal test cases

hyu (approved):
Looks good to me, and works as intended!

Also, thanks for fixing my missing semicolons 😬

PR #4386, opened 2021-03-30
engine: consider Unschedulable pods as being in error state

### LINKED ISSUE #4385: Detect when pods are stuck in an Unschedulable state
### Expected Behavior
* If a pod is **never** going to be scheduled, show the resource as having an error and fail `tilt ci`

### Current Behavior
* Resource will show as pending forever in `tilt up` and hang `tilt ci` indefinitely

### Steps to Reproduce
1. Taint your node, e.g. for kind: `kubectl taint nodes kind-control-plane key=value:NoSchedule --overwrite`
2. Run `tilt up` (or `tilt ci`) for a `Tiltfile` that deploys a pod
3. Observe that pod is stuck in pending forever (or `tilt ci` hangs)

### Extra Context
It's worth noting that a first attempt at addressing this via #4386 had to be reverted because it's not atypical for pods to end up in `Unschedulable` **transiently** for different reasons. That said, it'd be helpful to have better heuristics for when the situation isn't expected to imminently resolve and reflect that more usefully as an "error" within Tilt.

### About Your Use Case
With `tilt up`, this is more of a nuisance - we DO show the unschedulable event at least but don't show it as an error (so you might not notice for a while)
For `tilt ci`, it means things hang indefinitely

See also #437 - there used to be some logic around this.

milas:
If a pod is unschedulable for some reason (e.g. resource limits),
mark its status as an error instead of letting it sit in pending
indefinitely.

If the scheduling issue is resolved, the resource (i.e. the pod
from Tilt's perspective) should then transition to OK as usual.

See also #437 which originally introduced a check for this;
while the code was deleted in #4365, it was **dead code** at that
point, so this regressed at some indeterminate point in the past
when we stopped relying on watching individual containers.
(I'm assuming TBH - I don't actually know _how_ that container
watch code was used originally! 😬)

Closes #4385.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -90,7 +90,12 @@ jobs:
       - run: echo 'export PATH=/go/bin:$PATH' >> $BASH_ENV
       - setup_remote_docker:
           version: 19.03.12
-      - run: ctlptl create cluster kind --registry=ctlptl-registry && make integration
+      - run:
+          name: Execute integration tests
+          command: |
+            ctlptl create cluster kind --registry=ctlptl-registry
+            kubectl wait --for=condition=ready node --all --timeout=60s
+            make integration
       - store_test_results:
           path: test-results
       - slack/notify-on-failure:
@@ -104,7 +109,12 @@ jobs:
       - run: echo 'export PATH=/go/bin:$PATH' >> $BASH_ENV
       - setup_remote_docker:
           version: 19.03.12
-      - run: ctlptl create cluster kind --registry=ctlptl-registry && make install test-extensions
+      - run:
+          name: Execute extension tests
+          command: |
+            ctlptl create cluster kind --registry=ctlptl-registry
+            kubectl wait --for=condition=ready node --all --timeout=60s
+            make install test-extensions
       - slack/notify-on-failure:
           only_for_branches: master
 
--- a/integration/event_test.go
+++ b/integration/event_test.go
@@ -41,7 +41,11 @@ func markNodeSchedulable(f *k8sFixture, name string) {
 		"markNodeSchedulable")
 }
 
-func TestEvent(t *testing.T) {
+// TestUnschedulableEvent is testing two things:
+//	1) Important K8s events (such as unschedulable pods) are propagated to logs (at least until #3532 is done)
+//	2) Unschedulable pods are continued to be monitored by Tilt after the situation is resolved
+//		(see comments on assertions - this is hacky right now!)
+func TestUnschedulableEvent(t *testing.T) {
 	f := newK8sFixture(t, "event")
 	defer f.TearDown()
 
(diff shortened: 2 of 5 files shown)


nicks:
i would probably make the ergonomics of it `ctlptl wait` to match the ergonomics of kubectl, but sgtm

.circleci/config.yml
```diff
@@ -90,7 +90,12 @@ jobs:
       - run: echo 'export PATH=/go/bin:$PATH' >> $BASH_ENV
       - setup_remote_docker:
           version: 19.03.12
-      - run: ctlptl create cluster kind --registry=ctlptl-registry && make integration
+      - run:
+          name: Execute integration tests
+          command: |
+            ctlptl create cluster kind --registry=ctlptl-registry
+            kubectl wait --for=condition=ready node --all --timeout=60s
```
maiamcc, line 97: nice

integration/event_test.go
```diff
@@ -41,7 +41,11 @@ func markNodeSchedulable(f *k8sFixture, name string) {
 		"markNodeSchedulable")
 }
 
-func TestEvent(t *testing.T) {
+// TestUnschedulableEvent is testing two things:
+//	1) Important K8s events (such as unschedulable pods) are propagated to logs (at least until #3532 is done)
+//	2) Unschedulable pods are continued to be monitored by Tilt after the situation is resolved
```
maiamcc, line 46:
```suggestion
//	2) Unschedulable pods continue to be monitored by Tilt after the situation is resolved
```

maiamcc (approved):
🙌

nicks:
alternatively, i wonder if Tilt itself should spend some time waiting for the cluster to come up instead of ctlptl :thinking:

milas:
> alternatively, i wonder if Tilt itself should spend some time waiting for the cluster to come up instead of ctlptl 🤔

Most of the time, there's no reason to wait for cluster nodes to be ready, e.g. if you `ctltptl create ... && tilt up`, it's possible you'd see your resource in error for a moment (with the event showing it as unschedulable) and then it'd auto-resolve once the node finishes coming up.

That said, if you are doing `ctltpl create ... && tilt ci` like we are doing for Tilt integration tests + extension integration tests, you'll hit the same snag and need to add a `kubectl wait` in there. So for that case, it could be nice if Tilt automatically handled it, but it'd need to be restricted to local dev clusters IMO since any external/real cluster could reasonably expected to have some not-ready nodes for whatever reason.

I could at least add some guidance to https://docs.tilt.dev/cli/tilt_ci.html recommending a similar command if you're using a local dev cluster?

nicks:
not sure. 

This is a pretty common CI flow though, I would be very surprised if we were the only ones this change affected.

what if we made it a warning if LastTransitionTime > 0 and an error only if LastTransitionTime > 15s?

milas:
@maiamcc Pushed a small change to add a 15 sec grace period during which you'll see the warning from the K8s event in the log and it'll be in pending. (Thanks @nicks for suggestion!)

maiamcc (approved):
Ooh nice! we'll probably eventually want that grace period to be settable from the Tiltfile, but for now this is 🎉

PR #4389, opened 2021-03-31
web: break apart OverviewResourceBar

landism:
The resource bar is about to go away and these components moved elsewhere. This is some restructuring to reduce the size of the future PRs that contain the actual changes.

1. rename "ResourceBarEnd" to "TiltMenu"
2. rename "ResourceBarStatus" to "ResourceStatusSummary"
3. pull both of the above out to their own files


web/src/TiltMenu.tsx
```diff
@@ -0,0 +1,297 @@
...
+  tiltCloudTeamName: string
+  snapshot: SnapshotAction
+  showUpdate: boolean
+  suggestedVersion: string | null | undefined
+  runningBuild: Proto.webviewTiltBuild | undefined
+  showMetricsButton: boolean
+  metricsServing: MetricsServing | null | undefined
+}
+
+export function TiltMenu(props: TiltMenuProps) {
```
nicks, line 143: maybe call this GlobalNav after https://developer.mozilla.org/en-US/docs/Web/HTML/Element/nav ?

By convention, I would usually interpret `TiltMenu` to be "a generic Menu component with Tilt brand colors", which is not what you want to communicate here (and even then, I think using the product name in the component name is considered bad form)

nicks (approved):

PR #4392, opened 2021-03-31
runtimelog: write podlogstream to apiserver

nicks:
Hello @milas,

Please review the following commits I made in branch nicks/ch11686:

5a659adabb92a9e10631a4c295b1a0edcfb25816 (2021-03-31 14:58:22 -0400)
runtimelog: write podlogstream to apiserver


pkg/apis/core/v1alpha1/podlogstream_types.go
```diff
@@ -139,6 +139,8 @@ func (in *PodLogStreamList) GetListMeta() *metav1.ListMeta {
 }
 
 // PodLogStreamStatus defines the observed state of PodLogStream
+//
+// TODO(nick): rewrite this status field, i don't think this is quite right.
```
milas, line 143: 🤔

milas (approved):

pkg/apis/core/v1alpha1/podlogstream_types.go
nicks: we're streaming logs from different containers (both init containers and normal run containers)

So you could have an "active" init container log stream, but a pending run container log stream

I think it _probably_ makes sense to represent each container's log stream individually in the status, rather than trying to aggregate their statuses

nicks: (i still haven't convinced myself this is the way to go, but also recognize that we don't need to make a decision on this right now)

PR #4393, opened 2021-03-31
web: add headerbar to storybook

landism:
As part of merging tabs + pins, we're pulling resource status + nav menu up to the top bar, where the tabs currently are.

This PR adds a new HeaderBar component that matches this design, but only in the storybook. It won't be included in the actual UI until the new starred resources component is ready.

Screenshots:

image

image

NB: there are some differences in the resource status summary between this and the design in figma:
* no vertical dividing line between resources and tests
* no "pending"
* status counts are only colored if they meet their highlight criteria (for err/warn: # => 1, for healthy: # = 100%)

I figure sorting out the resource status summary is a separate change.
I did take out the explicit status labels (the existing tilt UI says, e.g., "10/10 healthy", and the new heaer bar doesn't show "healthy", since they took up a lot of horizontal space and the header bar felt cramped.

### DIFF
--- a/web/src/HeaderBar.tsx
+++ b/web/src/HeaderBar.tsx
@@ -1,14 +1,14 @@
 import React from "react"
+import { Link } from "react-router-dom"
 import styled from "styled-components"
 import { ReactComponent as LogoWordmarkSvg } from "./assets/svg/logo-wordmark.svg"
 import { GlobalNav } from "./GlobalNav"
 import { usePathBuilder } from "./PathBuilder"
 import { ResourceStatusSummary } from "./ResourceStatusSummary"
 import { useSnapshotAction } from "./snapshot"
-import {AnimDuration, Color, Font, FontSize, SizeUnit} from "./style-helpers"
+import { AnimDuration, Color, Font, FontSize, SizeUnit } from "./style-helpers"
 import { TargetType } from "./types"
 import { showUpdate } from "./UpdateDialog"
-import { Link } from "react-router-dom"
 
 const HeaderBarRoot = styled.div`
   display: flex;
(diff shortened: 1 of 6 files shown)


web/src/HeaderBar.tsx
```diff
@@ -0,0 +1,98 @@
...
+  const pb = usePathBuilder()
+
+  return (
+    <HeaderBarRoot>
+      <Link to='/overview'>
+        <Logo width="57px" />
+      </Link>
+      <HeaderDivider />
+      <AllResourcesLink to={pb.encpath`/r/(all)/overview`} >All Resources</AllResourcesLink>
+      <ResourceStatusSummaryContainer>
```
landism, line 92: `ResourceStatusSummary` needs `flex-grow: 1`. Previously, we put this on `ResourceStatusSummaryRoot`, but that's kind of awkward since it prevents its parent from controlling where its child goes. (ditto for `GlobalNav` in `OverviewResourceBar`). It seems good for `HeaderBar` to control the layout of the `HeaderBar` flexbox.

I didn't see an existing pattern for this, so I took a stab and added a couple `____Container` classes to control their childrens' layout and see how it feels.

Another couple options:
* expose `style` properties on components and then do `<ResourceStatusSummary style={{flexGrow: 1}}>`
* instead of defining a `ResourceStatusSummaryContainer` const, just stick an inline `<div style={{flexGrow: 1}}>` right here. This is kind of appealing since the const isn't really doing that much for us, but it does feel kind nice to keep this a css-free zone.

This is probably a bit bikesheddy and I don't think I have a strong preference and it I don't see a reason it matters much as long as we have some kind of convention (and even then, that won't matter unless the need for this comes up more often).

landism: realized I could remove these elements and deal with layout from the parent's styled component, like this:
https://github.com/tilt-dev/tilt/blob/f0fefeef99e0eba00b43464e9cdc22774a940189/web/src/HeaderBar.tsx#L21-L25

nicks (approved):

PR #4394, opened 2021-03-31
engine: make podlogmanager a reconciler

nicks:
Hello @milas,

Please review the following commits I made in branch nicks/ch11687:

ca5eb0f06b552c40a74134f18ef5a8079d8ca1de (2021-03-31 19:38:50 -0400)
engine: make podlogmanager a reconciler


internal/engine/runtimelog/podsource.go
```diff
@@ -0,0 +1,155 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+
+	"github.com/tilt-dev/tilt/internal/k8s"
+	"github.com/tilt-dev/tilt/pkg/logger"
+)
+
+// Helper struct that captures Pod changes and queues up a Reconcile()
```
nicks, line 18: there's definitely some big abstraction here that I'm missing

The controller-runtime has great tools for "if object X creates object Y, then trigger a Reconcile of X every time Y changes".

I feel like we have a common pattern of "object X has a field that references object Y, which means we need a reconcile every time Y change", and then need lots of custom mapping code to do that efficiently.

not sure what the right answer is yet...

milas: I get the sense that we might be working against the desired pattern here - since owner references are actually stored on the child, that's how it can map back efficiently on the `.Owns()` (which is more or less syntactic sugar over `.Watches()`).

The Kubebuilder book does have an example of adding an index on a field for client lookups, so I guess that _could_ be a possibility - index your pod field and then in the map func when the pod changes, look up the `PodLogStream`s using the index and emit the requests based off hits.

internal/engine/runtimelog/podlogmanager_test.go
```diff
@@ -250,7 +258,7 @@ func TestLogReconnection(t *testing.T) {
 
 	_, _ = writer.Write([]byte("hello world!"))
 	f.AssertOutputContains("hello world!")
-	assert.Equal(t, startTime, f.kClient.LastPodLogStartTime)
+	assert.Equal(t, startTime.Truncate(time.Second), f.kClient.LastPodLogStartTime)
```
milas, line 261: Just curious - is this result of (intentionally) reduced precision or something else?

milas (approved):

internal/engine/runtimelog/podsource.go
nicks: it's even more complicated than that, because the Pod comes from a different API server than the PodLogStream

but I think you're right that the long-term goal should be to have something like Owns() that handles all of this generically....there's nothing really Pod-specific here.

internal/engine/runtimelog/podlogmanager_test.go
nicks: ya, this is not documented super-well, but metav1.Time only provides second-level granularity https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1#Time.Rfc3339Copy
so all the underlying Log apis truncate this anyway

this test changed because we're serializing the API, which is where the truncation happens

nicks:
i'm going to wait on https://github.com/tilt-dev/tilt-apiserver/pull/29 before merging (though tbh i don't totally understand why that bug doesn't affect things now)

PR #4396, opened 2021-04-01
build: attach name of pipeline step to summary

maiamcc:
@theothertomelliott asked what the steps a la--
```
Step 1 - 5s
Step 2 - 1s
Step 3 - 0.1s
```
in the web UI corresponded to.

these are steps in our update pipeline (e.g. "building dockerfile [gcr.io/xyz]", "Deploying to k8s"), which ARE given names at some point, but you need to scroll back up through the logs to associate the step and its name with the duration given below

this PR attaches the step name to its duration when we print the pipeline summary

I don't feel great about how the way we store `pipelineStep` and `buildStep` have diverged, but I figure this is fine for now

cc @hyu for "does this look ugly" consult


maiamcc:
Screen Shot 2021-04-01 at 5 46 22 PM

internal/build/pipeline_state.go
```diff
@@ -59,27 +70,35 @@ func (ps *PipelineState) End(ctx context.Context, err error) {
...
-	ps.curPipelineStep++
+	stepName := fmt.Sprintf(format, a...)
+	ps.curPipelineStep = PipelineStep{
+		Name:  stepName,
+		Index: ps.curPipelineIndex() + 1,
+	}
+	line := logger.Blue(l).Sprintf("STEP %d/%d", ps.curPipelineIndex(), ps.totalPipelineStepCount)
+	l.Infof("%s — %s", line, stepName)
 	ps.curBuildStep = 1
 	ps.curPipelineStepStart = ps.c.Now()
```
landism, line 95: wanna pull this field into the `PipelineStep` struct?

internal/build/pipeline_state.go
```diff
@@ -35,7 +46,7 @@ const buildStepOutputPrefix = "     "
 
 func NewPipelineState(ctx context.Context, totalStepCount int, c Clock) *PipelineState {
 	return &PipelineState{
-		curPipelineStep:        1,
+		curPipelineStep:        PipelineStep{Index: 0},
```
landism, line 49: IIUC it's idiomatic go to omit initializations to 0-values like this (both the `Index` and the `PipelineStep` here are such initializations)

(this is irrelevant if you agree with my below comment to delete `curPipelineStep`)

web/src/subdir/index.test.js
```diff
@@ -0,0 +1,3 @@
+test("everything is okay", () => {
```
landism, line 1: 👀

landism (approved):
I like this change!

internal/build/pipeline_state.go
```diff
@@ -59,27 +70,35 @@ func (ps *PipelineState) End(ctx context.Context, err error) {
...
 	l := logger.Get(ctx)
-	line := logger.Blue(l).Sprintf("STEP %d/%d", ps.curPipelineStep, ps.totalPipelineStepCount)
-	l.Infof("%s — %s", line, fmt.Sprintf(format, a...))
-	ps.curPipelineStep++
+	stepName := fmt.Sprintf(format, a...)
+	ps.curPipelineStep = PipelineStep{
+		Name:  stepName,
+		Index: ps.curPipelineIndex() + 1,
+	}
+	line := logger.Blue(l).Sprintf("STEP %d/%d", ps.curPipelineIndex(), ps.totalPipelineStepCount)
```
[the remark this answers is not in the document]
maiamcc: yyyup good call

PR #4398, opened 2021-04-02
web: add starred resource bar to storybook

landism:
image

This is just the component + story - it's not yet actually used by the UI.

The overflow behavior is just to keep expanding vertically, which isn't ideal, but 1) isn't awful and 2) only affects users who star resources, and star many of them.


web/package.json
```diff
@@ -26,6 +26,7 @@
     "react-dom": "^16.9.0",
     "react-modal": "^3.10.1",
     "react-outline-manager": "^1.2.2",
+    "react-priority-navigation": "^1.0.2",
```
nicks, line 29: would it be possible to use a material-ui component for this?

i really don't want to have a hodge podge of random UI libraries with different styling conventions and mixed support for a11y, etc. I poked at this component and it doesn't seem to be actively maintained

landism: > i really don't want to have a hodge podge of random UI libraries with different styling conventions and mixed support for a11y, etc. I poked at this component and it doesn't seem to be actively maintained

Yeah, I'm not thrilled with this component. Probably 80% of the time I've spent on this epic has been trying to find a decent component to power the overflow menu.

I figured I'd spent too much time looking for existing options already and was on the fence between using this component, implementing our own, or just giving up on overflow and wrapping around to more lines. I figured using the component for now was the best option since it gives us the functionality now for ~free and we can choose another option later without much risk of its cost increasing in the meantime.

> would it be possible to use a material-ui component for this?

I would strongly prefer to! I looked through all of its components last week to see if any of them would fit and came up empty-handed. I'll dig a bit deeper.

nicks: > I figured using the component for now was the best option since it gives us the functionality now for ~free and we can choose another option later without much risk of its cost increasing in the meantime.

Hahaha, this is like the number one fallacy of UI engineering. It's a big part of what people are trying to push back against when they talk about "design systems" and "component libraries".

in practice, future engineers and designs will have to figure out how to this component fits in with any adjacent components -- and it's not really fair to ask them to clean up after you here.

though this is fundamentally a design system problem, so let me have a quick chat with Surbhi about what she'd like to do here.

landism: ok, dropped the overflow menu for now, which doesn't seem that awful

PR #4402, opened 2021-04-05
test: fix a flaky podlogstream test

nicks:
Hello @milas, @maiamcc,

Please review the following commits I made in branch nicks/flake:

08abb7fa9960a0a72b3ae2303650bdd70407d878 (2021-04-05 18:11:06 -0400)
test: fix a flaky podlogstream test
I think this is a good example of how having a public status
field will make this a bit easier to test in general

### DIFF
--- a/internal/engine/runtimelog/podlogmanager_test.go
+++ b/internal/engine/runtimelog/podlogmanager_test.go
@@ -100,10 +100,7 @@ func TestLogsFailed(t *testing.T) {
 	assert.Contains(t, f.out.String(), "my-error")
 
 	// Check to make sure the status has an error.
-	stream := &PodLogStream{}
-	streamNN := types.NamespacedName{Name: fmt.Sprintf("default-%s", podID)}
-	err := f.client.Get(f.ctx, streamNN, stream)
-	require.NoError(t, err)
+	stream := f.getPodLogStream(podID)
 	assert.Equal(t, stream.Status, PodLogStreamStatus{
 		ContainerStatuses: []ContainerLogStreamStatus{
 			ContainerLogStreamStatus{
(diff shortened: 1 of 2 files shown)


milas (approved):

milas:
Heads up that `TestLogs` is flaky for me (both before/after this PR) - issue looks like that because the logs are a `strings.Reader`, there's a race where it consumes them, and comes back around to the `isActive` check and then uses the `lastTerminationTime` which is NOT truncated to the second, so the time comparison fails: https://github.com/tilt-dev/tilt/blob/c847167f683ccb59d07e98773cb02659f382b429/internal/engine/runtimelog/podlogmanager_test.go#L55

```
=== Failed
=== FAIL: internal/engine/runtimelog TestLogs (0.01s)
    podlogmanager_test.go:55:
        	Error Trace:	podlogmanager_test.go:55
        	Error:      	Not equal:
        	            	expected: time.Time{wall:0x0, ext:63753328595, loc:(*time.Location)(0x361e400)}
        	            	actual  : time.Time{wall:0xc0134614e106c1f8, ext:52941950, loc:(*time.Location)(0x361e400)}

        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1,4 +1,4 @@
        	            	 (time.Time) {
        	            	- wall: (uint64) 0,
        	            	- ext: (int64) 63753328595,
        	            	+ wall: (uint64) 13840483135328272888,
        	            	+ ext: (int64) 52941950,
        	            	  loc: (*time.Location)({
        	Test:       	TestLogs

DONE 1 tests, 1 failure in 6.000s
```

Can probably fairly quickly hit it with `gotestsum ./internal/engine/runtimelog -run '^TestLogs$' -count 100 -failfast`.

I think in theory we should probably use an `io.Pipe` or some other non-closing stream across the tests except when we're actually testing that behavior (e.g. `TestLogsCanceledUnexpectedly` + `TestLogReconnection`).

nicks:
hmmm...let's fix that separately. i like your idea of fixing this in fakeClient

PR #4403, opened 2021-04-06
web: report analytics on sidebar name filter

landism:


web/src/OverviewSidebarOptions.tsx
```diff
@@ -118,6 +119,11 @@ export const ResourceNameFilterTextField = styled(TextField)`
   }
 `
 
+export const ClearResourceNameFilterButton = styled(CloseSvg)`
```
nicks, line 122: this should be a button with an SVG inside, take a look at how PinButton or SidebarTriggerButton structures the HTML here. we should probably have some generic components for this eventually

nicks (approved):

PR #4405, opened 2021-04-07
session: rename TiltRun to Session

milas:
This is a purely mechanical rename of the `TiltRun` type to `Session`.
Similarly, `exit.Controller` is now `session.Controller` to reflect
its expanded responsibilities.


internal/engine/session/controller.go
```diff
@@ -17,18 +17,26 @@ import (
...
-	tiltrun "github.com/tilt-dev/tilt/pkg/apis/core/v1alpha1"
+	session "github.com/tilt-dev/tilt/pkg/apis/core/v1alpha1"
 )
 
-// Controller handles normal process termination. Either Tilt completed all its work,
-// or it determined that it was unable to complete the work it was assigned.
+// Controller summarizes engine state of resources for the active Tilt session (i.e. invocation of up/ci).
+//
+// Part of the Session spec includes an exit condition, which is evaluated here and reflected on the Session status.
+// Additionally, the engine is signaled to exit (either with success or with error) based on the result of the exit
```
maiamcc, line 26: hm, nbd but right now it reads like the signaling-to-exit happens in this controller, which it doesn't (it happens in the existing control loop of the engine, right?), maybe clarify WHERE this happens or leave it out.

maiamcc (approved):

internal/engine/session/controller.go
milas: Not me forgetting that I moved that logic out of here 🙈 

(Good catch - updated the comment to reflect reality)

PR #4406, opened 2021-04-07
server: listen on the memory connection only

nicks:
Hello @milas,

Please review the following commits I made in branch nicks/memconn:

0bc947e0cea346dd57f4add56b442302687495ae (2021-04-07 00:00:22 -0400)
server: listen on the memory connection only
this will temporarily break 'tilt alpha get'
while we transition to the new system


milas (approved):
FYI integration test failure looks like it was due to OpenAPI being out of sync for `PodLogStream`, but `master` is back in sync if you rebase/merge

PR #4408, opened 2021-04-07
test: add a check to ensure codegen is in sync as part of CI

milas:
This script tries to be smart and only run if the commit changed something under `pkg/apis` or `pkg/openapi` so it won't slow down CI in most cases.

If there was a modified file, it runs the codegen script and ensures there's no modified files in the repo afterwards.

### DIFF
--- a/scripts/check-codegen.sh
+++ b/scripts/check-codegen.sh
@@ -1,0 +1,60 @@
+#!/usr/bin/env bash
+
+# This script ensures that codegen files are up-to-date based on changes to HEAD as
+# compared to origin/master.
+#
+# Because the codegen script is pretty slow, it bails early if no files in pkg/apis
+# (or pkg/openapi) were modified.
+#
+# If there were changed files, the codegen script is run and if there are any uncommitted
+# files as a result, it fails with an error.
+
+set -e
+
+dir=$(dirname "$0")
+cd "${dir}/.."
+
+codegen_regex="pkg/apis|pkg/openapi"
+
+function print_file_list() {
+    while IFS= read -r line
+    do
+      if [[ $line =~ $codegen_regex ]]; then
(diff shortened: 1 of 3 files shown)


scripts/check-codegen.sh
```diff
@@ -0,0 +1,60 @@
...
+    while IFS= read -r line
+    do
+      if [[ $line =~ $CODEGEN_REGEX ]]; then
+        printf "  - %s\n" "${line}"
+      fi
+    done < <(printf '%s\n' "${*}")
+}
+
+changes=""
+if [[ -n "${CIRCLECI}" ]]; then
```
nicks, line 27: hmmmm...would it be simpler to simply verify if any files are modified at the beginning?

not totally sure we need circleci-specific logic here

milas:
Gonna have to rethink this a bit - CircleCI doesn't run in `GOPATH` by default, and it looks like the magic they used to do to symlink into it doesn't exist anymore. (And doesn't seem worth it to me to try to set up a bunch of temporary symlinks for this step?)

Running via Docker is probably the right call to sidestep all that, but local volume + remote Docker means some changes are needed there 😵

nicks:
ya, running the k8s codegen stuff is a pain :\

milas:
Sample intentional failure output:
```sh
#!/bin/bash -eo pipefail
./scripts/check-codegen.sh
Found changed API files (compared to origin/master):
  - pkg/apis/core/v1alpha1/session_types.go

Running codegen to ensure up-to-date...

Found out of sync codegen files:
  - pkg/openapi/zz_generated.openapi.go

Run make update-codegen locally and push the changes.

Exited with code exit status 1
CircleCI received exit code 1
```

nicks (approved):

PR #4413, opened 2021-04-08
Revert "engine: consider Unschedulable pods as being in error state"

### LINKED ISSUE #4412: unexplained tilt ci exits
We've gotten reports of `tilt ci` exiting early like so:

```
Error: Pod <podname> in error state: Pending
```

It's not totally clear if `tilt ci` is legitimately exiting but just reporting it poorly, or if there's a bug in how it's deciding to exit.

milas:
Unfortunately, this is too aggressive and can cause `tilt ci` to fail even when everything _will_ come up successfully but isn't doing so fast enough.

The behavior we're reverting to is still technically correct - the pod _is_ pending. We'd like to frame that as an error if it's going to be waiting indefinitely, but we can't accurately make that determination with the state available.

Perhaps this would be better addressed as an optional flag/config for `tilt ci` of a "max resource startup duration" that could generically cause it to fail if a resource is still in pending after some user-determined amount of time to avoid a `tilt ci` process that runs indefinitely but avoid pitfalls like the one encountered here.

Reverts tilt-dev/tilt#4386.

Closes #4412.


nicks (approved):

maiamcc (approved):
F

PR #4414, opened 2021-04-08
api-server: generate types/boilerplate for PortForward [ch11822]

maiamcc:
Hello @nicks, @milas,

Please review the following commits I made in branch maiamcc/port-fwd-obj:

056514b0cf6d778e2dc33d357caa689367d8ebed (2021-04-08 15:24:11 -0400)
api-server: generate types/boilerplate for PortForward [ch11822]


pkg/apis/core/v1alpha1/portforward_types.go
```diff
@@ -0,0 +1,144 @@
...
+	// The name of the pod to port forward to/from. Required.
+	Pod string `json:"pod,omitempty"`
+
+	// The namespace of the pod to port forward to/from. Defaults to the kubecontext default namespace.
+	//
+	// +optional
+	Namespace string `json:"namespace,omitempty"`
+
+	// The port on the Kubernetes pod to connect to.
+	ContainerPort int
```
milas, line 65: Missing `json` struct tags from here down

pkg/apis/core/v1alpha1/portforward_types.go
```diff
@@ -0,0 +1,144 @@
...
+	metav1.TypeMeta `json:",inline"`
+	metav1.ListMeta `json:"metadata,omitempty"`
+
+	Items []PortForward `json:"items"`
+}
+
+// PortForwardSpec defines the desired state of PortForward
+type PortForwardSpec struct {
+	// The name of the pod to port forward to/from. Required.
+	Pod string `json:"pod,omitempty"`
```
milas, line 57: FYI I believe if you omit `omitempty`, apiserver will automatically enforce that it's non-empty.

pkg/apis/core/v1alpha1/portforward_types.go
```diff
@@ -0,0 +1,144 @@
...
+	}
+}
+
+func (in *PortForward) IsStorageVersion() bool {
+	return true
+}
+
+func (in *PortForward) Validate(ctx context.Context) field.ErrorList {
+	// TODO(user): Modify it, adding your API validation here.
+	// TODO(maia): verify that Pod is populated, ContainerPort and LocalPort are
```
milas, line 109: FYI whenever you add validation, you'll need to add a blessed sample here:
https://github.com/tilt-dev/tilt/blob/77ae37f8f66c4b2d6df5eb526bd71954be73ecf2/internal/hud/server/apiserver_test.go#L40-L49

milas (approved):
Looks good to me! Just a small note about JSON struct tags

matt-psaltis:
Will this support k8s kind: Service as well? Would love the ability to have the port forward at the service level rather than the pod so that it survives pod restarts and multi replica scenarios.

nicks:
@matt-psaltis good question! Our plan is to break this into separate APIs -- one api that simply drives the portforwarder, another api that update the port forward in response to pod restarts / new pods. let me add that to its own feature request thread so that we don't miss it - https://github.com/tilt-dev/tilt/issues/4422

PR #4415, opened 2021-04-08
web: add InstrumentedButton, instrument some buttons

landism:
### Changes in this PR
1. Add an `InstrumentedButton` class, which is just a `<button>` that reports clicks to analytics
2. Change most existing `<button>`s to `InstrumentedButton`s. Some of these buttons already had analytics and some did not.
3. For buttons that have tests, also make sure their analytics are tested.

### Notes
1. Some of the existing analytics was attached to the api call rather than the ui component (e.g., clearLogs, triggerUpdate). Changing this might be controversial. Maybe I'm just rationalizing the use of this new component, but I think attaching it to the ui component means we have some duplication of name/tags, but it also means a) each component controls how it is recorded and b) it's easier to find uninstrumented components (e.g., for now, cmd+f styled.button).
2. I've found myself wanting guidelines for event names, but that's a yak-shave.


web/src/analytics_test_helpers.ts
```diff
@@ -1,6 +1,16 @@
 import fetchMock from "jest-fetch-mock"
 import { Tags } from "./analytics"
 
+// at some point we might want to be clever about this and only mock the responses of calls to /api/analytics
```
nicks, line 4: seems like this is likely to bite us, and the fix for it would be shorter than this comment?
https://www.npmjs.com/package/jest-fetch-mock#conditional-mocking
tho i could be missing something

nicks (approved):

landism:
I've switched our tests from jest-fetch-mock to fetch-mock. They have similar numbers of weekly downloads and github stars, but it looks like fetch-mock seems to have three advantages:
1. With jest-fetch-mock, I see no way to define multiple routes in different calls. Each call to a mock function throws out any previous defined behavior. This means we'd need to come up with our own api to define analytics mocks in one place and test-specific mocks in another.
2. With jest-fetch-mock, if a test causes an unmocked fetch to be called, you get an error saying absolute paths are not supported (because it fell through to the original fetch implementation), but it doesn't say what the path is, and the path can be annoying to track down.
3. fetch-mock's api leads the user to match specific routes rather than all routes. With jest-fetch-mock, most of our tests were just matching all routes, which means a test might generate more fetches than expected and unintentionally mock them and allow them to succeed.

(fetch-mock's docs also point the reader at [fetch-mock-jest](https://github.com/wheresrhys/fetch-mock-jest) (which is not at all easy to confuse with jest-fetch-mock), but that 1) has much lower usage and 2) doesn't clearly add anything of value to this PR)

web/src/analytics_test_helpers.ts
landism: it looks like jest-fetch-mock's behavior with multiple calls is not great, so I've switched our tests to fetch-mock, which feels a lot nicer! Discussed further in [another comment](https://github.com/tilt-dev/tilt/pull/4415#issuecomment-817915482)

web/src/analytics_test_helpers.ts
```diff
@@ -1,12 +1,21 @@
-import fetchMock from "jest-fetch-mock"
+import fetchMock from "fetch-mock"
 import { Tags } from "./analytics"
 
+export function mockAnalyticsCalls() {
+  fetchMock.reset()
```
nicks, line 5: if you have multiple mock helpers, it seems like it would be bad if they all called fetchMock.reset() when they set up?

nicks (approved):

web/src/analytics_test_helpers.ts
landism: good poinit

PR #4417, opened 2021-04-09
vendor: latest tilt-apiserver and default to TLS connections

nicks:
Hello @milas,

Please review the following commits I made in branch nicks/vendor:

eee91baf2f74e61620882b07263f3929220dc229 (2021-04-09 15:32:23 -0400)
vendor: latest tilt-apiserver and default to TLS connections

0cdf20bee5b72e30fd8a8308b45b880044f6d6f6 (2021-04-09 15:20:01 -0400)
test: optimize engine tests a bit


nicks:
this makes the engine tests a bit slower, but i don't totally understand why yet.

i know a lot of the kubebuilder docs recommends a testing convention where you use ginkgo to create a new apiserver at the start of the package and keep it up in between tests, though i'm not sure if there's an easy way to do this without a major test refactor

nicks:
ugh, i think i see the problem. generating certs is very expensive and adds a lot of overhead per test. i think i need to refactor this api to generate the certs once

nicks:
OK, let's try this again! now with test certs in testing

milas (approved):

PR #4425, opened 2021-04-13
web: rename TabNav to ResourceNav and cut all the tab stuff

landism:
Cleaning up a bunch of stuff that we no longer need now that we don't have tabs.

I went a bit down the path of just removing the concept of a nav context entirely, but this seems better than trying to get history + pathbuilder into everywhere that needs it, and is still a step in that direction if we decide to do that in the future.

The TabNav.tsx/ResourceNav.tsx edits are a separate commit in this branch, if you want to see the changes to that file (since the branch diff didn't detect it as a rename).


web/src/ResourceNav.tsx
```diff
@@ -0,0 +1,78 @@
...
+})
+
+export function useResourceNav(): ResourceNav {
+  return useContext(resourceNavContext)
+}
+
+export let ResourceNavContextConsumer = resourceNavContext.Consumer
+export let ResourceNavContextProvider = resourceNavContext.Provider
+
+export function ResourceNavProvider(
```
nicks, line 34: seems like this new provider should have tests?

nicks (approved):

PR #4426, opened 2021-04-13
session: fix a bug that led to infinite loops on linux

nicks:
Hello @milas,

Please review the following commits I made in branch nicks/ch11871:

86d5057602802d5518b18919419851defaaa60e8 (2021-04-13 18:09:16 -0400)
session: fix a bug that led to infinite loops on linux

### DIFF
--- a/internal/controllers/fake/client.go
+++ b/internal/controllers/fake/client.go
@@ -1,0 +1,52 @@
+package fake
+
+import (
+	"context"
+	"encoding/json"
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	ctrlclient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// Wrap the fake client in other assertions we want to make.
+type fakeTiltClient struct {
+	ctrlclient.Client
+}
+
+// NOTE(nick): We've had a lot of bugs due to the way the apiserver
+// modifies and truncates objects on update.
+//
+// We simulate this behavior to catch this class of bug.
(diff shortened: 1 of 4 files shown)


nicks:
i am probably over thinking this but i'm so sick of this time truncation issues!!!

nicks:
i am probably over thinking this but i'm so sick of this time truncation issues!!!

internal/controllers/fake/client.go
```diff
@@ -0,0 +1,52 @@
...
+// Wrap the fake client in other assertions we want to make.
+type fakeTiltClient struct {
+	ctrlclient.Client
+}
+
+// NOTE(nick): We've had a lot of bugs due to the way the apiserver
+// modifies and truncates objects on update.
+//
+// We simulate this behavior to catch this class of bug.
+func (f fakeTiltClient) Update(ctx context.Context, obj client.Object, opts ...client.UpdateOption) error {
```
milas, line 20: Oh nice! I was looking through the fake client to see if it had any useful hooks (spoiler: it doesn't) but this is a good solution

milas (approved):

internal/controllers/fake/client.go
nicks: Yaaaaaah. I think the k8s team has put most of their effort into making it easier to use a real api server in tests, rather than making the fake client a good simulation of what the api server does.

But to make that work out, you also need to buy into a bunch of testing philosophies that our tests don't currently buy into... (ginkgo's per-package setup, ways to clear the server between tests, etc)

We might have to end up investing in our own fake client...

PR #4432, opened 2021-04-14
cli: add tilt alpha apply

nicks:
Hello @milas, @landism,

Please review the following commits I made in branch nicks/ch11865:

b26cb13b9e34a2b518ab7ac0b71cf61dd5fbff14 (2021-04-14 14:41:26 -0400)
cli: add tilt alpha apply

### DIFF
--- a/internal/cli/apply_test.go
+++ b/internal/cli/apply_test.go
@@ -1,0 +1,42 @@
+package cli
+
+import (
+	"bytes"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"k8s.io/apimachinery/pkg/types"
+
+	"github.com/tilt-dev/tilt/pkg/apis/core/v1alpha1"
+)
+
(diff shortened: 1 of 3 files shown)


internal/cli/apply_test.go
```diff
@@ -0,0 +1,34 @@
...
+
+	cmd := newApplyCmd()
+	cmd.streams.Out = out
+	c := cmd.register()
+	err := c.Flags().Parse([]string{"-f", f.JoinPath("sleep.yaml")})
+	require.NoError(t, err)
+
+	err = cmd.run(f.ctx, nil)
+	require.NoError(t, err)
+	assert.Contains(t, out.String(), `cmd.tilt.dev/my-sleep created`)
```
milas, line 33: Similar to delete, might be worthwhile to just check that `f.client.Get()` can find `my-sleep`

milas (approved):

internal/cli/apply_test.go
nicks: done

PR #4433, opened 2021-04-14
web: rename pins to stars

landism:


web/src/StarredResourcesContext.tsx
```diff
@@ -0,0 +1,110 @@
...
+        action: "pin",
+      })
+      return ret
+    })
+  }
+
+  function unstarResource(name: string) {
+    setStarredResources((prevState) => {
+      const ret = prevState.filter((n) => n !== name)
+      incr("ui.web.pin", {
```
nicks, line 91: should this also become ui.web.star?

nicks (approved):

web/src/StarredResourcesContext.tsx
landism: I'd been leaning towards keeping the name consistent to keep queries easier, but given the likelihood of such a query and the low cost of just doing two names, yeah, probably better just to change it.

PR #4436, opened 2021-04-14
tiltfile: add sys.executable

nicks:
Hello @landism,

Please review the following commits I made in branch nicks/ch11874:

3fcb25f3d0b61e866c9338f6d9138287bcfc6ece (2021-04-14 17:02:23 -0400)
tiltfile: add sys.executable
We want to write tilt extensions that use tilt get, tilt apply,
etc to mutate things on the fly, so we need a way for extensions
to reliably find the tilt binary.

### DIFF
--- a/internal/tiltfile/sys/sys.go
+++ b/internal/tiltfile/sys/sys.go
@@ -1,0 +1,53 @@
+package sys
+
+import (
+	"os"
+
+	"go.starlark.net/starlark"
+
+	"github.com/tilt-dev/tilt/internal/tiltfile/starkit"
+)
+
+// The starlark sys module.
+// Contains a subset of Python's sys module.
+// https://docs.python.org/3/library/sys.html
+type Extension struct {
+}
+
+func NewExtension() Extension {
+	return Extension{}
+}
+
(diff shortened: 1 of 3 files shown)


internal/tiltfile/sys/sys.go
```diff
@@ -0,0 +1,53 @@
...
+		values = append(values, starlark.String(arg))
+	}
+
+	list := starlark.NewList(values)
+	list.Freeze()
+	return list
+}
+
+// Full path to the Tilt executable.
+func executable() starlark.Value {
```
landism, line 47: Hm, I guess we should think about moving the `config` vars (`main_dir`, `main_path`, `tilt_subcommand`) into `sys`?

(I don't know if it's just me, but I've always struggled with the distinction between `os` and `sys` in Python, and now we've got that but also throwing in `config` with similar stuff)

landism (approved):

internal/tiltfile/sys/sys.go
nicks: eh, i wouldn't stress too much about it.

i think these made sense under `sys` because of the consistency with python. i think the other ones are less interesting, because they don't really have a python analog

PR #4438, opened 2021-04-14
cli: add tilt alpha create filewatch

nicks:
Hello @milas,

Please review the following commits I made in branch nicks/filewatch:

c19e97125bab67bf458672ed32e8002752e7dca4 (2021-04-14 18:51:40 -0400)
cli: add tilt alpha create filewatch

ed7bbc6a74e6d380594dc598a360f6219fca4875 (2021-04-14 17:32:58 -0400)
cli: add tilt alpha create

### DIFF
--- a/internal/cli/create.go
+++ b/internal/cli/create.go
@@ -82,7 +82,7 @@ func (c *createCmd) register() *cobra.Command {
 	c.options = o
 	c.cmd = cmd
 
-	// TODO(nick): Add subcommands
+	addCommand(cmd, newCreateFileWatchCmd())
 
 	return cmd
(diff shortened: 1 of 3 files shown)


nicks:
I was originally hoping this could be autogenerated from the data models

but the more i thought about it

the more i think it might be better to have very fine-tunes CLIs with better defaults 

tailored to only the data models you'd realistically want to create from the command-line

maybe will revisit auto-generating in the future

milas (approved):

nicks:
updated to remove the 'paths' flag and just specify paths as extra args

Issue #4435, opened 2021-04-14
opening a browser fails with "exit code 3"


nicks:
We received a report that when hitting _space_ at the tilt prompt, they get an error:

```
Opening browser: http://localhost:10350/
Error: exit status 3
```

The underlying browser code is just doing `xdg-open http://localhost:10350/`, exit code 3 means xdg-open wasn't able to find a browser
https://linux.die.net/man/1/xdg-open

but the person reports that xdg-open outside of tilt works ok....hmmmm...

landism:
Merged #4441 to hopefully get some useful output the next time a user encounters this.

PR #4441, opened 2021-04-16
prompt: log output of browser open command

landism:
### Problem

At the Tilt terminal prompt, if `browser.Open` produces any stdout/stderr, it gets swallowed, which might make errors like #4435 unnecessarily mysterious (at least, all the examples I saw on google of xdg-open exiting with status 3 also had output explaining why).

### Solution

fork github.com/pkg/browser so that we don't have to use its global stderr/stdout vars and can redirect its output appropriately

### Notes

The original code we had to redirect browser.openurl's output put it on logger.DebugLvl. For darwin, `open` doesn't output anything on success, so InfoLvl has no downside. I don't have easy access to a linux box to see if the same is true for xdg-open, but as long as it doesn't have a lot of output, it seems a small price to pay.


landism:
(also, after looking at the commit that added it, it made sense that we were setting `browser.Stdout` in the k8s client, but it was sure amusing to find it there!)

internal/browser/browser_linux.go
```diff
@@ -0,0 +1,5 @@
+package browser
+
+func openArgv(url string) []string {
+	return []string{"xdg-open", url}
```
nicks, line 4: i think the real upstream version is more complex now?
https://github.com/pkg/browser/blob/master/browser_linux.go#L9

internal/browser/README.md
```diff
@@ -0,0 +1 @@
+This is a fork of https://github.com/pkg/browser that allows changing stdout/stderr non-globally.
```
nicks, line 1: have you considered having a formal fork live at github.com/tilt-dev/browser? 

i don't think the changes we want here would be unreasonable upstream

nicks (approved):

internal/browser/browser_linux.go
landism: Ah, I was just grabbing it from `vendor`. I'll update.

internal/browser/README.md
landism: I figured I'd try an upstream PR and then make a formal fork if that goes poorly, but I guess I need a fork from which to make a PR anyway.

PR #4442, opened 2021-04-16
session: k8s resource status consistency

milas:
The revised logic for session (which currently is really just used
for `tilt ci` to monitor when it's time to exit) does not share
all the logic with the internal engine due to needing more
granularity.

There was a slight behavior difference here which made it
unintentionally stricter for K8s resources - as long as a Pod is
in the `Running` phase, in the engine model, it's `RuntimeStatusOK`
if all containers are ready and `RuntimeStatusPending` if not.

Some fall-through logic in session meant that it'd end up in a
weird state with both `Active` and `Terminated` with the container
error on `Terminated` which took precedence.

This makes the logic closer to the `K8sRuntimeState::RuntimeStatus()`
and returns early once it determines proper state, eliminating a lot
of the subtlety in the fall-through states.

An additional case is also added to mirror engine logic which is that
anything with `HasEverDeployedSuccessfully == false` will be pushed
to pending regardless of data from the K8s API. This _should_ be
redundant (if nothing has deployed, both `MostRecentPod()` and
`AllContainers()` should be empty, so will hit the fall-back case)
but will at least avoid the possibility of another edge case around
this.

### DIFF
--- a/internal/engine/session/controller_test.go
+++ b/internal/engine/session/controller_test.go
@@ -119,6 +119,56 @@ func TestExitControlCI_FirstRuntimeFailure(t *testing.T) {
 	f.store.requireExitSignalWithError("Pod pod-a in error state due to container c1: ErrImagePull")
 }
 
+func TestExitControlCI_PodRunningContainerError(t *testing.T) {
+	f := newFixture(t, store.EngineModeCI)
+	defer f.TearDown()
+
+	f.store.WithState(func(state *store.EngineState) {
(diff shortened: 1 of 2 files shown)


milas:
FYI to get a better sense of the modified logic, look at diff with whitespace changes ignored: https://github.com/tilt-dev/tilt/pull/4442/files?diff=split&w=1

(A bunch of the logic got nested into a conditional in addition to the early `return`s)

nicks (approved):

PR #4448, opened 2021-04-19
 engine: use new apiserver Pod model

milas:
See beb957c5cb45f24e649614c97b39116225ac11ed (#4445) for similar changes
for `store.Container`.

The TL;DR is that this is being done as a precursor to moving more
K8s logic (Pod discovery) to apiserver to limit the number of almost
the same but slightly different data models being converted and passed
around.

The apiserver Pod data model is almost identical to the store Pod
model aside from renaming a couple fields to better align with the
K8s API names, so there aren't really any logic changes. It is already
simplified from the K8s model in a couple ways (aggregated status
messages & `kubectl`-esque status) that don't compromise the richness
of the data, so are left as-is besides renaming.

Similar to the changes for `store.Container`, the type defs for `pod.Name`
and `pod.Namespace` are no longer used in the API model but still used
throughout other parts of the engine.


nicks (approved):

milas:
Fixed a couple mismatched JSON/field names in 5057ec9d63e1916446395f167c4883e046ee726a and made merged to make git happy

PR #4454, opened 2021-04-20
store: remove Pod alias and use canonical API type

milas:
Purely mechanical change to get rid of the transitional type alias.

See #4448 for the introduction of the type alias and migration of
logic to use the API type rather than the old `store.Pod` type.

Note that there are still some helper methods in `store` that _really_
shouldn't be there, but with the current package dependency tree,
there isn't any easy/obvious place to move them.

As part of broader Pod discovery API server refactoring, they'll find
a better home and get moved around; most of these should probably be
unexported functions that live alongside the logic that depends on
them.

### DIFF
--- a/internal/store/engine_state.go
+++ b/internal/store/engine_state.go
@@ -14,7 +14,7 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 
 	"github.com/tilt-dev/tilt/internal/k8s"
-	filewatches "github.com/tilt-dev/tilt/pkg/apis/core/v1alpha1"
+	"github.com/tilt-dev/tilt/pkg/apis/core/v1alpha1"
 
 	tiltanalytics "github.com/tilt-dev/tilt/internal/analytics"
 	"github.com/tilt-dev/tilt/internal/container"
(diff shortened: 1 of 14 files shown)


internal/store/engine_state.go
```diff
@@ -584,11 +584,11 @@ func (ms *ManifestState) StartedFirstBuild() bool {
 	return !ms.CurrentBuild.Empty() || len(ms.BuildHistory) > 0
 }
 
-func (ms *ManifestState) MostRecentPod() Pod {
+func (ms *ManifestState) MostRecentPod() filewatches.Pod {
```
maiamcc, line 587: hmm now that the v1alpha package contains way more than just filewatches, we should probably not alias it to this (/in general, should always refer to it as `v1alpha` for consistency)

maiamcc (approved):
noice

internal/store/engine_state.go
milas: Ha that's what happens when I let IntelliJ do the refactor for me! I'll remove this alias, thanks for catching.

But yeah, to the broader point, I guess the only time it's probably "appropriate" to alias the import is within a type-specific reconciler/controller package, so I should really never have used it here in the first place. And even within the reconciler package... 🤷 probably not worth it?

milas: 'tis gone as of ee3a48e38

PR #4458, opened 2021-04-21
api: helpers to avoid time granularity comparison issues

milas:
From K8s, there are two time types for apiserver: `metav1.Time` and
`metav1.MicroTime`. The former has second-level granularity when
serialized and the latter microsecond-level granularity when
serialized.

Internally, these are both wrappers around Go stdlib `time.Time` and
only truncate when serialized. This has caused several issues due to
holding onto unserialized versions that are later compared against a
deserialized version resulting in false positives.

This was already largely mitigated by changes in the test fakes (see #4426)
to detect this and return an error. However, as part of migration work,
we often have apiserver entities that _don't_ live in the apiserver
for a period of time, so this doesn't detect that. New constructors
that perform the truncation immediately can help ensure that the
representation is the same upfront as it will be post-serialization.

There is another class of bugs that remains, however, which is that
we often have stdlib times for tracking purposes that we then compare
against apiserver times, and the difference in granularity (precision)
can cause bugs. To help avoid this, a new package, `timecmp`, offers
methods that will normalize the time values to the lowest-precision
of the two before doing a comparison. This has a side benefit that
even when comparing two stdlib `time.Time` values, it strips off the
monotonic clock portion. Because this is never present on Windows
(and possibly other platforms), there are often subtle
platform-specific bugs that this should find sooner.


internal/engine/k8swatch/event_watch_manager.go
```diff
@@ -191,7 +193,7 @@ func (m *EventWatchManager) dispatchEventsLoop(ctx context.Context, ch <-chan *v
 			// TODO(nick): We might need to remove this check and optimize
 			// it in a different way. We want Tilt to be to attach to existing
 			// resources, and these resources might have pre-existing events.
-			if event.ObjectMeta.CreationTimestamp.Time.Before(tiltStartTime) {
+			if !timecmp.AfterOrEqual(event.ObjectMeta.CreationTimestamp, tiltStartTime) {
```
milas, line 196: Realistically, this logic was fine before - you could pretty safely ignore any events that happened the same second that Tilt started. But the stricter construction meant some tests failed, and since they use wall-clock, this was cleaner (and literally faster) than making them `time.Sleep(time.Second)`.

internal/engine/upper.go
```diff
@@ -412,8 +413,8 @@ func handleBuildCompleted(ctx context.Context, engineState *store.EngineState, c
 		}
 
 		bestPod := ms.MostRecentPod()
-		if store.AfterOrEqual(bestPod.CreatedAt.Time, bs.StartTime) ||
-			bestPod.UpdateStartedAt.Time.Equal(bs.StartTime) {
+		if timecmp.AfterOrEqual(bestPod.CreatedAt, bs.StartTime) ||
+			timecmp.Equal(bestPod.UpdateStartedAt, bs.StartTime) {
```
milas, line 417: This was the case that motivated me to do this -- build times are stdlib times while the Pod times are API times, so this comparison was problematic (it started breaking tests with more Pod changes).

internal/timecmp/cmp_test.go
```diff
@@ -0,0 +1,269 @@
...
+		"Microsecond (metav1.MicroTime <> metav1.MicroTime) comparison failed. Values:\n- x: %s\n- y: %s",
+		tc.x.String(), tc.y.String())
+	assert.Equal(t, tc.microsecond, tc.cmp(metav1.NewMicroTime(tc.x), tc.y),
+		"Microsecond (metav1.MicroTime <> stdlib) comparison failed. Values:\n- x: %s\n- y: %s",
+		tc.x.String(), tc.y.String())
+	assert.Equal(t, tc.microsecond, tc.cmp(tc.x, metav1.NewMicroTime(tc.y)),
+		"Microsecond (stdlib <> metav1.MicroTime) comparison failed. Values:\n- x: %s\n- y: %s",
+		tc.x.String(), tc.y.String())
+
+	assert.Equal(t, tc.second, tc.cmp(metav1.NewTime(tc.x), metav1.NewTime(tc.y)),
```
nicks, line 254: can you also add a test for the case with `*metav1.Time`? does that work ok?

nicks: (i have a feeling it will pass the compile check but panic)

internal/timecmp/cmp.go
```diff
@@ -0,0 +1,98 @@
...
+		return a.Truncate(time.Second), b.Truncate(time.Second)
+	}
+
+	_, aMicrosecondsGranularity := a.(metav1.MicroTime)
+	_, bMicrosecondsGranularity := b.(metav1.MicroTime)
+	if aMicrosecondsGranularity || bMicrosecondsGranularity {
+		return a.Truncate(time.Microsecond), b.Truncate(time.Microsecond)
+	}
+
+	// this will need new cases if a new time type is ever introduced
```
maiamcc, line 86: it's not super likely that we'll introduce a new time type anytime soon, but this will be a PITA to find and update if that every happens -- how annoying would it be to do this as a type switch? e.g.:
```
var anySeconds bool
var anyMicroseconds bool
for _, t := range []commonTime{a, b} {
    switch entity.Obj.(type) {
    case metav1.Time:
        anySeconds = true
    case metav1.MicroTime:
        anyMicroseconds = true
    case time.Time:
    default:
        panic("omgwtf")
}

if anySeconds {
    return a.Truncate(time.Second), b.Truncate(time.Second)
} else if anyMicroseconds...
```

maiamcc (approved):
nice!

internal/timecmp/cmp_test.go
milas: Yep, you are 100% correct. Cleaned up the logic + added pointer support (thanks @maiamcc) and added some (non-exhaustive) cases that test pointers of each type as well as pointer<>value.

internal/timecmp/cmp.go
milas: Oooo thank you, this was a great suggestion (especially given what Nick pointed out re: pointer types). I am embarrassed by my previous logic now 🙈

PR #4463, opened 2021-04-22
 engine: use v1alpha.Pod for pod change events

milas:
This is another incremental change; it _mostly_ shifts the conversion
logic out of the action handler. The other biggest change is that it
now completely replaces the tracked Pod objects instead of getting a
reference and mutating it - this shifts some of the logic about
whether to track up a level. This gets the logic closer to its
eventual state as part of a reconciler based on the patterns we've
been following.

The actual action is still a change for a single Pod rather than a
`PodDiscoveryStatus` update, but the changes are large enough for
this chunk that it made since to use this as an increment.

~~(⚠️ This needs the changes from #4458, so I'll rebase once that's merged; diff might be chaotic in the interim)~~

### DIFF
--- a/internal/engine/pod.go
+++ b/internal/engine/pod.go
@@ -18,7 +18,7 @@ import (
 	"github.com/tilt-dev/tilt/pkg/logger"
 )
 
-func handlePodDeleteAction(ctx context.Context, state *store.EngineState, action k8swatch.PodDeleteAction) {
+func handlePodDeleteAction(_ context.Context, state *store.EngineState, action k8swatch.PodDeleteAction) {
 	// PodDeleteActions only have the pod id. We don't have a good way to tie them back to their ancestors.
 	// So just brute-force it.
 	for _, target := range state.ManifestTargets {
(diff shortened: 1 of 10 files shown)


internal/engine/pod.go
```diff
@@ -175,28 +172,15 @@ func maybeTrackPod(ms *store.ManifestState, action k8swatch.PodChangeAction) (*v
...
-		}
-
-		runtime.Pods[podID] = podInfo
-
-		return podInfo, true
 	}
 
-	// CASE 3: This pod is already in the PodSet, nothing to do.
-	return podInfo, false
+	// Return the existing (if any) and track the new version
```
maiamcc, line 177: not sure i follow -- why do we return whatever was previously in the map, instead of the current thing we're tracking?

maiamcc:
FYI I changed the PR to compare to `milas/apiserver-time` so the diff looks reasonable, just remember to change that back to `master` before merging :)

internal/engine/pod.go
milas: The logic before was:
* If already tracked, return ref to that + `isNew: false`
* Otherwise, create stub + `isNew: true`

Then, caller would mutate it a bunch using its event to populate new info, and critically, in the `!isNew` case, look for container restarts by comparing the old and the new.

It was largely this way because the incoming object type (k8s `v1.Pod`) was different than the outgoing object type (`store.Pod`). Now that the incoming + outgoing are both apiserver `v1alpha.Pod`, it makes more sense IMO to just replace the object*, but we still need the old one for that container restart diff logic 😬 

> why do we return whatever was previously in the map, instead of the current thing we're tracking?
So the answer is sorta that previously it was returning both since we just kept mutating it, so would compare containers before overwriting them.

TBH I'm not thrilled with this, but I'm hoping the logic can be further simplified as things move closer to the reconciler model - right now this is both acting as a "reconciler" as well as the downstream consumer, which muddles the waters.


* Ideally we'd treat these as effectively immutable and actually store a `.DeepCopy()` and not let the caller keep mutating it - the only remaining hiccup there is `BaselineRestartCount` which will become the responsibility of the reconciler to populate

maiamcc: Gooootcha. then can you mention the reasoning in a comment? E.g.
```Return the existing pod (if any) so we can see if any containers restarted etc., and track the new version```

File: internal/engine/pod.go [hunk 2]
```diff
@@ -34,49 +34,49 @@ func handlePodChangeAction(ctx context.Context, state *store.EngineState, action
 		return
 	}
 
-	pod := action.Pod
+	podInfo := action.Pod
```
maiamcc, line 37: why `podInfo` and not just `pod`?

internal/engine/pod.go
maiamcc: (and maybe mention the motivation in the comment at the top of the func too)

internal/engine/pod.go
```diff
@@ -148,24 +148,21 @@ func matchPodChangeToManifest(state *store.EngineState, action k8swatch.PodChang
...
+// trackPod adds a Pod to the RuntimeState for a resource.
+//
+// If there are currently tracked Pods and the new Pod's ancestor UID differs from theirs,
+// the tracked Pods will be cleared and the ancestor UID updated. Otherwise, the Pod will
+// be upserted and the prior version returned.
+//
+// A mutable reference to the Pod to be tracked is stored so that further changes to it
+// can be made based on the diff from the returned prior version. The caller should *not*
+// hold their reference beyond the action handler.
+func trackPod(ms *store.ManifestState, action k8swatch.PodChangeAction) *v1alpha1.Pod {
```
maiamcc, line 160: tbh I've always had trouble reasoning about what this func returns, so if you have naming thoughts that might clarify, I'm v in favor. `trackPodAndReturnExisting` (or maybe better yet `trackPodAndReturnPrevious`) is... more readable, maybe annoyingly so? idk.

maiamcc (approved):

internal/engine/pod.go
milas: It was an attempt to make the diff more readable (the old var that _actually_ got used was named `podInfo`), but it's more confusing than helpful so I just stuck with `pod`

PR #4465, opened 2021-04-22
engine: track deployed object refs

milas:
This is in preparation for `PodDiscovery` API work.

The motivation here is that currently, the engine extracts some
details about the objects it's managing from the specs, i.e. what
it "expects" to deploy. These are gathered across all manifests to
determine which namespaces to watch in the cluster.

Indepedently, the deployer stores the UIDs, but that loses the
mapping back to namespace. Currently, it's not an issue because of
the aforementioned logic determining which namespaces to watch.

With the move to apiserver, the semantics will be slightly different;
each manifest will define the ancestor UID + namespaces it cares
about. The reconciler will handle ensuring only 1x watch per namespace
is created for efficiency, but the key is that the spec needs to
include the UID + namespace together.

By holding the deployed object refs, we now have this information, so
a targeted spec can be created.

(This will also enable resolving a long-standing bug/unexpected
behavior that the extra label selectors will be across _any_ namespace
that's being watched, even if the respective manifest doesn't deploy
anything there. It's a very minor issue, but will be nice to clean up
in the process for free.)

### DIFF
--- a/internal/store/build_result.go
+++ b/internal/store/build_result.go
@@ -4,9 +4,10 @@ import (
 	"fmt"
 	"sort"
 
+	v1 "k8s.io/api/core/v1"
+
 	"github.com/docker/distribution/reference"
 	dockertypes "github.com/docker/docker/api/types"
-	"k8s.io/apimachinery/pkg/types"
 
 	"github.com/tilt-dev/tilt/internal/container"
 	"github.com/tilt-dev/tilt/internal/dockercompose"
@@ -127,8 +128,8 @@ func NewDockerComposeDeployResult(id model.TargetID, containerID container.ID, s
 type K8sBuildResult struct {
 	id model.TargetID
 
-	// The UIDs that we deployed to a Kubernetes cluster.
-	DeployedUIDs []types.UID
+	// DeployedEntities are references to the objects that we deployed to a Kubernetes cluster.
+	DeployedEntities []v1.ObjectReference
 
 	// Hashes of the pod template specs that we deployed to a Kubernetes cluster.
 	PodTemplateSpecHashes []k8s.PodTemplateSpecHash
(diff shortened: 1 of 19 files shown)


milas:
Oops - I bundled in an extra change I didn't want here.

internal/store/build_result.go
```diff
@@ -139,8 +140,8 @@ type K8sBuildResult struct {
 func (r K8sBuildResult) TargetID() model.TargetID   { return r.id }
 func (r K8sBuildResult) BuildType() model.BuildType { return model.BuildTypeK8s }
 
-// For kubernetes deploy targets.
-func NewK8sDeployResult(id model.TargetID, uids []types.UID, hashes []k8s.PodTemplateSpecHash, appliedEntities []k8s.K8sEntity) BuildResult {
+// NewK8sDeployResult creates a deploy result for Kubernetes deploy targets.
```
maiamcc, line 143: ooh look at that properly formatted docstring!

internal/store/build_result.go
```diff
@@ -127,8 +128,8 @@ func NewDockerComposeDeployResult(id model.TargetID, containerID container.ID, s
 type K8sBuildResult struct {
 	id model.TargetID
 
-	// The UIDs that we deployed to a Kubernetes cluster.
-	DeployedUIDs []types.UID
+	// DeployedEntities are references to the objects that we deployed to a Kubernetes cluster.
+	DeployedEntities []v1.ObjectReference
```
maiamcc, line 132: checking my understanding: the main change in this PR (besides moving the UIDSet type to the k8s package) is storing ObjectReferences here instead of UIDs, since the former contains namespace info and we'll need that at some point in the near future?

maiamcc (approved):
lgtm!

internal/store/build_result.go
milas: Yes! Right now, the watch logic looks at the "things we parsed out of YAML that we think we'll deploy" to find namespaces to watch for entity changes and then uses "UIDs of things we _actually_ deployed" to decide what entities to actually react to changes thereof.

Since the apiserver model will have an explicit spec, it is kinda weird to have something like this:
```yaml
uids: ["abc123", "def456", "xyz789"]
namespaces: ["mix", "meow"]
```
(There's no relationship between indices of UID + namespace)
vs
```yaml
watch:
  - uid: "abc123"
     namespace: "meow"
  - uid: "def456"
     namespace: "mix"
  - uid: "xyz789"
     namespace: "meow"
```

internal/store/build_result.go
milas: IntelliJ aggressively highlights them so I miiiight opportunistically fix them up if I'm touching the method anyway 😅

PR #98, opened 2018-08-17
engine: don't push to a central registry when using docker-for-mac

nicks:
Hello @yuindustries, @jazzdan,

Please review the following commits I made in branch nicks/kube:

9e8ad19ed93c1ab1cb01932e2863d7840f708bcd (2018-08-17 18:18:15 -0400)
engine: don't push to a central registry when using docker-for-mac

PR #4466, opened 2021-04-22
Switch back to image's user after copying files from restart-helper

salim-runsafe:
This addresses https://github.com/tilt-dev/tilt-extensions/issues/98. There might be a more elegant set up steps than locally running "docker inspect", but this worked in my local ad hoc testing and `make test` passed.

### DIFF
--- a/integration/restart_process_different_user/tilt_modules/restart_process/Tiltfile
+++ b/integration/restart_process_different_user/tilt_modules/restart_process/Tiltfile
@@ -39,6 +39,11 @@ def docker_build_with_restart(ref, context, entrypoint, live_update,
     base_ref = '{}{}'.format(ref, base_suffix)
     docker_build(base_ref, context, **kwargs)
 
+    # get the user specified by the docker image
+    # this will often be an empty string, but when that is passed in below that will leave
+    # it as "root"
+    base_user = local("docker image inspect {} --format='{{{{.Config.User}}}}'".format(ref))
+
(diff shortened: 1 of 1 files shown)


salim-runsafe:
I'm realizing now that this should probably have been a PR over to tilt-dev/tilt-extensions. @maiamcc you opened #98 there, can you help me make sure I get this change in the right place(s)? Does it go over there and does it also need to exist as a change into the `integration` folder like in this PR?

maiamcc:
ooh, thanks for this! The PR should be against the extensions repo, but when you open it there, leave a note for us to pull the latest version of the extension into the integration tests etc.

You're welcome to keep using this PR to run integration tests against your change though, this is a great idea--and has uncovered a bug! Integration tests failed with:
> local: docker image inspect rpdu --format='{{.Config.User}}'
> → Error: No such image: rpdu

because a `docker_build` call in your Tiltfile doesn't immediately build the image, but rather registers the image-to-be-built with Tilt, and Tilt does all the building after the Tiltfile is done being read. your `docker image inspect` call failed b/c the image it's referring to hasn't yet been built.

hm, you might try an optional arg to `docker_build_with_restart` where you can pass the user to reset to? It's not great from a data duplication standpoint, but it'll get the job done (and if we as the Tilt team see a lot of people using that arg, we can look into a more first-class solution).

salim-runsafe:
Thanks @maiamcc, I'll make that change to pass in the user arg instead and will send it over to the correct repo.

PR #4470, opened 2021-04-27
podlogstream: fix an update loop

nicks:
Hello @milas,

Please review the following commits I made in branch nicks/update:

904588f5393c41a44b419dc1d86c260b80eae5b1 (2021-04-27 15:52:09 -0400)
podlogstream: fix an update loop

### DIFF
--- a/internal/controllers/core/podlogstream/podlogstreamcontroller.go
+++ b/internal/controllers/core/podlogstream/podlogstreamcontroller.go
@@ -9,7 +9,7 @@ import (
 
 	"github.com/tilt-dev/tilt/pkg/apis/core/v1alpha1"
 
-	"github.com/google/go-cmp/cmp"
+	"k8s.io/apimachinery/pkg/api/equality"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/types"
 	ctrl "sigs.k8s.io/controller-runtime"
@@ -42,6 +42,7 @@ type Controller struct {
 	watches         map[podLogKey]PodLogWatch
 	hasClosedStream map[podLogKey]bool
 	statuses        map[types.NamespacedName]*PodLogStreamStatus
+	lastUpdate      map[types.NamespacedName]*PodLogStreamStatus
 
 	newTicker func(d time.Duration) *time.Ticker
 	since     func(t time.Time) time.Duration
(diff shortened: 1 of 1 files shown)


nicks:
gahhhhh, i wonder if we should add some sort of integration test or metric for the number of updates, catching this class of error is a pain...

internal/controllers/core/podlogstream/podlogstreamcontroller.go
```diff
@@ -433,9 +435,16 @@ func (r *Controller) updateStatus(streamName types.NamespacedName) {
 		return
 	}
 
+	lastUpdate, hasLastUpdate := r.lastUpdate[streamName]
+	if hasLastUpdate && cmp.Equal(status, lastUpdate) {
+		return
+	}
+
+	r.lastUpdate[streamName] = status.DeepCopy()
```
milas, line 443: Shouldn't this only be updated if the status update actually succeeds below?

internal/controllers/core/podlogstream/podlogstreamcontroller.go
```diff
@@ -433,9 +435,16 @@ func (r *Controller) updateStatus(streamName types.NamespacedName) {
 		return
 	}
 
+	lastUpdate, hasLastUpdate := r.lastUpdate[streamName]
+	if hasLastUpdate && cmp.Equal(status, lastUpdate) {
```
milas, line 439: I realize this is actually the same as before, but any reason to use `cmp` vs the K8s `equality.Semantic.DeepEqual()`?

milas (commented):
I think the last update tracking needs to move to only be on success or we might miss some updates?

Also, would it make sense to, instead of storing copies of the last update result, compare to the entity you fetch?

```go
stream := &PodLogStream{}
err := r.client.Get(r.ctx, streamName, stream)
if err != nil {
	return
}
if equality.Semantic.DeepEqual(status, stream.Status) {
	return
}
// do update...
```

nicks:
Re: would it make sense to, instead of storing copies of the last update result, compare to the entity you fetch?

Nope! This is precisely the bug. The entity you fetch might be arbitrarily different than the entity you sent.

internal/controllers/core/podlogstream/podlogstreamcontroller.go
nicks: done

internal/controllers/core/podlogstream/podlogstreamcontroller.go
nicks: done

nicks:
fwiw, i do think we need to investigate other kubebuilder / controller patterns for making these updates more "natural". They're tricky to get right. "Don't read status when updating status" is a general good rule of thumb

milas (approved):

PR #4477, opened 2021-04-28
api: KubernetesDiscovery data model

milas:
To start, this will just handle Pod discovery, but it can be more
generally extended to other K8s resources Tilt cares about (e.g.
services for endpoints) in the future, which is why it's called
more generically `KubernetesDiscovery` rather than `PodDiscovery`
as originally intended.


pkg/apis/core/v1alpha1/kubernetesdiscovery_types.go
```diff
@@ -18,9 +18,138 @@ limitations under the License.
...
+}
+
+var _ resource.ObjectList = &KubernetesDiscoveryList{}
+
+func (in *KubernetesDiscoveryList) GetListMeta() *metav1.ListMeta {
+	return &in.ListMeta
+}
+
+// KubernetesDiscoveryStatus defines the observed state of KubernetesDiscovery
+type KubernetesDiscoveryStatus struct{}
```
milas, line 137: Status is temporarily empty to avoid any confusion - see #4478; for sanity + commit/PR size sake, I'm trying to aggressively chunk things and so actual status population/handling is next (first iteration is using the spec to drive discovery, second iteration will actually use status for reporting instead of ye old `PodChangeAction`).

But spoiler: it's just a list of `v1alpha.Pod` which are ⏬ in this file already

pkg/apis/core/v1alpha1/kubernetesdiscovery_types.go
```diff
@@ -18,9 +18,138 @@ limitations under the License.
...
+	"github.com/tilt-dev/tilt-apiserver/pkg/server/builder/resource/resourcestrategy"
 )
 
+// +genclient
+// +genclient:nonNamespaced
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// KubernetesDiscovery
+// +k8s:openapi-gen=true
+type KubernetesDiscovery struct {
```
nicks, line 38: you also need to add this to register.go i think

nicks (approved):

pkg/apis/core/v1alpha1/kubernetesdiscovery_types.go
milas: Ahh good catch - fyi had to update one of the CLI tests to be more lenient (output width/spacing is based on longest resource name, and our new friend `k17y` is annoyingly long) - b0ab895fdb99f8ebb9ad428af18f3fa0d9bc2d20

PR #4478, opened 2021-04-28
k8s: use KubernetesDiscovery spec to drive Pod discovery

milas:
This is another incremental change (but the largest) towards moving
Pod discovery fully into apiserver.

This adds `KubernetesDiscoveryXxxAction` action types for CRUD and
corresponding reducers that just update store (no apiserver
interaction yet).

Then, `k8swatch.ManifestSubscriber` is a new store subscriber that
takes on the job of creating `KubernetesDiscovery` specs for K8s
manifests. The existing `k8swatch.PodWatcher` subscriber now _only_
reacts to changes on `KubernetesDiscovery` objects and behaves like
a reconciler for them.

However, the status for `KubernetesDiscovery` is still empty and
no `KubernetesDiscoveryUpdateStatusAction` exists/is dispatched.
The existant `PodChangeAction` and its reducer are still used
downstream. This was just done to limit scope of this commit +
for code review sanity, and will be converted imminently.

Internally to `k8swatch.PodWatcher`, there were a few main chunks of
logic changes:
 * Diff logic acts on a per-entity basis rather than global and
   does not use the shared `watcherKnownState` helper anymore
 * Key everything off of `types.NamespacedName` and only map back
   to manifest for `PodChangeAction` dispatch as a bridge
 * Extra state tracking due to reconciling on a per-entity basis
   to ensure that namespace watches have appropriate lifecycle


internal/engine/k8swatch/manifest_subscriber.go
```diff
@@ -0,0 +1,178 @@
...
+			})
+		}
+	}
+
+	var labelSets [][]v1alpha1.LabelValue
+	if len(seenNamespaces) > 0 {
+		for i := range kt.ExtraPodSelectors {
+			l, err := labelsFromSelector(kt.ExtraPodSelectors[i])
+			if err != nil {
+				logger.Get(ctx).Debugf("Error processing extra pod selectors for %q: %v", kt.Name, err)
```
nicks, line 142: I don't think this is a reasonable way to handle this error. This is just going to infinite loop and log this on every change.

internal/engine/k8swatch/manifest_subscriber.go
```diff
@@ -0,0 +1,178 @@
...
+
+	if fieldErrs := kd.Validate(ctx); len(fieldErrs) != 0 {
+		var sb strings.Builder
+		for i, fieldErr := range fieldErrs {
+			if i != 0 {
+				sb.WriteString(", ")
+			}
+			sb.WriteString(fieldErr.Error())
+		}
+		logger.Get(ctx).Debugf("Manifest %q produced invalid spec: %s",
```
nicks, line 172: ditto here re: infinite loops

internal/engine/k8swatch/pod_watch.go
```diff
@@ -37,79 +65,196 @@ type PodWatcher struct {
...
+	return w.claimedUIDs[uid]
+}
+
+// reconcile manages namespace watches for the modified KubernetesDiscovery object.
+func (w *PodWatcher) reconcile(ctx context.Context, st store.RStore, key types.NamespacedName) {
+	state := st.RLockState()
+	kd := state.KubernetesDiscoveries[key]
+	st.RUnlockState()
+
+	if kd == nil {
```
nicks, line 151: i liked the other pattern you used in FileWatch where you checked "if deleted || specChanged { tearDown() }"... i had trouble reasoning about the piecemeal diffing below

internal/engine/k8swatch/pod_watch.go
```diff
@@ -37,79 +65,196 @@ type PodWatcher struct {
...
-		delete(w.watcherKnownState.namespaceWatches, teardown)
 	}
 
-	for _, setup := range taskList.setupNamespaces {
-		w.setupWatch(ctx, st, setup)
+	seenUIDs := make(map[types.UID]bool)
+	seenNamespaces := make(map[k8s.Namespace]bool)
+	for _, toWatch := range kd.Spec.Watches {
+		namespace := k8s.Namespace(toWatch.Namespace)
+		if !seenNamespaces[namespace] {
```
nicks, line 187: this check is probably redundant because it's handled internally

internal/engine/k8swatch/pod_watch.go
```diff
@@ -37,79 +65,196 @@ type PodWatcher struct {
...
+
+		id := types.UID(toWatch.UID)
+		if id == "" || seenUIDs[id] {
+			continue
+		}
+		seenUIDs[id] = true
+
+		if claimant, ok := w.claimedUIDs[id]; !ok || claimant != key {
+			// either this UID has not been claimed (so it's new to Tilt)
+			// or the claim has changed (so it's new to this KubernetesDiscovery object)
```
nicks, line 200: i think this is going to lead to infinite loops where claims bounce back and forth between objects? (this is what the old code was protecting against)

i wonder if the data model should simply allow multiple discovery objects to have the same UID, which would make this more consistent with how it works where you set up nsWatch and clean up abandoned ones

nicks (commented):
whew! this is exciting!

internal/engine/k8swatch/pod_watch.go
milas: RE: data model, absolutely - that was my intent, in subsequent changes I pushed the "claim" logic up to `k8swatch.ManifestSubscriber` so it handles not creating "duplicates" from its perspective, but the actual watch logic doesn't have to care.

I'd originally chunked it to limit scope of this, but yeahhh this is definitely wrong, so gonna pull those changes in now because it's so much easier to reason about that way

internal/engine/k8swatch/manifest_subscriber.go
milas: This now results in an `ErrorAction` as of 3f33963be7b563808a88358a01a2a71b31c7f0b5 - comments also updated to reflect that any errors returned from `kubernetesDiscoveryFromManifest` are expected to be fatal and some explanation re: why those errors are indicative of a Tilt bug and should be impossible unless we regress something.

internal/engine/k8swatch/manifest_subscriber.go
milas: This now results in an `ErrorAction` as of 3f33963be7b563808a88358a01a2a71b31c7f0b5 - comments also updated to reflect that any errors returned from `kubernetesDiscoveryFromManifest` are expected to be fatal and some explanation re: why those errors are indicative of a Tilt bug and should be impossible unless we regress something.

internal/engine/k8swatch/pod_watch.go
milas: Moved the logic around in d2f14cc and put in a test case around a bunch of behaviors here. There's some lingering weirdness (that's heavily documented) in `PodWatcher` for the extra label selector cases.

internal/engine/k8swatch/pod_watch.go
milas: Cleaned this up and made the UID stuff more consistent as well, so there's `setupNamespaceWatch` and `setupUIDWatch` which are both idempotent, so it just calls them both indiscriminately - ebf7462ef

internal/engine/k8swatch/pod_watch.go
milas: Check out the latest version of the diff and let me know if you think it's easier to follow.

There's a temporary snag that I can't just fully rip down the old version because of the logic to only send `PodChangeEvent` if it's the first time it's being seen by the watcher (i.e. manifest/`KubernetesDiscovery`).

That'll go away imminently though once it swaps to `KubernetesDiscoveryUpdateStatusAction` since it'll just always emit a status update when there's a spec change. 🎉

nicks (approved):
seems fine, though i think we should wait until after the release today to merge

Issue #4476, opened 2021-04-28
Tilt (post 19) truncates useful logs


adamf:
### Expected Behavior
A process' full logs should be available.

### Current Behavior
A process logs a lot of data - a few thousand lines, say - and while scrolling thru these logs to find the issue the logs suddenly truncate and jump to the last log lines, leaving only a page or two of logs left. This means we can't find important data - stack traces, failures in test suites, etc.

### Steps to Reproduce
1. Run a process that generates a lot of logs
2. Scroll around
3. The logs get cut off

#### `tilt doctor` Output
OS X latest, tile 0.19.1-6, hyperkit driver. 

#### About Your Use Case
We've had to roll back to 0.18.x. We can't use tilt if the logs don't work. This is particularly impactful in the test suite cases, where, say, 200 tests run and 4 fail early and log their traceback early in the run. Those tracebacks get lost with 0.19.x

That said, the logs do take up a bunch of browser memory over time. Perhaps the logs could truncate after x resource restarts, but always keep the full logs of a resource run?

adamf:
Also see https://kubernetes.slack.com/archives/CESBL84MV/p1617637043068000

nicks:
hmmm....thanks for reporting. For what it's worth, Tilt has always truncated logs. Otherwise it would grind to a halt. We can't realistically keep all logs. For an example, see: https://github.com/tilt-dev/tilt/issues/3936

I think the algorithm for truncation changed a bit in v0.19, so it would be helpful to have an example of a server configuration where tilt was discarding useful logs...

adamf:
@nicks for us it's the test case example - we have something like 200 unit tests for one of our applications and when you run the full test suite the log truncation in 19+ removes most of the logs a short time after the tests finish, which means you can't go back and see why some tests failed. This doesn't happen in 18 (it does truncate, but it leaves plenty for research).

nicks:
how long are the logs for the tests? i think the current cap is about 1MB. The test server I usually use logs about once every 10ms, and it goes awhile before Tilt truncates anything. I think the main thing that changed is what we decide to truncate first.

nicks:
(we could just bump it to 2MB, but i'm worried that something else is happening that's particular to your setup that we need to be able to repro to fix)

adamf:
[...]

PR #4480, opened 2021-04-28
logstore: raise the log limit, and truncate by both recency and length

nicks:
Hello @landism, @milas,

Please review the following commits I made in branch nicks/logstore:

cfb1e8b4ceab8491cc8fae42c9060d47766e5cd4 (2021-04-28 14:40:37 -0400)
logstore: raise the log limit, and truncate by both recency and length
fixes https://github.com/tilt-dev/tilt/issues/4476

### DIFF
--- a/pkg/model/logstore/logstore.go
+++ b/pkg/model/logstore/logstore.go
@@ -2,6 +2,7 @@ package logstore
 
 import (
 	"fmt"
+	"sort"
 	"strings"
 	"time"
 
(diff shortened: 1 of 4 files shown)


nicks:
I also wonder if we should get rid of the global log cap completely, and instead have a per-resource cap.

That would make things behave more like other log storage systems, but of course introduces the risk that you'll add tons of services and blow up perf that way.

milas (approved):

milas:
Per-resource would be nice but definitely seems like it'd open a can of worms on the UI side - based on the comments, the global cap was originally added to keep initial UI of `All` fast. At the very least I think we'd need to be able to load logs in arbitrary chunks (and be able to get a handle to latest) so we could lazily page stuff in/out from latest or something along those lines.

nicks:
ya, i think i basically agree that if this gets any more complex, we should scrap it and move to per-resource caps

PR #4481, opened 2021-04-28
test: fix a flakey test

nicks:
Hello @nicks,

Please review the following commits I made in branch nicks/timecmp:

5980b767a50271ab1dc091e65d25f0db74a73902 (2021-04-28 14:53:38 -0400)
test: fix a flakey test

### DIFF
--- a/internal/timecmp/cmp_test.go
+++ b/internal/timecmp/cmp_test.go
@@ -20,12 +20,16 @@ type tc struct {
 	nanosecond  bool
 }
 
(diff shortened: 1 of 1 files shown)


milas (approved):
I thought truncating would be sufficient to prevent weirdness like this but evidently not...I hate dealing with times 😭

nicks:
There were a couple cases where we were right on a boundary, so it got truncated in the wrong direction

PR #4483, opened 2021-04-28
portforwards: PortForwardSubscriber uses new api objects, stores current PortForwards on state

maiamcc:
this PR updates the `PortForwardSubscriber` to think about what PFs to start/stop in terms of `v1alpha1.PortForward` objects

the Subscriber also updates the list of current PortForwards on the engine state.

In a future, PR, we'll have the Subscriber create/delete PortForward objects in the API as well, and move all the logic for actually running a PortForward to the to-be-created `PortForwardReconciler`

### DIFF
--- a/internal/engine/portforward/subscriber.go
+++ b/internal/engine/portforward/subscriber.go
@@ -2,6 +2,7 @@ package portforward
 
 import (
 	"context"
+	"fmt"
 	"time"
 
 	"k8s.io/apimachinery/pkg/api/equality"
@@ -81,8 +82,9 @@ func (s *Subscriber) diff(ctx context.Context, st store.RStore) (toStart []portF
 					continue
 				}
 
-				// The port forward has changed, so remove the old version and re-add the new one
-				toShutdown = append(toShutdown, oldEntry)
+				// Every possible PortForward ought to have a unique name--it
+				// should be impossible to get two unequal PFs with the same name
+				panic(fmt.Sprintf("found two different PortForwards with same name %s (this should be impossible)", apiPf.Name))
 			}
 
 			toStart = append(toStart, entry)
--- a/internal/engine/portforward/subscriber.go
+++ b/internal/engine/portforward/subscriber.go
@@ -2,7 +2,6 @@ package portforward
 
 import (
 	"context"
-	"fmt"
 	"time"
 
 	"k8s.io/apimachinery/pkg/api/equality"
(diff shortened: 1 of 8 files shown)


pkg/apis/core/v1alpha1/portforward_types.go
```diff
@@ -76,6 +79,50 @@ type PortForwardSpec struct {
...
+			PodName:       podID,
+			Namespace:     ns,
+			ContainerPort: containerPort,
+			LocalPort:     localPort,
+			Host:          host,
+		},
+	}
+}
+
+// pfName generates a consistent and unique name for a port forward based on its component parts
```
nicks, line 105: hmmm...i'm not sure this restriction makes much sense, for a couple different reasons:

- we're moving towards a world where people can create their own port forwards. You _could_ try to enforce name and immutability restrictions, but it's not the "easy path", and I don't see a good reason to try to enforce them.

- the real restriction on port forwards is that you can't have two port-forwards binding to the same local port

have you seen how FileWatch handles this? I think a good approach here would be to make the port forward name = manifest name + index (e.g., storybook-1, storybook-2), and do a spec comparison like here: https://github.com/tilt-dev/tilt/blob/master/internal/controllers/core/filewatch/controller.go#L81

- -

pkg/apis/core/v1alpha1/portforward_types.go
```diff
@@ -54,19 +57,19 @@ type PortForwardList struct {
 // PortForwardSpec defines the desired state of PortForward
 type PortForwardSpec struct {
 	// The name of the pod to port forward to/from. Required.
-	Pod string `json:"pod"`
+	PodName string `json:"pod_name"`
```
nicks, line 60: it would be better to put schema changes on their own PR

internal/engine/portforward/subscriber.go
```diff
@@ -103,25 +111,28 @@ func (s *Subscriber) OnChange(ctx context.Context, st store.RStore, _ store.Chan
...
 		})
 
-		for _, forward := range entry.forwards {
-			entry := entry
-			forward := forward
-			go s.startPortForwardLoop(ctx, entry, forward)
-		}
+		go s.startPortForwardLoop(ctx, entry)
+
+		st.Dispatch(NewPortForwardCreateAction(entry.PortForward))
```
nicks, line 129: i'm pretty sure you're going to run into problems with this approach.

usually the data flow works like
subscriber reads EngineState -> creates API objects
reconciler reads API Objects -> sets up external work

The migration path we recommend tries to get the data flow working first

I think it's going to be really hard to start with the opposite data flow, and then try to reverse it

nicks: (this is what Milas and I both originally tried to do and both of us hit a wall :grimacing: )

maiamcc: > I think it's going to be really hard to start with the opposite data flow, and then try to reverse it

how do you figure this is the opposite data flow? This is me writing the API objects to the engine state. I think I'm missing something here...

pkg/apis/core/v1alpha1/portforward_types.go
maiamcc: just basing the index on how many other PFs for that manifest exist? That makes sense but seems like it'd be tricky because I need to know the state AND have PF's stored by manifest in order to make that go, hm.

i might call them "$POD-1" instead because that's at least information I have access to in the `PortForwadSubscriber`--I assume that still fits with the basic suggestion you're making here?

internal/engine/portforward/subscriber.go
nicks: @milas has done more of these than I have at this point and can probably give better guidance / pointers on next steps (or if he sees possible paths from here)

by "opposite", i mean this code creates the PortForward entry + loop first, then sends the PortForward API object to the EngineState. Usually, you'd go in the opposite direction - send the PortForward API object to the engine state, then read it off the engine state to set up the port forward in a separate reconciler.

internal/engine/portforward/subscriber.go
```diff
@@ -103,25 +111,28 @@ func (s *Subscriber) OnChange(ctx context.Context, st store.RStore, _ store.Chan
...
 	for _, entry := range toShutdown {
 		entry.cancel()
+
+		st.Dispatch(NewPortForwardDeleteAction(entry.Name))
+		// TODO(maia): delete PF object in API
 	}
 
 	for _, entry := range toStart {
 		// Treat port-forwarding errors as part of the pod log
 		ctx := logger.CtxWithLogHandler(entry.ctx, PodLogActionWriter{
```
nicks, line 121: fyi, this should eventually use  store.MustObjectLogHandler like PodLogStream does

pkg/apis/core/v1alpha1/portforward_types.go
maiamcc: (fwiw this naming constraint wouldn't need to exist for user-created PF's / PFs that went straight into the reconciler -- it's specifically for PFs created in the subscriber, so that if we make one for e.g. podA:8080->8080 and the next time around make one for podA:9090-9090, there should be an easy way to check whether the port forward exists or not. this wasn't a problem previously because there was only one entry per pod (which could contain multiple forwards) so it was easy to check if a given pod already had something on it. maybe i'll revert to doing it that way)

internal/engine/portforward/subscriber.go
maiamcc: oh yes that is my plan for the next PR -- i did this instead of a bigger PR where I both write the new object to engine state AND move over all the logic for running the port forwards to a different controller.

are you saying you'd rather see the latter type of PR, or that there's a different and less foot-gun-y way to do this incrementally?

maiamcc: talked to milas offline, I'm going to try again with an incremental approach, but one that doesn't risk infinite loops by dispatching events.

maiamcc:
closing in favor of https://github.com/tilt-dev/tilt/pull/4499, which is
a. smaller since I landed some intermediate PRs,
b. deviates less from existing behavior, and
c. takes in to account some comments from this PR.

PR #4486, opened 2021-04-29
test: move PodStatus test to k8sconv

milas:
Missed the test when moving the logic, so just moving it to where
the actual code is.

### DIFF
--- a/internal/engine/k8swatch/pod_watch_test.go
+++ b/internal/engine/k8swatch/pod_watch_test.go
@@ -2,7 +2,6 @@ package k8swatch
 
 import (
 	"context"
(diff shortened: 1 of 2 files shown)


nicks (approved):
fwiw, should probably add the 'tests:' prefix to goreleaser so this doesn't show up in the changelog
https://github.com/tilt-dev/tilt/blob/master/.goreleaser.yml#L51

milas:
Oooo I didn't know that was a feature and I definitely hadn't realized I was being different by pluralizing 😅 

Renamed this PR to filter it out in the squash message and opened #4490 since it was super easy/quick

PR #4487, opened 2021-04-29
apis: generate protobufs

nicks:
Hello @milas,

Please review the following commits I made in branch nicks/go-to-protobuf:

6ed194ae730ea2d3479cde74d39ef0fc896df311 (2021-04-29 16:17:30 -0400)
apis: generate protobufs


nicks:
weeeeeeeeeeeeeee

I'm not convinced we want to go with protobufs long-term. But for the purposes of jamming this into the web interface as fast as possible (which already uses protobufs), this seems like the easiest approach

nicks:
oof, this actually catches some bugs in how we were currently defining the types (K8s API types don't allow 2-d arrays, and don't allow Go ints)

milas:
> K8s API types don't allow 2-d arrays

I think `KubernetesDiscovery` is the only one with this at the moment from looking at the diff. I'll poke around K8s API, but do you have a sense for general pattern on avoiding? Wrap inner with an object with a single `values` field or something?

nicks:
rebased this on top of https://github.com/tilt-dev/tilt/pull/4491, which addresses some of the api rules

milas (approved):
Might consider updating the boilerplate so that it has the protobuf tags on meta/spec/status but not a big deal

PR #4488, opened 2021-04-29
k8s/portforward: FakePortForwardClient knows its whole call history

maiamcc:
splitting out some of the changes in #4483 to reduce headaches

this change makes the FakePortForwardClient aware of its whole call history rather than just the last call it made, which lets us test that we're correctly handling e.g. a single pod with multiple port forwards

### DIFF
--- a/internal/k8s/fake_client.go
+++ b/internal/k8s/fake_client.go
@@ -491,26 +491,60 @@ func (pf FakePortForwarder) ForwardPorts() error {
 }
 
 type FakePortForwardClient struct {
-	CreatePortForwardCallCount int
-	LastForwardPortPodID       PodID
-	LastForwardPortRemotePort  int
-	LastForwardPortHost        string
-	LastForwarder              FakePortForwarder
-	LastForwardContext         context.Context
+	mu               sync.Mutex
+	PortForwardCalls []PortForwardCall
+}
+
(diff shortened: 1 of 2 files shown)


internal/k8s/fake_client.go
```diff
@@ -491,26 +491,60 @@ func (pf FakePortForwarder) ForwardPorts() error {
 }
 
 type FakePortForwardClient struct {
-	CreatePortForwardCallCount int
-	LastForwardPortPodID       PodID
-	LastForwardPortRemotePort  int
-	LastForwardPortHost        string
-	LastForwarder              FakePortForwarder
-	LastForwardContext         context.Context
+	mu               sync.Mutex
```
milas, line 494: I'd probably make `PortForwardCalls` unexported + provide a helper method that takes the lock and returns a copy of them (thankfully `FakePortForwarder` looks safe to copy, so doesn't need its own locks as well 😅). The `CreatePortForwardCallCount` + `LastXxx` helpers also don't currently use the mutex before reads.

milas (approved):
🎉 for more tests

internal/k8s/fake_client.go
maiamcc: good call, done!

PR #4489, opened 2021-04-29
v1alpha1: one port forward object may contain multiple forwards

maiamcc:
as per https://github.com/tilt-dev/tilt/pull/4483#discussion_r623331797, when we have one pf API object per pair of ports to forward between, it can get awkward to diff existing vs. desired in the subscriber.

this PR has the api object behaving more in line with the current port forward subscriber behavior -- one PF API object contains one or more forwards to run on the specified pod.


maiamcc:
(note that this new data organization doesn't ENFORCE "one PortForward API object per pod", but it does allow the PortForwardSubscriber to create one per pod, which will make it easier to track/diff PFs against each other)

milas (approved):

pkg/apis/core/v1alpha1/portforward_types.go
```diff
@@ -54,19 +54,25 @@ type PortForwardList struct {
 // PortForwardSpec defines the desired state of PortForward
 type PortForwardSpec struct {
 	// The name of the pod to port forward to/from. Required.
-	Pod string `json:"pod"`
+	PodName string `json:"pod_name"`
```
milas, line 57:
```suggestion
	PodName string `json:"podName"`
```

FYI apiserver names are `camelCase` (standards doc still forthcoming)

pkg/apis/core/v1alpha1/portforward_types.go
```diff
@@ -54,19 +54,25 @@ type PortForwardList struct {
...
+	Forwards []Forward `json:"forwards"`
+}
 
+// Forward defines a port forward to execute on a given pod.
+type Forward struct {
 	// The port to expose on the current machine. Required.
 	LocalPort int `json:"local_port"`
 
+	// The port on the Kubernetes pod to connect to. Required.
+	ContainerPort int `json:"container_port"`
```
milas, line 74:
```suggestion
type Forward struct {
	// The port to expose on the current machine. Required.
	LocalPort int `json:"localPort"`

	// The port on the Kubernetes pod to connect to. Required.
	ContainerPort int `json:"containerPort"`
```

🐫

pkg/apis/core/v1alpha1/portforward_types.go
maiamcc: 🤯  cool cool ty

PR #4495, opened 2021-04-30
k8s: defer extra pod labels conversion to labels.Selector

milas:
Extra pod labels come in as one or more maps of label keys and values
(`[]labels.Set`). Once converted into a list of `labels.Selector`,
it can be used for matching, but introspecting it is more difficult,
as it requires conversion, which produces a list of
`labels.Requirement`. The biggest problem that introduces is that
`labels.Requirement` is more expressive, so backwards conversion
means a bunch of error-handling for (what should be) impossible
cases.

This just passes the `labels.Set`s as they're parsed from the
`Tiltfile` through and stores them that way on the target in state.
Once they're actually used, they get converted to the actual
`labels.Selector`s for matching.

This will simplify the new intermediary step where they'll get
converted to API label selector objects, so that conversion can
happen without hiccups or excessive boilerplate.

### DIFF
--- a/internal/engine/k8swatch/pod_watch.go
+++ b/internal/engine/k8swatch/pod_watch.go
@@ -70,7 +70,8 @@ func (w *PodWatcher) diff(ctx context.Context, st store.RStore) podWatchTaskList
 	var extraSelectors []ExtraSelector
 	if len(taskList.watchableNamespaces) > 0 {
 		for _, mt := range state.Targets() {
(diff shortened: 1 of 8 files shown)


milas:
This is just to make the conversion to `KubernetesDiscovery` easier so that instead of going from `labels.Set` -> `labels.Selector` -> `metav1.LabelSelector` -> `labels.Selector`, it'll be `labels.Set` -> `metav1.LabelSelector` -> `labels.Selector`.

nicks (approved):

PR #4505, opened 2021-05-03
test: blackhole useless logs and fix a race condition

nicks:
Hello @milas,

Please review the following commits I made in branch nicks/tests:

0c46c6a668b5ae12396eaad412ad0ca24e9b2c74 (2021-05-03 18:15:27 -0400)
test: blackhole useless logs and fix a race condition

### DIFF
--- a/internal/engine/upper_test.go
+++ b/internal/engine/upper_test.go
@@ -1679,12 +1679,10 @@ func TestPodUnexpectedContainerAfterSuccessfulUpdate(t *testing.T) {
 		StartTime:    f.Now(),
 		SpanID:       spanID0,
 	})
-	podStartTime := f.Now()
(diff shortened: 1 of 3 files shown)


internal/engine/upper_test.go
```diff
@@ -1679,12 +1679,10 @@ func TestPodUnexpectedContainerAfterSuccessfulUpdate(t *testing.T) {
 		StartTime:    f.Now(),
 		SpanID:       spanID0,
 	})
-	podStartTime := f.Now()
```
nicks, line 1682: this isn't a new problem, but is aggravated by the recent timecmp changes. we want to use time.Now() when the pod builder Build(), not define it upfront

milas (approved):

PR #4512, opened 2021-05-04
apis: send UIResource instead of Resource

nicks:
Hello @milas, @landism,

Please review the following commits I made in branch nicks/uisession3:

42e651196944fee3b9d0e07fd5506d113a5e6fac (2021-05-04 16:14:20 -0400)
apis: send UIResource instead of Resource

9dbf8d09ad0dc86a1eea9614f9e876ee26613c97 (2021-05-04 14:21:11 -0400)
apis: define fields for UIResourceStatus
This is a pretty straight-up port of what's currently
in the UI protobuf, but with some slight tweaks to better match
conventions for Kubernetes APIs.

See also:
https://github.com/tilt-dev/tilt/blob/8d9c1033b3bc308b51af865682dd1c6be6a81040/pkg/webview/view.proto

### DIFF
--- a/internal/hud/webview/convert_test.go
+++ b/internal/hud/webview/convert_test.go
@@ -8,7 +8,6 @@ import (
 
 	"github.com/tilt-dev/tilt/pkg/apis/core/v1alpha1"
 
-	"github.com/golang/protobuf/ptypes/timestamp"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
@@ -17,6 +16,7 @@ import (
 	"github.com/tilt-dev/tilt/internal/k8s/testyaml"
 	"github.com/tilt-dev/tilt/internal/store"
 	"github.com/tilt-dev/tilt/internal/testutils/tempdir"
+	"github.com/tilt-dev/tilt/internal/timecmp"
 	"github.com/tilt-dev/tilt/pkg/logger"
 	"github.com/tilt-dev/tilt/pkg/model"
 	proto_webview "github.com/tilt-dev/tilt/pkg/webview"
@@ -54,7 +54,7 @@ func TestStateToWebViewRelativeEditPaths(t *testing.T) {
 		}
 	v := stateToProtoView(t, *state)
 
-	require.Len(t, v.Resources, 2)
+	require.Len(t, v.UiResources, 2)
 }
 
 func TestStateToWebViewPortForwards(t *testing.T) {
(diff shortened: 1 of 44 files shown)


nicks:
this change has a lot of lines. But it's fairly straightforward and mechanical. We're converting from our old format (where the source of truth was protobufs) to the new format (where the source of truth is Go structs), and tweaking some of the representations a bit.

The major change is how Kubernetes represents time.

web/src/SidebarItemView.stories.tsx
```diff
@@ -45,7 +45,9 @@ function withBuildStatusOnly(status: ResourceStatus): optionFn {
     item.buildStatus = status
     item.runtimeStatus = ResourceStatus.None
     if (status === ResourceStatus.Building) {
-      item.currentBuildStartTime = new Date(Date.now() - 1).toISOString()
+      item.currentBuildStartTime = {
+        seconds: new Date(Date.now() - 1).toISOString(),
```
milas, line 49: I'm surprised TS compiler is allowing this! (I think it'll actually work though since we just pass the `seconds` value through to the component which can presumably handle both formats.

FWIW `fromDate()` might be helpful...assuming it exists on `MicroTime`, if not: `Math.floor(value.getTime() / 1000)`. Source: https://github.com/protocolbuffers/protobuf/blob/436bd7880e458532901c58f4d9d1ea23fa7edd52/src/google/protobuf/compiler/js/well_known_types_embed.cc#L92-L104

milas (approved):

web/src/SidebarItemView.stories.tsx
nicks: i changed my mind on this -- I'm going to try to hack MicroTime to serialize differently. hold on...

internal/hud/webview/convert_test.go
```diff
@@ -299,15 +301,17 @@ func TestBuildHistory(t *testing.T) {
...
-	for i, actual := range r.BuildHistory {
+	rs := r.Status
+	require.Len(t, rs.BuildHistory, 3)
+
+	for i, actual := range rs.BuildHistory {
 		expected := buildRecords[i]
-		require.Equal(t, mustTimeToProto(expected.StartTime), actual.StartTime)
-		require.Equal(t, mustTimeToProto(expected.FinishTime), actual.FinishTime)
+		require.Equal(t, metav1.NewMicroTime(expected.StartTime), actual.StartTime)
+		require.Equal(t, metav1.NewMicroTime(expected.FinishTime), actual.FinishTime)
```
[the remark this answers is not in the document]
nicks: ooh good point! added

web/src/OverviewItemView.tsx
```diff
@@ -80,46 +85,47 @@ export class OverviewItem {
   runtimeAlertCount: number
   hasEndpoints: boolean
   lastBuildDur: moment.Duration | null
-  lastDeployTime: string
-  pendingBuildSince: string
-  currentBuildStartTime: string
+  lastDeployTime: MicroTime
+  pendingBuildSince: MicroTime
+  currentBuildStartTime: MicroTime
```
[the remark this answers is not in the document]
nicks: i ended up having to revert this back to RFC3339-time :grimacing: https://github.com/tilt-dev/tilt/pull/4518

web/src/SidebarItemView.stories.tsx
nicks: ditto

PR #4513, opened 2021-05-04
test: emit Pod events instead of manually dispatching actions

milas:
The upper/buildcontroller tests (which share a fixture) were
simulating pod events by manually dispatching events, thus bypassing
the `PodWatcher` entirely.

That in and of itself isn't hugely problematic (although since these
are more integration than unit tests, it was skipping a lot of logic
that's useful to exercise).

The problem arises from the fact that events are dispatched per-pod
(`PodChangeAction`) currently, but as part of the changes for api
server, the `KubernetesDiscovery` status includes _all_ pods. Since
thigns are dispatched async, this presents new synchronization to
manage in the test.

Instead of trying to work around it further, this changes the tests
to rely on emitting Pod events from the fake K8s client and letting
`PodWatcher` see them and dispatch events.

This does present a complication because the pods generated by the
`PodWatcher` don't have any real relationship to the YAML in the
manifest. A new map on the fake build and deployer uses the provided
entities instead of trying to derive things from the YAML if
populated to accommodate this.

### DIFF
--- a/internal/engine/upper_test.go
+++ b/internal/engine/upper_test.go
@@ -17,6 +17,9 @@ import (
 	"testing"
 	"time"
 
+	"github.com/davecgh/go-spew/spew"
+	"k8s.io/apimachinery/pkg/labels"
+
 	"k8s.io/apimachinery/pkg/api/equality"
 
 	"github.com/tilt-dev/tilt/internal/timecmp"
@@ -213,6 +216,7 @@ type fakeBuildAndDeployer struct {
 	nextDockerComposeContainerID    container.ID
 	nextDockerComposeContainerState *dockertypes.ContainerState
 
+	targetObjectTree          map[model.TargetID]podbuilder.PodObjectTree
 	nextDeployedUID           types.UID
 	nextPodTemplateSpecHashes []k8s.PodTemplateSpecHash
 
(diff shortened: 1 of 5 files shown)


milas:
These changes were...frustrating to say the least. It still feels pretty icky but I was really struggling to get anything cleaner/better working - I've been having a ton of trouble getting the tests to behave well with my WIP branch for replacing `PodChangeAction` with `KubernetesDiscoveryUpdateStatusAction`, and in the end removing manual event dispatching from these tests seemed the best, since in theory they should now continue to pass untouched despite the internal refactoring for Pod watching (e.g. the aforementioned action change + imminent move to apiserver instead of store)

internal/engine/upper_test.go
```diff
@@ -2238,11 +2315,11 @@ func TestUpperBuildImmediatelyAfterCrashRebuild(t *testing.T) {
...
 
-	// Restart the pod with a new container id, to simulate a container restart.
-	f.podEvent(podbuilder.New(t, manifest).
-		WithPodID("pod-id").
-		WithContainerID("funnyContainerID").
-		Build(), manifest.Name)
+	// Simulate a container restart where the new container isn't running yet
+	// HACK(milas): by making the container ID an empty string, it'll get ignored; we should really just
+	// 	properly set the container status
+	pb = pb.WithContainerID("")
```
milas, line 2321: AFAICT the old logic here was actually subtly wrong - we want the same Pod, but a restarted container.

(The test actually started failing because our logic kicked in and pruned the other pod 😬)

nicks (approved):

PR #4516, opened 2021-05-05
k8s: use KubernetesDiscoveryUpdateStatusAction

milas:
This replaces `PodChangeAction`. Most tests remain unchanged as they
were already refactored to not dispatch `PodChangeAction` directly but
instead emit Pod events using `FakeK8sClient` and allowing the
`PodWatcher` to see them, dispatch events, and have the engine state
get updated. As a result, the inner changes of event type and
corresponding reducer are mostly invisible - wahoo!

Also eliminated is `PodDeleteAction` - since the `KubernetesDiscovery`
status object includes the full state (i.e. all Pods) each time, a
delete is simply a trimming of an item from the list and does not have
a dedicated action.

From a logic perspective, various things were up-leveled out of the
reducer. The biggest of these is container restart detection, which
still works in the same way, but now happens as part of reconciliation
when actually updating the object. Baseline container restarts
similarly is done as part of the reconciliation. Lastly, because the
status includes all Pods, each Pod includes its respective matched
ancestor UID since they might not all share one. (Within the actual
engine runtime state, there is still a "canonical" ancestor UID that
works the same as before, but the backing data has slightly richer
fidelity.)

Within the `PodWatcher`, there is a slight behavior change: previously,
extra pod label selectors were ONLY applied if the Pod did not have a
transitive UID match from any other watchers, and even then, only the
FIRST watcher to match on the label selectors would see it. Maintaining
this behavior would necessitate a lot of extra bookkeeping that isn't
warranted - the situations it will occur in are rare enough AND recent
changes mean that it won't actually break the engine, just result in
some duplicative data across manifests (i.e. logs for Pod seen both
places).

For similar reasons, the logic for handling update events on Pod delete
is not particularly efficient: it simply tries to dispatch updates to
EVERY watcher. This is done because there's no quick way to know which
ones might have referenced it. The update logic will short-circuit if
there are no changes, and this is actually very similar to how the
`handlePodDeleteAction` reducer worked previously: it'd look through
all manifests to see any that referenced it and remove it.


internal/engine/k8swatch/container_restart.go
```diff
@@ -0,0 +1,87 @@
...
+	"github.com/tilt-dev/tilt/pkg/model"
+)
+
+type Dispatcher interface {
+	Dispatch(action store.Action)
+}
+
+type ContainerRestartDetector struct{}
+
+func NewContainerRestartDetector() *ContainerRestartDetector {
```
nicks, line 23: i like this!

nicks (approved):
let's goooooo

milas:
FYI 20a35bd should hopefully fix up the tests - I forgot to remove `Pods` from `ChangeSummary`, which was never populated anymore (since `PodChangeAction` + `PodDeleteAction` no longer exist). Impact was that the store subscriber which creates `PodLogStream` objects would never run -- it now uses `summary.KubernetesDiscoveries.Changes` instead, which mimics what it will do once the KD reconciler exists + creates those objects 🙌 

(Actual subscriber logic didn't change at all -- it still uses the manifests vs. reading anything off KD objects. There were some small tweaks to the tests to ensure the event is populated properly.)

PR #4523, opened 2021-05-06
controllers/portforward: portforwards are run by the reconciler, reading off the engine state [ch11908]

maiamcc:
I've moved the "actually run the port forwards" logic into the PortForwardReconciler (reading desired PortForwards off the EngineState). This shouldn't be touched in prod yet -- will make a separate PR to remove the "actually run the port forwards" logic from the subscriber.


internal/controllers/core/portforward/reconciler.go
```diff
@@ -0,0 +1,199 @@
...
+			toShutdown = r.addToShutdown(toShutdown, existing)
+			continue
+		}
+	}
+
+	for name, desired := range statePFs {
+		existing, isActive := r.activeForwards[name]
+		if isActive {
+			// We're already running this PortForward -- do we need to do anything further?
+			if equality.Semantic.DeepEqual(existing.Spec, desired.Spec) && equality.Semantic.DeepEqual(existing.ObjectMeta, desired.ObjectMeta) {
```
nicks, line 65: comparing the ObjectMeta like this isn't right...it's going to compare things you don't expect, like ResourceVersion

internal/controllers/core/portforward/reconciler.go
```diff
@@ -0,0 +1,199 @@
...
+	delete(r.activeForwards, entry.Name)
+	return append(toShutdown, entry)
+}
+
+func (r *Reconciler) OnChange(ctx context.Context, st store.RStore, summary store.ChangeSummary) {
+	if !r.isActive {
+		return
+	}
+
+	if summary.IsLogOnly() {
```
nicks, line 96: I think your summary already has the list of PortForward names that have changed? can narrow this to len(summary.PortForwards) == 0

internal/controllers/core/portforward/reconciler.go
```diff
@@ -0,0 +1,199 @@
...
+	}
+
+	toStart, toShutdown := r.diff(ctx, st)
+	for _, entry := range toShutdown {
+		entry.cancel()
+	}
+
+	for _, entry := range toStart {
+		// Treat port-forwarding errors as part of the pod log
+		ctx := logger.CtxWithLogHandler(entry.ctx, PodLogActionWriter{
```
nicks, line 107: should probably use ObjectLogHandler instead of a custom one

nicks (approved):
nice!

internal/controllers/core/portforward/reconciler.go
maiamcc: Hmm fair. so the one thing i need to make sure stays the same is the ManifestName annotation -- would it be too weird to compare that directly? or should i duplicate that data on the state? or?

internal/controllers/core/portforward/reconciler.go
maiamcc: yep will do that after I'm changing state via events (right now i'm just setting the PortForwards directly on the state for tests)

internal/controllers/core/portforward/reconciler.go
maiamcc: lol I just lifted this straight from the original code. will do.

PR #4526, opened 2021-05-06
web: update the ui to read uiSession instead of view

nicks:
Hello @milas, @landism,

Please review the following commits I made in branch nicks/uisession7:

4d9b9844265cc51ec2d22b92a08d5812bee12684 (2021-05-06 16:41:50 -0400)
web: update the ui to read uiSession instead of view

0a7e237529d504b74e4ada9b3d74f0d73fdb9f62 (2021-05-06 16:41:22 -0400)
apis: add UISession (a straight port of the old view.proto)

### DIFF
--- a/web/src/HUD.tsx
+++ b/web/src/HUD.tsx
@@ -67,21 +67,7 @@ export default class HUD extends Component<HudProps, HudState> {
     })
 
     this.state = {
-      view: {
-        needsAnalyticsNudge: false,
-        fatalError: undefined,
-        runningTiltBuild: {
-          version: "",
-          date: "",
-          dev: false,
-        },
-        suggestedTiltVersion: "",
-        versionSettings: { checkUpdates: true },
-        featureFlags: {},
-        tiltCloudUsername: "",
-        tiltCloudSchemeHost: "",
-        tiltCloudTeamID: "",
-      },
+      view: {},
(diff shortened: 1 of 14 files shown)


web/src/HUD.tsx
```diff
@@ -248,16 +239,18 @@ export default class HUD extends Component<HudProps, HudState> {
...
     let resources = view?.uiResources ?? []
-    if (!resources?.length || !view?.tiltfileKey) {
+    if (!resources?.length || !session?.tiltfileKey) {
       return <HeroScreen message={"Loading…"} />
     }
 
-    let runningBuild = view?.runningTiltBuild
-    let suggestedVersion = view?.suggestedTiltVersion
-    const versionSettings = view?.versionSettings
+    let tiltfileKey = session?.tiltfileKey || ""
```
landism, line 250:
```suggestion
    let tiltfileKey = session.tiltfileKey
```

Maybe nitty, but I think the `!view?.tiltfileKey` condition in the `if` a few lines up is so that the compiler knows that `view.tiltfileKey` was a `string` in `tiltfileKeyContext.Provider` below, so the `|| ""` seems more confusing than anything. Or maybe something changed so that no longer worked?

landism (approved):

PR #4528, opened 2021-05-06
k8swatch: write KubernetesDiscovery objects to apiserver

milas:
Specs + status are now written to the apiserver. The `PodWatcher`
is not yet a reconciler.


milas:
Good chance that tests fail - I'm still finding + fixing some race-y tests that are more obvious due to being slowed down by writing to apiserver (there's a couple already fixed in this, but I've also seen `TestMultipleChangesOnlyDeployOneManifest` fail locally which is really confusing me at the moment).

Hoping to stress `internal/engine` tests a bit tomorrow and clean those up, but the logic here shouldn't change and is ready for review. (The primary issue in the tests I've encountered so far is they dispatch Pod events that they want in the next build but don't wait for them to get ack'd.)

milas:
I let the `internal/engine` tests run 100 times and didn't hit anymore failures* - I feel good about state of races in those.

I am a bit suspicious by the rate of integration test failures, but mostly they've been connection reset errors on the port forward, so I think are probably just local KIND wonkiness - the actual Pod deployment status from Tilt's perspective looks good when it dumps out the engine state at the end, so the `KubernetesDiscovery` looks like it's working as expected.

\* `TestTiltignoreRespectedOnError` failed once but it doesn't even use Pods so was just an unavoidable timeout thing (likely since my laptop was a bit overwhelmed)

integration/shortlived_pods_test.go
```diff
@@ -21,7 +21,6 @@ func TestShortlivedPods(t *testing.T) {
 	defer cancel()
 	f.WaitForAllPodsInPhase(ctx, "app=shortlived-pod-fail", v1.PodFailed)
 	f.WaitForAllPodsInPhase(ctx, "app=shortlived-pod-success", v1.PodSucceeded)
-	f.KillProcs()
```
nicks, line 24: ?

nicks (approved):

integration/shortlived_pods_test.go
milas: Test teardown kills them anyway, but first tries to dump engine state. This was killing them early, so when the test failed, you wouldn't get any useful output. There's no reason afaict to kill it before asserting - I think it's just an old test that might pre-date the `TearDown()`.

(This test was legitimately failing at one point because I broke podlogstream and it was annoying to debug 😓 )

PR #4534, opened 2021-05-07
test: namespace options in PodBuilder

milas:
Previously, all resources produced by PodBuilder had `default` as
the K8s namespace.

Now, it will examine the manifest YAML and try to find an explicitly
referenced namespace. If it finds more than one, it'll error out
(this is a safeguard - nothing trips it today; if ever hit, we can
add an _explicit_ namespace override to use since there is no sensible
default in those scenarios).

If it doesn't find a single explicit namespace, it'll fallback to a
"context" namespace value if set. This allows simulating the case
where you've got a non-default namespace set in `kubeconfig`. If
that is also not set, it'll fallback to `k8s.DefaultNamespace` (aka
`default`).

The "cluster namespace" has to always be passed to PodBuilder if
you want to use it because there's no practical way for PodBuilder
to infer that from the `Manifest`. This is fine because there's
really only a few isolated places where we care about this, so
those tests can use it, while all other tests get generally more
consistent data for free.

### DIFF
--- a/internal/testutils/podbuilder/podbuilder.go
+++ b/internal/testutils/podbuilder/podbuilder.go
@@ -67,6 +67,10 @@ type PodBuilder struct {
 	resourceVersion string
 	unknownOwner    bool
 
+	// contextNamespace allows setting a fallback default namespace instead of `default` to simulate
+	// a namespace override in the active config context.
+	contextNamespace k8s.Namespace
+
 	// keyed by container index -- i.e. the first container will have image: imageRefs[0] and ID: cIDs[0], etc.
 	// If there's no entry at index i, we'll use a dummy value.
 	imageRefs map[int]string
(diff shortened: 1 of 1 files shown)


internal/testutils/podbuilder/podbuilder.go
```diff
@@ -371,6 +387,28 @@ func (b PodBuilder) validateContainerIDs(numContainers int) {
...
+		}
+	}
+	if nsVal != "" {
+		return k8s.Namespace(nsVal)
+	}
+	if b.contextNamespace != "" {
+		return b.contextNamespace
+	}
+	// this is actually redundant as long as k8s.Namespace::String() is used which
+	// coerces empty string to this, but that behavior is actually a bit sketchy,
```
nicks, line 407: :nods:

nicks: would really like to get rid of k8s.Namespace, in retrospect it was a mistake and doesn't play well with other k8s code we have to interop with

nicks (approved):

internal/testutils/podbuilder/podbuilder.go
milas: I'd seen your (very helpful) comment on it previously which is what led me to adding this 😅

PR #4537, opened 2021-05-07
k8swatch: convert PodWatcher to reconciler

milas:
`PodWatcher` is now `kubernetesdiscovery.Reconciler`. Almost all
the logic is the same, though some stuff moved around. Most notable
change is that the check to decide whether to skip a status update
now happens against the API object instead of the store object.


milas:
There's a few more test cases from `pod_watcher_test.go` that vanished here that I want to re-introduce. Some of them (e.g. making sure we don't do bad `ResourceVersion` string comparisons) should be reconciler tests; some of them are more about manifest behavior and need to get covered by the `k8swatch.ManifestSubscriber` tests (I _think_ they all are, but made a list and need to compare to ensure match-y match-y behavior.)

FYI There are already (many) "integration" tests in `upper_test.go` which are currently passing (🎉) using the `k8swatch.ManifestSubscriber` store subscriber + `kubernetesdiscovery.Reconciler` apiserver controller.

nicks (approved):

milas:
This also fixes #4557. When there is no controller registered for a type, the `controller-runtime` client caching does not behave as expected, and will consistently return `NotFound` for some entities. When this happens, even adjusting the resync period to be very short doesn't cause it to become consistent.

We need to dive into the client caching behavior more, but this at least ensures things don't behave erratically.

PR #4563, opened 2021-05-18
test: dump apiserver state on integration test failure

milas:
Now that more and more state is in the apiserver, it's really
helpful to see its state when an integration test fails, particularly
for debugging issues where the engine state and apiserver are out
of sync.


nicks (approved):

nicks:
tho in fairness, this should probably be a built-in tilt command, similar to `kubectl cluster-info dump`

milas:
Yeah, it might even be handy to have a `tilt diag` that spits out doctor+engine+api+etc and zips it up

PR #4565, opened 2021-05-18
engine: fix a race condition in Tiltfile reload

nicks:
Hello @milas,

Please review the following commits I made in branch nicks/miss:

7b2ea03e1eb6cd9cd72c1d3d5bbb80706341347f (2021-05-18 17:45:30 -0400)
engine: fix a race condition in Tiltfile reload
PendingConfigFileChanges had its own change tracking,
which didn't get the fixes we applied in
https://github.com/tilt-dev/tilt/pull/4397

This removes that state completely and unifies
Tiltfile change tracking with other stuff

### DIFF
--- a/internal/engine/configs/configs_controller.go
+++ b/internal/engine/configs/configs_controller.go
@@ -83,10 +83,9 @@ func (cc *ConfigsController) needsBuild(ctx context.Context, st store.RStore) (b
 		reason = reason.With(model.BuildReasonFlagInit)
 	}
 
-	for _, changeTime := range state.PendingConfigFileChanges {
-		if changeTime.After(lastStartTime) {
-			reason = reason.With(model.BuildReasonFlagChangedFiles)
-		}
+	hasPendingChanges, _ := tfState.HasPendingChanges()
+	if hasPendingChanges {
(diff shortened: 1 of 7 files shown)


internal/engine/configs/configs_controller.go
```diff
@@ -101,9 +100,11 @@ func (cc *ConfigsController) needsBuild(ctx context.Context, st store.RStore) (b
 		return buildEntry{}, false
 	}
 
-	filesChanged := make([]string, 0, len(state.PendingConfigFileChanges))
-	for k := range state.PendingConfigFileChanges {
-		filesChanged = append(filesChanged, k)
+	filesChanged := []string{}
+	for _, st := range state.TiltfileState.BuildStatuses {
+		for k := range st.PendingFileChanges {
+			filesChanged = append(filesChanged, k)
```
milas, line 106: Is it possible for this to result in dupes now? (I don't think that'll actually harm anything even if it does)

milas (approved):

internal/engine/configs/configs_controller.go
nicks: nope! i don't think so! but the data model supports multiple Target Statuses, and certainly could imagine a future change where it does matter.

PR #4567, opened 2021-05-19
engine: store Pod update times on RuntimeState

milas:
This was migrated to KubernetesDiscovery but both does not make
sense there as well as doesn't actually work because the build
controller needs to set the value (in engine state), which will
then just get overwritten by the API value (always zero).

The logic here matches the current behavior, which is only used
for crashed container detection and is a bit odd, but this is
just to address the current regression.


pkg/apis/core/v1alpha1/generated.proto
```diff
@@ -516,9 +516,6 @@ message Pod {
   // Tilt uses this to associate Pods with the build that triggered them.
   optional string podTemplateSpecHash = 10;
 
-  // UpdateStartedAt is when Tilt started a deployment update for this Pod.
-  optional k8s.io.apimachinery.pkg.apis.meta.v1.Time updateStartedAt = 11;
```
landism, line 520: do you know how/if the generator avoids reusing deprecated field ids?

landism (approved):

pkg/apis/core/v1alpha1/generated.proto
milas: Did a quick test - any new fields will get +1 of the highest, so interim deleted fields will be ignored, so we'd need to be careful about deleting the highest field number.

(In general, we should probably avoid deleting fields and use multiple versions + conversions, but since this is brand new and totally wrong, removing it seemed to make more sense - nothing external should be relying on the value of this field right now.)

PR #4569, opened 2021-05-19
engine: store BaselineRestarts on RuntimeState

milas:
This is very similar to #4567 in terms of motivation. There's a
bit more changed here to clean up some unnecessary indirection in
the reconciler now that it's simpler.

The key part is to allow the build controller to adjust the baseline
restart for live updates, so that any restarts before the live
update get ignored. As a result, while it'd be nice to have the
KubernetesDiscovery API handle this, it needs to live in RuntimeState
(at least for now).


internal/engine/k8swatch/reducers.go
```diff
@@ -122,9 +123,19 @@ func maybeUpdateStateForPod(ms *store.ManifestState, pod *v1alpha1.Pod) bool {
...
-	if existing := runtime.Pods[podID]; existing != nil && equality.Semantic.DeepEqual(existing, pod) {
+	existing := runtime.Pods[podID]
+	if existing == nil {
+		// the first time a Pod is seen, the sum of restarts across all containers is used as the baseline
+		// restart count so that everything that happened before Tilt was aware of the Pod can be ignored
+		var baselineRestarts int32
+		for i := range pod.Containers {
+			baselineRestarts += pod.Containers[i].Restarts
+		}
+		runtime.BaselineRestarts[podID] = baselineRestarts
```
landism, line 137: IIUC, this is `store.AllPodContainerRestarts`

```suggestion
		runtime.BaselineRestarts[podID] = store.AllPodContainerRestarts(pod)
```

landism (approved):

internal/engine/k8swatch/reducers.go
milas: Ahhh right you are. (I was lazy and cut and pasted and forgot that the type changed so it could use the helper now.)

FYI I don't know how to edit a suggestion before accepting but this needed to be `*pod` so I just made the change manually as ef61c43.

Issue #4424, opened 2021-04-13
Connecting to Tilt Cloud via Tilt UI is not accessible via Voiceover


hyu:
I received a report via email:

> Connecting to the Tilt Cloud via the Tilt UI may not be accessible. I’ve tried with both Chrome and Safari, and VoiceOver recognizes that a dialog has opened (either when I click “Make Snapshot” or “Account” > “Connect Tilt To Tilt Cloud”), but I can’t seem to do any kind of interaction with the dialog at all.

We should check to make sure that the buttons to create the snapshot can be activated via Voiceover.

PR #4571, opened 2021-05-19
web: fix issue with UI becoming inaccessible to assistive tech when snapshot modal is opened

lizzthabet:
This PR fixes issue #4424, where a user reports that they have trouble interacting with the UI using VoiceOver after the `Make a snapshot` button is clicked. I was able to reproduce the issue with VO and saw that `aria-hidden=true` was set on the `body` whenever the snapshot modal was visible. :sob:

The docs for `ReactModal` specify that the (`setAppElement`)[http://reactcommunity.org/react-modal/accessibility/] should be called on the app root so that the component can hide the main app content while the modal is displaying. Updating that method call with the app root id, instead of the document `body` fixed the issue. 

~I'm not sure how we want to fix the tests since the app root isn't available in the testing environment. (Probably just the limits of my react testing knowledge!) `ReactModal` does expose an option to turn off its ARIA-property management which we could add support for. Not setting the app element is an option, too, but it adds a bunch of noisy warning logs to the tests.~

Thanks to @hyu for pairing with me!

### DIFF
--- a/web/src/ShareSnapshotModal.tsx
+++ b/web/src/ShareSnapshotModal.tsx
@@ -24,7 +24,7 @@ export default class ShareSnapshotModal extends PureComponent<props> {
         isOpen={this.props.isOpen}
         className="ShareSnapshotModal"
       >
-        <h2 className="ShareSnapshotModal-title">Share a Shapshot</h2>
+        <h2 className="ShareSnapshotModal-title">Share a Snapshot</h2>
         <section className="ShareSnapshotModal-pane u-flexColumn">
           <p className="ShareSnapshotModal-description">
             Get a link to a{" "}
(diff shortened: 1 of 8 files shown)


web/src/ShareSnapshotModal.tsx
```diff
@@ -24,7 +24,7 @@ export default class ShareSnapshotModal extends PureComponent<props> {
         isOpen={this.props.isOpen}
         className="ShareSnapshotModal"
       >
-        <h2 className="ShareSnapshotModal-title">Share a Shapshot</h2>
+        <h2 className="ShareSnapshotModal-title">Share a Snapshot</h2>
```
lizzthabet, line 27: Typo fix 🤓

nicks (approved):
nice! that was easier than i expected

PR #4573, opened 2021-05-20
local: do not run serve_cmd with auto_init=False on start

### LINKED ISSUE #4548: trigger_mode=TRIGGER_MODE_MANUAL, auto_init=False doesn't work for serve_cmd-only resources
```
local_resource('serve-cmd',
               serve_cmd='echo 🤐; sleep 100000',
               trigger_mode=TRIGGER_MODE_MANUAL, auto_init=False)
```

basically, neither of these commands ought to run without user intervention. the former (an update cmd only) behaves as expected; the latter (a serve_cmd) runs even tho it shouldn’t

You can work around it by adding an update cmd

milas:
If a `local_resource` had no `update_cmd` (i.e. just a `serve_cmd`)
but was set to `auto_init=False`, it would end up starting anyway
because the server controller saw that the dependency status was
`not_applicable` (technically true).

To fix this, the dependency status is set to `none` for local
resources with `auto_init=False` that have not yet had a build
(either from being manually triggered or having `TRIGGER_MODE_AUTO`
and a file change). Once a build has happened, it will resume
returning `not_applicable` to indicate that there is no longer a
reason to block on a dependency.

Fixes #4548.

### DIFF
--- a/internal/store/manifest_target.go
+++ b/internal/store/manifest_target.go
@@ -47,6 +47,16 @@ func (mt *ManifestTarget) UpdateStatus() v1alpha1.UpdateStatus {
 		if us == v1alpha1.UpdateStatusInProgress {
 			return v1alpha1.UpdateStatusPending
 		}
+
+		// If the local_resource has auto_init=False and has not built yet (i.e. it has
+		// not been manually triggered), use UpdateStatusNone to indicate that the resource
+		// has not yet had a reason to trigger so that it blocks the serve_cmd from executing.
+		// Once manually triggered, a no-op build will exist, and subsequent calls will return
+		// UpdateStatusNotApplicable so that the server controller knows it does not need to
+		// wait for anything.
+		if !m.TriggerMode.AutoInitial() && !mt.State.StartedFirstBuild() {
+			return v1alpha1.UpdateStatusNone
+		}
 		return v1alpha1.UpdateStatusNotApplicable
 	}
 
--- a/internal/store/manifest_target.go
+++ b/internal/store/manifest_target.go
@@ -29,11 +29,6 @@ func (mt *ManifestTarget) UpdateStatus() v1alpha1.UpdateStatus {
 	m := mt.Manifest
 	us := mt.State.UpdateStatus(m.TriggerMode)
 
-	if us == v1alpha1.UpdateStatusPending {
-		// A resource with no update command can still be in pending mode.
-		return us
-	}
-
 	if m.IsLocal() && m.LocalTarget().UpdateCmd.Empty() {
 		// NOTE(nick): We currently model a local_resource(serve_cmd) as a Manifest
 		// with a no-op Update. BuildController treats it like any other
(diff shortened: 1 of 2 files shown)


internal/store/manifest_target.go
```diff
@@ -47,6 +47,16 @@ func (mt *ManifestTarget) UpdateStatus() v1alpha1.UpdateStatus {
...
 			return v1alpha1.UpdateStatusPending
 		}
+
+		// If the local_resource has auto_init=False and has not built yet (i.e. it has
+		// not been manually triggered), use UpdateStatusNone to indicate that the resource
+		// has not yet had a reason to trigger so that it blocks the serve_cmd from executing.
+		// Once manually triggered, a no-op build will exist, and subsequent calls will return
+		// UpdateStatusNotApplicable so that the server controller knows it does not need to
+		// wait for anything.
+		if !m.TriggerMode.AutoInitial() && !mt.State.StartedFirstBuild() {
```
nicks, line 57: Hmmmm...i'm not sure this is right. a couple obvious smells that leap out to me:

- Currently mt.State.UpdateStatus() does this same check. why does this check live in both places?
- Other things read UpdateStatus too (like the UI), will they also do the right thing in this case?
- I'm also not sure I can articulate the difference between "UpdateStatusNone" and "UpdateStatusNotApplicable". can you?

nicks: a check that might make more sense to me here is

```
	if us == v1alpha1.UpdateStatusNone {
		return us
	}
```

which is closer to the check above

milas: Okay, I simplified the logic so that it only explicitly handles the special cases -- there's _technically_ slight semantic change here now which is that `UpdateStatusError` won't get coerced to `UpdateStatusNotApplicable` anymore, but I think that's a good thing because that'd be indicative of an engine bug (i.e. no-op build "failing" should be impossible) and would probably result in UI being in a confusing/inconsistent state.

RE: The approach, I agree it's odd. IMO while it is kludgy it's at least consistent vs a non-existent update being allowed to be `Pending` but not `None`, i.e. it should either _always_ be `NotApplicable` or it should generally follow the rules. (FWIW I'm also a bit confused why it coerces `InProgress` -> `Pending` instead of making `InProgress` a "hold" status in server controller, but that's a separate issue.)

Alternatively, we can add special logic to server controller for N/A statuses to do a `!m.TriggerMode.AutoInitial() && !mt.State.StartedFirstBuild()` check?

For difference between `None` / `NotApplicable` -- I agree, I think the name is unfortunate but the docs/usage are consistent since it's what gets sued for anything that's `auto_init=False`: https://github.com/tilt-dev/tilt/blob/87a7f57d91aa0685b076e80eb54cfcd63bf0e83f/pkg/apis/core/v1alpha1/updatestatus_types.go#L8-L10

Thankfully the UI does handle it appropriately (also confirmed by actually running): https://github.com/tilt-dev/tilt/blob/87a7f57d91aa0685b076e80eb54cfcd63bf0e83f/web/src/status.tsx#L12-L15

nicks (approved):

nicks:
ya, i like this approach!

PR #4584, opened 2021-05-24
engine: double-write uisession and uiresource

nicks:
Hello @milas, @maiamcc,

Please review the following commits I made in branch nicks/uisession14:

78f8d7384324980a07aeba96045702bcbf802344 (2021-05-24 17:35:21 -0400)
engine: double-write uisession and uiresource

### DIFF
--- a/internal/engine/uiresource/subscriber.go
+++ b/internal/engine/uiresource/subscriber.go
@@ -3,23 +3,28 @@ package uiresource
 import (
 	"context"
 	"fmt"
+	"strings"
 
-	"k8s.io/apimachinery/pkg/api/equality"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
+	ctrlclient "sigs.k8s.io/controller-runtime/pkg/client"
 
+	"github.com/tilt-dev/tilt/internal/controllers/apicmp"
 	"github.com/tilt-dev/tilt/internal/hud/webview"
 	"github.com/tilt-dev/tilt/internal/store"
 	"github.com/tilt-dev/tilt/pkg/apis/core/v1alpha1"
+	"github.com/tilt-dev/tilt/pkg/logger"
 )
 
 // Creates UIResource objects from the EngineState
 type Subscriber struct {
-	lastResources map[types.NamespacedName]*v1alpha1.UIResource
+	client ctrlclient.Client
 }
 
-func NewSubscriber() *Subscriber {
+func NewSubscriber(client ctrlclient.Client) *Subscriber {
 	return &Subscriber{
-		lastResources: make(map[types.NamespacedName]*v1alpha1.UIResource),
+		client: client,
 	}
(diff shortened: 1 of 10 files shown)


internal/engine/uiresource/subscriber.go
```diff
@@ -62,13 +72,38 @@ func (s *Subscriber) OnChange(ctx context.Context, st store.RStore, summary stor
 			// If there's a current version but no last version of this resource,
 			// create it.
 			st.Dispatch(NewUIResourceCreateAction(current))
+			err := s.client.Create(ctx, current.DeepCopy())
```
milas, line 75: I think for `Create` and `Update`, you'll want to _not_ `DeepCopy()` and only dispatch after the API operation. It'll mutate the object (e.g. populate UID, creation timestamp, resource version on create; increment resource version on update) so that's the version we should pass to engine (and store here to ensure that optimistic concurrency on update works as expected)

internal/engine/uiresource/subscriber.go
```diff
@@ -62,13 +72,38 @@ func (s *Subscriber) OnChange(ctx context.Context, st store.RStore, summary stor
...
 
 		if !equality.Semantic.DeepEqual(last.Status, current.Status) {
 			// If the current version is different than the last version, update it.
 			st.Dispatch(NewUIResourceUpdateStatusAction(current))
+
+			freshResource := v1alpha1.UIResource{}
+			err := s.client.Get(ctx, types.NamespacedName{Name: current.Name}, &freshResource)
+			if err != nil {
+				if apierrors.IsNotFound(err) {
+					delete(s.lastResources, name)
```
milas, line 92: I feel like this could cause trouble unless with the latest fixes cache mean we won't **ever** return not found for a new/existing object - if that happens, this will result in a create next invocation and get an already exists error (which will throw a store error).

That said, the alternative (ignoring, hoping it's stale cache and moving on) caused a whole different set of problems for KD when cache never caught up, so maybe we really do need to start doing some sort of upsert.

FWIW I just did some quick and dirty testing and am NOT seeing not founds anymore on `master` even with a bunch of entities, so this is probably fine as is.

internal/engine/uiresource/subscriber.go
```diff
@@ -62,13 +72,38 @@ func (s *Subscriber) OnChange(ctx context.Context, st store.RStore, summary stor
...
+					continue
+				}
+
+				st.Dispatch(store.NewErrorAction(fmt.Errorf("fetching resource %s: %v", current.Name, err)))
+				return
+			}
+
+			freshResource.Status = *freshResource.Status.DeepCopy()
+
+			err = s.client.Status().Update(ctx, &freshResource)
```
milas, line 102: Are these subscribers the exclusive updaters of resource/session respectively? If not, I think on the update at least it's reasonable to ignore a conflict (and not update internal state).

milas (approved):
Couple data flow suggestions/questions but given the current usage patterns nothing should explode 😉

nicks:
actually, i'm going to try to redo this without the lastSession stuff - i think this is the wrong pattern and just causes more problems. hang on

PR #4586, opened 2021-05-25
webview: only send updated objects

nicks:
Hello @landism, @milas,

Please review the following commits I made in branch nicks/incremental:

0dbdfd17046b9cbb00e62a6f70f3e3eeb43069d7 (2021-05-25 16:12:36 -0400)
webview: only send updated objects
This has two big benefits:

1) We want to make it easier to read directly from the APIServer,
   which uses List/Watch model for API access

2) Web UI performance should be much better and easier to optimize
   going forward, since we know exactly what changed and
   don't need to do a lot of JSON parsing/re-rendering
   for objects that didn't change


landism (approved):
nice! a lot simpler than I thought it was going to be!

PR #4589, opened 2021-05-26
Draft: (spike) Add logs filter input with sample implementation and notes

hyu:


lizzthabet:
Closing since this is just the spike code

Issue #4544, opened 2021-05-11
Can't start tilt with multi node minikube clusters using docker driver


code-merc:
### Expected Behavior

Running `tilt up` should work with a multi node cluster

### Current Behavior

`tilt up` executes, but the tilt web interface logs that the tiltfile could not be executed with this error
```
Failed to connect to Docker: Could not read docker env from minikube.
Did you forget to run `minikube start`?
X Exiting due to ENV_MULTINODE_CONFLICT: The docker-env command is incompatible with multi-node clusters. Use the 'registry' add-on: https://minikube.sigs.k8s.io/docs/handbook/registry/
```

### Steps to Reproduce

1. Create a minikube cluster like this ` minikube start --driver=docker --nodes=2 --addons=default-storageclass,registry,storage-provisioner`
2. Run `tilt up`
3. See problem above

#### `tilt doctor` Output

```
$ tilt doctor
Tilt: v0.20.0, built 2021-04-30
System: linux-amd64
---
Docker (cluster)
- Host: Error: Could not read docker env from minikube.
Did you forget to run `minikube start`?
X Exiting due to ENV_MULTINODE_CONFLICT: The docker-env command is incompatible with multi-node clusters. Use the 'registry' add-on: https://minikube.sigs.k8s.io/docs/handbook/registry/

---
Docker (local)
- Host: [default]
- Version: 1.40
- Builder: 2
---
Kubernetes
- Env: minikube
- Context: minikube
- Cluster Name: minikube
- Namespace: default
- Container Runtime: docker
- Version: v1.20.2
- Cluster Local Registry: none
---

```

I think  the issue is [here](https://github.com/tilt-dev/tilt/blob/77ae37f8f66c4b2d6df5eb526bd71954be73ecf2/internal/k8s/minikube.go#L68), running `minikube docker-env` on a multi node cluster with the docker driver fails with 
```
❌  Exiting due to ENV_MULTINODE_CONFLICT: The docker-env command is incompatible with multi-node clusters. Use the 'registry' add-on: https://minikube.sigs.k8s.io/docs/handbook/registry/
```

Enabling/disabling the registry addon doesn't make a difference either.

code-merc:
By the way this same issue comes up when running with a VM driver so `minikube start --driver=kvm2 --nodes=2 --addons=default-storageclass,registry,storage-provisioner` causes tilt to error the same as with docker.

Issue #4587, opened 2021-05-26
Add option to disable pushing images manually


zaro:
## Describe the Feature You Want
Right now tilt disables push only if k8s is minikube ( and few others). I've made a setup with `k3s --docker` which basically makes k3s to use the already installed docker daemon. This leads to a push which is not needed at all because after the build I always the get message ` Container image "..." already present on machine`. This is normal as both k3s and tilt use the same docker daemon.

### Current Behavior
Needless push if using tilt with `k3s  --docker` as dev cluster.

### Why Do You Want This?
I would like to skip waiting for a push which is not needed anyway.

nicks:
thanks for the report! I'm going to mark this as a duplicate of https://github.com/tilt-dev/tilt/issues/3654, since i think they're basically the same issue

Issue #3654, opened 2020-08-03
Support k3s in --docker mode


leoluk:
k3s can use a local Docker daemon in `--docker` mode.

This is similar to Docker for Desktop mode, where built images are immediately available in the cluster.

Currently, this isn't properly detected.

nicks:
Hmmm....can you link to documentation on this? 

This is the first I've heard of this, and we do talk to the K3d team about this sort of interop from time to time.

nicks:
(In general, what we would do in this case is ask the K3s team to provide some sort of protocol for determining when K3s is in this mode, like a ConfigMap created by the cluster, and then Tilt would read this ConfigMap...so I would need to read more about how they're advertising/documenting this)

leoluk:
By the way, I tried to pretend that's it's a `docker-for-windows` cluster by renaming the context, and it works flawlessly!

leoluk:
Reproduction instructions:

```
curl -sfL https://get.k3s.io | sh -s - --docker
export KUBECONFIG=/etc/rancher/k3s/k3s.yaml
sed -i 's/  name: default/  name: docker-desktop/g' $KUBECONFIG
sed -i 's/current-context: default/current-context: docker-desktop/g' $KUBECONFIG
sed -i 's/cluster: default/cluster: docker-desktop/g' $KUBECONFIG

tilt up
```

nicks:
Hmmm....those instructions are about setting up K3s on a Node.

I'm not sure this is a safe or recommended way to run K3s for local development, but let me point some people in the K3s project at this issue and see what they say.

We usually see people using K3d to run K3s for local development, see our instructions here: https://github.com/tilt-dev/k3d-local-registry

leoluk:
Perhaps the real feature request here is being able to disable image pushing?

Skaffold has a local override: https://skaffold.dev/docs/environment/local-cluster/#manual-override

nicks:
Ya, here's an overview of some of the problems with this approach:
https://minikube.sigs.k8s.io/docs/drivers/none/

You can disable image pushing today with Tilt with custom_build(disable_push) https://docs.tilt.dev/api.html#api.custom_build

[...]

Issue #1729, opened 2019-06-10
support minikube with --vm-driver=none


nicks:
Currently, if you start minikube with --vm-driver=none, Tilt does not work properly.

Seems to be some problem with how it's using `minikube docker-env`, needs investigation

Reported by Jonas in the k8s.slack.io #tilt channel

JohnnyCrazy:
Just to shed some more light here:

When running `tilt doctor`, this is the output with a vm-driver=none cluster:

```
➜ tilt doctor
Tilt: v0.8.9, built 2019-06-07
System: linux-amd64
---
Docker
- Host: Error: ProvideDockerEnv: Could not read docker env from minikube.
Did you forget to run `minikube start`?
X The docker host is currently not running

- Version: Error: ProvideDockerEnv: Could not read docker env from minikube.
Did you forget to run `minikube start`?
X The docker host is currently not running

- Builder: Error: ProvideDockerEnv: Could not read docker env from minikube.
Did you forget to run `minikube start`?
X The docker host is currently not running

---
Kubernetes
- Env: minikube
- Context: minikube
- Cluster Name: minikube
- Namespace: default
- Container Runtime: docker
- Version: v1.14.2
---
Thanks for seeing the Tilt Doctor!
Please send this info along when filing bug reports. 💗
```

Two problems have to be solved:
* With `vmdriver=none`, minikube has to be started with elevated rights `sudo`. Thus, running just `minikube docker-env` will not find the running cluster. I'm not entirely sure how we could solve this one...
* `sudo minikube docker-env` reports `'none' driver does not support 'minikube docker-env' command`. In this case, tilt should fall back to the hosts docker installation IMO?

nicks:
a simple thing to do might be to check if the machine id in `kubectl get nodes -o json` matches the current machine id. And if that's true, we skip the `minikube docker-env` call because we know it has to be the current env.

nicks:
If someone wants to contribute a PR for this, we'd accept it.

I'm concerned that --vm-driver=none is designed for the case "I built my own VM", NOT "I want to do local k8s development without a VM". And that the minikube team may make changes that break the second case. See:

https://github.com/kubernetes/minikube/blob/master/docs/vmdriver-none.md#can-the-none-driver-be-used-outside-of-a-vm

I wonder if we should be endorsing microk8s more strongly for the second case in the Tilt documentation

dlpatri:
[...]

PR #4598, opened 2021-05-27
docker: change how we model "image builds show up in the cluster immediately"

nicks:
Hello @maiamcc,

Please review the following commits I made in branch nicks/ch12079:

cb46eee7205d6078b33247a62901e2d8b47e731f (2021-05-27 17:28:54 -0400)
docker: change how we model "image builds show up in the cluster immediately"
We used to treat this as a property of the cluster type + the container runtime.

But this made it impossible to support clusters that sometimes use your docker
runtime, and sometimes do not.

For examples, see:
- https://github.com/tilt-dev/tilt/issues/4587
- https://github.com/tilt-dev/tilt/issues/3654
- https://github.com/tilt-dev/tilt/issues/1729
- https://github.com/tilt-dev/tilt/issues/4544

This changes the data model so that "image builds show up in the cluster"
is a property of the Docker client, not of the cluster.

This should be much more flexible and correct, and help us support multiple
clusters.

Fixes https://github.com/tilt-dev/tilt/issues/4544

### DIFF
--- a/internal/docker/env.go
+++ b/internal/docker/env.go
@@ -33,9 +33,31 @@ type Env struct {
 	// https://github.com/kubernetes/minikube/issues/4143
 	IsOldMinikube bool
 
+	// Some Kubernetes contexts have a Docker daemon that they use directly
+	// as their container runtime. Any images built on that daemon will
+	// show up automatically in the runtime.
+	//
+	// We store this as a field of the Docker Env, because this
+	// really affects how we interact with the Docker Env (rather than
+	// how we interact with the K8s Env).
+	//
+	// In theory, you can have more than one, but in practice,
+	// this is very difficult to set up.
+	BuildToKubeContexts []string
+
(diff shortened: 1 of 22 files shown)


internal/docker/env.go
```diff
@@ -33,11 +33,33 @@ type Env struct {
 	// https://github.com/kubernetes/minikube/issues/4143
 	IsOldMinikube bool
 
+	// Some Kubernetes contexts have a Docker daemon that they use directly
+	// as their container runtime. Any images built on that daemon will
+	// show up automatically in the runtime.
+	//
+	// We store this as a field of the Docker Env, because this
```
maiamcc, line 40:
```suggestion
	// We used to store this as a property of the k8s env but now store it as a field of the Docker Env, because this
```

maiamcc (approved):

PR #4600, opened 2021-05-28
cmd: add support for StartOn and UIButton

landism:
1. Rename `RestartOnSpec` to `TriggerSpec`
2. Add a `StartOn TriggerSpec` to `Cmd`
3. Add `UIButton` to `TriggerSpec`
4. Implement support for `StartOn` and `UIButton` in the `Cmd` reconciler


landism:
A couple notes:
1. There's a lot of ugly copy/paste for filewatch + uibutton in `TriggerManager`. I figure when we add a third type we can worry about making it a bit more generic.
2. There are at least three different rules we'll want for how a trigger interacts with an existing build:
  a. restart the process, no matter what (the existing `RestartOn` semantics)
  b. start the process if there isn't an existing process (the existing `StartOn` semantics)
  c. queue a new process to start after the existing process (the existing `local_resource` `update_cmd` semantics)
  Yesterday I had a fourth, but I can't recall at the moment. At some point we might want to change TriggerSpec to something like a list of name/kind/behaviorOnTrigger. That doesn't seem worth worrying about at the moment.

landism:
note: hit this error in CI and currently debugging: `panic: no error action allowed: syncing to apiserver: Operation cannot be fulfilled on cmds.tilt.dev "testcmd": object was modified`

nicks:
Hmmmm....not sure I agree with the `Rename RestartOnSpec to TriggerSpec` API change, that seems like a pretty major breaking change.

nicks:
can you change RestartOn and RestartManager stuff to the way it was? that doesn't seem related to the rest of the PR

nicks:
ah, wait, nevermind, ignore the last two comments, i understand now, i misread something

internal/controllers/core/cmd/controller.go
```diff
@@ -103,15 +103,26 @@ func (c *Controller) reconcile(ctx context.Context, name types.NamespacedName) e
...
 	proc, ok := c.procs[name]
+	// there is already a proc (which may or may not be running) for this cmd
 	if ok {
-		needsRestart := !equality.Semantic.DeepEqual(proc.spec, cmd.Spec) ||
-			lastEventTime.After(proc.lastEventTime)
+		// any change to the spec means we should restart the process to pick up the changes
+		specChanged := !equality.Semantic.DeepEqual(proc.spec, cmd.Spec)
+		// a new restart event happened
+		restartOnTriggered := lastRestartEventTime.After(proc.lastRestartEventTime)
+		// a new start event happened *and* the existing proc is finished
```
nicks, line 122: Hmmmm...I'm concerned this API isn't going to behave like you expect.

A good rule of thumb with this kind of distributed API is that it needs to be idempotent - if the controller accidentally gets the event twice or if the timing is a bit off, that's OK.

So we really don't want an API that behaves differently gets ignored if the cmd is running and does an action if the cmd is completed. I think that's a recipe for lots of unresolvable problems and race conditions.

landism:
> Hmmmm....not sure I agree with the `Rename RestartOnSpec to TriggerSpec` API change, that seems like a pretty major breaking change.

Whoops, I meant to be explicit about that - do we have any clients yet that we care about breaking?

Is it time for our first v1alpha2?

pkg/apis/core/v1alpha1/cmd_types.go
```diff
@@ -213,8 +220,8 @@ func (in CmdStatus) CopyTo(parent resource.ObjectWithStatusSubResource) {
 	parent.(*Cmd).Status = in
 }
 
-// RestartOnSpec indicates the set of objects that can trigger a restart of this object.
-type RestartOnSpec struct {
+type TriggerSpec struct {
 	// A list of file watches that can trigger a restart.
-	FileWatches []string `json:"fileWatches" protobuf:"bytes,1,rep,name=fileWatches"`
+	FileWatches []string `json:"fileWatches,omitempty" protobuf:"bytes,1,rep,name=fileWatches"`
+	UIButtons   []string `json:"uiButtons,omitempty" protobuf:"bytes,2,rep,name=uiButtons"`
```
nicks, line 226: can you add comments to this field?

pkg/apis/core/v1alpha1/cmd_types.go
```diff
@@ -93,7 +93,14 @@ type CmdSpec struct {
 	// Restarts can happen even if the command is already done.
 	//
 	// Logs of the current process after the restart are discarded.
-	RestartOn *RestartOnSpec `json:"restartOn,omitempty" protobuf:"bytes,5,opt,name=restartOn"`
+	RestartOn *TriggerSpec `json:"restartOn,omitempty" protobuf:"bytes,5,opt,name=restartOn"`
+
+	// Indicates objects that can trigger a run of this command.
```
nicks, line 98: We will probably need to support the case where:

1) User creates a Button and CmdA
2) User clicks the Button a few times
3) User deletes CmdA and creates CmdB

where we wouldn't want CmdB to run until the next button press. How would we model that scenario?

pkg/apis/core/v1alpha1/cmd_types.go
```diff
@@ -93,7 +93,14 @@ type CmdSpec struct {
...
+	//
+	// If there is a StartOn rule, the cmd's process will not be started until the first
+	// StartOn trigger occurs.
+	// StartOn triggers are ignored while the cmd's process is running.
+	StartOn *TriggerSpec `json:"startOn,omitempty" protobuf:"bytes,6,opt,name=startOn"`
```
nicks, line 103: I'm a little bit nervous about representing StartOn and RestartOn with the same data model...my personal rule of thumb is to always create a new data model unless there's seem deep foundational reason why they have to be the same

For example, I could imagine modelling StartOn as having some "After" time that tells it to ignore events before a certain time, or having RestartOn having some "restart on timeout" field that wouldn't make sense for StartOn.

thoughts?

nicks (commented):
excited about this! I have a few high-level API questions that we should probably get out of the way first.

pkg/apis/core/v1alpha1/cmd_types.go
landism: I've added a StartAfter to StartOn

pkg/apis/core/v1alpha1/cmd_types.go
landism: Yeah, makes sense. This was mostly motivated by wanting to use the same Restart/TriggerManager code for both, but 1) it's not that much code to repeat (and the introduction of StartAfter actually means their logic is slightly different now) and 2) if the repetition later becomes a problem we can deal with that in a way that doesn't expose it to the API.

landism:
whoops, merged in a way that screwed up history. ignore pushes for now.

PR #4607, opened 2021-06-02
docker: support DOCKER_HOST=ssh://... remotes

### LINKED ISSUE #4603: DOCKER_HOST using ssh protocol makes tilt up fail
### Expected Behavior

`tilt up` works when using SSH protocol in `DOCKER_HOST` environment variable

### Current Behavior

tilt up fails with

```
Beginning Tiltfile execution
Successfully loaded Tiltfile (10.47725ms)
Failed to connect to Docker: error during connect: Get "http://user%40host/v1.24/version": dial tcp: lookup user@host: no such host
```

when using `DOCKER_HOST=ssh://user@host`.

### Steps to Reproduce

1. set `DOCKER_HOST` to use ssh protocol (e.g. `export DOCKER_HOST=ssh://user@host`)
2. create Tiltfile containing `docker_build` function
3. run `tilt up`

#### `tilt doctor` Output

```
$ tilt doctor
Tilt: v0.20.4, built 2021-05-24
System: darwin-arm64
---
Docker
- Host: Error: error during connect: Get "http://user%40host/v1.24/version": dial tcp: lookup user@host: no such host
---
Kubernetes
- Env: k3d
- Context: k3d-clamav-rest
- Cluster Name: k3d-clamav-rest
- Namespace: default
- Version: v1.20.4+k3s1
---
```

#### About Your Use Case

I want to make use of a remote docker host via SSH.

milas:
Docker 18.09+ added support for SSH remote Docker connections.
The code to handle this isn't actually in the Moby client portions
that we use, but in the Docker CLI (which we already have a
dependency on).

This adapts some of the code from the Docker CLI that attempts
to create a `ConnectionHelper` if the host value is set. Currently,
the logic only creates a helper for `ssh://` hosts and returns nil
otherwise, in which case the former logic is used (i.e. pass the
host as-is to Moby). If a connection helper is created, it's used
to create a custom HTTP client/transport that will use SSH under
the hood.

Fixes #4603.

### DIFF
--- a/internal/docker/client.go
+++ b/internal/docker/client.go
@@ -13,6 +13,8 @@ import (
 	"sync"
 	"time"
 
+	"github.com/docker/cli/cli/connhelper"
+
 	"github.com/blang/semver"
 	"github.com/docker/cli/cli/command"
 	"github.com/docker/cli/cli/config"
@@ -237,7 +239,7 @@ func SupportsBuildkit(v types.Version, env Env) bool {
 // DOCKER_API_VERSION to set the version of the API to reach, leave empty for latest.
 // DOCKER_CERT_PATH to load the TLS certificates from.
 // DOCKER_TLS_VERIFY to enable or disable TLS verification, off by default.
-func CreateClientOpts(ctx context.Context, env Env) ([]client.Opt, error) {
+func CreateClientOpts(_ context.Context, env Env) ([]client.Opt, error) {
 	result := make([]client.Opt, 0)
 
 	if env.CertPath != "" {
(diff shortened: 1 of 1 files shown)


milas:
Unfortunately, there's no great way to test this in CI that I can find. CircleCI actually configures the remote docker to be SSHable ([docs](https://circleci.com/docs/2.0/building-docker-images/#accessing-the-remote-docker-environment)) but there's enough other complexity/weirdness with port forwarding etc that it doesn't seem practical.

landism:
> Unfortunately, there's no great way to test this in CI that I can find. CircleCI actually configures the remote docker to be SSHable ([docs](https://circleci.com/docs/2.0/building-docker-images/#accessing-the-remote-docker-environment)) but there's enough other complexity/weirdness with port forwarding etc that it doesn't seem practical.

If the next person to touch this is tempted to add a test, having some form of this comment in the code might save them some time

landism (approved):

milas:
Good call - added a note in 61089f5c8. I really wanted to get a variation of one of the integration tests working, but there's just too many oddities to make it practical right now

PR #4608, opened 2021-06-02
(web) Add term input field to filters bar to filter resource logs by string

lizzthabet:
*  Add support for filtering log spans based on a string input that behaves like the existing source and term filters
*  Add a term input field in the filter bar with a debounce delay; (this does behave a tiny bit quirky which i outlined in a code note. I think this is probably okay for mvp, but happy to chat more in-depth if needed.)

Styles and tests are still a wip, but otherwise code is ready for review. 

Term input field:
Screen Shot 2021-06-02 at 5 47 34 PM

Filtering logs by term and level filters:

Closes ticket #12046

### DIFF
--- a/web/src/OverviewActionBar.stories.tsx
+++ b/web/src/OverviewActionBar.stories.tsx
@@ -1,7 +1,13 @@
 import { createMemoryHistory } from "history"
 import React from "react"
 import { Router } from "react-router"
-import { FilterLevel, FilterSource, useFilterSet } from "./logfilters"
+import {
+  EMPTY_FILTER_TERM,
+  FilterLevel,
+  FilterSet,
+  FilterSource,
+  useFilterSet,
+} from "./logfilters"
 import OverviewActionBar from "./OverviewActionBar"
 import { StarredResourceMemoryProvider } from "./StarredResourcesContext"
 import { oneButton, oneResource } from "./testdata"
@@ -41,7 +47,11 @@ export default {
   },
 }
 
-let defaultFilter = { source: FilterSource.all, level: FilterLevel.all }
+let defaultFilter: FilterSet = {
+  source: FilterSource.all,
+  level: FilterLevel.all,
+  term: EMPTY_FILTER_TERM,
+}
 
 export const OverflowTextBar = () => {
   let filterSet = useFilterSet()
--- a/web/src/OverviewActionBar.stories.tsx
+++ b/web/src/OverviewActionBar.stories.tsx
@@ -33,12 +33,14 @@ export default {
   ],
   argTypes: {
     source: {
+      name: "Source",
       control: {
         type: "select",
         options: [FilterSource.all, FilterSource.build, FilterSource.runtime],
       },
     },
     level: {
+      name: "Level",
       control: {
         type: "select",
         options: [FilterLevel.all, FilterLevel.warn, FilterLevel.error],
--- a/web/src/OverviewActionBar.tsx
+++ b/web/src/OverviewActionBar.tsx
@@ -1,25 +1,48 @@
+import {
+  debounce,
+  InputAdornment,
+  InputProps,
+  TextField,
+} from "@material-ui/core"
 import Menu from "@material-ui/core/Menu"
 import MenuItem from "@material-ui/core/MenuItem"
 import { PopoverOrigin } from "@material-ui/core/Popover"
 import { makeStyles } from "@material-ui/core/styles"
 import ExpandMoreIcon from "@material-ui/icons/ExpandMore"
 import moment from "moment"
-import React, { useRef, useState } from "react"
+import { History } from "history"
+import React, { ChangeEvent, useRef, useState } from "react"
 import { useHistory, useLocation } from "react-router"
 import styled from "styled-components"
 import { Alert } from "./alerts"
 import { incr } from "./analytics"
 import { ReactComponent as CheckmarkSvg } from "./assets/svg/checkmark.svg"
+import { ReactComponent as CloseSvg } from "./assets/svg/close.svg"
 import { ReactComponent as CopySvg } from "./assets/svg/copy.svg"
+import { ReactComponent as FilterSvg } from "./assets/svg/filter.svg"
 import { ReactComponent as LinkSvg } from "./assets/svg/link.svg"
 import { InstrumentedButton } from "./instrumentedComponents"
 import { displayURL } from "./links"
 import LogActions from "./LogActions"
-import { FilterLevel, FilterSet, FilterSource } from "./logfilters"
+import {
+  EMPTY_TERM,
+  FilterLevel,
+  FilterSet,
+  FilterSource,
+  TermState,
+} from "./logfilters"
 import { useLogStore } from "./LogStore"
 import OverviewActionBarKeyboardShortcuts from "./OverviewActionBarKeyboardShortcuts"
 import { usePathBuilder } from "./PathBuilder"
-import { AnimDuration, Color, Font, FontSize, SizeUnit } from "./style-helpers"
+import SrOnly from "./SrOnly"
+import {
+  AnimDuration,
+  Color,
+  Font,
+  FontSize,
+  mixinResetButtonStyle,
+  SizeUnit,
+} from "./style-helpers"
 import { ResourceName } from "./types"
 
 type UIResource = Proto.v1alpha1UIResource
@@ -73,9 +96,7 @@ function FilterSourceMenu(props: FilterSourceMenuProps) {
   let l = useLocation()
   let onClick = (e: any) => {
     let source = e.currentTarget.getAttribute("data-filter")
-    let search = new URLSearchParams(l.search)
-    search.set("source", source)
-    search.set("level", level)
+    const search = createLogSearch(l.search, { source, level })
     history.push({
       pathname: l.pathname,
       search: search.toString(),
@@ -236,6 +257,43 @@ export let ButtonRightPill = styled(ButtonRoot)`
   border-radius: 0 4px 4px 0;
 `
 
+const FilterTermTextField = styled(TextField)`
+  & .MuiOutlinedInput-root {
+    border-radius: ${SizeUnit(0.125)};
+    border: 1px solid ${Color.grayLighter};
+    background-color: ${Color.gray};
+    transition: border-color ${AnimDuration.default} ease;
+
+    &:hover {
+      border-color: ${Color.blue};
+    }
+    & fieldset {
+      border-color: 1px solid ${Color.grayLighter};
+    }
+    &:hover fieldset {
+      border: 1px solid ${Color.grayLighter};
+    }
+    & .Mui-focused fieldset {
+      border: 1px solid ${Color.grayLighter};
+    }
+    & .MuiOutlinedInput-input {
+      padding: ${SizeUnit(0.2)};
+    }
+  }
+
+  & .MuiInputBase-input {
+    font-family: ${Font.monospace};
+    color: ${Color.gray7};
+    font-size: ${FontSize.small};
+  }
+`
+
+const ClearFilterTermTextButton = styled(InstrumentedButton)`
+  ${mixinResetButtonStyle}
+  align-items: center;
+  display: flex;
+`
+
 type FilterRadioButtonProps = {
   // The level that this button toggles.
   level: FilterLevel
@@ -245,6 +303,32 @@ type FilterRadioButtonProps = {
 
   // All the alerts for the current resource.
   alerts?: Alert[]
+}
+
+function createLogSearch(
+  currentSearch: string,
+  {
+    level,
+    source,
+    term,
+  }: { level?: FilterLevel; source?: FilterSource; term?: string }
+) {
+  // Start with the existing search params
+  const newSearch = new URLSearchParams(currentSearch)
+
+  if (level !== undefined) {
+    newSearch.set("level", level)
+  }
+
+  if (source !== undefined) {
+    newSearch.set("source", source)
+  }
+
+  if (term !== undefined) {
+    newSearch.set("term", term)
+  }
+
+  return newSearch
 }
 
 export function FilterRadioButton(props: FilterRadioButtonProps) {
@@ -291,9 +375,10 @@ export function FilterRadioButton(props: FilterRadioButtonProps) {
   let history = useHistory()
   let l = useLocation()
   let onClick = () => {
-    let search = new URLSearchParams(l.search)
-    search.set("level", level)
-    search.set("source", "")
+    const search = createLogSearch(l.search, {
+      level,
+      source: FilterSource.all,
+    })
     history.push({
       pathname: l.pathname,
       search: search.toString(),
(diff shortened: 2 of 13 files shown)


web/src/OverviewLogPane.tsx
```diff
@@ -66,11 +66,26 @@ let LogEnd = styled.div`
 
 let anser = new Anser()
 
+// Currently unused / placeholder if we want to implement text highlighting
+function newLineText(text: string, textToHighlight?: string): string {
```
lizzthabet, line 70: I'll clean this up!

web/src/OverviewLogPane.tsx
```diff
@@ -375,6 +375,7 @@ export class OverviewLogComponent extends Component<OverviewLogComponentProps> {
   }
 
   resetRender() {
+    // console.log("running resetRender")
```
lizzthabet, line 378: Will remove this ✂️

landism (commented):
The epic and docs say:
1. you can filter by logs span EXCLUDING [term]
2. the search term is a raw string *unless* surrounded by `//`, in which case it's a regexp

Are these still planned?

web/src/OverviewLogPane.tsx
```diff
@@ -399,6 +400,20 @@ export class OverviewLogComponent extends Component<OverviewLogComponentProps> {
...
   }
 
+  matchesTermFilter(line: LogLine): boolean {
+    const { filterTerm } = this.props.filterSet.term
+
+    if (!filterTerm) {
+      return true
+    }
+
+    if (line.text && filterTerm.test(line.text)) {
```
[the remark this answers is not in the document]
landism: > The epic and docs say:
> 
> 1. you can filter by logs span EXCLUDING [term]
> 2. the search term is a raw string _unless_ surrounded by `//`, in which case it's a regexp
> 
> Are these still planned?

I guess if we want to use the Chrome console filter as an example, that gives answers to lots of the comments on this PR (like a tooltip that says "e.g. /error/ ~rectangler resource:storage"), but maybe that work comes in future PRs.

web/src/SrOnly.tsx
```diff
@@ -0,0 +1,28 @@
+import { Typography } from "@material-ui/core"
+import React, { ElementType } from "react"
+
+// Question for the audience: Is there a type that covers just normal React-supported html properties, so I don't have to make my own type? This component may need to receive ARIA markup and other properties, like `htmlFor`.
```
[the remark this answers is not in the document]
landism: this is actually arcane enough that I might add a comment along the lines of `// copypasta'd from Typography`

lizzthabet:
> The epic and docs say:
> 
>     1. you can filter by logs span EXCLUDING [term]
> 
>     2. the search term is a raw string _unless_ surrounded by `//`, in which case it's a regexp
> 
> 
> Are these still planned?

They are! I broke the epic into smaller stories, so this first one is #12046 and just covers the string search. I was planning to do the other tickets (including regexp) in the epic as fast follows.

web/src/OverviewLogPane.tsx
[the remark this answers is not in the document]
lizzthabet: I like this! I don't have a strong opinion about what users might expect by default, but if there is any / some frustration about it matching on the manifest name, it'd be easy enough to change.

web/src/OverviewActionBar.tsx
```diff
@@ -326,6 +405,91 @@ export function FilterRadioButton(props: FilterRadioButtonProps) {
...
+        aria-invalid={globalTerm.invalid}
+        id="FilterTermTextInput"
+        InputProps={inputProps}
+        onChange={onChange}
+        placeholder="Filter logs by string"
+        value={filterTerm}
+        variant="outlined"
+      />
+      <SrOnly component="label" htmlFor="FilterTermTextInput">
+        Filter resource logs by string
```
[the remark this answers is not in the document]
lizzthabet: Ah ya! I will add that to this label when I add support for regexp.

web/src/logfilters.ts
```diff
@@ -61,9 +86,23 @@ export function filterSetFromLocation(l: Location): FilterSet {
       break
   }
 
+  const sourceTerm = params.get("term")
+  if (sourceTerm) {
+    filters.term = {
+      sourceTerm,
+      filterTerm: parseFilterTerm(sourceTerm),
+      invalid: false,
```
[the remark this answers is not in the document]
lizzthabet: Yeah, I was thinking that if a user enters a regexp that we can't parse, we could set `invalid` to `true`. (This pr is a little funny because it's only parsing strings and we technically won't hit an error state just yet.) The design doesn't include an error state, but at the minimum, I wanted to pass that error state to an ARIA property on the term field.

I like the state pattern you outlined and will update my code. 🔧

web/src/OverviewActionBar.tsx
```diff
@@ -326,6 +405,91 @@ export function FilterRadioButton(props: FilterRadioButtonProps) {
   )
 }
 
+function FilterTermField(props: { filterSet: FilterSet }) {
+  const globalTerm = props.filterSet.term
```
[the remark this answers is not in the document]
lizzthabet: It's referring to the term object on the global state, rather than the local value of the input field. I had to separate the two to implement debouncing, but it does create a bit of a mismatched state (like using the `invalid` that from the global term, while the local input value may be different).

On a somewhat related note, I don't think the `debounce` function that comes with material-ui is working like I think it should... let me dig into that first.

web/src/SrOnly.tsx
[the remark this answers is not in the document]
lizzthabet: Ooohh this is magic. 👀 I am going to have to look at this more. Could you point me to where you grabbed it from?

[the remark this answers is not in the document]
landism: I pretty much just went to the [definition of `Typography`](https://github.com/mui-org/material-ui/blob/next/packages/material-ui/src/Typography/Typography.d.ts#L91) to get:
```
declare const Typography: OverridableComponent<TypographyTypeMap>
```
and then the [definition of `OverridableComponent`](https://github.com/mui-org/material-ui/blob/next/packages/material-ui/src/OverridableComponent.d.ts#L10-L21) to get:
```
export interface OverridableComponent<M extends OverridableTypeMap> {
  <C extends React.ElementType>(
    props: {
      /**
       * The component used for the root node.
       * Either a string to use a HTML element or a component.
       */
      component: C;
    } & OverrideProps<M, C>
  ): JSX.Element;
  (props: DefaultComponentProps<M>): JSX.Element;
}
```
and then plugged `TypographyTypeMap` from the first in for `M` in the second and fiddled with stuff a bit until it worked.

web/src/logfilters.ts
```diff
@@ -23,9 +23,37 @@ export enum FilterSource {
...
+
+export const EMPTY_TERM = ""
+export const EMPTY_FILTER_TERM: FilterTerm = {
+  input: EMPTY_TERM,
+  state: TermState.Empty,
+}
+
+export function parseFilterTerm(term: string): RegExp {
+  // Filter terms are case-insensitive and can match multiple instances
+  return new RegExp(term, "gi")
```
lizzthabet, line 56: (Commenting to document a discussion @landism and I had yesterday)

My plan was to transform a user's input string into a regex. That input can either be treated as a string literal or as a regex and passed to the `RegExp` constructor, but downstream consumers don't need to worry about what kind of filter term a user enters; they'll have the same interface to filter text by.

For example:
*  the user input `fail(ed)?` would be `new RegExp("fail\\(ed\\)\\?")` and match only on strings that include that exact string
*  while the user input `/fail(ed)?/` would be `new RegExp("fail(ed)?")` and match strings that include `fail` and `failed`

I had incorrectly assumed based on incomplete testing on my part that I could pass this input directly to the `RegExp` constructor, and it would treat the input as a string literal unless surrounded by `/`. That's not the case. 😅 Any regex-specific characters need to be escaped. I'm going to implement this and add tests around these cases.

web/src/OverviewActionBar.stories.tsx
```diff
@@ -43,7 +51,11 @@ export default {
   },
 }
 
-let defaultFilter = { source: FilterSource.all, level: FilterLevel.all }
+let defaultFilter: FilterSet = {
+  source: FilterSource.all,
+  level: FilterLevel.all,
+  term: EMPTY_FILTER_TERM,
+}
```
lizzthabet, line 58: @nicks, lmk if you'd like to see any other storybook changes made! The filter term input field is functional, though we don't have a view where the filters change what logs display.

web/src/OverviewActionBar.tsx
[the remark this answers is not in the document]
landism: I'm not sure about calling that state global, either.

Do you like any of these options?
1. change props to `props: { initialValue: string }` and change the caller to pass in filterSet.term.input (which also
2. just get rid of this line and pass `props.filterSet.term` to `useState` directly, avoiding the naming problem
3. rename `globalTerm` to `termFromUrl` (though that feels like then this function knows too much about what its caller is doing - but that's already a problem with the comment, but I think that'd let you eliminate the comment and the introduction of the term "global")

web/src/OverviewActionBar.tsx
```diff
@@ -326,6 +411,94 @@ export function FilterRadioButton(props: FilterRadioButtonProps) {
...
+      <InputAdornment position="start" disablePointerEvents={true}>
+        <FilterSvg fill={Color.grayDark} role="presentation" />
+      </InputAdornment>
+    ),
+  }
+
+  // If there's a search term, add a button to clear that term
+  if (filterTerm) {
+    const onClearClick = () => {
+      // Clear the input field's value
```
landism, line 441: I'd be tempted to replace this and the onChange function to a call to something like this:
```
  const setTerm = (term: string) => {
    setFilterTerm(term)
    const search = createLogSearch(location.search, { term })
    if (term === EMPTY_TERM) {
      history.push({
        pathname: location.pathname,
        search: search.toString(),
      })
    } else {
      /**
       * Note: debouncing allows us to wait to execute log filtration until a set
       * amount of time has passed without the filter term changing. To implement
       * debouncing, it's necessary to separate the term field's value from the url
       * search params, otherwise the field that a user types in doesn't update.
       * The term field updates without any debouncing, while the url search params
       * (which actually triggers log filtering) updates with the debounce delay.
       */
      debounceFilterLogs(history, search.toString())
    }
  }
```
just since there are a number of things going on and I found myself wondering if they both did all the same things

web/src/OverviewActionBar.tsx
```diff
@@ -224,6 +245,43 @@ export let ButtonRightPill = styled(ButtonRoot)`
   border-radius: 0 4px 4px 0;
 `
 
+const FilterTermTextField = styled(TextField)`
```
landism, line 248: the right end of the placeholder text is getting cut off in my browser
image

web/src/OverviewActionBar.tsx
```diff
@@ -326,6 +411,94 @@ export function FilterRadioButton(props: FilterRadioButtonProps) {
   )
 }
 
+export const FILTER_INPUT_DEBOUNCE = 500 // in ms
+export const FILTER_FIELD_ID = "FilterTermTextInput"
+
+const debounceFilterLogs = debounce((history: History, search: string) => {
+  history.push({ pathname: location.pathname, search })
```
landism, line 418: Apparently when this happens the textfield loses focus, which seems like it'd be really annoying! If you pause typing for half a second you have to go click in the textfield again to continue typing.

I haven't looked into this too much and don't have much idea what the fix will look like.

web/src/OverviewActionBar.tsx
landism, line 248: When I start entering text, the width of the box changes a fair amount:
image
image
This would be nice to fix but is probably safe to ignore.

web/src/OverviewActionBar.tsx
```diff
@@ -224,6 +245,43 @@ export let ButtonRightPill = styled(ButtonRoot)`
...
 `
 
+const FilterTermTextField = styled(TextField)`
+  & .MuiOutlinedInput-root {
+    border-radius: ${SizeUnit(0.125)};
+    border: 1px solid ${Color.grayLighter};
+    background-color: ${Color.gray};
+    transition: border-color ${AnimDuration.default} ease;
+
+    &:hover {
```
landism, line 255: The border colors here are inconsistent with the resource name filter in the sidebar. I don't know if one of us failed to follow the design or the design didn't specify or the design is inconsistent. Would you mind tracking that down? (We should probably make a component that standardizes these so that we don't have to think about it, but that doesn't need to be in this PR)

landism (commented):
Looking pretty good!

hyu (approved):
Nice! It took me a little while to read through this code, and I think the PR would have been a little easier to understand if it was somehow split into more PRs. (e.g., separate PR for RegEx functionality / debouncing). I appreciate your comments, like those on debouncing, since that did help me understand the code better. 

I added a note re: a fix for the width change on the input that Matt also noticed.

web/src/OverviewActionBar.tsx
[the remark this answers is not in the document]
lizzthabet: Yes! It absolutely makes sense to change this to `initialTerm` and just pass that down as props. Thanks for that suggestion.

web/src/OverviewActionBar.tsx
lizzthabet: Nick fixed this issue with #4641. 🎉

web/src/OverviewActionBar.tsx
lizzthabet: Yeaaaah, I noticed this too. The resource filter field behaves the same, so I think it's okay to leave as-is for now.

web/src/OverviewActionBar.tsx
lizzthabet: Surbhi's design has this input styled like the other filter buttons (which have less-rounded corners and blue borders), rather than the resource filter field. Let me know if that doesn't resolve your question!

web/src/OverviewActionBar.tsx
```diff
@@ -224,6 +245,43 @@ export let ButtonRightPill = styled(ButtonRoot)`
   border-radius: 0 4px 4px 0;
 `
 
+const FilterTermTextField = styled(TextField)`
+  & .MuiOutlinedInput-root {
+    border-radius: ${SizeUnit(0.125)};
```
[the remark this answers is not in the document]
lizzthabet: Sweet! Thank you.

web/src/OverviewActionBar.tsx
lizzthabet: That's fair! I've combined the logic like you've outlined and removed the debouncing delay when there's an empty term that will be the new state.

web/src/OverviewLogPane.tsx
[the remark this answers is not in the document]
lizzthabet: I'm going to leave this as-is for now, but we can revisit this functionality if/when we get user feedback.

nicks (approved):

PR #4611, opened 2021-06-04
websocket: ignore empty loglists

### LINKED ISSUE #4604: Sticky loading popover in v0.20.5
### Expected Behavior
The loading popover screen in the Web UI should go away after connecting to the local tilt daemon.

### Current Behavior
`Loading...` appears until most (all?) of the containers are booted.

#### `tilt doctor` Output

```
$ tilt doctor
[erik/exclude-health+][~/dev/reverb]❥  tilt doctor
Tilt: v0.20.5, built 2021-05-28
System: darwin-amd64
---
Docker
- Host: [default]
- Version: 1.41
- Builder: 2
---
Kubernetes
- Env: kind-0.6+
- Context: kind-kind
- Cluster Name: kind-kind
- Namespace: default
- Container Runtime: read-failure
- Version: Error: Get "https://127.0.0.1:60700/version?timeout=32s": dial tcp 127.0.0.1:60700: connect: connection refused
---

---
Analytics Settings
- User Mode: opt-out
- Machine: 9dba14d67dd4c502e594a05aa1a0e5ec
- Repo: HqkW338Dth3PB3xaNOmm8Q==
```

landism:
Fixes #4604.

### Problem
A race condition can cause the HUD to wipe out its state and go back to "Loading..."
My understanding of how this happens is:
1. Store writes a LogEvent to the LogStore.
2. Store notifies WebsocketSubscriber with a ChangeSummary w/ Log: true for the LogEvent from (1)
3. Store writes a second LogEvent to the LogStore
4. WebsocketSubscriber processes the ChangeSummary from (2), asks LogStore for the LogList, gets the messages from both (1) and (3), and updates its checkpoint to indicate it's fully caught up.
5. Store notifies WebsocketSubscriber with a ChangeSummary w/ Log: true for the LogEvent from (2)
6. WebsocketSubscriber processes the ChangeSummary from (5), asks LogStore for the LogList since ws's checkpoint. Since there are no new log messages, LogStore [returns -1, -1](https://github.com/tilt-dev/tilt/blob/31f52e847be1583418893ac0e8b04f662510893a/pkg/model/logstore/logstore.go#L522-L525). WebsocketSubscriber then [sets its own fromCheckpoint to -1](https://github.com/tilt-dev/tilt/blob/db7bebb023d6e9d11141d159e71fb3e10a7c5abf/internal/hud/server/websocket.go#L210-L212)
7. The next time a log message, comes through, WebsocketSubscriber asks for logs w/ a checkpoint of -1, which gets [bounded to 0](https://github.com/tilt-dev/tilt/blob/31f52e847be1583418893ac0e8b04f662510893a/pkg/model/logstore/logstore.go#L184-L193).
8. WebsocketSubscriber sends that LogList w/ fromCheckpoint=0 to the frontend, which assumes the incoming update is a full update w/ uisession and uiresources, and [wipes out its previous uisession and uiresources](https://github.com/tilt-dev/tilt/blob/c7a500bbeae4807e796b9dd7064f0112bc59b3ec/web/src/HUD.tsx#L373-L382)
9. Since uisession.tiltfileKey is unset, [the HUD shows "Loading..."](https://github.com/tilt-dev/tilt/blob/c7a500bbeae4807e796b9dd7064f0112bc59b3ec/web/src/HUD.tsx#L217-L219)

### Solution
If logstore.ToLogList returns [-1, -1), just return immediately.

### Notes
1. I'm tempted to instead change the "no new logs to send down" contract to return `{fromCheckpoint: fromCheckpoint, toCheckpoint: fromCheckpoint}` instead of -1, -1. That seems like it'd be harder for a caller to screw up. I ended up not going that route when realizing that contract made it into [the proto docs](https://github.com/tilt-dev/tilt/blob/ba1de77456a46ff72f5de798aaac092f47f481e7/pkg/webview/log.proto#L56-L57).
2. Inferring `isRefresh` from `!logListUpdate?.fromCheckpoint` still feels error-prone. AFAICT, we only ever send a full view on WS initialization, while `fromCheckpoint` could could technically be 0 in other cases (like, a bunch of logs fill the buffer and truncate the existing checkpoint out of the buffer). It might be worth formalizing the notion of a refresh.

### DIFF
--- a/internal/hud/server/websocket.go
+++ b/internal/hud/server/websocket.go
@@ -132,6 +132,11 @@ func (ws *WebsocketSubscriber) OnChange(ctx context.Context, s store.RStore, sum
 		return // Not much we can do on error right now.
 	}
 
+	// [-1,-1) means there are no logs
+	if view.LogList.ToCheckpoint == -1 && view.LogList.FromCheckpoint == -1 {
+		return
+	}
+
 	ws.sendView(ctx, view)
 
 	// A simple throttle -- don't call ws.OnChange too many times in quick succession,
--- a/internal/hud/server/websocket_test.go
+++ b/internal/hud/server/websocket_test.go
@@ -12,9 +12,9 @@ import (
 	"github.com/stretchr/testify/require"
 
 	"github.com/tilt-dev/tilt/internal/controllers/fake"
-	"github.com/tilt-dev/tilt/internal/testutils"
-
 	"github.com/tilt-dev/tilt/internal/store"
+	"github.com/tilt-dev/tilt/internal/testutils"
+	"github.com/tilt-dev/tilt/pkg/logger"
 )
 
 func TestWebsocketCloseOnReadErr(t *testing.T) {
@@ -39,10 +39,10 @@ func TestWebsocketCloseOnReadErr(t *testing.T) {
 
 	conn.AssertNextWriteMsg(t).Ack()
 
-	st.NotifySubscribers(ctx, store.ChangeSummary{Log: true})
+	writeLogAndNotify(ctx, st)
 	conn.AssertNextWriteMsg(t).Ack()
 
-	st.NotifySubscribers(ctx, store.ChangeSummary{Log: true})
+	writeLogAndNotify(ctx, st)
 	conn.AssertNextWriteMsg(t).Ack()
 
 	conn.readCh <- readerOrErr{err: fmt.Errorf("read error")}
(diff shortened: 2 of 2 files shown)


nicks (approved):
I wonder if it would make more sense to have a more explicit "isIncremental" / "isFull" websocket message, which would also fix this issue (since then you don't have "empty" logs treated as a full reload)

landism:
> I wonder if it would make more sense to have a more explicit "isIncremental" / "isFull" websocket message, which would also fix this issue (since then you don't have "empty" logs treated as a full reload)

Yep, that's what I was attempting to communicate in note (2) above! That more directly addresses the problem. I mostly went this route since I wasn't super comfortable with the logstore code and wasn't certain of the problem being solved by the refresh block.

I'll merge this change and put up another PR with that change.

PR #4613, opened 2021-06-07
reconcile api objects instead of state objects [ch12124]

maiamcc:
Hello @milas,

Please review the following commits I made in branch maiamcc/reconcile-pfs:

ec728c3cc09bcbc400a132c59b17d12fbf0b8785 (2021-06-07 11:30:24 -0400)
reconcile in response to API changes and update tests

66be159d7e4bdf31b58d1924fd4c10748938ff18 (2021-06-04 19:34:47 -0400)
preserve old behavior

c133e9696923a6b25463f95fc652c29080b6fcbd (2021-06-04 19:24:28 -0400)
wip


internal/controllers/core/portforward/reconciler.go
```diff
@@ -16,88 +21,75 @@ import (
...
 	// map of PortForward object name --> running forward(s)
-	activeForwards map[string]portForwardEntry
+	activeForwards map[types.NamespacedName]portForwardEntry
 }
 
-func NewReconciler(kClient k8s.Client) *Reconciler {
+var _ store.TearDowner = &Reconciler{}
+var _ reconcile.Reconciler = &Reconciler{}
+
+func NewReconciler(store store.RStore, kClient k8s.Client, ctrlClient ctrlclient.Client) *Reconciler {
```
milas, line 35: You won't be able to take `ctrlclient.Client` as a constructor param, you have to let it be "injected" via `SetClient()`. (Your `Reconcile()` method won't be called _before_ that though, so you don't have to worry about it being nil.)

There's a weird chicken-and-egg party going on between the reconcilers + the controller manager (manager creates client but not until it's actually initialized/running which is _after_ the clients get constructed), see #4250 for how we ended up here.

internal/controllers/core/portforward/reconciler.go
```diff
@@ -16,88 +21,75 @@ import (
...
-					desired.ObjectMeta.Annotations[v1alpha1.AnnotationManifest] {
-				// Nothing has changed, nothing to do
-				continue
-			}
-
-			// There's been a change to the spec for this PortForward, so tear down the old version
-			toShutdown = r.addToShutdown(toShutdown, existing)
+	if active, ok := r.activeForwards[name]; ok {
+		if equality.Semantic.DeepEqual(active.Spec, pf.Spec) &&
+			equality.Semantic.DeepEqual(active.ObjectMeta.Annotations, pf.ObjectMeta.Annotations) {
```
milas, line 70: FWIW I think comparing spec + _just_ the `v1alpha1.AnnotationManifest` annotation (rather than all) from the old code is a bit less error-prone in the event that some{one,thing} starts manipulating annotations across resources.

internal/controllers/core/portforward/reconciler_test.go
```diff
@@ -32,14 +37,21 @@ func TestCreatePortForward(t *testing.T) {
...
-
-	f.onChange()
+	pf := f.makeSimplePF(pfFooName, 8000, 8080)
+	f.Create(pf)
+
+	// gonna want to test that
+	//   a. the PFs are started as expected (r.activeForwards etc.) and
+	//   b. the status is reflected in the API
+	time.Sleep(time.Second)
+	f.requireState(pfFooName, func(pf *PortForward) bool {
```
milas, line 50: I think the `time.Sleep()` can go away?

Also, this needs a real assertion 😬

internal/controllers/core/portforward/reconciler_test.go
```diff
@@ -215,57 +218,81 @@ func TestMultipleForwardsMultiplePods(t *testing.T) {
...
+		kCli:              kCli,
+		ctrlCli:           ctrlCli,
+		r:                 r,
+		out:               out,
 	}
 }
 
-func (f *pfrFixture) onChange() {
-	f.r.OnChange(f.ctx, f.st, store.LegacyChangeSummary())
+func (f *pfrFixture) wait() {
```
milas, line 263: TBH I think the tests are likely to suffer from timing failures from this. I think it might be better to replace this method with something like `requirePfStarted(name string)` which uses `requireState()` to check that `!pf.Status.StartedAt.IsZero()`; i.e. use it as control flow/synchronization and then do normal assertions after.

milas (changes requested):

internal/controllers/core/portforward/reconciler.go
maiamcc: good call, will do

internal/controllers/core/portforward/reconciler_test.go
maiamcc: lol that was supposed to go away 😅

internal/controllers/core/portforward/reconciler_test.go
maiamcc: hm yeah given that i haven't yet implemented status yet, i'm just doing "you can get the port forward back from the API server" which is maybe not THE most robust but better than the existing, wdyt?

maiamcc:
@milas this is ready for you when you get a chance

milas (approved):
🌮

PR #4620, opened 2021-06-08
deps: upgrade to k8s 1.21.1

milas:
Required Changes:
* Refactor deferred client init to not use `ClientBuilder`
  (see https://github.com/kubernetes-sigs/controller-runtime/pull/1409)
* Add stub `RemoveStreams` method on `fakePortForward`

Mechanical Changes:
* Upgrade K8s dependencies to v0.21.1 (aka 1.21.1)
* Upgrade controller-runtime to 0.9.0 (required for K8s v0.21.1 deps)
* Upgrade tilt-apiserver to v0.4.0 (requires for K8s v0.21.1 deps)
* Remove `moby/term` replace (bug was fixed upstream and older
  version no longer compiles)


go.mod
```diff
@@ -128,9 +127,5 @@ replace (
 	go.opencensus.io => github.com/tilt-dev/opencensus-go v0.22.5-0.20200904175236-275b1754f353
 	golang.org/x/crypto => golang.org/x/crypto v0.0.0-20191206172530-e9b2fee46413
 
-	// Workaround for:
-	// https://github.com/moby/term/issues/15
-	golang.org/x/sys => golang.org/x/sys v0.0.0-20200826173525-f9321e4c35a6
```
milas, line 133: One of the updated dependencies was using a new function that wasn't available. The Darwin compilation issue in `moby/term` was fixed and the new version compiles on macOS again. I didn't see any odddities in terminal rendering using the TUI either.

internal/controllers/manager.go
```diff
@@ -85,14 +93,17 @@ func (m *TiltServerControllerManager) SetUp(ctx context.Context, st store.RStore
...
+		ClientDisableCacheFor:   m.uncachedObjects,
 		Logger:                  logr,
 		GracefulShutdownTimeout: &timeout,
 	})
 	if err != nil {
 		return fmt.Errorf("unable to create controller manager: %v", err)
 	}
 
+	// provide the deferred client with the real client now that it has been initialized
+	m.deferredClient.initialize(mgr.GetClient())
```
milas, line 105: `ClientBuilder` is gone and `ctrl.NewManager` can handle uncached objects, so this just explicitly initializes the deferred client after setting that up. Uncached objects are injected via `ProvideUncachedObjects` (currently nil for release, except in tests where we explicitly call this constructor and pass Filewatch)

nicks (approved):

PR #4621, opened 2021-06-08
core/portforward: update port forward status [ch11825]

maiamcc:
Hello @milas,

no code to review yet, but stubbed out where i think port forward status updates ought to go, does this seem right/am i missing anything?

### DIFF
--- a/internal/controllers/core/portforward/reconciler.go
+++ b/internal/controllers/core/portforward/reconciler.go
@@ -2,6 +2,7 @@ package portforward
 
 import (
 	"context"
+	"fmt"
 	"time"
 
 	"k8s.io/apimachinery/pkg/api/equality"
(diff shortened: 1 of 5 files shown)


milas (commented):
Makes sense, though I think there might need to be a slight refactor around `ForwardPorts()` and the unexported `forward()` because they don't actually write errors to `ReadyChan` [can't comment on them directly because GitHub].

Right now, those return `error` and actually only use `ReadyChan` to close it [never sending data on it] to signal being ready. Since they never return a non-nil error after the point of "ready", I feel like they should have no return type and write everything to `ReadyChan`. (Alternatively, make it something like `StatusChan()` and send over a "ready" message or an "error" message.)

milas:
See #4658 for final implementation

Issue #3708, opened 2020-08-16
--update-mode=exec/container inconsistent file permissions


leoluk:
#### Problem

We use a Dockerfile that runs as non-root (i.e. `USER 1000`).

With `--update-mode=container`, synced files are - presumably - created via `docker exec` as root and can't be read by the application user because they're owned by root.

With `--update-mode=exec`, k8s defaults to the container user and everything works as expected.

See also: https://github.com/tilt-dev/tilt/issues/3060

#### Proposed Solution

Always use the container user to sync files

maiamcc:
Thanks for the report @leoluk! Let me dig into how this looks in our code and get back to you.

maiamcc:
Am I right in thinking that you set `USER 1000` early in Dockerfile execution, such that any files you copy into your image are owned by user 1000? Your proposed solution works great here, but conflicts with the needs of #3060 where a user is creating all the files in their image as root, and only switching to a non-root user to execute the entrypoint: in their case, syncing files as the container user causes permission errors as Tilt attempts to overwrite a file owned by root as a non-root user.

I agree that the different file permissions behavior between the two update modes isn't great :-/ but unfortunately a blanket solution of either "run all syncs/runs as root," OR "run all syncs and runs as the container user" is going to fail for some set of users.

I might instead propose a Tiltfile-wide setting a la `update_settings(use_container_user=True)` to tell Tilt to run live updates (syncs and runs) with the container user, if present--would that work for your use case?

Unfortunately, I'm not sure we'll be able to prioritize this feature in the near term; does the workaround of running with `--update-mode=exec` work for you for the time being?

leoluk:
> Am I right in thinking that you set `USER 1000` early in Dockerfile execution, such that any files you copy into your image are owned by user 1000? 

> I might instead propose a Tiltfile-wide setting a la `update_settings(use_container_user=True)` to tell Tilt to run live updates (syncs and runs) with the container user, if present--would that work for your use case?

Yes, the files in the container are owned by the unprivileged user.

Agreed that it should be configurable.

> Unfortunately, I'm not sure we'll be able to prioritize this feature in the near term; does the workaround of running with `--update-mode=exec` work for you for the time being?

Yes, the workaround works just fine.

mrsimo:
[...]

PR #4623, opened 2021-06-08
containerupdate: use the exec API instead of the cp API for copying files

nicks:
Hello @maiamcc,

Please review the following commits I made in branch nicks/ch8967:

c8c15dcbd00827dcda42b3a6089a087c88e5a7c2 (2021-06-08 17:15:23 -0400)
containerupdate: use the exec API instead of the cp API for copying files
Fixes https://github.com/tilt-dev/tilt/issues/3708

### DIFF
--- a/internal/containerupdate/docker_container_updater.go
+++ b/internal/containerupdate/docker_container_updater.go
@@ -40,16 +40,22 @@ func (cu *DockerUpdater) UpdateContainer(ctx context.Context, cInfo store.Contai
 		return errors.Wrap(err, "rmPathsFromContainer")
 	}
 
-	// TODO(maia): catch errors -- CopyToContainer doesn't return errors if e.g. it
-	// fails to write a file b/c of permissions =(
-	err = cu.dCli.CopyToContainerRoot(ctx, cInfo.ContainerID.String(), archiveToCopy)
+	// Use `tar` to unpack the files into the container.
+	//
+	// Although docker has a copy API, it's buggy and not well-maintained
+	// (whereas the Exec API is part of the CRI and much more battle-tested).
+	// Discussion:
+	// https://github.com/tilt-dev/tilt/issues/3708
+	err = cu.dCli.ExecInContainer(ctx, cInfo.ContainerID, model.Cmd{
+		Argv: tarArgv(),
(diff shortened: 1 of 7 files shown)


internal/docker/fake_client.go
```diff
@@ -203,7 +203,14 @@ func (c *FakeClient) ContainerRestartNoWait(ctx context.Context, containerID str
 	return nil
 }
 
-func (c *FakeClient) ExecInContainer(ctx context.Context, cID container.ID, cmd model.Cmd, out io.Writer) error {
+func (c *FakeClient) ExecInContainer(ctx context.Context, cID container.ID, cmd model.Cmd, in io.Reader, out io.Writer) error {
+	if cmd.Argv[0] == "tar" {
```
maiamcc, line 207: nice!

internal/containerupdate/docker_container_updater.go
```diff
@@ -40,16 +40,16 @@ func (cu *DockerUpdater) UpdateContainer(ctx context.Context, cInfo store.Contai
 		return errors.Wrap(err, "rmPathsFromContainer")
 	}
 
-	// TODO(maia): catch errors -- CopyToContainer doesn't return errors if e.g. it
-	// fails to write a file b/c of permissions =(
-	err = cu.dCli.CopyToContainerRoot(ctx, cInfo.ContainerID.String(), archiveToCopy)
+	err = cu.dCli.ExecInContainer(ctx, cInfo.ContainerID, model.Cmd{
+		Argv: []string{"tar", "-C", "/", "-x", "-f", "-"},
```
maiamcc, line 44: this is the same invocation as in the exec updater, right? nbd but might be smart to make this a shared arg list so that we force the two invocations to stay in sync.

maiamcc (approved):
good catch!

internal/containerupdate/docker_container_updater.go
maiamcc: also probably worth a comment why we're using exec+tar instead of the docker client's builtin Copy functionality

nicks: added!

PR #4626, opened 2021-06-09
k8s: wait between delete and create in pod replacement

### LINKED ISSUE #4171: Kubernetes deployer does not wait for pod delete to finish before create
### Expected Behavior
* If a pod is being replaced, and tilt does a Delete + Create, it should wait for the Delete to finalize before attempting the Create

### Current Behavior
* Tilt does not wait for delete so can encounter errors when running the create:
```
Build Failed: kubernetes create: object is being deleted: pods "hi-pod" already exists
```

### Steps to Reproduce
1. Create K8s YML with a single pod:
```
apiVersion: v1
kind: Pod
metadata:
  name: site
spec:
  containers:
    - name: site
      image: nginx
      imagePullPolicy: IfNotPresent
```
2. Create a `Tiltfile` that uses this:
```
k8s_yaml('site.yaml')
k8s_resource('site')
```
3. Run `tilt up`
4. Make a change to a file or hit reload in Tilt UI

### Context
Most users won't hit this because they're using `Deployment`s so the pods have unique names (i.e. old ones can still be terminating while new ones are coming up)

#### About Your Use Case
This came from a user report by Mehran prs in Slack

lucieleblanc:
Fixes #4171.


milas (approved):
🎉 🎉 🎉 🎉

PR #4628, opened 2021-06-09
api: hide generated code from diffs

landism:
I finally got tired of scrolling through 30 pages of pb.go in `git diff master`.


nicks (approved):
woooo

PR #4634, opened 2021-06-10
web: add UIButton to view, read updates from websocket

landism:


landism:
Or maybe it'd be better to have the web ui read the button objects from the api directly and not do the whole fake reconciler thing?

nicks (approved):
LGTM

ya, I think that's right. eventually, I'd like to have the webui use the autogenerated REST API. Though we're still going to need the View proto for snapshots.

Issue #4549, opened 2021-05-12
working directory argument for local/local_resource


nicks:
Most modern subprocess APIs have a way to set the working directory when you invoke the process. Examples:
Docker's workdir: https://docs.docker.com/engine/reference/builder/#workdir
Go's Cmd.Dir https://golang.org/pkg/os/exec/#Cmd
Node's CWD https://nodejs.org/api/child_process.html#child_process_child_process_exec_command_options_callback

Tilt's local/local_resource should also have a way to set the working directory

The workaround is to use 'cd' in the script

nicks:
now that Cmd is a first-class object (https://api.tilt.dev/core/cmd-v1alpha1.html), i wonder if we should just allow local() to accept a cmd object as well as a string

PR #4644, opened 2021-06-14
tiltfile: add working directory option to local and local_resource

lucieleblanc:
Adds a working directory option as specified in #4549.

Documentation change in [tilt.build/#872](https://github.com/tilt-dev/tilt.build/pull/872).

Given the following directory structure: 
```
.
├── Tiltfile
├── site.yaml
├── some_dir
│   └── test-greeting.txt
└── test-greeting.txt
```
And the contents of `Tiltfile`:
```
k8s_yaml('site.yaml')
k8s_resource('site')
local_resource('greeting', cmd='cat test-greeting.txt', dir='some_dir')
```
The "greeting" `local_resource` prints the contents of `some_dir/test-greeting.txt`.

Specifying `dir=None` defaults the working directory to the Tiltfile's directory, meaning the "greeting" resource will print the contents of the Tiltfile-level `test-greeting.txt` instead. This is equivalent to omitting the `dir` operation altogether.

The following will produce errors:
```
# local_resource('greeting', cmd='cat test-greeting.txt', dir=['inside'])         # Tiltfile error
# local_resource('greeting', cmd='cat test-greeting.txt', dir=4)                  # Tiltfile error
# local_resource('greeting', cmd='cat test-greeting.txt', dir='nonexistent_dir')  # resource error
```

### DIFF
--- a/internal/tiltfile/local_resource.go
+++ b/internal/tiltfile/local_resource.go
@@ -84,8 +84,8 @@ func (s *tiltfileState) localResource(thread *starlark.Thread, fn *starlark.Buil
 		"env?", &updateEnv,
 		"serve_env?", &serveEnv,
 		"readiness_probe?", &readinessProbe,
-		"cmd_dir?", &updateCmdDirVal,
-		"serve_cmd_dir?", &serveCmdDirVal,
+		"dir?", &updateCmdDirVal,
+		"serve_dir?", &serveCmdDirVal,
 	); err != nil {
 		return nil, err
 	}
--- a/internal/tiltfile/local_resource.go
+++ b/internal/tiltfile/local_resource.go
@@ -44,6 +44,7 @@ func (s *tiltfileState) localResource(thread *starlark.Thread, fn *starlark.Buil
 	var updateEnv, serveEnv value.StringStringMap
 	var triggerMode triggerMode
 	var readinessProbe probe.Probe
+	var updateCmdDirVal, serveCmdDirVal starlark.Value
 
 	deps := value.NewLocalPathListUnpacker(thread)
 
@@ -83,6 +84,8 @@ func (s *tiltfileState) localResource(thread *starlark.Thread, fn *starlark.Buil
 		"env?", &updateEnv,
 		"serve_env?", &serveEnv,
 		"readiness_probe?", &readinessProbe,
+		"cmd_dir?", &updateCmdDirVal,
+		"serve_cmd_dir?", &serveCmdDirVal,
 	); err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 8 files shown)


lucieleblanc:
Given the following directory structure: 
```
.
├── Tiltfile
├── site.yaml
├── some_dir
│   └── test-greeting.txt
└── test-greeting.txt
```
And the contents of `Tiltfile`:
```
k8s_yaml('site.yaml')
k8s_resource('site')
local_resource('greeting', cmd='cat test-greeting.txt', cmd_dir='some_dir')
```
The "greeting" `local_resource` prints the contents of `some_dir/test-greeting.txt`.

Specifying `cmd_dir=None` or omitting the `cmd_dir` operation altogether defaults to the Tiltfile's directory, meaning the "greeting" resource will print the contents of the Tiltfile-level `test-greeting.txt` instead.

The following will produce errors:
```
# local_resource('greeting', cmd='cat test-greeting.txt', cmd_dir=['inside'])         # Tiltfile error
# local_resource('greeting', cmd='cat test-greeting.txt', cmd_dir=4)                  # Tiltfile error
# local_resource('greeting', cmd='cat test-greeting.txt', cmd_dir='nonexistent_dir')  # resource error
```

internal/tiltfile/local_resource.go
```diff
@@ -83,6 +84,8 @@ func (s *tiltfileState) localResource(thread *starlark.Thread, fn *starlark.Buil
 		"env?", &updateEnv,
 		"serve_env?", &serveEnv,
 		"readiness_probe?", &readinessProbe,
+		"cmd_dir?", &updateCmdDirVal,
+		"serve_cmd_dir?", &serveCmdDirVal,
```
milas, line 88: For consistency with `env` + `serve_env`, I think maybe we should name the actual Tiltfile params `dir` + `serve_dir`.

(FWIW The `_bat` ones are sort of different/special because they're actually _variations_ of the `cmd` value itself to use on Windows vs *nix/macOS. Of course, that said, both `env` + `dir` *are* actually part of the command, so arguably what you picked is more precise, buuut I think consistency/conciseness might be better since I don't think they're otherwise super ambiguous.)

milas (commented):
Logic looks good! Small note about the argument naming.

Can you add a few positive unit test cases (i.e. where `dir=` arguments are specified) to `internal/tiltfile/tiltfile_test.go`?

internal/tiltfile/local_resource.go
lucieleblanc: That makes sense -- I'll change it

milas (approved):

PR #4649, opened 2021-06-14
engine: don't block tilt ci for local_resource with auto_init=False

### LINKED ISSUE #4637: Tilt CI waits indefinitely on un-initialised resources
### Expected Behavior
`tilt ci` should wait for all auto-initialised resources to complete or fail, and then exit accordingly.

### Current Behavior
When a `local_resource` is not auto-initialised, but contains a serve command `tilt ci` never exits.

### Steps to Reproduce
1. Put this basic Tiltfile in an empty directory (replace the commands and `serve_cmd` with anything you like).
```
local_resource("echo-app", "echo hello")

local_resource("echo-service",
               "echo hello",
               serve_cmd="yes world",
               auto_init=False)
```
1. Run `tilt ci`
1. It waits indefinitely.
1. Remove `serve_cmd`
1. Run `tilt ci` again
1. It exits as expected.

#### `tilt doctor` Output

```
Tilt: v0.20.4, built 2021-05-24
System: darwin-amd64
---
Docker
- Host: [default]
- Version: 1.41
- Builder: 2
---
Kubernetes
- Env: unknown
- Context: ***
- Cluster Name: ***
- Namespace: default
- Container Runtime: docker
- Version: v1.18.10
```

#### About Your Use Case
We have been using Tilt CI to initialise headless cypress testing in CI, and to manually start cypress UI locally. Since updating (from Tilt 0.17 to 0.20) our CI has been timing out waiting for the Cypress UI to be manually triggered.

milas:
If a `local_resource` has `auto_init=False` and a `serve_cmd`,
the target in `Session` (used for `tilt ci`) would erroneously
be assigned a waiting state (with an unknown reason since it
wasn't _really_ waiting).

Now the waiting fallback is only used if the resource has
`auto_init=True` (default) OR has ever had a build triggered
for any reason (e.g. `TRIGGER_MODE_AUTO` + file change or via
UI). The latter avoids a potential race condition between the
build being invoked and the serve_cmd actually getting launched.

Fixes #4637.

### DIFF
--- a/internal/engine/session/controller_test.go
+++ b/internal/engine/session/controller_test.go
@@ -374,38 +374,78 @@ func TestExitControlCI_JobSuccess(t *testing.T) {
 }
 
 func TestExitControlCI_TriggerMode_Local(t *testing.T) {
+	type tc struct {
+		triggerMode model.TriggerMode
+		serveCmd    bool
+	}
+	var tcs []tc
 	for triggerMode := range model.TriggerModes {
-		t.Run(triggerModeString(triggerMode), func(t *testing.T) {
+		for _, hasServeCmd := range []bool{false, true} {
+			tcs = append(tcs, tc{
+				triggerMode: triggerMode,
+				serveCmd:    hasServeCmd,
+			})
+		}
+	}
+
+	for _, tc := range tcs {
+		name := triggerModeString(tc.triggerMode)
+		if !tc.serveCmd {
+			name += "_EmptyServeCmd"
+		}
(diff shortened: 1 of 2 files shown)


nicks (approved):
we should probably have an integration test for this? this kind of bug seems particularly susceptible to problems where components integrate

milas:
Merging this & opening a new PR with a much more comprehensive `tilt ci` integration test (diff is noisy so want to keep separate)

PR #4654, opened 2021-06-15
store: add a return value to OnChange, so we can add re-queueing logic

nicks:
Hello @milas,

Please review the following commits I made in branch nicks/onchange2:

ee5c3892e71b4a725234854d972cc335320fd423 (2021-06-15 13:44:10 -0400)
store: add a return value to OnChange, so we can add re-queueing logic

### DIFF
--- a/internal/cli/updog.go
+++ b/internal/cli/updog.go
@@ -164,7 +164,8 @@ func (s *updogSubscriber) SetUp(ctx context.Context, _ store.RStore) error {
 	}
 	return nil
 }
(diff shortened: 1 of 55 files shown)


nicks:
I'm not 100% sure we need this, but i think it will give us more peace of mind to be sure that tilt will eventually retry if it hits an apiserver error

nicks:
also note that there's very little logic in this change, it's mainly updating signatures and making the linter happy

milas (approved):

PR #4656, opened 2021-06-16
controllers: factor out an indexer API

nicks:
Hello @milas,

Please review the following commits I made in branch nicks/indexer:

d675d52a6e9747822bfa266a021bc1903f715cc0 (2021-06-15 20:17:55 -0400)
controllers: factor out an indexer API
Now that we're doing dependencies more, I felt bad duplicating this code.


milas (approved):
Nice!

(I do wish that `ctrlclient.Client` didn't have weird, deferred initialization so that we could just read the scheme off of it instead of having to pass it around again, but not much we can do about that.)

PR #4657, opened 2021-06-16
api: fill out PortForwardStatus type

milas:
It's possible to define multiple forwards on a single `PortForward`
spec, so the status conveys each of these independently, as it's
not possible to transactionally set up all forwards at once, meaning
some might be in a failure state while others are running.

It's also possible to let Tilt pick a random free port for the
local port, in which case, the status is also the only way to find
out what port got picked, so having per-forward granularity is
also necessary.


pkg/apis/core/v1alpha1/portforward_types.go
```diff
@@ -125,14 +130,29 @@ func (in *PortForwardList) GetListMeta() *metav1.ListMeta {
 
 // PortForwardStatus defines the observed state of PortForward
 type PortForwardStatus struct {
-	// Time at which we started trying to run the Port Forward (potentially distinct
-	// from the time the Port Forward successfully connected)
-	StartedAt metav1.MicroTime `json:"startedAt,omitempty" protobuf:"bytes,1,opt,name=startedAt"`
```
milas, line 130: I just deleted this field since it wasn't being populated yet, but I did force the protobuf numbering to re-start at 2 just to be safe.

nicks (approved):

PR #4658, opened 2021-06-17
k8s: update PortForwardStatus after initializing forwards

milas:
There are 3 cases where status gets updated for a forward:
 * Forward started successfully (StartedAt set as well as local
   port, which is useful if a randomized one was requested)
 * Forwarder instance can't be created (bad config)
 * Forwarder returns an error after start

To simplify error handling and keep the status update logic
as centralized as possible, the actual initialization changed
slightly to happen in a single method.

There is some very basic logic to avoid spamming the API
server with updates on errors: it will only attempt to do
an update once a second if continuously failing. (Even then,
it will attempt to avoid a no-op update if the error hasn't
actually changed.)

(This is based off Maia's original stub PR: #4621)

### DIFF
--- a/internal/controllers/core/portforward/reconciler.go
+++ b/internal/controllers/core/portforward/reconciler.go
@@ -2,7 +2,13 @@ package portforward
 
 import (
 	"context"
+	"sort"
+	"sync"
 	"time"
+
+	"github.com/tilt-dev/tilt/internal/timecmp"
+	"github.com/tilt-dev/tilt/pkg/apis"
+	"github.com/tilt-dev/tilt/pkg/logger"
 
 	"k8s.io/apimachinery/pkg/api/equality"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
@@ -17,7 +23,6 @@ import (
 
 	"github.com/tilt-dev/tilt/internal/k8s"
 	"github.com/tilt-dev/tilt/internal/store"
-	"github.com/tilt-dev/tilt/pkg/logger"
 )
 
 type Reconciler struct {
@@ -26,7 +31,7 @@ type Reconciler struct {
 	ctrlClient ctrlclient.Client
 
 	// map of PortForward object name --> running forward(s)
-	activeForwards map[types.NamespacedName]portForwardEntry
+	activeForwards map[types.NamespacedName]*portForwardEntry
 }
 
 var _ store.TearDowner = &Reconciler{}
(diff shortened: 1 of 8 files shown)


internal/controllers/core/portforward/reconciler.go
```diff
@@ -128,23 +126,92 @@ func (r *Reconciler) startPortForwardLoop(ctx context.Context, entry portForward
...
+	// the doneCh ensures we don't leak the goroutine if ForwardPorts() errors out early without
+	// ever becoming ready
+	doneCh := make(chan struct{}, 1)
+	go func() {
+		readyCh := pf.ReadyCh()
+		if readyCh == nil {
+			return
+		}
+		select {
+		case <-doneCh:
```
nicks, line 185: should we also select on ctx.Done()?

internal/controllers/core/portforward/reconciler.go
```diff
@@ -104,20 +107,15 @@ func (r *Reconciler) startPortForwardLoop(ctx context.Context, entry portForward
...
+			// If the context was canceled, there's nothing more to do;
+			// we cannot even update the status because we no longer have
+			// a valid context, but that's fine because that means this
+			// PortForward is being deleted.
 			return
 		}
 
-		// Otherwise, repeat the loop, maybe logging the error
-		if err != nil {
-			logger.Get(ctx).Infof("Reconnecting... Error port-forwarding %s (%d -> %d): %v",
```
nicks, line 116: hmm....it seems like we still want this logging to tell the user that we're reconnecting?

right now, it looks like we never print the error message at all....

milas: Yep, good catch. Re-added in 91a5eb6fa7ef9c74acf3bbe318014d29f3a72484.

FWIW I kept the logic the same in that it will log every invocation (that "maybe" in the comment was incorrect), but it might make sense to make this conditional along with API update to avoid flooding the logs (if you pick a privileged port it's...a lot)

internal/controllers/core/portforward/reconciler.go
milas: _Shouldn't_ matter (if context is canceled, `ForwardPorts()` should return, so `doneCh` will get closed) but can't hurt anything and might avoid a nasty surprise in the future if we refactor - added in 83f8bc3a3f40d5a00d4eb3e3630a2d9ee58ca524

nicks (approved):

PR #4660, opened 2021-06-17
web: add Table Overview at "secret" URL /table

hyu:
Implements a Table Overview using the [react-table](https://react-table.tanstack.com/) library.

### 1. On Sub-Components

Although the Star, Trigger, and Trigger Mode components used in this Table could theoretically be shared with other use cases (e.g., SidebarItem), in this PR I created Table-specific versions of these components prefixed with "OverviewTable." 

For these components to work in the table, they needed changes (e.g., taking the useContext hook as a prop rather than using it directly in the component, to avoid using a hook within a hook). And the original components had baked-in assumptions specific to Sidebar (e.g., Trigger Mode visibility depends on hover status of Sidebar item parent). 

So, creating separate components was also attempting to add complexity to the original component. I'm not opposed to refactoring these components in some future PR so there can be more code sharing, but I punted on that for now.

### 2. On Trigger Mode
Trigger Mode is visible on all items, rather than only on Tests (which is the case with Sidebar Items). This seemed reasonable to me because we support this in the API, and the Trigger Mode column was a part of Surbhi's mocks. I did not update Sidebar Item accordingly, to contain all changes in this PR to the Table only.

### 3. URL

This table is visible at URL /table. Once this PR is merged in, a separate PR will set the Table View as default in lieu of the Grid.


web/src/OverviewTable.tsx
```diff
@@ -0,0 +1,482 @@
...
+  ${mixinResetButtonStyle};
+  padding-top: ${SizeUnit(0.5)};
+  padding: ${SizeUnit(0.25)};
+
+  svg {
+    fill: ${Color.gray6};
+  }
+`
+
+function columnDefs(): Column<RowValues>[] {
```
nicks, line 163: would it be simpler to make these global components? e.g.,

```
function StarColumn({ row }: CellProps<RowValues>) {
  let ctx = useStarredResources()
  return (
            <TableStarResourceButton
              resourceName={row.values.name}
              analyticsName="ui.web.overviewStarButton"
              ctx={ctx}
            />
          )
        }
}

let starColumnDef = {Header: "Starred", Cell: StarColumn}
```

AFAICT, you've added two layers of indirection: a factory method layer that returns React components, and a React.memo layer that memoizes the factory methods. But the layers cancel each other out, so I'm not sure I understand what they're buying you....though I might be missing something.

web/src/OverviewTableStarResourceButton.tsx
```diff
@@ -0,0 +1,89 @@
...
+
+  if (isStarred) {
+    icon = <ActiveStarIcon />
+    title = "Unstar"
+  } else {
+    icon = <InactiveStarIcon />
+    title = "Star"
+  }
+
+  function onClick(e: any) {
```
nicks, line 71: i wonder if the starred resource context should have a toggleStar() method that you can use directly.

web/src/OverviewTableTriggerButton.tsx
```diff
@@ -0,0 +1,147 @@
...
+  if (!isDisabled) {
+    if (props.hasPendingChanges && isManualTriggerMode) {
+      shouldBeClicked = true
+    } else if (!props.hasBuilt && isManualTriggerIncludingInitial) {
+      shouldBeClicked = true
+    }
+  }
+
+  function triggerUpdate(name: string) {
+    let url = `//${window.location.host}/api/trigger`
```
nicks, line 105: you can shorten this to `/api/trigger`

nicks (commented):
in the future, it would be nice if this was incremental rather than one 1K+ line PR

web/src/OverviewTableStarResourceButton.tsx
hyu: Good point, @nicks!

web/src/OverviewTable.tsx
hyu: Thanks for pointing this out, @nicks. To be honest, I'm a little confused about this as well because my understanding of how and why useMemo is used here is a bit shaky. I ended up with this code structure mostly through following the sample code from React Table's [docs](https://react-table.tanstack.com/docs/quick-start), and adapting it for our needs. The library author writes that memoization is required for “referential equality from render to render,” whereas I had thought of memoization as mostly for optimization to avoid excessive re-rendering of tons of data. And when I didn't have `useMemo` in there, I kept seeing a ton of errors in the console.

If you can see a way to do this more sensibly, I'd like to pair on that. Maybe you can see a way to get around whatever's required by React Table while not getting into too much indirection.

nicks: ya, want to pair on it this afternoon?

web/src/OverviewTable.tsx
```diff
@@ -0,0 +1,482 @@
...
+`
+const ResourceTableData = styled.td`
+  font-family: ${Font.monospace};
+  font-size: ${FontSize.smallest};
+  color: ${Color.gray6};
+  box-sizing: border-box;
+`
+const ResourceTableHeader = styled(ResourceTableData)`
+  color: ${Color.gray7};
+  font-size: ${FontSize.smallester};
```
lizzthabet, line 85: This is a small note (no pun intended 🤦🏻). The headers are a little small for me to read as a user, even on smaller screens. What would you think about bumping this up to `smallest` instead of `smallester`? You may not want to mess with style tweaks at this stage of feedback, so happy to revisit this later!

web/src/OverviewTable.tsx
```diff
@@ -0,0 +1,482 @@
...
+          return <div>{endpoints}</div>
+        },
+      },
+      {
+        Header: "Trigger Mode",
+        accessor: "triggerMode",
+        width: "50px",
+        Cell: ({ row }: CellProps<RowValues>) => {
+          return (
+            <TableTriggerModeToggle
```
lizzthabet, line 340: Should the trigger mode toggle render conditionally? (This may be a lack of knowledge of my end.) When I test locally, I see toggles showing up for all resources, including the Tiltfile, and not all of the toggles successfully toggle when I click on them.

nicks (approved):

web/src/OverviewTable.tsx
nicks: pushed the changes we paired on

lizzthabet (approved):
Looks great to me! I tested locally and just noted a couple design tweaks that are non-blocking. I did have trouble accessing the table contents while navigating with VoiceOver, but I'm not sure if that's user error or something particular with react-table.

hyu:
Thank you so much for the careful review, @lizzthabet! I will create separate PRs to address the issues you found.

PR #4661, opened 2021-06-17
web: add support for filtering logs by regular expression

lizzthabet:
This PR:
*  Fix a bug where navigating to an individual resource from sidebar view would clear the filter term in the url, but not the filter input field
*  Add support for using regular expressions to filter log spans
*  Display an error state with helpful text if parsing regexp errors, using Material UI's `FormHelperText` component

Fyi, the final style implementation does look different than the Figma files. While pairing with Han, we decided that displaying the error message styled as tooltip would be the easiest approach, and Surbhi approved.

Basic functionality:

Final styles:
Screen Shot 2021-06-21 at 12 13 57 PM

Closes ch12047

### DIFF
--- a/web/src/OverviewActionBar.tsx
+++ b/web/src/OverviewActionBar.tsx
@@ -24,7 +24,13 @@ import { ReactComponent as LinkSvg } from "./assets/svg/link.svg"
 import { InstrumentedButton } from "./instrumentedComponents"
 import { displayURL } from "./links"
 import LogActions from "./LogActions"
-import { EMPTY_TERM, FilterLevel, FilterSet, FilterSource } from "./logfilters"
+import {
+  EMPTY_TERM,
+  FilterLevel,
+  FilterSet,
+  FilterSource,
+  TermState,
+} from "./logfilters"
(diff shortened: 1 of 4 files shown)


web/src/OverviewActionBar.tsx
```diff
@@ -251,36 +260,56 @@ export let ButtonRightPill = styled(ButtonRoot)`
...
   }
 
   & .MuiInputBase-input {
-    font-family: ${Font.monospace};
     color: ${Color.gray7};
+    font-family: ${Font.monospace};
     font-size: ${FontSize.small};
   }
+
+  & .MuiFormHelperText-root {
```
nicks, line 299: fyi, i think the & bit here is redundant?

nicks (approved):

web/src/OverviewActionBar.tsx
lizzthabet: Ah yes! I ended up moving these styles to a new selector, but ty for pointing out.

nicks (approved):

PR #4662, opened 2021-06-17
web: optimize log rendering

nicks:
Hello @hyu, @lizzthabet,

Please review the following commits I made in branch nicks/react-optimize2:

c6960421c234c77d200da5ab53df259653560d26 (2021-06-17 15:31:31 -0400)
web: optimize log rendering

### DIFF
--- a/web/src/HUD.tsx
+++ b/web/src/HUD.tsx
@@ -352,7 +352,7 @@ function mergeObjectUpdates<T extends { metadata?: Proto.v1ObjectMeta }>(
 export function mergeAppUpdate<K extends keyof HudState>(
   prevState: Readonly<HudState>,
   stateUpdates: Pick<HudState, K>
-): Pick<HudState, K> {
+): null | Pick<HudState, K> {
   // All fields are optional on a HudState, so it's ok to pretent
   // a Pick<HudState> and a HudState are the same.
   let state = stateUpdates as HudState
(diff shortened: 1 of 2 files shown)


nicks:
I'm hoping this will fix some of the complaints in the user research about the web ui using a lot of CPU when there are lots of logs.

setState(null) will ensure the app doesn't re-render when the only change that came in was new logs

web/src/HUD.tsx
```diff
@@ -413,5 +413,26 @@ export function mergeAppUpdate<K extends keyof HudState>(
...
+  }
+
   return result
 }
+
+function hasChange(result: any, prevState: any): boolean {
+  for (let k in result) {
+    let resultV = result[k] as any
+    let prevV = prevState[k] as any
+    if (resultV != prevV) {
```
lizzthabet, line 433: Does strict equality make a difference here? (Just curious.)

lizzthabet:
That makes sense! Is there any way you'd recommend testing this out locally?

nicks:
The best way to test it manually is to 
1) Install the react profiler (it's a browser extension)
2) run an app with a lot of logs
3) verify that the profiler doesn't see any react re-rendering when the logs come in

web/src/HUD.tsx
nicks: i don't think it will make a difference, but it certainly doesn't hurt

lizzthabet (approved):
Sweet! I'm not sure I see any difference in the profiling, but I think I need to read up on how to understand the profiler output better. 😅

Issue #1113, opened 2019-02-05
Namespace Support


jazzdan:
Tilt should have build in support for setting a Kubernetes namespace that all objects get created in for the duration of the session.

Tiltfile syntax could look like this:

```python
set_namespace("dan-dev")
```

Tilt would remember the namespace that you specify and append it as a command line flag every time it invokes `kubectl`.

drubin:
For now, I am using the following 

```python
def set_namespace(namespace):
  command = 'kubectl config set-context $(kubectl config current-context) --namespace=%s' % (namespace)
  current_context = str(local(command)).rstrip('\n')

def set_context(context, namespace=''):
  command = 'kubectl config use-context %s' % (context)
  current_context = str(local(command)).rstrip('\n')
  if namespace:
    set_namespace(namespace)

set_context('docker-for-desktop')
# set_context('docker-for-desktop', 'dev')
set_namespace('dev')
```

victorwuky:
@jazzdan : Would it make sense for that command `set_namespace`, to create the namespace if it didn’t already exist? See https://kubernetes.slack.com/archives/CESBL84MV/p1589465928081300?thread_ts=1589454049.073200&channel=CESBL84MV&message_ts=1589465928.081300

jazzdan:
@victorwuky yup I think that could make sense.

drubin:
When has a tool that sets a setting every created it?

What would happen if kubens created the namespace for every typo that a user typed?

These should be seperate functions/commands  Linux philosophy here do one thing and one thing well. If people want the namespace they could use a create_namespace

Just for the record I would love to tilt to be able to create namespaces but I would hate it creating namespaces for me when I am trying to set the namespace/context.

PR #4668, opened 2021-06-21
cli: add namespace flag

lucieleblanc:
A very basic default namespace flag, inspired by the change suggested in #1113. 
```
tilt up --namespace=my-namespace
```
Documentation change in [tilt.build #876](https://github.com/tilt-dev/tilt.build/pull/876).

### DIFF
--- a/internal/cli/flags.go
+++ b/internal/cli/flags.go
@@ -14,8 +14,10 @@ import (
 
 var defaultWebHost = "localhost"
 var defaultWebPort = model.DefaultWebPort
+var defaultNamespace = ""
 var webHostFlag = ""
 var webPortFlag = 0
+var namespaceFlag = ""
 
 func readEnvDefaults() error {
 	envPort := os.Getenv("TILT_PORT")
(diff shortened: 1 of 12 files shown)


internal/k8s/client.go
```diff
@@ -41,6 +42,7 @@ import (
 )
 
 type Namespace string
+type NamespaceFlag string
```
milas, line 45: Minor, but I think for clarity it might make sense to call this `NamespaceOverride` (the fact that it comes from a flag when run from CLI is sorta incidental)

internal/cli/flags.go
```diff
@@ -62,8 +64,16 @@ func addDevServerFlags(cmd *cobra.Command) {
 	cmd.Flags().Var(&webModeFlag, "web-mode", "Values: local, prod. Controls whether to use prod assets or a local dev server. (If flag not specified: if Tilt was built from source, it will use a local asset server; otherwise, prod assets.)")
 }
 
+func addNamespaceFlag(cmd *cobra.Command) {
+	cmd.Flags().StringVar(&namespaceFlag, "namespace", defaultNamespace, "Namespace for the resources Tilt brings up.")
```
milas, line 68: This is tricky to convey concisely, but I think it'd be good to mention that it's a default, e.g. "Default namespace for Kubernetes resources (overrides default namespace from active context in kubeconfig)"

milas (approved):
Nice! Left two very minor suggestions - feel free to take or leave 🙃

PR #4670, opened 2021-06-21
store: add backoff logic to subscribers

nicks:
Hello @milas, @landism,

Please review the following commits I made in branch nicks/onchange:

69909e11093c984d6cedf1212148fdc789aedccd (2021-06-21 12:15:25 -0400)
store: add backoff logic to subscribers

### DIFF
--- a/internal/store/subscriber.go
+++ b/internal/store/subscriber.go
@@ -4,7 +4,12 @@ import (
 	"context"
 	"fmt"
 	"sync"
+	"time"
+
+	"github.com/tilt-dev/tilt/pkg/logger"
 )
+
+const MaxBackoff = time.Second * 15
 
 // A subscriber is notified whenever the state changes.
 //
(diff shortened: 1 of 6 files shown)


nicks:
this is the more "kubernetes"-y way to do error handling. Rather than do a lot of work to handle errors correctly and make sure controllers start in the right order, just back off and try again if anything goes wrong.

internal/store/subscriber.go
```diff
@@ -185,10 +191,30 @@ func (e *subscriberEntry) notify(ctx context.Context, store *Store) {
...
+	}
+
+	// Backoff on error
+	backoff := activeChange.LastBackoff * 2
+	if backoff == 0 {
+		backoff = time.Second
+	} else if backoff > MaxBackoff {
+		backoff = MaxBackoff
+	}
+	logger.Get(ctx).Errorf("Problem processing change. Backing off %v. Error: %v", backoff, err)
```
milas, line 206: Should we include the subscriber type here? Also, what do you think about `Debug` if `backoff == 0` (i.e. first time) and then error `Error` after that?

milas (approved):
Omg this will make migrations much easier (and more consistent 😅)

internal/store/subscriber.go
nicks: ooh, i like the Debug idea! done.

i left a TODO for the subscriber type - i think that's a good idea too, but i don't think we necessarily have reliable subscriber type names right now (they'll all just be "Controller")

PR #4672, opened 2021-06-21
web: add analytics information for filter term input

lizzthabet:


web/src/analytics.ts
```diff
@@ -46,5 +51,8 @@ export let navigationToTags = (location: any, action: string): Tags => {
   if (filterSet.source != FilterSource.all) {
     tags.source = filterSet.source
   }
+  if (filterSet.term.state !== TermState.Empty) {
+    tags.term = filterSet.term.input
```
lizzthabet, line 55: Not sure if we actually want the user input or just the term state, which would either be `parsed` or `error`. I'll let other chime in here, since I don't have a good sense of how tags are used.

landism: > Not sure if we actually want the user input

Nope! Including the user input would be against what we describe in our [telemetry faq](https://docs.tilt.dev/telemetry_faq.html#what-kinds-of-data-does-tilt-record-if-i-opt-in). We aim to include no user-generated strings in any of our analytics data.

> just the term state, which would either be parsed or error

I think it's pretty unlikely we make a decision based on that data, but it's also really cheap to add. You might as well.

I think the more interesting bit of data is whether the search is plaintext or regexp - if it turns out we get very few regexp users, we could consider cutting that part of the feature or improving its discoverability.

lizzthabet: Ahh, great point! I can refactor to record that data instead.

lizzthabet: This should be g2g now.

landism (approved):

PR #4677, opened 2021-06-22
engine: write ImageMaps to the apiserver

nicks:
Hello @milas, @landism,

Please review the following commits I made in branch nicks/tiltfile2:

f42326d1a396ed70cd5f0c0b1dfbe3942e8409db (2021-06-22 13:00:23 -0400)
engine: write ImageMaps to the apiserver

861df8800fc08a34baa4f41d93b99bc3f18cbb71 (2021-06-22 12:40:53 -0400)
engine: make the tiltfile loader write to the API server


milas (approved):

internal/engine/configs/api.go
```diff
@@ -98,9 +100,35 @@ func toKubernetesApplyObjects(tlr tiltfile.TiltfileLoadResult) typedObjectSet {
...
+		}
+	}
+	return result
+}
+
 // Fetch all the existing API objects that were generated from the Tiltfile.
 func getExistingAPIObjects(ctx context.Context, client ctrlclient.Client) (objectSet, error) {
 	result := objectSet{}
+
+	// TODO(nick): There's got to be a more generic way to do this.
```
landism, line 131: oy, I take it the dynamic client would be ugly to use here?

landism (approved):

internal/engine/configs/api.go
nicks: ya, the List object interface isn't super helpful here
https://pkg.go.dev/github.com/tilt-dev/tilt-apiserver/pkg/server/builder/resource#ObjectList

PR #4684, opened 2021-06-24
watch: use WalkDir to speed up file listing

milas:
`WalkDir` is new in Go 1.16 and avoids calling `os.Lstat` on
every visited file and directory. In most cases, we don't need
that info, so this will help reduce I/O when listing files,
which can be helpful for particularly big monorepos.

I happened to notice this in `listdir` when looking at Tiltfile
code this week and it's an extremely straightforward change.
Probably won't be noticeable most of the time, but no downside
to it either!


nicks (approved):

nicks:
we probably need to update the extensions ci for go 1.16

PR #4688, opened 2021-06-24
test: share K8s native PodBuilder

milas:
Several of the controllers interact with K8s Pods; the existing
PodBuilder used in manifest tests isn't appropriate as it's meant
to create Pods for manifests.

The PodLogStream reconciler has a good implementation of code to
generate realistic looking K8s pods, so moving it to the controller
`fake` package for use across reconcilers.


milas:
This is just moving the code + exporting it. Only new thing is that it sets a `UID` on the `Pod` (will be necessary for `KubernetesDiscovery` to share this momentarily) and varargs options to handle common other things to set at construction time (namespace, labels)

nicks (approved):

milas:
Just kidding - I was doing something else silly in tests and _don't_ need to share this, so going to avoid doing so for now and close this 🙃

PR #4690, opened 2021-06-24
api: PodLogStream lifecycle from KubernetesDiscovery reconciler

milas:
Create + delete `PodLogStream` objects from `KubernetesDiscovery`
object status during reconcile based on discovered pods. This
eliminates the `runtimelog.PodLogManager` which was a store
subscriber that performed the same task - in fact, the data on
the store it was using was populated _by_ the `KubernetesDiscovery`
reconciler, so this brings it closer to the source and eliminates
a potential source of desync.


internal/controllers/core/kubernetesdiscovery/reconciler.go
```diff
@@ -539,6 +575,95 @@ func (w *Reconciler) handlePodDelete(ctx context.Context, namespace k8s.Namespac
...
+
+		if err := w.createPodLogStream(ctx, kd, pod); err != nil {
+			errs = append(errs, fmt.Errorf("failed to create PodLogStream for Pod %s:%s for KubernetesDiscovery %s: %v",
+				pod.Namespace, pod.Name, kd.Name, err))
+		}
+	}
+
+	for podKey, pls := range plsByPod {
+		if !seenPods[podKey] {
+			if err := w.ctrlClient.Delete(ctx, &pls); err != nil {
```
nicks, line 612: should we ignore notFound errors?

nicks (approved):
:clap:

internal/controllers/core/kubernetesdiscovery/reconciler.go
milas: ah yeah - added in 307f582

PR #4693, opened 2021-06-25
kubernetesapply: reconcile objects

nicks:
Hello @milas,

Please review the following commits I made in branch nicks/ka:

f14dd9da7b50a58628fc2c1ba818316a7aaad883 (2021-06-25 13:28:50 -0400)
kubernetesapply: reconcile objects

### DIFF
--- a/internal/controllers/core/kubernetesapply/reconciler.go
+++ b/internal/controllers/core/kubernetesapply/reconciler.go
@@ -3,10 +3,10 @@ package kubernetesapply
 import (
 	"context"
 	"fmt"
-
-	"sigs.k8s.io/controller-runtime/pkg/builder"
+	"sync"
 
 	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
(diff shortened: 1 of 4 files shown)


internal/controllers/core/kubernetesapply/reconciler.go
```diff
@@ -76,23 +83,160 @@ func (r *Reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
...
 func (r *Reconciler) ForceApply(
+	ctx context.Context,
+	nn types.NamespacedName,
+	spec v1alpha1.KubernetesApplySpec,
+	imageMaps map[types.NamespacedName]*v1alpha1.ImageMap) (v1alpha1.KubernetesApplyStatus, error) {
+	status := r.forceApplyHelper(ctx, spec, imageMaps)
+
+	result := Result{
+		Spec:   spec,
+		Status: status,
```
milas, line 206: I think this usage is actually provably safe right now (status has no reference types + based on code paths we never actually use both the return + the version stored in results), but it might be good to `*status.DeepCopy()` here anyway to avoid a future surprise (or incur the wrath of the race detector)

milas (approved):

internal/controllers/core/kubernetesapply/reconciler.go
nicks: done

PR #4694, opened 2021-06-25
ci: use pinned tilt-extensions-ci image

milas:
Avoid breakage from changes to the extensions CI runner image
by pinning to a specific version (ideally kept in sync with
`tilt-extensions` as much as possible).


milas:
This will fail until we merge tilt-dev/tilt-extensions#186

nicks (approved):

PR #4696, opened 2021-06-28
configs: create filewatches during tiltfile creation

nicks:
Hello @milas,

Please review the following commits I made in branch nicks/fw:

9af8f1d2de8b363c8f9e20e9a0763640295c88cb (2021-06-28 11:55:02 -0400)
configs: create filewatches during tiltfile creation


nicks:
this is mainly taking the code in manifest subscriber and moving it over to tiltfile creation...

nicks:
this should be much simpler and easier to reason about in general because:
1) the tiltfile loader already reconciles against the apiserver, handling garbage collection, etc, and
2) we really only want to change these filewatches based on the tiltfile, nothing else

internal/store/engine_state.go
```diff
@@ -492,7 +491,6 @@ func NewState() *EngineState {
 	}
 
 	ret.Cmds = make(map[string]*Cmd)
-	ret.FileWatches = make(map[types.NamespacedName]*v1alpha1.FileWatch)
```
milas, line 495: It's so nice to see these go away

milas (approved):

internal/store/engine_state.go
nicks: :clap:

Issue #4456, opened 2021-04-21
cmd should delete its process group if the init process terminates


nicks:
Currently, we have a root Cmd type. It's the underlying implementation of local_resource(serve_cmd)

Here's how it currently works:
- Tilt runs the command specified by Cmd, then marks the Cmd "Running"
- Tilt uses Golang's Cmd Wait() to wait until the process is terminated, then marks the Cmd "Terminated"

There's a subtle problem here - Golang's Cmd Wait() semantics are slightly different than Unix's Wait() semantics. In Unix, the wait() syscall just waits for the main process to terminate. In Golang, Go ALSO waits for all the stdout/stderr streams to close.
https://golang.org/pkg/os/exec/#Cmd.Wait

This can lead to some surprising behavior if there are child processes interacting with the stream -- e.g., you "kill" the main process, and Tilt doesn't restart it.

I think we should change Cmd to work more like containers - if you the "root" process terminates for any reason, all child processes die immediately. 
https://lwn.net/Articles/532748/

This would be a breaking change. But I think this would fix a wide variety of weird robustness issues around Cmd when subprocesses are left running and, say, hanging onto a port.

nicks:
Related issue: https://github.com/tilt-dev/tilt/issues/4455

I'm not 100% confident that that bug is caused by that issue (since I'm still unable to reproduce this bug).

But I do think we should move away from relying on the semantics of Golang's cmd Wait(). Even the Go team is considering changing it, because the behavior not what most people expect: https://github.com/golang/go/issues/23019

PR #4697, opened 2021-06-28
cmd: fix flaky grandkid test

landism:
This test failed for me in CI.

I'm not sure a test change is the right path forward, but it's not immediately clear to me how to fix the underlying code in a way that's copacetic with #4456, so I figured I'd throw up a PR to discuss the issue.

(FWIW, I added logging to the `io.Writer` passed as stdout and confirmed its `Write` was being called with `BACKGROUND $!` after `c.Process.Wait()` had returned)

### DIFF
--- a/internal/controllers/core/cmd/execer_unix_test.go
+++ b/internal/controllers/core/cmd/execer_unix_test.go
@@ -27,6 +27,29 @@ echo BACKGROUND $!
 '`)
 	f.waitForStatus(Done)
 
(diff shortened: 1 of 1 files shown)


milas (approved):
This test fix seems reasonable to me - I looked at the Go stdlib sources and I actually don't know if we can truly resolve it in the general case (i.e. make execer block until I/O is consumed)? It looks like it'll only ever close the writers if you call `Cmd::Wait()` 😢

landism:
> I actually don't know if we can truly resolve it in the general case

Yeah, short of bypassing `exec.Cmd` and going straight to `os.StartProcess` and friends.

PR #4701, opened 2021-06-28
k8s: remove our homegrown meta Accessor in favor of apimachinery's

nicks:
Hello @landism, @milas,

Please review the following commits I made in branch nicks/entity:

60b63fcbc1a670905f4653adff5846eb3a4f4248 (2021-06-28 19:00:43 -0400)
k8s: remove our homegrown meta Accessor in favor of apimachinery's

### DIFF
--- a/internal/k8s/entity.go
+++ b/internal/k8s/entity.go
@@ -3,11 +3,10 @@ package k8s
 import (
 	"fmt"
 	"net/url"
-	"reflect"
 	"sort"
 	"strings"
-	"testing"
 
+	"k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/kubernetes/scheme"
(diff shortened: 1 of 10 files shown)


internal/k8s/entity.go
```diff
@@ -118,27 +101,14 @@ func (e K8sEntity) GVK() schema.GroupVersionKind {
...
-	}
-
-	if structured, _ := e.maybeStructuredMeta(); structured != nil {
-		return structured
-	}
-
-	return emptyMeta{}
-}
-
 func (e K8sEntity) maybeUnstructuredMeta() *unstructured.Unstructured {
```
milas, line 112: In theory, should this method (+ `maybeStructuredMeta`) go away? It looks like it's now only* used by `SetUID` - should that also be using `e.Meta().SetUID()`?

\* I do see it also gets used in a CRD test but I think that's easily in-lineable

milas (approved):

internal/k8s/entity.go
nicks: ooh good point. removed!

PR #4712, opened 2021-06-30
docker-compose: do not force recreate on up

milas:
From the Docker Compose[1] docs:
```
--force-recreate    Recreate containers even if their configuration
                    and image haven't changed.
```

We are currently always passing this flag when Tilt is handling the
build (vs `docker-compose build`). However, Docker Compose already
checks if the image it's using is outdated; if it is, it will stop
the existing container and create a new one with the updated image
(and/or config). If the image and config are the same, it will not
replace them.

It looks like the flag has been in place since the original
implementation of Docker Compose code in Tilt, so wasn't a reaction
to any other sort of regression.

In general, it doesn't matter because Tilt only calls
`docker-compose up` after it has done a build. However, when
re-launching Tilt where an already running `docker-compose` session
is active and changes haven't been made, this means after the (fully
cached) image rebuild happens, we'll force the containers to get
recreated, which slows things down.

[1]: https://docs.docker.com/compose/reference/up/

### DIFF
--- a/internal/dockercompose/client.go
+++ b/internal/dockercompose/client.go
@@ -77,14 +77,6 @@ func (c *cmdDCClient) Up(ctx context.Context, configPaths []string, serviceName 
 	runArgs := append([]string{}, genArgs...)
 	runArgs = append(runArgs, "up", "--no-deps", "--no-build", "-d")
 
-	if !shouldBuild {
-		// !shouldBuild implies that Tilt will take care of building, which implies that
-		// we should recreate container so that we pull the new image
-		// NOTE(maia): this is maybe the WRONG thing to do if we're deploying a service
-		// but none of the code changed (i.e. it was just a dockercompose.yml change)?
-		runArgs = append(runArgs, "--force-recreate")
-	}
-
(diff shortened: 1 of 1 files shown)


milas:
A question was asked about restarting Tilt when you've got an active Compose session in support Slack, so I did a couple quick tests around behavior and improving it is straight-forward.

I verified that unchanged image means it leaves it alone, but by manually rebuilding the image while Tilt was not running, Docker Compose did the right thing and picked up the new image.

This _should_ be safe for all versions of Docker Compose because historically the workflow when you're using remote images with rolling tags has been `docker-compose pull && docker-compose up` precisely because there is no `--pull` flag on `up`, but it _does_ notice changes.

nicks (approved):
ya, it's possible that we either made a mistake when verifying the original behavior. it's also possible that the docker-compose behavior has changed since we built this - Docker-team has put a lot of work into docker-compose recently

milas:
It was definitely a reasonable safe starting point, and it _really_ only matters if you're restarting Tilt and have a lot of containers, so I'm not surprised nobody has commented on it until now 🙃

PR #4716, opened 2021-07-01
Accessibility fixes for Table Overview

hyu:
Thanks to @lizzthabet for finding some issues, which will improve accessibility:
- Show star icon on keyboard focus
- Make Header font easier to read
- Clicking on status takes you to logs

Other fixes in this PR: 
- Refine tooltip text for Trigger and Trigger Mode
- Visual tweak for Endpoints
- Tiltfile should not have Trigger Mode

### DIFF
--- a/web/src/OverviewTable.tsx
+++ b/web/src/OverviewTable.tsx
@@ -258,7 +258,11 @@ function TablePodIDColumn({ row }: CellProps<RowValues>) {
         readOnly={true}
         onClick={() => selectPodIdInput(row.values.podId)}
       />
-      <PodIdCopy onClick={copyClick} analyticsName="ui.web.overview.copyPodID">
+      <PodIdCopy
+        onClick={copyClick}
+        analyticsName="ui.web.overview.copyPodID"
+        title="Copy Pod ID"
+      >
         {icon}
       </PodIdCopy>
     </PodId>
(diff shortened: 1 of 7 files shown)


web/src/OverviewTable.tsx
```diff
@@ -277,16 +282,19 @@ function TableEndpointColumn({ row }: CellProps<RowValues>) {
         key={ep.url}
       >
         <StyledLinkSvg />
-        <DetailText>{ep.name || displayURL(ep)}</DetailText>
```
nicks, line 280: hmm...this removes the overflow handling for long urls. was that intentional?

hyu: Oops! That was a mistake. I did just test it and realized that it wasn't working in the first place because I needed to set a width for the parent. 

Thanks for catching this!

web/src/OverviewTable.tsx
```diff
@@ -277,16 +289,21 @@ function TableEndpointColumn({ row }: CellProps<RowValues>) {
...
+        <DetailText title={ep.name || displayURL(ep)}>
+          {ep.name || displayURL(ep)}
+        </DetailText>
       </Endpoint>
     )
   })
-  return <div>{endpoints}</div>
+  return endpoints
```
nicks, line 298: i'm surprised this works? i would have expected this to need a `<>{endpoints}</>` so that it returns a valid react fragment (and is a valid standalone component)

nicks (approved):

web/src/OverviewTable.tsx
hyu: I'll revert it to ` <>{endpoints}</>`...I'd removed the containing div while trying to figure out why I couldn't get keyboard focus to work on the endpoint link. (Turns out it may just be an issue with Firefox). But yes, in retrospect this is surprising this worked, even with multiple endpoints!

Issue #2953, opened 2020-02-13
Abort running local_resource


rpocase:
I'm hitting an issue with a `local_resource` getting stuck in progress indefinitely. The root problem for the indefinite process is on my end. From the CLI, I can kill the parent process (`kill <pid>`), preventing the need to restart the tilt process. Tilt recognizes this (which is great!), but it would be nice to not have to drop to another terminal.

nicks:
Oof. Ya, this isn't good, but not totally sure yet what the right way to fix it is.
- A cancel UI for local()/local_resource() processes that take a long time?
- A default timeout that you can configure for long tasks?

rpocase:
Those are the two thoughts I had as well. They both seem viable for different use cases. 

I was initially concerned with clutter on the UI for a cancel, but the refresh symbol is already removed for a running resource, so adding a cancel wouldn't be too bad.

landism:
FWIW, when we tackle this, I don't think it's specific to local_resource. If one accidentally puts a long-running command in a Dockerfile, 1) there's no way to cancel and 2) the refresh symbol is removed, leaving the same space for a cancel

PR #4720, opened 2021-07-01
engine: run local update commands with the reconciler

nicks:
Hello @milas, @landism,

Please review the following commits I made in branch nicks/cmd:

d6c502ec3e1540ca6429e2ef66217aea965ad3d4 (2021-07-01 14:38:41 -0400)
engine: run local update commands with the reconciler


nicks:
after this change, it should be possible to implement https://github.com/tilt-dev/tilt/issues/2953 as an extension, by looking up commands annotated for a manifest, and killing them by PID

milas (approved):
Dunno if I mentioned in the other PR that was similar, but I actually think this pattern (of reconciler exporting some logic for the engine to share) is nice - it's unconventional but very easy to reason about/follow

nicks:
i'm slightly worried that this pattern will cause problems down the line (since we'll have a bunch of objects in the system that don't behave correctly when you edit them), but it definitely makes the migration easier

PR #4721, opened 2021-07-01
apis: add PodLogStreamTemplateSpec, PortForwardTemplateSpec

nicks:
Hello @milas,

Please review the following commits I made in branch nicks/pls:

5e57145a5799b28c0a6d8028718fa8d73d62638a (2021-07-01 15:22:23 -0400)
apis: add PodLogStreamTemplateSpec, PortForwardTemplateSpec
The basic types here should be non-controversial, and are something
you would want to have as a field on different types of objects.

The more controversial part of this PR is whether they should be fields of
KubernetesDiscovery, or of some more top-level type. I go back and forth on
this, but think we might end up playing around with different ownership models.


milas (approved):
All right, let's do it! 😝

PR #4722, opened 2021-07-01
api: add fields to UIButton for nav bar support

milas:
* Optional `icon` field - will be primary for navbar buttons with
  `text` as hover; for resource icons, will be alongside
* New constant for component type: `Global` (expected component ID
  for navbar will be `nav`)
* Validation to ensure `icon` is populated for `Global:nav` buttons


pkg/apis/core/v1alpha1/uibutton_types.go
```diff
@@ -56,8 +57,16 @@ type UIButtonSpec struct {
...
+	// Text to appear on the button itself or as hover text (depending on button location).
 	Text string `json:"text" protobuf:"bytes,2,opt,name=text"`
+
+	// Icon to appear next to button text or on the button itself (depending on button location).
+	//
+	// Valid values are icon font ligature names from the Material Icons set.
+	// See https://fonts.google.com/icons for the full list of available icons.
+	//
+	// +optional
+	Icon string `json:"icon,omitempty" protobuf:"bytes,3,opt,name=icon"`
```
nicks, line 69: maybe call this IconName?

tbh, it's pretty fashionable among icon hipsters to embed SVGs everywhere. they're pretty small. so can we have IconSVG too?

nicks (approved):

pkg/apis/core/v1alpha1/uibutton_types.go
nicks: would probably call this 'IconName'

pkg/apis/core/v1alpha1/uibutton_types.go
```diff
@@ -129,6 +139,16 @@ func (in *UIButton) Validate(_ context.Context) field.ErrorList {
 			locField.Child("componentType"), "Parent component type is required"))
 	}
 
+	if strings.EqualFold(string(in.Spec.Location.ComponentType), string(ComponentTypeGlobal)) {
+		switch strings.ToLower(in.Spec.Location.ComponentID) {
+		case "nav":
+			if in.Spec.Icon == "" {
+				fieldErrors = append(fieldErrors, field.Required(
```
nicks, line 146: also - i think it would be way funnier to pick an icon at random, based on, say, a hash of the button text

nicks: or always default to a goose icon or smth

milas: Ah that's fair, we can pick a default to make it friendlier - I'll remove this

pkg/apis/core/v1alpha1/uibutton_types.go
milas: I like the SVG idea - especially for the "help button" it might be nice to have a branded button, and shouldn't be too difficult on the frontend either.

(FYI To make the API friendlier to use, I didn't make it exclusive but documented SVG will take precedence if both are defined.)

pkg/apis/core/v1alpha1/uibutton_types.go
```diff
@@ -56,8 +56,26 @@ type UIButtonSpec struct {
...
+	//
+	// Valid values are icon font ligature names from the Material Icons set.
+	// See https://fonts.google.com/icons for the full list of available icons.
+	//
+	// If both IconSVG and IconName are specified, IconSVG will take precedence.
+	//
+	// +optional
+	IconName string `json:"iconName,omitempty" protobuf:"bytes,3,opt,name=iconName"`
+
+	// IconSVG is an SVG to use as the icon to appear next to button text or on the button itself (dependong on button
```
landism, line 72:
```suggestion
	// IconSVG is an SVG to use as the icon to appear next to button text or on the button itself (depending on button
```

landism: It seems like it'd be tempting for a user to put a path to an SVG file here. It'd be good to specify in the comment that this wants raw xml not a filename and maybe also validate that it's XML in `Validate` to catch people who miss the note in the docs?

landism (approved):

pkg/apis/core/v1alpha1/uibutton_types.go
milas: Good points - clarified in docs and added ridiculously basic validation (`strings.Contains(in.Spec.IconSVG, "<svg")`) + fixed that typo in a6d0254e8 (needed to re-run codegen so didn't accept GitHub suggestion 🤪).

PR #4724, opened 2021-07-02
web: Replace Grid with Table Overview

hyu:
The main purpose of this PR is to replace the Grid with the Table. As cleanup, it also includes these changes:
- Replace the Grid Storybook with Table Storybook
- Delete Grid components and related storybooks + tests: OverviewPane > OverviewGrid > OverviewItemView

**Also note:** After deleting the Overview Grid, I had to fix the TriggerModeToggle test that was mounting the Grid. With this PR, it mounts `SidebarResources` instead. I also renamed this component to `SidebarTriggerModeToggle` to indicate that it's really only used in the Sidebar, and refactored accordingly.

&nbsp;

Since this effectively releases the feature, I also fixed this high-priority item from [Surbhi's QA List](https://docs.google.com/document/d/1XSsWdbmGbw8LeSOSjl8doTYp8HUEopyOMUwIERyoRak/edit#):
- Better color contrast for SortBy triangle (and more sensible tooltip)

----


web/src/SidebarTriggerModeToggle.tsx
```diff
@@ -87,16 +87,16 @@ function TriggerModeToggle(props: TriggerModeToggleProps) {
   }
 
   return (
-    <TriggerModeToggleRoot
+    <StyledSidebarTriggerModeToggle
```
hyu, line 90: We've been using the naming convention suffix of `Root` to avoid naming conflicts between React Components and styled-components. But I noticed in the official [docs](https://styled-components.com/docs/basics#define-styled-components-outside-of-the-render-method) that the convention is to use a prefix of `Styled`. 

Since I had to change the name here anyhow, I've added the `Styled` prefix because I think it's more obvious than `Root`.

nicks (approved):
wooooo

PR #4725, opened 2021-07-02
web: add SVG support for UIButton icons

milas:
SVG will take precedence over an icon name. If neither is provided,
behavior depends on button location: resource buttons will simply
have no icon, while navbar buttons will get a default fallback of
`smart_button`[1].

[1]: https://fonts.google.com/icons?selected=Material+Icons+Outlined:smart_button


milas:
I did have to pull in a new small package to handle safely creating `<svg>` DOM elements dynamically. (Can't just `` because then you can't style with CSS, so plain `path` SVGs won't get automatically colored for the UI theme.)

web/src/ApiButton.tsx
```diff
@@ -1,12 +1,34 @@
-import { Icon } from "@material-ui/core"
+import { Icon, SvgIcon } from "@material-ui/core"
 import moment from "moment"
 import React, { useState } from "react"
+import InlineSVG from "react-inlinesvg"
```
nicks, line 4: this component seems to do a ton of things we don't really need

https://github.com/gilbarbara/react-inlinesvg#props

have you considered using a data url directly? we can use the vanillajs library, which is great: http://vanilla-js.com/

nicks: ah, saw your above comment, i think you just want to use react-from-dom, not the svg wrapper library around it

nicks: which looks like it's just a thin wrapper over https://developer.mozilla.org/en-US/docs/Web/API/DOMParser/parseFromString

milas: This was the route I originally went - simply `` but the big limitation is that SVGs embedded that way don't respect CSS (understandably). To have CSS apply, so they are the correct color, they have to be actual `<svg>` elements in the DOM.

That said, you're right that our use case is very limited, so I can do the DOM manipulation manually and save a dependency 🙃

nicks: ya, i would be fine with the react-from-dom dependency, though i honestly think we could get away with dangerouslySetInnerHTML per https://www.npmjs.com/package/react-from-dom

milas: Hmmm...so I guess there's two paths here:

* Create a wrapper element with React and then `dangerouslySetInnerHTML` with the user SVG data
  * Would need to try to replicate some of the CSS stuff `<SvgIcon>` is doing for us
  * Should sanitize the input first

* Use `react-from-dom` which uses a `DOMParser` to get a DOM node and then maps that to a React element
  * Pro: uses `DOMParser` which eliminates one class of escape issues
  * Pro: easy to integrate with `<SvgIcon>`
  * Should still sanitize the input (see below)

For sanitization, we should probably run everything through [DOMPurify](https://github.com/cure53/DOMPurify) regardless - apparently you can include `<script>` tags in SVG!?

milas: All right - check out a61510b which gets rid of `react-inlinesvg` and just uses `react-from-dom` directly (bonus: it's actually stricter on the mime-type + selector for the parser 😛)

nicks (approved):

PR #4727, opened 2021-07-02
kubernetesdiscovery: propagate PortForward template

nicks:
Hello @milas, @landism,

Please review the following commits I made in branch nicks/pf2:

27780ba24c08472638756280492ff64df3b6f46e (2021-07-02 18:34:35 -0400)
kubernetesdiscovery: propagate PortForward template

### DIFF
--- a/internal/controllers/core/kubernetesdiscovery/portforwards.go
+++ b/internal/controllers/core/kubernetesdiscovery/portforwards.go
@@ -1,0 +1,176 @@
+package kubernetesdiscovery
+
+import (
+	"context"
+	"fmt"
+
+	v1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	errorutil "k8s.io/apimachinery/pkg/util/errors"
+	ctrlclient "sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+
+	"github.com/tilt-dev/tilt/internal/controllers/apicmp"
+	"github.com/tilt-dev/tilt/internal/store"
+	"github.com/tilt-dev/tilt/pkg/apis/core/v1alpha1"
+)
+
+func (r *Reconciler) manageOwnedPortForwards(ctx context.Context, kd *v1alpha1.KubernetesDiscovery) error {
+	var pfList v1alpha1.PortForwardList
+	err := r.ctrlClient.List(ctx, &pfList, ctrlclient.InNamespace(kd.Namespace),
+		ctrlclient.MatchingFields{ownerKey: kd.Name})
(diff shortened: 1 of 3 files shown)


nicks:
this is mostly adapted from https://github.com/tilt-dev/tilt/blob/master/internal/engine/portforward/subscriber.go

internal/controllers/core/kubernetesdiscovery/portforwards.go
```diff
@@ -0,0 +1,180 @@
...
+		if bestPod == nil || isBetterPortForwardPod(&pod, bestPod) {
+			bestPod = &pod
+		}
+	}
+	return bestPod
+}
+
+// Check if podA is better than podB for port-forwarding.
+func isBetterPortForwardPod(podA, podB *v1alpha1.Pod) bool {
+	// A deleting pod is always better than a non-deleting pod.
```
milas, line 164:
```suggestion
	// A non-deleting pod is always better than a deleting pod.
```
I might have not yet had enough coffee, but I think this should be the opposite? That seems to be what the logic is actually doing - comment just caught me off-guard 😛

internal/controllers/core/kubernetesdiscovery/reconciler.go
```diff
@@ -179,6 +189,9 @@ func (w *Reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
 			w.teardown(key)
 			w.cleanupAbandonedNamespaces()
 		}
+
+		// TODO(nick): Delete any orphaned PortForwards or PodLogStreams.
```
milas, line 193: Ahh whoops, I can't believe I spent the time tracking down the lack of automatic GC on owner references in our apiserver and then didn't actually write the logic to handle it manually 🤦

milas (approved):

internal/controllers/core/kubernetesdiscovery/portforwards.go
```diff
@@ -0,0 +1,180 @@
...
+	errs := []error{}
+	foundDesired := false
+	for _, existingPF := range pfList.Items {
+		matchesPF := pf != nil && existingPF.Name == pf.Name
+		if matchesPF {
+			foundDesired = true
+
+			// If this PortForward is already in the APIServer, make sure it's up-to-date.
+			if apicmp.DeepEqual(pf.Spec, existingPF) &&
+				existingPF.ObjectMeta.Annotations[v1alpha1.AnnotationManifest] ==
```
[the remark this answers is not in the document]
nicks: oooh, i like that idea. done!

internal/controllers/core/kubernetesdiscovery/portforwards.go
nicks: oops, fixed

internal/controllers/core/kubernetesdiscovery/reconciler.go
nicks: :laughing:

PR #4742, opened 2021-07-08
web: Add basic display of labeled resources in Storybook behind feature flag

lizzthabet:
This PR adds basic resource grouping by label using mocked data in Storybook and the `labels` feature flag (which can be toggled in Storybook) for [ch12250](https://app.clubhouse.io/windmill/story/12250/display-label-headings-and-resource-groups-behind-feature-flag). I haven't added any label-specific styles yet, since I want to use the MaterialUI [accordion](https://material-ui.com/components/accordion/#accordion) component and thought that would make this PR too large. (I'll tackle the styles and accordion functionality next for [ch12252](https://app.clubhouse.io/windmill/story/12252/add-expand-collapse-functionality-for-label-groups)).

With feature flag off, everything should remain the same:
Screen Shot 2021-07-08 at 6 10 57 PM

With feature flag on, resources appear under each label they have, with unlabeled resources at the bottom:
Screen Shot 2021-07-08 at 6 11 12 PM

### DIFF
--- a/web/src/SidebarResources.tsx
+++ b/web/src/SidebarResources.tsx
@@ -1,5 +1,7 @@
-import React, { Dispatch, SetStateAction } from "react"
+import React, { Dispatch, PropsWithChildren, SetStateAction } from "react"
 import styled from "styled-components"
+import Features, { FeaturesContext, Flag } from "./feature"
+import { orderLabels } from "./labels"
 import { PersistentStateProvider } from "./LocalStorage"
 import { OverviewSidebarOptions } from "./OverviewSidebarOptions"
 import PathBuilder from "./PathBuilder"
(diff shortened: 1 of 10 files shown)


web/src/SidebarResources.tsx
```diff
@@ -48,6 +49,72 @@ export function SidebarListSection(
...
+        }
+
+        labelsToResources[label].push(item)
+      })
+    } else {
+      unlabeled.push(item)
+    }
+  })
+
+  const labels = Object.keys(labelsToResources)
```
nicks, line 101: how much do we care about the order of labels? feels weird to implicitly rely on whatever the JS implementation order is.

web/src/SidebarResources.tsx
```diff
@@ -48,6 +49,72 @@ export function SidebarListSection(
...
+    <>
+      {labels.map((label) => (
+        <SidebarLabelListSection
+          {...props}
+          key={`sidebarItem-${label}`}
+          label={label}
+          items={labelsToResources[label]}
+        />
+      ))}
+      <SidebarLabelListSection {...props} label="unlabeled" items={unlabeled} />
```
nicks, line 113: probably more of a question for Surbhi, but I feel like it's going to be confusing to have both an "unlabeled" section and "uncategorized" resource

web/src/labels.ts
```diff
@@ -0,0 +1,19 @@
...
+// and why we filter out labels with prefixes
+export function getUiLabels({ labels }: UILabels) {
+  if (!labels) {
+    return []
+  }
+
+  return Object.keys(labels)
+    .filter((labelKey) => {
+      const labelHasPrefix = labelKey.includes("/")
+      console.log
```
nicks, line 15: ?

web/src/SidebarResources.tsx
```diff
@@ -48,6 +49,72 @@ export function SidebarListSection(
...
+  const unlabeled: SidebarItem[] = []
+
+  props.items.forEach((item) => {
+    if (item.labels.length) {
+      item.labels.forEach((label) => {
+        if (!labelsToResources.hasOwnProperty(label)) {
+          labelsToResources[label] = []
+        }
+
+        labelsToResources[label].push(item)
```
nicks, line 94: i wonder if it's going to be weird to have an item in multiple groups

nicks (approved):

web/src/labels.ts
lizzthabet: lol ⏳will fix momentarily

web/src/SidebarResources.tsx
lizzthabet: ya, that's a good question! @SurbhiSGupta may have user stories/feedback that she can share. right now, i just have the `unlabeled` section implemented.

web/src/SidebarResources.tsx
lizzthabet: hmm, how do you feel about alphabetizing the labels? that way, labels show up consistently in the same order (which i would find helpful as a user and as a user, i'd be able to manipulate the alphabet system if i want my labels to display in a specific order). otherwise, in theory, the labels will appear in the order of whatever resources we iterate over first.

web/src/SidebarResources.tsx
lizzthabet: fwiw it hasn't seemed odd to me while looking at all the designs. i think that only displaying a resource under the first label it has might be an unintuitive behavior, but it'll be interesting to get user feedback!

web/src/SidebarResources.tsx
lizzthabet: i went ahead and added label sorting code! we can revisit if there's a better suggestion for sorting.

lizzthabet:
i don't think any of the discussions here are blocking, so i'll merge in now and can address any more feedback in subsequent prs.

Issue #4754, opened 2021-07-10
non-deterministic resource loading


nicks:
### Steps to Reproduce
1. Create a Tiltfile like:

```
local('sleep 10')
...
# register 5 resources
```

2. Start tilt with `tilt up --web-mode=prod`
3. Connect to the tilt webserver before the resources finish loading

### Expected Behavior
The resources should load in a deterministic order

### Current Behavior
The resources load in a random order

### Context
You can currently work around this problem by refreshing the page, which will load them in the right order. There's a bug in how the web ui is handling incremental updates.

PR #4756, opened 2021-07-12
web: sort objects by 'order'

nicks:
Hello @landism, @lizzthabet,

Please review the following commits I made in branch nicks/ch12353:

ad564fbc8e7b860a387cc4c435c9d889b0c970d2 (2021-07-12 17:05:03 -0400)
web: sort objects by 'order'


nicks:
someone complained about this on friday (see: https://github.com/tilt-dev/tilt/issues/4754)

as we get more into helping users to organize resources in the way they care about, i suspect ordering is going to become more important...for now, i'm just trying to do something deterministic so that people aren't caught off-guard.

nicks:
fwiw, i don't totally understand why this PR introduces a bunch of typescript warnings in files i didn't modify, but i went ahead and updated those too

nicks:
ah, nm, i see what happened, we had some manual edits in view.d.ts that got overwritten when i regenerated it

lizzthabet:
> ah, nm, i see what happened, we had some manual edits in view.d.ts that got overwritten when i regenerated it

oohhh this is my bad! if i remember correctly, i edited the file manually cause i wanted `labels` to be typed as `[key: string]: string` instead of a generic object. i can put up a pr to fix, so we don't have to `any` type the label code as a workaround.

nicks:
AFAICT, there is SOME way to set up the code-generator to give `labels` the right type.

the code generator that the kuberentes project uses does it that way: https://github.com/kubernetes-client/javascript/blob/master/src/gen/model/v1ObjectMeta.ts

but we're using a different code generator that doesn't support this...

web/src/HUD.tsx
```diff
@@ -332,6 +332,18 @@ export function HUDFromContext(props: React.PropsWithChildren<{}>) {
   return <HUD history={history} interfaceVersion={interfaceVersion} />
 }
 
+function compareObjects(a: any, b: any): number {
```
lizzthabet, line 335: (optional nitpick) What about a more descriptive name, like `compareObjectsOrder`, so it's easier to tell how this function is being used?

It may not be worth the Typescript finagling, but I _think_ it would work to type this as `<T extends { status?: Proto.v1alpha1UIResourceStatus, metadata?: Proto.v1ObjectMeta }>`? 🤞🏻

lizzthabet (approved):
Lgtm! Minor question and suggestion, and apologies for the merge conflicts I probably just introduced, but that should solve all the label issues you saw.

internal/hud/webview/convert.go
```diff
@@ -188,25 +188,28 @@ func ToUISession(s store.EngineState) *v1alpha1.UISession {
...
 		}
 
-		r, err := ToUIResource(mt, state)
+		r, err := toUIResource(mt, state)
 		if err != nil {
 			return nil, err
 		}
 
+		// The Tiltfile has Order -1, all "normal" resources have Order >= 1
+		r.Status.Order = int32(i + 1)
```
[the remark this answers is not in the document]
nicks: yep! I think it depends on the order they're registered in the tiltfile itself.

web/src/HUD.tsx
nicks: i like that, done!

got something that kinda works...

PR #4757, opened 2021-07-13
web: type the label data better to handle the generic generated type

lizzthabet:
These are some slight Typescript gymnastics to cast the generic `object` to `{ [key: string]: string }` for label data. 🙃 I'm not sure this is phenomenally better than using `any` types for labels, but it does let us keep stricter type information.

### DIFF
--- a/web/src/labels.ts
+++ b/web/src/labels.ts
@@ -1,10 +1,35 @@
 // Helper functions for working with labels
 
-type UILabels = Pick<Proto.v1ObjectMeta, "labels">
+// The generated type for labels is a generic object,
+// when in reality, it is an object with string keys and values,
+// so we can use a predicate function to type the labels object
+// more precisely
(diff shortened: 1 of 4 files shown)


web/src/labels.ts
```diff
@@ -1,10 +1,37 @@
...
+type UILabelsGenerated = Pick<Proto.v1ObjectMeta, "labels">
+interface UILabels extends UILabelsGenerated {
+  labels: { [key: string]: string } | undefined
+}
+
+function isUILabels(
+  labelsWrapper: UILabelsGenerated
+): labelsWrapper is UILabels {
+  const { labels } = labelsWrapper
+  if (labels === undefined || typeof labels === "object") {
```
nicks, line 16: can also just return the condition directly

nicks (approved):

PR #4765, opened 2021-07-14
web: add accordion functionality for resources grouped by labels

lizzthabet:
(feat) Add accordion functionality for resources grouped by labels in sidebar with styles and status summary

This PR uses Material UI's accordion component to display collapse-able and expand-able groups of resources in the sidebar. It also adds styles to align with the design and a small refactoring of resource summary display to cover different use cases. You can take a look at the `Resources with Labels` storybook entry to test. The resources sidebar should not look any different in the Tilt UI (outside of Storybook), cuz even though the feature flag is enabled in `storybook.tiltfile`, no resources have labels defined in the Tiltfile.

~There are a couple todos (noted in code) that I'd like to take care of before merging this in, but wanted to get this up for code review first.~ ✅ done! (minus the accessibility note, that i will address later)

**update:** Styles have been to address the long label name issue, add Tiltfile resource on its own, and italicize the `unlabeled` resources:
Screen Shot 2021-07-15 at 4 36 22 PM

Closes [ch12252](https://app.clubhouse.io/windmill/story/12252/add-expand-collapse-functionality-for-label-groups)

### DIFF
--- a/web/src/OverviewResourceSidebar.stories.tsx
+++ b/web/src/OverviewResourceSidebar.stories.tsx
@@ -64,7 +64,8 @@ export const ResourcesWithLabels = () => {
   const view = nResourceView(10)
   for (let i = 0; i < 10; i++) {
     const labels: { [key: string]: string } = {}
-    if (i < 5) {
+    // The first item is a Tiltfile, so don't apply a label to it
+    if (i > 0 && i < 5) {
       labels["frontend"] = "frontend"
     }
     if (i % 2) {
@@ -72,15 +73,10 @@ export const ResourcesWithLabels = () => {
     }
 
     if (i === 3) {
-      labels["some_really_really_really_long_label_name"] =
-        "some_really_really_really_long_label_name"
+      labels["very_long_long_long_label"] = "very_long_long_long_label"
     }
 
-    const resourceMetadata: Proto.v1ObjectMeta = {
-      name: `resource_${i}`,
-      labels,
-    }
-    view.uiResources[i].metadata = resourceMetadata
+    view.uiResources[i].metadata!.labels = labels
   }
 
   // Non-happy path resources
@@ -99,6 +95,8 @@ export const ResourcesWithLabels = () => {
   }
   crashedStart.metadata!.name = "resource_12"
   view.uiResources.push(crashedStart)
+
+  // TODO: Add a Tiltfile resource
 
   return <OverviewResourceSidebar name={"vigoda_1"} view={view} />
 }
--- a/web/src/OverviewResourceSidebar.stories.tsx
+++ b/web/src/OverviewResourceSidebar.stories.tsx
@@ -96,8 +96,6 @@ export const ResourcesWithLabels = () => {
   crashedStart.metadata!.name = "resource_12"
   view.uiResources.push(crashedStart)
 
-  // TODO: Add a Tiltfile resource
-
   return <OverviewResourceSidebar name={"vigoda_1"} view={view} />
 }
 
--- a/web/src/OverviewResourceSidebar.stories.tsx
+++ b/web/src/OverviewResourceSidebar.stories.tsx
@@ -71,6 +71,11 @@ export const ResourcesWithLabels = () => {
       labels["test"] = "test"
     }
 
+    if (i === 3) {
+      labels["some_really_really_really_long_label_name"] =
+        "some_really_really_really_long_label_name"
+    }
+
     const resourceMetadata: Proto.v1ObjectMeta = {
       name: `resource_${i}`,
       labels,
--- a/web/src/ResourceStatusSummary.tsx
+++ b/web/src/ResourceStatusSummary.tsx
@@ -8,7 +8,7 @@ import { ReactComponent as WarningSvg } from "./assets/svg/warning.svg"
 import { FilterLevel } from "./logfilters"
 import { usePathBuilder } from "./PathBuilder"
 import SidebarItem from "./SidebarItem"
-import { summaryStatus, summaryStatusFromAllStatuses } from "./status"
+import { buildStatus, combinedStatus, runtimeStatus } from "./status"
 import {
   Color,
   Font,
@@ -287,18 +287,14 @@ export function ResourceStatusSummary(props: ResourceStatusSummaryProps) {
   let resources = props.view.uiResources || []
   const allStatuses: ResourceStatus[] = []
 
-  let testResources = new Array<UIResource>()
   const testStatuses: ResourceStatus[] = []
-  let otherResources = new Array<UIResource>()
   const otherStatuses: ResourceStatus[] = []
   resources.forEach((r) => {
-    const status = summaryStatus(r)
+    const status = combinedStatus(buildStatus(r), runtimeStatus(r))
     allStatuses.push(status)
     if (r.status?.localResourceInfo?.isTest) {
-      testResources.push(r)
       testStatuses.push(status)
     } else {
-      otherResources.push(r)
       otherStatuses.push(status)
     }
   })
@@ -335,10 +331,7 @@ export function ResourceSidebarStatusSummary(
   // Because SidebarItems have already-calculated statuses,
   // pass those directly to determine their summary status
   const statuses: ResourceStatus[] = props.items.map((item) =>
-    summaryStatusFromAllStatuses({
-      buildStatus: item.buildStatus,
-      runtimeStatus: item.runtimeStatus,
-    })
+    combinedStatus(item.buildStatus, item.runtimeStatus)
   )
 
   return (
--- a/web/src/ResourceStatusSummary.tsx
+++ b/web/src/ResourceStatusSummary.tsx
@@ -7,7 +7,8 @@ import { ReactComponent as PendingSvg } from "./assets/svg/pending.svg"
 import { ReactComponent as WarningSvg } from "./assets/svg/warning.svg"
 import { FilterLevel } from "./logfilters"
 import { usePathBuilder } from "./PathBuilder"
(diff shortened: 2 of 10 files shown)


web/src/status.tsx
```diff
@@ -52,16 +52,32 @@ function runtimeStatus(r: UIResource): ResourceStatus {
...
-  if (bs !== ResourceStatus.Healthy && bs !== ResourceStatus.None) {
-    return bs
+function summaryStatus(res: UIResource): ResourceStatus {
+  return summaryStatusFromAllStatuses({
+    buildStatus: buildStatus(res),
+    runtimeStatus: runtimeStatus(res),
+  })
+}
+
+function summaryStatusFromAllStatuses({
```
lizzthabet, line 62: Very open to function naming suggestions

web/src/ResourceStatusSummary.tsx
```diff
@@ -283,31 +283,38 @@ type ResourceStatusSummaryProps = {
...
 export function ResourceStatusSummary(props: ResourceStatusSummaryProps) {
-  // Count the statuses.
+  // Count and calculate the combined statuses.
   let resources = props.view.uiResources || []
+  const allStatuses: ResourceStatus[] = []
 
   let testResources = new Array<UIResource>()
+  const testStatuses: ResourceStatus[] = []
   let otherResources = new Array<UIResource>()
+  const otherStatuses: ResourceStatus[] = []
```
landism, line 293: it looks like `testResources` and `otherResources` are now unused and can be removed?

web/src/status.tsx
landism: What do you think of keeping the original function but changing the signature so that you can pass it SidebarItem's info? e.g.:
```
function combinedStatus(buildStatus: ResourceStatus, runtimeStatus: ResourceStatus)
```
and then the two calls become
```
  resources.forEach((r) => {
    const status = combinedStatus(buildStatus(r), runtimeStatus(r))
```
and
```
  const statuses: ResourceStatus[] = props.items.map((item) =>
    combinedStatus(item.buildStatus, item.runtimeStatus)
  )
```

web/src/OverviewResourceSidebar.stories.tsx
```diff
@@ -71,6 +71,11 @@ export const ResourcesWithLabels = () => {
       labels["test"] = "test"
     }
 
+    if (i === 3) {
```
landism, line 74: I don't know if this is a new issue in this PR, and not really a blocker, but if it's cheap to fix: this surfaces that the status icon for this group is a bit unaligned with the status icons for the other groups:
image

web/src/ResourceStatusSummary.tsx
lizzthabet: yes! i will fix.

web/src/assets/svg/caret.svg
```diff
@@ -0,0 +1,3 @@
+<svg width="9" height="9" viewBox="0 0 9 9" fill="none" xmlns="http://www.w3.org/2000/svg">
+<path d="M0 0L0 9L9 4.5L0 0Z" fill="#586E75"/>
```
nicks, line 2: fwiw, i usually do class="fillStd" on my SVG paths rather than hardcode the color. That makes it easier to change the color, add transitions, etc.

web/src/ResourceStatusSummary.tsx
```diff
@@ -283,31 +283,38 @@ type ResourceStatusSummaryProps = {
 }
 
 export function ResourceStatusSummary(props: ResourceStatusSummaryProps) {
-  // Count the statuses.
+  // Count and calculate the combined statuses.
   let resources = props.view.uiResources || []
+  const allStatuses: ResourceStatus[] = []
 
   let testResources = new Array<UIResource>()
+  const testStatuses: ResourceStatus[] = []
```
nicks, line 291: i wonder if it would make sense to rip out the test stuff first, so you don't have to stress about it.

web/src/SidebarResources.tsx
```diff
@@ -39,6 +49,69 @@ const NoMatchesFound = styled.li`
...
+
+  &.MuiAccordion-root,
+  &.MuiAccordion-root.Mui-expanded {
+    margin: ${SizeUnit(1 / 3)} 0;
+  }
+`
+
+const SummaryIcon = styled(CaretSvg)`
+  padding: ${SizeUnit(1 / 4)};
+  transition: all 300ms cubic-bezier(0.4, 0, 0.2, 1) 0ms; /* Copied from MUI accordion */
```
nicks, line 65: i dunno that i trust "transition: all", seems too easy to accidentally transition a property you don't expect

web/src/ResourceStatusSummary.tsx
lizzthabet: that's not a bad idea at all! how about i merge this one in as-is and then address that ticket next?

web/src/assets/svg/caret.svg
lizzthabet: good idea, done!

web/src/SidebarResources.tsx
lizzthabet: fixed 👍🏻

web/src/status.tsx
lizzthabet: i like this much better. thanks for the suggestion!

web/src/SidebarItemView.tsx
```diff
@@ -38,6 +38,11 @@ export const SidebarItemRoot = styled.li`
   ${StarResourceButtonRoot} {
     margin-right: ${SizeUnit(1.0 / 12)};
   }
+
+  /* groupIndent is used for un-grouped items in the grouped sidebar view */
+  &.groupIndent {
```
landism, line 43: I think I understand what the code is doing but I'm finding myself confused about what "group" means.
1. By itself, "groupIndent is used for un-grouped items" reads pretty strangely. I guess the "group" there is referring to the view rather than the items, but it took me a while to get there. Maybe "groupedSidebarResourceIndent"?
2. The comment here says it's used for ungrouped items, but [below](https://github.com/tilt-dev/tilt/pull/4765/files#diff-2114c24a4cdf4a64c9f1ce43f4a80ab57f605472e53a40564b0f2dccf7f44acfR275) it seems to only be used when `inGroup` is true, which seems like the opposite of what the comment says?
3. But then [further down](https://github.com/tilt-dev/tilt/pull/4765/files#diff-363468d2eab0757279a1799e1c908efe2d100ba43fce7e10c34e593046610c2aR220), it turns out `inGroup` is only true for the Tiltfile, which I would have thought was the only resource not in a group.

web/src/OverviewResourceSidebar.stories.tsx
```diff
@@ -95,6 +96,8 @@ export const ResourcesWithLabels = () => {
   crashedStart.metadata!.name = "resource_12"
   view.uiResources.push(crashedStart)
 
+  // TODO: Add a Tiltfile resource
```
landism, line 99: 👀

web/src/SidebarResources.tsx
```diff
@@ -112,6 +216,9 @@ function SidebarGroupedByLabels(props: SidebarProps) {
         />
       ))}
       <SidebarLabelListSection {...props} label="unlabeled" items={unlabeled} />
+      <SidebarListSection name="Tiltfile">
```
landism, line 219: nit - if the user types something that doesn't match into the resource name filter box, this will still show the "tiltfile" section w/ no resources. This is inconsistent with the label groups, which disappear when no resources match.

web/src/ResourceStatusSummary.tsx
nicks: :+1:

web/src/SidebarResources.tsx
```diff
@@ -94,9 +193,14 @@ function SidebarGroupedByLabels(props: SidebarProps) {
 
         labelsToResources[label].push(item)
       })
-    } else {
+    } else if (!item.isTiltfile) {
       unlabeled.push(item)
     }
+
+    // Display the Tiltfile outside of the label groups
```
nicks, line 200: i wonder if the Tiltfile itself should have a special built-in label...in the future, we might have more than one tiltfile resource

nicks (approved):

landism (approved):

web/src/SidebarResources.tsx
lizzthabet: this is definitely something to keep in mind! i think right now, the label grouping code can handle multiple tiltfiles, as long as the `isTiltfile` property is set properly further up stream.

i'm curious, from a technical standpoint, is it possible for a user to add labels to a tiltfile from the `Tiltfile`? 👀

web/src/SidebarItemView.tsx
lizzthabet: that's totally fair 😅 i found myself struggling a bit with the naming. this class is more like "indent un-grouped items to align with grouped items in grouped view". what do you think about the `inGroup` option prop becoming `groupView`? i can update the comment with a clearer explanation.

web/src/SidebarResources.tsx
lizzthabet: ahh, thank you! i forgot to test with the resource filter. let me take a look and fix.

lizzthabet: this should be fixed now, so that we don't display the grouped view when a resource name filter is applied.

web/src/SidebarItemView.tsx
lizzthabet: i improved the comment and naming, so hopefully it's more clear what it does!

Issue #4736, opened 2021-07-07
Regression in dc_resource link support 


ebenoist:
### Expected Behavior
Docker compose resource links should be present in the UI as per: https://github.com/tilt-dev/tilt/pull/4175

### Current Behavior
Links declared in `dc_resource` calls are no longer present (open ports declared in the docker-compose.yml still provide links)

### Steps to Reproduce
1. Download latest tilt
2. Add a link to an existing dc_resource
3. Inspect the UI
```
dc_resource(
    'foozles',
    links=[
      link('https://foozles.devo.test')
      ],
    )
```

#### `tilt doctor` Output

```
Tilt: v0.21.2, built 2021-07-06
System: darwin-amd64
---
Docker
- Host: [default]
- Server Version: 20.10.7
- API Version: 1.41
- Builder: 2
---
Kubernetes
- Env: docker-for-desktop
- Context: docker-desktop
- Cluster Name: docker-desktop
- Namespace: default
- Container Runtime: docker
- Version: v1.21.2
- Cluster Local Registry: none
---
Thanks for seeing the Tilt Doctor!
Please send the info above when filing bug reports. 💗

The info below helps us understand how you're using Tilt so we can improve,
but is not required to ask for help.
---
Analytics Settings
--> (These results reflect your personal opt in/out status and may be overridden by an `analytics_settings` call in your Tiltfile)
- User Mode: default
- Machine: 9dba14d67dd4c502e594a05aa1a0e5ec
- Repo: HqkW338Dth3PB3xaNOmm8Q==
```

hyu:
Hi @ebenoist, I just did a quick test using the latest release ([v0.21.2](https://github.com/tilt-dev/tilt/releases/tag/v0.21.2)), and I do see the specified link in the main overview UI:

<img width="1457" alt="Screen Shot 2021-07-08 at 4 24 21 PM" src="https://user-images.githubusercontent.com/20349/124986053-2da44580-e009-11eb-85bc-9f6c68ddc372.png">

Is that the UI you were referring to, or did you mean something else?

ebenoist:
This is the UI I'm talking about, but I'm not seeing the custom links like you have listed. I wish I did! It does look like you're on v0.21.1 based on your screenshot though?

![image](https://user-images.githubusercontent.com/1656713/125000662-7f070180-e016-11eb-9ce9-f6637be3ea8c.png)

```
dc_resource(
    'rql',
    links=[
      link('https://rql.devo.test')
[...]

Issue #3738, opened 2020-08-31
make resource settings editable in the Tiltfile


landism:
### Problem

If my Tiltfile loads another Tiltfile (or extension), and the loaded Tiltfile declares a resource, I can't change that resource's options (e.g., add objects to it or rename it or make it manually triggered). The only recourse is to change the loaded Tiltfile (or extension) to be more parameterized and plumb through whatever settings I want to configure.

The concrete use case that inspired this issue is that KOGI wants to [add more objects to a k8s_resource created by an extension](https://github.com/tilt-dev/tilt/issues/3714#issuecomment-677941174)

### Proposal

Make it possible to edit resource's options. This would allow me to load an extension that defines a resource and then change that resource's options, without having to edit the extension and get a PR merged.

### Implementations

The main options that come to my mind are:

* Make all `*_resource` parameters optional
* Expose resource settings as starlark values
* Do something like the golang [functional options pattern](https://www.sohamkamani.com/golang/options-pattern/) and have builtins for resource options and a builtin for adding options to a resource

#### Make all `*_resource` arguments optional

i.e., change their type to `Union[T, None]`.
When a `_resource` function is called, if there are existing settings, merge any non-`None` args with those existing settings.

Cons:
* This feels like it'd make the API pretty messy and complicated.

#### Expose resource settings as starlark values

e.g., `resource_options['servantes'].objects.append('foobar')`

Cons:
* We don't currently have a way to track usage of starlark values, so we'd lose visibility into which options are being used.

#### Functional Options Pattern

e.g., `resource_options['servantes'].options.append(resource_objects('foobar'))

Cons:
* feels verbose. might be fixable.

Note: As written, of the three only the "starlark values" one allows modifying the values of existing options (e.g., add to the list of objects in a k8s resource or mutate a local_resource's cmd, rather than replace)

KOGI:
Personally, (without knowing what affects it would have on the backend), I like option #1 the best. From the user's perspective, it definitely feels the cleanest and most understandable.

If you're worried about it blowing out the complexity of the API, maybe there could be a separate function for amending resources after they're created:
```python
k8s_resource('foo')
amend_resource('foo', objects=[...], new_name='bar', ...)
```

KOGI:
Any movement on this at all? Thoughts about path forward?

nicks:
I don't see any good reason why k8s_resource can't be additive

k8s_resource('foo', objects=['object1'])
[...]

PR #4766, opened 2021-07-15
tiltfile: dc_resource and k8s_resource should both be additive

nicks:
Hello @hyu, @landism,

dd8b1b13744e5000f01760e28ca4b5d34f51bc87 (2021-07-15 15:59:39 -0400)
tiltfile: dc_resource and k8s_resource should both be additive
Before this change,
- If you had 2 k8s_resource calls for the same resource, it would be an error.
- If you had 2 dc_resource calls for the same resource, some properties
  would merge and some would overwrite.

This behavior was confusing.

This changes both functions to "aggregate" options.

Fixes https://github.com/tilt-dev/tilt/issues/4736
Fixes https://github.com/tilt-dev/tilt/issues/3738

I think that has much better semantics in a world with extensions that want to
"add" things to a resource, and is a good stopgap as we add more things
to the API server.

### DIFF
--- a/internal/tiltfile/k8s.go
+++ b/internal/tiltfile/k8s.go
@@ -71,7 +71,6 @@ type k8sResourceOptions struct {
 	extraPodSelectors []labels.Set
 	triggerMode       triggerMode
 	autoInit          bool
-	tiltfilePosition  syntax.Position
 	resourceDeps      []string
 	objects           []string
 	manuallyGrouped   bool
@@ -285,14 +284,18 @@ func (s *tiltfileState) k8sResource(thread *starlark.Thread, fn *starlark.Builti
 	if !ok {
 		// Set default options.
 		o.autoInit = true
-		o.tiltfilePosition = thread.CallFrame(1).Pos
 	}
 
-	manuallyGrouped := o.manuallyGrouped
-	if workload == "" {
+	newManualGroup := false
+	if workload == "" && !o.manuallyGrouped {
 		resourceName = newName.String()
 		// If a resource doesn't specify an existing workload then it needs to have objects to be valid
-		manuallyGrouped = true
+		newManualGroup = true
+		o.manuallyGrouped = true
+	}
+
+	if resourceName == "" {
+		return nil, fmt.Errorf("Resource name missing. Must give a name for an existing resource or a new_name to create a new resource.")
 	}
 
 	portForwards, err := convertPortForwards(portForwardsVal)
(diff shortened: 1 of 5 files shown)


internal/tiltfile/k8s.go
```diff
@@ -281,7 +281,14 @@ func (s *tiltfileState) k8sResource(thread *starlark.Thread, fn *starlark.Builti
 	}
 
 	resourceName := workload.String()
-	manuallyGrouped := false
+	o, ok := s.k8sResourceOptions[resourceName]
+	if !ok {
+		// Set default options.
+		o.autoInit = true
+		o.tiltfilePosition = thread.CallFrame(1).Pos
```
landism, line 288: I think the only use we still have for this field is the error when k8s_resource specifies a `new_name` that's already taken: https://github.com/tilt-dev/tilt/blob/d06c6881a9a614c3814212e71c8d60dbfd9e91be/internal/tiltfile/tiltfile_state.go#L706

For this error message, we don't want the first k8s_resource for the resource, but the one that specifies `new_name`. What do you think of:
1. Delete other uses of this field
2. Rename this field to something like `newNameTiltfilePosition`
3. Set this on every call, if newName is set

(or maybe however we solve the previous comment makes this all irrelevant)
(or just remove the field entirely - it's kind of an annoying amount of bookkeeping for what's probably not much value)

internal/tiltfile/k8s.go
```diff
@@ -281,7 +281,14 @@ func (s *tiltfileState) k8sResource(thread *starlark.Thread, fn *starlark.Builti
 	}
 
 	resourceName := workload.String()
-	manuallyGrouped := false
+	o, ok := s.k8sResourceOptions[resourceName]
```
landism, line 284: Here are a couple bugs to mull:
```
k8s_resource(new_name='a', objects=['config'])
# generates: Error: k8s_resource has only non-workload objects but doesn't provide a new_name
k8s_resource('a', objects=['config2'])
```
```
k8s_resource('a', objects=['config2'])
# doesn't generate an error, but the resulting resource has 'config2' and not 'config'
k8s_resource(new_name='a', objects=['config2'])
```

Relatedly, we should figure out what semantics we want when a call to `k8s_resource` sets a `new_name` - do subsequent calls to `k8s_resource` access it by its workload name or its `new_name`? Right now, `new_name` doesn't really take effect until assembly (except in the case of workload-less resources, as in the above bugs).

landism (changes requested):

internal/tiltfile/k8s.go
nicks: ya, i think this bookkeeping isn't work keeping. removed!

internal/tiltfile/k8s.go
nicks: Added test cases for both of those!

ya, i think this question speaks a lot to the workshop Han is working on, and making sure we have good nudges to push people in the right direction when they get the name wrong.

landism (approved):

ebenoist:
Thank you for doing this! I'll rollback my changes!

PR #4767, opened 2021-07-16
tiltfile: add support for labels in *_resource functions

lizzthabet:
Thanks @landism for driving this work!

Closes [ch12256](https://app.clubhouse.io/windmill/story/12256/add-tiltfile-support-of-labels-as-argument-to-resource-functions)


internal/tiltfile/value/label.go
```diff
@@ -0,0 +1,87 @@
...
+
+type LabelValue string
+
+func (lv *LabelValue) Unpack(v starlark.Value) error {
+	str, ok := AsString(v)
+	if !ok {
+		return fmt.Errorf("Value should be convertible to string, but is type %s", v.Type())
+	}
+
+	validationErrors := validation.IsQualifiedName(str)
```
nicks, line 19: should this use IsValidLabelValue?

the way you've defined labels, they need to be both valid label names and valid label values

internal/engine/configs/api.go
```diff
@@ -205,6 +207,32 @@ func toCmdObjects(tlr tiltfile.TiltfileLoadResult) typedObjectSet {
...
 }
 
+// Pulls out all the UIResource objects generated by the Tiltfile.
+func toUIResourceObjects(tlr tiltfile.TiltfileLoadResult) typedObjectSet {
+	result := typedObjectSet{}
+
+	for _, m := range tlr.Manifests {
+		name := string(m.Name)
+		labels := make(map[string]string)
+		for _, l := range m.Labels {
```
nicks, line 217: it seems odd for the tiltfile to use two different data types as the source and the destination of labels - seems likely to lead to mistakes

lizzthabet: do you have a suggestion for how to handle the source type vs the destination type? the source type is only relevant in the context of the tiltfile, but we do the same data transformation in the `webview` and `api`.

nicks: i mean that the tiltfile tracks labels as []string, but they're ultimately sent as map[string]string. Let's just use a map everywhere

lizzthabet: gotcha! i paired with matt to move the mapping logic and typing into the starlark unpacking func, so there aren't two data types for labels. does that address your concern?

nicks (approved):

internal/engine/configs/api.go
```diff
@@ -205,6 +207,29 @@ func toCmdObjects(tlr tiltfile.TiltfileLoadResult) typedObjectSet {
...
 }
 
+// Pulls out all the UIResource objects generated by the Tiltfile.
+func toUIResourceObjects(tlr tiltfile.TiltfileLoadResult) typedObjectSet {
+	result := typedObjectSet{}
+
+	for _, m := range tlr.Manifests {
+		name := string(m.Name)
+		labels := m.Labels
+		labels[LabelOwnerKind] = LabelOwnerKindTiltfile // TODO(lizz): move this label assignment to the Tiltfile
```
nicks, line 217: CI is complaining that you're trying to assign to a nil map here. what you probably need to do is make sure that all Manifests have non-nil label maps when you create them.

pkg/model/manifest.go
```diff
@@ -222,6 +224,14 @@ func (m Manifest) LocalPaths() []string {
 	return sliceutils.DedupedAndSorted(paths)
 }
 
+func (m Manifest) WithLabels(labels map[string]string) Manifest {
```
nicks, line 227: i think what you probably want to do is call this function from tiltfile_state.go, instead of assigning Labels directly

lizzthabet: yes! this should be fixed now. thanks for your help. 🎉

Issue #4577, opened 2021-05-21
Live update without injecting an image


ninjaprox:
## Describe the Feature You Want

Allow to config `imagePullPolicy` whenever Tilt's managing an image build

### Current Behavior

`imagePullPolicy` is always set to `IfNotPresent`

### Why Do You Want This?

I have a development version of an image, e.g. `example.com/my-app:dev` and use `sync_file_only` extension to sync local changes to the container and hot reload the app with new changes in the container without having to restart the pod. `sync_file_only` does the trick with `custom_build` (it actually doesn't build anything) for this, which causes Tilt to override `imagePullPolicy` to `IfNotPresent`. When I have some changes in `example.com/my-app:dev`, the container doesn't pull the latest version due to `imagePullPolicy: NotIfPresent`.

I read through the discussion in https://github.com/tilt-dev/tilt/issues/3209. It looks like `Always` is fast enough to use (I tested with a custom Tilt build). According to [this documentation](https://kubernetes.io/docs/concepts/configuration/overview/#container-images), `Always` doesn't pull the image if the image digest unchanged.

Therefore I'd suggest having the ability to config `imagePullPolicy` or not overriding it to `IfNotPresent`.

landism:
Thanks for the report!

It looks like:
1. The current behavior was predicated [at least partially](https://github.com/tilt-dev/tilt/issues/3209#issuecomment-614064746) on the understanding that `imagePullPolicy: Always` meant it'd always pull. As you point out, the documentation now says that it is not the case (the docs were updated after that comment - no idea when/if the actually behavior change happened)
2. There is a separate discussion of whether [using ":latest" in dev is a good idea](https://github.com/tilt-dev/tilt/issues/3209#issuecomment-616801423). As Nick says, teams generally try to keep their builds hermetic and avoid specifying unpinned versions of any dependencies in their build configs. As you may be aware, at some point these dependencies will have non-backwards-compatible changes, which causes all kinds of annoying problems such as breaking devs' environments until they update, making CI runs non-reproducible, and making it hard/impractical to bisect or check behavior at an old commit when tracking down the source of a behavior change.

Of course, all decisions are tradeoffs, and it's also annoying to specify the image digest in your config in git and have it automatically bumped when new images are pushed. I just wanted to make sure to bring that up since the discussion here had only covered the speed side of things.

I think we'll need to look into this and talk about it a bit.

[...]

PR #4768, opened 2021-07-16
engine: introduce a system for live-updating images we didn't deploy

nicks:
Hello @landism,

Please review the following commits I made in branch nicks/ch12386:

7b1dbae447968c04f5009b6905b8078e1f3fb0e9 (2021-07-16 14:42:19 -0400)
engine: introduce a system for live-updating images we didn't deploy
This is the first step towards a general-purpose live-update API
that's independent of a general-purpose image build API, and suggests
how they might interoperate.

Fixes https://github.com/tilt-dev/tilt/issues/4577

I played around with a couple different ways to test this effectively,
but ultimately, the most valuable test was the file_sync_only
integration test we already have. Almost all of our live-update
specs are ultimately designed to catch bugs in how we do the live-update,
rather than asserting things about the live-update spec itself.

### DIFF
--- a/pkg/model/image_target.go
+++ b/pkg/model/image_target.go
@@ -3,18 +3,44 @@ package model
 import (
 	"fmt"
 
+	"github.com/docker/distribution/reference"
+	"github.com/google/go-cmp/cmp"
+
 	"github.com/tilt-dev/tilt/internal/container"
 	"github.com/tilt-dev/tilt/internal/sliceutils"
 	"github.com/tilt-dev/tilt/pkg/apis"
 	"github.com/tilt-dev/tilt/pkg/apis/core/v1alpha1"
 )
 
+type LiveUpdateSpec struct {
+	// A KubernetesDiscovery maps deployed objects to pods.
+	//
+	// The live-updater watches the named KubernetesDiscovery object to find pods to
+	// live update into.
+	KubernetesDiscoveryName string
+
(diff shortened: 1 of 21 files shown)


pkg/model/image_target.go
```diff
@@ -3,19 +3,45 @@ package model
...
 	BuildDetails BuildDetails
 
+	// In a live-update-only image, we don't inject the image into the Kubernetes
+	// deploy, we only live-update to the deployed object. See this issue:
+	//
+	// https://github.com/tilt-dev/tilt/issues/4577
+	//
+	// This is a hacky way to model this right now until we
+	// firm up how images work in the apiserver.
+	IsLiveUpdateOnly bool
```
landism, line 43: Doesn't matter much since this is a hack, but is this already represented by `BuildDetails == nil`? (which maybe we should rename to `BuildSpec`) That seems like it'd make the "some targets have live update but not image build and vice versa" thing represented a bit more symmetrically.

landism (approved):
well that's nice and simple!

pkg/model/image_target.go
nicks: nope, it can't be nil, because the BuildDetails contains the Live Update steps.

But yeah, this isn't how I would model it from the get-go, but this data model is going to get thrown away anyway so it doesn't really matter.

PR #4773, opened 2021-07-19
tiltfile: pass TILT_HOST + TILT_PORT to local() commands

### LINKED ISSUE #4737: Invoking `tilt` from `local` breaks with custom port
### Expected Behavior
* Calling `local()` to shell out to Tilt (e.g. to interact with API) should work if Tilt was invoked with `tilt up --port=10351`

### Current Behavior
* No way to determine port to pass-through to `local()` call

### Steps to Reproduce

1. Add `local('%s get session Tiltfile' % (sys.executable,))` to `Tiltfile`
2. Ensure no instances of Tilt running on default port (10350)
3. Run `tilt up --port=10351` (or any non-standard port)

### Context
Per @nicks we already pass some extra environment variables through to `local()` calls and Tilt understands the `TILT_PORT` environment variable as-is, so the easiest solution for now is to just populate this explicitly.

milas:
Various Tilt extensions have been using `local()` to shell back
out to Tilt to use the API (e.g. `tilt apply -f ...`). This does
not play nice with custom ports if Tilt was invoked using the
`--port` argument.

Now, the current Tilt instance's port is passed as the `TILT_PORT`
argument so that the invoked command will automatically connect to
the same instance of Tilt.

Fixes #4737.

### DIFF
--- a/internal/tiltfile/files.go
+++ b/internal/tiltfile/files.go
@@ -8,6 +8,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"strconv"
 	"strings"
 
 	"github.com/tilt-dev/tilt/internal/k8s"
@@ -82,16 +83,9 @@ func (s *tiltfileState) execLocalCmd(t *starlark.Thread, cmd model.Cmd, logOutpu
 
 	// if Tilt was invoked with `tilt up --port=XXXXX`, local() calls to use the Tilt API will fail due to trying to
 	// connect to the default port, so explicitly populate the TILT_PORT environment variable if it isn't already
-	hasTiltPort := false
-	for _, e := range c.Env {
-		if strings.HasPrefix(e, "TILT_PORT") {
-			hasTiltPort = true
-			break
-		}
-	}
-	if !hasTiltPort {
-		c.Env = append(c.Env, fmt.Sprintf("TILT_PORT=%d", s.webPort))
-	}
+	addEnvIfNotPresent(c, "TILT_PORT", strconv.Itoa(int(s.webPort)))
+	// some Tilt commands, such as `tilt dump engine`, also require the host
+	addEnvIfNotPresent(c, "TILT_HOST", string(s.webHost))
 
 	err = c.Run()
 	if err != nil {
@@ -315,4 +309,16 @@ func getHelmCRDs(path string) ([]string, error) {
 func hasYAMLExtension(fname string) bool {
 	ext := filepath.Ext(fname)
 	return strings.EqualFold(ext, ".yaml") || strings.EqualFold(ext, ".yml")
+}
+
+func addEnvIfNotPresent(c *exec.Cmd, key, value string) bool {
+	prefix := key + "="
+	for _, e := range c.Env {
+		if strings.HasPrefix(e, prefix) {
+			return false
+		}
+	}
+
+	c.Env = append(c.Env, key+"="+value)
+	return true
 }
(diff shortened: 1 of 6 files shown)


internal/tiltfile/files.go
```diff
@@ -80,6 +80,19 @@ func (s *tiltfileState) execLocalCmd(t *starlark.Thread, cmd model.Cmd, logOutpu
 		c.Stderr = io.MultiWriter(stderr, logOutput)
 	}
 
+	// if Tilt was invoked with `tilt up --port=XXXXX`, local() calls to use the Tilt API will fail due to trying to
+	// connect to the default port, so explicitly populate the TILT_PORT environment variable if it isn't already
+	hasTiltPort := false
+	for _, e := range c.Env {
+		if strings.HasPrefix(e, "TILT_PORT") {
```
nicks, line 87: should this be HasPrefix(e, "TILT_PORT=") ? note the '='

internal/tiltfile/files.go
```diff
@@ -80,6 +80,19 @@ func (s *tiltfileState) execLocalCmd(t *starlark.Thread, cmd model.Cmd, logOutpu
 		c.Stderr = io.MultiWriter(stderr, logOutput)
 	}
 
+	// if Tilt was invoked with `tilt up --port=XXXXX`, local() calls to use the Tilt API will fail due to trying to
+	// connect to the default port, so explicitly populate the TILT_PORT environment variable if it isn't already
+	hasTiltPort := false
+	for _, e := range c.Env {
```
nicks, line 86: it might make sense to do this in PrepareEnv with the other Env injections?
https://github.com/tilt-dev/tilt/blob/master/pkg/logger/env.go#L19

though I guess you'd have do pass webport to that function...

nicks (approved):

internal/tiltfile/files.go
```diff
@@ -80,6 +80,19 @@ func (s *tiltfileState) execLocalCmd(t *starlark.Thread, cmd model.Cmd, logOutpu
...
+		if strings.HasPrefix(e, "TILT_PORT") {
+			hasTiltPort = true
+			break
+		}
+	}
+	if !hasTiltPort {
+		c.Env = append(c.Env, fmt.Sprintf("TILT_PORT=%d", s.webPort))
```
landism, line 93: would it be annoying to do TILT_HOST as well?

landism (approved):

internal/tiltfile/files.go
milas: Not really - we've already got `webHost` available on the Tiltfile state (for links display).

I'll add it so that `tilt dump engine` and friends work too.

milas: Added in f3d9caa

internal/tiltfile/files.go
milas: Yeah I thought about that but it seemed weird since that's `logger.PrepareEnv` and pretty specific to xterm type stuff.

I (_very_ briefly) looked at doing it in `Execer` and having `local()` use that but `Execer` is way more oriented to use for the BaDer for daemons and it seemed more headache/risky than it was worth. I think maybe the right thing in the future is to replace the pure `localexec.ExecCmd` with a struct that does the preparation so that it can hold state of things to inject (bonus: you also wouldn't need to pass a logger every time).

internal/tiltfile/files.go
milas: Good call - tweaked that when making it more generic to handle `TILT_HOST` so it'll need to be `${key}=` in f3d9caa

PR #4777, opened 2021-07-20
docker-compose: use compose-go to parse Compose YAML

milas:
Docker has formalized the Compose spec and has the `compose-go`
library to parse it. This is what the new v2 Compose CLI uses
itself (being activated on Docker for Mac as of the latest
releases).

Schema-wise, it is capable of parsing both v2 and v3 versions
of the spec as used by the `docker-compose` v1 CLI. That's all
we need it for!

The only notable behavior change is that the service iteration
order, while still respecting dependencies, is NOT influenced
by order within the YAML. This same change impacts the Compose
v2 CLI; in fact, before these changes, Tilt would be broken with
it as `docker-compose config --services` (what it was using)
is totally random on Compose v2! So, while this is a behavior
change, relying on YAML ordering was never a guarantee of Compose.

There was also a bug in Tilt's manual parsing with port mappings
that used a randomized host port, e.g. `3000/tcp`, where it
treated that as `3000:3000/tcp`, which is NOT the case. The only
impact of this was that previously we'd erroneously have shown a
non-functional endpoint link in the UI, which will now no longer
show.

As a bonus, this should speed up parsing operations that involve
Docker Compose, as we now do everything from native Go code instead
of shelling out 2x to Python!

But mainly, this will ensure that Tilt consistently parses the
Compose YAML (regardless of Docker Compose version), which can avoid
bugs like the above but also means unit tests won't start randomly
failing when Compose slightly changes the YAML output format (and
also leaving us in a weird situation where either every dev + CI
needs to upgrade or remain on an old version).

Tilt still shells out to the Compose CLI for other operations such
as streaming logs, as these are not part of the `compose-go` "spec"
library, and the `compose-cli` project is realistically not a
dependency we want :)


milas:
I got frustrated by tests breaking again after upgrading Docker for Mac and this turned out to not be a lot of work and should hopefully make things more robust. (See #4775 for a real world instance where our manual Compose parsing was causing trouble!)

internal/tiltfile/docker_compose_test.go
```diff
@@ -20,24 +20,24 @@ func TestParseConfigPreservesServiceOrder(t *testing.T) {
...
-e
-d
-c
-b
-a`
-	services := f.parse(output, servicesOutput)
+    image: imgf
+`
+
+	services := f.parse(output)
```
milas, line 40: Implicit YAML ordering was never a guarantee of Compose - it implicitly worked in v1 due to Python parsing, but is totally gone in v2 -- current Tilt + Compose v2 is actually broken in this respect right now.

Tilt will launch services in the same order as Compose v2 now, which will respect dependencies but otherwise be alphabetical.

nicks (approved):
nice!

you might be able to remove some tests from the skiplargetiltfiletests tag, since what was making those tests slow was invoking docker-compose

milas:
> you might be able to remove some tests from the `skiplargetiltfiletests` tag

Good call - the `internal/tiltfile/tiltfile_docker_compose_test.go` tests take ~14sec on average for me on `master` and  ~600ms after the swap to `compose-go` 🎉  (cc @landism - iirc you hate those slow tests 😛)

landism (approved):
nice!

Issue #4751, opened 2021-07-09
probe never reaches or communicates error state


nicks:
Report:
> when a check fails forever, the resource just hangs giving the user no feedback that a readiness check has continued to fail or how to fix it. What would be great is we could have an overall timeout value on the resource getting to "success". If it does not reach success in that time, it would mark the resource failed and give the user an error about the check that was being run.

This seems to directly contradict the docs, which say that the probe has failure thresholds:
https://docs.tilt.dev/api.html#api.probe 

This also seems inconsistent with Kubernetes probes, which are supposed to mark the pod as Unready:
https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#configure-probes

**Additional context**
As I recall, there was a long bit of discussion about the proper behavior of probes, because users complained that they were too chatty:
https://github.com/tilt-dev/tilt/issues/4146

That issue says that they're supposed to communicate when the status changes...

I'm not totally sure what parts of this issue are working as intended or if there are parts of how probes work that need to be rethought or exposed in a different way.

scottwlyft:
The doc says

> failure_threshold ( int ) – Minimum number of consecutive failures for the result to be considered failing **after having succeeded** (minimum value is 1).

My reading of this is the check is continually made after the resource starts successfully? Then if it fails this number of times it will then go into a "failed" state?

nicks:
@scottwlyft yep, that's right

nicks:
oh wait maybe that's not right? it seems inconsistent with the kubernetes probe documentation

milas:
Currently, `local_resource` only has _readiness_ probes, so they only affect whether the runtime state is "not ready" (shown as pending/pulsing gray in UI) or "ready" (green). Within Tilt, K8s readiness probes behave identically.

(We don't currently have anything like a liveness probe for local resources, e.g. it'll never restart due to failure like K8s does to containers that fail liveness probe checks.)

In terms of logging: currently, _all_ `local_resource` readiness probe logging only shows if you're running with `--verbose`. I'll open a PR in a bit to improve that so that transitions from success -> failure show up at a warn level (and failure -> success @ info) so that it's clear when there are important events without totally flooding the logs during normal development. 

[...]

PR #4780, opened 2021-07-21
probe: log transitions at higher level

milas:
This does not change any of the behavior/logic for probes.

Previously, ALL probe activity was logged at `Verbose`. Now,
transitions are logged at `Warn` (if transitioning TO failure)
or `Info` (if transitioning TO success). Sequential failures
continue to be logged at `Verbose` while sequential successes
continue to be suppressed.

See #4751 for inspiration (this doesn't fully resolve the concern
raised there, but helps make failures more visible without being
too spammy.)

~⚠️ Draft until tilt-dev/probe#17 is merged and `go.mod` updated and re-vendored here.~


internal/controllers/core/cmd/controller.go
```diff
@@ -355,37 +353,56 @@ func (c *Controller) runInternal(ctx context.Context,
...
 	l := logger.Get(ctx)
-	if !l.Level().ShouldDisplay(logger.VerboseLvl) {
+	if level == logger.NoneLvl || !l.Level().ShouldDisplay(level) {
 		return
 	}
 
+	w := l.Writer(level)
 	if err != nil {
-		l.Verbosef("[readiness probe error] %v", err)
+		_, _ = io.WriteString(w, fmt.Sprintf("[readiness probe error] %v\n", err))
```
nicks, line 404: fwiw, i would usually use fmt.Fprintf for this

nicks (approved):

PR #4781, opened 2021-07-21
web: remove test and non-test default resource grouping

lizzthabet:
This PR removes the test / non-test resource grouping and filtering, since that delineating will be replaced with labels.

If we want to keep the test / non-test filters code, but turn it off instead of deleting it, I can do that instead.

Closes [ch12251](https://app.clubhouse.io/windmill/story/12251/hide-non-test-test-resource-code-throughout-ui-with-labels-feature-flag)

### DIFF
--- a/web/src/OverviewSidebarOptions.tsx
+++ b/web/src/OverviewSidebarOptions.tsx
@@ -127,7 +127,6 @@ export const ClearResourceNameFilterButton = styled(InstrumentedButton)`
 `
 
 type OverviewSidebarOptionsProps = {
-  showFilters: boolean
   options: SidebarOptions
   setOptions: Dispatch<SetStateAction<SidebarOptions>>
 }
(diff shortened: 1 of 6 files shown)


nicks (approved):

web/src/OverviewSidebarOptions.tsx
```diff
@@ -260,13 +205,10 @@ export function OverviewSidebarOptions(
...
     <OverviewSidebarOptionsRoot>
-      <OverviewSidebarOptionsButtonsRoot
-        className={!props.showFilters ? "is-filterButtonsHidden" : ""}
-      >
-        {props.showFilters ? filterOptions(props) : null}
+      <OverviewSidebarOptionsButtonsRoot className="is-filterButtonsHidden">
         <AlertsOnTopToggle
           className={props.options.alertsOnTop ? "is-enabled" : ""}
-          onClick={(e) => setAlertsOnTop(props, !props.options.alertsOnTop)}
+          onClick={(_e) => setAlertsOnTop(props, !props.options.alertsOnTop)}
```
landism, line 211: nit: I think it's more common in our codebase to do `(_) =>` or just `() =>`, but I don't know if there's a wider JS convention

landism (approved):

web/src/OverviewSidebarOptions.tsx
lizzthabet: that is good to know! i think the wider convention that i'm used to is putting an `_` before variable names that are not used (but available).

PR #4782, opened 2021-07-21
store: stop retrying subscriber notifications on shutdown

landism:
repro: `go test ./internal/engine -run TestConfigChange_ManifestWithPendingChangesBuildsIfTriggerModeChangedToAuto -count 100`

Sometimes on failures, I've gotten 100s of these in my output:
```
Problem processing change. Backing off 15s. Error: context canceled
Problem processing change. Backing off 15s. Error: context canceled
Problem processing change. Backing off 15s. Error: context canceled
Problem processing change. Backing off 15s. Error: context canceled
Problem processing change. Backing off 15s. Error: context canceled
Problem processing change. Backing off 15s. Error: context canceled
Problem processing change. Backing off 15s. Error: context canceled
Problem processing change. Backing off 15s. Error: context canceled
Problem processing change. Backing off 15s. Error: context canceled
Problem processing change. Backing off 15s. Error: context canceled
Problem processing change. Backing off 15s. Error: context canceled
Problem processing change. Backing off 15s. Error: context canceled
Problem processing change. Backing off 15s. Error: context canceled
Problem processing change. Backing off 15s. Error: context canceled
Problem processing change. Backing off 15s. Error: context canceled
```

It seems reasonable to skip retrying when the context is done.

### DIFF
--- a/internal/store/subscriber.go
+++ b/internal/store/subscriber.go
@@ -188,6 +188,13 @@ func (e *subscriberEntry) notify(ctx context.Context, store *Store) {
 		return
 	}
 
+	select {
+	case <-ctx.Done():
+		// don't keep retrying after context is done
+		return
+	default:
+	}
+
 	// Backoff on error
 	// TODO(nick): Include the subscriber name in the error message.
 	backoff := activeChange.LastBackoff * 2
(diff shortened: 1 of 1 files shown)


internal/store/subscriber.go
```diff
@@ -188,6 +188,13 @@ func (e *subscriberEntry) notify(ctx context.Context, store *Store) {
 		return
 	}
 
+	select {
+	case <-ctx.Done():
```
nicks, line 192: fwiw, i usually do this as 

```
if ctx.Err() != nil {
  // context finished
}
```

nicks (approved):

PR #4784, opened 2021-07-21
engine: prepare for multiple tiltfiles

nicks:
Hello @landism, @milas,

fdbf28c3123ed818d1435d407233dd1bddf1bf5d (2021-07-21 19:59:24 -0400)
engine: prepare for multiple tiltfiles

### DIFF
--- a/internal/engine/upper.go
+++ b/internal/engine/upper.go
@@ -487,14 +487,18 @@ func handleConfigsReloadStarted(
 	state *store.EngineState,
 	event configs.ConfigsReloadStartedAction,
 ) {
+	ms, ok := state.TiltfileStates[event.Name]
+	if !ok {
+		return
+	}
+
 	status := model.BuildRecord{
 		StartTime: event.StartTime,
 		Reason:    event.Reason,
 		Edits:     event.FilesChanged,
 		SpanID:    event.SpanID,
 	}
-
-	state.TiltfileState.CurrentBuild = status
+	ms.CurrentBuild = status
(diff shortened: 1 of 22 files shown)


nicks:
there are still some ideas i want to play around with on https://github.com/tilt-dev/tilt.specs/pull/30, but i think the general functionality of having config loaders that execute independently is a good thing to have.

internal/engine/upper.go
```diff
@@ -511,10 +515,14 @@ func handleConfigsReloaded(
 		manifestNames[m.Name] = true
 	}
 
-	b := state.TiltfileState.CurrentBuild
```
landism, line 514: I wish github let you comment outside of changed lines...

Some things in this function that I think will be buggy with multiple Tiltfiles:
1. If an extension loads a Tiltignore, it'll replace the main Tiltfile's Tiltignore (but also maybe it looks like `state.Tiltignore` could be deleted?)
2. The GC that happens with `newDefOrder` looks like it'll be broken unless we give `m.Source` a different value per Tiltfile
3. `state.Features` and `state.ConfigFiles` should probably be additive across Tiltfiles (they're already additive in the `err != nil` case) (also I think we can remove `state.ConfigFiles`)
4. Maybe TelemetrySettings, VersionSettings, et al should also be additive, but that seems a lot hairier

landism (approved):
LGTM!

I don't know what scope you want on this PR vs the next. TODOing issues is fine since these changes won't cause problems until it's actually possible to add a second Tiltfile.

Issue #4472, opened 2021-04-27
clear logs does not clear warnings


nicks:
### Steps to Reproduce
1. Check out https://github.com/tilt-dev/tilt.build
2. Run tilt up
3. Kill one of the pods so it restarts
4. Observe that the resource status turns yellow
5. Click clear logs

### Expected Behavior
The resource status should be green

### Current Behavior
The resource status remains yellow

#### About Your Use Case
I feel like this used to be something we handled well, but dropped it on the move to the new UI

andymartin-sch:
just want to bump this issue once :)

the combination of this + https://github.com/tilt-dev/tilt/issues/3483 means that "normal"/successful use of Tilt results in a few warning/yellow colored services in the left sidebar and there's nothing that the user can do to get the color back to green once the warnings are resolved / no longer an issue

nicks:
thanks @andymartin-sch ! let me poke at this a bit. we've done some major revamps to how the tilt ui works that might make this easier

PR #4802, opened 2021-07-27
web: generate error and warning counts from logstore

nicks:
Hello @lizzthabet, @milas,

Please review the following commits I made in branch nicks/ch11942:

c2ecba0a5a25ac9048029b2107bedff1792461f3 (2021-07-27 16:47:08 -0400)
web: generate error and warning counts from logstore
Fixes https://github.com/tilt-dev/tilt/issues/4472


nicks:
There wasn't an easy way to do this incrementally. :\

Before this change, we counted errors and warnings by digging around in the metadata and deciding client-side if there was an error or warning.

After this change, we just count the number of logs with level = WARN or level = ERROR

This should overall be much simpler, and also make the "Clear Logs" button behave more like people expect.

internal/hud/webview/convert.go
```diff
@@ -6,23 +6,20 @@ import (
...
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	ctrlclient "sigs.k8s.io/controller-runtime/pkg/client"
 
 	"github.com/tilt-dev/tilt/internal/cloud/cloudurl"
+	"github.com/tilt-dev/tilt/internal/k8s"
 	"github.com/tilt-dev/tilt/internal/store"
+	"github.com/tilt-dev/tilt/internal/store/k8sconv"
+	"github.com/tilt-dev/tilt/pkg/apis"
```
milas, line 18: Not important, but did you do some manual re-arranging to get the imports "correctly" sorted? I'm sorta surprised `goimports` hadn't placed them here before, so just wondering if there was magic involved

web/src/LogStore.ts
```diff
@@ -266,6 +280,15 @@ class LogStore {
...
       span.lastLineIndex = this.lines.length
       this.lines.push(candidate)
+
+      // If this starts a warning or error, index it now.
+      let level = newSegment.level
+      if (
+        newSegment.anchor &&
+        (level === LogLevel.WARN || level === LogLevel.ERROR)
+      ) {
+        span.alerts.push({ level: level, lineIndex: this.lines.length })
```
milas, line 290: AFAICT we never use the `lineIndex` so it's not causing problems, but I think it should be `span.lastLineIndex` because the lines have already been pushed to `this.lines`, so this will be the index _after_ the alert, no?

milas (approved):
Nice! I think this is much more straight-forward behavior-wise.

internal/hud/webview/convert.go
nicks: there was some manual reordering :grimacing:

ya, might dig into sometime why goimports doesn't do this on its own

web/src/LogStore.ts
nicks: oops, yes!

you're right it's not used now; it was just much easier to debug when this object had some info about where the alert came from

web/src/LogStore.ts
```diff
@@ -736,5 +760,8 @@ const logStoreContext = React.createContext<LogStore>(new LogStore())
 export function useLogStore(): LogStore {
   return useContext(logStoreContext)
 }
+export function useLogAlertIndex(): LogAlertIndex {
```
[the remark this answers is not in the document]
nicks: Good question! Mostly to help with testing. FakeAlertIndex is way simpler than trying to support the full LogStore API.

I played around with having a separate AlertIndex (that LogStore owns and keeps up-to-date). Making AlertIndex an interface made the overall PR smaller. But it might still make sense to separate them in the future.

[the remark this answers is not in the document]
lizzthabet: ah, ok! i think that makes sense. it might be worth adding a comment above `useLogAlertIndex` to explain why/where you'd use it instead of `useLogStore`.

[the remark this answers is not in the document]
nicks: comments added!

internal/hud/webview/convert.go
landism: I'm a bit late to the party but this has bugged me for a long time: https://github.com/golang/go/issues/20818

FWIW when I notice this in a file I just delete all the blank lines in the import block and then goimports will do the right thing.

PR #4804, opened 2021-07-28
pod rollout: don't show completed jobs as pending

### LINKED ISSUE #3513: pod rollout tracking for Jobs stops at "pending"
When I run the [echo-hi job in servantes](https://github.com/tilt-dev/servantes/blob/master/deploy/job.yaml), the log looks like this when it finishes:

```
STEP 1/1 — Deploying
     Applying via kubectl:
     → echo-hi:job

     Step 1 - 0.21s
     DONE IN: 0.21s 


Tracking new pod rollout (echo-hi-rr7xd):
     ┊ Scheduled       - <1s
     ┊ Initialized     - <1s
     ┊ Ready           - (…) Pending
```

which leaves the impression that it's still waiting on something

`kubectl` says the pod is Completed:
```
$ kc get pod
NAME            READY   STATUS      RESTARTS   AGE
echo-hi-rr7xd   0/1     Completed   0          15s
```

landism:
### Problem
Fixes #3513
When a job's pod finishes, we get a PodCondition that says it's no longer in the ready state. This means we go back and update the pod rollout's "Ready" line in the log to "pending", which is very confusing. This is what the log shows for a completed job:
```
Tracking new pod rollout (echo-hi-rr7xd):
     ┊ Scheduled       - <1s
     ┊ Initialized     - <1s
     ┊ Ready           - (…) Pending
```

### Solution
Treat PodCondition w/ Type="Ready" and Reason="PodCompleted" as meaning we're done with the pod, and munge "completed" into the "ready" field. I don't love this solution except for its cheapness, and am not sure doing something fancier is worth it right now.

### DIFF
--- a/internal/engine/k8srollout/podmonitor.go
+++ b/internal/engine/k8srollout/podmonitor.go
@@ -6,13 +6,12 @@ import (
 	"strings"
 	"time"
 
-	"github.com/tilt-dev/tilt/pkg/apis/core/v1alpha1"
-
 	"github.com/google/go-cmp/cmp"
 	v1 "k8s.io/api/core/v1"
 
 	"github.com/tilt-dev/tilt/internal/k8s"
 	"github.com/tilt-dev/tilt/internal/store"
+	"github.com/tilt-dev/tilt/pkg/apis/core/v1alpha1"
 	"github.com/tilt-dev/tilt/pkg/logger"
 	"github.com/tilt-dev/tilt/pkg/model"
 	"github.com/tilt-dev/tilt/pkg/model/logstore"
@@ -119,6 +118,19 @@ func (m *PodMonitor) printCondition(ctx context.Context, name string, cond v1alp
 	}
 
 	if cond.Status == string(v1.ConditionTrue) {
+		l.Infof("%s┊ %s%s- %s", indent, name, spacer, duration)
+		return
+	}
+
+	// PodConditions unfortunately don't represent Jobs well:
+	// 1) If a Job runs quickly enough, we might never observe the pod in a ready state (i.e., no updates have Type="Ready" and Status="True")
+	// 2) After a Job finishes, we get a pod update with Type="Ready", Status="False", and LastTransitionTime=the job end time,
+	//    meaning that we lose the time at which the pod actually transitioned to the ready state.
+	// Rather than invest in more state to track these for the Job case, let's just replace "Ready" with "Completed"
+	// and reconsider if/when a user cares.
+	if cond.Type == "Ready" && cond.Reason == "PodCompleted" {
+		name = "Completed"
+		spacer = strings.Repeat(" ", spacerMax-len(name))
 		l.Infof("%s┊ %s%s- %s", indent, name, spacer, duration)
 		return
 	}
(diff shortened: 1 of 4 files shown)


landism:
fwiw, here's what the Conditions look like for a completed Job's Pod:
```
        "conditions": [
            {
                "lastProbeTime": null,
                "lastTransitionTime": "2021-07-28T21:30:23Z",
                "reason": "PodCompleted",
                "status": "True",
                "type": "Initialized"
            },
            {
                "lastProbeTime": null,
                "lastTransitionTime": "2021-07-28T21:30:35Z",
                "reason": "PodCompleted",
                "status": "False",
                "type": "Ready"
            },
            {
                "lastProbeTime": null,
                "lastTransitionTime": "2021-07-28T21:30:35Z",
                "reason": "PodCompleted",
                "status": "False",
                "type": "ContainersReady"
            },
            {
                "lastProbeTime": null,
                "lastTransitionTime": "2021-07-28T21:30:23Z",
                "status": "True",
                "type": "PodScheduled"
            }
        ],
```

steve-bitdotio:
question: If a job has completed, should it be considered ready (or should jobs be special cased from other pods)? The job pod exists on completion, so labeling it as "ready" might imply that its still running? Thoughts?

landism:
> question: If a job has completed, should it be considered ready (or should jobs be special cased from other pods)? The job pod exists on completion, so labeling it as "ready" might imply that its still running? Thoughts?

Good question!
Two things:
1. This feature logs the progress of the pod rollout, so it's showing the state transitions as the pod becomes scheduled, initialized, and ready, even after those states are no longer the current state. 
If you see this in your log, do you think it means the pod is necessarily still ready?
```
Tracking new pod rollout (echo-hi-nrck6):
     ┊ Scheduled       - <1s
     ┊ Initialized     - <1s
     ┊ Ready           - 1s
```
2. This PR actually specifically special cases jobs so that after they've completed the rollout tracker shows "completed" instead of "ready". Here's what this PR changes the display of a completed job to:
```
Tracking new pod rollout (echo-hi-nrck6):
     ┊ Scheduled       - <1s
     ┊ Initialized     - <1s
     ┊ Completed       - 1s
```

Does that seem reasonable?

steve-bitdotio:
> 2. This PR actually specifically special cases jobs so that after they've completed the rollout tracker shows "completed" instead of "ready". Here's what this PR changes the display of a completed job to:
> 
> ```
> Tracking new pod rollout (echo-hi-nrck6):
>      ┊ Scheduled       - <1s
>      ┊ Initialized     - <1s
>      ┊ Completed       - 1s
> ```
> 
> Does that seem reasonable?

Ah, I see @landism ! That makes sense to me. I think I just misunderstood your PR description, so thanks for clarifying 😁

nicks (approved):

milas:
> question: If a job has completed, should it be considered ready (or should jobs be special cased from other pods)? The job pod exists on completion, so labeling it as "ready" might imply that its still running? Thoughts?

See #4809 for an issue about improved status reporting (i.e. the icon in the resource sidebar) for `Job`s

PR #4807, opened 2021-07-29
apis: add a ConfigMap type

nicks:
Hello @landism, @milas,

Please review the following commits I made in branch nicks/configmap:

9841b23013c078f43b2eba2d378b11caef5e58f3 (2021-07-29 09:56:23 -0400)
apis: add a ConfigMap type
This is a simplified version of Kubernetes' ConfigMap.
It's a dumping ground for unstructured data to read and write
without necessarily needing do real reconcilation.


pkg/apis/core/v1alpha1/configmap_types.go
```diff
@@ -0,0 +1,108 @@
...
+// +k8s:openapi-gen=true
+type ConfigMap struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty" protobuf:"bytes,1,opt,name=metadata"`
+
+	// Data contains the configuration data.
+	// Each key must consist of alphanumeric characters, '-', '_' or '.'.
+	// Values with non-UTF-8 byte sequences must use the BinaryData field.
+	// The keys stored in Data must not overlap with the keys in
+	// the BinaryData field, this is enforced during validation process.
```
milas, line 49:
```suggestion
```
(Unless you meant to add `BinaryData` as well)

milas (approved):
I don't have a great alternative, but I'm slightly afraid of naming this identically to K8s and causing confusion (when I saw the PR title I briefly thought this was for explicit K8s `ConfigMap` support inside Tilt)

pkg/apis/core/v1alpha1/configmap_types.go
nicks: oops yes

nicks:
re: giving it the same name as Kubernetes' ConfigMap.

ya, I went back and forth on this. I also considered using the "v1" apiVersion instead of tilt.dev/v1alpha1

but i think other tools in this space are also using ConfigMap directly (e.g., https://github.com/kcp-dev/kcp), so I suspect that this is going to end up being a common pattern

PR #4812, opened 2021-07-29
web: add helpful tooltips for trigger mode toggle and regexp log filter

lizzthabet:
This pr adds two informational tooltips for trigger mode and for regexp log filtering, plus helpers for generating Tilt docs links.

Screen Shot 2021-07-29 at 1 46 11 PM
Screen Shot 2021-07-29 at 1 50 12 PM

### DIFF
--- a/web/src/OverviewActionBar.tsx
+++ b/web/src/OverviewActionBar.tsx
@@ -483,14 +483,6 @@ function FilterTermFieldError({ error }: { error: string }) {
   )
 }
 
-const FilterTermInfoTooltip = styled(TiltInfoTooltip)`
-  margin: 0 ${SizeUnit(1 / 4)};
-
-  & .fillStd {
-    fill: ${Color.grayLight};
-  }
-`
-
 const filterTermTooltipContent = (
   <>
     RegExp should be wrapped in forward slashes, is case-insensitive, and is{" "}
@@ -598,7 +590,7 @@ export function FilterTermField({ termFromUrl }: { termFromUrl: FilterTerm }) {
       <SrOnly component="label" htmlFor={FILTER_FIELD_ID}>
         Filter resource logs by text or /regexp/
       </SrOnly>
-      <FilterTermInfoTooltip
+      <TiltInfoTooltip
         id={FILTER_FIELD_TOOLTIP_ID}
         title={filterTermTooltipContent}
         placement="right-end"
--- a/web/src/OverviewActionBar.tsx
+++ b/web/src/OverviewActionBar.tsx
@@ -491,6 +491,19 @@ const FilterTermInfoTooltip = styled(TiltInfoTooltip)`
   }
 `
 
+const filterTermTooltipContent = (
+  <>
+    RegExp should be wrapped in forward slashes, is case-insensitive, and is{" "}
+    <a
+      href="https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions"
+      target="_blank"
+    >
+      parsed in JavaScript
+    </a>
+    .
+  </>
+)
+
 const debounceFilterLogs = debounce((history: History, search: string) => {
   // Record the action for analytics
   incr("ui.web.filterTerm", { action: "edit" })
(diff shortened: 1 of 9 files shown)


web/src/OverviewActionBar.tsx
```diff
@@ -481,6 +483,14 @@ function FilterTermFieldError({ error }: { error: string }) {
   )
 }
 
+const FilterTermInfoTooltip = styled(TiltInfoTooltip)`
+  margin: 0 ${SizeUnit(1 / 4)};
+
+  & .fillStd {
+    fill: ${Color.grayLight};
+  }
```
lizzthabet, line 491: i _think_ there's a way to pass props to a styled component's styles, but i wasn't able to figure out how to do it in a way that made typescript happy. @hyu do you have any suggestions, or is this style override an acceptable solution?

web/src/OverviewActionBar.tsx
```diff
@@ -574,6 +585,11 @@ export function FilterTermField({ termFromUrl }: { termFromUrl: FilterTerm }) {
       <SrOnly component="label" htmlFor={FILTER_FIELD_ID}>
         Filter resource logs by text or /regexp/
       </SrOnly>
+      <FilterTermInfoTooltip
+        id={FILTER_FIELD_TOOLTIP_ID}
+        title="RegExp should be wrapped in forward slashes, is case-insensitive, and parsed in JavaScript."
```
lizzthabet, line 590: @ellenkorbes i'm not exactly sure how to describe the flavor of regexp that we're using. 😅 do you have any suggestions?

web/src/OverviewTable.tsx
```diff
@@ -375,6 +385,30 @@ const columns: Column<RowValues>[] = [
   },
 ]
 
+const columnNameToInfoTooltip: {
+  [key: string]: NonNullable<React.ReactNode>
+} = {
+  "Trigger Mode": (
+    <>
+      Trigger mode can be toggled through the UI. To set it persistantly, see{" "}
```
landism, line 393:
```suggestion
      Trigger mode can be toggled through the UI. To set it persistently, see{" "}
```

web/src/OverviewActionBar.tsx
landism: if you can put a link in the tooltip, linking to a js regexp reference would both be useful and make the specific wording here less important

lizzthabet: done!

web/src/OverviewTable.tsx
lizzthabet: fixed ✔️

hyu (approved):
@lizzthabet — TIL about aria-describedby! 

In response to your question about the style override, I updated a Tilt [thread](https://tilt-dev.slack.com/archives/CD9C65ME0/p1627587564022800?thread_ts=1627581267.021100&cid=CD9C65ME0) with Surbhi and also responded on DM re: general approach to style overrides for shared components.

That said, I Approve this PR since it's a definite improvement! I have a branch with the proposed style change I Slacked about, which I can push. Or you can just merge this one and we can tweak the style later.

web/src/OverviewActionBar.tsx
hyu: (Also DM'd this, but adding here for reference)

Here the style override (i.e., for margin and color) is minor enough that it seems like you could get away with the same default padding, and I think it’s actually better if color is consistent. 

In general, if there’s a shared component, I prefer if you can change the appearance in all instances just by looking in the one component itself. Vs, like, having to check for custom styles applied in each usage.

(i.e., you create a new prop like you did with `size`, and then inside the shared component you add some conditional style when a certain prop is coming through. styled-components does have that native method you mentioned, but usually we just use the prop to fiddle with the className. Like, if you wanted to change the color you might toggle a class like “isHighlighted” or something.)

web/src/OverviewActionBar.tsx
ellenkorbes: this? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions

landism:
Merging on Lizz's behalf since she's out today and we want this in today's release.

web/src/OverviewActionBar.tsx
lizzthabet: yup! that's what i linked to.

PR #4815, opened 2021-07-29
controllers: trigger tiltfile reconcilation from the configs controller

nicks:
Hello @milas, @landism,

ea18baf2f2288d2dd41cfbffe4fe48dbaa22a32c (2021-07-29 14:59:53 -0400)
controllers: trigger tiltfile reconcilation from the configs controller

### DIFF
--- a/internal/controllers/core/tiltfile/buildsource.go
+++ b/internal/controllers/core/tiltfile/buildsource.go
@@ -1,0 +1,69 @@
+package tiltfile
+
+import (
+	"context"
+	"sync"
+
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/util/workqueue"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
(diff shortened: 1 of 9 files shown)


internal/controllers/core/tiltfile/buildsource.go
```diff
@@ -0,0 +1,69 @@
...
+}
+
+func (s *BuildSource) Start(ctx context.Context, handler handler.EventHandler, q workqueue.RateLimitingInterface, ps ...predicate.Predicate) error {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	s.q = q
+	return nil
+}
+
+func (s *BuildSource) Entry() *BuildEntry {
```
milas, line 54: AFAICT this is unused - what's the reason for holding onto the latest vs just having an `AddEntry()` and writing to the queue?

milas (approved):

internal/controllers/core/tiltfile/buildsource.go
nicks: yep - it will be used in the next PR : )

PR #4819, opened 2021-07-30
tiltfile: add exit() built-in to stop execution

### LINKED ISSUE #4501: stop tiltfile execution without exiting
Currently, the Tiltfile has a fail() function that exits immediately with a failure code

For symmetry, it seems like there should be an exit() function (or similar) that is allowed to exit with a success code (e.g., if you've disabled the rest of the tiltfile)

milas:
Currently we have `fatal()` for terminating Tiltfile execution
with an error.

However, there are some circumstances where it's desirable to
stop Tiltfile execution but NOT trigger an error. For example,
when debugging a problem, it's useful to disable everything
subsequent in the Tiltfile.

This adds `exit()` which will stop execution in a "successful"
manner.

An optional message can be provided which will be printed before
exiting.

Closes #4501.

### DIFF
--- a/internal/tiltfile/print/print.go
+++ b/internal/tiltfile/print/print.go
@@ -25,6 +25,10 @@ func (Extension) OnStart(env *starkit.Environment) error {
 	if err != nil {
 		return err
 	}
+	err = env.AddBuiltin("exit", exit)
+	if err != nil {
+		return err
+	}
 	return nil
 }
 
@@ -53,4 +57,23 @@ func warn(thread *starlark.Thread, fn *starlark.Builtin, args starlark.Tuple, kw
 	logger.Get(ctx).Warnf("%s", msg)
 
 	return starlark.None, nil
+}
+
+func exit(thread *starlark.Thread, fn *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
+	var msg string
+	err := starkit.UnpackArgs(thread, fn.Name(), args, kwargs, "msg?", &msg)
+	if err != nil {
+		return nil, err
+	}
+
+	ctx, err := starkit.ContextFromThread(thread)
+	if err != nil {
+		return nil, err
+	}
+
+	if msg != "" {
+		logger.Get(ctx).Infof("%s", msg)
+	}
+
+	return starlark.None, starkit.ErrStopExecution
 }
--- a/internal/tiltfile/print/print.go
+++ b/internal/tiltfile/print/print.go
@@ -60,8 +60,8 @@ func warn(thread *starlark.Thread, fn *starlark.Builtin, args starlark.Tuple, kw
 }
 
 func exit(thread *starlark.Thread, fn *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
-	var msg string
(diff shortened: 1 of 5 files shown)


internal/tiltfile/print/print.go
```diff
@@ -54,3 +58,22 @@ func warn(thread *starlark.Thread, fn *starlark.Builtin, args starlark.Tuple, kw
 
 	return starlark.None, nil
 }
+
+func exit(thread *starlark.Thread, fn *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
+	var msg string
```
nicks, line 63: not sure i feel strongly about this, but the Python version of this function takes an exit code rather than a message

https://docs.python.org/3/library/constants.html#exit

milas:
This has come up as particularly desirable for workshop/tutorial where we want to halt loading of everything but the workshop meta-resource until a condition is met [and not require you to rewrite 99% of your Tiltfile to be in a function]

nicks (approved):

landism:
it's probably worth having a test that calls `exit` from an included Tiltfile

internal/tiltfile/print/print.go
milas: Good point (for some reason I thought it was `quit()` in Python which I now see is an alias).

And it's actually even worse than that!

It passes the code to `SystemExit` which says:
> If the value is an integer, it specifies the system exit status (passed to C’s exit() function); if it is None, the exit status is zero; if it has another type (such as a string), the object’s value is printed and the exit status is one.

```sh
$ python3 -c "exit(123)"; echo $?
123

$ python3 -c "exit('see ya')"; echo $?
see ya
1
```

I wonder if I should just pick a different name entirely (open for suggestions here) - I am afraid making it truly API compatible (i.e. in terms of behavior, not just types) will just cause confusion/frustration since we have `fail` and so you'd expect `exit("foo")` to be the non-error version, but having a built-in that behaves differently is also bad.

Alternatively, could make it accept NO args so you just need to `print` + `exit()`?

nicks: fwiw, i think it would be reasonable to accept any value and print that

re: consistency - I personally think the value of consistency here is not API compatibility, but rather more about exploiting existing conventions and memory (for both search-ability and readability), so that people don't need to constantly be searching around the api docs. So I'd rather avoid inventing a new name.

milas:
> it's probably worth having a test that calls `exit` from an included Tiltfile

Added in d4a8dca. I'll be sure to call this out in the corresponding docs API update, but the behavior is that `exit()` is more like `sys.exit()` than `return` -- so a call in a loaded `Tiltfile` will stop the root Tiltfile from executing as well.

landism (approved):

landism:
> For example,
> which debugging a problem, it's useful to disable everything
> subsequent in the Tiltfile.

yes! I've often gotten annoyed at having to comment out a bunch of crap

internal/tiltfile/print/print.go
milas: Okay, d4a8dca chages it to have an optional param named `code` that accepts `starlark.Value` and it just outputs the Starlark-stringified value (assuming something other than `None` was passed).

Regardless of what you pass, it's not an error, i.e. `exit(1)` will log `1` but WON'T be an error.

PR #4826, opened 2021-08-02
tiltfile: limit scenarios where `[no output]` is shown

milas:
Currently, `[no output]` is ALWAYS logged if a command from
`local()` / `helm()` / `kustomize()` returns no output.

This can be annoying with `local()` if you have a silent command
using `echo_off=True` + `quiet=True` because it'll result in
`[no output]` in the logs with no context.

At the least, this should be tied to `quiet` - if you've already
requested to NOT receive output, it's weird to get output when
there is no output.

This change also goes one step further - if `echo_off=True` and
`quiet=False`, it _also_ omits the `[no output]` because there's
no way to know what ran and didn't output anything since the
command itself wasn't echoed.

### DIFF
--- a/internal/tiltfile/files.go
+++ b/internal/tiltfile/files.go
@@ -25,6 +25,15 @@ import (
 
 const localLogPrefix = " → "
 
+type execCommandOptions struct {
+	// logOutput writes stdout and stderr to logs if true.
+	logOutput bool
+	// logCommand writes the command being executed to logs if true.
+	logCommand bool
+	// logCommandPrefix is a custom prefix before the command (default: "Running: ") used if logCommand is true.
+	logCommandPrefix string
+}
+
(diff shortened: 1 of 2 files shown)


milas:
This was annoying me while working on workshop - it triggers lots of orchestration commands that are silent (e.g. `local('...', quiet=True, echo_off=True)` but you end up with a `(Tiltfile)` that looks like this:

```
Beginning Tiltfile execution
 → [no output]
 → [no output]
 → [no output]
 → [no output]
 → [no output]
 → [no output]
 → uibutton.tilt.dev/Tutorial-workshop-play created
 → cmd.tilt.dev/btn-Tutorial-workshop-play created
 → uibutton.tilt.dev/Tutorial-workshop-replay created
 → cmd.tilt.dev/btn-Tutorial-workshop-replay created
 → uibutton.tilt.dev/Tutorial-workshop-rewind created
 → cmd.tilt.dev/btn-Tutorial-workshop-rewind created
Successfully loaded Tiltfile (765.457439ms)
```

nicks (approved):

PR #4830, opened 2021-08-03
extensionrepo: add a basic reconciler

nicks:
Hello @milas, @landism,

Please review the following commits I made in branch nicks/repo2:

d54d27c14fdd6cd8fe0217fcc3dbd05152510771 (2021-08-03 17:18:11 -0400)
extensionrepo: add a basic reconciler

### DIFF
--- a/internal/controllers/core/extensionrepo/reconciler.go
+++ b/internal/controllers/core/extensionrepo/reconciler.go
@@ -2,18 +2,35 @@ package extensionrepo
 
 import (
 	"context"
+	"fmt"
+	"os"
+	"strings"
+	"time"
 
+	"github.com/tilt-dev/go-get"
+	"github.com/tilt-dev/wmclient/pkg/dirs"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/types"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
 	ctrlclient "sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
+	"github.com/tilt-dev/tilt/internal/controllers/apicmp"
+	"github.com/tilt-dev/tilt/pkg/apis"
 	"github.com/tilt-dev/tilt/pkg/apis/core/v1alpha1"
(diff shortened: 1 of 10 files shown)


internal/controllers/core/extensionrepo/reconciler.go
```diff
@@ -40,9 +65,116 @@ func (r *Reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
...
+
+func getImportPath(repo *v1alpha1.ExtensionRepo) (string, error) {
+	url := repo.Spec.URL
+	if strings.HasPrefix(url, "https://") {
+		return strings.TrimPrefix(url, "https://"), nil
+	}
+	if strings.HasPrefix(url, "http://") {
+		return strings.TrimPrefix(url, "http://"), nil
+	}
+	return "", fmt.Errorf("malformed url: %v", url)
```
milas, line 179: Are you still planning on `file://` for the future?

milas (approved):

internal/controllers/core/extensionrepo/reconciler.go
landism, line 180: Maybe something like "url must start with http:// or https://"? I think it's relatively common for people to (technically incorrectly, but often practically) refer to strings without protocols as urls and the problem might not be obvious from the error message.

nicks: yes!!! I forgot about that. will leave a todo for it

internal/controllers/core/extensionrepo/reconciler.go
```diff
@@ -40,9 +65,117 @@ func (r *Reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
...
+	if err != nil && !os.IsNotExist(err) {
+		return ctrl.Result{}, err
+	}
+
+	// If the destination path does not exist, download it now.
+	if os.IsNotExist(err) {
+		lastFetch := r.fetches[nn]
+		lastBackoff := r.backoffs[nn]
+		if time.Since(lastFetch) < lastBackoff {
+			return ctrl.Result{RequeueAfter: lastBackoff}, nil
```
landism, line 109: should this be `lastBackoff - time.Since(lastFetch)`? or maybe it can just skip the RequeueAfter since there's still theoretically a requeued request coming in from the call that set the backoff in the first place? I guess this probably doesn't matter in practice.

landism (approved):

PR #4833, opened 2021-08-05
uibutton: opens link to trigger mode docs in a new tab

LucasPerso:
This link in the Trigger mode tooltip (shown in picture) is now opening in a new tab.

example


lizzthabet (approved):
thanks, @LucasPerso!

PR #4840, opened 2021-08-06
tiltfile:  Propagate labels and owner references from the parent tiltfile

nicks:
Hello @milas, @landism,

Please review the following commits I made in branch nicks/tiltfileowner:

4eb56cdedaeb09044f3a87212937bc3b07ab85ba (2021-08-06 17:25:54 -0400)
tiltfile:  Propagate labels and owner references from the parent tiltfile


nicks:
i also needed to clean up the UIResource life-cycle, since having two different systems creating UIResources was causing issues

internal/controllers/core/kubernetesapply/reconciler.go
```diff
@@ -142,7 +143,8 @@ func (r *Reconciler) shouldDeployOnReconcile(
 	nn types.NamespacedName,
 	ka *v1alpha1.KubernetesApply,
 	imageMaps map[types.NamespacedName]*v1alpha1.ImageMap) bool {
-	if ka.ObjectMeta.Labels[v1alpha1.LabelOwnerKind] == v1alpha1.LabelOwnerKindTiltfile {
+	owner := metav1.GetControllerOf(ka)
+	if owner != nil && owner.Kind == v1alpha1.OwnerKindTiltfile {
```
milas, line 147: I feel like this is a pretty cool milestone!

internal/controllers/core/tiltfile/api.go
```diff
@@ -47,15 +50,37 @@ type typedObjectSet map[string]object
...
+				return err
+			}
+
+			if len(tf.Spec.Labels) > 0 {
+				labels := obj.GetLabels()
+				if labels == nil {
+					labels = make(map[string]string)
+				}
+				for k, v := range tf.Spec.Labels {
+					labels[k] = v
```
milas, line 71: I don't have a compelling argument on either side - just checking that this is desired behavior to override any overlapping labels?

milas (approved):

internal/controllers/core/kubernetesapply/reconciler.go
nicks: YAS

internal/controllers/core/tiltfile/api.go
nicks: good question! ya, I changed it so that labels from tiltfile execution take precedence. I also don't really have a compelling argument either way, but this seems like more what i'd expect

PR #4842, opened 2021-08-09
cmd: error instead of panic when empty

landism:
### Problem

`Env.ExecCmd` panics if given an empty `Cmd`. This reproduces with `local('')` or `local([])`.

### Solution

Have it return an `error` instead.

We could instead check for empty at the starlark builtin level, but it seems good to also avoid panics if someone, e.g. creates a `Cmd` object with an empty argv.

### DIFF
--- a/internal/controllers/core/cmd/execer.go
+++ b/internal/controllers/core/cmd/execer.go
@@ -150,14 +150,23 @@ func (e *processExecer) processRun(ctx context.Context, cmd model.Cmd, w io.Writ
 	defer close(statusCh)
 
 	logger.Get(ctx).Infof("Running cmd: %s", cmd.String())
-	c := e.localEnv.ExecCmd(cmd, logger.Get(ctx))
+	c, err := e.localEnv.ExecCmd(cmd, logger.Get(ctx))
+	if err != nil {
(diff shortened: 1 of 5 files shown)


internal/controllers/core/cmd/execer.go
```diff
@@ -150,14 +150,23 @@ func (e *processExecer) processRun(ctx context.Context, cmd model.Cmd, w io.Writ
 	defer close(statusCh)
 
 	logger.Get(ctx).Infof("Running cmd: %s", cmd.String())
-	c := e.localEnv.ExecCmd(cmd, logger.Get(ctx))
+	c, err := e.localEnv.ExecCmd(cmd, logger.Get(ctx))
+	if err != nil {
+		logger.Get(ctx).Errorf("%s invalid cmd: %v", cmd.String(), err)
```
nicks, line 155:
```suggestion
		logger.Get(ctx).Errorf("%q invalid cmd: %v", cmd.String(), err)
```

i usually quote the comment with %q, so it's easier to see how whitespace plays in

nicks (approved):
LGTM

ya, i'm always on the fence how functions should handle invalid inputs, agreed this seems like it should be handled further upstream

PR #4844, opened 2021-08-10
tiltfile: add reconciler tests, and fix up filewatch init/deletion

nicks:
Hello @milas, @landism,

Please review the following commits I made in branch nicks/delete:

b4cdf2f87bb28e45ef03a3e1bf2d1650ea8b6706 (2021-08-10 12:43:19 -0400)
tiltfile: add reconciler tests, and fix up filewatch init/deletion

50aa214830523ec34b98efd9f44c9decc7c6353b (2021-08-10 12:24:45 -0400)
engine: dependency-inject EngineMode


internal/controllers/core/tiltfile/filewatch.go
```diff
@@ -99,18 +97,30 @@ func ToFileWatchObjects(watchInputs WatchInputs) typedObjectSet {
 		}
 	}
 
+	paths := []string{}
 	if len(watchInputs.ConfigFiles) > 0 {
+		paths = append(paths, watchInputs.ConfigFiles...)
+	} else if watchInputs.TiltfilePath != "" {
+		// A complete ConfigFiles set should include the Tiltfile. If it doesn't,
+		// add it to the watch list now.
+		paths = append(paths, watchInputs.TiltfilePath)
```
milas, line 106: Given the comment, should we use a set and/or check to see if slice already contains `watchInputs.TiltfilePath`?

milas (approved):

internal/controllers/core/tiltfile/filewatch.go
nicks: mayyyyybe? tbqh, I'm not totally convinced we should even auto-reload the tiltfile anymore, and mainly trying to leave it alone right now.

PR #4845, opened 2021-08-10
web: Add resource grouping by labels to table view

lizzthabet:
This PR:
* Adds expandable label groups to table view in Storybook with tests
* Refactors some resource grouping code in detail view to share methods and styles with table view

Groups in table view, expanded:
Screen Shot 2021-08-11 at 3 07 04 PM

Groups in table view, collapsed:
Screen Shot 2021-08-11 at 3 07 28 PM

Storybook:
Screen Shot 2021-08-11 at 3 08 04 PM

Closes [ch12498](https://app.clubhouse.io/windmill/story/12498/add-expandable-resource-groups-to-table-view)


web/src/labels.ts
```diff
@@ -46,3 +64,89 @@ export function getUiLabels({ labels }: UILabels): string[] {
...
+
+  if (resources === undefined) {
+    return false
+  }
+
+  return resources.some((r) => getLabels(r).length > 0)
+}
+
+// Shared resource grouping styles
+// TODO: Consider moving these to a resource grouping specific file
```
lizzthabet, line 88: Any opinions here? I don't mind keeping the resource grouping + label together, but don't want it to be confusing.

web/src/types.ts
```diff
@@ -148,3 +148,9 @@ export type SidebarOptions = {
   // Filter options
   resourceNameFilter: string
 }
+
+// TODO (lizz): Centralize other redundant type definitions
```
lizzthabet, line 152: I think doing this now will make the diff too big to be useful. 🤓 I'll do this in a follow-up pr.

web/src/ResourceGroupsContext.tsx
```diff
@@ -0,0 +1,69 @@
...
+    // Note: this warning should only show in development
+    console.warn(
+      "Resource group context is not set. Did you forget to provide it?"
+    )
+  },
+  getGroup: () => {
+    // Note: this warning should only show in development
+    console.warn(
+      "Resource group context is not set. Did you forget to provide it?"
+    )
```
lizzthabet, line 28: I don't have strong opinions about keeping this warning message, but I found it useful in my development.

web/src/ResourceGroupsContext.tsx
```diff
@@ -0,0 +1,69 @@
...
+      ...prevState,
+      [groupLabel]: nextGroupState,
+    }))
+  }
+
+  function getGroup(groupLabel: string) {
+    return expanded[groupLabel] ?? DEFAULT_GROUP_STATE
+  }
+
+  const defaultValue = { expanded: {}, setGroup, getGroup }
```
nicks, line 62: i think 'expanded' is always empty?

web/src/ResourceGroupsContext.tsx
```diff
@@ -0,0 +1,69 @@
+import { createContext, PropsWithChildren, useContext } from "react"
+import { AnalyticsAction, AnalyticsType, incr } from "./analytics"
+import { usePersistentState } from "./LocalStorage"
+
+const GROUPS_LOCAL_STORAGE_NAME = "resource-groups"
+const DEFAULT_GROUP_STATE = true
+
+export type GroupExpandedState = { [key: string]: boolean }
+
+type ResourceGroupsContext = {
```
nicks, line 10: It seems extremely surprising to me that a method named getGroup/setGroup does collapse/expand? I feel like this file needs a naming/comment revamp (as well as tests)

nicks (commented):
I have some high-level architecture concerns about this PR. I think it would be a lot easier to move this forward if you broke up the state management and UI components into separate PRs.

web/src/labels.ts
```diff
@@ -46,3 +64,89 @@ export function getUiLabels({ labels }: UILabels): string[] {
...
+  }
+
+  if (resources === undefined) {
+    return false
+  }
+
+  return resources.some((r) => getLabels(r).length > 0)
+}
+
+// Shared resource grouping styles
```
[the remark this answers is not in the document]
lizzthabet: yup, i think that's generally a correct read on how i approached this implementation. i extracted out the shared styling between the two views because there are a lot of material ui component style overrides. each view implementation adds additional styling that's specific to the view.

i didn't think about how reusable the components are outside of these two views because we're only using grouping in these places. it seemed like an okay trade-off to me, but i'm happy to revisit that, either to create more reusable versions of the components or organize the code in a different manner.

there is a lack of commenting and documentation about the relationship bw the components that i can see is causing friction!

web/src/ResourceGroupsContext.tsx
lizzthabet: oh, that's my goof! it should be `expanded` from the persistent state above. (i implemented the shared context before the local storage state and forgot to change that.)

web/src/ResourceGroupsContext.tsx
lizzthabet: ah, yeah! 🤦🏻 my thinking was that we might want to track other properties (besides an `expanded` state) related to groups in the future, but i didn't want to over-engineer by creating a more complex state than we need right now. i can make these methods/state clearer and add comments.

landism:
Some design questions after playing around with it for a bit: (which might have already been considered and rejected or are probably just todos/future work anyway)
* Is it worth showing the status summary on collapsed groups, like we do in the sidebar?
* If I have a tiltfile with no labels, I can sort by alerts to ensure broken resources show at the top. If I add labels, is there a way to do that? Or is it only possible to sort within a group? (maybe a checkbox to control whether we list in groups?)
* Is it worth adding a tooltip advertising groups on the table view like we have in the sidebar?
* The collapsed group headers on the table view are kind of hard to distinguish from the background:
image
WDYT of making them look like the sidebar group headers?
image
(this is at least partly driven by it feeling weird that the button expands all the way across horizontally, but is pretty much invisible, which is a pretty minor nit)

lizzthabet:
i'm going to split this pr into two prs, so it'll be easier to review!

lizzthabet:
> Some design questions after playing around with it for a bit: (which might have already been considered and rejected or are probably just todos/future work anyway)
> 
>     * Is it worth showing the status summary on collapsed groups, like we do in the sidebar?
> 
>     * If I have a tiltfile with no labels, I can sort by alerts to ensure broken resources show at the top. If I add labels, is there a way to do that? Or is it only possible to sort within a group? (maybe a checkbox to control whether we list in groups?)
> 
>     * Is it worth adding a tooltip advertising groups on the table view like we have in the sidebar?
> 
>     * The collapsed group headers on the table view are kind of hard to distinguish from the background:
>       image
>       WDYT of making them look like the sidebar group headers?
>       image
>       (this is at least partly driven by it feeling weird that the button expands all the way across horizontally, but is pretty much invisible, which is a pretty minor nit)

@landism, these are super great questions! (sorry, i converted this back into draft mode before i saw your comments.)

* each resource group will have a status summary next to it and there will be the same info icon and tooltip that links to the resource grouping docs. i was planning to do those in separate prs.
* in the designs, you can sort by status and when you sort one label group, it'll sort all the label groups by the same criteria. i'm not 100% sure how global sorting state will work with react-table, but that's the functionality i'm going to aim for! (this work is outlined in [ch12496](https://app.clubhouse.io/windmill/story/12496/add-support-for-status-sorting-in-table-view)). i think that'll address the missing "sort by alerts" functionality in table view.
* to your point about 100% width but visually indistinguishable buttons, i had the same thought! i was planning to bring this up with surbhi during our design review, when i have a couple more of the resource grouping features implemented. it's good to know that i wasn't the only one who noticed that.

web/src/labels.ts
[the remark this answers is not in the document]
nicks: two patterns i've seen are:
1) the one we use in SidebarItemView, where we create a single component with top-level CSS classes to control variants on styling (https://github.com/tilt-dev/tilt/blob/master/web/src/SidebarItemView.tsx#L68)
2) the material-ui styling hooks, which we use in a few places: https://material-ui.com/styles/basics/#hook-api

i agree that the main goal is to figure out what's the best way to move this PR forward, and the way to do that might be to use this approach for now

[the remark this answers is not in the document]
nicks: i think it would also be totally reasonable to make these two completely separate components with no shared code. i was reminded of this old advice this week: https://twitter.com/zkat__/status/1425859526616616960

lizzthabet:
thank you for all your feedback, @nicks and @landism! i've updated this pr to include only the addition of expandable resource groups in table view from storybook. i pulled the shared styles (which are 95% removing the material ui default styles) into a separate `ResourceGroups` file and added comments. this solution isn't the _most_ dry or reusable, but i think we can revisit the code if/when we use it in a third place. 

i'll open a separate pr for the persistent expanded group state work.

web/src/ResourceGroupsContext.tsx
lizzthabet: i will address this in a separate pr with the persistent state code!

web/src/OverviewTable.tsx
```diff
@@ -551,13 +618,54 @@ function viewToRowValues(
...
-export default function OverviewTable(props: OverviewTableProps) {
-  let logStore = useLogStore()
-  const data = React.useMemo(() => viewToRowValues(props.view, logStore), [
-    props.view.uiResources,
-    props.view.uiButtons,
-  ])
+function getResourceLabels(resource: UIResource) {
+  const uiLabels = asUILabels({ labels: resource.metadata?.labels })
+  return getUILabels(uiLabels)
+}
```
landism, line 624: this function is kind of confusing!
1. the names `getResourceLabels`, `asUILabels`, and `getUILabels` don't really distinguish themselves from each other
2. `getUILabels(uiLabels)` - getting ui labels from ui labels doesn't make much sense!

Some options:
* At a glance, it seems pretty clear this is just getting labels from a resource so maybe just adding a `: []string` return type annotation would clear things up a bit
* The only other place that calls `asUILabels` also just passes its result into `getUILabels`, so maybe just `asUILabels`, `getUILabels`, and `getResourceLabels` into one function, to clean up the naming?
* rename `UILabels`/`asUILabels`, which doesn't seem like the right name anyway - at that point, it represents all labels, not just ones used by the UI

web/src/OverviewTable.tsx
```diff
@@ -68,16 +99,52 @@ type OverviewTableStatus = {
...
 
   td:first-child {
     padding-left: ${SizeUnit(1)};
   }
   td:last-child {
     padding-right: ${SizeUnit(1)};
   }
+
+  &.isGroup {
+    /* border: 1px yellow solid; */
```
landism, line 144: 👀

landism (commented):
didn't have time to finish reviewing before I had to stop for the day, but here are some comments

web/src/OverviewTable.test.tsx
```diff
@@ -35,3 +49,170 @@ it("shows buttons on the appropriate resources", () => {
...
+          resourcesInLabelGroup = unlabeled.map((resource) => resource.name)
+        } else if (groupName === TILTFILE_LABEL) {
+          resourcesInLabelGroup = tiltfile.map((resource) => resource.name)
+        } else {
+          resourcesInLabelGroup = labelsToResources[groupName].map(
+            (resource) => resource.name
+          )
+        }
+
+        expect(resourcesInTable).toEqual(resourcesInLabelGroup)
```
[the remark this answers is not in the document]
lizzthabet: this makes sense to me!

web/src/OverviewTable.test.tsx
```diff
@@ -35,3 +49,170 @@ it("shows buttons on the appropriate resources", () => {
...
+      const resourceGroups = wrapper.find(OverviewGroup)
+
+      resourceGroups.forEach((group) => {
+        // Find the label group name
+        const groupName = group.find(OverviewGroupName).text()
+        // Find the resource list displayed in the table
+        const table = group.find(Table)
+        const resourcesInTable = table
+          .find(TableNameColumn)
+          .map((resourceName) => resourceName.text())
```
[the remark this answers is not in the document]
lizzthabet: yeah, that's a good point. i hesitate to hardcode expected cases based on shared mocked data because, (at least in other codebases i've worked on), when that data is changed for an unrelated reason, it can cause a bunch of tests to falsely fail. (though one might argue that false negatives are better than false positives when it comes to tests passing!)

i think in this case, because the `toEqual()` values are generated from the `view`, we should get a correct failure if `OverviewTable` changes. i'll keep this in mind for future tests i write.

web/src/OverviewTable.tsx
```diff
@@ -634,3 +747,82 @@ export default function OverviewTable(props: OverviewTableProps) {
...
+export function TableGroupedByLabels(props: OverviewTableProps) {
+  const logStore = useLogStore()
+  const data = useMemo(
+    () =>
+      resourcesToTableCells(
+        props.view.uiResources,
+        props.view.uiButtons,
+        logStore
+      ),
+    [props.view.uiResources]
```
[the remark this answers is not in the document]
lizzthabet: thanks for catching these!

web/src/OverviewTable.test.tsx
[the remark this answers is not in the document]
landism: Oops, you're right, I got my expected and actual reversed. Sorry!

(maybe I've become too reliant on the pattern of explicitly naming expected and actual vars "expected" and "actual")

web/src/OverviewTable.tsx
lizzthabet: you're right! i know the typescript gymnastics are a little goofy with both the type predicate and the casting function. i'll added better commenting and created a single helper function that's used instead of `as`/`isUILabels`.

web/src/labels.ts
```diff
@@ -46,3 +59,22 @@ export function getUiLabels({ labels }: UILabels): string[] {
...
+// and a label accessor function
+export function resourcesHaveLabels<T>(
+  features: Features,
+  resources: T[] | undefined,
+  getLabels: (resource: T) => string[]
+): boolean {
+  // Labels on resources are ignored if feature is not enabled
+  if (!features.isEnabled(Flag.Labels)) {
+    return false
+  }
```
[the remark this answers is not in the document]
lizzthabet: i think this is okay to leave as-is. maybe a little bit of personal preference here! in my mind, if the labels feature flag is off, then it's the same as a resource not having labels.

web/src/OverviewTable.test.tsx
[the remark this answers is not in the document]
lizzthabet: gotchu! it'll be helpful if i rename the vars for better clarity.

web/src/OverviewTable.tsx
```diff
@@ -568,23 +667,29 @@ export default function OverviewTable(props: OverviewTableProps) {
...
-    if (c.Header === "Widgets" && widgetsVisible !== c.isVisible) {
-      c.toggleHidden(!widgetsVisible)
-    }
-  })
+  // const widgetsVisible = hasWidgets(props.view)
+  // columns.forEach((c) => {
+  //   if (c.Header === "Widgets" && widgetsVisible !== c.isVisible) {
+  //     c.toggleHidden(!widgetsVisible)
+  //   }
+  // })
```
landism, line 687: I think it's fine to just delete this and worry about it if/when someone gives us a reason to

landism (approved):
🪗 🎉

PR #4849, opened 2021-08-11
web: show resource uibuttons on table view

landism:
For any uibuttons that specify a resource as their location, render them in the table view.
1. Add a "widgets" column iff there are any buttons.
2. Button text currently only shows up in the tooltip, to save room.

image

FWIW, I only invested in the "ton of buttons" rendering case enough to make it usable, but not good. Here's what it currently looks like:
image


web/src/OverviewTable.tsx
```diff
@@ -327,7 +340,36 @@ function TableTriggerModeColumn({ row }: CellProps<RowValues>) {
...
+    )
+  })
+  return <WidgetCell>{buttons}</WidgetCell>
+}
+
+// https://react-table.tanstack.com/docs/api/useTable#column-options
+// The docs on this are not very clear!
+// "accessor" should return a primitive, and that primitive is used for sorting and filtering
+// the Cell function can get whatever it needs to render via row.original
+// best evidence I've (Matt) found: https://github.com/tannerlinsley/react-table/discussions/2429#discussioncomment-25582
```
lizzthabet, line 371: thanks for this comment 👌🏻

web/src/OverviewTable.tsx
```diff
@@ -478,29 +536,53 @@ function resourceTypeLabel(r: UIResource): string {
...
-      columns,
+      columns: columnDefs,
       data,
       autoResetSortBy: false,
     },
     useSortBy
   )
 
+  // if there are no widgets, hide the widgets column
+  // TODO(matt) is this actually what we want?
```
milas, line 578: fwiw this seems reasonable to me

milas (approved):

web/src/OverviewTable.tsx
```diff
@@ -478,29 +536,53 @@ function resourceTypeLabel(r: UIResource): string {
...
+  const widgetsVisible = hasWidgets(props.view)
```
lizzthabet, line 579: this makes sense to me, but i'll let @SurbhiSGupta and others chime in too!

web/src/OverviewTable.tsx
```diff
@@ -431,6 +483,11 @@ function uiResourceToCell(r: UIResource, alertIndex: LogAlertIndex): RowValues {
   let currentBuildStartTime = res.currentBuild?.startTime ?? ""
   let isBuilding = !isZeroTime(currentBuildStartTime)
   let hasBuilt = lastBuild !== null
+  let buttons = allButtons.filter(
+    (b) =>
+      b.spec?.location?.componentType?.toLowerCase() === "resource" &&
+      b.spec?.location?.componentID === r.metadata?.name
+  )
```
lizzthabet, line 490: (nitpick) it'd be nice to have this as a helper function related to the custom ui buttons code, maybe like `isResourceButton`?

lizzthabet (commented):
this looks good! i had one outstanding ux question and may have some a11y suggestions based on your answer.

web/src/OverviewTable.tsx
landism: yeah, it's at least fine. my main hesitations are:
1) it's inconsistent with other columns (e.g., we don't hide endpoints or podids when all of their cells are empty)
2) showing an empty column allows space for advertising the feature, though I'm not sure we'd do that at this point anyway

web/src/OverviewTable.tsx
```diff
@@ -327,7 +340,36 @@ function TableTriggerModeColumn({ row }: CellProps<RowValues>) {
...
+      content = (
+        <TiltTooltip title={b.spec.text}>
+          <span>{content}</span>
+        </TiltTooltip>
+      )
+    }
+    return (
+      <CustomActionButton key={b.metadata?.name} button={b}>
+        {content}
+      </CustomActionButton>
```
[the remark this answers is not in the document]
landism: Nope, not intentional. Thanks!

web/src/OverviewTable.tsx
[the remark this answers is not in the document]
landism: Yes, for sure. Added a TODO.

web/src/OverviewTable.tsx
landism: yeah! though I went with the less general `buttonsForResource` (which also does the filter) since for now that's the only way we'd use it and I'm not sure if that will change.

web/src/OverviewTable.tsx
landism: I'm happy to change this later, but don't think it matters enough either way to block merging on

PR #4854, opened 2021-08-11
tiltfile: add builtins for extension_repo, extension

nicks:
Hello @milas, @landism,

Please review the following commits I made in branch nicks/tiltfileapi:

7ba55ef879a359b7586a67d321973420dba4a4fb (2021-08-11 19:23:05 -0400)
tiltfile: add builtins for extension_repo, extension

279cd7a7cdff45fa247eceef951187fb6ea1cebd (2021-08-11 18:28:26 -0400)
tiltfile: disambiguate tiltextension and starkit.Extension


nicks:
this puts everything together and should be enough for me to write some cool demos with, while setting us up to codegen some of this later

internal/tiltfile/tiltfile.go
```diff
@@ -186,6 +189,9 @@ func (tfl tiltfileLoader) Load(ctx context.Context, filename string, userConfigS
 	tlr.Manifests = manifests
 	tlr.TeamID = s.teamID
 
+	objectSet, _ := v1alpha1.GetState(result)
+	tlr.ObjectSet = objectSet
```
milas, line 193: Why do we ignore errors from all these? It's consistent with the others so not suggesting changing now, but would it make sense to create an aggregate error for assignment to `tlr.Error` in the future?

web/storybook.tiltfile
```diff
@@ -1,6 +1,10 @@
 # Enforce versioning so that labels are supported
 version_settings(constraint='>=0.22.1')
 
+# Uncomment to try the cancel button extension
+#v1alpha1.extension_repo(name='default', url='file:///home/nick/src/tilt-extensions')
```
milas, line 5: Harmless but just commenting in case you meant to remove

milas (approved):

internal/tiltfile/tiltfile.go
nicks: ya, I'm going to change them to all use MustState (which panics on error)

The problem is that an error here is a hard-stop programmer error - in particular, that you're trying to get the state of a plugin, but never registered that plugin.

Honestly, my vision for this is that TiltfileLoadResult goes away entirely, and is almost entirely a collection of API objects and/or Plugin state

web/storybook.tiltfile
nicks: nope! just showing what this looks like. the v1alpha1 in the module name is a bit weird IMHO, but it's a k8s convention that i guess people are familiar with? and will put pressure on us to make it v1

internal/tiltfile/tiltfile.go
nicks: wait, nope, the original code was correct. I added this comment:

```
	// All data models are loaded with GetState. We ignore the error if the state
	// isn't properly loaded. This is necessary for handling partial Tiltfile
	// execution correctly, where some state is correctly assembled but other
	// state is not (and should be assumed empty).
```

nicks: (it would not be correct to aggregate + report the GetState errors, since that would just obscure the original error)

milas: ahhhh ok that makes sense

PR #4857, opened 2021-08-12
api: add UIInputs to UIButton

landism:
### Feature

Updates the spec to allow a UIButton to have associated text inputs.

### Next steps

Update the cmd controller to grab current values off the button when it triggers
Update the UI to render the fields

### Alternatives

* Make UIInput a standalone api object
  * Pros:
    * we'll want this if we want to support, e.g., a local_resource cmd consuming a text field directly
    * don't need the weird parallel `[]UIInputSpec` `[]UIInputStatus` (but maybe that's not weird in k8s-land?)
  * Cons:
    * seems like a lot more bookkeeping and things to go wrong for the button use case


pkg/apis/core/v1alpha1/uibutton_types.go
```diff
@@ -174,12 +187,74 @@ func (in *UIButtonList) GetListMeta() *metav1.ListMeta {
...
+type UITextInputStatus struct {
+	// The content of the text input.
+	Value string `json:"value" protobuf:"bytes,1,opt,name=value"`
+}
+
+// Defines an Input to render in the UI.
+// If UIButton is analogous to an HTML <form>,
+// UIInput is analogous to an HTML <input>.
+type UIInputSpec struct {
+	// An ID to distinguish this input from others.
```
nicks, line 211: fwiw, I think the typical kubernetes convention would be to call this a 'Name' instead of an 'ID'. See:
https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#lists-of-named-subobjects-preferred-over-maps

pkg/apis/core/v1alpha1/uibutton_types.go
```diff
@@ -174,12 +187,74 @@ func (in *UIButtonList) GetListMeta() *metav1.ListMeta {
...
+		fieldErrors = append(fieldErrors, field.Invalid(path, in, "must specify exactly one input type"))
+	}
+
+	return fieldErrors
+}
+
+// The status corresponding to a UIInputSpec
+type UIInputStatus struct {
+	// The ID of the input to which this status belongs
+	ID string `json:"id" protobuf:"bytes,1,opt,name=id"`
```
nicks, line 239: ditto

nicks (approved):

PR #4860, opened 2021-08-13
cmd: when a UIButton starts a Cmd, add its UIInputs to the proc's env

landism:


internal/controllers/core/cmd/controller.go
```diff
@@ -151,11 +152,17 @@ func (c *Controller) lastStartEventTime(startOn *StartOnSpec, buttons map[string
...
+		if !lastEventTime.Time.Before(startOn.StartAfter.Time) && lastEventTime.Time.After(latestTime) {
+			latestTime = lastEventTime.Time
+			latestButton = b
 		}
 	}
-	return cur
+
+	var inputs []v1alpha1.UIInputStatus
+	if latestButton != nil {
+		inputs = latestButton.Status.Inputs
```
nicks, line 163: this seems surprising to me? i would have expected you'd want to grab the inputs from all buttons, rather than the most recently clicked button

internal/controllers/core/cmd/controller.go
```diff
@@ -138,10 +138,11 @@ func (c *Controller) fileWatches(ctx context.Context, cmd *v1alpha1.Cmd) (map[st
 }
 
 // Fetch the last time a start was requested from this target's dependencies.
-func (c *Controller) lastStartEventTime(startOn *StartOnSpec, buttons map[string]*v1alpha1.UIButton) time.Time {
-	cur := time.Time{}
+func (c *Controller) lastStartEvent(startOn *StartOnSpec, buttons map[string]*v1alpha1.UIButton) (time.Time, []v1alpha1.UIInputStatus) {
+	latestTime := time.Time{}
+	var latestButton *v1alpha1.UIButton
```
nicks, line 143: UIButtons also have a DefaultValue field. how does that play out here? i kind of expected the DefaultValue would be injected into the env, even if the button wasn't pressed. though maybe I'm misunderstanding the intended behavior of DefaultValue

internal/controllers/core/cmd/controller.go
landism: I don't think I have a strong opinion on this and I think I'm fine going either way until we come up with a concrete use case for a cmd with multiple buttons in its `StartOn`.

I went this way because it seemed like it'd surprise me if I clicked a button and my process got values from other buttons that seem (to me) not really related.

My mental model is a button is like an html `<form>` with a mandatory `input type="submit"`, and if two forms have the same `action`, they don't get each others `input`s.

Or, from another angle - a cmd has some set of env vars it consumes, which we could call the api it exposes to buttons that want to trigger it. If a cmd consumes inputs from all of its buttons then things get messy if two buttons representing two workflows try to set the same env var (or maybe not *that* messy - we just take the most recent one for each).

I'm fine shipping as-is and changing as needed, or changing to match your expectations if you have a strong expectation it should go that way, or chatting. I've been trying to think of use cases that'd tilt it one way or another and am starting to think having one cmd triggered by multiple buttons doesn't currently have a compelling use case (vs declaring multiple cmds, one per button)

nicks: i buy that argument! ya, I haven't totally thought through what it means to to have multiple buttons with inputs

nicks (approved):

internal/controllers/core/cmd/controller.go
landism: Whoops, I typed an answer to this but I guess I didn't submit?

The `UITextInput`'s `DefaultValue` is passed as the `TextField`'s `defaultValue`. When the user clicks the button, the frontend updates the `UITextInput`'s status with the `TextField`'s current value, which will already be making use of the default as appropriate.

PR #4866, opened 2021-08-19
web: add persistent state to resource grouping

lizzthabet:
This PR adds state that tracks which label groups are expanded / collapsed and persists across views within the dashboard and across user sessions. If there isn't any saved state for a label group, then it'll default to expanded. As a user toggles a group closed and open, the `ResourceGroupsContext` will track that label's expanded state in local storage, under the "resource-groups" key.

Resource groups aren't live in table view yet, so the best way to test locally is in Storybook. You can toggle [OverviewResourcePane > Ten Resources With Labels](http://localhost:9009/?path=/story/new-ui-overviewresourcepane--ten-resources-with-labels) and [Overview > OverviewTable > Ten Resources With Labels](http://localhost:9009/?path=/story/new-ui-overview-overviewtable--ten-resource-with-labels) and see that the expanded state is the same across views.

I also added tests for the label helpers.

Closes [ch12556](https://app.clubhouse.io/windmill/story/12556/add-persistent-expanded-state-for-resource-grouping).

### DIFF
--- a/web/src/ResourceGroupsContext.test.tsx
+++ b/web/src/ResourceGroupsContext.test.tsx
@@ -1,0 +1,152 @@
+import { mount, ReactWrapper } from "enzyme"
+import React from "react"
+import { AnalyticsAction, AnalyticsType } from "./analytics"
+import {
+  cleanupMockAnalyticsCalls,
+  expectIncrs,
+  mockAnalyticsCalls,
+} from "./analytics_test_helpers"
+import {
+  DEFAULT_GROUP_STATE,
+  ResourceGroupsContextProvider,
+  useResourceGroups,
+} from "./ResourceGroupsContext"
+
+const GROUP_STATE_ID = "test-group-state"
+const LABEL_STATE_ID = "test-label-state"
+
+// This is a very basic test component that prints out the state
+// from the ResourceGroups context and provides buttons to trigger
+// methods returned by the context, so they can be tested
+const TestConsumer = (props: { labelName?: string }) => {
+  const { groups, getGroup, toggleGroupExpanded } = useResourceGroups()
+
+  return (
+    <>
+      <p id={GROUP_STATE_ID}>{JSON.stringify(groups)}</p>
+      {/* Display the label state if a specific label is present */}
+      {props.labelName && (
+        <p id={LABEL_STATE_ID}>{JSON.stringify(getGroup(props.labelName))}</p>
+      )}
+      {/* Display a button to toggle the label state if a specific label is present */}
+      {props.labelName && (
+        <button
+          onClick={() =>
+            toggleGroupExpanded(props.labelName || "", AnalyticsType.Grid)
+          }
+        />
+      )}
+    </>
+  )
+}
+
+describe("ResourceGroupsContext", () => {
+  let wrapper: ReactWrapper<typeof TestConsumer>
+
+  beforeEach(() => {
+    mockAnalyticsCalls()
+  })
+
+  afterEach(() => {
+    localStorage.clear()
+    cleanupMockAnalyticsCalls()
+  })
+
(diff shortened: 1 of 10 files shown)


web/src/ResourceGroupsContext.test.tsx
```diff
@@ -0,0 +1,152 @@
...
+      {props.labelName && (
+        <button
+          onClick={() =>
+            toggleGroupExpanded(props.labelName || "", AnalyticsType.Grid)
+          }
+        />
+      )}
+    </>
+  )
+}
```
lizzthabet, line 41: i'm not sure if this is the best way to test a provider/context, but looking around our codebase and (very briefly) the internet, it seems like this is a common approach. open to suggestions if there's a better way to test!

web/src/ResourceGroupsContext.tsx
```diff
@@ -0,0 +1,81 @@
...
+const resourceGroupsContext = createContext<ResourceGroupsContext>({
+  groups: {},
+  toggleGroupExpanded: () => {
+    console.warn("Resource group context is not set.")
+  },
+  getGroup: () => {
+    console.warn("Resource group context is not set.")
+    return { ...DEFAULT_GROUP_STATE }
+  },
+})
```
lizzthabet, line 31: @nicks, i addressed [your feedback](https://github.com/tilt-dev/tilt/pull/4845#discussion_r687157169) here by changing the shape of the resource groups state from 
```
state: { labelName: boolean }
```
 to 
```
state: { labelName: { expanded: boolean } }
```
so it's clearer and more extensible if there's other state we want to track for label groups. the method names are also renamed to be clearer. (it may be a little awkward that there's a `getGroup` but not a `setGroup` where you can pass in the desired group state? i figured a `toggle` method is more useful for the implementation now, but i can refactor that.)

web/src/ResourceGroupsContext.test.tsx
nicks: this seems like a reasonable approach. we've also done this:
https://github.com/tilt-dev/tilt/blob/master/web/src/ResourceNav.test.tsx#L15
which uses the "Consumer" react component (rather than hooks) and attaches the state to a class (rather than serializing/deserializing). but i don't think either approach is necessarily more correct.

i might add a helper method for the stanza:

```
      const labelState = wrapper.find(`#${LABEL_STATE_ID}`).text()
      expect(labelState).toBe(JSON.stringify({ expanded: true }))
```

web/src/ResourceGroupsContext.tsx
```diff
@@ -0,0 +1,81 @@
...
+import { usePersistentState } from "./LocalStorage"
+
+export type GroupState = { expanded: boolean }
+
+export type GroupsState = {
+  [key: string]: GroupState
+}
+
+type ResourceGroupsContext = {
+  groups: GroupsState
```
nicks, line 12: i'm surprised you exposed both `groups` and `getGroups`? seems like you'd want people to use the accessor function.

web/src/ResourceGroupsContext.tsx
nicks: :+1:

nicks (approved):

web/src/ResourceGroupsContext.tsx
lizzthabet: that's a good point! i thought it'd be useful for a context to still expose its whole state, but i don't have any strong opinions about it. (it is useful for testing.)

web/src/ResourceGroupsContext.test.tsx
lizzthabet: ah, cool! it's good to see another example for testing. added a couple small helpers!

PR #4869, opened 2021-08-19
web: sort uibuttons

### LINKED ISSUE #4868: UIButton rendering order is nondeterministic
### Steps to Reproduce
Add multiple buttons to a resource, e.g.:
```
local_resource('hi', 'echo hi')
load('ext://uibutton', 'cmd_button', 'location', 'text_input')
for i in range(1, 20):
  cmd_button('test' + str(i), resource='hi', argv=['echo', 'foo', '$str'], text='Hi! ' + str(i), inputs=[text_input('str')]) #, icon_name='travel_explore')
```

### Current Behavior
Every time you refresh the tilt ui, the buttons are in a different / seemingly random order.

### Expected Behavior
The buttons appear in a consistent order.

### Context
We had a similar problem with UIResource: #4756
We can't cheaply use that solution for UIButton since they're not all generated in one place at one time. We could get part of the way there with the global_vars extension or by, on button creation, querying all buttons and setting order to the max order plus 1.
For now it's probably fine to just sort by alpha and worry about this if/when users have enough buttons that they want to control their order.

landism:
Fixes #4868

### DIFF
--- a/web/src/ApiButton.tsx
+++ b/web/src/ApiButton.tsx
@@ -95,16 +95,25 @@ export const ApiButton: React.FC<ApiButtonProps> = (props) => {
   )
 }
 
-export function buttonsForResource(
+export function buttonsForComponent(
   buttons: UIButton[] | undefined,
-  resourceName: string | undefined
+  componentType: string,
+  componentID: string | undefined
 ): UIButton[] {
-  if (!resourceName || !buttons) {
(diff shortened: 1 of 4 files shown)


nicks (approved):
it might make sense to have an 'order' field like we do for uiresource, so users can specify relative order of buttons. but this is a good change either way.

PR #4870, opened 2021-08-19
web: add resource status summaries to groups in table view

lizzthabet:
This PR adds resource status summaries to groups in table view and refactors the core status summary code to take a template type so it can be used for resources in different data forms.

The best way to test this code is to look at Storybook in [Overview > OverviewTable > Ten Resources With Labels](http://localhost:9009/?path=/story/new-ui-overview-overviewtable--ten-resource-with-labels), since groups are available in the main table view just yet!

Collapsed and expanded groups:
Screen Shot 2021-08-19 at 3 32 54 PM

Collapsed groups:
Screen Shot 2021-08-19 at 3 32 40 PM

Closes [ch12501](https://app.clubhouse.io/windmill/story/12501/display-resource-status-summary-for-each-label-group).

### DIFF
--- a/web/src/ResourceStatusSummary.tsx
+++ b/web/src/ResourceStatusSummary.tsx
@@ -7,6 +7,7 @@ import { ReactComponent as PendingSvg } from "./assets/svg/pending.svg"
 import { ReactComponent as WarningSvg } from "./assets/svg/warning.svg"
 import { FilterLevel } from "./logfilters"
 import { useLogStore } from "./LogStore"
+import { RowValues } from "./OverviewTable"
 import { usePathBuilder } from "./PathBuilder"
 import SidebarItem from "./SidebarItem"
 import { buildStatus, combinedStatus, runtimeStatus } from "./status"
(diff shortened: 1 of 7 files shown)


web/src/ResourceStatusSummary.tsx
```diff
@@ -286,58 +287,87 @@ function ResourceMetadata(props: { counts: StatusCounts }) {
   return <></>
 }
 
-type ResourceStatusSummaryProps = {
-  view: Proto.webviewView
+/**
+ * The ResourceStatusSummary component takes a template type
```
nicks, line 291: have you considered adding a `statuses: ResourceStatus[]` parameter instead?

i found all the type templating difficult to follow. are the type templates there mainly so you can get statuses from different object types, or are there other benefits?

nicks (approved):

web/src/ResourceStatusSummary.tsx
lizzthabet: yup, that's why i added the template type! my thinking is that we want to be able to provide resource summaries throughout the app, but the resource data is going to come in different shapes. so this function accepts some resource type and a status getter function. but it sounds like that's creating level of abstraction that's difficult to follow, so i can simplify based on your suggestion.

PR #4873, opened 2021-08-20
engine: don't look in completed pods for running containers

nicks:
Hello @milas, @landism,

Please review the following commits I made in branch nicks/jobs:

ef490cca875b509964c65f54d655f3b890c8546b (2021-08-19 20:12:49 -0400)
engine: don't look in completed pods for running containers
I was playing around with live-update Jobs.

I saw weird behavior where:
- builds would get stuck pending forever on already completed jobs
- tilt would think that a pod had crash when a job completed normally

Note that this changes the behavior of crash rebuild.
Before, if a pod crashed, Tilt would immediately begin a new build.
Now, if a pod crashes, Tilt waits for the controller to replace
it before starting the crash rebuild.
I think this behavior is probably more correct, given
that Tilt can't know if the controller is going to replace the Pod.

### DIFF
--- a/internal/engine/upper_test.go
+++ b/internal/engine/upper_test.go
@@ -2275,66 +2275,6 @@ func TestUpperPodRestartsBeforeTiltStart(t *testing.T) {
 	assert.NoError(t, err)
 }
 
-// This tests a bug that led to infinite redeploys:
-// 1. Crash rebuild
-// 2. Immediately do a container build, before we get the event with the new container ID in (1). This container build
-//    should *not* happen in the pre-(1) container ID. Whether it happens in the container from (1) or yields a fresh
-//    container build isn't too important
-func TestUpperBuildImmediatelyAfterCrashRebuild(t *testing.T) {
-	f := newTestFixture(t)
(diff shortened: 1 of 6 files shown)


internal/engine/upper_test.go
```diff
@@ -2275,66 +2275,6 @@ func TestUpperPodRestartsBeforeTiltStart(t *testing.T) {
 	assert.NoError(t, err)
 }
 
-// This tests a bug that led to infinite redeploys:
-// 1. Crash rebuild
-// 2. Immediately do a container build, before we get the event with the new container ID in (1). This container build
-//    should *not* happen in the pre-(1) container ID. Whether it happens in the container from (1) or yields a fresh
-//    container build isn't too important
-func TestUpperBuildImmediatelyAfterCrashRebuild(t *testing.T) {
```
nicks, line 2283: i removed this test because I feel like the conditions it's testing (a pod crashes and tilt immediately starts building) shouldn't happen anymore. but open to suggestions on what might replace it.

milas (approved):

PR #4874, opened 2021-08-20
web: have InstrumentedButton use MUI Button instead of a plain ol' button

landism:
In a separate branch, I'm trying to use MUI's [ButtonGroup](https://material-ui.com/components/button-group/), which needs MUI `Button`s rather than `button`s.

This might be a bit controversial since `InstrumentedButton` is kind of controversial in the first place for mixing analytics and rendering, but I don't think this really commits us further to it?

(fwiw (and at the risk of sparking that controversy in this PR) - I went down the path of getting rid of `InstrumentedButton` and replacing it with an `instrumentedOnClick` that takes a click handler and wraps it with a call that reports the click to analytics, but didn't like result - it seemed too much up to every single component to remember to call `instrumentedOnClick`, while with `InstrumentedButton`, the compiler ensures we specify analytics for the vast majority of our buttons)

### DIFF
--- a/web/src/style-helpers.ts
+++ b/web/src/style-helpers.ts
@@ -104,11 +104,15 @@ export const mixinResetButtonStyle = `
   cursor: pointer;
 
   // undo Material UI's Button styling
+  // TODO - maybe we should be doing it like this? https://material-ui.com/customization/globals/#css
   letter-spacing: normal;
   min-width: 0px;
   text-transform: none;
   line-height: normal;
   font-size: ${FontSize.smallest};
+  &:hover {
+    background-color: transparent;
+  }
 `
 
 export const mixinTruncateText = `
--- a/web/src/style-helpers.ts
+++ b/web/src/style-helpers.ts
@@ -102,6 +102,13 @@ export const mixinResetButtonStyle = `
   margin: 0;
   font-family: inherit;
   cursor: pointer;
+
+  // undo Material UI's Button styling
+  letter-spacing: normal;
+  min-width: 0px;
+  text-transform: none;
+  line-height: normal;
+  font-size: ${FontSize.smallest};
 `
 
 export const mixinTruncateText = `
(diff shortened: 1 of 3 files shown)


milas (approved):
Seems fine to me - I think this might help move towards easier styling/component hierarchy since now the (Material) styled button is the base instead of the other way around? That said, we don't reaaallly buy-in to the Material UI theming (as evidenced by some of the reset CSS in this 😛) so maybe it doesn't fundamentally change things, but that seems like a largely independent issue, and I don't see this as being any worse

web/src/style-helpers.ts
```diff
@@ -102,6 +102,13 @@ export const mixinResetButtonStyle = `
   margin: 0;
   font-family: inherit;
   cursor: pointer;
+
+  // undo Material UI's Button styling
+  letter-spacing: normal;
+  min-width: 0px;
+  text-transform: none;
+  line-height: normal;
+  font-size: ${FontSize.smallest};
```
lizzthabet, line 111: i think there might be some fun material ui click animations to reset too. 😬 it's pretty subtle, but i noticed this kind of radiating shadow when i clicked on the trigger mode toggle (on both views) and the restart button (on detail view only). i don't think this needs to block this pr.

lizzthabet (approved):
lgtm! i'm not aware of the controversial conversation 😅, but i don't have an issue with a lightweight wrapper of buttons that submits analytics.

lizzthabet:
whoops, also i just noticed that there's now a hover background style applied to the star resource buttons (which is VERY subtle).

on storybook, the button sizing is off and there's a border on buttons, but i think that's probably going to be fixed with your other pr that corrects the style hierarchy in storybook. this screenshot shows the shadow that i'm seeing locally and in storybook, plus the weird sizing and borders.
Screen Shot 2021-08-20 at 2 15 49 PM

web/src/style-helpers.ts
landism: I think this is taken care of by `disableRipple` in f0906f8

landism:
> a hover background style applied to the star resource buttons

nice eye! fixed!

PR #4879, opened 2021-08-23
web: show options forms on uibuttons w/ inputs

landism:
When a UIButton's spec defines one or more inputs:
1. Attach a little arrow next to the button in the UI, which pops up a form allowing the user to set those inputs' values.
2. Include those inputs' values in the UIButtonStatus when the button is clicked.

The design/flow is currently a bit awkward but probably good enough to have people play around with it. The main goal was to come up one design that'd work for all button locations for now.

Here's how it looks in the nav bar:

In table view:

In overview action bar (on resource pane):

In action:

### DIFF
--- a/web/src/ApiButton.tsx
+++ b/web/src/ApiButton.tsx
@@ -1,14 +1,37 @@
-import { Icon, SvgIcon } from "@material-ui/core"
+import { ButtonGroup, Icon, SvgIcon, TextField } from "@material-ui/core"
+import ArrowDropDownIcon from "@material-ui/icons/ArrowDropDown"
 import moment from "moment"
-import React, { useState } from "react"
+import React, { useRef, useState } from "react"
 import { convertFromNode, convertFromString } from "react-from-dom"
+import styled from "styled-components"
+import FloatDialog from "./FloatDialog"
 import { InstrumentedButton } from "./instrumentedComponents"
+import { Color, FontSize, SizeUnit } from "./style-helpers"
 
 type UIButton = Proto.v1alpha1UIButton
+type UIInputSpec = Proto.v1alpha1UIInputSpec
+type UIInputStatus = Proto.v1alpha1UIInputStatus
+
+const ApiButtonFormRoot = styled.span`
+  z-index: 20;
+`
+const ApiButtonFormFooter = styled.div`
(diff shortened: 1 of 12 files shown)


web/src/ApiButton.tsx
```diff
@@ -1,15 +1,38 @@
...
+  font-size: ${FontSize.smallester};
+`
+export const ApiButtonLabel = styled.span``
 
 type ApiButtonProps = { className?: string; button: UIButton }
 
 type ApiIconProps = { iconName?: string; iconSVG?: string }
 
+export const ApiButtonInputsToggleButton = styled(InstrumentedButton)`
+  &&&& {
```
nicks, line 31: ?

web/src/ApiButton.tsx
```diff
@@ -1,15 +1,38 @@
...
+import styled from "styled-components"
+import FloatDialog from "./FloatDialog"
 import { InstrumentedButton } from "./instrumentedComponents"
+import { Color, FontSize, SizeUnit } from "./style-helpers"
 
 type UIButton = Proto.v1alpha1UIButton
+type UIInputSpec = Proto.v1alpha1UIInputSpec
+type UIInputStatus = Proto.v1alpha1UIInputStatus
+
+const ApiButtonFormRoot = styled.span`
```
nicks, line 15: you really shouldn't put block elements (like div) inside inline elements (like span) - browsers will behave somewhat erratically when you do this. Here's a good doc:
https://developer.mozilla.org/en-US/docs/Web/HTML/Inline_elements

web/src/ApiButton.tsx
```diff
@@ -19,6 +42,97 @@ const svgElement = (src: string): React.ReactElement => {
...
 }
 
+type ApiButtonInputProps = {
+  spec: UIInputSpec
+  status: UIInputStatus | undefined
+  value: string | undefined
+  setValue: (name: string, value: string) => void
+}
+
+function ApiButtonInput(props: ApiButtonInputProps) {
```
nicks, line 52: i'm surprised you didn't inline this; i'm not sure it's adding much value as an independent component.

nicks (approved):

Issue #4885, opened 2021-08-24
resources in sidebar jumping around in order


milas:
Got a report from a user that the resources in the sidebar started changing order endlessly at a pretty high speed.

At the time it happened, they did have several resources starred and the active resource was one of the starred ones, but that might be a red herring. Additionally, many services were in the process of building/deploying or pending state.

Restarting Tilt resolved the issue.

Version: 0.22.4

landism:
Theoretically resources have an 'order' attached to them, which the UI sorts by and keeps them stable.

I don't have a good guess as to what's causing the jumping. Maybe the backend failing to set the order field, or setting it inconsistently, or maybe the frontend is not respecting the order field appropriately. If a user observes this and they can grab some output from `tilt get uiresource -ojson --watch` while it's happening, that should help narrow down where the problem is.

Depending on what's going wrong, #4886 might effectively fix it so that it's no longer evident to users without actually getting at the cause

PR #4886, opened 2021-08-24
web: fix object update sorting when Orders are equal

landism:
I spotted this when briefly looking into #4885. I don't know if it will affect that issue. If it is caused by resources having equal Order, then it will eliminate the "jumping around" symptom, which will be nice for the user but make the actual issue harder to debug.

This makes the sorting added in #4869 redundant, so removing that.

Also updated the UIResource sorting test to be less likely to pass by chance.


nicks (approved):

milas (approved):
Nice find!

PR #4898, opened 2021-08-27
web: fix navbar custom buttons jumping when clicked

landism:
### Problem

1. When a custom button is clicked, we temporarily disable it.
2. `InstrumentedButton` hardcodes all buttons to `variant="outlined".
3. In the header, we override the style to not have a border, but when the button enters the disabled state, the MUI style takes over and adds an invisible border, causing the nav bar to shift by a pixel.

This was introduced in #4874

### Solution

Plumb ButtonProps up from InstrumentedButton through ApiButton so that we can use `variant="contained"` in the navbar.

(the majority of the diff is renaming `ApiButtonProps.button => uiButton`)


lizzthabet (approved):
the click jump is fixed! however, i did notice a number of other style and functionality goofs when testing locally. if you want to merge this in and address these in a follow-up, that'd work for me.

#### nav bar custom buttons:
- title display on hover overlaps and can lead the user to click the wrong button
- (very subtle) drop shadow on buttons that shouldn't be there
- ripple displays on click and shouldn't be there
- icon color changes to black ("flickers") on click, though this could be related to the disabled/loading state
- the click area is a little strange for the button groups. the click area is large for the command button itself, but small for the dropdown button. i think we could solve this by making the size of the command button / button group smaller, but i didn't play around with this solution.

^ here i just added a background color to each individual button and changed the color on the header bar to make it easier to see the shadows


#### table and detail view custom buttons:

- focus border is only three sides
- hover border needs transition animation for color on right side (to align with other three sides of border)
- ripple displays on click and shouldn't be there
- icon color changes to black ("flickers") on click, though this could be related to the disabled/loading state

landism:
> if you want to merge this in and address these in a follow-up, that'd work for me.

yep! filed ch12672 to track

PR #4906, opened 2021-08-30
web: add InstrumentedTextField

landism:
Extract `OverviewSidebarOptions`' `TextField` instrumentation to a separate component, since we have other TextFields we want to instrument.

### DIFF
--- a/web/src/instrumentedComponents.tsx
+++ b/web/src/instrumentedComponents.tsx
@@ -1,23 +1,35 @@
-import { Button, ButtonProps } from "@material-ui/core"
-import React from "react"
+import {
+  Button,
+  ButtonProps,
+  debounce,
+  TextField,
+  TextFieldProps,
+} from "@material-ui/core"
+import React, { useMemo } from "react"
 import { AnalyticsAction, incr, Tags } from "./analytics"
 
-export type InstrumentedButtonProps = ButtonProps & {
+// Shared components that implement analytics
+// 1. Saves callers from having to implement/test analytics for every interactive
+//    component.
+// 2. Allows wrappers to cheaply require uses specify analytics params.
+
+type InstrumentationProps = {
   analyticsName: string
   analyticsTags?: Tags
 }
 
-export function InstrumentedButton(props: InstrumentedButtonProps) {
+export function InstrumentedButton(props: ButtonProps & InstrumentationProps) {
   const { analyticsName, analyticsTags, onClick, ...buttonProps } = { ...props }
-  const tags = { action: AnalyticsAction.Click, ...(analyticsTags ?? {}) }
-  const instrumentedOnClick = (
-    e: React.MouseEvent<HTMLButtonElement, MouseEvent>
-  ) => {
-    incr(analyticsName, tags)
+  const instrumentedOnClick: typeof onClick = (e) => {
     if (onClick) {
       onClick(e)
     }
+    incr(analyticsName, {
+      action: AnalyticsAction.Click,
+      ...(analyticsTags ?? {}),
+    })
   }
+
   return (
     <Button
       variant="outlined"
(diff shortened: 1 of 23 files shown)


web/src/instrumentedComponents.tsx
```diff
@@ -1,23 +1,35 @@
...
+  TextFieldProps,
+} from "@material-ui/core"
+import React, { useMemo } from "react"
 import { AnalyticsAction, incr, Tags } from "./analytics"
 
-export type InstrumentedButtonProps = ButtonProps & {
+// Shared components that implement analytics
+// 1. Saves callers from having to implement/test analytics for every interactive
+//    component.
+// 2. Allows wrappers to cheaply require uses specify analytics params.
```
lizzthabet, line 14: i'm not totally sure what your second note is trying to say 😅 maybe:
```suggestion
// 2. Allows wrappers to cheaply require and use specific analytics params.
```
?

web/src/instrumentedComponents.tsx
```diff
@@ -27,3 +39,33 @@ export function InstrumentedButton(props: InstrumentedButtonProps) {
...
+  // we have to memoize the debounced function so that incrs reuse the same debounce timer
+  const debouncedIncr = useMemo(
+    () =>
+      // debounce so we don't send analytics for every single keypress
+      debounce(() => {
+        incr(analyticsName, {
+          action: AnalyticsAction.Edit,
+          ...(analyticsTags ?? {}),
+        })
+      }, 5000),
```
lizzthabet, line 59: (optional suggestion) it might be nice to extract this time into a `const` for readability and for use in the tests.

also, five seconds seems a little long to wait for text input to change. i think in other places, we're using half a second?

lizzthabet (approved):
lgtm! were you also planning on refactoring the [`FilterTermTextField`](https://github.com/tilt-dev/tilt/blob/master/web/src/OverviewActionBar.tsx#L445-L490) to use this component? (the logic may be a little bit more complicated, but happy to pair on it if you'd like!)

web/src/instrumentedComponents.tsx
landism: > it might be nice to extract this time into a const for readability and for use in the tests.

yeah, fair

> also, five seconds seems a little long to wait for text input to change. i think in other places, we're using half a second?

Those are for the purposes of re-rendering / applying filters, yeah? I think it makes sense to have dramatically different debounces for those.
e.g., if the user spends 5 seconds typing in a new filter value:
* we want to incrementally update the filtered results as they type, so they can see the impact of what they've typed and judge whether they should type more to further narrow results
* the analytics are invisible to the user and logging four different "change" events doesn't seem particularly useful

landism: hm, or maybe I misunderstood the comment?

lizzthabet: ah, yeah! that's a good point. i forgot about the debounce being used both to apply the filter and log the event. it makes sense to keep the filter term field logic on its own. then the goal of the five second delay here is to attempt to log only one event per input interaction, or rather to prevent extraneous multiple events from a single interaction.

web/src/instrumentedComponents.tsx
landism: ah, was trying to use "use" as a noun, like "require uses of this to specify analytics params" or "require that things that use this specify analytics params", but yeah, that sucks

landism:
> were you also planning on refactoring the FilterTermTextField to use this component? (the logic may be a little bit more complicated, but happy to pair on it if you'd like!)

I figured that could be a separate PR, but might as well throw it in here. It shouldn't be complicated - the goal of this component is you just swap `TextField` out for `InstrumentedTextField` and add an `analyticsName`. I guess FilterTermTextField will be a good test of this, to see if the two debounces play nicely.

PR #4907, opened 2021-08-30
web: make info buttons hideable by user

landism:
### Problem

Info tooltips are really helpful for introducing users to new features! However, they can be kind of distracting and annoying if you're already familiar with their information.

### Solution

Add a button to the info tooltip to allow the user to hide the info button.

image

When clicked, a bool will be written to localStorage and the user will no longer see that little (i) or the tooltip (unless they do something like switch browsers or clear local storage).

Notes:
1. There is currently no way to undo this other than digging into your browser's localStorage.
2. Until now, all of the info Tilt stores in localStorage has had its key namespaced by the path to the currently running Tiltfile. That seems like it'd be much more annoying than useful for this feature, so dismissals do not use the Tiltfile in their localStorage key.

### DIFF
--- a/web/src/LocalStorage.tsx
+++ b/web/src/LocalStorage.tsx
@@ -33,22 +33,35 @@ export function accessorsForTesting<S>(name: string) {
   }
 }
 
-// Like `useState`, but backed by localStorage and namespaced by the tiltfileKey
-// maybeUpgradeSavedState: transforms any state read from storage - allows, e.g., filling in default values for
-//                         fields added since the state was saved
+// Like `useState`, but backed by localStorage
+// options:
+// - maybeUpgradeSavedState: transforms any state read from storage - allows, e.g., filling in default values for
+//                           fields added since the state was saved
+// - keyedByTiltfile: this state is saved per tiltfile - defaults to true
 export function usePersistentState<S>(
   name: string,
   defaultValue: S,
-  maybeUpgradeSavedState?: (state: S) => S
+  options?: {
+    maybeUpgradeSavedState?: (state: S) => S
+    keyedByTiltfile?: boolean
+  }
 ): [state: S, setState: Dispatch<SetStateAction<S>>] {
-  const tiltfileKey = useContext(tiltfileKeyContext)
-  let [state, setState] = useStorageState<S>(
-    localStorage,
-    makeKey(tiltfileKey, name),
-    defaultValue
-  )
-  if (maybeUpgradeSavedState) {
-    state = maybeUpgradeSavedState(state)
+  let state: S
+  let setState: Dispatch<SetStateAction<S>>
+  if (options?.keyedByTiltfile !== false) {
+    const tiltfileKey = useContext<TiltfileKey>(tiltfileKeyContext)
+
+    ;[state, setState] = useStorageState<S>(
+      localStorage,
+      makeKey(tiltfileKey, name),
+      defaultValue
+    )
+  } else {
+    ;[state, setState] = useStorageState<S>(localStorage, name, defaultValue)
+  }
+
+  if (options?.maybeUpgradeSavedState) {
+    state = options.maybeUpgradeSavedState(state)
   }
   return [state, setState]
 }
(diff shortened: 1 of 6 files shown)


web/src/Tooltip.tsx
```diff
@@ -55,22 +64,61 @@ const InfoIcon = styled(InfoSvg)`
...
+    content = (
+      <>
+        <div>{title}</div>
+        <DismissButton
+          size="small"
+          analyticsName={"ui.web.dismissTooltip"}
+          analyticsTags={{ tooltip: props.dismissId }}
+          onClick={() => setDismissed(true)}
+        >
+          Hide this info button
```
landism, line 111: This is a bit longer than some of my earlier iterations (like "dismiss" or "hide this"). Since this is non-undoable action, I think it's extra important the user understands what clicking the button will do.  Happy to hear alternatives (or cheap ways to allow undo!)

lizzthabet: wdyt about `Don't show again`? or perhaps `Don't show tip again`? i think that "don't show" language is kinda standard for software tutorials and users will understand what it means!

web/src/LocalStorage.tsx
```diff
@@ -33,22 +33,35 @@ export function accessorsForTesting<S>(name: string) {
...
 }
 
-// Like `useState`, but backed by localStorage and namespaced by the tiltfileKey
-// maybeUpgradeSavedState: transforms any state read from storage - allows, e.g., filling in default values for
-//                         fields added since the state was saved
+// Like `useState`, but backed by localStorage
+// options:
+// - maybeUpgradeSavedState: transforms any state read from storage - allows, e.g., filling in default values for
+//                           fields added since the state was saved
+// - keyedByTiltfile: this state is saved per tiltfile - defaults to true
```
lizzthabet, line 40: isn't there already a pattern for `global` tilt settings used in the font settings? i'm not opposed to configuring this hook to key globally, but i do find the implementation difficult to read.

web/src/Tooltip.tsx
```diff
@@ -55,22 +64,62 @@ const InfoIcon = styled(InfoSvg)`
...
 `
+const DismissButton = styled(InstrumentedButton)`
+  ${mixinResetButtonStyle};
+
+  .MuiButton-label {
+    font-size: ${FontSize.smallester};
+    font-style: italic;
+    color: ${Color.grayLight};
+  }
+`
```
lizzthabet, line 75: playing around with this locally, i kinda like the dismiss button right-aligned, since it better separates it from the tip content. i'd be curious to hear what @hyu or @SurbhiSGupta think if they have time to look, but i'd be sold either way!

Screen Shot 2021-08-31 at 2 11 42 PM

lizzthabet (commented):
i think this is a great improvement for the info tips! i had one larger question about the global local storage settings, and i did notice some funny behavior when dismissing a tip on table view (just ignore the fact that they're all open lol):


^ i wonder also if that "dismiss" click is propagating to the column sort? not a big deal if it is, but worth noting.

web/src/LocalStorage.tsx
landism: so there is! yeah that seems a lot simpler

web/src/Tooltip.tsx
landism: Yes! Do you have the styling for that handy? That's what I initially went for and spent 15 minutes trying to get that to work and decided it wasn't worth the effort at the moment.

landism:
> i did notice some funny behavior when dismissing a tip on table view

Found this on SO, sigh...
> Some browsers don't support the storage event, and most of the browsers that do support it will only call it when the storage is changed by a different window.

web/src/Tooltip.tsx
landism: good!

landism:
> i did notice some funny behavior when dismissing a tip on table view

I spent two hours trying to fix this and gave up and filed https://app.clubhouse.io/windmill/story/12654/modifying-usepersistentstate-state-doesn-t-update-other-components-in-the-same-page-using-the-same-key

It currently only manifests in table header tooltips with resource grouping enabled, and only lasts until refresh.

web/src/Tooltip.tsx
landism: oh that wasn't so bad!

landism:
Here's how it looks now:

image

web/src/ResourceGroups.tsx
```diff
@@ -131,7 +131,13 @@ export function ResourceGroupsInfoTip(props: InfoTooltipProps) {
 
   return (
     <BottomLeftContainer>
-      <TiltInfoTooltip title={tooltipInfo} displayShadow={true} {...props} />
+      <TiltInfoTooltip
+        title={tooltipInfo}
+        displayShadow={true}
+        idForIcon={props.idForIcon}
+        dismissId="resource-groups"
+        open={true}
```
lizzthabet, line 139: i think you want to remove `open={true}` 😅

lizzthabet (approved):
looks great! i noted one leftover piece of testing code that should be removed before merging, but otherwise g2g.

PR #4908, opened 2021-08-31
api: move tiltfile args to the API

nicks:
Hello @landism,

Please review the following commits I made in branch nicks/api:

743419fd50f9a40dceca773291bfdf5eb5c14247 (2021-08-31 10:34:45 -0400)
api: move tiltfile args to the API


landism (approved):
I'm a bit surprised to see this on ExtensionSpec! I thought:
* In the short term, extensions should ignore args, and only the main tiltfile should get them, which seems like the reverse of this.
* In the long term, we want to kill args anyway (maybe that's more about `tilt args` and the `config` tiltfile api, though)

If you just wanna say "this will make sense in a PR or two", that's cool.

nicks:
the trick: extensions will have independent args! So the kubefwd extension might have the args:

```
v1alpha1.extension('kubefwd', args=['--namespaces=my-ns'])
```

and `tilt args` will become a light wrapper around a patch on the main tiltfile spec.

PR #4909, opened 2021-08-31
web: add info tooltip to widgets column header

landism:
image

(will be dismissable with #4907)

Copy advice welcome!

### DIFF
--- a/web/src/OverviewTable.tsx
+++ b/web/src/OverviewTable.tsx
@@ -534,18 +534,15 @@ const columnNameToInfoTooltip: {
   ),
   Widgets: (
     <>
-      <div>You can add custom buttons here to easily perform actions.</div>
-      <div>
-        See{" "}
-        <a
-          href={linkToTiltDocs(TiltDocsPage.CustomButtons)}
-          target="_blank"
-          rel="noopener noreferrer"
-        >
-          custom buttons docs
-        </a>
-        .
-      </div>
+      Buttons can be added to resources to easily perform custom actions. See{" "}
+      <a
+        href={linkToTiltDocs(TiltDocsPage.CustomButtons)}
+        target="_blank"
+        rel="noopener noreferrer"
+      >
+        buttons docs
+      </a>
+      .
     </>
   ),
 }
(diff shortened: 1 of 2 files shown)


web/src/OverviewTable.tsx
```diff
@@ -532,6 +532,22 @@ const columnNameToInfoTooltip: {
...
+        <a
+          href={linkToTiltDocs(TiltDocsPage.CustomButtons)}
+          target="_blank"
+          rel="noopener noreferrer"
+        >
+          custom buttons docs
+        </a>
+        .
+      </div>
+    </>
```
lizzthabet, line 549: wdyt about 
```
Buttons can be added to resources to easily perform custom actions. See button docs.
```
? maybe? i don't think it needs a line break or needs to be wrapped in a `div`.

web/src/OverviewTable.tsx
```diff
@@ -532,6 +532,22 @@ const columnNameToInfoTooltip: {
       .
     </>
   ),
+  Widgets: (
```
lizzthabet, line 535: (maybe bikeshedding?) this tooltip text is making me wonder if this column should just be named `Custom actions`? i forgot what naming we originally talked about... but honestly, feel free to ignore this suggestion / thought.

lizzthabet (approved):
🎉 ℹ️

web/src/OverviewTable.tsx
```diff
@@ -532,6 +532,22 @@ const columnNameToInfoTooltip: {
...
+    <>
+      <div>You can add custom buttons here to easily perform actions.</div>
+      <div>
+        See{" "}
```
milas, line 539: I'm both surprised I've never seen this trick before and also wish I hadn't 🙈

milas (approved):

web/src/OverviewTable.tsx
landism: I'm not a fan of putting "custom" in the names of things, but haven't really had a good conversation about it and might be wrong. I'm happy to talk about it, but am also pretty sure it's not an urgent thing and we can change it later, so not going to block.

web/src/OverviewTable.tsx
landism: Thanks!

PR #4914, opened 2021-08-31
web: increase font size and add padding between table cells in table view

lizzthabet:
This PR makes two style changes I chatted about with @SurbhiSGupta: 
* increase font size so text is easier to read
* add a little bit of padding in between columns / table cells so text from different columns doesn't run into other columns

before:
Screen Shot 2021-08-31 at 3 59 29 PM

after:
Screen Shot 2021-08-31 at 3 58 14 PM

Screen Shot 2021-08-31 at 3 41 57 PM

### DIFF
--- a/web/src/OverviewTable.tsx
+++ b/web/src/OverviewTable.tsx
@@ -114,6 +114,11 @@ const ResourceTable = styled.table`
     padding-right: ${SizeUnit(1)};
   }
 
+  td + td {
+    padding-left: ${SizeUnit(1 / 4)};
+    padding-right: ${SizeUnit(1 / 4)};
+  }
+
   &.isGroup {
     border: 1px ${Color.grayLighter} solid;
     border-radius: 0 ${SizeUnit(1 / 4)};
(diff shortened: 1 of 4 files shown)


landism:
LGTM other than one thing:

On my screen, this is forcing successful build durations into two lines, which feels a bit weird:

image

Maybe bump that [width](https://github.com/tilt-dev/tilt/blob/388e307b59e2b1f7fb931cf44fd72d4234703d49/web/src/OverviewTable.tsx#L495) up by a bit?
(also we should probably `text-align: left` that?)

landism (approved):

lizzthabet:
> LGTM other than one thing:
> 
> On my screen, this is forcing successful build durations into two lines, which feels a bit weird:
> 
> image
> 
> Maybe bump that [width](https://github.com/tilt-dev/tilt/blob/388e307b59e2b1f7fb931cf44fd72d4234703d49/web/src/OverviewTable.tsx#L495) up by a bit?
> (also we should probably `text-align: left` that?)

yeah! great catch. i made that column a bit wider and left-aligned the content.

PR #4915, opened 2021-08-31
configs: replicate trigger queue to apiserver

nicks:
Hello @milas,

Please review the following commits I made in branch nicks/trigger:

a34eb5d38201486dd02546c2043495171cadb50b (2021-08-31 16:14:51 -0400)
configs: replicate trigger queue to apiserver


nicks:
That's a good question!

I think eventually, all the restart buttons should use the UIButton API. In a world with independent controllers, I don't think it makes sense to have a central trigger queue.

nicks:
(but of course, we have to do something in the transitionary period)

PR #4916, opened 2021-09-01
docker: exit early under more conditions in pruner

milas:
Recently received a pprof from a user who was experiencing high CPU
while Tilt wasn't really doing anything.

One of the offenders is the Docker Pruner: due to being an older
subscriber, it didn't have the newer optimization to skip running on
log-only store events, so would fire at a high-rate even when it
had no work to do.

Beyond that, collecting image references is surprisingly expensive
due to using a bunch of regex stuff under the hood. While it should
get called orders of magnitude less due to skipping log-only updates,
further changes are included to try and optimize this to exit early
without doing this work if possible.

### DIFF
--- a/internal/engine/dockerprune/docker_pruner.go
+++ b/internal/engine/dockerprune/docker_pruner.go
@@ -63,58 +63,59 @@ func (dp *DockerPruner) SetUp(ctx context.Context, _ store.RStore) error {
 	return nil
 }
 
-func (dp *DockerPruner) OnChange(ctx context.Context, st store.RStore, _ store.ChangeSummary) error {
-	if dp.disabledForTesting || dp.disabledOnSetup {
+// OnChange determines if any Tilt-built Docker images should be pruned based on settings and invokes the pruning
+// process if necessary.
+//
+// Care should be taken when modifying this method to not introduce expensive operations unless necessary, as this
+// is invoked for EVERY store action change batch. Because of this, the store (un)locking is done somewhat manually,
+// so care must be taken to avoid locking issues.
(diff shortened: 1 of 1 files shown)


internal/engine/dockerprune/docker_pruner.go
```diff
@@ -63,58 +63,59 @@ func (dp *DockerPruner) SetUp(ctx context.Context, _ store.RStore) error {
...
-
-	// Don't prune while we're building or about to build something, in case of weird side-effects.
-	if buildInProg || nextToBuild != "" {
+	// Exit early if possible if any of the following is true:
+	// 	* Pruning is disabled entirely
+	// 	* Engine is currently building something
+	// 	* There are NO `docker_build`s in the Tiltfile
+	// 	* Something is queued for building
+	if !settings.Enabled || len(state.CurrentlyBuilding) > 0 || !state.HasDockerBuild() || buildcontrol.NextManifestNameToBuild(state) != "" {
+		st.RUnlockState()
```
landism, line 85: can these all be moved into a `defer` after the `RLockState`?

milas: The issue would be holding the lock during the _actual_ call to `Prune`.

That said, since subscribers already run non-blocking in their own goroutines, we could fire off the actual prune in a goroutine [would need to add some slight extra synchronization then to make sure we don't run it multiple times simultaneously].

Or could do something kinda hacky with a `sync.Once(st.RUnlockState())` - call it via `defer` or early in the path that actually prunes.

Thoughts?

landism: I'd lean towards that all being more complexity than it's worth. I missed the lock on Prune issue.

PR #4917, opened 2021-09-01
k8s: allow rancher-desktop context by default

milas:
[Rancher Desktop](https://rancherdesktop.io/) is a new local K8s distribution.

See also tilt-dev/tilt.build#931 for a blog post explaining how to use it with Tilt.


milas:
I've never actually added a new context before so I'm not sure if there is anything I missed, but I was able to run a local project using Rancher Desktop with these changes at least!

nicks (approved):
i look forward to adding this to ctlptl too! :sweat_smile:

PR #4920, opened 2021-09-02
controllers: factor out restarton/starton interpreters

nicks:
Hello @landism, @milas,

Please review the following commits I made in branch nicks/restarton:

81bb74ae84eb000ba55ef8aeeba4b8c1aa5d7fa7 (2021-09-02 13:57:26 -0400)
controllers: factor out restarton/starton interpreters

### DIFF
--- a/internal/controllers/apis/restarton/restarton.go
+++ b/internal/controllers/apis/restarton/restarton.go
@@ -1,0 +1,147 @@
+package restarton
+
+import (
+	"context"
+	"time"
+
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/tilt-dev/tilt/internal/sliceutils"
+	"github.com/tilt-dev/tilt/pkg/apis/core/v1alpha1"
+)
+
+// Fetch all the buttons that this object depends on.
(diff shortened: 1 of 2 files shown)


internal/controllers/apis/restarton/restarton.go
```diff
@@ -0,0 +1,141 @@
...
+			continue
+		}
+		lastEventTime := b.Status.LastClickedAt
+		if lastEventTime.Time.After(cur) {
+			cur = lastEventTime.Time
+			latestButton = b
+		}
+	}
+
+	return cur, latestButton
```
milas, line 117: I wonder if the return here should be a union type that has nil-able button/fw fields - it seems weird that in the case of FW being the last triggering event, you just get a time and `nil`. (It could also be `time.Time, runtime.Object` I guess but that would probably make it a PITA to actually use)

milas (approved):

internal/controllers/apis/restarton/restarton.go
nicks: ya...let me leave some comments on this.

i think this is more of a product ambiguity...if you have a cmd that reads inputs from a button, but the most recent trigger is a filewatch, what should happen? does the filewatch imply 0 inputs, or does the filewatch imply "use the most recent button input"

PR #4921, opened 2021-09-02
configs: also store trigger queue reasons in the apiserver

nicks:
Hello @milas,

Please review the following commits I made in branch nicks/triggerqueue:

9f0907ffe54df9478a29535a177669ca5d0c21a2 (2021-09-02 17:25:37 -0400)
configs: also store trigger queue reasons in the apiserver


milas (approved):
I wonder if it's worth renaming the other keys as well, e.g. `%d-name` or something so that the pattern is consistent. (That said, it's pretty straight-forward as-is.)

nicks:
sure, that seems reasonable to me. done!

PR #4922, opened 2021-09-03
controllers: move all rebuild triggers to the tiltfile controller

nicks:
Hello @milas, @landism,

Please review the following commits I made in branch nicks/configs2:

0fbfb9701b233c56d9db5a343618fa8eab30676f (2021-09-02 20:22:46 -0400)
controllers: move all rebuild triggers to the tiltfile controller

### DIFF
--- a/internal/controllers/core/tiltfile/reconciler.go
+++ b/internal/controllers/core/tiltfile/reconciler.go
@@ -20,6 +20,8 @@ import (
 	"github.com/pkg/errors"
 
 	"github.com/tilt-dev/tilt/internal/controllers/apicmp"
+	"github.com/tilt-dev/tilt/internal/controllers/apis/configmap"
+	"github.com/tilt-dev/tilt/internal/controllers/apis/restarton"
 	"github.com/tilt-dev/tilt/internal/controllers/indexer"
 	"github.com/tilt-dev/tilt/internal/docker"
 	"github.com/tilt-dev/tilt/internal/engine/buildcontrol"
@@ -41,6 +43,7 @@ type Reconciler struct {
 	indexer      *indexer.Indexer
 	buildSource  *BuildSource
 	engineMode   store.EngineMode
+	loadCount    int // used to differentiate spans
 
 	runs map[types.NamespacedName]*runStatus
 }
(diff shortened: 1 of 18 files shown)


internal/controllers/core/tiltfile/reconciler.go
```diff
@@ -151,6 +165,69 @@ func (r *Reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
...
+// Modeled after BuildController.needsBuild and NextBuildReason(). Check to see that:
+// 1) There's currently no Tiltfile build running,
+// 2) There are pending file changes, and
+// 3) Those files have changed since the last Tiltfile build
+//    (so that we don't keep re-running a failed build)
+// 4) OR the command-line args have changed since the last Tiltfile build
+// 5) OR user has manually triggered a Tiltfile build
+//
+// In some cases, there may be race conditions where a tiltfile has finished executing,
+// but its results haven't bee
```
milas, line 177: Curious to know the full details of this race condition 😅

internal/controllers/core/tiltfile/reconciler.go
```diff
@@ -151,6 +165,69 @@ func (r *Reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
...
+	}
+
+	if step == runStepNone {
+		reason = reason.With(model.BuildReasonFlagInit)
+	} else {
+		filesChanged = restarton.FilesChanged(tf.Spec.RestartOn, fileWatches, lastStartTime)
+		if len(filesChanged) > 0 {
+			reason = reason.With(model.BuildReasonFlagChangedFiles)
+		} else if lastRestartEvent.After(lastStartTime) {
+			reason = reason.With(model.BuildReasonFlagTriggerUnknown)
```
milas, line 198: I'm guessing there's broader implications (e.g. handling in UI), but should we add some sort of `BuildReasonFlagTriggerExternal` as a catch-all in the future?

milas (approved):

internal/controllers/core/tiltfile/reconciler.go
milas: Ah nevermind - I was thinking that "Unknown" was the default/0 value, i.e. would more indicative of a logic error that we forgot to populate it, but I see we already use(d) it as the umbrella, so it's really as you described it 👍

PR #4924, opened 2021-09-03
engine: remove old opencensus/opentelemetry metrics code

nicks:
Hello @landism,

Please review the following commits I made in branch nicks/ot:

6036097f9a3e374ac00588675af6f7de44ae9b08 (2021-09-03 11:51:34 -0400)
engine: remove old opencensus/opentelemetry metrics code
the ecosystem has evolved substantially since I wrote this code. Even if we were
going to revive it, we wouldn't end up using it this way. We'd either use:

- a more CLI-focused opentelemetry library, or
- the opentelemetry built into apiserver itself (https://kubernetes.io/blog/2021/09/03/api-server-tracing/)


internal/cli/cli.go
```diff
@@ -97,21 +96,14 @@ func preCommand(ctx context.Context, cmdName model.TiltSubcommand) (context.Cont
...
-	stats.Record(ctx, CommandCountMeasure.M(1))
-
 	a, err := wireAnalytics(l, cmdName)
 	if err != nil {
 		l.Errorf("Fatal error initializing analytics: %v", err)
 		os.Exit(1)
 	}
 
 	ctx = tiltanalytics.WithAnalytics(ctx, a)
+	cleanup := func() error { return nil }
```
landism, line 106: it looks like `cleanup` was only used for metrics and we can just get rid of the return value

internal/user/api.go
```diff
@@ -1,13 +1,9 @@
 package user
```
landism, line 1: FFTI - Did you consider deleting this whole package? I would guess if we decided to add prefs we'd want to do something more like `ExtensionRepo`s.

landism (approved):

internal/user/api.go
nicks: good point!

internal/cli/cli.go
nicks: done

PR #4925, opened 2021-09-03
tiltfile: add extension(args=...)

nicks:
Hello @landism,

Please review the following commits I made in branch nicks/extargs:

51a77545aa083123f4ef67964b57a5279389858a (2021-09-03 13:14:50 -0400)
tiltfile: add extension(args=...)

### DIFF
--- a/internal/tiltfile/value/string.go
+++ b/internal/tiltfile/value/string.go
@@ -58,7 +58,7 @@ func (s *StringList) Unpack(v starlark.Value) error {
 	for iter.Next(&item) {
 		sv, ok := AsString(item)
 		if !ok {
-			return fmt.Errorf("list should contain only strings, but element %q was of type %s", item.String(), item.Type())
+			return fmt.Errorf("value should contain only strings, but element %q was of type %s", item.String(), item.Type())
 		}
 		*s = append(*s, sv)
 	}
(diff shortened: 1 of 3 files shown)


internal/tiltfile/value/string.go
```diff
@@ -32,6 +32,41 @@ func AsString(x starlark.Value) (string, bool) {
 	return starlark.AsString(x)
 }
 
+type StringList []string
+
+var _ starlark.Unpacker = &StringList{}
+
+// Unpack an argument that can either be expressed as
+// a string or as a list of strings.
```
landism, line 40:
```suggestion
// a tuple or list of strings.
```

internal/tiltfile/value/string.go
```diff
@@ -32,6 +32,41 @@ func AsString(x starlark.Value) (string, bool) {
...
+	default:
+		return fmt.Errorf("value should be a string or List or Tuple of strings, but is of type %s", v.Type())
+	}
+
+	defer iter.Done()
+	var item starlark.Value
+	for iter.Next(&item) {
+		sv, ok := AsString(item)
+		if !ok {
+			return fmt.Errorf("list should contain only strings, but element %q was of type %s", item.String(), item.Type())
```
landism, line 62:
```suggestion
			return fmt.Errorf("value should contain only strings, but element %q was of type %s", item.String(), item.Type())
```

landism (approved):

internal/tiltfile/value/string.go
nicks: fixed

PR #4926, opened 2021-09-03
cli: add a convenience function over extension repos

nicks:
Hello @milas,

Please review the following commits I made in branch nicks/generator:

e3220afe93feb6a3dd3c69f3e2eb0eb2bf2a4bd5 (2021-09-03 13:45:49 -0400)
cli: add a convenience function over extension repos

### DIFF
--- a/internal/cli/create_repo.go
+++ b/internal/cli/create_repo.go
@@ -1,0 +1,85 @@
+package cli
+
+import (
+	"context"
+	"time"
+
+	"github.com/spf13/cobra"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/tilt-dev/tilt/internal/analytics"
+	engineanalytics "github.com/tilt-dev/tilt/internal/engine/analytics"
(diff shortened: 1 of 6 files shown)


internal/cli/create_repo.go
```diff
@@ -0,0 +1,85 @@
...
+
+	c.helper.addFlags(cmd)
+
+	return cmd
+}
+
+func (c *createRepoCmd) run(ctx context.Context, args []string) error {
+	a := analytics.Get(ctx)
+	cmdTags := engineanalytics.CmdTags(map[string]string{})
+	a.Incr("cmd.create-repo", cmdTags.AsMap())
```
milas, line 62: Might be helpful to have a `url_type` tag with whether it's a `file`/git URL. (Only other possible data point that could be added would be a boolean indicating whether a ref was pinned, but I dunno if there's any value in that?)

milas (approved):

internal/cli/create_repo.go
nicks: i'm going to try to add some analytics deeper in the extension reconciler that i think will address exactly this

PR #4930, opened 2021-09-03
web: table view: allow sorting by status, fix sorting by endpoints

landism:
Basically just applying [this](https://github.com/tilt-dev/tilt/blob/844f193f6539966ea6c0c70aa619d04aea2b288f/web/src/OverviewTable.tsx#L490-L495) to the columns that weren't already returning primitives from accessor, following the pattern that seems to be working for uibuttons.


lizzthabet (approved):
looks great to me!

PR #4931, opened 2021-09-03
scripts: decode bytes to string properly

nicks:
Hello @landism,

Please review the following commits I made in branch nicks/pythonnnnnnn:

cbad4e25d41df073811475a24a92ee6679e3b6a6 (2021-09-03 17:21:25 -0400)
scripts: decode bytes to string properly

### DIFF
--- a/scripts/upload-assets.py
+++ b/scripts/upload-assets.py
@@ -21,9 +21,10 @@ parser.add_argument('version', help='A version string like "v1.2.3".')
 args = parser.parse_args()
 version = args.version
 if version == "latest":
-  version = str(subprocess.check_output([
+  version = subprocess.check_output([
     "git", "describe", "--tags", "--abbrev=0"
-  ])).strip()
+  ]).decode('utf-8').strip()
+
 dir_url = ("https://storage.googleapis.com/tilt-static-assets/%s/" % version)
 
 if args.clean:


scripts/upload-assets.py
```diff
@@ -21,9 +21,10 @@
 args = parser.parse_args()
 version = args.version
 if version == "latest":
-  version = str(subprocess.check_output([
+  version = subprocess.check_output([
     "git", "describe", "--tags", "--abbrev=0"
-  ])).strip()
+  ]).decode('utf-8').strip()
+
```
landism, line 27: we could drop this in here, too, which would protect both against weird `git describe` results and against weird versions passed as argv[0]:

(also need the `import re`)

```
if not re.match(r'^v\d+\.\d+\.\d+', version):
  sys.exit('error: version %r is not in format v#.#.#' % (version))
```

landism (approved):

scripts/upload-assets.py
landism: oh wait that's a bad idea - we set the version to SHAs on a most of these

nicks: i still have a dream that we can use the `go embed` stuff and get rid of a lot of the uploading (tho we'll still need it for snapshots)

PR #4936, opened 2021-09-07
web: fix borders on table view api buttons

landism:
Fix the focus/hover border weirdness spotted in [this PR comment](https://github.com/tilt-dev/tilt/pull/4898#pullrequestreview-742973668)

MUI is kind of annoying for this:
In a button group, each line between buttons is on the same 1px from each neighbor button. MUI does this by adding `margin-left: 1px;` to each button in the group other than the first.
Button 2's left border gets rendered on top of button 1's right border.
My understanding of the way this works in normal MUI is:
1. give the buttons a transparent background, so button 2's background doesn't wipe out button 1's right border
2. make the normal border color transparent, so that when button 2's left border is rendered on top of button 1's right border, you still see most of button 1's right border.

I'd previously tried to solve this with z-index, but then the transition on the right border is wonky.

> ripple displays on click and shouldn't be there

I wasn't able to observe this, which maybe means it's not worth worrying about? Or maybe my eyes are going bad and/or I wasn't looking at the right place.

> icon and button seems to darken/flicker on click, though this could be related to the disabled/loading state

I think I'm fine leaving this as-is, if you don't feel strongly about it.

### DIFF
--- a/web/src/OverviewButton.tsx
+++ b/web/src/OverviewButton.tsx
@@ -1,6 +1,7 @@
 import {
   AnimDuration,
   Color,
+  ColorRGBA,
   Font,
   mixinResetButtonStyle,
 } from "./style-helpers"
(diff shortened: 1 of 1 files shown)


lizzthabet (approved):
the borders look great! 🎉 i also did not see the ripple click this time around... i wonder if maybe the different branches i was reviewing / switching from locally caused an issue?

i'm okay with not addressing the flickering button for now! i'm pretty sure it's the `.Mui-disabled` class that's being added when the button is getting disabled. (it makes sense that `disabled` buttons would have different styles.)

Screen Shot 2021-09-07 at 4 16 59 PM

PR #4941, opened 2021-09-08
web: save uibutton inputs to localstorage

landism:
### Problem

If you set a button's input values in one view (i.e., table or resource) and then switch to the other view, the values are gone!

The values are saved via React's `useState`, and since in the other view, they're different components, each button has a separate set of values per view.

### Solution

Store button values in localStorage.

This also means they'll stick around between tilt runs, which seems like it's prooooobably a good thing? (pros: you still have the stuff you typed into that button three weeks ago, which might be what you want 99% of the time. cons: accidentally running with some values you haven't looked at in three weeks). We could switch to sessionStorage if it's a problem.


web/src/ApiButton.tsx
```diff
@@ -274,7 +275,9 @@ export function ApiButton(props: React.PropsWithChildren<ApiButtonProps>) {
   const { className, uiButton, ...buttonProps } = props
 
   const [loading, setLoading] = useState(false)
-  const [inputValues, setInputValues] = useState(new Map<string, any>())
+  const [inputValues, setInputValues] = usePersistentState<{
```
landism, line 278: it turns out `JSON.stringify` on a `Map` just returns `"{}"` :(

web/src/ApiButton.test.tsx
```diff
@@ -142,6 +149,43 @@ describe("ApiButton", () => {
     expect(actualStatus).toEqual(expectedStatus)
   })
 
+  it("reads options from local storage", async () => {
```
lizzthabet, line 152: do these tests need to be `async`? i'm curious because i didn't need to use it when i was writing tests with local storage, but i may be missing something!

web/src/ApiButton.tsx
lizzthabet: (there isn't a 😭 reaction)

lizzthabet (approved):
lgtm! a couple super minor / optional comments.

> cons: accidentally running with some values you haven't looked at in three weeks

i'm also curious about this, but i don't have any sense for what intuitive behavior users might expect. we'll see if there's any feedback! :)

web/src/ApiButton.test.tsx
landism: Nope, I just copied them from a test that happened to need to be async. Thanks!

PR #4943, opened 2021-09-08
docker-compose: log tailing robustness fixes

### LINKED ISSUE #4904: docker-compose: logs polled on exited containers
### Expected Behavior
When some docker-compose container exist correctly (exit code 0), Tilt should stop processing container logs. 

### Current Behavior
When d-c container exits, Tilt tries to follow its logs although the container is not running anymore.
`docker-compose logs --no-color -f containername` terminates immediately but Tilt tries to attach logs reader over and over.

### Steps to Reproduce
The minimal repoducer:
```
# Tiltfile:
docker_compose("./docker-compose.yml")

# ./docker-compose.yml
version: '3'
services:
  nginx:
    image: nginx
  one-off:
    image: alpine
    command: sh -c "date; echo Some command"
```
1. Run `tilt up`
2. Navigate to UI and check the `one-off` service
3. You'll see the very same line (with the same timestamp) repeating
4. check `pstree -a $(pgrep tilt)` (if available on your platform:
```
tilt up
  ├─docker-compose -f /tmp/test/docker-compose.yml logs --no-color -f nginx
  │   └─docker-compose -f /tmp/test/docker-compose.yml logs --no-color -f nginx
  │       └─2*[{docker-compose}]
  ├─docker-compose -f /tmp/test/docker-compose.yml logs --no-color -f one-off
  │   └─docker-compose -f /tmp/test/docker-compose.yml logs --no-color -f one-off
  └─18*[{tilt}]
```

CPU load is high because running `docker-compose logs` every second is a pretty expensive operation.

#### `tilt doctor` Output

```
$ tilt doctor
Tilt: v0.22.6, built 2021-08-27
System: linux-amd64
---
Docker
- Host: [default]
- Server Version: 20.10.7
- API Version: 1.41
- Builder: 2
---
Kubernetes
- Env: k3d
- Context: k3d-default
- Cluster Name: k3d-default
- Namespace: gooddata
- Version: v1.20.6+k3s1
---
```
#### Docker-compose
```
$ docker-compose version
docker-compose version 1.25.4, build 8d51620a
docker-py version: 4.1.0
CPython version: 3.7.5
OpenSSL version: OpenSSL 1.1.0l  10 Sep 2019
```
#### About Your Use Case
Besides long-living services (db, app, web etc.) we also have several containers declared in our `docker-compose.yml`, that will just run one command and then exit. They have various purposes, like loading DB and app fixtures, checking that everything works, and so on).

We'd like to use Tilt to improve productivity for our developers while leveraging existing docker-compose manifests used for local development.

milas:
The core issue this aims to solve is that for a stopped container,
`docker-compose logs` exits, and Tilt would continuously attempt
to re-launch it, resulting in high CPU usage due to repeatedly
invoking the command.

However, there were actually several intermingled issues that this
exposed/worsened. These mostly tie back to the behavior of
`docker-compose logs`: there is no built-in support for time
filtering (unlike `docker logs --since ...`), and since it exits
when a container stops, once invoked again, all the old logs will
be returned again. This means if a container crashes/restarts,
we'd re-emit all the old logs each time. Similarly, if there is
a "job" container that naturally runs to completion, whenever
re-triggered from the UI to run it, all the logs from prior
invocations would be re-logged. (This could also mean when attaching
on start up that you'd get a bunch of logs prior to the last
[re]start, which isn't super useful either.)

To solve this issue, `docker-compose logs` is now invoked with the
`--timestamps` argument, which prepends an RFC3339Nano timestamp
before each log line. Of course, there are caveats: first, Compose
v2 includes a whitespace before the timestamp and second, meta logs
from Compose itself about container events do NOT have a timestamp.
As a result, the attempt to parse/filter by the timestamp is _very_
forgiving. (There's an additional wrinkle which is that evidently
the `StartedAt` time reported by Docker for a container is not
precise, so you can actually get logs with earlier timestamps!)

Tying back to the original problem that once `docker-compose logs`
exits, Tilt tries to continuously re-invoke it, the watcher behavior
has been altered to more closely mirror the Pod log stream watcher.
When reading, if EOF is reached and/or the context is canceled,
it stops. The former case means that `docker-compose logs` exited
normally, indicating that the container is no longer running. The
latter case means the manifest is being torn down, so we no longer
want logs. Any errors during reading will cause it to retry, e.g.
if `docker-compose logs` crashes or is terminated abnormally.
(To support propagating process execution errors, the way that
logs are read has been adjusted slightly, which actually resolves
a race condition on reading logs and the `docker-compose logs`
process terminating, which ensures that short-lived containers
get all their logs!)

In the subscriber, when looking at already created watchers, if
the watcher has stopped running, a new one will only be created
if the corresponding container's start time is AFTER the last
time it started. This is the key to not continuously starting
up new `docker-compose logs` processes! (Finally.) We can't
key off container ID because that's consistent across restarts,
including both crashes and job containers that are run multiple
times without changes.

Finally, there is a fix for the Docker Compose event watcher
which was inadvertently broken (by yours truly) a while back.
It was not actually _ever_ running, so Tilt wouldn't notice
external changes (e.g. `docker-compose restart foo`), which
meant state was out of sync. This was not really noticeable
before since we were insanely aggressive about re-establishing
`docker-compose logs` sessions. After these fixes, however,
it meant that any container events _not_ initiated via Tilt
would go unnoticed, so the logs would not get resumed, as the
new container start time would never get picked up.

TL;DR Docker Compose logs should behave much nicer now both
      in terms of CPU usage as well as under various common
      scenarios like container restarts or job containers

Fixes #4904.

### DIFF
--- a/internal/dockercompose/client.go
+++ b/internal/dockercompose/client.go
@@ -50,6 +50,8 @@ func NewDockerComposeClient(env docker.LocalEnv) DockerComposeClient {
 
 func (c *cmdDCClient) Up(ctx context.Context, configPaths []string, serviceName model.TargetName, shouldBuild bool, stdout, stderr io.Writer) error {
 	var genArgs []string
+	// TODO(milas): this causes docker-compose to output a truly excessive amount of logging; it might
+	// 	make sense to hide it behind a special environment variable instead or something
 	if logger.Get(ctx).Level().ShouldDisplay(logger.VerboseLvl) {
 		genArgs = []string{"--verbose"}
 	}
@@ -118,13 +120,17 @@ func (c *cmdDCClient) Down(ctx context.Context, configPaths []string, stdout, st
 }
 
 func (c *cmdDCClient) StreamLogs(ctx context.Context, configPaths []string, serviceName model.TargetName) (io.ReadCloser, error) {
-	// TODO(maia): --since time
-	// (may need to implement with `docker log <cID>` instead since `d-c log` doesn't support `--since`
 	var args []string
 	for _, config := range configPaths {
 		args = append(args, "-f", config)
 	}
-	args = append(args, "logs", "--no-color", "-f", serviceName.String())
+	// NOTE: we can't practically remove "--no-color" due to the way that Docker Compose formats colorful lines; it
+	// 		 will wrap the entire line (including the \n) with the color codes, so you end up with something like:
+	//			\u001b[36mmyproject_my-container_1 exited with code 0\n\u001b[0m
+	// 		 where the ANSI reset (\u001b[0m) is _AFTER_ the \n, which doesn't play nice with our log segment logic
+	// 		 under some conditions - adding a final \n after stdout is closed would probably be sufficient given the
+	// 		 current pattern of how Compose colorizes stuff, but it's really not worth the headache to find out
+	args = append(args, "logs", "--no-color", "--no-log-prefix", "--timestamps", "--follow", serviceName.String())
 	cmd := c.dcCommand(ctx, args)
 	stdout, err := cmd.StdoutPipe()
 	if err != nil {
@@ -136,18 +142,27 @@ func (c *cmdDCClient) StreamLogs(ctx context.Context, configPaths []string, serv
 
 	err = cmd.Start()
 	if err != nil {
-		return nil, errors.Wrapf(err, "`docker-compose %s`",
+		return nil, errors.Wrapf(err, "cmd `docker-compose %s` failed to start",
 			strings.Join(args, " "))
 	}
 
+	// extra pipe is created so that we can ensure we've read all output from stdout _before_ calling Cmd::Wait()
+	// AND so that we can propagate errors (vs. just silently logging them) to the reader so they can choose to
+	// restart on failures
+	r, w := io.Pipe()
 	go func() {
-		err = cmd.Wait()
-		if err != nil {
-			logger.Get(ctx).Debugf("cmd `docker-compose %s` exited with error: \"%v\" (stderr: %s)",
-				strings.Join(args, " "), err, errBuf.String())
+		// prefer reporting process errors to copy errors as they're more likely to be the root cause
+		_, copyErr := io.Copy(w, stdout)
+		if cmdErr := cmd.Wait(); cmdErr != nil {
+			_ = w.CloseWithError(fmt.Errorf("cmd `docker-compose %s` exited with error: \"%v\" (stderr: %s)",
+				strings.Join(args, " "), cmdErr, errBuf.String()))
+		} else if copyErr != nil {
+			_ = w.CloseWithError(fmt.Errorf("failed to read docker-compose logs: %v", copyErr))
+		} else {
+			_ = w.Close()
 		}
 	}()
-	return stdout, nil
+	return r, nil
 }
 
 func (c *cmdDCClient) StreamEvents(ctx context.Context, configPaths []string) (<-chan string, error) {
--- a/internal/dockercompose/client.go
+++ b/internal/dockercompose/client.go
@@ -28,7 +28,7 @@ import (
 type DockerComposeClient interface {
 	Up(ctx context.Context, configPaths []string, serviceName model.TargetName, shouldBuild bool, stdout, stderr io.Writer) error
 	Down(ctx context.Context, configPaths []string, stdout, stderr io.Writer) error
-	StreamLogs(ctx context.Context, configPaths []string, serviceName model.TargetName) (io.ReadCloser, error)
+	StreamLogs(ctx context.Context, configPaths []string, serviceName model.TargetName) io.ReadCloser
 	StreamEvents(ctx context.Context, configPaths []string) (<-chan string, error)
 	Project(ctx context.Context, configPaths []string) (*types.Project, error)
 	ContainerID(ctx context.Context, configPaths []string, serviceName model.TargetName) (container.ID, error)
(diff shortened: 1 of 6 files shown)


internal/engine/dcwatch/event_watcher.go
```diff
@@ -26,12 +26,12 @@ func NewEventWatcher(dcc dockercompose.DockerComposeClient, docker docker.LocalC
 	}
 }
 
-func (w *EventWatcher) OnChange(ctx context.Context, st store.RStore, _ store.ChangeSummary) error {
-	if !w.watching {
+func (w *EventWatcher) OnChange(ctx context.Context, st store.RStore, summary store.ChangeSummary) error {
+	// TODO(nick): This should respond dynamically if the path changes.
+	if summary.IsLogOnly() || w.watching {
```
milas, line 31: This is the fix for the event watcher never running - I inadvertently flipped the condition back in #4364 🙀

internal/dockercompose/client.go
```diff
@@ -136,18 +142,27 @@ func (c *cmdDCClient) StreamLogs(ctx context.Context, configPaths []string, serv
...
+	// AND so that we can propagate errors (vs. just silently logging them) to the reader so they can choose to
+	// restart on failures
+	r, w := io.Pipe()
 	go func() {
-		err = cmd.Wait()
-		if err != nil {
-			logger.Get(ctx).Debugf("cmd `docker-compose %s` exited with error: \"%v\" (stderr: %s)",
-				strings.Join(args, " "), err, errBuf.String())
+		// prefer reporting process errors to copy errors as they're more likely to be the root cause
+		_, copyErr := io.Copy(w, stdout)
```
nicks, line 155: this seems very strange.

I've read a ton of code that uses the Go os/exec standard lib. 

I've never seen code that used it this way, in particular, blocking a cmd.Wait() on an io.Copy.

Can you talk a little bit more about the conditions where the copy failed but the command succeeded? That would be very surprising to me, based on my understanding of how Cmd/Stdout is implemented internally.

i think i would have expected something more like:
```
r, w := io.Pipe()
cmd.Stdout = w
go func() {
		err = cmd.Wait()
...
}()
return r, nil
```

and not use the built-in StdoutPipe() at all

internal/engine/runtimelog/docker_compose_log_manager.go
```diff
@@ -104,100 +121,173 @@ func (m *DockerComposeLogManager) OnChange(ctx context.Context, st store.RStore,
...
-	// cID             container.ID
-	// cName           container.Name
+	ctx            context.Context
+	cancel         func()
+	name           model.ManifestName
+	dc             model.DockerComposeTarget
+	startWatchTime time.Time
+}
+
+func (w *dockerComposeLogWatch) Done() bool {
```
nicks, line 166: i usually do `w.ctx.Err() != nil`, which is equivalent

internal/dockercompose/client.go
```diff
@@ -118,13 +120,17 @@ func (c *cmdDCClient) Down(ctx context.Context, configPaths []string, stdout, st
...
 		args = append(args, "-f", config)
 	}
-	args = append(args, "logs", "--no-color", "-f", serviceName.String())
+	// NOTE: we can't practically remove "--no-color" due to the way that Docker Compose formats colorful lines; it
+	// 		 will wrap the entire line (including the \n) with the color codes, so you end up with something like:
+	//			\u001b[36mmyproject_my-container_1 exited with code 0\n\u001b[0m
+	// 		 where the ANSI reset (\u001b[0m) is _AFTER_ the \n, which doesn't play nice with our log segment logic
+	// 		 under some conditions - adding a final \n after stdout is closed would probably be sufficient given the
+	// 		 current pattern of how Compose colorizes stuff, but it's really not worth the headache to find out
+	args = append(args, "logs", "--no-color", "--no-log-prefix", "--timestamps", "--follow", serviceName.String())
```
nicks, line 133: i wonder if we should give up on trying to get the logs from `docker-compose`...instead, we'd get the containers ID from docker-compose, then use the docker client logs API to stream logs

(i would generally expect the docker logs API to be way more fully baked, since that's what CRI uses -- and it'd look a lot closer to our k8s log streaming code)

internal/engine/dcwatch/event_watcher.go
nicks: :grimacing:

nicks (approved):

internal/dockercompose/client.go
milas: I _ALMOST_ did that - we already have the container ID in state, so it'd be very straightforward to use the Docker client logs API.

My decision for not was
   A. mistakenly thinking this wasn't going to be a huge can of worms until too late
   B. in theory, the new Compose V2 lets you `docker-compose up` against Azure ACI/AWS ECS, so the CLI will handle using appropriate cloud API for operations [that said, Live Update won't work in its current form since it uses Docker API directly vs shelling out to `docker-compose cp` - so not sure if this even a use case we want to support]

But, yeah, I think how we approach Docker Compose might be interesting to re-visit when we do the API migration for it. At the very least, if we pull in the core of Compose V2 as a dependency now that it's in Go, we can avoid all the pitfalls of shelling out via CLI. (Maybe it could become its own operator extension vs being built-in to keep dependencies under control?)

internal/dockercompose/client.go
milas: Agh, yeah, I originally had something similar to that and then was having issues with something unrelated and eventually ended up in this weird hybrid state 😧

Massively simplified in 058b0561b.

internal/dockercompose/client.go
mouchar: @milas For the record, I just tested this fix, and the need for docker-compose 1.28.0+ (because of `--no-log-prefix`) should be explicitly noted somewhere - I spent 20 minutes investigating why log-watching processes are still restarting.

milas: @mouchar Good catch - I have added that to the release notes. We currently don't have version detection logic but have considered adding it due to Compose v2 having known bugs - I'll make note of this as a case to detect as well.

PR #4944, opened 2021-09-08
extension: api-level event for extension load

nicks:
Hello @milas, @landism,

Please review the following commits I made in branch nicks/metrics2:

5efa2f2a0e7ad14253b02e703c525e5c80cd2fbe (2021-09-08 17:37:28 -0400)
extension: api-level event for extension load

f305a99fb91eafa71ae0a6622df89b8567a67e57 (2021-09-08 17:37:28 -0400)
extension: propagate extension arg changes correctly

### DIFF
--- a/internal/controllers/core/extension/reconciler.go
+++ b/internal/controllers/core/extension/reconciler.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"strings"
 
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -17,6 +18,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 	"sigs.k8s.io/controller-runtime/pkg/source"
 
+	"github.com/tilt-dev/tilt/internal/analytics"
 	"github.com/tilt-dev/tilt/internal/controllers/apicmp"
 	"github.com/tilt-dev/tilt/internal/controllers/indexer"
 	"github.com/tilt-dev/tilt/internal/ospath"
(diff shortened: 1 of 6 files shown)


internal/controllers/core/extension/reconciler.go
```diff
@@ -106,19 +110,31 @@ func (r *Reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
...
 	update := ext.DeepCopy()
 	mutateFn(&(update.Status))
 
 	if apicmp.DeepEqual(update.Status, ext.Status) {
-		return update, nil
+		return update, false, nil
 	}
 	err := r.ctrlClient.Status().Update(ctx, update)
-	return update, err
+	return update, true, err
```
milas, line 137: I feeeeel like this should only return `true` if `err == nil` but it doesn't actually matter since we always early return in that case

internal/controllers/core/tiltfile/reconciler.go
```diff
@@ -268,7 +268,11 @@ func (r *Reconciler) run(ctx context.Context, nn types.NamespacedName, tf *v1alp
 	}
 
 	tlr := r.tfl.Load(ctx, entry.TiltfilePath, userConfigState)
-	if tlr.Error == nil && len(tlr.Manifests) == 0 {
+
+	// If the user is executing an empty main tiltfile, that probably means
+	// they need a tutorial. For now, we link to that tutorial, but a more interactive
+	// system might make sense here.
+	if tlr.Error == nil && len(tlr.Manifests) == 0 && tf.Name == model.MainTiltfileManifestName.String() {
```
milas, line 275: 😅

milas (approved):

internal/controllers/core/extension/reconciler.go
nicks: done!

internal/controllers/core/tiltfile/reconciler.go
nicks: going to add more and more comments in the code talking about the better tutorial until it's launched :man_dancing:

PR #4946, opened 2021-09-09
web: add enable/disable button to resource view

landism:
The button currently doesn't actually do anything other than cycle through enable/disable/confirm.

It's hidden behind a disabled feature flag for now, so users won't see it.

image

### DIFF
--- a/web/src/OverviewActionBar.tsx
+++ b/web/src/OverviewActionBar.tsx
@@ -1,4 +1,4 @@
-import { debounce, InputAdornment, InputProps } from "@material-ui/core"
+import { debounce, Icon, InputAdornment, InputProps } from "@material-ui/core"
 import Menu from "@material-ui/core/Menu"
 import MenuItem from "@material-ui/core/MenuItem"
 import { PopoverOrigin } from "@material-ui/core/Popover"
@@ -17,6 +17,7 @@ import { ReactComponent as CloseSvg } from "./assets/svg/close.svg"
 import { ReactComponent as CopySvg } from "./assets/svg/copy.svg"
 import { ReactComponent as FilterSvg } from "./assets/svg/filter.svg"
 import { ReactComponent as LinkSvg } from "./assets/svg/link.svg"
+import { Flag, useFeatures } from "./feature"
 import {
   InstrumentedButton,
   InstrumentedTextField,
@@ -45,7 +46,7 @@ import {
   mixinResetButtonStyle,
   SizeUnit,
 } from "./style-helpers"
-import { TiltInfoTooltip } from "./Tooltip"
+import TiltTooltip, { TiltInfoTooltip } from "./Tooltip"
 import { ResourceName } from "./types"
 
 type UIResource = Proto.v1alpha1UIResource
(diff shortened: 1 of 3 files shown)


web/src/OverviewActionBar.tsx
```diff
@@ -659,6 +660,88 @@ export function OverviewWidgets(props: { buttons?: UIButton[] }) {
...
+  margin-left: auto;
+  padding-left: ${SizeUnit(0.25)};
+`
+const EnableDisableResourceButtonContent = styled.div`
+  .MuiIcon-root {
+    font-size: ${FontSize.small};
+  }
+  display: flex;
+  align-items: center;
+`
```
lizzthabet, line 686: is there a reason why you need a `div` with these styles inside the `button`? (afaik, having a `div` inside a `button` isn't good practice, but i forget the exact html spec wording!) 

i think you could get away with styling the `Icon` directly with `font-size` and drop the `div` + display styles.

web/src/OverviewActionBar.tsx
```diff
@@ -659,6 +660,88 @@ export function OverviewWidgets(props: { buttons?: UIButton[] }) {
...
+      color: ${Color.white};
+      background-color: ${Color.red};
+    }
+  `};
+  margin-left: auto;
+  padding-left: ${SizeUnit(0.25)};
+`
+const EnableDisableResourceButtonContent = styled.div`
+  .MuiIcon-root {
+    font-size: ${FontSize.small};
```
lizzthabet, line 682: (design nitpick) how about adding `padding-right: ${SizeUnit(1 / 8)}` to the icon, so there's a little space between the icon and button text?

web/src/OverviewActionBar.tsx
```diff
@@ -659,6 +660,88 @@ export function OverviewWidgets(props: { buttons?: UIButton[] }) {
...
+}>`
+  ${(props) =>
+    props.confirmingDisable &&
+    `
+    color: ${Color.white};
+    background-color: ${Color.red};
+
+    &:hover {
+      color: ${Color.white};
+      background-color: ${Color.red};
```
lizzthabet, line 674: (design nitpick) we may want to choose a different `border-color` for the `confirmingDisable` state in hover/active/non-hover, since the light gray (non-hover) and the blue (hover) clash with the red button. i think making the border red on regular state and darker red on hover/active states would work?

this is _very_ nitpicky, so feel free to not address it now or bring it up in a design review with @SurbhiSGupta!

web/src/OverviewActionBar.tsx
```diff
@@ -659,6 +660,88 @@ export function OverviewWidgets(props: { buttons?: UIButton[] }) {
...
+    }
+  `};
+  margin-left: auto;
```
lizzthabet, line 677: (design nitpick) i'm not a huge fan of the layout this margin creates, since it acts a little bit like a `justify-content: space-around`. from the designs, it looks like both the clear logs menu and disable button are right-aligned, so i think you could get away with `margin-left: ${SizeUnit(1 / 2)}` for now.

lizzthabet (commented):
looks good! i had one question about the state structure and a couple small design improvements that could be done now or punted till later.

web/src/OverviewActionBar.tsx
```diff
@@ -659,6 +660,88 @@ export function OverviewWidgets(props: { buttons?: UIButton[] }) {
...
+  }
+  display: flex;
+  align-items: center;
+`
+
+function EnableDisableResource(props: { resource: UIResource }) {
+  // TODO(matt) - get disabled value from resource status, set disabled value by calling api
+  const [disabled, setDisabled] = useState(false)
+
+  const [confirmingDisable, setConfirmingDisable] = useState(false)
```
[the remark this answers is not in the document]
landism: It is weird to have the two separate states in this PR! Once the backend is in place, the disabled/setDisabled state will be in the backend, while confirming disable will be only on the frontend.

web/src/OverviewActionBar.tsx
landism: It does look ugly. I'll just disable the hover behavior for now.

Logged this in shortcut [here](https://app.clubhouse.io/windmill/story/12695/resolve-design-questions?cf_workflow=500000005&ct_workflow=all&vc_group_by=day#activity-12700).

web/src/OverviewActionBar.tsx
landism: ah yeah that's simpler, thanks!

web/src/OverviewActionBar.tsx
landism: actually, border-color: white seems reasonable

nicks (commented):
high-level question: how hard would it be to make our built-in button specs support this flow?

it seems like a big red flag that we're adding a new button model that's inconsistent with our existing UIButton model

(that doesn't necessarily mean that the enable/disable button needs to be a button written to the APIServer, but it seems like it should be driven by the same UI components)

nicks:
is this PR waiting on anything? (mainly wanted to make sure it's not blocked on me)

PR #4947, opened 2021-09-09
cli: dump openapi spec from a headless server

nicks:
Hello @milas,

Please review the following commits I made in branch nicks/openapi:

873ed29eb7a54938edeaed13745e5a2030143e58 (2021-09-09 11:13:56 -0400)
cli: dump openapi spec from a headless server


nicks:
(this should make it way easier to use in scripts)

internal/cli/openapi.go
```diff
@@ -78,7 +84,48 @@ func (c *openapiCmd) run(ctx context.Context, args []string) error {
...
+	hudsc                *server.HeadsUpServerController
+	loopbackClientConfig *rest.Config
+}
+
+func newHeadlessServer(ctx context.Context) (*headlessServer, error) {
+	tmpdir, err := ioutil.TempDir("", "tilt-headless")
+	if err != nil {
+		return nil, err
+	}
+	dir := dirs.NewTiltDevDirAt(tmpdir)
```
milas, line 103: Not a huge deal as-is (and I can very much see the argument against adding complexity here), but we could change `HeadsUpServerController::addToAPIServerConfig` -> `maybeAddToAPIServerConfig` (same with remove) and only do work if `configAccess != nil` and then everything would work here w/o needing to generate temporary files.

milas (approved):
👏

internal/cli/openapi.go
nicks: ooh that's clever! done

PR #4948, opened 2021-09-09
api: add types for DisableSource/Status

landism:
1. Adds DisableSource / DisableStatus to the API, as described in [the spec](https://github.com/tilt-dev/tilt.specs/blob/master/enable_disable_resources.md#disablesource)
  * I couldn't think of good use cases for `DisableSource.Value` or `ConfigMapDisableSource.Optional`, so I figured I'd leave those out until needed.
2. Adds `DisableSource` / `DisableStatus` to `UIResource.Status`, to allow the frontend to do what it needs to do.
  * It might be a little weird to put `DisableSource` in `UIResource.Status`, since all of its other appearances will be in objects' `Spec`s. It seems like the best way to tell the frontend how to disable the resource, though.


pkg/apis/core/v1alpha1/uiresource_types.go
```diff
@@ -178,6 +178,12 @@ type UIResourceStatus struct {
 	//
 	// +optional
 	Order int32 `json:"order,omitempty" protobuf:"varint,15,opt,name=order"`
+
+	// Indicates how to disable this resource.
+	DisableSource DisableSource `json:"disableSource,omitempty" protobuf:"bytes,17,opt,name=disableSource"`
```
nicks, line 183: according to the spec, this should be in the Spec of a UIResource, not the Status of a UIResource?

https://github.com/tilt-dev/tilt.specs/blob/master/enable_disable_resources.md#disablesource

But i don't think it really makes sense in UIResource at all because it doesn't control anything....

nicks: maybe DisableStatus in the spec should also contain a reference to the ConfigMap that set the status? i think that would be closer to what you want here

landism: > according to the spec, this should be in the Spec of a UIResource, not the Status of a UIResource?
> https://github.com/tilt-dev/tilt.specs/blob/master/enable_disable_resources.md#disablesource
> But i don't think it really makes sense in UIResource at all because it doesn't control anything....

Oh, I didn't interpret the spec as having an opinion on this. Before really thinking about it, I went to add it to `UIResourceSpec` and found [this comment](https://github.com/tilt-dev/tilt/blob/faf011201df39be650876d3abe1f03de5a632e94/pkg/apis/core/v1alpha1/uiresource_types.go#L58-L62), which led me to reconsider and put it in `Status`, since, as you say, it doesn't control anything.

> But i don't think it really makes sense in UIResource at all because it doesn't control anything....
> maybe DisableStatus in the spec should also contain a reference to the ConfigMap that set the status?

Hm, why does the fact that it doesn't control anything mean it doesn't belong in UIResource? DisableSource describes how an object is disabled. As I see it, adding the ConfigMap to DisableStatus effectively means mirroring DisableSource's structs inside of DisableStatus, but DisableSource still feels like it'd be the right name for that (a type that describes how the UIResource gets disabled). We could fork DisableSource into two clones - DisableSourceSpec and DisableSourceStatus to allow for the possibility they'll diverge in the future. Or did you have something simpler in mind?

nicks: i think the closest analogue to UIResourceStatus is something like how ReplicaSetStatus or DeploymentStatus are a reflection of many aggregate pod statuses. (The analogy here isn't perfect, bc the owner relationships are different, but i think it gets at the high-level problem with this approach.)

i.e., you typically don't copy the ContainerSpec and ContainerStatus from each pod up to the ReplicaSetStatus. rather, you create some aggregate object with the bits you need

in this case, maybe we could do something like

```
DisableResourceStatus struct {
  EnableCount int
  DisableCount int
   
  []DisableSource
}
```

or even

```
DisableResourceStatus struct {
  EnableCount int
  DisableCount int
   
  // Non-nil in if all objects represented in this Resource can be controlled by a single source.
  SingleDisableSource *DisableSource
}
```

landism: ah, didn't understand from the original comment that the issue was aggregation. makes sense, thanks!

nicks (approved):

PR #4949, opened 2021-09-09
tiltfile: Pass the Tiltfile model into TiltfileLoader

nicks:
Hello @landism, @milas,

Please review the following commits I made in branch nicks/fetcher2:

5f94b0f148e6acc5a15917665b3a97363e7e5974 (2021-09-09 12:27:43 -0400)
tiltfile: Pass the Tiltfile model into TiltfileLoader

### DIFF
--- a/internal/tiltfile/tiltfile_test.go
+++ b/internal/tiltfile/tiltfile_test.go
@@ -19,9 +19,12 @@ import (
 	appsv1 "k8s.io/api/apps/v1"
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/labels"
+	ctrlclient "sigs.k8s.io/controller-runtime/pkg/client"
 
 	tiltanalytics "github.com/tilt-dev/tilt/internal/analytics"
 	"github.com/tilt-dev/tilt/internal/container"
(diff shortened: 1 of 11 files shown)


internal/tiltfile/tiltfile_test.go
```diff
@@ -5879,6 +5882,7 @@ type fixture struct {
 	k8sContext k8s.KubeContext
 	k8sEnv     k8s.Env
 	webHost    model.WebHost
+	ctrlclient ctrlclient.Client
```
milas, line 5885: Is this intentional (maybe going to be imminently used)? AFAICT the TiltfileLoader doesn't actually need the client (yet?)

milas (approved):

internal/tiltfile/tiltfile_test.go
nicks: ya, i have a branch that unifies the reconciler with the existing extension system so you can do:

extension_repo(name='default', url='file://path/to/my/repo')
extension(name='kubefwd', repo_name='default', path='kubefwd')
load('ext://my-ext', ...)

and the tiltfile loader will ask the reconciler to load the extension instead. i think that will be a lot more flexible in general.

PR #4950, opened 2021-09-09
engine: handle all link types + don't invalidate build on change

### LINKED ISSUE #4945: Tilt crashed with a not http link
### Expected Behavior
When adding a not HTTP link in a resource, Tilt should show that link

### Current Behavior
The web app becomes unresponsive and it's hard to stop the process with CTRL+c.

### Steps to Reproduce

1. I've added a MySQL link in one of my resource:
```
dc_resource('mainapp',
            links=[
              link('mysql://root:password@localhost:3306/mydbname', 'Local DB')
            ],
            labels=['MainApp']
) 
```
2. TIlt try to rebuild the Tiltfile, but the process never completes
Tilt not working

3. If I try to reload the page, it doesn't work
4. If I try to close the process on the terminal, it doesn't work immediately (It does after a couple of ctrl+c)
5. If I try to run `tilt up` again, the process start, but `(space)` doesn't open the terminal
6. If I change the URL from `mysql://` to `http://` and restart Tilt, it works, but that's not the link I need.

#### `tilt doctor` Output

```
$ tilt doctor
Tilt: v0.22.7, built 2021-09-03
System: darwin-amd64
---
Docker
- Host: [default]
- Server Version: 20.10.8
- API Version: 1.41
- Builder: 2
---
Kubernetes
- Env: docker-for-desktop
- Context: docker-desktop
- Namespace: default
- Container Runtime: docker
- Version: v1.21.4
---

---
Analytics Settings
- User Mode: opt-in
- Machine: 89d34649937db2fa2f05bc53d5b0593b
- Repo: 1Gy8Wn6qHhxrZpKcmQlMsw==
```

#### About Your Use Case
Would be helpful for my team to have the db link on our resources to access the local (and staging DB) using a client (TablePlus or SequelPRO for example)

milas:
In some cases, `url.URL` can't be compared: case reported was a
URL with a `user@` portion, which populates the `User` field of
type `url.Userinfo` that contains unexported fields. (This came
up when trying to add a link for a local DB connection string.)

While one possible fix would be to add a custom comparer for
`Link` which uses `Link::URLString()` to side-step this, since
the comparers are used for build invalidation, we can actually
just ignore the `Link` type entirely for comparison purposes,
as it does not cause a build to be invalidated.

So, as a bonus, you can now freely change `links` in `Tiltfile`
without triggering new builds unnecessarily!

Fixes #4945.

### DIFF
--- a/pkg/model/manifest.go
+++ b/pkg/model/manifest.go
@@ -5,17 +5,15 @@ import (
 	"net/url"
 	"strings"
 
+	"github.com/docker/distribution/reference"
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
 	"github.com/pkg/errors"
 	"k8s.io/apimachinery/pkg/api/validation/path"
 	"k8s.io/apimachinery/pkg/labels"
 
-	"github.com/docker/distribution/reference"
-
 	"github.com/tilt-dev/tilt/internal/container"
 	"github.com/tilt-dev/tilt/internal/sliceutils"
-
-	"github.com/google/go-cmp/cmp"
-	"github.com/google/go-cmp/cmp/cmpopts"
 )
 
 // TODO(nick): We should probably get rid of ManifestName completely and just use TargetName everywhere.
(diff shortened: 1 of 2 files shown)


landism (approved):
> It is intended to only be used in tests, as performance is not a goal and it may panic if it cannot compare the values.

:(

nicks (approved):
unit test?

tbh i think we're pretty close to having all the relevant stuff here as API models with API machinery doing invalidation

milas:
Test case added in 09be71549 which ensures it doesn't invalidate and will catch regressions on the comparison

PR #4953, opened 2021-09-10
docker-compose: upgrade compose-go

milas:
There's been a lot of fixes for different edge cases in the
`compose-go` project recently, so I pulled in the new version.

Only necessary change on Tilt side is to explicitly request that
the loader(s) resolve all paths -- v1 Compose uses relative paths,
v2 Compose early RCs (via `compose-go`) used absolute paths, but
have since switched the default _back_ to relative paths.
Thankfully, there's an option to request path resolution, so it
get set explicitly now. (Test coverage already detects if this
goes awry, which is how I discovered the flip in behavior in the
first place. Hurray!)


milas:
This was motivated by me unintentionally finding another edge case where `compose-go` is broken while working on the log streaming this week. (Spoiler: it's still broken, so I filed an upstream issue @ https://github.com/docker/compose - #8605.)

But I looked through the diff quickly and there's a bunch of fixes so still seemed worth it to pull in: https://github.com/compose-spec/compose-go/compare/efad24e892ce...813d4ccb40f8

nicks (approved):

PR #4956, opened 2021-09-10
web: fix rendering of disabled api buttons

### LINKED ISSUE #4954: disabled buttons broken at master
button.yaml

```
apiVersion: tilt.dev/v1alpha1
kind: UIButton
metadata:
  name: sample
spec:
  text: sample
  disabled: true
  location:
    componentType: resource
    componentID: "(Tiltfile)"
  inputs:
  - name: example
    label: example
    bool: {}
 ```

```
tilt apply -f button.yaml
```

Output:
Screenshot from 2021-09-10 12-30-49

### PR DESCRIPTION
**Pull request by landism**
Fixes #4954 

### Problems
1. In the overview action bar, when a uibutton is disabled, its submit button is effectively invisible. Also, its options button gets no disabled styling.
2. In the nav bar, there's no visual indication that a button is disabled.
3. The UI was not showing the appropriate cursor on disabled buttons.

### Solutions
1. Fix the disabled styling for overview buttons and apply the same to the options button.
  1. I accomplished the latter by just disabling the options button, which is hacky, but was almost free and seems unlikely to have much user impact. If a user notices this we can spend the time to apply the styling without disabling the button, or if anyone thinks that's worth the time now, I'll do it.
2. Add some opacity to the disabled styling for nav bar buttons.
3. Apply the [mui workaround](https://material-ui.com/components/buttons/#cursor-not-allowed) to fix the cursor on disabled buttons.

Here's a screenshot with some disabled buttons:
image

### DIFF
--- a/web/src/ApiButton.tsx
+++ b/web/src/ApiButton.tsx
@@ -38,7 +38,14 @@ const ApiButtonFormFooter = styled.div`
 `
 const ApiIconRoot = styled.div``
 export const ApiButtonLabel = styled.div``
-export const ApiButtonRoot = styled(ButtonGroup)`
+// MUI makes it tricky to get cursor: not-allowed on disabled buttons
+// https://material-ui.com/components/buttons/#cursor-not-allowed
+export const ApiButtonRoot = styled(ButtonGroup)<{ disabled?: boolean }>`
+  ${(props) =>
+    props.disabled &&
+    `
+    cursor: not-allowed;
+  `}
   ${ApiIconRoot} + ${ApiButtonLabel} {
     margin-left: ${SizeUnit(0.25)};
   }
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: web/src/ApiButton.tsx [hunk 1]
```diff
@@ -38,7 +38,14 @@ const ApiButtonFormFooter = styled.div`
 `
 const ApiIconRoot = styled.div``
 export const ApiButtonLabel = styled.div``
-export const ApiButtonRoot = styled(ButtonGroup)`
+// MUI makes it tricky to get cursor: not-allowed on disabled buttons
+// https://material-ui.com/components/buttons/#cursor-not-allowed
+export const ApiButtonRoot = styled(ButtonGroup)<{ disabled?: boolean }>`
+  ${(props) =>
```
nicks, line 44: we usually do this with CSS cascading?

nicks (approved):

web/src/ApiButton.tsx
landism: Yeah, that's how we've been doing it. I've been wanting to try out Styled Components props for a while, but hadn't figured out how to get them to play nicely with TypeScript.

I think the extra bit of typing might make them a bit nicer than class names - the compiler can catch issues and IDEs can jump between use to definition. If you've got a strong preference for sticking with class names, we can do that. Merging now - it's easy to change later.

PR #4960, opened 2021-09-14
tiltfile: clarify unexpected positional args error

milas:
This sounds like a Starlark syntax error, but it's actually about
the CLI args passed to `tilt up` if there's no Tiltfile config
positional option defined.

The text has been adjusted slightly to explicilty refer to "CLI
args" and links to a (new) section in the docs (see
tilt-dev/tilt.build#941).

No behavior has changed.

### DIFF
--- a/internal/tiltfile/config/config_def.go
+++ b/internal/tiltfile/config/config_def.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"fmt"
 	"os"
+	"strings"
 
 	jsoniter "github.com/json-iterator/go"
 	"github.com/pkg/errors"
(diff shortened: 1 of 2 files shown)


internal/tiltfile/config/config_def.go
```diff
@@ -119,7 +120,9 @@ func (cd ConfigDef) parseArgs(args []string) (ret configMap, output string, err
 
 	if len(fs.Args()) > 0 {
 		if cd.positionalSettingName == "" {
-			return nil, w.String(), errors.New("positional args were specified, but none were expected (no setting defined with args=True)")
+			return nil, w.String(), fmt.Errorf(
+				"invalid tiltfile config args: positional CLI args (%q) were specified, but none were expected.\n"+
+					"See https://docs.tilt.dev/tiltfile_config.html#positional-arguments for examples.", strings.Join(fs.Args(), " "))
```
landism, line 125: wdyt of taking the prefixes out of `parseArgs` and adding an `error.Wrap(err, "invalid tilitfile config args:")` at the call site in `incorporateArgs`?

(parseArgs` has three places that can return errors, and now two of them have (differing) prefixes)

milas: Good call - see latest (b2fb211fe) for changes

### parse error
formatting has been improved here - before you got the error + usage as INFO (not ERROR) and then the error again (w/o usage) as ERROR; now everything is 1x in ERROR
arg parse error

### unexpected positional args
unexpected positional args error

### invalid positional args
(bonus: added a missing `)` on this error message)
invalid positional args error

landism (approved):

PR #4964, opened 2021-09-16
storybook: add prettier and lint buttons to storybook Tiltfile as shortcuts

lizzthabet:
Screen Shot 2021-09-16 at 3 48 35 PM

i've been thinking about making the lint-related local resources into buttons for a while. i'm not sure it's a huge improvement (especially since when you run commands from the button, you don't get the exit status of the command), but i thought i might try it out for a bit. open to suggestions or even closing this if there are any strong opinions against!


landism:
Yeah, definitely worth trying!

> especially since when you run commands from the button, you don't get the exit status of the command

This seems like a good feature request!

web/storybook.tiltfile
```diff
@@ -14,6 +14,33 @@ local_resource(
...
+  argv=['sh', '-c', 'node_modules/.bin/tsc -p .'],
+  text='1. Verify compile',
+  location=location.RESOURCE,
+  icon_name='handyman',
+)
+
+cmd_button(
+  name='3check-js',
+  resource='storybook',
+  argv=['sh', '-c', 'cd .. && make check-js'],
```
landism, line 38: The underlying `cmd` already supports a dir, but the uibutton extension just hardcodes it to `config.main_dir`. We should probably allow the user to set that so they don't have to do weird stuff like that - any interest in taking a stab at that? shall I?

lizzthabet: oh yeah! i can take a stab at that.

lizzthabet:
> > especially since when you run commands from the button, you don't get the exit status of the command
> 
> This seems like a good feature request!

💥🧠 i can create an issue for that! i guess in this case, what i could do is just `tilt trigger` the corresponding `local_resource` and then i'd see the status of those commands on those resources... which maybe begs the question about what utility exactly the buttons serve, if they're just providing another button to trigger a resource. eh. i think it's still worth playing around with buttons in our own codebase and workflows.

PR #4966, opened 2021-09-17
api: configure DisableSource on eligible API objects

landism:
When a Tiltfile creates API objects:
1. Create ConfigMaps to be used for DisableSources for each resource.
2. Configure `DisableSource` in the specs of all objects that need it.

DisableSource is not yet read by anything.

It's a little weird that with this change, the TiltfileLoadResult has full specs for these objects, and we're adding in the DisableSource at the end.
Alternatives:
1. We could construct the spec with the appropriate DisableSource in the first place, but we don't know the manifest name until assembly time.
2. We could have the tiltfile loader itself iterate over all these objects and set the DisableSources as appropriate. The main reason I chose a different approach in this PR is that this code already knows about the list of API objects, while doing it in the Tiltfile loader would involve adding a second place that we're iterating over types / objects.

I'm also a bit on the fence about how this is getting ConfigMaps into the API:
1. returning an empty GetSpec is weird! (though teeechnically accurate)
2. the special case for updating ConfigMaps is also not amazing

The alternative is basically duplicating all the create/update/delete logic for the ConfigMaps. I figure we can worry about that if/when the Tiltfile generates other ConfigMaps.


internal/controllers/core/tiltfile/api.go
```diff
@@ -348,6 +392,15 @@ func updateNewObjects(ctx context.Context, client ctrlclient.Client, newObjects,
 			labelsChanged := !apicmp.DeepEqual(old.GetLabels(), obj.GetLabels())
 			if specChanged || labelsChanged {
 				obj.SetResourceVersion(old.GetResourceVersion())
+				if cm, ok := obj.(*v1alpha1.ConfigMap); ok {
+					// Tiltfiles can create ConfigMaps with default values, but
+					// they shouldn't blow away values that were modified elsewhere.
+					// At some point, we might have a use case that contradicts this and
+					// need to make this more complicated.
```
milas, line 399: Ah, nice - I feel like this will be important for things like trigger mode as well (since currently after toggling via UI it'll get reset next Tiltfile execution). Perhaps `Patch()` / SMD would be the more general solution here in the future

milas (approved):

PR #4972, opened 2021-09-20
tiltfile: add built-in for filewatch

nicks:
Hello @milas,

Please review the following commits I made in branch nicks/filewatchcodegen:

22cc7165582aa74fa81b01cdf32a745632594941 (2021-09-20 16:33:30 -0400)
tiltfile: add built-in for filewatch

### DIFF
--- a/internal/tiltfile/v1alpha1/types.go
+++ b/internal/tiltfile/v1alpha1/types.go
@@ -1,6 +1,8 @@
 package v1alpha1
 
 import (
+	"fmt"
+
 	"go.starlark.net/starlark"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
@@ -20,6 +22,10 @@ func (p Plugin) registerSymbols(env *starkit.Environment) error {
 		return err
 	}
 	err = env.AddBuiltin("v1alpha1.extension_repo", p.extensionRepo)
+	if err != nil {
+		return err
+	}
+	err = env.AddBuiltin("v1alpha1.file_watch", p.fileWatch)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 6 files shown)


internal/tiltfile/v1alpha1/types.go
```diff
@@ -73,3 +79,111 @@ func (p Plugin) extensionRepo(t *starlark.Thread, fn *starlark.Builtin, args sta
...
+type IgnoreDef struct {
+	*starlark.Dict
+	Value v1alpha1.IgnoreDef
+	t     *starlark.Thread // instantiation thread for computing abspath
+}
+
+func (o *IgnoreDef) Unpack(v starlark.Value) error {
+	obj := v1alpha1.IgnoreDef{}
+
+	mapObj, ok := v.(*starlark.Dict)
```
milas, line 119: Did you consider having `tilt-starlark-codegen` make struct types for nested complex types vs `dict`s?

nicks: i'm also going to add a built-in v1alpha1.ignore_def() that returns this object

re: structs vs dicts - i think it would be totally reasonable to accept both structs and dicts here. I started with `dict` because i expect python programmers to be more familar with it (as opposed to struct, which is starlark-only). i don't think it's too hard to convert between them.

https://docs.bazel.build/versions/main/skylark/lib/struct.html

pkg/apis/core/v1alpha1/filewatch_types.go
```diff
@@ -55,7 +56,10 @@ type FileWatchList struct {
 // FileWatchSpec defines the desired state of FileWatch
 type FileWatchSpec struct {
 	// WatchedPaths are paths of directories or files to watch for changes to. It cannot be empty.
+	//
+	// +tilt:local-path=true
```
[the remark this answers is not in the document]
nicks: :)

PR #4973, opened 2021-09-20
tiltfile: kubecontext protection shouldn't apply when loading extensions

nicks:
Hello @landism,

Please review the following commits I made in branch nicks/allowed:

af810d6f5d1412cc361a7f5731662e08f573ca10 (2021-09-20 19:54:41 -0400)
tiltfile: kubecontext protection shouldn't apply when loading extensions
Fixes https://github.com/tilt-dev/tilt-extensions/issues/253

### DIFF
--- a/internal/controllers/apis/tiltfile/main.go
+++ b/internal/controllers/apis/tiltfile/main.go
@@ -2,13 +2,31 @@ package tiltfile
 
 import (
 	"fmt"
+	"path/filepath"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
+	"github.com/tilt-dev/tilt/internal/ospath"
 	"github.com/tilt-dev/tilt/pkg/apis"
 	"github.com/tilt-dev/tilt/pkg/apis/core/v1alpha1"
 	"github.com/tilt-dev/tilt/pkg/model"
 )
+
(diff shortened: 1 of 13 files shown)


landism:
The code looks good.

Sorry for a bunch of words incoming about a relatively minor thing - just a bit of handwringing about the PR title/description - I feel like they set me off on the wrong foot when trying to understand the change. To make this more accessible to users reading the changelog, it might be worth 1) making clear this is a workaround rather than our desired long-term policy and 2) clarifying what this actually affects.
Maybe something like:
1. change the title to "disable kubecontext protection to extensions loaded outside of the main tiltfile context"
2. add a bit to the description describing that it's a workaround and that what "outside of the main tiltfile context" means. We have a lot of existing extensions for which I'd expect users want this protection and it'd be nice to communicate we're not breaking that.

(nb: I have been struggling to come up with better names to clarify our dual meanings of "api" (i.e., tilt api and tiltfile api) and "extensions" (i.e., via load and via api objects))

(also maybe this protection on `local` doesn't matter much anyway - it seems notable that we don't have this protection on `local_resource` and no one has asked for it! We added it relatively early in Tilt's life and maybe it's become less important as the Tiltfile API has become more flexible?)

landism (approved):

PR #4975, opened 2021-09-21
cmd: reconciler uses DisableSource/Status

landism:
When a `Cmd`'s disable `ConfigMap` is set, the `Cmd` will disable/enable and set its `Status.DisableStatus` as appropriate.

### DIFF
--- a/internal/controllers/apis/configmap/disable.go
+++ b/internal/controllers/apis/configmap/disable.go
@@ -1,0 +1,65 @@
+package configmap
+
+import (
+	"context"
+	"fmt"
+
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/tilt-dev/tilt/pkg/apis/core/v1alpha1"
+)
+
+func disableStatus(ctx context.Context, client client.Client, disableSource *v1alpha1.DisableSource) (isDisabled bool, reason string, err error) {
+	if disableSource == nil {
+		return false, "object does not specify a DisableSource", nil
+	}
+
+	if disableSource.ConfigMap == nil {
+		return false, "DisableSource specifies no ConfigMap", nil
+	}
+
+	cm := &v1alpha1.ConfigMap{}
+	err = client.Get(ctx, types.NamespacedName{Name: disableSource.ConfigMap.Name}, cm)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			return false, fmt.Sprintf("ConfigMap %q does not exist", disableSource.ConfigMap.Name), nil
+		}
+		return false, fmt.Sprintf("error reading ConfigMap %q", disableSource.ConfigMap.Name), err
+	}
+
+	cmVal, ok := cm.Data[disableSource.ConfigMap.Key]
+	if !ok {
+		return false, fmt.Sprintf("ConfigMap %q has no key %q", disableSource.ConfigMap.Name, disableSource.ConfigMap.Key), nil
+	}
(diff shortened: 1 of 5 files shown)


internal/controllers/core/cmd/controller_test.go
```diff
@@ -941,14 +1057,14 @@ func (f *fixture) assertCmdMatches(name string, matcher func(cmd *Cmd) bool) *Cm
 		return matcher(cmd)
 	}, timeout, interval)
 
-	return f.assertCmdMatchesInAPI(name, matcher)
+	return f.requireCmdMatchesInAPI(name, matcher)
 }
 
-func (f *fixture) assertCmdMatchesInAPI(name string, matcher func(cmd *Cmd) bool) *Cmd {
+func (f *fixture) requireCmdMatchesInAPI(name string, matcher func(cmd *Cmd) bool) *Cmd {
```
landism, line 1063: I made this change because I was getting spurious / misleading error messages from subsequent code when the assertion failed, and terminating seems better. If someone has a preference for `assert` mechanics in this method, 
I can do it another way.

internal/controllers/apis/configmap/disable.go
```diff
@@ -0,0 +1,65 @@
...
+		}
+		return false, fmt.Sprintf("error reading ConfigMap %q", disableSource.ConfigMap.Name), err
+	}
+
+	cmVal, ok := cm.Data[disableSource.ConfigMap.Key]
+	if !ok {
+		return false, fmt.Sprintf("ConfigMap %q has no key %q", disableSource.ConfigMap.Name, disableSource.ConfigMap.Key), nil
+	}
+	// TODO(matt)
+	// checking `== "true"` rather than strconv.ParseBool because we lack a good way to surface errors
```
nicks, line 38: Hmmm...I'm having trouble reconciling the statement "we lack a good way to surface errors" and then....most of this function is surfacing errors in the reason field? what am i missing?

seems like we could use strconv and then surface any errors in 'reason'

internal/controllers/core/cmd/controller.go
```diff
@@ -224,6 +243,15 @@ func (c *Controller) reconcile(ctx context.Context, name types.NamespacedName) e
 //
 // Blocks until the command is finished, then returns its status.
 func (c *Controller) ForceRun(ctx context.Context, cmd *v1alpha1.Cmd) (*v1alpha1.CmdStatus, error) {
+	disableStatus, err := configmap.MaybeNewDisableStatus(ctx, c.client, cmd.Spec.DisableSource, cmd.Status.DisableStatus)
+	if err != nil {
+		return nil, err
+	}
+	if disableStatus.Disabled {
+		// TODO(matt) have the build controller ensure we don't reach this point on disabled commands (ch12734)
```
nicks, line 251: hmmmm...this comment makes it sound like we shouldn't check this in ForceRun at all? The name "ForceRun()" implies to me that we've already done the necessary checks. this seems premature and likely to bite us with race conditions later.

internal/controllers/core/cmd/controller.go
```diff
@@ -311,7 +339,10 @@ func (c *Controller) runInternal(ctx context.Context,
 
 	stillHasSameProcNum := proc.stillHasSameProcNum()
 	c.updateStatus(name, func(status *CmdStatus) {
-		*status = CmdStatus{Waiting: &CmdStateWaiting{}}
+		*status = CmdStatus{
```
nicks, line 342: seems like we should change this to use

status.Waiting = &CmdStateWaiting{}
status.Running = nil
status.Terminated = nil

like we do elsewhere in this file. i.e., only modify the fields we want to modify.

nicks (approved):

internal/controllers/apis/configmap/disable.go
landism: Ah, yeah, this code is confusing and silly.

Thanks!

PR #4978, opened 2021-09-21
telemetry: upgrade to otel 1.0

milas:
There's a substantial number of changes here due to moving from
v0.2.0 to v1.0.0. Notably, the way that we were marshaling spans
to JSON doesn't work anymore, so there's a decent amount of bridge
code, which is a fork of the structs and marshaling code as of
v0.2.0 with helper functions to convert from the newer types into
them.

Eventually, we should probably provide an option to the
`experimental_telemetry_cmd` (or replacement) to control the
serialization format and use the canonical OTLP JSON format (but
this is not yet finalized).

Beyond this, the span collector is now a `SpanExporter` rather
than a `SpanProcessor` and is invoked by the built-in batch
processor. This actually impacts tests more than production code!

**serialization comparison**
otel span serialization diff


milas:
I was trying to pull in a more recent version of Docker Go API client today and it relies on a more recent `opentelemetry-go` (who happened to tag v1.0.0 yesterday 🎉)...there was some fun glue needed to not break `experimental_metrics_cmd`.

(As it turns out, I did not need to upgrade Docker Go lib for what I needed :face_palm::skin-tone-2: but anyway this was already done, so yay?)

milas:
For reviewing: the `exptel` package is 99% code forked from `opentelemetry-go` v0.2.0. The `NewXxxFromOtel` methods are the Tilt-specific bridge code to do conversions from v1.0.0 -> v0.2.0 types.

nicks (approved):
i hope you didn't spend too much time on this. I'm not sure the telemetry serialization was worth preserving. Very, very few teams use it (according to analytics). And it's based on a bunch of assumptions about opentelemetry that have changed (in particular, as you noticed, the json wire format).

nicks:
(tho this unblocks the upgrade to k8s v1.21 support libraries, so it certainly isn't wasted!)

milas:
The v1.0.0 `Span` isn't actually super different (it's just an interface and doesn't JSON marshal the same in Go), so converting between formats was actually very straightforward + quick - figuring out the new package paths was what was annoying 😛

milas:
(When I first did the upgrade, the tests were failing due to the serialization differences, so I actually almost halted and proposed deprecating/removing that, but then saw it was easy enough to fork off the old versions to give us a transition period)

PR #4979, opened 2021-09-22
apis: add LiveUpdateSpec fields

nicks:
Hello @milas, @landism,

Please review the following commits I made in branch nicks/lu3:

84349a3dc9a40a297824ee2803154672b7e06d8d (2021-09-21 20:38:27 -0400)
apis: add LiveUpdateSpec fields
This is a fork of the LiveUpdate model under pkg/model, but with some light
naming and structure changes based on how users currently use it. We expect this
to replace pkg/model/LiveUpdate in-place.


pkg/apis/core/v1alpha1/liveupdate_types.go
```diff
@@ -53,6 +53,67 @@ type LiveUpdateList struct {
...
+	// of container discovery objects (like Docker Compose container discovery).
+	//
+	// +optional
+	KubernetesDiscoveryName string `json:"kubernetesDiscoveryName,omitempty" protobuf:"bytes,3,opt,name=kubernetesDiscoveryName"`
+
+	// A list of relative paths that will immediately stop the live-update for the
+	// current container.
+	//
+	// Used to detect file changes that invalidate the entire container image,
+	// forcing a complete rebuild.
```
landism, line 86: I don't have a good sense how this will fit into the flow or what "stop the live-update" will mean in practice, would something like this be accurate?

```
A list of relative paths that will cause this live-update to pause until the container has been redeployed.
```

(also the existing wording makes it easy to think the liveupdate does the complete rebuild)

If that's accurate, maybe "container has been redeployed" is a weird condition and we want the PodTemplateSpecHash somewhere?

landism (approved):
Some comments on docs that are probably not that important for the code and mostly me trying to figure out how this fits in.

FWIW, I'm struggling to guess at what a LiveUpdate's lifecycle might look like once we have other pieces in place. As you're aware, this has a bunch of fields that are not in the existing model. Most of the ways in which this could work I've been able to think of have seemed racy. I've found this has made it tricky to feel like a useful reviewer on this and similar PRs.

Is the goal here similar to what we've done with Cmd - to define an api object to allow us to have a reconciler and do api tricks, and not having the api actually own existing live updates? (e.g., in the medium term, we write a LiveUpdate reconciler that is functional and can be used by extensions, but don't actually run it on live_updates specified in Tiltfiles?)

It's reasonable if you want to merge / don't think it's worth spelling out here / think it'll be cheaper to make clear via future PRs - I just wanted to put that qualifier on the value of my review here.

pkg/apis/core/v1alpha1/liveupdate_types.go
nicks: this exists in the model today: https://docs.tilt.dev/api.html#api.fall_back_on

nicks:
Hmmm...did you read the PR description? it says:

> This is a fork of the LiveUpdate model under pkg/model, but with some light
> naming and structure changes based on how users currently use it. We expect this
> to replace pkg/model/LiveUpdate in-place.

I'm trying to reconcile that with your statement:

> As you're aware, this has a bunch of fields that are not in the existing model.

Which fields are you referring to? I'm not aware of any that don't exist in the existing model.

nicks:
(i  constructed this PR by going through the existing model, and copying fields over in a way that made sense with the k8s api conventions, and copying/rewriting the documentation for them)

landism:
> Which fields are you referring to? I'm not aware of any that don't exist in the existing model.

I was looking at this existing model:
```
type LiveUpdate struct {
	Steps   []LiveUpdateStep
	BaseDir string // directory where the LiveUpdate was initialized (we'll use this to eval. any relative paths)
}
```

Sorry, "a bunch" was a bad impression from just looking at the number of fields without considering this PR explodes Steps.
To get more concrete, AFAICT, "FileWatchName" and "KubernetesDiscoveryName" aren't at this level in the current model. Inside a BaD, we already have this information from elsewhere, and populating them on this model seems like it'd just open us to race conditions. Are those fields intended for use inside a BaD? Am I missing an easy way to use them without race conditions?

milas (approved):

pkg/apis/core/v1alpha1/liveupdate_types.go
milas: (More generally, I'm curious how the reconciler for this will work, both short-term and long-term)

PR #4988, opened 2021-09-24
cli: add demo subcommand

milas:
`tilt demo` is meant for first-time Tilt user who want to explore
without needing to get a local K8s cluster up and running, intended
for use as part of the revamped tutorial.

It will create an ephemeral cluster using k3d+k3s against the local
Docker daemon, which is torn down on exit. (In the event that
something goes awry, `tilt demo --teardown` will clean up any k3d
clusters whose names start with `tilt-demo-`.)

Then, it will clone the `tilt-avatars` project (using `go-get`
that we use for cloning extensions) and simulate a `tilt up` against
it, using the generated kubeconfig.

invoking `tilt demo` via CLI


nicks:
I'm finding it hard to find a continuous block of time in my schedule long enough to review this PR, since there are a lot of moving independent pieces. Would it be possible to break it up a bit? (e.g., the Docker client new functionality from the other stuff)

internal/cli/demo/k3d.go
```diff
@@ -0,0 +1,165 @@
...
+	err := k.command(ctx, []string{"kubeconfig", "get", clusterName}, &kubeconfigBuf, stderr, true)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get kubeconfig: %v", err)
+	}
+	return kubeconfigBuf.Bytes(), nil
+}
+
+func (k *K3dClient) command(ctx context.Context, cmd []string, stdout io.Writer, stderr io.Writer, wait bool) error {
+	k.ensurePulled.Do(func() {
+		ref, err := k.cli.ImagePull(ctx, k.k3dImage)
```
nicks, line 125: is there a reason why you chose to pull the image up-front rather than Pull as part of the Run?

internal/cli/demo/k3d.go
```diff
@@ -0,0 +1,165 @@
...
+		if err != nil {
+			logger.Get(ctx).Errorf("failed to pull %q image: %v", k.k3dImage, err)
+		} else {
+			k.k3dImage = ref
+		}
+	})
+
+	runOpts := docker.RunOptions{
```
nicks, line 133: is the idea of this to avoid tooling to install k3d?

I'm a little a bit about having k3d state living inside of a container and being hard to cleanup

milas:
Yes, I'll chunk this up! Sorry, it was much longer than expected once done

internal/cli/demo/k3d.go
```diff
@@ -0,0 +1,165 @@
...
+	cmd := []string{
+		"cluster",
+		"create", clusterName,
+		"--registry-create",
+		"--kubeconfig-switch-context",
+		"--kubeconfig-update-default",
+		"--no-hostip",
+		"--no-image-volume",
+		"--no-lb",
+		"--label", fmt.Sprintf("%s@%s", docker.BuiltByTiltLabelStr, "server[0]"),
```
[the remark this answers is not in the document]
milas: Will add a comment - it's a "node filter" for K3d, which is meant to let you label agents/server/LB/etc differently. There's only a single node (`server[0]`) in this configuration, which ironically makes K3d emit a cranky warning that it doesn't have a node filter, so this is explicit to suppress that

internal/cli/demo/k3d.go
milas: Just to speed things up a bit because we run multiple k3d commands, each one in an ephemeral container, so it's redundant after the first time. This isn't _hugely_ critical at startup, but on shutdown, we have < 2 sec to fire off the Docker API request before the signal handler forcibly terminates the process 😬

(For reference, it takes >500ms to `ImagePull()` on something I've already got pulled, which I'm guessing is largely dependent on local I/O to find the image digest and then the network request to registry API to compare.)

internal/cli/demo.go
```diff
@@ -0,0 +1,193 @@
...
+
+	//
+	// 0. Prepare environment
+	//
+	err = client.CheckConnected()
+	if err != nil {
+		return fmt.Errorf("tilt demo requires Docker to be installed and running: %v", err)
+	}
+	if client.Env().Host != "" && !strings.HasPrefix(client.Env().Host, "unix://") {
+		// to properly support remote Docker hosts, we'd need to:
```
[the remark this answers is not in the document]
milas: Ahhh that's interesting! I definitely didn't dive into this deeply; comment was intended more as a "thar be dragons" than anything. I adjusted/clarified that and mentioned ctlptl here

internal/cli/demo/k3d.go
milas: > is the idea of this to avoid tooling to install k3d?

Precisely - the value prop of "download tilt and run `tilt demo` to try it out" vanishes if you need to download k3d/minikube/KIND - at that point, you might as well actually create the cluster yourself 🙃 

(TBH I think that an approach either like this in Docker OR that actually downloads k3d/kind/minikube before execing if needed would probably be a good addition to `ctlptl`, and ideally we could use that here.)

> I'm a little a bit about having k3d state living inside of a container and being hard to cleanup

There's not any state other than the Docker containers it creates - each invocation of `k3d` actually runs in an ephemeral container and gets thrown away. Clean-up wise, I delegated to `k3d cluster delete` for ease, but it actually might be more sane to just use some labels and delete the containers directly based off that, which is presumably what it's doing anyway. (The generated `kubeconfig` is at a tmp path to not interfere with any the user might already have, so there's no config cleanup needed.)

LMK if that makes sense or if I'm overlooking something here!

milas:
(This is rebased and JUST contains the k3d/demo CLI stuff now so is shorter - I'm still making some tweaks based on initial feedback, but LMK if it'd be useful to further split out the k3d client portion into its own PR!)

internal/cli/demo.go
```diff
@@ -0,0 +1,193 @@
...
+	k3dCli := demo.NewK3dClient(client)
+
+	//
+	// 0. Prepare environment
+	//
+	err = client.CheckConnected()
+	if err != nil {
+		return fmt.Errorf("tilt demo requires Docker to be installed and running: %v", err)
+	}
+	if client.Env().Host != "" && !strings.HasPrefix(client.Env().Host, "unix://") {
```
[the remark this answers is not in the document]
milas: Switched in f94ba5b with a TODO to switch to use shared lib (`github.com/tilt-dev/cluster`) once ready

internal/cli/demo/k3d.go
milas: I went ahead and opened #4995 with the k3d-via-Docker client portion to (hopefully?) make it easier to discuss

nicks (approved):

PR #4989, opened 2021-09-27
docker: add image pull method

milas:
This is in preparation for directly creating+running containers
with Docker, where it's necessary to do a pull first to ensure
the image exists before creating the container (this is automatic
behavior from the CLI but not when using the API).

(This was split out of #4988.)

### DIFF
--- a/internal/docker/client.go
+++ b/internal/docker/client.go
@@ -10,14 +10,14 @@ import (
 	"os"
 	"path/filepath"
 	"strconv"
+	"strings"
 	"sync"
 	"time"
-
-	"github.com/docker/cli/cli/connhelper"
 
 	"github.com/blang/semver"
 	"github.com/docker/cli/cli/command"
 	"github.com/docker/cli/cli/config"
+	"github.com/docker/cli/cli/connhelper"
 	cliflags "github.com/docker/cli/cli/flags"
 	"github.com/docker/distribution/reference"
 	"github.com/docker/docker/api/types"
@@ -28,6 +28,7 @@ import (
 	"github.com/moby/buildkit/identity"
 	"github.com/moby/buildkit/session"
 	"github.com/moby/buildkit/session/auth/authprovider"
+	"github.com/opencontainers/go-digest"
 	"github.com/pkg/errors"
 
 	"github.com/tilt-dev/tilt/internal/container"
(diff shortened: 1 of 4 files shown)


internal/docker/client.go
```diff
@@ -384,41 +386,110 @@ func (c *Cli) ServerVersion() types.Version {
...
+	if err != nil {
+		return nil, fmt.Errorf("connection error while pulling image %q: %v", image, err)
+	}
+
+	imgInspect, _, err := c.ImageInspectWithRaw(ctx, image)
+	if err != nil {
+		return nil, fmt.Errorf("failed to inspect after pull for image %q: %v", image, err)
+	}
+
+	digestInfo := strings.SplitN(imgInspect.RepoDigests[0], "@", 2)
```
nicks, line 456: can you add some comments on why you need to hand-write a digest parser here? I thought the docker/distribution parser takes care of this.

nicks (approved):

internal/docker/client.go
milas: Sortaaa, `reference.Parse(imgInspect.RepoDigests[0])` + cast to `reference.Canonical` _will_ work, but you'll lose repo (if it's `library.docker.io`) and tag (which is arguably redundant, but useful in this context if you plan to immediately use the value to create+run containers so that it shows up in Docker engine).

(e.g. with the impl here you get `docker.io/nginx:1.21.3@sha256:<hash>` vs just `nginx@sha256:<hash>`)

If you think that's preferable, I can swap over. Unfortunately, `docker/distribution` doesn't have better helpers here (unlike containerd: https://github.com/containerd/containerd/blob/6416cc35c1f0a92a3561411de4ccfb71df2c8308/reference/reference.go#L155-L166)

nicks: does ParseAnyReference work?

The docker.io/nginx vs nginx is called "Familiarization" in the API, and you can usually handle it with https://pkg.go.dev/github.com/docker/distribution@v2.7.1+incompatible/reference#FamiliarString)

and ya, the image reference spec is a mess, i've been meaning to find one of the authors at kubecon sometime and pry the story out of them of why it's such a mess.

milas: Okay, tweaked this - it uses `ParseNormalizedNamed` (`ParseAnyReference` would delegate to this in this case so saves extra cast/indirection), casts to `reference.Canonical` and then uses `reference.WithDigest(original, inspectResult.Digest()`.

Roundabout but best that can be done afaict with their APIs that will ensure:
 * Qualified name
 * Tag preserved if provided to pull
 * Digest attached

nicks (approved):

PR #4990, opened 2021-09-28
vendor: upgrade tilt-apiserver

nicks:
Hello @milas,

Please review the following commits I made in branch nicks/vendor:

80912f5e042547f05235faf00d9260ebaf24a636 (2021-09-28 11:14:18 -0400)
vendor: upgrade tilt-apiserver
Unfortunately, I needed to rollback opentelemetry to v0.20.0, because
that's what k8s uses. The difference between v1.0.0 and v0.20.0 isn't THAT big
but it did requires some bridge code.


milas:
> Unfortunately, I needed to rollback opentelemetry to v0.20.0

Ooof - hopefully the ecosystem upgrades to v1.0.0 so we don't get stuck between incompatible versions between deps at some point 😓

milas (approved):

PR #4991, opened 2021-09-28
docker: add run (container create+start) method

milas:
Single-shot method to create and run a container, similar to
`docker run`. The container create + container run semantics of
the Docker API are more low-level than we care about right now,
so it's easier to encapsulate these into a single method. It
also has some convenience bits around getting container logs and
the execution result (which has quirky semantics).

This will be used by #4988.


internal/docker/run.go
```diff
@@ -0,0 +1,66 @@
...
+	// If an image will only be used once, this is a convenience to avoid calling ImagePull first.
+	// If an image will be used multiple times (across containers), prefer explicitly calling ImagePull
+	// to avoid the overhead of calling the registry API to check if the image is up-to-date every time.
+	Pull bool
+	// ContainerName is a unique name for the container. If not specified, Docker will generate a random name.
+	ContainerName string
+	// ContainerConfig is the main configuration for the container.
+	//
+	// The Image field MUST be populated at a minimum.
+	ContainerConfig mobycontainer.Config
```
nicks, line 24: it might make sense to have narrower settings than the whole Config?

i feel like there are settings here that are incompatible with how you're running the container (e.g., AutoCleanup and RestartPolicy)

internal/docker/client.go
```diff
@@ -645,3 +649,99 @@ func (c *Cli) ExecInContainer(ctx context.Context, cID container.ID, cmd model.C
...
+	if err != nil {
+		return RunResult{}, fmt.Errorf("could not create container: %v", err)
+	}
+	var containerStarted bool
+	defer func() {
+		// make an effort to clean up any container we create but don't successfully start
+		if containerStarted {
+			return
+		}
+		if err := c.Client.ContainerRemove(ctx, createResp.ID, types.ContainerRemoveOptions{}); err != nil {
```
nicks, line 686: do we also need to remove the container when it exits? maybe this should be a Close() method of the RunResult?

nicks (approved):

internal/docker/client.go
milas: I like that - combined with reducing surface area of options it makes it harder to misuse. Added in f758868

internal/docker/run.go
milas: Fair - the Docker types are giant, so I was a little afraid of having an obnoxious amount of mapping code, but I think realistically Tilt will only ever want/need to use a handful of options - refactored in f758868

nicks (approved):

PR #4992, opened 2021-09-28
api: uibutton: add "hidden" input type

landism:
This is basically HTML's `<input type="hidden>`, but for UIButton.

The motivation for this is so that when we update UIButtonStatus for a toggle button, that button status will also include whether the user was clicking "on" or "off".


nicks (approved):
i like this idea!

PR #4995, opened 2021-09-29
demo: add k3d-via-Docker client to provision clusters

milas:
To simplify trying out Tilt when following the new tutorial, we
want an "all-in-one" `tilt demo` command that will launch a
temporary K8s cluster, clone the sample project, and `tilt up`
against it.

To avoid extra dependencies within Tilt, this is an experimental
k3d client that executes k3d commands by running k3d itself via
Docker with the Docker socket mounted. This eliminates the need
for the user to have k3d downloaded (or Tilt to play "package
manager" and grab the appropriate binary from GitHub releases.)

There's a few quirks/limitations with this approach, namely that
k3d can't interact with the host system to e.g. update kubeconfig.
However, that's actually desirable in this instance as we want to
keep things isolated.

Longer-term, particularly if we want Tilt to truly manage clusters
as part of the dev environment (that are meant to be more robust
and "permanent" than for a tutorial), it'd probably make sense to
use `ctlptl` as a library and explore options within `ctlptl` to
avoid requiring `k3d` itself as a dependency (note: `ctlptl` does
not actually support `k3d` today).

### DIFF
--- a/internal/cli/demo/k3d.go
+++ b/internal/cli/demo/k3d.go
@@ -1,0 +1,172 @@
+package demo
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"strings"
+	"sync"
+
+	"github.com/docker/distribution/reference"
+	"github.com/docker/docker/api/types/mount"
+
+	"github.com/tilt-dev/tilt/internal/docker"
+	"github.com/tilt-dev/tilt/pkg/logger"
+)
+
+const defaultK3dImage = "docker.io/rancher/k3d:v4.4.7"
+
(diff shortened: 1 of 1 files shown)


nicks (approved):

nicks:
also - fwiw - i do have a ctlptl branch that basically has k3d working, but there are some subtleties with the registry interop (k3d sets up their registry slightly differently than the others)

milas:
Nice! Also if you haven't already seen, I noticed recently they've got RCs up for v5.x -> https://github.com/rancher/k3d/releases (nothing raised an eyebrow re: `ctlptl` compat tho)

PR #5000, opened 2021-09-29
tilt: rename/organize codegen makefile targets

landism:
We have three codegen make targets:
* proto
* update-codegen
* update-codegen-starlark

### Problems
1. I find it easy to forget to run some / run them in the wrong order
2. "codegen" reasonably describes a few things that are needed, but "update-codegen" only does some of them
3. ditto "proto"
4. CI (I think) only checks `update-codegen` (and only half of it!)
5. If you have a merge conflict in the files generated by `update-codegen`, the (I think) natural solution is to run it to regenerate it from scratch, but it'll first run `update-codegen-starlark`, which tries to parse that package and fails because of the merge conflict's `>>>>`s

### Changes
1. Rename them to make them more apparently siblings
2. Have `make update-codegen` run them all
3. Check 1.5/3 of them instead of 1/3 in CI, and add TODOs for the other 1.5
  a. I started this branch specifically to get to 3/3 and failed, but it seems like a step in the right direction, anyway.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.PHONY: all proto install lint test test-go check-js test-js test-storybook integration wire-check wire ensure check-go goimports proto-webview proto-webview-ts vendor shellcheck release-container update-codegen
+.PHONY: all install lint test test-go check-js test-js test-storybook integration wire-check wire ensure check-go goimports vendor shellcheck release-container update-codegen update-codegen-go update-codegen-starlark update-codegen-ts
 
 all: check-go check-js test-js test-storybook
 
@@ -14,9 +14,6 @@ GO_PARALLEL_JOBS := 4
 CIRCLECI := $(if $(CIRCLECI),$(CIRCLECI),false)
 
 GOIMPORTS_LOCAL_ARG := -local github.com/tilt-dev/tilt
-
-proto:
-	toast proto-ts
 
 # Build a binary the current commit SHA
 install:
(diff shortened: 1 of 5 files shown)


landism:
d'aww:
```
go-to-protobuf -h ./hack/boilerplate.go.txt --packages github.com/tilt-dev/tilt/pkg/apis/core/v1alpha1 --proto-import /go/src/github.com/tilt-dev/tilt/vendor --apimachinery-packages -k8s.io/apimachinery/pkg/util/intstr,-k8s.io/apimachinery/pkg/api/resource,-k8s.io/apimachinery/pkg/runtime/schema,-k8s.io/apimachinery/pkg/runtime,-k8s.io/apimachinery/pkg/apis/meta/v1
scripts/update-protobuf-helper.sh: line 10: go-to-protobuf: command not found
```

That might be a bigger yak than I want to tackle at the moment.

milas:
Can you `go install "k8s.io/code-generator/cmd/go-to-protobuf@v0.22.2"` in CI [or rather add to CI Docker image]?

nicks (approved):

landism:
> Can you `go install "k8s.io/code-generator/cmd/go-to-protobuf@v0.22.2"` in CI [or rather add to CI Docker image]?

* `go install "k8s.io/code-generator/cmd/go-to-protobuf@v0.22.2"` in CI - didn't work, since other stuff was missing
* add to CI Docker image - increases the size of this yak more than I want to deal with at the moment

PR #5003, opened 2021-10-01
test: improve how we infer pod template hashes

nicks:
Hello @landism,

Please review the following commits I made in branch nicks/lu12:

247f0fd7ac42aab89120eb51af1ddcf9e04ba17c (2021-10-01 12:30:36 -0400)
test: improve how we infer pod template hashes

d80e6ee232c4c1e4919a547fce266deee8798412 (2021-09-30 16:42:25 -0400)
store: create helper objects for summarizing Discovery+Apply
Right now, we decide which pods to LiveUpdate based on a complex
set of information from Discovery+Apply. We want to move all the relevant
info into its own object, separate from the old K8sRuntimeState

### DIFF
--- a/internal/store/store.go
+++ b/internal/store/store.go
@@ -141,6 +141,15 @@ func (s *Store) Loop(ctx context.Context) error {
 	}
 	defer s.subscribers.TeardownAll(context.Background())
 
+	// Set up a defer handler, and make sure to unlock the state
+	// if the control loop is interrupted by a panic.
+	hasStateLock := false
+	defer func() {
+		if hasStateLock {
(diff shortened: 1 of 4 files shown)


internal/store/store.go
```diff
@@ -141,6 +141,15 @@ func (s *Store) Loop(ctx context.Context) error {
 	}
 	defer s.subscribers.TeardownAll(context.Background())
 
+	// Set up a defer handler, and make sure to unlock the state
```
nicks, line 144: this bug has been driving me crazy!!! it's what makes the tests timeout on panic

landism (approved):

PR #5009, opened 2021-10-04
script: update Toast configuration for Go 1.16+

milas:
We've been using an old image for Toast tasks based on Go 1.12
(the Dockerfile was updated but the digest still pointed to the
old one).

Things broke recently because `toast wire` was touching code that
required Go 1.16+ so couldn't compile anymore.

There's a bunch of annoying changes required for the protobuf stuff
to work, but everything is in order now. The versions of packages
here are newer than before, so the tasks have been re-run. The
generated protobuf code looks almost entirely different, for example.


milas:
Draft until I can run `make tilt-toast-container` and update the digest SHA, but I need to get GCP permissions sorted first 🤪

nicks:
re: "GCP permissions sorted first" - what command are you running, and what error are you getting? i just checked the permissions and it should be ok

nicks (approved):

PR #5013, opened 2021-10-05
engine: don't treat Jobs as "OK" until they complete

### LINKED ISSUE #4809: Improved status reporting for Jobs
Currently, Tilt does not have (much) special logic for K8s `Job` objects. In general, it follows the same paths as for a `Deployment`, where the child `Pod`(s) are discovered and their status used to determine the overall resource status.

This can cause weirdness (#3513, #4788) under normal circumstances for status reporting.

But unlike a normal `Pod`, the fact that it's "ready" just means that the job is _running_, and while it's useful to know it started up okay, it'd often be helpful to know that it's in progress and then finished.

This also plays into resource dependencies, imagine `db` (Deployment), `db:migrate` (Job), `app` (Deployment) resources. Currently, if `app` has a dependency on `db:migrate`, it'll start as _soon_ as the Job is "ready" (i.e. executing) but _not_ when it's complete, which is arguably the desired behavior!

### Current Behavior
`Job` entities are treated the same as `Deployment` for most purposes despite having different semantics.

### Why Do You Want This?
* More accurate status reporting
* Increased utility for resource dependencies (this is arguably a separate feature enabled by this)

landism:
Fixes #4809 

1. Jobs now have a runtime status of "pending" until the pod completes (they are currently "ok" as soon as the pod is ready)
2. A resource that has a resource_dep on a Job resource will now start only once the job's pod succeeds (it currently starts as soon as the job's pod is ready)

### DIFF
--- a/internal/engine/k8swatch/reducers.go
+++ b/internal/engine/k8swatch/reducers.go
@@ -101,7 +101,14 @@ func maybeUpdateStateForPod(ms *store.ManifestState, pod *v1alpha1.Pod) bool {
 	// Track the new version
 	runtime.Pods[podID] = pod
 
-	if len(pod.Containers) != 0 && (store.AllPodContainersReady(*pod) || pod.Phase == string(v1.PodSucceeded)) {
+	isReadyOrSucceeded := false
+	if ms.K8sRuntimeState().PodReadinessMode == model.PodReadinessComplete {
+		// for jobs, we don't care about whether it's ready, only whether it's succeeded
+		isReadyOrSucceeded = pod.Phase == string(v1.PodSucceeded)
+	} else {
+		isReadyOrSucceeded = len(pod.Containers) != 0 && store.AllPodContainersReady(*pod)
+	}
+	if isReadyOrSucceeded {
 		runtime.LastReadyOrSucceededTime = time.Now()
 	}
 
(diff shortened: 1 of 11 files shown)


internal/engine/k8swatch/reducers.go
```diff
@@ -101,7 +101,14 @@ func maybeUpdateStateForPod(ms *store.ManifestState, pod *v1alpha1.Pod) bool {
 	// Track the new version
 	runtime.Pods[podID] = pod
 
-	if len(pod.Containers) != 0 && (store.AllPodContainersReady(*pod) || pod.Phase == string(v1.PodSucceeded)) {
+	isReadyOrSucceeded := false
+	if ms.K8sRuntimeState().PodReadinessMode == model.PodReadinessComplete {
```
nicks, line 105: the model is "Complete" but the condition is "Succeeded".

Are we assuming "Complete" and "Succeeded" are interchangeable here?

(I know airflow calls this ALL_DONE vs ALL_SUCCESS - https://airflow.apache.org/docs/apache-airflow/stable/concepts/dags.html?highlight=all_done#trigger-rules)

landism: good point! changed model to "Succeeded"

PR #5014, opened 2021-10-05
api: implement ToggleButton reconciler

landism:


internal/controllers/core/togglebutton/reconciler.go
```diff
@@ -0,0 +1,291 @@
...
+				Name: types.NamespacedName{Name: toggleButton.Spec.StateSource.ConfigMap.Name},
+				GVK:  bGVK,
+			})
+		}
+	}
+	return result
+}
+
+func (r *Reconciler) Reconcile(ctx context.Context, request reconcile.Request) (reconcile.Result, error) {
+	r.mu.Lock()
```
nicks, line 79: i don't think you need a mutex here? there's nothing async in this code

internal/controllers/core/togglebutton/reconciler.go
```diff
@@ -0,0 +1,291 @@
...
+	}
+
+	return ctrl.Result{}, err
+}
+
+func (r *Reconciler) processClick(ctx context.Context, tb v1alpha1.ToggleButton) error {
+	uiButton := v1alpha1.UIButton{}
+	err := r.ctrlClient.Get(ctx, types.NamespacedName{Name: uibuttonName(tb.Name)}, &uiButton)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
```
nicks, line 117: usually you'd use the client.IgnoreNotFound idiom, as demonstrated here:
https://book.kubebuilder.io/cronjob-tutorial/controller-implementation.html#1-load-the-cronjob-by-name

internal/controllers/core/togglebutton/reconciler.go
```diff
@@ -0,0 +1,291 @@
...
+	}
+
+	// if there's a new click, pass the new value through to the ConfigMap
+	if uiButton.Status.LastClickedAt.After(r.lastClickProcessTimes[tb.Name]) {
+		foundInput := false
+		var newOnValue bool
+		for _, input := range uiButton.Status.Inputs {
+			if input.Name == actionUIInputName {
+				if input.Hidden == nil {
+					return fmt.Errorf("button %q input %q was not of type 'Hidden'", uiButton.Name, input.Name)
```
nicks, line 131: i don't think returning an error is what you want here. that tells the controller to requeue and try again later. you probably want this to be some sort of an invalid state on your ToggleButton?

internal/controllers/core/togglebutton/reconciler.go
```diff
@@ -0,0 +1,291 @@
...
+		}
+
+		ss := tb.Spec.StateSource.ConfigMap
+		if ss == nil {
+			return fmt.Errorf("ToggleButton %q has nil Spec.StateSource.ConfigMap", tb.Name)
+		}
+		var cm v1alpha1.ConfigMap
+		err := r.ctrlClient.Get(ctx, types.NamespacedName{Name: ss.Name}, &cm)
+		if err != nil {
+			return errors.Wrap(err, "fetching ToggleButton ConfigMap")
```
nicks, line 157: who's responsible for creating the ConfigMap? if there's no ConfigMap, should ToggleButton create one?

(probably want to handle NotFound separately from other errors)

internal/controllers/core/togglebutton/reconciler.go
```diff
@@ -0,0 +1,291 @@
...
+			default:
+				newOn = tb.Spec.DefaultOn
+			}
+		} else {
+			newOn = tb.Spec.DefaultOn
+		}
+
+		if newOn != tb.Status.On {
+			tb.Status.On = newOn
+			tb.Status.LastChange = metav1.NowMicro()
```
nicks, line 217: i think this is going to have race conditions and idempotency problems.

i'm not sure why this is part of Status at all though? i think these libraries generally try to be very level-based.

internal/controllers/core/togglebutton/reconciler.go
```diff
@@ -0,0 +1,291 @@
...
+		err := r.ctrlClient.Delete(ctx, &v1alpha1.UIButton{ObjectMeta: metav1.ObjectMeta{Name: uibuttonName(nn.Name), Namespace: nn.Namespace}})
+		return ctrlclient.IgnoreNotFound(err)
+	}
+
+	desiredUIButton, err := r.toDesiredUIButton(tb)
+	if err != nil {
+		return err
+	}
+
+	b := &v1alpha1.UIButton{ObjectMeta: metav1.ObjectMeta{Name: desiredUIButton.Name}}
```
nicks, line 239: the relationship between desiredUIButton and b is kind of twisty

would it be easier if toDesiredUIButton was toDesiredUIButtonSpec?

PR #5037, opened 2021-10-13
api: set UIResource.Status.DisableStatus counts

landism:
### Problem

UIResource needs to indicate whether the resource is disabled

### Solution

1. When the Tiltfile controller creates a UIResource object, populate its `Status.DisableStatus.Sources` field with the DisableSources of the resource's api objects (KubernetesApply, Cmd, etc)
2. In the UIResource reconciler, check those sources and update `Status.DisableStatus.{Enabled,Disabled}Count`

This might be a controversial approach!
1. If we create an object in the resource and give it a different DisableSource, it will not be reflected in UIResource's DisableStatus.
2. This means UIResource's DisableStatus reflects the *configured* disabled state (what's in the ConfigMap) rather than the *observed* disabled state (whether the individual api objects actually report that they're disabled) - this has pros and cons, but I think on balance this leaves less room for weirdness, ui-wise.
3. I seem to recall seeing something about it being problematic to read from and write to the same object's status in a reconciler, but I can't find anything about that now.

I like it because:
1. It seems vastly simpler than having UIResource watch / iterate over all the types that might belong to a resource
2. We can swap it out for a more sophisticated approach later w/o changing the api

If there's a good reason not to go this route for now, I'll start looking into having the reconciler watch / iterate over all the types.


landism:
DisableStatus now gets updated without involving the reconciler, so we should be good on the cycle situation.

It's still a bit weird that the subscriber is reading sources from DisableStatus, but at least things are more manageable.

PR #5044, opened 2021-10-14
engine: cmdserver cancels cmds when resource is disabled

landism:
[This](https://github.com/tilt-dev/tilt/compare/mlandis/sc-12735/cmdserver-handles-disablesource-status?expand=1#diff-78c811a3ae9e9c8910344b3cc909b39c4e6e8152cc73586015a40006451f7f66R127) tempts me to store a single DisableSource on ManifestState further upstream


internal/testutils/manifestbuilder/manifestbuilder.go
```diff
@@ -97,6 +99,11 @@ func (b ManifestBuilder) WithLocalServeCmd(cmd string) ManifestBuilder {
 	return b
 }
 
+func (b ManifestBuilder) WithLocalDisableSource() ManifestBuilder {
```
nicks, line 102: can you add some comments on when you'd want to call this and when you wouldn't want to call this? I don't totally understand why it's an option at all, rather than something that all Manifests get.

internal/store/uiresources/reducers.go
```diff
@@ -1,11 +1,16 @@
...
 import (
+	"fmt"
+
 	"github.com/tilt-dev/tilt/internal/store"
 )
 
 func HandleUIResourceUpsertAction(state *store.EngineState, action UIResourceUpsertAction) {
 	n := action.UIResource.Name
+	if state.UIResources[n] != nil && len(state.UIResources[n].Status.DisableStatus.Sources) == 0 && len(action.UIResource.Status.DisableStatus.Sources) != 0 {
+		fmt.Printf("adding UIResource %s w/ DisableSource for first time\n", n)
```
nicks, line 12: did you intend to delete this?

internal/engine/local/servercontroller.go
```diff
@@ -37,6 +40,9 @@ type ServerController struct {
 	createdTriggerTime map[string]time.Time
 	client             ctrlclient.Client
 
+	mu         sync.Mutex
+	cmdServers map[string]CmdServer
```
nicks, line 44: this seems really bizarre - why does this PR add new state to the ServerController? i tried reading the code a few times and couldn't figure it out, and also couldn't find any comments to explain it.

internal/testutils/manifestbuilder/manifestbuilder.go
landism: the reason for this no longer exists and doing it for all manifests is good. thanks!

internal/engine/local/servercontroller.go
landism: I added some comments.

It might be more consistent to replicate this back to EngineState rather than storing them here, but I couldn't think of a reason it'd be worth the time/code.

nicks: ah, ok

nicks (approved):

PR #5058, opened 2021-10-21
Use default text input value

alisaifee:
# Description
When a `UIButton` has a text input which contains a `defaultValue` this value was displayed but never used in the final command execution.


alisaifee:
@landism thanks! 

I wasn't sure if it was intended to be handled on the front end only - but now that I look at it the defaultValue for boolean inputs was also not handled in the controller. 

I'll close this PR!

PR #5086, opened 2021-10-27
web: show what resources are waiting on

milas:
The `UIResource` API now exposes waiting info including _which_
other objects it's waiting on. Currently, there are two types of
possible objects: `ImageMap` and `UIResource`. The former is
used for shared/base image builds and the latter for `resource_deps`
from the Tiltfile.


hyu (approved):
Very cool! 

Current behavior here is that when you have a really long resource name, the status section will expand to be taller. That seems fine for now, though I can imagine an edge case where if a bunch of services are all waiting on a resource with a really long name, it might mess up your sidebar layout as the status cycles from "waiting on" back to pending.

So, entirely optional for this PR but we might want to truncate this in the future if it proves problematic.
Screen Shot 2021-10-29 at 3 40 34 PM

milas:
resource dependency on resource with very long name truncated in Tilt UI

Pushed e2a3173 which adds truncation to the status line; the tooltip provides the full text still

PR #5097, opened 2021-10-29
localexec: provide a facade for local command exec

milas:
This is a Tilt-centric local process executor meant for commands
that run to completion. The `Cmd` controller, which manages
long-running processes for `local_resource`'s `serve_cmd`, has
its own unexported implementation that's focused more on async
signaling of process status for a daemon-style process that runs
perpetually.

There is overlap between these two, specifically in that they
take care to use process groups and terminate the entire group to
avoid leaking resources.

The Tiltfile loader has been updated to use this for running
`local()` commands, which should be invisible to consumers, but
does mean that it's now possible to inject a fake execer for
Tiltfile loading and return pre-canned responses for `local()`
calls, which could be useful.


internal/localexec/execer.go
```diff
@@ -0,0 +1,200 @@
...
+		return -1, err
+	}
+
+	// there's a data race on this value if the process termination + context cancellation happen
+	// very close together, so all access should be done via atomic
+	//
+	// NOTE: if this condition occurs, it's possible the process will be reported as killed (127)
+	// even though it actually had already exited because we force an exit code to prevent returning
+	// an exit code of 0 despite having triggered a SIGKILL in the common case
+	var exitCode int64
```
nicks, line 96: this is pretty complex and difficult to reason about...have you considered using a sync.Once? smth like:

```
var once sync.Once
defer cancel()

go func() {
  <-ctx.Done()
  once.Do(func() { exitCode = 137; procutil.KillProcessGroup(osCmd); cancel() })
}()
```

internal/localexec/execer.go
```diff
@@ -0,0 +1,200 @@
...
+			procutil.KillProcessGroup(osCmd)
+		}
+	}()
+
+	// this WILL block on child processes, but that's ok since we handle the timeout termination in a goroutine above
+	// and it's preferable vs using Process::Wait() since that complicates I/O handling (Cmd::Wait() will
+	// ensure all I/O is complete before returning)
+	err = osCmd.Wait()
+	procDone <- struct{}{}
+	close(procDone)
```
nicks, line 116: i don't think the synchronization here will behave like you expect, because you're using a buffered channel

PR #5098, opened 2021-10-29
web: reorganize sidebar components

lizzthabet:
While working on the display of disabled resources in the resource detail view, I reorganized the sidebar components to simplify the logic between group/non-group sidebar views, as well as move display logic from the parent sidebar component into children components. This should make it easier to display disabled components for group/non-group views. 

This PR just includes the refactoring work, so the disable resources display PR doesn't get too huge. 🙂

### DIFF
--- a/web/src/SidebarResources.tsx
+++ b/web/src/SidebarResources.tsx
@@ -3,7 +3,7 @@ import {
   AccordionDetails,
   AccordionSummary,
 } from "@material-ui/core"
-import React, { ChangeEvent, PropsWithChildren } from "react"
+import React, { ChangeEvent } from "react"
 import styled from "styled-components"
 import { AnalyticsType } from "./analytics"
 import { FeaturesContext, Flag } from "./feature"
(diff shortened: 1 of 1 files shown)


web/src/SidebarResources.tsx
```diff
@@ -108,20 +122,31 @@ const SidebarGroupDetails = styled(AccordionDetails)`
...
 }
 
-function SidebarItemsView(props: SidebarProps & { groupView?: boolean }) {
+function SidebarListSectionItems(props: SidebarSectionProps) {
+  if (
+    props.items.length === 0 &&
+    props.resourceListOptions.resourceNameFilter.length > 0
+  ) {
+    return (
+      <SidebarListSectionItemsRoot>
```
nicks, line 142: i like the move of this logic into the component

web/src/SidebarResources.tsx
```diff
@@ -108,20 +122,31 @@ const SidebarGroupDetails = styled(AccordionDetails)`
...
+      {props.sectionName && (
+        <SidebarListSectionName>{props.sectionName}</SidebarListSectionName>
+      )}
+      <SidebarListSectionItems {...props} />
+    </>
   )
 }
 
-function SidebarItemsView(props: SidebarProps & { groupView?: boolean }) {
+function SidebarListSectionItems(props: SidebarSectionProps) {
```
nicks, line 136: tbh, "Items" feels like a weird name for a standalone react component

lizzthabet: haha yeah, naming all these components in this file is challenging! i opted for `items` to better describe the `li` in `ul`, but i may tweak the names a bit more in the pr following this one.

PR #5104, opened 2021-11-01
ci: use circleci orb for Go on macOS

milas:
We were only using Homebrew for installing Go, so using CircleCI orb seems easier than trying to debug whatever is wonky with the caching.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -5,6 +5,7 @@ orbs:
   win: circleci/windows@2.2.0
   gcp-cli: circleci/gcp-cli@1.8.3
   kubernetes: circleci/kubernetes@0.9.1
+  go: circleci/go@1.7.0
 jobs:
   build-linux:
     resource_class: medium+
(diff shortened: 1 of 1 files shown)


milas:
@nicks TBH CircleCI is the CI system I've got the least familiarity with, so not sure if there's issues with the Orb stuff, but this seemed more straightforward than trying to debug Homebrew inside CircleCI 😛

nicks (approved):

.circleci/config.yml
```diff
@@ -118,21 +119,8 @@ jobs:
...
-      # 2) hits github in a way that leads to flakyness
-      - run: brew info ${HOMEBREW_PACKAGES} || brew update || true
-      - run: HOMEBREW_NO_AUTO_UPDATE=true brew install ${HOMEBREW_PACKAGES}
-      - save_cache:
-          paths:
-            - /usr/local/Homebrew
-          key: homebrew_cache_v14
-      - run: echo 'export PATH="/usr/local/opt/go@1.17/bin:$PATH"' >> $BASH_ENV
+      - go/install:
+          version: "1.17.2"
```
nicks, line 123: can you also add cache: true? https://circleci.com/developer/orbs/orb/circleci/go

nicks: oh wait nm cache: true is the default

milas:
On the bright side, the Orb install (which grabs the official binary installer) takes 6 secs so faster than restoring Homebrew cache 😅

PR #5110, opened 2021-11-01
api: custom apply command support for KubernetesApply

milas:
Allow deeper integration with custom deploy tools (e.g. Helm) by
invoking a command to perform the apply instead of having Tilt
call the K8s API directly.

The invoked command MUST return YAML for applied entities on
stdout and return an exit code of 0.

### DIFF
--- a/integration/custom_deploy/Tiltfile
+++ b/integration/custom_deploy/Tiltfile
@@ -1,0 +1,26 @@
+
+v1alpha1.file_watch(
+    name='custom-deploy-fw',
+    watched_paths=['./deploy.yaml']
+)
+
+v1alpha1.kubernetes_apply(
+    name='custom-deploy-int-test',
+    timeout='10s',
+    port_forward_template_spec=v1alpha1.port_forward_template_spec(
+        forwards=[
+            v1alpha1.forward(
+                local_port=54871,
+                container_port=80
+            )
+        ]
+    ),
+    cmd=v1alpha1.kubernetes_apply_cmd(
+        args=['sh', '-c', 'kubectl apply -o yaml -f "${DEPLOY_FILE}"'],
+        dir=config.main_dir,
+        env=['DEPLOY_FILE=deploy.yaml']
+    ),
+    restart_on=v1alpha1.restart_on_spec(
+        file_watches=['custom-deploy-fw']
+    )
+)
--- a/internal/controllers/core/kubernetesapply/reconciler.go
+++ b/internal/controllers/core/kubernetesapply/reconciler.go
@@ -1,6 +1,7 @@
 package kubernetesapply
 
 import (
+	"bytes"
 	"context"
 	"fmt"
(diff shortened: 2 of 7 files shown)


milas:
(Need to wrap up integration test, so draft for now)

internal/controllers/core/kubernetesapply/reconciler.go
```diff
@@ -346,8 +347,40 @@ func (r *Reconciler) runYAMLDeploy(ctx context.Context, spec v1alpha1.Kubernetes
 }
 
 func (r *Reconciler) runCmdDeploy(ctx context.Context, spec v1alpha1.KubernetesApplySpec) ([]k8s.K8sEntity, error) {
-	// TODO(milas): implement Cmd apply
-	return nil, errors.New("Cmd apply not supported")
+	cmd := model.Cmd{
+		Argv: append([]string(nil), spec.Cmd.Args...),
```
nicks, line 351: i thought we could generally assume kubernetes model objects are immutable?

internal/controllers/core/kubernetesapply/reconciler.go
```diff
@@ -346,8 +347,40 @@ func (r *Reconciler) runYAMLDeploy(ctx context.Context, spec v1alpha1.Kubernetes
...
+	if timeout == 0 {
+		timeout = v1alpha1.KubernetesApplyTimeoutDefault
+	}
+	ctx, cancel := context.WithTimeout(ctx, timeout)
+	defer cancel()
+
+	var stdoutBuf bytes.Buffer
+	runIO := localexec.RunIO{
+		Stdout: &stdoutBuf,
+		Stderr: logger.Get(ctx).Writer(logger.WarnLvl),
```
nicks, line 366: i wouldn't use WarnLvl here...WarnLvl is more for discrete warnings that we want the UI to call attention to. Stderr is generally more informational. Let's use InfoLvl

nicks (approved):

milas:
FYI made small changes based on code review comments and added an integration test (currently using the `v1alpha.kubernetes_apply` helpers with a command that calls `kubectl apply -oyaml -f ...`)

integration/custom_deploy/Tiltfile
```diff
@@ -0,0 +1,26 @@
+
+v1alpha1.file_watch(
+    name='custom-deploy-fw',
+    watched_paths=['./deploy.yaml']
+)
+
+v1alpha1.kubernetes_apply(
```
nicks, line 7: nice!

nicks (approved):

integration/custom_deploy/Tiltfile
milas: It was really fast/enjoyable to use these helpers! I didn't really refer to docs and the only mistake I made was originally passing a number instead of duration string for `Timeout`

PR #5114, opened 2021-11-02
api: add helper for registering RestartOn watches

milas:
I did not actually add a watch when adding `RestartOn` to the
`KubernetesApply` spec, so it wouldn't actually work.

I realized that the Tiltfile reconciler was also only watching
`FileWatch` and not `UIButton` so thought a helper was probably
warranted to ensure these all stay in sync.

### DIFF
--- a/internal/controllers/apis/restarton/restarton.go
+++ b/internal/controllers/apis/restarton/restarton.go
@@ -6,7 +6,10 @@ import (
 
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/source"
 
 	"github.com/tilt-dev/tilt/internal/controllers/indexer"
 	"github.com/tilt-dev/tilt/internal/sliceutils"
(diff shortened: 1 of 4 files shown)


internal/controllers/core/cmd/controller.go
```diff
@@ -557,19 +568,6 @@ func indexCmd(obj client.Object) []indexer.Key {
 	return result
 }
 
-func (r *Controller) CreateBuilder(mgr ctrl.Manager) (*builder.Builder, error) {
```
milas, line 560: (I moved this to the top of the file so it's next to the constructor like the others for consistency/ease of finding it)

internal/controllers/apis/restarton/restarton.go
```diff
@@ -235,3 +247,18 @@ func FilesChanged(restartOn *v1alpha1.RestartOnSpec, fileWatches map[string]*v1a
...
+// It should be called by a controllers CreateBuilder() method.
+//
+// Controllers MUST also call ExtractKeysForIndexer as part of their indexer initialization.
+func RegisterWatches(builder *builder.Builder, indexer *indexer.Indexer) {
+	for _, t := range restartOnTypes {
+		// this is arguably overly defensive, but a copy of the type object stub is made
+		// to avoid sharing references of it across different reconcilers
+		obj := t.DeepCopyObject().(client.Object)
+		builder.Watches(&source.Kind{Type: obj},
+			handler.EnqueueRequestsFromMapFunc(indexer.Enqueue))
```
nicks, line 262: feels like we're missing some underlying abstraction. 

This implicitly relies on RestartOn already being registered with the indexer.

I wonder if the indexer should have some sort of plugin system where you're registering both Watches and Key indexers?

nicks (approved):

internal/controllers/apis/restarton/restarton.go
milas: Yeah, if the `indexer.Indexer` accepted multiple `KeyFunc`s and then took the union of them all, we could do this in one shot.

You'll still need an "extractor" func to go from `client.Object` -> `(*v1alpha1.RestartOnSpec, *v1alpha1.StartOnSpec)` but that's not the worst 🤷 

I'll play around with it briefly and see if I can open another PR that combines them

PR #5116, opened 2021-11-02
server: cache certs a bit better on startup

nicks:
Hello @milas, @landism,

Please review the following commits I made in branch nicks/headless:

b73311246150d59b5410edf59f4575d40a6faee7 (2021-11-02 11:33:19 -0400)
server: cache certs a bit better on startup
cert generation is very cpu intensive!


milas (approved):
This has made me think that we should have a `TILT_INTEGRATION_PORT` env var to pick a specific port vs a randomized one (yes, I realize the irony as the one who introduced the randomized port behavior 😝) so that CI can generate 1x and re-use it across each integration test

PR #5120, opened 2021-11-02
engine: remove disabled manifests from trigger queue

landism:


internal/engine/configs/trigger_queue_subscriber.go
```diff
@@ -70,3 +76,26 @@ func (s *TriggerQueueSubscriber) OnChange(ctx context.Context, st store.RStore,
...
+				st.Dispatch(store.NewLogAction(mn, spanID, logger.InfoLvl, nil, []byte(msg)))
+
+				st.Dispatch(RemoveFromTriggerQueueAction{Name: mn})
+			}
+		}
+	}
+}
+
+type RemoveFromTriggerQueueAction struct {
+	Name model.ManifestName
```
nicks, line 98: shouldn't this be handled in the reducer loop? it seems weird for this state to bounce out to a subscriber, which then dispatches an action back into the reducer.

landism: yeah, that makes more sense

nicks (approved):

PR #5122, opened 2021-11-02
api: set annotation on disable buttons

landism:
Allow the UI to distinguish disable buttons from other buttons.

The UI design for this feature involves treating disable buttons different from other UIButtons:
* disable buttons go on a different row in the resource pane action bar, and on the far right
* disable buttons are not shown on the table view (those will be handled via a new design Surbhi is working on)

Having a list of special button types feels a bit hacky, but none of the alternatives I could think of feel great, either:
* we could just add an annotation instead of a spec field, but that seems like a thing to stay away from, as a rule
* the frontend could just pattern-match the disable button name format
* we could add a new UI location for "bottom right of resource pane action bar, and not table view", but I don't know how likely it is we'll ever want to put another button there

I'd love a better solution if one comes to mind!

On the bright side, this approach is easy to back off from if/when we decide to go another route once we have more cases like this and have a better idea of a more general solution - we just stop reading/writing this field and use that solution instead.

edit: going with an annotation after all!


nicks (approved):

nicks:
i'm fine with this. 

but fwiw, a common pattern in kubernetes is to use annotations to specify "special" per-controller behavior that doesn't make sense in the general-purpose object schema. Here are some examples: https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/annotations/

landism:
> but fwiw, a common pattern in kubernetes is to use annotations to specify "special" per-controller behavior that doesn't make sense in the general-purpose object schema

oh, great, I'll use an annotation, then

nicks (approved):

PR #5126, opened 2021-11-03
web: fix a bug in log filtering

nicks:
Hello @lizzthabet,

Please review the following commits I made in branch nicks/logfilter:

91a0c8c407e427532b9ba0d033bea1ddde2f299e (2021-11-03 09:53:36 -0400)
web: fix a bug in log filtering

### DIFF
--- a/web/src/logfilters.test.ts
+++ b/web/src/logfilters.test.ts
@@ -80,6 +80,14 @@ describe("Log filters", () => {
           )
         })
 
+        it("is not stateful", () => {
+          let parsed = parseTermInput("line 1")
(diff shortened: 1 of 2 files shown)


lizzthabet (approved):
i learned something new today! 💡 

> [T]he g flag sets the `lastIndex` property as well.
> A very important side effect of this is if you are reusing the same regex instance against a matching string, it will eventually fail because it only starts searching at the lastIndex.

nicks:
i knowwww

nicks:
i knowww but only because i've learned this too many times the hard way :sob:

PR #5129, opened 2021-11-03
engine: remove duplicate PortForward definition from K8sTarget

milas:
Rely on the `PortForwardTemplateSpec` as the canonical source for
port forwards rather than storing a duplicate version of them on
the target.

(Note: the API `PortForward` currently does NOT get the name/path
propagated to it, but this is a straightforward addition in the
future if we want that data exposed there.)


nicks (approved):
woooo

PR #5130, opened 2021-11-03
engine: remove "to-be-deployed" ObjectRefs from K8sTarget

milas:
We won't always have this anymore since the "to-be-applied" might
be invoking an external deploy command rather than a YAML blob.

There is some impact to this change (unlike the wholly redundant
removal of port forwards here recently):

  1. The `Session` object (used for Tilt CI) will _change_ a target
     type from "server" to "job" since it no longer knows up front
     until it gets deployed. This is a bit odd; longer-term, maybe
     it makes sense to provide explicit user control here and in
     the `k8s_yaml()` case to infer it up front to set a value.

  2. K8s event + service watches won't be started for a given
     namespace until _after_ the first deploy. In practice, this
     should not mean missed events, as the watch will return the
     existing ones when started.

### DIFF
--- a/internal/engine/session/conv.go
+++ b/internal/engine/session/conv.go
@@ -44,13 +44,8 @@ func k8sRuntimeTarget(mt *store.ManifestTarget) *session.Target {
 
 	target := &session.Target{
 		Name:      fmt.Sprintf("%s:runtime", mt.Manifest.Name.String()),
+		Type:      k8sTargetType(mt),
 		Resources: []string{mt.Manifest.Name.String()},
-	}
-
-	if mt.Manifest.IsK8s() && mt.Manifest.K8sTarget().HasJob() {
-		target.Type = session.TargetTypeJob
-	} else {
-		target.Type = session.TargetTypeServer
 	}
 
 	// a lot of this logic is duplicated from K8sRuntimeState::RuntimeStatus()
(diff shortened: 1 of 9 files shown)


internal/engine/session/conv.go
```diff
@@ -44,15 +44,10 @@ func k8sRuntimeTarget(mt *store.ManifestTarget) *session.Target {
 
 	target := &session.Target{
 		Name:      fmt.Sprintf("%s:runtime", mt.Manifest.Name.String()),
+		Type:      k8sTargetType(mt),
 		Resources: []string{mt.Manifest.Name.String()},
 	}
 
-	if mt.Manifest.IsK8s() && mt.Manifest.K8sTarget().HasJob() {
```
nicks, line 50: fwiw: i suspect what's eventually going to happen is that:
1) we're going to add PodReadinessMode to KubenetesApply/KubernetesDiscovery Spec
2) There will be a PodReadinessMode="auto" that infers based on YAML
3) KubernetesApplyStatus will reflect whether the inferred readiness is Server or Job
4) KuberentesApply will have a metav1.Condition{Name: "Ready"} that reflects readiness

and so a lot of this inference will be in the API

nicks (approved):

internal/engine/session/conv.go
milas: That's what I was imagining, so glad we're on the same page.

(The short-term impact here is not my favorite...as a user, I would generally assume a `type` field would be immutable. But since the data within any given version of a `Session` obj will be consistent, it doesn't actually introduce any races so 🤷)

PR #5132, opened 2021-11-04
engine: remove DisplayNames from K8sTarget

milas:
This is now a computed property on `K8sRuntimeState` that uses
the deployed references.

These are included in `UIResource`, but not shown in the web
UI. The TUI does use them, so there's a slight behavior change
here that they won't show up until after applied. (Note that
they are also only used in the case when you've set pod readiness
mode to `ignore`.)


milas:
This is the last of fields that need to get eliminated from `K8sTarget` to enable smoother `custom_deploy` support 🎉

nicks (approved):

PR #5133, opened 2021-11-04
engine: allow file dependencies for K8sTarget

milas:
This will be used for `custom_deploy`, where the target has a
command instead of YAML.

(In the `k8s_yaml()` case, the file dependencies actually belong
to the Tiltfile because there's additional K8s entity <> resource
mapping logic that requires the "global" Tiltfile state, so it's
not sufficient to only trigger the target.)

### DIFF
--- a/pkg/model/k8s_target.go
+++ b/pkg/model/k8s_target.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"reflect"
 
+	"github.com/tilt-dev/tilt/internal/sliceutils"
 	"github.com/tilt-dev/tilt/pkg/apis/core/v1alpha1"
 )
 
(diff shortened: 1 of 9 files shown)


pkg/model/k8s_target.go
```diff
@@ -44,7 +45,18 @@ type K8sTarget struct {
...
+
+	// pathDependencies are files required by this target.
+	//
+	// For Tiltfile-based, YAML-driven (i.e. `k8s_yaml()`) resources, this is
+	// NOT used because it's not sufficient to reload the YAML and re-deploy;
+	// there is a lot of post-Tiltfile-load logic for resource assembly, image
+	// locator injection, etc. As a result, these resources have their YAML
+	// files registered as "config files", which cause the Tiltfile to be
+	// re-evaluated.
+	pathDependencies []string
```
nicks, line 58: this seems fine to me, but how hard would it be to add a FileWatchSpec to the K8sTarget? that way you wouldn't need to do the two-step Tiltfile -> private fields, private fields -> FileWatch

nicks (approved):

pkg/model/k8s_target.go
milas: It's sorta icky in either case - knowledge of its existence never needs to go out of the `tiltfile` package, but since the `ToFileWatchObjects` and friends methods operate primarily off manifests, it's the most convenient way to make it visible.

At the same time, bypassing `ToFileWatchObjects` completely means missing any post-processing it does, which includes setting up the `DisableSource`, so I _think_ what might be useful is to make it ergonomic for Tiltfile functions to create API objects, and then ensure the same post-processing logic runs across implicitly generated objects as well as any explicitly created ones (e.g. by setting `DisableSource` for known types based on annotations - this also allows the API helper functions exposed to end users to benefit from that, although maybe that's a step too far 🤔)

milas: Going to think on this some more though - I _might_ punt and merge as-is to get `custom_deploy` over the line and then revisit since it should be a straightforward refactor whichever direction we go

milas: (I did end up merging just now - going to revisit next week!)

PR #5135, opened 2021-11-04
api: fix Live Update selector inference for LU-only targets

milas:
There is no corresponding `ImageMap` for Live Update-only targets,
so it can be skipped.

### DIFF
--- a/integration/live_update_only/Tiltfile
+++ b/integration/live_update_only/Tiltfile
@@ -1,0 +1,21 @@
+
+# TODO(milas): this test is failing with live_update_v2 because the LU Reconciler
+#   stops after seeing a path that it can't sync, but a full BaD never happens,
+#   and the resource gets stuck in update Pending
+# disable_feature('live_update_v2')
+
+custom_build(
+    ref='nginx',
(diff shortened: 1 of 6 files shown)


milas:
@nicks This fixes one bug but there's still an issue for LU-only targets - the integration test will reproduce it.

To reproduce:
1. `go run ./cmd/tilt up -f ./integration/live_update_only/Tiltfile`
2. `touch ./integration/live_update_only/web/index.html`
3. `touch ./integration/live_update_only/special.txt`

It'll get stuck in a pending state and won't ever fallback to a full image build (no-op) & deploy. To be fair, it's a super weird case because the deploy won't actually do anything since the YAML hasn't changed and we don't have an image build! (And also having deps outside your sync path that are unusable is illogical.)

However...it matters for `custom_deploy` because that _will_ result in the custom command being re-invoked, but it gets hung.

nicks:
hmmm...digging into it a bit, something deeply weird is happening

nicks:
sent as https://github.com/tilt-dev/tilt/pull/5137

milas:
Fix got included (along with more) as part of #5137

PR #5136, opened 2021-11-04
tiltfile: add custom_deploy function

milas:
`custom_deploy` allows deploying resources to K8s using a custom
command (instead of having Tilt use the K8s API directly).

The resource supports Live Update, pod log streaming, as well as
standard status information.


milas:
Still cleaning this up and doing some more testing, but it's functional in its current state!

I put it behind a FF so that we can also merge and do some things incrementally without worrying too much about any limitations/issues.

milas:
Okay, tests are passing on this and it's behind a FF so I think this is a reasonable place for a review + merge.

Still to do (as separate PRs) before this is finalized:
- [ ] Add a "teardown" command for `tilt down` / KA reconciler GC
- [ ] Add a `cmd` func/struct to avoid having 8 command related args

internal/feature/flags.go
```diff
@@ -71,6 +72,10 @@ var MainDefaults = Defaults{
 		Enabled: false,
 		Status:  Active,
 	},
+	CustomDeploy: Value{
```
nicks, line 75: what's the point of putting this behind the flag? i feel like features are more for changes in behavior that we don't want to expose yet. we don't usually do them for new tiltfile functions.

internal/tiltfile/tiltfile_state.go
```diff
@@ -351,6 +363,7 @@ const (
 	k8sKindN                    = "k8s_kind"
 	k8sImageJSONPathN           = "k8s_image_json_path"
 	workloadToResourceFunctionN = "workload_to_resource_function"
+	customDeployN               = "custom_deploy"
```
nicks, line 366: we should probably put kubernetes/k8s in the name somewhere? since it's expected to return k8s yaml

internal/feature/flags.go
milas: I need to make a couple breaking changes, specifically new required teardown command, and as part of that I want to switch to exclusively use `cmd` objects from #5143 to avoid having 8 command args.

Given timing w/ release, I didn't want to have a Tilt version with an exposed "beta" `custom_deploy`, so was using this as a way to chunk some of the Tiltfile stuff.

(Happy to also roll those changes into this PR and/or just accept Tiltfile API breakage in this instance)

internal/tiltfile/tiltfile_state.go
milas: That's a great point - `k8s_custom_deploy` seem reasonable?

internal/tiltfile/tiltfile_state.go
```diff
@@ -1110,6 +1118,67 @@ func (s *tiltfileState) translateK8s(resources []*k8sResource, updateSettings mo
...
+
+	var deps []string
+	if r.customDeploy != nil {
+		deps = r.customDeploy.deps
+		applySpec.Cmd = &v1alpha1.KubernetesApplyCmd{
+			Args: r.customDeploy.cmd.Argv,
+			Dir:  r.customDeploy.cmd.Dir,
+			Env:  r.customDeploy.cmd.Env,
+		}
+		applySpec.RestartOn = &v1alpha1.RestartOnSpec{
```
nicks, line 1152: fwiw - the pattern that i've found most helpful is to generate references to other API objects in the tiltfile reconciler, rather than in the tiltfile executor code. like so: https://github.com/tilt-dev/tilt/blob/master/internal/controllers/core/tiltfile/api.go#L56

mostly because it's easier to test that way.

internal/tiltfile/tiltfile_state.go
nicks: :+1:

internal/tiltfile/custom_deploy.go
```diff
@@ -0,0 +1,97 @@
...
+		if imageSelector == "" {
+			return nil, fmt.Errorf("custom_deploy: image_selector cannot be empty")
+		}
+
+		ref, err := container.ParseNamed(imageSelector)
+		if err != nil {
+			return nil, fmt.Errorf("can't parse %q: %v", imageSelector, err)
+		}
+
+		img := &dockerImage{
```
nicks, line 80: ack, hopefully we can get rid of this hack soon

internal/feature/flags.go
nicks: eh, if the flag makes you more comfortable, that's ok - but i don't think the existence of the flag changes anything. (the kinds of people that find an undocumented function will also find the flag that turns it on, and you'll break those people anyway)

internal/tiltfile/tiltfile_state.go
```diff
@@ -1110,6 +1118,67 @@ func (s *tiltfileState) translateK8s(resources []*k8sResource, updateSettings mo
 	return result, nil
 }
 
+func (s *tiltfileState) k8sDeployTarget(targetName model.TargetName, r *k8sResource, imageTargets []model.ImageTarget, updateSettings model.UpdateSettings) (model.K8sTarget, error) {
```
nicks, line 1121: woooo

nicks (approved):

internal/tiltfile/tiltfile_state.go
milas: Ahhh interesting, I'll have to play around with that approach

internal/tiltfile/custom_deploy.go
milas: Ugh, yeah, it's sketchy. On the bright side, this will _hopefully_ make the refactor easier in the future since there will be more test coverage around some of the more extreme use cases 🤞 

(I also just added a `HACK` comment explaining the `:` command 🙊)

nicks: i think we're mostly there - just need to be able to add LiveUpdates to K8sTarget (instead of ImageTarget)

PR #5137, opened 2021-11-04
buildcontrol: now that LiveUpdate fires BuildStarted events, we can't synchronize based on number of BuildStarted events

nicks:
Hello @milas,

Please review the following commits I made in branch nicks/lu-only-fix:

2f74377a9200e1d6e21d3344b76cde58c0973e8c (2021-11-04 19:57:35 -0400)
buildcontrol: now that LiveUpdate fires BuildStarted events, we can't synchronize based on number of BuildStarted events

12fa319d6c0e61b9ca317aa40aa0743e993a5dcb (2021-11-04 17:39:47 -0400)
api: fix Live Update selector inference for LU-only targets
There is no corresponding `ImageMap` for Live Update-only targets,
so it can be skipped.

### DIFF
--- a/integration/live_update_only/Tiltfile
+++ b/integration/live_update_only/Tiltfile
@@ -1,0 +1,21 @@
+
+# TODO(milas): this test is failing with live_update_v2 because the LU Reconciler
+#   stops after seeing a path that it can't sync, but a full BaD never happens,
+#   and the resource gets stuck in update Pending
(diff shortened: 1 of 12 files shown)


milas (approved):

milas:
Oh, nice! I definitely had never seen that synchronization code before 🙈

PR #5139, opened 2021-11-05
api: fix race during FileWatch cancellation

milas:
It's possible for the monitor to get cancelled while in the midst
of reporting an event, which will cause the API operations to fail.

We don't want to treat these as fatal errors: since the monitor
was cancelled, the failed update was stale anyway and we should
just silently ignore it.

### DIFF
--- a/internal/controllers/core/filewatch/controller_test.go
+++ b/internal/controllers/core/filewatch/controller_test.go
@@ -6,6 +6,7 @@ import (
 	"reflect"
 	"runtime"
 	"strconv"
+	"strings"
 	"testing"
 	"time"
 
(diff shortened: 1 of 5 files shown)


internal/controllers/core/filewatch/controller_test.go
```diff
@@ -348,3 +349,42 @@ func TestController_Disable_Ignores_File_Changes(t *testing.T) {
 	f.MustGet(key, &fwAfterDisable)
 	require.Equal(t, 0, len(fwAfterDisable.Status.FileEvents))
 }
+
+func TestController_IgnoreErrorsOnCancel(t *testing.T) {
```
milas, line 353: I can be convinced this test should not exist because it's doing some deeply brittle/sketchy things reliant on the internals

internal/controllers/fake/client.go
```diff
@@ -39,6 +45,12 @@ type fakeStatusWriter struct {
 }
 
 func (f fakeStatusWriter) Update(ctx context.Context, obj client.Object, opts ...client.UpdateOption) error {
+	if ctxErr := ctx.Err(); ctxErr != nil {
+		// controller-runtime fake ignores context; check it here to allow controllers to test
+		// handling of cancellation related errors
+		return ctxErr
+	}
```
milas, line 52: FYI There's a non-zero chance this is going to eventually expose races in controller tests 😬

nicks (approved):
fwiw - i feel like we'll eventually want to get rid of ErrorAction, and instead have some way for the UI to report when a FileWatch is stuck and needs to be rebooted. but this seems like a good fix for now.

milas:
Yeah, I think the pattern where we update internal state and then queue a reconcile which has built-in retry semantics is much more robust.

FYI I had to apply a similar change in a2771e27e to the `Cmd` reconciler - in practice, it wasn't an issue since it was on the global context, which would only be canceled during shutdown, but it was causing tests to fail since the context gets canceled on test cleanup but a goroutine might still try to do an update after that 🙈

PR #5140, opened 2021-11-05
liveupdate: print pod name and container name instead of container id

nicks:
Hello @milas, @landism,

Please review the following commits I made in branch nicks/lu1:

58c9fa3d8357ea0e65322be71d23b0304a303895 (2021-11-05 11:56:47 -0400)
liveupdate: print pod name and container name instead of container id
container id just isn't that meaningful to people

### DIFF
--- a/internal/container/container.go
+++ b/internal/container/container.go
@@ -2,7 +2,6 @@ package container
 
 import (
 	"fmt"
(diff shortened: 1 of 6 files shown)


nicks:
converted a few more that were fixable

PR #5143, opened 2021-11-05
tiltfile: add cmd() built-in

milas:
Tiltfile functions that accept commands currently require 4 arguments
per command (cmd, env, dir, cmd_bat). In some cases, like
`local_resource` where there are multiple commands, that gets unwieldy
fast.

The `cmd` built-in lets us group these into a single object, which
maps to the `model.Cmd`. Anywhere that currently accepts a `cmd`
can now optionally accept this instead. However, to avoid breaking
Tiltfile compatibility, the separate arguments are still available,
but CANNOT be used in conjunction with the `cmd()` built-in for a
given command assembly.

There's a side benefit here when working with API types in a Tiltfile,
which is that this can be used as a bridge/helper to get things into
the expected format. For example, the APIs all take `Args` as an
argv-style string array, and don't support the auto-magical
string -> shell wrapper, so callers can now make a `cmd` object and
then use `cmd.args` for the field value. Ditto for env, where it's
nicer in Starlark to make a `dict`, but that then needs to be mapped
to a `KEY=value` string list, which can be accessed via `cmd.env`.

### DIFF
--- a/internal/tiltfile/value/cmd.go
+++ b/internal/tiltfile/value/cmd.go
@@ -1,0 +1,72 @@
+package value
+
+import (
+	"fmt"
+
+	"go.starlark.net/starlark"
+	"go.starlark.net/starlarkstruct"
+
+	"github.com/tilt-dev/tilt/internal/tiltfile/starkit"
+	"github.com/tilt-dev/tilt/pkg/model"
+)
+
+const typeCmd = "cmd"
+
+type CmdPlugin struct{}
+
+func NewCmdPlugin() CmdPlugin {
+	return CmdPlugin{}
+}
+
+var _ starkit.Plugin = CmdPlugin{}
+
+func (p CmdPlugin) OnStart(env *starkit.Environment) error {
+	if err := env.AddBuiltin("cmd", p.cmd); err != nil {
+		return fmt.Errorf("could not add cmd builtin: %v", err)
+	}
+	return nil
+}
+
+func (p CmdPlugin) cmd(thread *starlark.Thread, fn *starlark.Builtin, args starlark.Tuple,
+	kwargs []starlark.Tuple) (starlark.Value, error) {
+	var argsVal, argsBatVal starlark.Value
+	// TODO(milas): should we also except ['FOO=bar'] style list?
+	var env StringStringMap
+	var dir starlark.Value
+
+	err := starkit.UnpackArgs(
+		thread, fn.Name(), args, kwargs,
+		"args", &argsVal,
+		"env?", &env,
+		"dir?", &dir,
+		"args_bat?", &argsBatVal,
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	cmd, err := ValueGroupToCmdHelper(thread, argsVal, argsBatVal, dir, env)
+	if err != nil {
+		return nil, err
+	}
+
+	return Cmd{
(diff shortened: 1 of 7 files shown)


internal/tiltfile/value/cmd.go
```diff
@@ -0,0 +1,72 @@
...
+	var argsVal, argsBatVal starlark.Value
+	// TODO(milas): should we also except ['FOO=bar'] style list?
+	var env StringStringMap
+	var dir starlark.Value
+
+	err := starkit.UnpackArgs(
+		thread, fn.Name(), args, kwargs,
+		"args", &argsVal,
+		"env?", &env,
+		"dir?", &dir,
```
landism, line 41: IIUC, custom_build ignores its command's dir and env:
https://github.com/tilt-dev/tilt/blob/2b9a8539f211bd9441c54a7109b78d53e7c5cbaa/internal/build/custom_builder.go#L77-L79
https://github.com/tilt-dev/tilt/blob/333ce8741d1b6285add1625f3ed004e8290d14f8/internal/tiltfile/docker.go#L361

(really, the CustomBuild struct probably shouldn't have a WorkDir field - I assume that predates Command having other fields)

It'd be great to fix, but also maybe a bigger yak than you want in this PR, in which case maybe add an error if someone passes a `cmd()` to `custom_build`

nicks:
i have a few high-level concerns about this PR:

1) i think having a Cmd (https://api.tilt.dev/core/cmd-v1alpha1.html) and a cmd with very different APIs and semantics is going to be massively confusing.

2) i understand the argument for "bridge/helper" ergonomics. But I think it's a bad idea to write out those ergonomics by hand. There's too much risk of subtle implementation details introducing bugs. To take an example from this PR - is `dir` evaluated relative to where `cmd()` is called, or where `cmd()` is used? Your hand-written code takes the latter stance, while the auto-generated code takes the former stance.

internal/tiltfile/value/cmd.go
landism: on the plus side, yay for this change making it more likely we have consistent support for cmd customization across builtins!

nicks:
one possible way to resolve these problems is to have some namespace for helpers, similar to how kubernetes apimachinery uses metav1

so maybe something like:

```
type Exec struct {
  // +tilt:parser=ArgList
  Args []string

  // +tilt:local-path=true
  Dir string
}
```

```
type ArgList struct {
  // +tilt:parser=ShellCmd
  Args []string
  
  // +tilt:parser=BatCmd
  Bat []string
}
```

which would then auto-generate to:

```
cmd=types.exec(args=types.args(['foo'], bat=['bar']), env=['FOO=1'])
```

though i'm just batting ideas around

nicks:
(i'm also obviously interested in directions that allow us to auto-generate starlark docs)

milas:
> i think having a Cmd (https://api.tilt.dev/core/cmd-v1alpha1.html) and a cmd with very different APIs and semantics is going to be massively confusing

Fair! I'm fine with punting on this and covering it more uniformly with auto-generated API helpers in the future.

FWIW The ergonomics for generating shell-type commands for API types was actually just a side benefit/realization, but can definitely be handled in another way if it becomes a more pressing issue. I mostly just felt guilty writing a _new_ Tiltfile function with 8 command-related arguments, but I can get over that 😉

PR #5148, opened 2021-11-08
use brew bundle and new tilt formula

stephengroat:
brew bundle handles both install and upgrade
uses new tilt formula in homebrew-core https://github.com/Homebrew/homebrew-core/pull/88977

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -27,14 +27,7 @@ function copy_binary() {
 
 function brew_install_or_upgrade() {
   set -x
-  brew tap tilt-dev/tap
-  if test -d "$(brew --prefix tilt-dev/tap/tilt)"; then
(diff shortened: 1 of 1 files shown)


stephengroat:
👋  sorry, starting to have issues with too many taps these days

there's https://github.com/dawidd6/action-homebrew-bump-formula, but I see that you're using circleci. you can use https://github.com/Homebrew/brew/blob/master/Library/Homebrew/dev-cmd/bump-formula-pr.rb directly, but you'll need a github access token to be able to perform the automated actions in CI

otherwise, you can run it command line

nicks:
cool! let's try it once manually and see how it goes. our releaser job already runs on a bot account so it should be OK to use bump-formula-pr /cc @landism

PR #5152, opened 2021-11-09
api: run DeleteCmd when KubernetesApply object deleted

milas:
When a custom `DeployCmd` is provided, a corresponding `DeleteCmd`
is needed to handle cleanup.

Although Tilt is able to track + diff the managed entities to determine
what to delete (which is still the case when using YAML), the
`DeployCmd` might create additional state, e.g. a Helm release in
Tiller/`Secret`, so this provides the tool with the means to clean
up the external state _as well as all relevant entities in the
cluster_.

### DIFF
--- a/internal/controllers/core/kubernetesapply/reconciler.go
+++ b/internal/controllers/core/kubernetesapply/reconciler.go
@@ -36,6 +36,11 @@ import (
 	"github.com/tilt-dev/tilt/pkg/model"
 )
 
+type garbageCollectMeta struct {
+	entities  []k8s.K8sEntity
+	deleteCmd *v1alpha1.KubernetesApplyCmd
+}
+
 type Reconciler struct {
 	st          store.RStore
 	dkc         build.DockerKubeConnection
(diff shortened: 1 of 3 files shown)


internal/controllers/core/kubernetesapply/reconciler.go
```diff
@@ -36,6 +36,11 @@ import (
 	"github.com/tilt-dev/tilt/pkg/model"
 )
 
+type garbageCollectMeta struct {
```
nicks, line 39: what does 'meta' signify here?

nicks: maybe it should be called DeleteSpec or smth

internal/controllers/core/kubernetesapply/reconciler.go
```diff
@@ -524,46 +523,72 @@ func (r *Reconciler) updateResult(nn types.NamespacedName, result *Result) []k8s
...
 	// Go through all the results, and check to see which objects
 	// we're not managing anymore.
-	var toDeleteMap objectRefSet
-	if existing != nil {
-		toDeleteMap = existing.AppliedObjects.clone()
-		for _, result := range r.results {
-			for objRef := range result.AppliedObjects {
-				delete(toDeleteMap, objRef)
-			}
+	// TODO(milas): in the case that the KA object was deleted, should we respect `tilt.dev/down-policy`?
```
nicks, line 542: i'd vote yes, though agreed that the current code doesn't respect this

nicks (approved):

internal/controllers/core/kubernetesapply/reconciler.go
milas: 'meta' signifies I was having a hard time naming this object 😉 

(Renamed to `deleteSpec`, thanks for suggestion)

PR #5155, opened 2021-11-09
tiltfile: add delete_cmd to k8s_custom_deploy

milas:
This wires up the `Tiltfile` + `tilt down` side of things (the API
support is in #5152).

The `delete_cmd` is **required** by `k8s_custom_deploy` since otherwise
there's no way to clean up resources via `tilt down`.

The feature flag is also removed _entirely_ - it was added as a
guard pending this change, but an interim Tilt release never
actually shipped, so there's no need to obsolete it first.


integration/k8s_custom_deploy/Tiltfile
```diff
@@ -1,15 +1,15 @@
 
 include('../Tiltfile')
 
-enable_feature('k8s_custom_deploy')
-
 k8s_custom_deploy(
     'custom-deploy-int-test',
     # forcibly delete and re-create the deployment whenever run
-    cmd='kubectl delete --ignore-not-found=true --wait=true -f "${DEPLOY_FILE}" >/dev/null && kubectl create -oyaml -f "${DEPLOY_FILE}"',
+    deploy_cmd='kubectl delete --ignore-not-found=true --wait=true -f "${DEPLOY_FILE}" >/dev/null && kubectl create -oyaml -f "${DEPLOY_FILE}"',
```
nicks, line 7: i wonder if we should call this `apply` instead of `delete` to better match the `kubectl` verbs...

internal/cli/down.go
```diff
@@ -75,7 +75,31 @@ func (c *downCmd) down(ctx context.Context, downDeps DownDeps, args []string) er
 		return err
 	}
 
-	entities, err := engine.ParseYAMLFromManifests(tlr.Manifests...)
+	if err := deleteK8sEntities(ctx, tlr.Manifests, downDeps, c.deleteNamespaces); err != nil {
```
nicks, line 78: i feel bad about continuing to add more and more complicated logic to `tilt down`.

possible idea: once enable/disable services is out, change the implementation of `tilt down` to create all resources initially disabled, and wait until the reconcilers reach a steady state of disabling them. maybe this is weird though

nicks (approved):

integration/k8s_custom_deploy/Tiltfile
milas: Renamed everywhere in e4adb2efb to be `apply_cmd` in Starlark and `ApplyCmd` in API

PR #5157, opened 2021-11-09
apis: add DockerImageSpec

nicks:
Hello @milas, @landism,

Please review the following commits I made in branch nicks/dockerimage2:

612022bbd346b49cebc90cf9b33226affa5c56d0 (2021-11-09 18:53:02 -0500)
apis: add DockerImageSpec


nicks:
this is mostly a straight port of https://github.com/tilt-dev/tilt/blob/master/pkg/model/image_target.go#L258, with some light renaming to better match what's in the Docker CLI

pkg/apis/core/v1alpha1/dockerimage_types.go
```diff
@@ -51,8 +51,95 @@ type DockerImageList struct {
...
+	Target string `json:"target,omitempty" protobuf:"bytes,4,opt,name=target"`
+
+	// Pass SSH secrets to docker so it can clone private repos.
+	//
+	// https://docs.docker.com/develop/develop-images/build_enhancements/#using-ssh-to-access-private-data-in-builds
+	//
+	// Equivalent to `--ssh` in the docker CLI.
+	//
+	// +optional
+	SSHSpecs []string `json:"sshSpecs,omitempty" protobuf:"bytes,5,rep,name=sshSpecs"`
```
nicks, line 98: i've been trying to come up with a better name for this. SSHSecrets? SSHKeys? 'SSHs' seems too weird

milas: BuildKit refers to it as `AgentConfig` internally - `SSHAgentConfigs`? `SSHAgentSpecs`? FWIW I think `SSHSpecs` is also fine/consistent with K8s nomenclature.

Also, do you think this should be normalized, i.e. `ID` + `Paths` tuple? (I actually think it's fine as a `string`, only potential upside to separate fields would be ability to explicitly mark the paths as file paths for the codegen helper, but I can't think of any use case beyond debugging where you'd need to actually use these values for anything from the API, so probably not worth the extra nesting/structs)

https://github.com/moby/buildkit/blob/a87b801ca08f1f8f34afeda915138b47ca054421/session/sshforward/sshprovider/agentprovider.go#L23-L27

milas (approved):

pkg/apis/core/v1alpha1/dockerimage_types.go
nicks: ooh, i like SSHAgentConfigs. done!

TBH, one of my big frustrations with Docker is how the CLI is so different from the API, and feel like a lot of the CLI is an admission that the API needs to be simpler. i'm trying to keep it closer to the CLI when it makes sense to, since that's what lots of people are used to.

pkg/apis/core/v1alpha1/dockerimage_types.go
```diff
@@ -51,8 +51,95 @@ type DockerImageList struct {
 	Items []DockerImage `json:"items" protobuf:"bytes,2,rep,name=items"`
 }
 
-// DockerImageSpec defines the desired state of DockerImage
+// DockerImageSpec describes how to build a Docker image with `docker_build`.
+//
+// Most fields of this spec directly correspond to the Docker CLI.
 type DockerImageSpec struct {
+	// The named reference of the image.
+	Ref string `json:"ref" protobuf:"bytes,12,opt,name=ref"`
```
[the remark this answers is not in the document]
nicks: i also wonder this! also been thinking about generalizing this to something like +tilt:parser=MyParser to help with different kinds of parsers

pkg/apis/core/v1alpha1/dockerimage_types.go
```diff
@@ -51,8 +51,95 @@ type DockerImageList struct {
...
+
+	// Dockerfile contains the complete contents of the Dockerfile.
+	//
+	// TODO(nick): We should also support referencing the Dockerfile as a path.
+	//
+	// +optional
+	DockerfileContents string `json:"dockerfileContents,omitempty" protobuf:"bytes,1,opt,name=dockerfileContents"`
```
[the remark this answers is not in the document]
nicks: ya, i think this is a big open question - how image dependencies will work in this system. i really like how current tilt auto-infers dependencies. trying to punt on this now to get image status in.

PR #5167, opened 2021-11-12
api: tiltfile reconciler should gc ToggleButtons

landism:
### Problem

When a disable button is no longer needed (resource gets deleted or feature flag gets disabled), it's not removed from the api

### Solution

Add ToggleButton to the list of types that the Tiltfile reconciler should reconcile, so that it gcs them.

### Note

I'm kind of tempted to make `typesToReconcile` a field on the Reconciler that just stores the union of all types ever returned by `toAPIObjects`, but maybe it's good to be more explicit about it?


lizzthabet (approved):
lgtm! i don't have any technical approach-related opinions, but i tested locally and everything worked. ✨

nicks (approved):

PR #5175, opened 2021-11-15
apis: add schema for DockerImageStatus

nicks:
Hello @milas, @landism,

Please review the following commits I made in branch nicks/dockerimage3:

987b2dc09672d97e0a83831c243551999a5f2091 (2021-11-15 18:19:22 -0500)
apis: add schema for DockerImageStatus


pkg/apis/core/v1alpha1/dockerimage_types.go
```diff
@@ -186,6 +186,30 @@ func (in *DockerImageList) GetListMeta() *metav1.ListMeta {
...
 // DockerImageStatus defines the observed state of DockerImage
 type DockerImageStatus struct {
+	// A fully-qualified image reference of a built image, as seen from the local
+	// network.
+	//
+	// Usually includes a name and an immutable tag.
+	//
+	// NB: If we're building to a particular registry, this may
+	// have a different hostname from the Spec `Ref` field.
+	Ref string `json:"ref,omitempty" protobuf:"bytes,1,opt,name=ref"`
```
landism, line 196: should this should be `+optional`? IIUC, during the first build, we'll want to set `Building` but not `Ref`.

landism (approved):

pkg/apis/core/v1alpha1/dockerimage_types.go
milas: ^ That's interesting - should the `Ref` actually be part of `DockerImageStateCompleted`?

pkg/apis/core/v1alpha1/dockerimage_types.go
```diff
@@ -201,3 +225,68 @@ var _ resource.StatusSubResource = &DockerImageStatus{}
...
+	// +optional
+	Cached bool `json:"cached,omitempty" protobuf:"varint,2,opt,name=cached"`
+
+	// The timestamp when we started working on the stage.
+	// +optional
+	StartedAt *metav1.MicroTime `json:"startedAt,omitempty" protobuf:"bytes,6,opt,name=startedAt"`
+
+	// The timetsamp when we completed the work on the stage.
+	// +optional
+	FinishedAt *metav1.MicroTime `json:"finishedAt,omitempty" protobuf:"bytes,7,opt,name=finishedAt"`
```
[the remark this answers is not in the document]
nicks: The way I've been thinking about it is:

1) If the owner object is required, and the timestamp may not be available yet, make it optional. As in CronJobStatus timestamps
https://pkg.go.dev/k8s.io/kubernetes/pkg/apis/batch#CronJobStatus

2) If the owner object is optional, and the existence of the owner object implies the timestamp exists, make it required. As in ContainerStateTerminated timestamps:
https://pkg.go.dev/k8s.io/api/core/v1#ContainerStateTerminated

In this case, I could easily imagine us reporting intermediate buildkit stages that weren't finished yet.

pkg/apis/core/v1alpha1/dockerimage_types.go
landism: I don't think so - I think when one build has finished and the next build is running, we want Ref pointing at the last successful build and the DockerImageStates set describing the current build. i.e., there would be no DockerImageStateCompleted (or it might have an error). It might be good to add some comments to clear this up.

nicks: ya, I could go either way on this. We could also have it in both fields, similar to how ContainerStatus has some things duplicated. it doesn't really matter that much, since we expect to write the result to ImageMapStatus

PR #5178, opened 2021-11-16
web: display disabled resources in Table View when feature flag is enabled

lizzthabet:
Display disabled resources in Table View when the feature flag is enabled. If the feature flag is not enabled, disabled resources will not be visible. This PR also:
* hides trigger, trigger mode, pod ID, and endpoint column information for disabled resources
* adds custom sort logic so that disabled resource rows appear at the bottom of each table when it's unsorted or when table is sorted by status
* adds support for `Disabled` as a runtime and build status
* adds a couple disabled resources + support for feature flag in `OverviewTable` story
* includes a couple small `TODO` comments that I'll follow up on after this PR

To test, you can check out this code locally and spin up the storybook `Tiltfile`. (The feature flag is already enabled.) 

Example of Table View with several disabled resources:
Screen Shot 2021-11-16 at 2 05 03 PM

Example of Table View (without groups), disabling resources, and resource name filter:

Closes [CH12869](https://app.shortcut.com/windmill/story/12869/table-view-displays-disabled-resources)


nicks (approved):

web/src/SidebarItem.tsx
```diff
@@ -34,7 +34,7 @@ class SidebarItem {
   queued: boolean
   lastBuild: Build | null = null
   hold: Hold | null = null
-  disabled: boolean
+  disabled: boolean // TODO (lizz): Refactor sidebar logic to refer to other status fields, since this is redundant
```
landism, line 37: FWIW, I suspect we should find a way to use this class (or something like it) for the table view so that we don't keep having to implement everything twice

landism (approved):

web/src/SidebarItem.tsx
lizzthabet: i totally agree. it would be nice to have a common interface that takes a `UIResource` and returns a view-agnostic instance.

web/src/OverviewTable.tsx
```diff
@@ -825,6 +877,29 @@ function resourceTypeLabel(r: UIResource): string {
   return "Unknown"
 }
 
+function resourceListByStatus(
+  resources: UIResource[] = [],
+  disabledResourcesEnabled: boolean
```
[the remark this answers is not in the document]
lizzthabet: good suggestion!

PR #5182, opened 2021-11-16
web: adjust disable button location / visuals

landism:
1. render the disable button on the far right of the bottom row of the overview action bar, instead of with the other buttons
2. change appearance to match figma (include "Resource", remove icon)


web/src/ApiButton.tsx
```diff
@@ -410,19 +411,40 @@ export function ApiButton(props: React.PropsWithChildren<ApiButtonProps>) {
   }
 }
 
+// UIButtons for a location, sorted into types
+export type ButtonSet = {
+  normal: UIButton[]
+  toggleDisable?: UIButton
+}
```
lizzthabet, line 418: i find `normal` a little odd/confusing. what do you think about `default` or `userDefined`?

web/src/OverviewActionBar.tsx
```diff
@@ -186,6 +186,20 @@ const CustomActionButton = styled(ApiButton)`
...
+    ${OverviewButtonMixin};
+  }
+
+  .MuiButton-label {
+    // hardcode a width to workaround this bug:
+    // https://app.shortcut.com/windmill/story/12912/uibuttons-created-by-togglebuttons-have-different-sizes-when-toggled
+    width: ${SizeUnit(3.6)};
+  }
+
+  margin-left: ${SizeUnit(0.25)};
```
lizzthabet, line 200: i would recommend a larger button margin here, like maybe `0.5` or even up to `0.75` instead.

pictured below with `0.5`:
Screen Shot 2021-11-17 at 4 56 36 PM

this is more a note for @SurbhiSGupta and doesn't need to be resolved in this pr! i find the placement of that button somewhat un-intuitive within the log part of the action bar, and i'm curious if others feel the same. it feels like the button gets a little lost on the far right. i tried it on both the left side and the left side with the log options centered (the latter of which may create too many layout issues on smaller screens). here are those views for reference:

Screen Shot 2021-11-17 at 4 54 41 PM

Screen Shot 2021-11-17 at 4 53 57 PM

lizzthabet (approved):
this looks great! one minor comment on naming, and a design note.

web/src/ApiButton.tsx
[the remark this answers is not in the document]
landism: yeah, 'default' works

fwiw, I prefer avoiding terms like "userDefined" since - who's the user? those buttons might come from tilt code, or from extensions, or from a tiltfile author, or from someone who's just running tilt and has never looked at a tiltfile

PR #5183, opened 2021-11-16
dockerimage: write image build performance to the apiserver

nicks:
Hello @milas, @landism,

Please review the following commits I made in branch nicks/dockerimage4:

d918439a8b47cf74a7307b7d8389308c997dd1cc (2021-11-16 17:51:36 -0500)
dockerimage: write image build performance to the apiserver


nicks:
this was even more straightforward than i expected it to be!

i wonder if we should also add Durations to the api even if they're redundant

milas (approved):

landism (approved):

Issue #4882, opened 2021-08-24
`tilt up` without a Kubernetes cluster takes 20 seconds to start


robertmarsal:
### Expected Behavior
When doing `tilt up` and only using docker-compose resources tilt starts straight away

### Current Behavior
At the moment `tilt` is trying to reach a Kubernetes cluster and waiting for a timeout delaying the startup time

### Steps to Reproduce

1.  Use a machine where there is no Kubernetes cluster
2. Use a Tiltfile that only imports a docker-compose file
3. Run `tilt up` and notice how it takes 20 seconds to process the tiltfile and then bring up the services

nicks:
thanks for filing @robertmarsal !

we haven't seen this before. is it possible that you have a ~/.kube/config that's pointing at a dead cluster? i wonder if tilt sees you have a valid config, tries to connect to it, and times out.

(it's still a bug that tilt blocks 20 seconds trying to connect to it...but i'm curious if deleting the config fixes your problem for now).

robertmarsal:
Hi @nicks you are correct! I still had a config file pointing to a dead cluster. Removing that file fixes the problem. Thanks

hookenz:
@nicks I just had this problem when roaming with my laptop running a local microk8s cluster.
It still seems to be happening with a recent version of Tilt.

v0.33.22, built 2025-01-03

PR #5184, opened 2021-11-16
Defer registry lookup until after Tiltfile is loaded

nicksieger:
When only using docker compose as in #4882, looking up the registry is unnecessary and appears to block the UI until after the tiltfile is fully executed. 

- tiltfile: defer finding registry until needed
- vendor: make vendor updates


internal/tiltfile/tiltfile_state.go
```diff
@@ -1000,12 +1001,23 @@ func (s *tiltfileState) k8sResourceForName(name string) (*k8sResource, error) {
 // Otherwise, we'll return the zero value of `s.defaultReg`, which is an empty registry.
 // It has side-effects (a log line) and so should only be called once.
 func (s *tiltfileState) decideRegistry() container.Registry {
-	if s.orchestrator() == model.OrchestratorK8s && !s.localRegistry.Empty() {
+	if s.orchestrator() != model.OrchestratorK8s {
+		return s.defaultReg
+	}
+
+	// Defer lookup of registry until now
+	if s.localRegistry == nil {
```
nicks, line 1009: for what it's worth - i'm pretty sure LocalRegistry() is already memoized in the k8sClient itself, so no need to store it as a field on tiltfileState

nicks (approved):
LGTM!

internal/tiltfile/tiltfile_state.go
nicksieger: Yeah, I wasn't sure whether it made sense to keep that struct field. I'm going to remove and clean up.

milas (approved):
🎉 🎉 🎉 🎉

hookenz:
@nicksieger I have a local kubeconfig that is pointing to a cluster address that no longer exists (due to me roaming).
The 20s startup time has either re-appeared or is still happening.

PR #5186, opened 2021-11-17
web: add the location to button click analytics

nicks:
Hello @lizzthabet, @landism,

Please review the following commits I made in branch nicks/metrics:

551c4192c20c8c72779cf9878a979f41953470a7 (2021-11-17 13:25:49 -0500)
web: add the location to button click analytics
also, clean up some test warnings

### DIFF
--- a/web/src/ApiButton.tsx
+++ b/web/src/ApiButton.tsx
@@ -169,6 +169,9 @@ function ApiButtonWithOptions(props: ApiButtonWithOptionsProps & ButtonProps) {
   const { submit, uiButton, setInputValue, getInputValue, ...buttonProps } =
     props
 
+  let componentType = uiButton.spec?.location?.componentType
+  let tags = { component: componentType }
+
   return (
     <>
       <ApiButtonRoot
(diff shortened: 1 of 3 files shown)


web/src/ApiButton.tsx
```diff
@@ -169,6 +169,9 @@ function ApiButtonWithOptions(props: ApiButtonWithOptionsProps & ButtonProps) {
   const { submit, uiButton, setInputValue, getInputValue, ...buttonProps } =
     props
 
+  let componentType = uiButton.spec?.location?.componentType
+  let tags = { component: componentType }
```
lizzthabet, line 173: wdyt about adding the `annotations` property to the tags? i'm thinking about how we'd want to distinguish between user-defined buttons vs. tilt-defined buttons, like the enable / disable toggle. (this would introduce a dependency between this and #5182, so it'd be fine if you wanted to do that in a separate pr!)

lizzthabet (approved):
i'm excited to have button analytics ahead of enable/disable launching. 🌟

web/src/ApiButton.tsx
nicks: i think that's a good idea! but we'd have to be a little careful to make sure we don't accidentally include personally-identifiable data (like if the annotation contains something about the user's project)

lizzthabet: oh yeah! maybe we could just allowlist the known tilt annotations and ignore everything else.

PR #5187, opened 2021-11-17
vendor: update apiserver

milas:
Pull in optimistic concurrency fix from tilt-dev/tilt-apiserver#51


nicks (approved):

nicks:
also it's fine to do PRs like this w/o review since they're basically auto-generated!

PR #5193, opened 2021-11-17
test: fix flaky TestPodEventOrdering test

milas:
One of the test cases creates + deletes a pod before sending the
final real pod that it wants in the steady state, but was determining
it'd reached steady state by looking at the pod count, so might have
caught it when the count matched due to the about-to-be-deleted pod,
but the pod it really wanted hadn't been seen yet.


landism (approved):
👐

nicksieger (approved):

PR #5200, opened 2021-11-18
test: improve pod event synchronization

milas:
This is a collection of fairly random tweaks to `upper_test.go`
primarily.

The big one is to eliminate usage of `MostRecentPod()` and instead
always explicitly look for the pod by name (via a new helper that
consolidates some oft-duplicated logic and hopefully provides more
helpful failure message with the pod name, particularly for tests
that are interleaving multiple pods).

There's some additional upfront sanity checks as well, e.g. when
dispatching a pod log event, it'll make sure that the pod is known
first or fail early.

A small change was made in the `KubernetesDiscovery` reconciler to
check for pod restarts _before_ dispatching the upsert event; this
won't actually affect anything real world but ensures that things
happen in a deterministic order for the sake of the tests.

### DIFF
--- a/internal/controllers/core/kubernetesdiscovery/reconciler.go
+++ b/internal/controllers/core/kubernetesdiscovery/reconciler.go
@@ -363,8 +363,8 @@ func (w *Reconciler) updateStatus(ctx context.Context, watcherID watcherID) erro
 		return fmt.Errorf("failed to update KubernetesDiscovery status for %q: %v", watcherID, err)
 	}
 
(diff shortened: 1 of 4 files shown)


milas:
I think #5203 will largely obsolete this

nicks:
sorry! was trying to thread through the PodOwner stuff, and realized it was easier to replace the old pod ingestion code than trying to modify it to handle PodOwners correctly.

milas:
No worries! (Almost) always better to get rid of code 🙃

PR #5204, opened 2021-11-19
cli: update template for cli docs

hyu:
The CLI docs require a new line in their front matter to render properly.


nicks (approved):
good find! LGTM

PR #5209, opened 2021-11-19
contributing: add instructions for bumping brew formula

landism:
Now that we have an upstream brew formula (#5148), we have an extra step to bump its tilt version

I didn't get around to adding auth and putting this command in CI, and figured the manual step might as well be documented in the meantime.


nicks (approved):
re: "I didn't get around to adding auth" - is there some additional auth necessary? in theory, the CI releaser should already have a github api token so it can commit all the doc-update PRs

landism:
> re: "I didn't get around to adding auth" - is there some additional auth necessary?

oh, maybe not!

landism:
no longer relevant, since Nick added it to CI

PR #5211, opened 2021-11-22
kubernetesapply: improve user-facing logging

### LINKED ISSUE #5192: k8s_custom_deploy should print what resources it deployed
After k8s_custom_deploy finishes, it should print the names of resources it deployed to the cluster (similar to what k8s_yaml prints today)

### Current Behavior

Right now it prints:

```
STEP 1/1 — Deploying

     Step 1 - 2.43s (Deploying)
     DONE IN: 2.43s
```

which is weird (afaik, it's not injecting any images)

milas:
* Result of apply now consistently logged for both YAML + Cmd deploys
  * This slightly changes things for YAML, which used to show the
    objects it was _going_ to apply
* Image injection message only shown when relevant and moved to debug
  * More detailed, including the entity and the transformation that
    was applied
* Consistent/improved indentation across YAML + Cmd deploys

### `k8s_yaml` example
YAML deploy

### `k8s_custom_deploy` example
Custom deploy command

Fixes #5192.

### DIFF
--- a/internal/engine/buildcontrol/image_build_and_deployer.go
+++ b/internal/engine/buildcontrol/image_build_and_deployer.go
@@ -310,8 +310,6 @@ func (ibd *ImageBuildAndDeployer) deploy(
 	ps.StartPipelineStep(ctx, "Deploying")
 	defer ps.EndPipelineStep(ctx)
 
-	ps.StartBuildStep(ctx, "Injecting images into Kubernetes YAML")
-
 	kTargetNN := types.NamespacedName{Name: kTargetID.Name.String()}
 	// Note: `KubernetesApply` object may not exist yet when this `ForceApply` is called
 	// and may cause a race-condition-related "Not found" error here.
--- a/internal/controllers/core/kubernetesapply/reconciler.go
+++ b/internal/controllers/core/kubernetesapply/reconciler.go
@@ -343,16 +343,26 @@ func (r *Reconciler) forceApplyHelper(
 	}
 
 	var deployed []k8s.K8sEntity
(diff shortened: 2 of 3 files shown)


milas:
A couple behavioral questions:
* For YAML, does the new behavior make sense? Or should we log beforehand - I _could_ see this being more useful for failure cases, for example, so you'd have "here's what I'm going to do" and then the error
* _Should_ the image injection logs be at `DEBUG` as I made them? Image injection/rewriting is often a source of confusion, so I think it's helpful to expose this more directly (yay API objects making this really easy), so maybe it should actually be `INFO`?

internal/engine/buildcontrol/image_build_and_deployer.go
```diff
@@ -310,8 +310,6 @@ func (ibd *ImageBuildAndDeployer) deploy(
 	ps.StartPipelineStep(ctx, "Deploying")
 	defer ps.EndPipelineStep(ctx)
 
-	ps.StartBuildStep(ctx, "Injecting images into Kubernetes YAML")
```
milas, line 313: AFAICT there was no corresponding `ps.EndPipelineStep()` call for this

nicks (approved):
To speak to the high-level questions:
1) I think the main reason we print out the YAML names is so that people can understand at a glance what's in this resource, and maybe use it to organize the resources differently. So printing after rather than before is probably fine.
2) A lot of the image-injection logs were written for a different time, when it was less clear what tilt was doing, and there were no tools to inspect it, but I'm less worried about this now.

PR #5212, opened 2021-11-22
web: propagate analytics tags a bit better

nicks:
Hello @lizzthabet, @landism,

Please review the following commits I made in branch nicks/metrics4:

bac8d06c96c55dd82e16392a8d655e20844f374b (2021-11-22 16:07:21 -0500)
web: propagate analytics tags a bit better

### DIFF
--- a/web/src/ApiButton.tsx
+++ b/web/src/ApiButton.tsx
@@ -85,6 +85,7 @@ type ApiButtonInputProps = {
   status: UIInputStatus | undefined
   value: any | undefined
   setValue: (name: string, value: any) => void
+  analyticsTags: Tags
 }
 
 function ApiButtonInput(props: ApiButtonInputProps) {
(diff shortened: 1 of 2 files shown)


web/src/ApiButton.tsx
```diff
@@ -188,8 +191,8 @@ function ApiButtonWithOptions(props: ApiButtonWithOptionsProps & ButtonProps) {
           onClick={() => {
             setOpen((prevOpen) => !prevOpen)
           }}
-          analyticsName="ui.web.uiButton.inputMenu"
-          analyticsTags={tags}
+          analyticsName="ui.web.uibutton.inputMenu"
```
nicks, line 194: i also fixed this typo, to make it consistent with the other metrics

landism (approved):

lizzthabet (commented):
when testing locally, i noticed that text input debouncing doesn't seem to be working properly on button inputs. the analytics events are delayed, but fire multiple times when i type into the field without pausing. there's not an event for every keystroke, but it's more than what i'd expect.

here's an example with storybook comparing the events fired on button input fields to the resource name filter:



i'm not sure why this is happening because the shared `InstrumentedTextField` is implementing the debounce, but maybe the memo parameters are changing? 🤔

lizzthabet:
whoops! @nicks, sorry i didn't see you merged in while i was testing.

nicks:
@lizzthabet i believe that! do you want to file a bug? i didn't really touch text input fields at all when working with this PR

lizzthabet:
@nicks, sorry i forgot to reply here! i have a fix pr up now. :)

PR #5213, opened 2021-11-22
api: do not log controller-runtime messages after context cancel

milas:
This is mostly annoying in tests when trying to reproduce an issue
and running with something like `-count 100`. With the async nature
of the reconcilers, it's pretty inevitable that some things will
still be in-flight when the tests are done (and so cancel the context
to clean up resources), and a bunch of irrelevant errors that are
really just wrapped "context canceled" get logged out.

### DIFF
--- a/internal/controllers/manager.go
+++ b/internal/controllers/manager.go
@@ -64,6 +64,14 @@ func (m *TiltServerControllerManager) SetUp(ctx context.Context, _ store.RStore)
 	// we don't care about.
 	ctxLog := logger.Get(ctx)
 	logr := genericr.New(func(e genericr.Entry) {
+		if ctx.Err() != nil {
+			// when the root context is canceled, any in-flight reconcile operations
+			// will inevitably return context-cancellation-wrapping errors, which
+			// results in a bunch of unhelpful log messages to that effect, so all
+			// logs after teardown has started are ignored
+			return
+		}
+
 		if e.Error != nil {
 			// Print errors to the global log on all builds.
 			//


milas:
FYI this isn't perfect, you still can get stuff like:
```
E1122 16:27:26.765708   72779 wrap.go:54] timeout or abort while handling: method=GET URI="/apis/tilt.dev/v1alpha1/uiresources/%28Tiltfile%29" audit-ID="201a6f2b-df7d-405a-8ba3-2c86feb57f01"
E1122 16:27:26.765806   72779 timeout.go:135] post-timeout activity - time-elapsed: 11.253µs, GET "/apis/tilt.dev/v1alpha1/uiresources/(Tiltfile)" result: <nil>
```
I'm guessing these are doing something sketchy like writing directly to `os.Stderr` 😬 

We already have some other string patterns we ignore, I held off on adding these for the moment, but I think `timeout or abort while handling` and `post-timeout activity` are probably both phrases we can add to the filter.

landism (approved):
huzzah! this was really spammy!

> you still can get stuff like

fwiw, it looks like these are just writing directly to klog w/ severity 2, which doesn't seem toooo sketchy and also wouldn't go through the filter, I assume?
https://github.com/tilt-dev/tilt/blob/948bda756a8f0b8e3fc30cec78894333eab9d3df/vendor/k8s.io/apiserver/pkg/server/filters/wrap.go#L54
https://github.com/tilt-dev/tilt/blob/948bda756a8f0b8e3fc30cec78894333eab9d3df/vendor/k8s.io/klog/v2/klog.go#L1456

milas:
Interesting - I was still seeing them sporadically in the logs when testing. I guess something else funky is going on there 🤷

PR #5218, opened 2021-11-23
web: implement UIButton.RequiresConfirmation

landism:
note: I added an "X" to the righthand side of the "confirm" mode of the button since it felt weird to have "confirm" as a safety check with no way to indicate "wait no that's not what I wanted"
image


web/src/ApiButton.tsx
```diff
@@ -287,17 +296,6 @@ async function updateButtonStatus(
...
-  uiButton: UIButton,
-  inputValues: { [name: string]: any }
-) {
-  uiButton.spec?.inputs?.forEach((i) => {
-    if (i.hidden && i.name) {
-      inputValues[i.name] = i.hidden.value
-    }
-  })
-}
-
```
landism, line 300: this is unrelated to this PR - I just noticed it was made unused by a previous PR and didn't seem worth its own PR

web/src/ApiButton.tsx
```diff
@@ -368,6 +377,37 @@ export function ApiButton(props: React.PropsWithChildren<ApiButtonProps>) {
 
   const disabled = loading || uiButton.spec?.disabled
 
+  if (confirming) {
+    return (
+      <ApiButtonRoot
```
lizzthabet, line 382: the disadvantage of returning a different dom node for the confirmation state is that in privileges mouse users. if i'm interacting with the disable button using a keyboard, then that button disappears and i'm no longer focused on it. to confirm or dismiss, i have to navigate back in the tab order. if i'm using an assistive device, there's no immediate (non-visual) indication that i need to confirm the action i just tried to perform.

ideally, (i think) the enable/disable button is always rendered inside a button group, and just the button text changes when you click `disable`. the `cancel` button is added to the button group during the confirmation state. the `i think` part is mostly because i can't quickly/cleanly map this "button with a confirmation and cancel state" to any of the aria docs examples. there is an [`alertdialog` role](https://www.w3.org/TR/wai-aria-1.1/#alertdialog), which indicates to the user that there's an alert that requires their attention/interaction, so that could be useful on the button group level (but then that overrides the mui component's `group` role). 🤔 

a workaround for the code as-is might be to manually manage focus in the click handler, so entering the confirmation state focuses on the confirmation button or button group. but that may end up being hacky bc that element doesn't exist on the dom just yet.

...ok, end of mini-essay! with all that being said, i don't think this is a blocker for this pr, but it's something i think we should look into (perhaps as a side ticket in bulk actions?). do you have any thoughts? you may have tried some of this stuff already and gone with this approach because of other things i'm not considering!

web/src/ApiButton.test.tsx
```diff
@@ -335,6 +333,59 @@ describe("ApiButton", () => {
 
     expect(error).toEqual("Error submitting button click: broken!")
   })
+
+  it("when requiresConfirmation is set, a second confirmation click to submit", async () => {
+    const b = makeUIButton()
+    b.spec!.requiresConfirmation = true
```
lizzthabet, line 339: (small nit) it might be nice to have `requiresConfirmation` as a param that gets passed into this function! my not-so-secret "secret" agenda is to move the test data functions into being more configurable, so we don't have to do manual manipulation and tweaking in tests.

web/src/ApiButton.tsx
```diff
@@ -57,6 +58,14 @@ export const LogLink = styled(Link)`
...
   padding-left: ${SizeUnit(0.5)};
 `
+const ConfirmButton = styled(InstrumentedButton)`
+  && {
+    background-color: ${Color.red};
+    &:hover {
+      background-color: ${Color.red};
+    }
+  }
+`
```
lizzthabet, line 68: there are probably some color contrast a11y issues with this button. i played around a bit and the closest thing i found that i liked was making the text black (for all states) with a dark gray border and light red border on hover/focus/active.



here's the code for that:

```scss
&& {
    background-color: ${Color.red};
    border-color: ${Color.gray};
    color: ${Color.black};

    &:hover, &:active, &:focus {
      background-color: ${Color.red};
      border-color: ${Color.redLight};
      color: ${Color.black};
    }
  }

  .fillStd {
    fill: ${Color.black} !important; /* there's some style somewhere with a higher level of specificity than this one that i did not spend time trying to find... */
  }
```

@SurbhiSGupta is welcome to chime in, too! (but we can also address stuff in the design review sesh.)

lizzthabet (approved):
looks good! i think the only thing that really should be addressed is button styling.

web/src/ApiButton.test.tsx
landism: yes, thanks!

web/src/ApiButton.tsx
landism: This all sounds right and is not a thing I was considering. I agree this isn't a blocker for this pr and we can tackle it later.

web/src/ApiButton.tsx
landism: Thanks, that looks much better (and still probably deserves a pass in the design review anyway since since the X wasn't in the figma)

PR #5220, opened 2021-11-23
store: log when a resource becomes enabled or disabled

landism:
I don't love the actual text being logged, but also couldn't come up with a good alternative.
1. `Resource "foo" is disabled` - why did it become enabled? How do I reenable it?
2. `User disabled resource "foo"` - we don't actually know it was changed by user action, vs some operator writing to the API

Went with the former since if the user clicked, hopefully they'll remember, though maybe it'll still be disabled two days later and they'll have forgotten and not be sure how to reenable. Probably not worth worrying about unless someone is confused by it.

### DIFF
--- a/internal/store/uiresources/reducers.go
+++ b/internal/store/uiresources/reducers.go
@@ -17,14 +17,15 @@ func HandleUIResourceUpsertAction(state *store.EngineState, action UIResourceUps
 		oldCount := old.Status.DisableStatus.DisabledCount
 		newCount := uir.Status.DisableStatus.DisabledCount
 
-		message := ""
+		verb := ""
 		if oldCount == 0 && newCount > 0 {
-			message = fmt.Sprintf("Resource %q disabled.\n", n)
+			verb = "disabled"
 		} else if oldCount > 0 && newCount == 0 {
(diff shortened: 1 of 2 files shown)


lizzthabet (approved):
> Went with the former since if the user clicked, hopefully they'll remember, though maybe it'll still be disabled two days later and they'll have forgotten and not be sure how to reenable. Probably not worth worrying about unless someone is confused by it.

yeah, i had the same thought! i don't know if users are _really_ gonna forget about their disabled resources, but i might air on the side of providing more context rather than less. i'm sure we'll tweak this a bunch along the way.

 wdyt about a slightly more detailed message that includes instructions for flipping the switch again?

`Resource "foo" is now disabled/enabled. This resource can be enabled/disabled again through the Tilt dashboard.`

or perhaps:

`Resource "foo" is now disabled/enabled. To enable/disable it, use the Tilt dashboard.`

landism:
> wdyt about a slightly more detailed message that includes instructions for flipping the switch again?
> Resource "foo" is now disabled/enabled. This resource can be enabled/disabled again through the Tilt dashboard.

That's a good solution! Though I went with "Tilt Web UI" since I'm not sure we've ever used the term "Tilt dashboard".

PR #5226, opened 2021-11-24
engine: make Tiltfile.Spec.Args single source of truth

landism:
### Old Behavior
When we run `tilt args`, it goes through a server.go api which edits the Tiltfile API object and also dispatches a SetTiltfileArgsAction to the engine to update EngineState.UserConfigState.Args.

### New Behavior
The engine sets EngineState.UserConfigState.Args based on the main Tiltfile's Spec.Args, with no separate action needed.

This means we can just edit the Tiltfile API object without having to have a `Store`.

### DIFF
--- a/internal/store/engine_state.go
+++ b/internal/store/engine_state.go
@@ -106,10 +106,7 @@ type EngineState struct {
 
 	TelemetrySettings model.TelemetrySettings
 
-	// args initially passed to Tilt
-	// This field is only used to initially create the Tiltfile API object
-	// subsequent updates to the args will be reflected in the Tiltfile API object, but not here
-	InitialTiltArgs []string
+	UserConfigState model.UserConfigState
 
 	// The initialization sequence is unfortunate. Currently we have:
 	// 1) Dispatch an InitAction
--- a/internal/store/engine_state.go
+++ b/internal/store/engine_state.go
@@ -106,7 +106,10 @@ type EngineState struct {
 
 	TelemetrySettings model.TelemetrySettings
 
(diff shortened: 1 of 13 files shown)


internal/store/engine_state.go
```diff
@@ -106,7 +106,10 @@ type EngineState struct {
 
 	TelemetrySettings model.TelemetrySettings
 
-	UserConfigState model.UserConfigState
```
nicks, line 109: ahhhh i'm glad you're excited for this change! Unfortunately, there are some tools that still read this off the engine state (most notably `tilty`), so we can't remove it yet. will send you a reference about it privately. but i think the other parts of this change are ok

landism:
Per Nick's comment, I've reverted the `EngineState` changes.

The main effect of this PR is now that `UpsertTiltfileAction` also updates `EngineState.UserConfigState.Args`, which means we can simply edit the Tiltfile args in the API and don't also have to dispatch a separate `SetTiltfileArgsAction` (which means we can change args w/o having an instance of `Store`).

internal/store/tiltfiles/reducers.go
```diff
@@ -18,6 +18,10 @@ func HandleTiltfileUpsertAction(state *store.EngineState, action TiltfileUpsertA
 		}
 	}
 
+	if mn == model.MainTiltfileManifestName {
+		state.UserConfigState.Args = action.Tiltfile.Spec.Args
```
nicks, line 22: haha i like this solution

nicks (approved):

PR #5228, opened 2021-11-24
web: refactor single test resource generation into a single, configurable function call

lizzthabet:
I find myself doing a bunch of manual manipulation of generated resources in tests or copy-pasta'ing another test data function and changing things slightly for what I need. Rather than have multiple `<some-description>Resource()` with slight variability, this PR refactors the `oneResourceNoAlerts()` into a parametrized function that returns a configured resource. 

So something like this:
```js
// multiple generators with similar code
oneResourceNoAlerts() {...}
oneResourceNoAlertsWithLabels() {...}

// or manually manipulating the test resource
const myResource = oneResourceNoAlerts()
myResource.name = "bleep"
myResource.metadata.labels = ["bloop", "beep"]
```
becomes this with this PR:
```js
oneResource({ name: "bleep", labels: 2 })
```

I expect that tests will still need to manipulate the returned resource to some extent, since it'll be difficult to cover every scenario, but this should be a move in the right direction! I opted to keep this PR just to single resource generation and get some feedback, before tackling `nResourceView` functions as well.

(This PR does a little bit of miscellaneous cleanup of test files, too.)

Most of the diff is related to find + replace, and the meaningful changes are in `testdata.tsx`.

### DIFF
--- a/web/src/HeaderBar.stories.tsx
+++ b/web/src/HeaderBar.stories.tsx
@@ -3,7 +3,7 @@ import { MemoryRouter } from "react-router"
 import HeaderBar from "./HeaderBar"
 import {
   nResourceView,
-  oneResourceTest,
+  oneTestResource,
   tenResourceView,
   twoResourceView,
 } from "./testdata"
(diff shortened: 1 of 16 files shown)


nicks (approved):
:tada:

PR #5229, opened 2021-11-29
web: hide ApiButtons on Table View when a resource is disabled

lizzthabet:
Hide the widgets column content if a resource is disable, plus add top margin to buttons to alleviate a style pet peeve of mine. 🤓


nicksieger (approved):
Makes sense to me!

PR #5231, opened 2021-11-29
Add go_test_changes resource to Tiltfile

nicksieger:
Adds a local resource to run tests in recently changed go packages. Uses a new makefile target to run `go test` with the most recent set of changes reported by Tilt.

- Makefile: add 'testchanges' which checks Tilt for changed pkgs
- Tiltfile: remove unused code
- Tiltfile: add go_test_changed resource

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -49,7 +49,7 @@ shorttest:
 
 # Run recent changes as seen by Tilt, called from Tiltfile
 uniq = $(if $1,$(firstword $1) $(call uniq,$(filter-out $(firstword $1),$1)))
-seen_files = $(shell tilt get filewatch local:go_test_changed -o json | jq -r '.status.fileEvents[-1].seenFiles[]')
+seen_files = $(shell tilt get filewatch local:go_test_changes -o json | jq -r '.status.fileEvents[-1].seenFiles[]')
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -47,6 +47,22 @@ test: test-go test-js
 shorttest:
 	go test -mod vendor -p $(GO_PARALLEL_JOBS) -short -tags skipcontainertests,skiplargetiltfiletests -timeout 100s ./...
 
+# Run recent changes as seen by Tilt, called from Tiltfile
+uniq = $(if $1,$(firstword $1) $(call uniq,$(filter-out $(firstword $1),$1)))
```
nicks, line 51: would it make more sense to do this in a shell script?

i feel like i've gotten in trouble with stuff that slows down evaluation of the makefile -- and there doesn't seem to be a strong reason to do it in the makefile.

this is fine too though

nicks (approved):
neat!

Makefile
nicksieger: Compromise in 09ff0bb6f715bf9be53f9fbafd88e4095fe7f255 that only makes the target visible if run from inside Tilt/Tiltfile so that regular make commands are unaffected by the extra processing?

PR #5232, opened 2021-11-29
tiltfile: ignore readiness probe for local resource w/ no serve_cmd

milas:
If a local resource has no `serve_cmd`, it cannot logically have a
readiness probe, as there's no server to check for readiness.

The actual impact of this was that if another resource had a
dependency on a local resource in this state, the dependent resource
would get stuck waiting on the local resource indefinitely.

Now, the readiness probe will be ignored and a warning logged.
(I chose not to make this an error, as this came from a user report,
so there are `Tiltfile`s out there configured this way, and didn't want
to break them.)

### DIFF
--- a/internal/engine/upper_test.go
+++ b/internal/engine/upper_test.go
@@ -3677,6 +3677,40 @@ func TestDisabledResourceRemovedFromTriggerQueue(t *testing.T) {
 	})
 }
 
+func TestLocalResourceNoServeCmdDeps(t *testing.T) {
+	f := newTestFixture(t)
+	defer f.TearDown()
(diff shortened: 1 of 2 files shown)


nicks (approved):

nicks:
(tho i think a test here would be good)

nicksieger (approved):

milas:
> (tho i think a test here would be good)

Added in 212860666 - it's in `upper_test.go` because what we really care about here is that status is reported correctly so that `resource_deps` work, so didn't want to just assert on Tiltfile assembly

PR #5233, opened 2021-11-29
dockercompose: upgrade compose-go to v1.0.8

milas:
The `compose-go` library reached v1! It's also got fixes for
(hopefully) all the big regressions from the Python version,
particularly environment variable resolution.

There are a couple changes to the client, predominantly to share
more logic around loader options between the three code paths
(`compose-go`, CLI fallback, fake/tests).

Importantly, the `compose-go` logic for `Dockerfile` paths changed,
such that they are no longer made to be absolute (unless an absolute
path is used, at which point it won't be touched). They're assumed
to be relative to the `context` in the general case. This actually
only affects the assertions in the tests - Compose has flip-flopped
behavior here multiple times, so Tilt already has logic to join the
`dockerfile` path with the `context` path if the former is relative.
(In theory, this is now the final behavior, as it was restoring v1
semantics to address compatibility with WSL, but I think I said the
same thing last time they changed the path resolution semantics!)


nicks (approved):
wooooo path resolution semantics

PR #5234, opened 2021-11-30
engine: add a ready condition to all uiresources

nicks:
Hello @milas,

Please review the following commits I made in branch nicks/readycond:

06ab713339c5693cad5e15d088cfb62ebf6e36e2 (2021-11-30 12:56:03 -0500)
engine: add a ready condition to all uiresources

### DIFF
--- a/internal/hud/webview/convert.go
+++ b/internal/hud/webview/convert.go
@@ -198,7 +198,7 @@ func ToUIResourceList(state store.EngineState, disableSources map[string][]v1alp
 			continue
 		}
 
-		r := TiltfileResourceProtoView(name, ms, state.LogStore)
+		r := TiltfileResource(name, ms, state.LogStore)
 		r.Status.Order = int32(len(ret) + 1)
 		ret = append(ret, r)
 	}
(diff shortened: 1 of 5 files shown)


internal/hud/webview/convert.go
```diff
@@ -295,12 +295,55 @@ func toUIResource(mt *store.ManifestTarget, s store.EngineState, disableSources
...
+	}
+
+	if r.RuntimeStatus == v1alpha1.RuntimeStatusOK {
+		c.Status = metav1.ConditionTrue
+		return c
+	}
+
+	if r.RuntimeStatus == v1alpha1.RuntimeStatusNotApplicable && r.UpdateStatus == v1alpha1.UpdateStatusOK {
+		c.Status = metav1.ConditionTrue
+		return c
```
milas, line 326: The reasons in the failure cases are nifty - I wonder if we should have them in the success cases as well

internal/hud/webview/convert.go
```diff
@@ -327,17 +370,18 @@ func TiltfileResourceProtoView(name model.ManifestName, ms *store.ManifestState,
...
 	return tr
 }
 
 func populateResourceInfoView(mt *store.ManifestTarget, r *v1alpha1.UIResource) error {
 	r.Status.UpdateStatus = mt.UpdateStatus()
 	r.Status.RuntimeStatus = v1alpha1.RuntimeStatusNotApplicable
 
-	if mt.Manifest.PodReadinessMode() == model.PodReadinessIgnore {
-		return nil
-	}
```
milas, line 339: Do we need any extra logic in the K8s case below to handle this correctly?

(Previously it looks like `K8sResourceInfo` would be nil in this case but will now get populated, not sure of impact of reporting on the Pod status when we're ignoring its readiness will be, if any)

milas (approved):

internal/hud/webview/convert.go
nicks: ya, i think this behavior was wrong, and is leftover from when RuntimeStatus didn't properly handle PodReadinessMode. and i think we do want K8sResourceInfo to get populated with the resources we deployed!

PR #5236, opened 2021-11-30
engine: add an up-to-date condition to all uiresources

nicks:
Hello @milas,

Please review the following commits I made in branch nicks/updated:

a0899bcffa825893a86ede8903f49eb55f1f00d0 (2021-11-30 15:35:37 -0500)
engine: add an up-to-date condition to all uiresources

### DIFF
--- a/internal/cloud/io_test.go
+++ b/internal/cloud/io_test.go
@@ -81,6 +81,11 @@ func TestWriteSnapshotTo(t *testing.T) {
           "order": 1,
           "conditions": [
             {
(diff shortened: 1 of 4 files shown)


nicks:
I'm ambivalent about the name for this. i could also imagine calling it "Updated". open to better ideas though

milas (approved):
Name seems fine to me, I can't think of anything better - thought about `Current`, but it's rather cryptic. `LatestUpdateDone`? 🤷

PR #5237, opened 2021-11-30
web: add additional analytics to toggle `ApiButton` interactions

lizzthabet:
To prep for the disable resources launch:
* Add `toggleValue` property to toggle button analytics to indicate what the value for toggle is when the user clicks on it
* Add toggle button analytics tags to the confirm/cancel buttons
* Reorganize and add more types and constants for `ApiButton`

The `toggleValue` property will help us know if a user is clicking `disable` or `enable` on a resource. I could see this value being potentially confusing because it doesn't represent the value that will result from the user interaction. But because a user could confirm or cancel their toggle, I think it's fine to record the value "pre" toggle. 

Here's what a disable resource button click looks like:
```json
{
  "action": "click",
  "buttonType": "DisableToggle",
  "component": "Resource",
  "specHash": "9ac7c2a31917ba7ca7ca",
  "toggleValue": "on"
}
```
and a confirmation button click:
```json
{
  "action": "click",
  "confirm": "true",
  "buttonType": "DisableToggle",
  "component": "Resource",
  "specHash": "9ac7c2a31917ba7ca7ca",
  "toggleValue": "on"
}
```
A cancel button click will have `confirm: "false"`, but otherwise look the same as above.

Will link a related pr for the schema documentation shortly!

Closes [CH13015](https://app.shortcut.com/windmill/story/13015/non-launch-blocker-analytics-test).

### DIFF
--- a/web/src/ApiButton.tsx
+++ b/web/src/ApiButton.tsx
@@ -8,7 +8,7 @@ import {
 import ArrowDropDownIcon from "@material-ui/icons/ArrowDropDown"
 import moment from "moment"
 import { useSnackbar } from "notistack"
-import React, { useRef, useState } from "react"
+import React, { useMemo, useRef, useState } from "react"
 import { convertFromNode, convertFromString } from "react-from-dom"
 import { Link } from "react-router-dom"
 import styled from "styled-components"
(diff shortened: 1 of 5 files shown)


web/src/ApiButton.tsx
```diff
@@ -307,6 +334,41 @@ async function updateButtonStatus(
...
+  let toggleInput: UIInputSpec | undefined
+  if (button.spec?.inputs) {
+    toggleInput = button.spec.inputs.find(
+      (input) => input.name === UIBUTTON_TOGGLE_INPUT_NAME
+    )
+  }
+
+  let toggleValue: ApiButtonToggleState | undefined
+  if (toggleInput !== undefined) {
+    toggleValue = toggleInput.hidden?.value as ApiButtonToggleState
```
nicks, line 361: per documentation here:
https://docs.tilt.dev/telemetry_faq.html#what-kinds-of-data-does-tilt-record-if-i-opt-in

i think we want to avoid sending personalized user data in anonymized tags. instead, can you whitelist the values that you want to send? like "on" / "off"?

lizzthabet: ah, yes! i didn't realize there could be other values there.

web/src/ApiButton.tsx
```diff
@@ -307,6 +333,49 @@ async function updateButtonStatus(
...
+    if (
+      definedValue === ApiButtonToggleState.On ||
+      definedValue === ApiButtonToggleState.Off
+    ) {
+      toggleValue = definedValue
+    }
+  }
+
+  return {
+    buttonType,
```
nicks, line 372: would it be better to leave off the tag completely if it's not defined, like the original code did? otherwise, i think you might end up with a bunch of {tag: "undefined"} in analytics?

nicks (approved):

web/src/ApiButton.tsx
lizzthabet: that's a great question! i believe that serializing to json drops properties that are `undefined` (`JSON.parse(JSON.stringify(tags))` does), but i'm not 100% sure of that. it's probably faster for me to make the code change than test my assumption out.

if anyone else knows, lmk for my curiosity. 🕵🏻‍♂️

PR #5238, opened 2021-11-30
cli: add 'tilt wait' to wait on resources to become ready

nicks:
Hello @milas,

Please review the following commits I made in branch nicks/wait:

3393803952243553ffcd4544c0350ea38bf2978b (2021-11-30 16:04:29 -0500)
cli: add 'tilt wait' to wait on resources to become ready

### DIFF
--- a/internal/cli/wait.go
+++ b/internal/cli/wait.go
@@ -52,8 +52,9 @@ var (
 		# Wait for the tiltfile to load.
 		tilt wait --for=condition=Ready "uiresource/(Tiltfile)"
 
-		# Wait for a deployment to become ready.
-		tilt wait --for=condition=Ready "uiresource/my-deployment"`))
+		# When used with a Kubernetes resource, waits for the pod
+    # to deploy, start running, and pass all readiness probes.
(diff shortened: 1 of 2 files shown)


internal/cli/wait.go
```diff
@@ -0,0 +1,109 @@
...
+
+		A successful message will be printed to stdout indicating when the specified
+    condition has been met. You can use -o option to change to output destination.`))
+
+	waitExample = templates.Examples(i18n.T(`
+		# Wait for the tiltfile to load.
+		tilt wait --for=condition=Ready "uiresource/(Tiltfile)"
+
+		# Wait for a deployment to become ready.
+		tilt wait --for=condition=Ready "uiresource/my-deployment"`))
```
milas, line 56: Should this be resource instead of deployment?

Might be useful to add an explainer for K8s, e.g. "for Kubernetes, this will wait until the Pod is deployed and passing health checks"

(but also this is fine as-is tbh)

milas (approved):
Gonna be super useful for scripting 🙂

internal/cli/wait.go
nicks: done, made this a bit clearer

Issue #4288, opened 2021-03-08
Tilt daemon?


markgarrigan:
## Describe the Feature You Want

`tilt up` run as a background process
`tilt stop` would stop the process. Same as ctrl+c during a tilt up process.

Maybe this already exists?? My apologies for not searching hard enough.
### Current Behavior

tilt up takes my console

### Why Do You Want This?

As a developer, I might want to `tilt up` on one microservice. Then as I'm working I realize that I want another microservice tilting. I have to open another console and run the `tilt args` command and I have to remember the `tilt up` command so that `tilt args` doesn't stop tilting things that are already tilting. Then if I wanted to do this again. I'd go to the console where I ran `tilt args` and up arrow then edit the last command.

I'm working on a little CLI where a developer can interactively choose the microservices they want to tilt. Then it runs the `tilt up` command with those microservices. If tilt was running in the background then the developer could just be in the same console run a command from this CLI and choose to add more microservices and I'd run the `tilt args` command for them.

Hope that makes sense.

**Additional context**

landism:
This feature doesn't already exist. Thanks for filing!

The title and first two sections all sound like #4278.

The "why do you want this" section makes me wonder if your concern is more about how awkward it is to edit the set of running enabled services, along the lines of #4039, and maybe the daemonization is less important?

TimothyLoyer:
> What if... thinking out loud.... there was a way to start tilt with NO resources started.

I've been working on a setup script, and I was attempting to make this exact thing happen. `tilt args` has the `--clear` flag which leaves your Tilt session in a similar state, but as far as I could tell there was no way to _start_ Tilt with no resources. If there is a way, it would be really useful. I ended up started it with a single local resource I needed to generate certificates.

nicks:
some related things we've shipped recently:

cli commands `tilt enable` and `tilt disable`: 
https://docs.tilt.dev/cli/tilt_enable.html
https://docs.tilt.dev/cli/tilt_disable.html

the tiltfile api `config.clear_enabled_resources()`, which lets you specify that `tilt up` should start with nothing enabled
https://docs.tilt.dev/api.html#modules.config.clear_enabled_resources

i still would like to make some improvements here to the cli where you can see all the running tilt instances (sort of like `kubectl config get-contexts`), and "down" an instance that's running in the background

springroll12:
[...]

Issue #4278, opened 2021-03-05
Automatically join multiple projects together in the UI


kevin-lindsay-1:
## Describe the Feature You Want

Right now, `tilt up` has the behavior of accepting a `port` argument, which lets you have multiple instances of tilt running at the same time.

However, it looks like recent updates of tilt have become slightly greedier with ports. Previously I could run `tilt up --port 10350` and `tilt up --port 10351` and they would run simultaneously without issue.

Now what appears to happen is that `tilt up --port 10350` will use up port `10350`, but also a few additional as necessary, making me have to kind of guess which port to select.

I know that I could simply figure out the port behavior and call it good, but I'd prefer that if running `tilt up` on multiple folders talked to the current "main" session, and merged the two sessions together in the UI. That way I don't need multiple tabs for each of these sessions, when really I'm trying to use all of them at once, and being split out into multiple tabs is really not the ideal experience, rather it's just kind of the default expectation when running local services.

Not saying it would be easy, but if I could just `tilt up` everything I want and have each individual session communicate with a "main" session (and be broken down in the UI accordingly), that would be a really great UX, especially for those of us that spin up many projects at once, condensing everything down into one tab.

It would also be fairly nice to be able to `tilt down` a session/list of resources via a button in the UI once this is exposed, because the session itself becomes an entity in the UI. The only question of this bit is if the first `tilt up` is the "main" session, what happens if you run `tilt down` on it?

### Why Do You Want This?

- No fussing with ports
- `tilt` effectively becomes an "always there" service (at least, in the user's head); a single tab that is almost always open, rather than feeling like one of many local servers.
- No rummaging through folders in the CLI to spin up commonly-used projects

### Doors/cans of worms this could open
- `tilt up/down` previous sessions/projects in the UI via history/Least Recently Used
- tilt automatically starts as a service, runs local scripts, spins up projects at boot
- Project-level dependencies?

landism:
Making tilt more daemon-y and giving the user more control over adding and removing things is a thing we've discussed and are interested in but haven't yet prioritized.

> Now what appears to happen is that tilt up --port 10350 will use up port 10350, but also a few additional as necessary, making me have to kind of guess which port to select.

This sounds really annoying and I think is worth sorting out sooner rather than later as a bug/separate issue.

Also, even without the recent behavior change you describe, fussing with even just the one port is annoying and we should fix that. Linking #3434 as related issue.

nicks:
[...]

PR #5239, opened 2021-11-30
cli: make 'up' and server commands choose a free port

nicksieger:
For those Tilt users that run multiple Tilt processes on their machines (and leave them running), this PR obviates the need to manage the `--port` on which to run. `tilt up` will find and advertise the first available free port starting with 10350.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -55,9 +55,11 @@ else
 	gotestsum --format standard-quiet --junitfile test-results/unit-tests.xml --rerun-fails=2 --rerun-fails-max-failures=10 --packages="./..." -- -mod vendor -count 1 -p $(GO_PARALLEL_JOBS) -short -tags skipcontainertests,skiplargetiltfiletests -timeout 100s
 endif
 
+TEST_RUN_ARG = $(if $(TEST_RUN),-run $(TEST_RUN),)
+
 integration:
 ifneq ($(CIRCLECI),true)
-		go test -mod vendor -v -count 1 -p $(GO_PARALLEL_JOBS) -tags 'integration' -timeout 700s ./integration
+		go test -mod vendor -v -count 1 -p $(GO_PARALLEL_JOBS) $(TEST_RUN_ARG) -tags 'integration' -timeout 700s ./integration
 else
 		mkdir -p test-results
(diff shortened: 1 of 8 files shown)


nicksieger:
Thoughts on this @nicks, @milas, @landism? Selfishly looked into what this would take because I find myself running multiple Tilts often. Don't know if there was some historical reason to not pick an open port.

nicks:
eh, i think most tools i work with on a day-to-day basis use a static port rather than a dynamic port.

i feel like every time i've written a dev server that used a dynamic port, people ended up complaining to me all the time about it. i know it sounds nice like it'd be easier, but what happens in practice is:

1) Alice runs Tilt. It starts on port X.
2) Alice forgets she has Tilt running, and starts a second instance. It starts on port X+1
3) Alice has her browser open to port X and is SUPER confused about what's happening, spends hours debugging, etc, etc

Because it's semantically weird for the behavior of Tilt to be a silent function of how many other Tilt instances you have running.

I like the pattern i've seen on a lot of nodejs devtools recently (e.g., Storybook) to notice that the port is taken and ask you to pick another port, like so:

Screenshot from 2021-12-01 12-15-51

since that helps disambiguate between the user story of "i intended to run two instances" vs "i forgot i had an instance running"

nicks:
(the comment above is all re: semantics of using sequential dynamic ports rather than a totally random port)

nicksieger:
That Storybook case is an interesting pattern and might be do-able. After thinking about it, the problem of auto-choosing a port is probably more of a fringe case of a power user that is intentionally running multiple Tilts.

Also somewhere in here is the ergonomics of something like #4278 or #4288 where maybe you always only run one Tilt per machine but you have the ability to dynamically add project(s)/Tiltfiles to it.

I'll leave this for a bit and see if there is more feedback, and consider the prompt-for-open-port option.

nicks:
re: "the ability to dynamically add project(s)/Tiltfiles" - in theory we support multiple Tiltfiles in a single tilt instance right now. you can `tilt apply` new Tiltfile objects to the Tilt server. https://api.tilt.dev/core/tiltfile-v1alpha1.html . But it's not really heavily advertized/supported in the CLI, and there are probably some rough edges if you tried to do it in practice.

nicksieger:
Yeah, and I don't think adding multiple Tiltfiles would be very useful unless/until 1) the UI handles that situation and groups resources accordingly, 2) you could specify different k8s clusters to connect to per-Tiltfile.

nicks:
yesssss :innocent:

PR #5240, opened 2021-12-01
docker: use an fssync server for uploading files

nicks:
Hello @landism,

Please review the following commits I made in branch nicks/fsprovider:

b4dabc5a12171a98fc20bb388f658d2bd3c732e7 (2021-12-01 13:08:38 -0500)
docker: use an fssync server for uploading files
This makes builds orders of magnitude faster when you're
working with a remote docker daemon, because the fssync
server can diff what it has against what buildkit already has.


landism (approved):
neat!

PR #5241, opened 2021-12-01
cli: `tilt args` allows editing args in $EDITOR

landism:
Previously:
* `tilt args` w/ no args would just generate an error telling the user to use `--clear` if they wanted to clear the args
* `tilt args` operated via a gorilla-routed api

Now:
* `tilt args` w/ no args opens the current args in $EDITOR and applies the result to the Tiltfile (using a chunk of the code that powers `kubectl edit`)
* `tilt args` operates via the Tilt API server

Notes:
* I haven't implemented the "on error, re-open the file and iterate" behavior that `kubectl edit` has - that seemed like more of a nice-to-have and better as a separate PR.
* Suggestions on wording (the initial comment/instructions placed in the editor, error messages, help) welcome!

### DIFF
--- a/internal/cli/args.go
+++ b/internal/cli/args.go
@@ -1,27 +1,27 @@
 package cli
 
 import (
-	"bytes"
 	"context"
-	"encoding/json"
 	"fmt"
-	"io"
-	"net/http"
-	"os"
+	"strings"
 
+	"github.com/kballard/go-shellquote"
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/kubectl/pkg/cmd/util/editor"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 
+	"github.com/tilt-dev/tilt/pkg/apis/core/v1alpha1"
 	"github.com/tilt-dev/tilt/pkg/model"
 )
 
 type argsCmd struct {
 	clear bool
-	post  httpPoster
 }
 
 func newArgsCmd() *argsCmd {
-	return &argsCmd{post: http.Post}
+	return &argsCmd{}
 }
 
 func (c *argsCmd) name() model.TiltSubcommand { return "args" }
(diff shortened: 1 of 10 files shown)


internal/cli/args.go
```diff
@@ -48,44 +54,86 @@ tilt args -- --foo=bar frontend backend
...
+	ctrlclient, err := client.New(cfg, client.Options{Scheme: v1alpha1.NewScheme()})
+	if err != nil {
+		return nil, err
+	}
+
+	return ctrlclient, err
+}
+
+func parseEditResult(b []byte) ([]string, error) {
+	lines := strings.Split(string(b), "\n")
```
milas, line 77: Might be worth swapping to a `bufio.Scanner` here - I'm guessing Windows editors will be `\r\n` line-terminated?

internal/cli/args.go
```diff
@@ -48,44 +54,86 @@ tilt args -- --foo=bar frontend backend
...
+
+	if c.clear {
+		if len(args) != 0 {
+			return errors.New("--clear cannot be specified with other values")
 		}
-		return fmt.Errorf("http request to Tilt failed: %s", res.Status)
+		args = nil
+	} else if len(args) == 0 {
+		input := fmt.Sprintf("# edit args for the running Tilt here\n%s\n", shellquote.Join(tf.Spec.Args...))
+		e := editor.NewDefaultEditor([]string{"EDITOR"})
```
milas, line 120: Probably unnecessary, but I know some apps support an app-specific variant, e.g. `GIT_EDITOR`, so we could have a `TILT_EDITOR`?

(Also, the word "editor" now looks fake to me after reading it so many times.)

internal/cli/args_test.go
```diff
@@ -1,72 +1,141 @@
...
-func TestArgsEmptyNewValueNoClear(t *testing.T) {
-	f := newArgsFixture()
-	err := f.cmd.run(context.Background(), nil)
-	require.Error(t, err)
-	require.Contains(t, err.Error(), "no args specified.")
-	require.Contains(t, err.Error(), "run `tilt args --clear`")
-}
+func TestArgsEdit(t *testing.T) {
+	if runtime.GOOS == "windows" {
+		t.Skip("haven't figured out an appropriate $EDITOR for windows")
```
milas, line 67: It's always Windows shell scripting 😬

milas (approved):
I think a lot of folks are gonna be very excited about this!

internal/cli/args.go
landism: oy, I guess I should figure out a windows test

PR #5242, opened 2021-12-02
dockercompose: fix spurious full rebuild instead of Live Update

### LINKED ISSUE #5196: Live Sync Sporadically Does Full Rebuild
### Expected Behavior

Several `touch`es or edits to a `live_update` `sync`'d file should only use existing container and not trigger a build or recreate.

### Current Behavior

Sporadically, and seemingly non-deterministically, Tilt will decide to do a full build/recreate of a container even though we expect it to simply be doing a `live_update`+`sync`.  This is true with a very aggressive `.dockerignore`, too. These even happens for trivial changes like simply saving the same file repeatedly. It'll be fine the first few saves (i.e. get live sync'd), but then sporadically when you go to save, Tilt decides to recreate the whole container. This is even while all working on the same file without switching between files.

### Steps to Reproduce

```
$ git clone https://github.com/rwoll/tilt-live-sync-triggers-full-rebuild-repro.git
$ cd tilt-live-sync-triggers-full-rebuild-repro
$ tilt up
```

Once Tilt is up and the `ui` resource is green, start editing `src/App.tsx` a bunch (saving every couple of edits). You can also simply try running: 

```
$ for i in `seq 1 100`; do echo $i; touch src/App.tsx; sleep 10; done
```

In the Tilt UI, most of the time  you'll see it live sync but ever couple of saves it will do a full re-rcreate:

repro-example-tilt

The above happened when I simply did a touch on `src/App.tsx` 3 times in a row waiting a bit in between. The first two times it did what you'd expect, but the 3rd time it decided to do a full build+recreate.

### Context

Some things to note:

1. The linked repro was created via Create React App:

    ```
    $ npx create-react-app my-app --template typescript --use-npm
    ```
2. We're using Docker Compose, not k8s.
3. We're using a specifically named container in our DC config
4. This was repo'able on 3 different macOS machines (2 of which are M1s)
5. `tilt describe filewatches docker-compose:ui` didn't output anything unexpected. It just lists the `src/App.tsx` changes.

#### `tilt doctor` Output

```
$ tilt doctor
Tilt: v0.23.0, built 2021-11-12
System: darwin-amd64
---
Docker
- Host: [default]
- Server Version: 20.10.8
- API Version: 1.41
- Builder: 2
---
Kubernetes
- Env: docker-for-desktop
- Context: docker-desktop
- Namespace: default
- Container Runtime: docker
- Version: v1.21.5
---
```

#### About Your Use Case

We can't use Tilt with our UI dev for now so it has a more involved setup than the rest of our Tilt workflow (which is basically just `tilt up` since Tilt is generally so awesome! <3 ) We gave UI dev in Tilt a shot, but the sporadic slower updates when we hit this case made us decide to switch back to running the UI in a different way.

Overall, we're loving Tilt so we'd appreciate any help debugging this! Thanks so much for your work! 😄

milas:
Occasionally, when using Live Update with a Docker Compose project,
a file change would trigger a full rebuild instead of Live Update
despite matching a sync path.

The root cause here is that the Docker Compose target watched files
in addition to the image target, so there was a race: if the Docker
Compose target saw the file but the image target hadn't yet, the
Docker Compose builder would take over and do a full rebuild. This
has probably been exacerbated by the API transition as there's now
greater potential for timing-related issues like this to manifest.

The reason the Docker Compose target watches for files is because
in some cases, there IS no image target for a Docker Compose service.
If there's no `docker_build` or `custom_build` configured, and the
Docker Compose service has a `build:` section in its YAML, the Docker
Compose build and deployer watched for file changes and then triggered
a build via passing the `--build` flag to the `docker-compose up`
call.

To address this, the Docker Compose target now behaves more like a
Kubernetes target. It does not watch for any files. Instead, image
targets are created for both services that will be managed by Tilt
(via `docker_build` or `custom_build`) as well as those that will
be built by Docker Compose. (NOTE: If a Docker Compose service has
no `build:` section and there's no corresponding `docker_build` or
`custom_build` call for it, there will be no target, as no image
builds will occur. This is commonly seen for infra deps that are
pulled from a registry.)

The image targets for Docker Compose managed builds have a new
`BuildDetails` type of `DockerComposeBuild` to go along with the
pre-existing `DockerBuild` and `CustomBuild` types. This ensures
that all file changes are tied to the image, so it will pass through
the `LiveUpdateBuildAndDeployer` first, and fallback to the
`DockerComposeBuildAndDeployer` if not Live Update-eligible. (This
is always the case for `DockerComposeBuild` targets currently!)
The `DockerComposeBuildAndDeployer` looks at the image targets and
either builds the set of Tilt-managed images for the service OR
delegates to `docker-compose up` by including the `--build` flag.

The `DockerComposeBuild` targets are created during Docker Compose
assembly during Tiltfile loading if no Tilt-managed builder exists.
This is kinda icky - it'd probably be nicer to create these as
stubs when `docker_compose()` is called and allow `docker_build`
and `custom_build` to override them, but that's a bigger yak shave
than practical for this fix.

Fixes #5196.


milas:
Apologies for the number of changes here - there's unfortunately lots of tiny ripples [especially in tests], and I couldn't figure out a great way to chunk it because it's a pretty atomic cut-over.

Also, there is currently a regression, which is that local/bind mount volumes aren't ignored for rebuilds - I need to re-introduce `LocalIgnoredDirectories` and use that for `DockerComposeBuild` targets.

nicksieger (approved):
Nice cleanup! Explanation makes sense to me, code and functionality are in better shape than before even with the caveats.

nicks:
high-level question: how do you think we should represent this in the api server?

i was originally assuming that the DockerComposeUp object would know which images come from a DockerImage (handled by tilt) and which images are handled by the DockerComposeUp object (handled by docker-compose). But this PR introduces a second possibility - that we have some sort of DockerComposeImage object in the system. But I'm not sure. what do you think?

milas:
Good question! FWIW my decision for approach of modeling this as close to the image builds for K8s was based on hopefully making API transitions easier.

I actually see a few possibilities:
 * Create a `DockerComposeImage` API type as mentioned
   * In practice, this would really just be shelling out to `docker-compose build ...` which sounds suspiciously like a custom build, so see ⬇️ 
 * Auto create `CustomBuild` objects that shell out to `docker-compose build ...` for services w/o explicit build directives in Tiltfile
   * This could be done now and would Just Work ™️ - we'd never use the `--build` flag for `up` calls, knowing that we've always pre-built everything
 * Eliminate Docker Compose managed builds entirely - create full `DockerBuild` image targets from DC services
   * This would probably be a "breaking" change just because of likelihood of subtle differences in some potential esoteric cases
   *  Users could always define a `custom_build` that calls `docker-compose build <service>` if they want DC to manage the build

The common thread here is that I believe it makes sense to split the build from the deploy long-term, so that we can have consistent semantics across K8s + DC.

internal/engine/buildcontrol/docker_compose_build_and_deployer.go
```diff
@@ -79,23 +111,23 @@ func (bd *DockerComposeBuildAndDeployer) BuildAndDeploy(ctx context.Context, st
...
 	hasReusedStep := len(reused) > 0
 	if hasReusedStep {
 		numStages++
 	}
 
-	haveImage := len(iTargets) > 0
-
 	ps := build.NewPipelineState(ctx, numStages, bd.clock)
-	defer func() { ps.End(ctx, err) }()
+	defer ps.End(ctx, err)
```
nicks, line 130: this change isn't right -- i think you're confused about defer evaluation order

before this change, err gets evaluated at the end of the function.

after this change, err gets evaluated immediately.

nicks (approved):

internal/engine/buildcontrol/docker_compose_build_and_deployer.go
nicks: here's a good explanation of it: https://medium.com/@manandharsabbir/go-lang-defer-statement-arguments-evaluated-at-defer-execution-b2c4a1687c6c

milas: Whoops 🤦 good catch - fixed in 4d76c6e2e

PR #5245, opened 2021-12-02
fix: lookup hosts before passing to NewOnAddresses

### LINKED ISSUE #5043: binding port-forwards to a non-localhost DNS fails
Start `tilt` like

```
tilt up --host dev.localhost --port=10450
```

if you have port forwards, you'll get an error:

```
Reconnecting... Error port-forwarding api (5734 -> 5000): error forwarding port: dev.localhost is not a valid IP
```

The error is coming from here:

https://kubernetes.io/docs/reference/generated/kubectl/kubectl-commands#port-forward

> Only accepts IP addresses or localhost as a value. When localhost is supplied, kubectl will try to bind on both 127.0.0.1 and ::1 and will fail if neither of these addresses are available to bind.

I don't totally understand why this restriction exists, or even if this is a bug. It seems like it would be trivial to resolve the DNS to an IP, so I am probably missing something. Needs further investigation.

nicksieger:
After trying multiple ways of looking up and caching the provided value for `--host`/`providedWebHost`, this turned out to be the simplest option to just look up the host before creating the portforward, so it will handle both the defaulted web host and specified hosts. It has the added benefit of creating portforwards on IPV6 addresses if any are resolved for the given host.

- Fixes #5043.
- Also fixes port-forwarding when passing an IPv6 literal like `tilt up --host='[::1]'`
- Only sets up port-forwards on addresses that can be listened, and returns an error if none of the resolved addresses can.
- Clarify error message between port collision and other errors
```
$ tilt up --port 10350
Tilt started on http://localhost:10350/
v0.23.1-dev, built 2021-12-02
Error: Tilt cannot start because you already have another process on port 10350
If you want to run multiple Tilt instances simultaneously,
use the --port flag or TILT_PORT env variable to set a custom port
Original error: listen tcp 127.0.0.1:10350: bind: address already in use

$ tilt up --host '[::1' --port 10351
Tilt started (without browser UI)
v0.23.1-dev, built 2021-12-02
Error: listen tcp: address [::1:10351: missing ']' in address
```

### DIFF
--- a/internal/hud/server/apiserver.go
+++ b/internal/hud/server/apiserver.go
@@ -92,10 +92,13 @@ func ProvideConfigAccess(dir *dirs.TiltDevDir) clientcmd.ConfigAccess {
 func ProvideWebListener(host model.WebHost, port model.WebPort) (WebListener, error) {
 	webListener, err := net.Listen("tcp", fmt.Sprintf("%s:%d", string(host), int(port)))
 	if err != nil {
-		return nil, fmt.Errorf("Tilt cannot start because you already have another process on port %d\n"+
-			"If you want to run multiple Tilt instances simultaneously,\n"+
-			"use the --port flag or TILT_PORT env variable to set a custom port\nOriginal error: %v",
-			port, err)
+		if strings.HasSuffix(err.Error(), "address already in use") {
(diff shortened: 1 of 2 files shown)


nicks (approved):
hehe

PR #5246, opened 2021-12-02
web: fix multiple analytics calls on instrumented text field input

lizzthabet:
[Originally noted here](https://github.com/tilt-dev/tilt/pull/5212#pullrequestreview-813993395). The analytics events for typing into an `ApiButton`'s text input field are delayed, but they fire multiple times. There's not an event for every keystroke, but it's way more than what I'd expect with a five second debounce delay.

This PR removes `useMemo`'s dependencies to make sure debounce analytics calls fires only once. (The dependencies aren't really needed, since the debounce function has access to the current values for `analyticsName` and `analyticsTags`.)

### DIFF
--- a/web/src/instrumentedComponents.tsx
+++ b/web/src/instrumentedComponents.tsx
@@ -46,7 +46,7 @@ export function InstrumentedButton(props: ButtonProps & InstrumentationProps) {
 // event per this duration. These don't need to be submitted super
 // urgently, and we want to be closer to sending one per user intent than
 // one per keystroke.
-const textFieldEditDebounceMilliseconds = 5000
+export const textFieldEditDebounceMilliseconds = 5000
 
 export function InstrumentedTextField(
   props: TextFieldProps & InstrumentationProps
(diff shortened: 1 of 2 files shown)


web/src/instrumentedComponents.tsx
```diff
@@ -63,7 +63,7 @@ export function InstrumentedTextField(
           ...(analyticsTags ?? {}),
         })
       }, textFieldEditDebounceMilliseconds),
-    [analyticsName, analyticsTags]
+    []
```
nicks, line 66: hmmm...i think this will cause other problems? now multiple fields will share a single debouncer, which is not what you want.

nicks: let's make the debouncer a global, and pass in all the arguments (rather than using the lambda to capture arguments from the outer scope)

that should make the state much easier to reason about, rather than having to think through how the debouncer state interacts with react state management

lizzthabet: the debouncing function is local to every instance of the `InstrumentedTextField` component, so we shouldn't hit any issues where intervals are reset by other fields' interactions, but i can refactor to make this code a little cleaner.

nicks (approved):
ah OK - i misunderstood something. still would be good to have a test, since there's already a test harness for these incr()s

lizzthabet:
@nicks, good point! thanks for mentioning that. i added a test + verified it fails when the debounce function is shared.

PR #5247, opened 2021-12-02
liveupdate: add container name selector support to reconciler

milas:
Allow selecting containers by their name instead of the image.

Nice and easy 😉


nicks (approved):
:tada:

PR #5248, opened 2021-12-03
k8s_custom_deploy: add container name selector support

milas:
In some instances, when using a custom deploy tool, you have no
underlying knowledge of the image name it'll produce, but there
is a predictable container name it'll use for the deployment, so
a `container_selector` argument allows providing the name of a
container in lieu of the image, which will be used to determine
which containers are targeted for Live Update operations.

### DIFF
--- a/pkg/model/image_target.go
+++ b/pkg/model/image_target.go
@@ -110,7 +110,7 @@ func (i ImageTarget) Validate() error {
 			return fmt.Errorf("[Validate] Image %q missing build path", confRef)
 		}
 	case CustomBuild:
-		if bd.Command.Empty() {
+		if !i.IsLiveUpdateOnly && bd.Command.Empty() {
 			return fmt.Errorf(
 				"[Validate] CustomBuild command must not be empty",
 			)
--- a/internal/tiltfile/k8s_custom_deploy.go
+++ b/internal/tiltfile/k8s_custom_deploy.go
@@ -3,6 +3,7 @@ package tiltfile
 import (
 	"fmt"
 
+	"github.com/docker/distribution/reference"
 	"github.com/pkg/errors"
 	"go.starlark.net/starlark"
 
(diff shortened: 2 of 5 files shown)


pkg/model/image_target.go
```diff
@@ -3,14 +3,32 @@ package model
...
+
+	// ContainerNameSelector means Live Update should select containers based
+	// on their name rather than their image.
+	//
+	// Because the spec is updated with inferred selectors from the ImageTarget
+	// in Manifest::InferLiveUpdateSelectors(), this needs to be stored
+	// independently for use there without relying on circular reading of the
+	// spec that is being updated.
+	ContainerNameSelector string
+}
```
milas, line 30: This felt icky - let me know if you have suggestions here.

Originally I had the `k8s_custom_deploy` partially populate the `.Spec.Selector.Kubernetes` struct and then `Manifest::InferLiveUpdateSelectors` would not unconditionally overwrite it, but that felt worse because it was creating a weird almost-circular situation 😕

nicks:
Yeah, this seems to take us in the opposite direction of where we want to go (i.e., it creates new data models that aren't part of the API server). Fixing it will essentially be a revert of this PR.

I expected that would would happen is that:
- the tiltfile function would populate the selector
- the InferSelector() function would skip inference if the selector was already populated
but it sounds like you tried that approach and it didn't work? can you elaborate on why?

milas:
My concern was reading parts of the `LiveUpdateSpec` while populating/inferring values for it and making that logic brittle and harder to reason about (which values can you pre-set? what is the interaction between those and inferred values?).

That said, yeah, this is moving in the wrong direction - let me swap back to that approach. It's definitely not untenable 🙃

milas:
Okay, revamped to pre-populate `.spec.selector.kubernetes.containerName` and hinge image inference off that being populated!

nicks (approved):

PR #5250, opened 2021-12-03
apis: add a Cluster object

nicks:
Hello @milas, @nicksieger,

Please review the following commits I made in branch nicks/clusterdef:

4150728ed810dee28a5d487aac1209000cdd0e76 (2021-12-03 15:33:35 -0500)
apis: add a Cluster object


pkg/apis/core/v1alpha1/cluster_types.go
```diff
@@ -0,0 +1,159 @@
...
+}
+
+// ClusterSpec defines how to find the cluster we're running
+// containers on.
+//
+// Tilt currently supports connecting to an existing Kubernetes
+// cluster or an existing Docker Daemon (for Docker Compose).
+type ClusterSpec struct {
+	// Connection spec for an existing cluster.
+	Connect *ClusterConnection `json:"connect,omitempty" protobuf:"bytes,1,opt,name=connect"`
```
milas, line 63: Total nit, but s/Connect/Connection here for consistency with object type?

pkg/apis/core/v1alpha1/cluster_types.go
```diff
@@ -0,0 +1,159 @@
...
+	// If not specified, will use the namespace in the kubeconfig.
+	Namespace string `json:"namespace,omitempty" protobuf:"bytes,2,opt,name=namespace"`
+}
+
+type DockerClusterConnection struct {
+	// The docker host to use.
+	//
+	// If not specified, will read the DOCKER_HOST env or use the default docker
+	// host.
+	Host string `json:"host,omitempty" protobuf:"bytes,1,opt,name=host"`
```
milas, line 92: Not critical, but I'd add a couple examples of valid Docker connection strings here - "host" is ambiguous (Docker's fault, I realize), so it'd be helpful for anyone reading the docs to realize this is something like `unix:///var/run/docker.sock` or `tcp://localhost:12345` and not just a hostname like `docker.cluster.local`

pkg/apis/core/v1alpha1/cluster_types.go
```diff
@@ -0,0 +1,159 @@
...
+	return &in.ListMeta
+}
+
+// ClusterStatus defines the observed state of Cluster
+type ClusterStatus struct {
+	// The preferred chip architecture of the cluster.
+	//
+	// On Kubernetes, this will correspond to the kubernetes.io/arch annotation.
+	//
+	// On Docker, this will be the Architecture of the Docker daemon.
```
milas, line 143: Do we need to normalize between these or do they use the same values? 🙏

milas (approved):

pkg/apis/core/v1alpha1/cluster_types.go
nicks: done

pkg/apis/core/v1alpha1/cluster_types.go
nicks: done

pkg/apis/core/v1alpha1/cluster_types.go
nicks: clarified that we're going to use GOARCH style!

pkg/apis/core/v1alpha1/cluster_types.go
```diff
@@ -0,0 +1,159 @@
+/*
+Copyright 2020 The Tilt Dev Authors
```
nicksieger, line 2:
```suggestion
Copyright 2021 The Tilt Dev Authors
```
old copyright template. .emacs aside: `(add-hook 'write-file-hooks 'copyright-update)` 😉

pkg/apis/core/v1alpha1/cluster_types.go
nicksieger: Since arch comes from nodes in the cluster, we're all good until someone makes a multi-node, multi-arch cluster. Maybe note that fact/assumption of uniform, "cluster"-wide arch?

nicksieger (approved):
LGTM! Exciting to see.

pkg/apis/core/v1alpha1/cluster_types.go
nicks: yep, didn't intend to suggest that this was the only arch, added a clarification

PR #5255, opened 2021-12-06
tiltfile: by default, local servers can start in parallel

nicks:
Hello @landism,

Please review the following commits I made in branch nicks/allowparallel:

af2afc1678b44f376cc0fa4108987edf6b3ab26a (2021-12-06 18:07:44 -0500)
tiltfile: by default, local servers can start in parallel


nicks:
once we have a restart mechanism that doesn't rely on buildcontrol, it will be much easier to fix the underlying problem here (local servers that depend on a local cmd)

landism (approved):

PR #5254, opened 2021-12-06
dockercompose: greenlight dc version 2.2 or later

nicksieger:
Fixes #4890.

Based on [this comment](https://github.com/tilt-dev/tilt/issues/4993#issuecomment-981637127) on #4993 as well as commentary at the end of docker/compose#8554, version 2.2 or greater appears to be stable.

PR #5256, opened 2021-12-06
down: support specifying services w/ DC

landism:
### Repro
w/ a DC config:
`tilt up foo bar baz`
`tilt down foo`

### Expected
tilt stops "foo"

### observed
tilt stops all services specified in the DC config

### Solution
AFAICT the only option to stop individual DC services is `docker-compose rm`, so use that if only a subset of the DC services are specified.

### DIFF
--- a/internal/cli/down.go
+++ b/internal/cli/down.go
@@ -12,8 +12,10 @@ import (
 
 	"github.com/tilt-dev/tilt/internal/analytics"
 	ctrltiltfile "github.com/tilt-dev/tilt/internal/controllers/apis/tiltfile"
+	"github.com/tilt-dev/tilt/internal/dockercompose"
 	"github.com/tilt-dev/tilt/internal/k8s"
 	"github.com/tilt-dev/tilt/internal/localexec"
+	"github.com/tilt-dev/tilt/internal/sliceutils"
 	"github.com/tilt-dev/tilt/pkg/logger"
 	"github.com/tilt-dev/tilt/pkg/model"
 )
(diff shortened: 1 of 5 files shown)


nicksieger:
Looks like docker-compose v2 has a `stop` command that you could use. I just turned on support for v2 in #5254 today.

internal/cli/down.go
```diff
@@ -80,25 +82,55 @@ func (c *downCmd) down(ctx context.Context, downDeps DownDeps, args []string) er
...
-	if !model.IsEmptyDockerComposeProject(dcProject) {
+	if len(dcServices) > 0 {
 		dcc := downDeps.dcClient
-		err = dcc.Down(ctx, dcProject, logger.Get(ctx).Writer(logger.InfoLvl), logger.Get(ctx).Writer(logger.InfoLvl))
+		// at the time of writing, a Tiltfile can only have one DC project
+		proj := dcServices[0].Project
+
+		// `docker-compose down` removes all services in the config
+		// if we are only stopping a subset, we use `docker-compose rm` instead
+		// (don't *always* use `rm` because it doesn't remove networks)
```
nicks, line 99: hmmmm...what's are the pros/cons of removing the networks? is it possible to garbage collect the networks ourselves?

i always get worried when there are two codepaths that are supposed to do the same thing but are implemented differently. Because then end-users on the team get very confused about why things are mysteriously failing, because it's non-obvious there are different codepaths.

but i guess it's fine for now, bc this unblocks things

internal/sliceutils/sliceutils.go
```diff
@@ -52,6 +52,31 @@ func StringSliceEquals(a, b []string) bool {
 	return true
 }
 
+// return true iff `a` has the same elements as `b`, ignoring order
```
nicks, line 55: can we use `cmp` for this?
https://pkg.go.dev/github.com/google/go-cmp/cmp#example-Option-SortedSlice
generally more idiomatic than writing our own cmp functions

internal/cli/down.go
```diff
@@ -80,25 +82,55 @@ func (c *downCmd) down(ctx context.Context, downDeps DownDeps, args []string) er
...
+		useDown, err := isAllServicesInProject(ctx, dcc, proj, dcServices)
 		if err != nil {
-			return errors.Wrap(err, "Running `docker-compose down`")
+			return err
+		}
+		if useDown {
+			err := dcc.Down(ctx, proj, logger.Get(ctx).Writer(logger.InfoLvl), logger.Get(ctx).Writer(logger.InfoLvl))
+			if err != nil {
+				return errors.Wrap(err, "Running `docker-compose down`")
```
nicks, line 107: i think FormatError in the Down() method already does this wrapping? feels redundant

nicks (approved):

landism:
> Looks like docker-compose v2 has a stop command that you could use.

Thanks! I don't see any advantages there worth dropping v1 compatibility for - am I missing something?

nicksieger:
My previous comment wasn't accurate, apparently v1 has `stop` also? Is this the behavior you want? Maybe `rm` is closer.
```
$ docker-compose-v1 stop --help
Stop running containers without removing them.

They can be started again with `docker-compose start`.

Usage: stop [options] [--] [SERVICE...]

Options:
  -t, --timeout TIMEOUT      Specify a shutdown timeout in seconds.
                             (default: 10)
```

internal/cli/down.go
landism: > what's are the pros/cons of removing the networks?

I don't have much confidence in an answer to this, but I suspect it doesn't matter much in practice. I mostly figured that behavior was worth preserving because people like to have a tool fully clean up after itself.

> is it possible to garbage collect the networks ourselves?

It looks like this is *mostly* `docker network prune --filter=label=com.docker.compose.project=$PROJECT_NAME`, but I haven't found a good way to get `$PROJECT_NAME`.

> but i guess it's fine for now, bc this unblocks things

I actually don't need the down.go changes to unblock anything, I just hoped it'd be a simple drive-by fix as long as I was adding `rm` support to the docker compose client. This is looking like a bigger yak than I want to shave at the moment, especially for an issue that no user has actually reported.

nicksieger: I agree it's not probably worth shaving that yak right now. Just thought I'd point out that the [default project name is a normalized form of the name of the project directory](https://github.com/compose-spec/compose-go/blob/master/cli/options.go#L341-L350). I'm codifying this as well to make it easier to provide our own values for the project name in my PR #5259.

nicks:
what's the current status of this PR? mainly making sure you're not waiting on me for something

landism:
I've pulled the DC client changes out into #5296, which is the actual critical path I'm on, and will take a look at actually fixing this issue after the holidays. Closing for now.

PR #5258, opened 2021-12-07
engine: trigger builds of resources when they become enabled

landism:
We found a few cases where reenabling a resource did not properly restore it to its previous state, including:
* The resource's FileWatches are disabled, so any files that changed while the resource was disabled won't be automatically rebuild on re-enable.
* The k8s reconciler won't apply k8s outside of the BaD flow.
* DC resources don't even have a reconciler that could reenable them.
* If the initial build was canceled by disabling the resource, Tilt records that it tried and failed to build that resource, which means it wouldn't rebuild on re-enable.

There are probably more we haven't found. We could (and probably should) fix these individually, but for now let's just force a build when a resource is re-enabled.


nicks:
in the past, these kinds of modelling shortcuts have caused tons of pain down the line. They tend to require more kludges, and cover up real bugs that are time-consuming to debug and fix.

Have you considered making it so that disabling a resource deletes the build history or deletes the ManifestState? That's closer to what the reconcilers do (e.g., https://github.com/tilt-dev/tilt/blob/master/internal/controllers/core/cmd/controller.go#L180) and seems more correct.

landism:
> Have you considered making it so that disabling a resource deletes the build history or deletes the ManifestState? That's closer to what the reconcilers do (e.g., https://github.com/tilt-dev/tilt/blob/master/internal/controllers/core/cmd/controller.go#L180) and seems more correct.

I had not! I suspect this is a can of worms and will break a bunch of assumptions made by various code that touches ManifestState, but we've probably already breaking a bunch of things by disabling resources without wiping it out, and just haven't noticed yet.

Wiping ManifestState seems like the right long-term path. I'll see how much progress I make in a couple days and might be tempted next week to advocate for the hack to get it into users' hands sooner.

PR #5259, opened 2021-12-07
dockercompose: get `docker_compose(Blob)` wired up

### LINKED ISSUE #4999: Support docker_compose(blob)
Currently, `k8s_yaml` accepts either a filename or a Blob of dynamically constructed YAML
https://docs.tilt.dev/api.html#api.k8s_yaml

`docker_compose` only accepts a filename.
https://docs.tilt.dev/api.html#api.docker_compose

It would be useful for `docker_compose to also accept a blob, for teams that dynamically generate docker-compose YAML from a template.

AFAIR, Maia mostly implemented dynamically-generated docker-compose YAML in this PR:
https://github.com/tilt-dev/tilt/pull/3200
as part of an unrelated effort to add injected entrypoints for Docker Compose. But we ended up not needing injected entrypoints, and scrapped the PR. I don't think it would be too hard to revive the part that handled dynamic YAML.

nicksieger:
Fixes #4999.

Supports any combination of compose file paths and blobs, with the following path resolution:
1. The compose project directory will be set to the parent directory of the first file, if any.
2. Otherwise, the compose project directory will be set to the parent directory of the executing Tiltfile.

Caveats that still exist:
- Multiple `docker_compose` files, blobs, and declarations with any combination of files and blobs are merged into a single compose project with the directory inferred as above.

### DIFF
--- a/internal/dockercompose/fake_client.go
+++ b/internal/dockercompose/fake_client.go
@@ -151,7 +151,7 @@ func (c *FakeDCClient) Project(_ context.Context, m model.DockerComposeProject) 
 	workDir := c.WorkDir
 	projectName := m.Name
 	if projectName == "" {
-		projectName = workDir
+		projectName = model.NormalizeName(workDir)
 	}
 	if projectName == "" {
 		projectName = "fakedc"
--- a/internal/dockercompose/fake_client.go
+++ b/internal/dockercompose/fake_client.go
@@ -5,8 +5,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
-	"path/filepath"
-	"regexp"
 	"strings"
 	"testing"
 	"time"
@@ -151,18 +149,9 @@ func (c *FakeDCClient) Project(_ context.Context, m model.DockerComposeProject) 
 	}
 
 	workDir := c.WorkDir
-	projectNameOpt := func(opt *loader.Options) {
-		if workDir != "" {
-			name := filepath.Base(workDir)
-			// normalization logic from https://github.com/compose-spec/compose-go/blob/c39f6e771fe5034fe1bec40ba5f0285ec60f5efe/cli/options.go#L366-L371
-			r := regexp.MustCompile("[a-z0-9_-]")
-			name = strings.ToLower(name)
-			name = strings.Join(r.FindAllString(name, -1), "")
-			name = strings.TrimLeft(name, "_-")
-			opt.Name = name
-		} else {
-			opt.Name = "fakedc"
-		}
+	projectName := workDir
+	if projectName == "" {
+		projectName = "fakedc"
 	}
 
 	p, err := loader.Load(types.ConfigDetails{
@@ -173,7 +162,7 @@ func (c *FakeDCClient) Project(_ context.Context, m model.DockerComposeProject) 
 			},
 		},
 		Environment: opts.Environment,
-	}, dcLoaderOption, projectNameOpt)
+	}, dcLoaderOption(projectName))
 	return p, err
 }
 
--- a/internal/dockercompose/fake_client.go
+++ b/internal/dockercompose/fake_client.go
@@ -149,7 +149,10 @@ func (c *FakeDCClient) Project(_ context.Context, m model.DockerComposeProject) 
 	}
 
 	workDir := c.WorkDir
-	projectName := workDir
+	projectName := m.Name
+	if projectName == "" {
+		projectName = workDir
+	}
 	if projectName == "" {
 		projectName = "fakedc"
 	}
--- a/internal/tiltfile/docker_compose.go
+++ b/internal/tiltfile/docker_compose.go
@@ -116,7 +116,7 @@ func (s *tiltfileState) dockerCompose(thread *starlark.Thread, fn *starlark.Buil
 		Project:      project,
 		configPaths:  project.ConfigPaths,
 		services:     services,
-		tiltfilePath: starkit.CurrentExecPath(thread),
+		tiltfilePath: currentTiltfilePath,
 	}
 
 	return starlark.None, nil
--- a/internal/tiltfile/docker_compose.go
+++ b/internal/tiltfile/docker_compose.go
@@ -54,13 +54,18 @@ func (s *tiltfileState) dockerCompose(thread *starlark.Thread, fn *starlark.Buil
 	}
 
 	dc := s.dc
-	project := model.DockerComposeProject{ConfigPaths: dc.configPaths, ProjectPath: dc.Project.ProjectPath}
 
 	currentTiltfilePath := starkit.CurrentExecPath(thread)
 	if dc.tiltfilePath != "" && dc.tiltfilePath != currentTiltfilePath {
 		return starlark.None, fmt.Errorf("Cannot load docker-compose files from two different Tiltfiles.\n"+
 			"docker-compose must have a single working directory:\n"+
 			"(%s, %s)", dc.tiltfilePath, currentTiltfilePath)
+	}
+
+	project := model.DockerComposeProject{
+		ConfigPaths: dc.configPaths,
+		ProjectPath: dc.Project.ProjectPath,
+		Name:        model.NormalizeName(filepath.Base(currentTiltfilePath)),
 	}
 
 	for _, val := range paths {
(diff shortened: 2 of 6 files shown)


nicksieger:
@nicks do you think we should support any of these cases?

```python
# Path + Blob
docker_compose(['docker-compose1.yml', read_file('docker-compose2.yml')])
# Multiple blobs
docker_compose([read_file('docker-compose1.yml'), read_file('docker-compose2.yml')])
# Multiples via multiple docker_compose declarations
docker_compose('docker-compose1.yml')
docker_compose(read_file('docker-compose2.yml'))
docker_compose(read_file('docker-compose3.yml'))
```

internal/tiltfile/docker_compose.go
```diff
@@ -37,42 +37,57 @@ type dcResourceSet struct {
...
 
 	dc := s.dc
+	project := model.DockerComposeProject{ConfigPaths: dc.configPaths, YAML: dc.Project.YAML, ProjectPath: dc.Project.ProjectPath}
+
+	for _, val := range paths {
+		switch v := val.(type) {
+		case nil:
+			continue
+		case io.Blob:
+			project.YAML = v.String()
```
nicks, line 61: is it intended that this should wipe out previous yaml? Or can you pass in multiple yaml files separate by "\n---\n" like you can with kubernetes?

nicks (commented):
for now it's fine to assume that all params are config paths or all params are yaml, and emit an error if the user specifies both.

long-term, i think it would be reasonable to have the tiltfile assembler always define a Project as YAML + ProjectPath (rather than ConfigPaths)

internal/tiltfile/docker_compose.go
nicksieger: Just incomplete implementation at this point, I was waiting to decide whether to combine with previous YAML or generate an error.

That's an interesting idea about a YAML stream. Not sure if DC supports it but I'll check.

internal/tiltfile/docker_compose.go
```diff
@@ -37,33 +40,72 @@ type dcResourceSet struct {
...
-	allConfigPaths = append(allConfigPaths, configPaths.Value...)
-	project := model.DockerComposeProject{ConfigPaths: allConfigPaths}
+	for _, val := range paths {
+		switch v := val.(type) {
+		case nil:
+			continue
+		case io.Blob:
+			yaml := v.String()
+			message := "unable to store yaml blob"
+			tmpfile, err := os.CreateTemp("", fmt.Sprintf("%s-docker-compose-*.yml", filepath.Base(filepath.Dir(currentTiltfilePath))))
```
nicksieger, line 73: One minor issue is tempfile proliferation -- we never clean them up. Is there an appropriate place that could happen, or do we just live with it?

internal/dockercompose/fake_client.go
```diff
@@ -151,18 +149,12 @@ func (c *FakeDCClient) Project(_ context.Context, m model.DockerComposeProject)
...
-			name = strings.ToLower(name)
-			name = strings.Join(r.FindAllString(name, -1), "")
-			name = strings.TrimLeft(name, "_-")
-			opt.Name = name
-		} else {
-			opt.Name = "fakedc"
-		}
+	projectName := m.Name
+	if projectName == "" {
+		projectName = workDir
```
milas, line 154: I think you'll need to bring back the normalization logic for this case. It's possible that `strings.Lower(filepath.Base(workDir))` might be sufficient, but the test tempdirs definitely don't make it happy by default.

Alternatively, we could just always hardcode the project name to `fakedc` here - I was trying to keep behavior fairly realistic by inferring project name from the working dir, but I don't think anything test-related _actually_ cares

internal/tiltfile/docker_compose.go
```diff
@@ -37,33 +40,77 @@ type dcResourceSet struct {
...
+			project.ConfigPaths = append(project.ConfigPaths, tmpfile.Name())
+		default:
+			path, err := value.ValueToAbsPath(thread, val)
+			if err != nil {
+				return starlark.None, fmt.Errorf("expected blob | path (string). Actual type: %T", val)
+			}
+
+			// Set project path to dir of first compose file, like DC CLI does
+			if project.ProjectPath == "" {
+				project.ProjectPath = filepath.Dir(path)
```
milas, line 99: I think this should also be where `project.Name` is set - currently it's going to name after the `Tiltfile` dir

internal/tiltfile/docker_compose.go
milas: What springs to mind:
 * Could (lazily) create a tempdir for the current Tilt process first time needed (stored on `tiltfileState`) and then in the loader set up a long-running goroutine that just blocks on `ctx.Done()` and then deletes the whole dir
 * Name these files deterministically (e.g. hash the string) so that we don't keep creating new files every Tiltfile invocation, only when the blob changes

.circleci/Dockerfile
```diff
@@ -14,6 +14,10 @@ RUN curl -sL https://deb.nodesource.com/setup_16.x | bash -
 
 RUN apt-get install -y nodejs
 
+RUN mv /usr/bin/docker-compose /usr/bin/docker-compose-v1
+ARG DOCKER_COMPOSE_VERSION=v2.2.2
```
milas, line 18: Just checking: we don't have a hard dependency on v2 now, do we?

internal/tiltfile/docker_compose.go
```diff
@@ -37,33 +40,77 @@ type dcResourceSet struct {
...
+		case io.Blob:
+			yaml := v.String()
+			message := "unable to store yaml blob"
+			tmpfile, err := os.CreateTemp("", fmt.Sprintf("%s-docker-compose-*.yml", filepath.Base(filepath.Dir(currentTiltfilePath))))
+			if err != nil {
+				return nil, errors.Wrap(err, message)
+			}
+			_, err = tmpfile.WriteString(yaml)
+			if err != nil {
+				return nil, errors.Wrap(err, message)
```
milas, line 84: Should close the file handle here as well to avoid leaking

milas (approved):
The tempfile approach seems reasonable as a step here - ideally we can eventually switch to just holding on to the YAML directly to pass to the loader, and then re-serialize to provide to `docker-compose` via Stdin, but unfortunately `compose-go` re-serialized data isn't compatible with Docker Compose v1 IIRC (it gets mad about some fields it doesn't know about, I think)

.circleci/Dockerfile
nicksieger: Not that I am aware of -- this is to allow us to test primarily against v2. If the fallback CLI loader is used and `which docker-compose` still returns a v1 composer, we'd still use that.

internal/dockercompose/fake_client.go
nicksieger: Right, okay. That's easy enough to do now.

nicksieger: Fixed in e757a7d718665cb7f20ccd042b3a91204debcab4

.circleci/Dockerfile
nicksieger: Do we think it would be useful to still run some CI against v1?

internal/tiltfile/docker_compose.go
nicksieger: Yeah, I had a previous version where I was hashing the string, that would definitely cut down on the number of files at least.

internal/tiltfile/docker_compose.go
nicksieger: Yeah, since we only support one compose spec per Tilt instance right now, I made the specific decision to always name the project after the directory the Tiltfile is in. We could do this once we support multiple compose specs per instance...

nicksieger:
@milas can you take a look at my recent changes? Any more suggestions? Otherwise I think this is ready to go. Thanks.

internal/tiltfile/tiltfile_docker_compose_test.go
```diff
@@ -109,6 +126,60 @@ func TestDockerComposeManifest(t *testing.T) {
...
+	assert.Equal(t, 2, len(f.loadResult.Manifests))
+}
+
+func TestDockerComposeBlobAndFileUsesFileDirForProjectPath(t *testing.T) {
+	f := newFixture(t)
+	defer f.TearDown()
+
+	f.dockerfile(filepath.Join("foo", "Dockerfile"))
+	f.file("docker-compose.yml", simpleConfig)
+	f.file("Tiltfile", fmt.Sprintf(`docker_compose([blob("""\n%s\n"""), 'docker-compose.yml'])`, barServiceConfig))
```
milas, line 171: Ooo this is a great test case ✨

milas (approved):
:shipit:

internal/tiltfile/docker_compose.go
nicksieger: Yeah, I think a follow-on PR to add some more of the DC command-line options -- eg. #5265 -- is a strong possibility.

PR #5260, opened 2021-12-08
cluster: add cluster reconciler and create connections

nicks:
Hello @milas,

Please review the following commits I made in branch nicks/cluster2:

81c5297116e3f0229839598ec68d7222370df229 (2021-12-07 19:19:33 -0500)
cluster: add cluster reconciler and create connections


internal/controllers/core/cluster/reconciler.go
```diff
@@ -22,20 +37,144 @@ func (r *Reconciler) CreateBuilder(mgr ctrl.Manager) (*builder.Builder, error) {
...
 
 	return ctrl.Result{}, nil
 }
+
+// Creates a docker connection from the spec.
+func (r *Reconciler) createDockerConnection(ctx context.Context, obj *v1alpha1.DockerClusterConnection) *connection {
+	// If no Host is specified, use the default Env from environment variables.
+	env := docker.Env(r.localDockerEnv)
+	if obj.Host != "" {
+		env = docker.Env{Host: obj.Host}
```
milas, line 96: Won't you generally also need TLS client certs in this case?

milas (approved):

internal/controllers/core/cluster/reconciler.go
nicks: yep! i think there are some notes on this in the API structs - we're eventually going to need to support all the major docker client options. But for now i just wanted to have a placeholder here to demo what it would look like.

nicks: (my current old-school remote cluster uses TCP+an authenticated tailscale tunnel without any client certs)

Issue #5252, opened 2021-12-06
auto_init=False does not show trigger button


nicks:
## Repro steps

Create a resource with `auto_init=False`

## Expected behavior

The resource should have a button to start it manually

## Actual behavior

No button appears

## Other notes

As a workaround, you can still trigger the resource as `tilt trigger [resource-name]` from the CLI

PR #5261, opened 2021-12-08
web: better unify all the trigger buttons

nicks:
Hello @lizzthabet,

Please review the following commits I made in branch nicks/ch13060:

e85f34f14be06994b6a9b557e7d25df00575d1d1 (2021-12-08 09:24:09 -0500)
web: better unify all the trigger buttons
fixes https://github.com/tilt-dev/tilt/issues/5252

### DIFF
--- a/web/src/OverviewTableTriggerButton.tsx
+++ b/web/src/OverviewTableTriggerButton.tsx
@@ -42,7 +42,7 @@ export let TriggerButtonRoot = styled(InstrumentedButton)`
   &.is-disabled:active svg {
     transform: scale(1);
   }
-  &.is-bold .fillStd {
+  &.is-emphasized .fillStd {
     fill: ${Color.blue};
   }
 `
@@ -72,12 +72,12 @@ function OverviewTableTriggerButton(props: TriggerButtonProps) {
     !props.isBuilding && // currently building
     !(isAutoInit && !props.hasBuilt) // waiting to perform its initial build
 
-  let isBold = false
+  let isEmphasized = false
   if (clickable) {
     if (props.hasPendingChanges && isManual) {
-      isBold = true
+      isEmphasized = true
     } else if (!props.hasBuilt && !isAutoInit) {
-      isBold = true
+      isEmphasized = true
     }
   }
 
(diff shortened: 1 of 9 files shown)


nicks:
there's still a bunch of copy-and-pasted code between SidebarTriggerButton and OverviewTriggerButton, but i tried to make it so at least the code is the SAME copy-and-pasted code :grimacing:

web/src/OverviewTableTriggerButton.tsx
```diff
@@ -64,80 +55,48 @@ type TriggerButtonProps = {
...
-    if (props.hasPendingChanges && isManualTriggerMode) {
-      shouldBeClicked = true
-    } else if (!props.hasBuilt && isManualTriggerIncludingInitial) {
-      shouldBeClicked = true
+  let isBold = false
+  if (clickable) {
+    if (props.hasPendingChanges && isManual) {
+      isBold = true
+    } else if (!props.hasBuilt && !isAutoInit) {
+      isBold = true
```
lizzthabet, line 80: fyi, there is a slight discrepancy between table view and detail view when a manual resource has been triggered once, even though it looks like this logic is the same! i did check on the main branch and it seems like this is happening there too, so it's not really anything to fix here.

detail view fills the icon with blue:
Screen Shot 2021-12-08 at 11 58 09 AM

table view does not (but it is blue when the resource is manual init and has not been built):
Screen Shot 2021-12-08 at 11 58 02 AM

lizzthabet (approved):
happy to see some code cleanup with a bug fix! 🎉 

is there any reason in the future to keep these as separate components? the styling is slightly different between the two views, but at first glance, i don't see major functionality differences. (asking for my curiosity, and not because i think this pr should do more.)

i may create a backlog ticket for myself to clean up the hover/focus states because i don't think the colors are accessible, and i could tackle other cleanup.

web/src/OverviewTableTriggerButton.tsx
```diff
@@ -64,80 +55,48 @@ type TriggerButtonProps = {
...
+    !props.isBuilding && // currently building
+    !(isAutoInit && !props.hasBuilt) // waiting to perform its initial build
 
-  let shouldBeClicked = false
-  if (!isDisabled) {
-    if (props.hasPendingChanges && isManualTriggerMode) {
-      shouldBeClicked = true
-    } else if (!props.hasBuilt && isManualTriggerIncludingInitial) {
-      shouldBeClicked = true
+  let isBold = false
```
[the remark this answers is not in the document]
nicks: done!

web/src/OverviewTableTriggerButton.tsx
nicks: haha -- yahhhhh, there are a bunch of discrepancies here. honestly i meant this to be a 2-line PR but it was hard to resist not fixing some of them (tho there are still some left)

nicks:
:+1:

nicks:
speaking to: "is there any reason in the future to keep these as separate components?"

i tend to lean on the side of https://sandimetz.com/blog/2016/1/20/the-wrong-abstraction - i.e., to bias towards code duplication if you're not sure how to come up with the right abstraction for an attribute. tho reasonable people differ on this. i wouldn't be opposed to trying to unify them, but also wouldn't take for granted that it would be an improvement.

PR #5262, opened 2021-12-08
k8s: wait for delete to be complete on trigger/full rebuild

milas:
This adapts the code originally added in #4626 to apply to all
delete cases. Previously, it was only done in `deleteAndCreate`.
However, when `Delete` is called explicitly, it's done as part of
a resource trigger/full rebuild, so the same eventual issue applies
where Tilt is going to (almost) immediately re-create the entities
it just deleted and might hit conflicts as a result.

Without this, you currently see errors like this (from @Chimney42):
Tilt UI showing an error about build failing because object already exists but is being deleted


nicks (approved):

nicks:
i think my one (minor) concern is that `tilt down` uses this codepath. I'm not sure if we want it to wait or not? but let's try it and see how it goes.

milas:
Hmm yeah it's also used for GCing old entities. I'll add a `waitForDelete bool` arg. (I must have been fuzzier than I realized this morning because I could have sworn that GoLand was showing the image BaDer as the only usage 😵)

nicksieger (approved):

PR #5270, opened 2021-12-10
ci: test against docker-compose v1+v2

nicksieger:
- integration: upgrade go to work with arm64/aarch64
- ci: Install DC v2 on Windows
- ci: Install DC v2 on CI images

### DIFF
--- a/internal/dockercompose/client.go
+++ b/internal/dockercompose/client.go
@@ -115,7 +115,9 @@ func (c *cmdDCClient) Up(ctx context.Context, spec model.DockerComposeUpSpec, sh
 	c.mu.Lock()
 	defer c.mu.Unlock()
 	runArgs := append([]string{}, genArgs...)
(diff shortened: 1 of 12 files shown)


internal/dockercompose/client.go
```diff
@@ -115,7 +115,9 @@ func (c *cmdDCClient) Up(ctx context.Context, spec model.DockerComposeUpSpec, sh
 	c.mu.Lock()
 	defer c.mu.Unlock()
 	runArgs := append([]string{}, genArgs...)
-	runArgs = append(runArgs, "up", "--no-deps", "--no-build", "-d")
+	// Omit --no-build for now to get v2 working.
```
nicks, line 118: can we check the docker compose version here and use --no-build in the v1 case?

nicksieger: Yes, I'll get that in here before preparing to merge.

nicks (approved):

PR #5272, opened 2021-12-10
tiltfile: infer the registry during reconciliation rather than during tiltfile loading

nicks:
Hello @milas, @nicksieger,

Please review the following commits I made in branch nicks/cluster2:

5e4625c0cb9263a802b37bb08541fd522e8e00b6 (2021-12-10 14:02:44 -0500)
tiltfile: infer the registry during reconciliation rather than during tiltfile loading
We want to support a world where
- k8s connections can be reloaded
- there are multiple k8s connections
- we can change image parameters based on arch

and the reconciler (rather than the tiltfile loader) is in a better position to
do all of that.

### DIFF
--- a/internal/tiltfile/tiltfile.go
+++ b/internal/tiltfile/tiltfile.go
@@ -11,6 +11,7 @@ import (
 	wmanalytics "github.com/tilt-dev/wmclient/pkg/analytics"
 	"go.starlark.net/starlark"
 
+	"github.com/tilt-dev/tilt/internal/container"
 	"github.com/tilt-dev/tilt/internal/controllers/apiset"
 	"github.com/tilt-dev/tilt/internal/tiltfile/tiltextension"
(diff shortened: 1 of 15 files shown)


nicks:
mostly this is simply moving code around, taking a bunch of checks that happened while the ImageTarget was being built, moving them to a single InferImageProperties() function, and calling that function after the tiltfile executes

internal/tiltfile/tiltfile.go
```diff
@@ -94,7 +95,6 @@ func ProvideTiltfileLoader(
 	env k8s.Env) TiltfileLoader {
 	return tiltfileLoader{
 		analytics:     analytics,
-		kCli:          kCli,
```
milas, line 97: 🎉

milas (approved):

nicksieger (approved):
Sweet!

PR #5281, opened 2021-12-13
ci: add exportloopref linter and upgrade to golangci-lint v1.43.0

milas:
* `exportloopref` - detects captures of loop variable without
                    re-assignment
  * NOTE: There can be false negatives with this linter to avoid being
          overly strict and annoying!
* ~~`unconvert` - unnecessary/redundant type conversions~~

Also upgraded `golangci-lint` to latest (v1.43.0 published 2021-11-03).


milas:
I don't think either of these are contentious, but let me know if you've got objections! I mostly wanted the `exportloopref` one after fixing a bug related to that recently, so did a quick scan of the available linters and `unconvert` seemed pretty easy/safe to add in as well 🙃

nicksieger:
FYI #5270 -- ok if I merge that first? You'll need to rebuild the ci containers again

milas:
Yep, no problem! I need to figure out what to do about the `goimports` stuff anyway so put this in draft for a bit until I have some down time

.golangci.yml
```diff
@@ -62,6 +62,8 @@ linters:
     - structcheck
     - ineffassign
     - misspell
+    - exportloopref
+    - unconvert
```
landism, line 66: I'd probably lean against including unconvert, despite my temptation to scratch the itch to fix this every time my IDE points one out - it feels like a style thing whose value would be outweighed by annoying CI failures.

Or maybe I'm in the minority in that I only rarely run lint locally rather than relying on CI.

internal/controllers/core/kubernetesdiscovery/container_restart.go
```diff
@@ -53,8 +53,8 @@ func (c *ContainerRestartDetector) logRestarts(dispatcher Dispatcher, mn model.M
 
 func podsByUID(status v1alpha1.KubernetesDiscoveryStatus) map[types.UID]*v1alpha1.Pod {
 	pods := make(map[types.UID]*v1alpha1.Pod)
-	for _, p := range status.Pods {
-		pods[types.UID(p.UID)] = &p
+	for i := range status.Pods {
+		pods[types.UID(status.Pods[i].UID)] = status.Pods[i].DeepCopy()
```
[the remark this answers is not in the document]
milas: Naw, linter would have been happy with what you wrote too.

This was a "free" optimization; since I wanted to use `DeepCopy()`, it made since to eliminate the intermediate copy-by-value from the `range` iteration variable IMO

.golangci.yml
milas: That's fair! Removed from this PR in 1c96f61. I really wanted the `exportloopref` linter and noticed `unconvert` in the list, which personally annoys me (due to the IDE squiggles), but yeah, it's totally a stylistic thing rather than indicative of a bug so I don't _really_ care 😛

milas:
FYI I fixed up CI and removed the `unconvert` linter (finally 😓)

landism (approved):

PR #5282, opened 2021-12-13
tiltfile: create a default cluster connection after the main tiltfile loads

nicks:
Hello @milas,

Please review the following commits I made in branch nicks/cluster3:

342b1dbb845e89a71187ec8ef4a539499350f2fc (2021-12-13 15:12:23 -0500)
tiltfile: create a default cluster connection after the main tiltfile loads


internal/controllers/core/tiltfile/reconciler.go
```diff
@@ -410,6 +416,14 @@ func (r *Reconciler) decideRegistry(ctx context.Context, tlr *tiltfile.TiltfileL
 	return tlr.DefaultRegistry
 }
 
+// The kubernetes connection defined by the CLI.
+func (r *Reconciler) defaultK8sConnection() *v1alpha1.KubernetesClusterConnection {
+	return &v1alpha1.KubernetesClusterConnection{
+		Context:   string(r.k8sContextOverride),
+		Namespace: string(r.k8sNamespaceOverride),
+	}
+}
```
milas, line 425: I'm thinking this through, and I think we might need to create a `ClientConfig` early in process start and then capture the resolved initial context+namespace, so you have an absolute ordering of:
1. Tiltfile defined context/namespace
2. CLI flag context/namespace [or is this actually #1?]
3. Env context/namespace _at time of `tilt up`_ (or at time of reconciler start)

Otherwise, if you change your kubeconfig context/namespace at runtime, a subsequent reconciliation [assuming a spec change somehow] could end up connecting somewhere different.

(This doesn't need to block this PR, and lemme know if I'm talking crazy here 🤪)

milas (approved):

internal/controllers/core/tiltfile/reconciler.go
nicks: ya, i think that's a good question. i'm honestly not sure of the right answer.

this is kind of like how, if you delete a pod from a deployment, the replacement pod will read configmaps and possibly get a different value. 

i think the canonical kubernetes API answer is that you're balancing two competing needs:

- the desire to always read from the source of truth (in this case, the default context in the kubeconfig)
- the desire to avoid task flapping, infinite loops, and disruptive restarts without clear user intent (in this case, you don't want to break the connection and clear the k8s caches a bunch of times if the user is in the middle of modifying the kubeconfig)

there are cases where we WANT users to change their kubeconfig and tilt will connect to the new cluster, but i'm not sure yet we have a clear way to distinguish between the cases we want and the cases we don't want

PR #5284, opened 2021-12-13
controllers: add cluster client cache

milas:
The `ConnectionManager` is the internal struct used by the
Cluster reconciler and has a `ConnectionCache` interface for use
by the other reconcilers that allows read-only access to the
cached Kubernetes/Docker client for a Cluster object.

Once wired up, dependent reconcilers will still be expected to set
up a watch on the Cluster and re-evaluate when it changes to handle
runtime changes to the Cluster config. However, they can all share
the same underlying client(s), as they do now, to reduce number of
connections and share any client-level caching.


internal/controllers/core/cluster/reconciler.go
```diff
@@ -104,7 +109,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
 }
 
 // Creates a docker connection from the spec.
-func (r *Reconciler) createDockerConnection(ctx context.Context, obj *v1alpha1.DockerClusterConnection) *connection {
+func (r *Reconciler) createDockerConnection(ctx context.Context, obj *v1alpha1.DockerClusterConnection) connection {
```
nicksieger, line 112: Curious about Go type signature design here -- what led you to change this from `*connection` to `connection` (and corresponding cascading changes)? The connection struct is big enough that it would seem like passing copies of it around would be a little more expensive -- aren't you duplicating all the possible state in the k8s/docker client structs as well?

internal/controllers/core/cluster/cache.go
```diff
@@ -0,0 +1,104 @@
...
+
+// ClientCache provides cached clients for use by reconcilers.
+//
+// All clients are goroutine-safe.
+type ClientCache interface {
+	// GetK8sClient returns the Kubernetes client for the cluster or an error for unknown clusters, connections
+	// in a transient error state, or if the connection is of a different type (i.e. Docker).
+	GetK8sClient(clusterKey types.NamespacedName) (k8s.Client, error)
+
+	// GetDockerClient returns the Docker client for the cluster or an error for unknown clusters, connections
```
nicks, line 27: is there a reason we need to expose the `docker.Client`? I'd probably avoid it if we can.

The phrase "the Docker client for the cluster" already means something very different for the rest of the codebase.

You're using it here to mean "the Docker instance that Docker Compose runs its container on."

In other places in the codebase, we use it to mean "the Docker instance that Minikube runs its containers on"

internal/controllers/core/cluster/reconciler.go
```diff
@@ -58,43 +55,51 @@ func (r *Reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
...
-		delete(r.connections, nn)
+		r.connManager.delete(nn)
 		return ctrl.Result{}, err
 	}
 
 	// The apiserver is the source of truth, and will ensure the engine state is up to date.
 	r.store.Dispatch(clusters.NewClusterUpsertAction(&obj))
 
-	connection, hasConnection := r.connections[nn]
+	var connectionChanged bool
```
nicks, line 65: why bother tracking connectionChanged? is it just an optimization, or is it preventing some sort of loop?

nicks (approved):

internal/controllers/core/cluster/cache.go
milas: I removed it from the interface and renamed the (now-inaccessible) method to `GetComposeDockerClient` in 7a4b44be9, which is awkward but hopefully clear(er) re: purpose.

(We will need it eventually when adding DC support to LU API, no? So I didn't want to totally remove the logic/tests for it. But if I'm misunderstanding and it's totally unnecessary lmk and I'll purge it)

internal/controllers/core/cluster/reconciler.go
milas: Removed in 08d3b3247f23443b985045bdd32fcfe4565095fb to unconditionally overwrite to make semantics simpler.

(It was mostly a vestige that didn't make as much sense as implementation grew - `Cluster` shouldn't be getting reconciled excessively so I'm not super concerned about thrashing the map)

internal/controllers/core/cluster/reconciler.go
milas: I switched to value types for `connection` to avoid goroutine-safety pitfalls - I think a lot of the Tilt codebase has used this pattern to good effect where we copy a value type, mutate, and then wholesale replace the entry in the shared map, so that we don't have to worry it.

From a performance perspective, the client types are interfaces, so they're both pointers and we won't actually copy the full structs. The prevailing wisdom I've seen for Go code is that preferring value types where appropriate is the right place to start and then change if it presents a performance problem. Copying the value and keeping it on stack is often actually much faster than having stuff on heap!

nicksieger: Thanks! That helps a lot.

PR #5285, opened 2021-12-14
buildcontrol: wait on cluster connection

nicks:
Hello @landism, @milas,

Please review the following commits I made in branch nicks/clusterwait:

909cce1c920a408abfd47740c2d325fbc3b23a90 (2021-12-13 19:08:46 -0500)
buildcontrol: wait on cluster connection

### DIFF
--- a/internal/controllers/core/cluster/reconciler.go
+++ b/internal/controllers/core/cluster/reconciler.go
@@ -30,6 +30,9 @@ type Reconciler struct {
 	localDockerEnv docker.LocalEnv
 	store          store.RStore
 
+	fakeK8sClient    k8s.Client
+	fakeDockerClient docker.Client
+
 	// TODO(nick): We should have all reconcilers share a client cache, and
 	// be able to wait on the connection to complete.
 	connections map[types.NamespacedName]*connection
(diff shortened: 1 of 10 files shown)


internal/controllers/core/cluster/reconciler.go
```diff
@@ -50,6 +53,11 @@ func NewReconciler(ctrlClient ctrlclient.Client, store store.RStore, localDocker
 	}
 }
 
+func (r *Reconciler) SetFakeClientsForTesting(k8sClient k8s.Client, dockerClient docker.Client) {
+	r.fakeK8sClient = k8sClient
+	r.fakeDockerClient = dockerClient
+}
```
milas, line 59: No objection to this right now since it's not particularly invasive, but do you think we should move to having factory interfaces for the clients eventually?

milas (approved):

internal/controllers/core/cluster/reconciler.go
nicks: i kind of suspect that we'll inject a fake client cache (as in https://github.com/tilt-dev/tilt/pull/5284) and use that everywhere?

nicks: (assuming that the client cache is more like a client factory)

internal/engine/upper_test.go
```diff
@@ -4538,6 +4544,19 @@ func (s fixtureSub) OnChange(ctx context.Context, st store.RStore, _ store.Chang
 	return nil
 }
 
+func (f *testFixture) ensureCluster() {
+	_ = f.ctrlClient.Create(f.ctx, &v1alpha1.Cluster{
```
landism, line 4548: are we assuming the client can't return errors in tests?

landism (approved):

PR #5289, opened 2021-12-14
build: fix a race condition in constructing tarballs

nicks:
Hello @milas, @landism,

Please review the following commits I made in branch nicks/shortread:

54cb443ec1b190fc43f48ea8bfd00a2248a3c188 (2021-12-14 16:27:51 -0500)
build: fix a race condition in constructing tarballs
i was able to reproduce this with a tilt environment
with a local resource that rapidly writes to a file
and a live-update rule that syncs it.

### DIFF
--- a/internal/build/tar.go
+++ b/internal/build/tar.go
@@ -23,6 +23,9 @@ type ArchiveBuilder struct {
 	tw     *tar.Writer
 	filter model.PathMatcher
 	paths  []string // local paths archived
+
+	// A shared I/O buffer to help with file copying.
+	copyBuf *bytes.Buffer
 }
 
 func NewArchiveBuilder(writer io.Writer, filter model.PathMatcher) *ArchiveBuilder {
@@ -31,7 +34,7 @@ func NewArchiveBuilder(writer io.Writer, filter model.PathMatcher) *ArchiveBuild
 		filter = model.EmptyMatcher
 	}
 
-	return &ArchiveBuilder{tw: tw, filter: filter}
+	return &ArchiveBuilder{tw: tw, filter: filter, copyBuf: bytes.NewBuffer(nil)}
 }
 
 func (a *ArchiveBuilder) Close() error {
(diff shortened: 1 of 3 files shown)


internal/build/tar.go
```diff
@@ -232,6 +233,27 @@ func (a *ArchiveBuilder) writeEntry(entry archiveEntry) error {
...
+	// We want to balance two needs:
+	//
+	// 1) Don't blow up heap size when copying large binary blobs.
+	// 2) If there's a race condition (e.g., a user is actively
+	//    writing to a file while we tar it up), we want to make sure
+	//    there's not a mismatch between the size of the file contents
+	//    in the tar, and the size we declare in the header.
+	//
+	// So for sufficiently small files, we copy the file into an in-memory buffer
+	// and use the size as the source of truth.
```
landism, line 245: It's not super clear to me why we're doing this or what happens for large files. Is something like this accurate?

```
// The size header must match the number of bytes. There is room for a race condition here if something writes to the file after we've read the file size.
// For small files, we avoid this by first copying the file into a buffer.
// For larger files, we don't want to copy the whole thing into a buffer, so we just let the error happen.
```

landism:
> After copying the file to the TarWriter, is it worth checking that the file on disk is still what we expect, and erroring if not?

oh, I guess that's handled by the `Flush`

nicks:
there's a related, but separable, issue here - the underlying APIs around tar and kubectl exec pass around asynchronous pipes. 

This creates problems when you're trying to figure out where an error came from - e.g., the exec Stream() doesn't do a great job separating between an error on the stdin pipe (on the client side) and an error in execution (on the server side). https://github.com/tilt-dev/tilt/blob/master/internal/k8s/exec.go#L34

internal/build/tar.go
```diff
@@ -240,11 +262,12 @@ func (a *ArchiveBuilder) writeEntry(entry archiveEntry) error {
 		return errors.Wrapf(err, "%s: writing header", path)
 	}
 
-	// N.B. intentionally do not limit the number of bytes - TarWriter::Write() will
```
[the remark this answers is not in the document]
nicks: ya, though it says the same things as the comment above and below it

internal/build/tar.go
nicks: added this to the comment!

internal/build/tar.go
```diff
@@ -232,6 +233,30 @@ func (a *ArchiveBuilder) writeEntry(entry archiveEntry) error {
...
+	//
+	// For larger files, we don't want to copy the whole thing into a buffer,
+	// because that would blow up heap size. There is some danger that this
+	// will lead to a spurious error when the tar writer validates the sizes.
+	// That error will be disruptive but will be handled as best as we
+	// can downstream.
+	useBuf := header.Size < 5000000
+	var buf *bytes.Buffer
+	if useBuf {
+		buf = bytes.NewBuffer(make([]byte, 0, header.Size))
```
milas, line 252: Arguably a premature optimization, but given that this will likely be the predominate case in most projects, worth storing the buf on `ArchiveBuilder` and re-using it?

milas (approved):
Nice!

internal/build/tar.go
nicks: good point! Also added a test case to catch this issue

PR #5292, opened 2021-12-15
api: concurrent update fixes

milas:
Most of the heavy-lifting is from the `tilt-apiserver` update here,
which fixes three important issues:
 * Fix race condition at storage-level that would result in two
   conflicting updates both being accepted
 * Fix optimistic concurrency check for subresources (i.e. status)
 * Fix `PATCH` semantics so it can't fail due to resource versioning

On the actual Tilt side, the only change currently is to start using
`PATCH` for status updates in the `KuberenetesApply` reconciler,
which has multiple concurrent writers between the standard
reconcilation process and the `ForceApply` calls from the
`BuildAndDeployer`. The latter doesn't support elegant retry, so
any optimistic concurrency failures result in a failed build. By
using `PATCH`, we prevent these failures, which prevents a host of
race conditions, particularly at Tilt startup.

As with the prior round of optimistic concurrency fixes, there's the
possibility that more "object was modified" errors will be seen than
before - these were previously silent and probably resulting in some
odd bugs/inconsistent state behind the scenes.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -51,7 +51,7 @@ require (
 	github.com/tilt-dev/go-get v0.2.1
 	github.com/tilt-dev/localregistry-go v0.0.0-20200615231835-07e386f4ebd7
 	github.com/tilt-dev/probe v0.3.1
-	github.com/tilt-dev/tilt-apiserver v0.5.2
+	github.com/tilt-dev/tilt-apiserver v0.5.4
 	github.com/tilt-dev/wmclient v0.0.0-20201109174454-1839d0355fbc
 	github.com/tonistiigi/fsutil v0.0.0-20200724193237-c3ed55f3b481
 	github.com/tonistiigi/units v0.0.0-20180711220420-6950e57a87ea
--- a/go.sum
+++ b/go.sum
@@ -1135,8 +1135,8 @@ github.com/tilt-dev/localregistry-go v0.0.0-20200615231835-07e386f4ebd7 h1:ysHGL
 github.com/tilt-dev/localregistry-go v0.0.0-20200615231835-07e386f4ebd7/go.mod h1:bZecZHy8tHzZWeF+MEQMugKbGhi9cF1A7tuaZNIxoDs=
 github.com/tilt-dev/probe v0.3.1 h1:PQhXSBkgcGBUU/eKt4vgAUKsAWWjBr2F53xNAc0E7zs=
 github.com/tilt-dev/probe v0.3.1/go.mod h1:F53NFbqblwu5oyIk2t+BPkswiboxuF8e5D3wbPnY4JA=
-github.com/tilt-dev/tilt-apiserver v0.5.2 h1:QRLB8bt4v76MURLNCUs4bqNhsZLfbP9/gIJuCQ9GR6U=
-github.com/tilt-dev/tilt-apiserver v0.5.2/go.mod h1:WfOq4XyMpYqXpMaIXZZodR9GX6clNkueNQCV2u1q5os=
+github.com/tilt-dev/tilt-apiserver v0.5.4 h1:IXWPPZFUrOh4+QAXkUQNyZsSsvBJ5QZpXjTbgAQ/HFY=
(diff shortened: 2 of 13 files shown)


milas:
I'm cautiously optimistic this will make a noticeable improvement in CI reliability!

I was primarily debugging using `TestBuildControllerResourceDeps` which I could consistently get to fail with `-count 100` previously, but have now done several _thousands_ of runs post-fixes without failure 🎉

nicks (approved):

milas:
Based on CI failures, had to use PATCH a few other places:
* KubernetesDiscovery
* Cmd
* FileWatch

These all make sense as they have background goroutines and arguably need to be refactored at some point to use the much more reliable design of some of the newer reconcilers where goroutines/async operations update internal state and trigger a reconcile, which evaluates the full state.

The use of `PATCH` here is arguably a slight improvement over the old behavior - all of these were already pretty susceptible to potentially persisting a "stale" status as-is (even without optimistic concurrency bugs - since they're happening outside a reconcile, they're generally not really considering the spec). That said, I am trying to be cautious and limit changing calls to `PATCH` to places where things are now erroring because they were relying on broken, almost-patch-like semantics previously (unintentionally due to the apiserver bugs).

nicks:
hmmm...i'm not sure i understand...why was this introducing new test failures originally? it seems like a straight vendor update wouldn't cause problems...?

milas:
Because optimistic concurrency was broken on subresource/status updates (tilt-dev/tilt-apiserver#52), anywhere we were doing status updates from outside the reconciler (e.g. monitor goroutines and `ForceApply` callers) would never fail with `Conflict` errors previously.

Effectively, we we had "replace with last-write wins" semantics.

Now, we have "patch with last-write wins" semantics in those places, which is an improvement.

Example: the test failure I started debugging originally for this manifested by KA reconciler + KA-force-apply-via-BaD running simultaneously - reconciler would set DisableStatus while KA-force-apply-via-BaD would set ResultYAML.
With Update/replace, it was last-write-wins (due to aforementioned OC bug), with the impact that one of the writes was lost.
Now, both will succeed and not clobber each other.
⬆️ This is actually the ideal case, fixing OC without changing to patch "broke" this because BaD won't retry like a reconciler

The Goroutine cases are less cut-and-dry -- arguably, these need to be refactored to NOT do API operations at all, but just update internal state + trigger reconcile (as some of the newer reconcilers do). However, if we kept using update (as I was originally), now that the OC bugs are fixed, some were dispatching fatal errors (`Cmd`) while others would more-or-less lose the update. Switching to patch restores the "no conflicts, last-write wins" semantics they unintentionally had before, so it's really the best we can do w/o adding retry or refactoring them.

Hopefully that makes sense?? TL;DR They only "worked" previously due to flawed behavior in the apiserver which meant they would actually never fail in practice, and now that they get OC errors, things fall apart (fatal errors and/or missed updates)

milas:
Set this as draft temporarily - I want to wait until after today's release to merge.

While I'm confident in the actual fixes and have had many clean local test runs + several clean CI runs (including integration/extensions), I think there might still be some places we're doing concurrent status updates and relying on the fact that they'd previously never fail with `Conflict`, so I want to get more internal soak time on it and avoid introducing a support headache during the holidays.

nicks (approved):
Ya, i'm definitely worried that we're just papering over more fundamental problems with this code (as you mentioned in your own comments). we definitely did this more robustly after we had written a few reconcilers, but the early ones don't handle failure well.

milas:
FWIW I'm not _super_ concerned about the places that I had to change to `PATCH` tbh - it's a crutch and icky but it's the same behavior it had previously w/ the broken `Update()` logic (albeit slightly improved in e.g. the `DisableSource` cases since they won't overwrite other fields).

I'm more worried about the code paths I _wasn't_ forced to change that are still using `Update()` but have never encountered `Conflict` before, e.g. `PodLogStream`. These presumably have much rarer (maybe impossible!) possibility for conflicts so haven't been seen in CI/my testing but depending on their failure behavior could be problematic, such as if they're dispatching `StoreError`s.

PR #5294, opened 2021-12-16
kubernetesapply: add images to custom commands

nicks:
Hello @milas,

Please review the following commits I made in branch nicks/kapp-imagemap:

b59556a1259ca7f6f29515ef5453de44b2a52af4 (2021-12-16 15:47:45 -0500)
kubernetesapply: add images to custom commands

### DIFF
--- a/internal/controllers/apis/imagemap/imagemap.go
+++ b/internal/controllers/apis/imagemap/imagemap.go
@@ -1,0 +1,31 @@
+package imagemap
+
+import (
+	"fmt"
+
+	"k8s.io/apimachinery/pkg/types"
+
+	"github.com/tilt-dev/tilt/pkg/apis/core/v1alpha1"
+	"github.com/tilt-dev/tilt/pkg/model"
+)
+
+// Inject completed ImageMaps into the environment of a local process that
+// wants to deploy to a cluster..
+//
+// Current env vars:
+// TILT_IMAGE_i - The reference to the image #i from the point of view of the cluster container runtime.
+// TILT_IMAGE_MAP_i - The name of the image map #i with the current status of the image.
+//
+// where an env may depend on arbitrarily many image maps.
+func InjectIntoDeployEnv(cmd *model.Cmd, imageMapNames []string, imageMaps map[types.NamespacedName]*v1alpha1.ImageMap) error {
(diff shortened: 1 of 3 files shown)


internal/controllers/apis/imagemap/imagemap.go
```diff
@@ -0,0 +1,31 @@
...
+	"github.com/tilt-dev/tilt/pkg/apis/core/v1alpha1"
+	"github.com/tilt-dev/tilt/pkg/model"
+)
+
+// Inject completed ImageMaps into the environment of a local process that
+// wants to deploy to a cluster..
+//
+// Current env vars:
+// TILT_IMAGE_i - The reference to the image #i from the point of view of the cluster container runtime.
+// TILT_IMAGE_MAP_i - The name of the image map #i with the current status of the image.
```
milas, line 17: Is the intended usage here that you'd know your image map name, so iterate through the `TILT_IMAGE_MAP_i` variables until you find the one you want and then get the value from the corresponding `TILT_IMAGE_i`?

milas (approved):

internal/controllers/apis/imagemap/imagemap.go
nicks: Good question! I'm hoping that in most cases, it will look something like this:

```
k8s_custom_deploy(
   image_deps=['be', fe'],
   apply_cmd="""
./inject.sh "$TILT_IMAGE_0"
./inject.sh "$TILT_IMAGE_1"
./deploy.sh
""")
```

or use `tilt get imagemap "$TILT_IMAGE_MAP_0"` to get the full imagemap data

i know Make + Bazel both use the approach of having a variable expansion syntax 
https://docs.bazel.build/versions/main/be/make-variables.html

so i guess we could also do something like `$(image fe)` to expand to the full image name BEFORE the command runs.

But i didn't want to go down that path quite yet...having numbered env variables seemed a good place to start that let us change things later.

milas: Ah right, the order is deterministic, so that makes it easy.

Definitely agree that adding another level of variable expansion is excessive for the moment, but that's pretty neat (and would definitely help with POSIX vs Windows 😬)

PR #5296, opened 2021-12-16
engine: stop DC services when disabled

landism:


integration/disable_test.go
```diff
@@ -23,21 +24,57 @@ func TestDisable(t *testing.T) {
...
+	}
+
+	f.WaitUntil(ctx, "service up", func() (string, error) {
+		return f.dockerCmdOutput(psArgs)
+	}, "disabletest")
+
+	f.WaitUntil(ctx, "disable configmap available", func() (string, error) {
+		cmd := exec.Command("tilt", "--port", fmt.Sprintf("%d", f.tilt.port), "get", "cm")
+		out, err := cmd.CombinedOutput()
+		return string(out), err
```
milas, line 59: FYI you can do `f.tilt.Get(ctx, "cm", "disabletest-disable")` (will give back JSON output as `[]byte` and err if any)

internal/dockercompose/client.go
```diff
@@ -162,6 +163,42 @@ func (c *cmdDCClient) Down(ctx context.Context, p model.DockerComposeProject, st
...
+	if logger.Get(ctx).Level().ShouldDisplay(logger.VerboseLvl) {
+		args = append(args, "--verbose")
+	}
+
+	var serviceNames []string
+	for _, s := range specs {
+		serviceNames = append(serviceNames, s.Service)
+	}
+
+	args = append(args, []string{"rm", "--stop", "--force"}...)
```
milas, line 187: Feels like they need `"--please", "--but-really"`

In non-trolling comments, `rm` doesn't make any comment about timeout, while `stop` + `down` both have a timeout argument that's 10 seconds by default. I'm GUESSING (though too lazy to look at the source right now) that means when `rm --stop` is called, you'll get the default 10 second timeout.

I don't think there's much we can do around that, I'm only noting because if you disable a service that's super slow to respond to `SIGTERM` and then disable another shortly after, you might not see any activity on the second for ~10 secs, but 🤷

internal/engine/dcwatch/disable_watcher.go
```diff
@@ -0,0 +1,85 @@
...
+	state := st.RLockState()
+	project := state.DockerComposeProject()
+
+	if model.IsEmptyDockerComposeProject(project) {
+		st.RUnlockState()
+		return nil
+	}
+
+	var specsToDisable []model.DockerComposeUpSpec
+	for _, uir := range state.UIResources {
```
milas, line 40: Since `UIResources` are the source-of-truth, you can:
 * Short-circuit (instead of log only) if `len(summary.UIResources.Changes) == 0`
 * Iterate over `summary.UIResources.Changes` and then look up key in `state.UIResources` to only look at modified `UIResource`s

internal/engine/dcwatch/disable_watcher.go
```diff
@@ -0,0 +1,85 @@
...
+
+	if len(specsToDisable) > 0 {
+		// Upon disabling, the DC event watcher will notice the container has stopped and update
+		// the resource's RuntimeStatus, preventing it from being re-added to specsToDisable.
+		// There is a race here, since this OnChange might get called again before EngineState
+		// gets updated with the new RuntimeStatus. This is fine--calling `docker-compose rm` on
+		// a down service is a no-op.
+		var out bytes.Buffer
+		err := w.dcc.Rm(ctx, specsToDisable, &out, &out)
+		if err != nil {
```
milas, line 72: In the non-error case, should we `logger.Get(ctx).Infof("%s", out.String())`?

milas (approved):
🎉 🎉 🎉

internal/dockercompose/fake_client.go
```diff
@@ -68,6 +90,19 @@ func (c *FakeDCClient) Down(ctx context.Context, p model.DockerComposeProject, s
...
 }
 
+func (c *FakeDCClient) Rm(ctx context.Context, specs []model.DockerComposeUpSpec, stdout, stderr io.Writer) error {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	c.rmCalls = append(c.rmCalls, RmCall{specs})
+	if c.RmError != nil {
+		err := c.RmError
+		c.RmError = err
```
lizzthabet, line 100: i'm curious -- what's the purpose of setting the error again on the client? (i've seen this pattern a couple times in this file.)

internal/engine/dcwatch/disable_watcher.go
```diff
@@ -0,0 +1,85 @@
...
+	"github.com/tilt-dev/tilt/internal/store"
+	"github.com/tilt-dev/tilt/pkg/apis/core/v1alpha1"
+	"github.com/tilt-dev/tilt/pkg/model"
+)
+
+type DisableSubscriber struct {
+	dcc dockercompose.DockerComposeClient
+}
+
+// How does the `New` function relate to the subscriber we're creating?
```
lizzthabet, line 19: you can remove my 🧠 🤔 questions (😅)

lizzthabet (approved):
this worked beautifully! i tested with `servantes` docker-compose resources. the only thing that could be confusing to users is that the `UIResource` status gets updated and logged as `disabled` before the docker rm completes. 

when i disabled a resource that takes a couple seconds to bring down, i get confirmation in the ui, and then a couple seconds later (maybe four in my case?), the container is actually brought down and another log prints, saying it exited.
```
Resource "fortune" disabled. To enable/disable it, use the Tilt Web UI.
```

definitely not a blocker, but something i anticipate we could get feedback from users about.

integration/disable_test.go
landism: nice

internal/dockercompose/client.go
landism: Oh, good note. I don't want to distract by digging into this now, but I've added this comment to the code:
```
	// `docker-compose rm` does not support a `--timeout` option, so it possibly defaults to 10,
	// like `docker-compose stop` or `docker-compose down`.
	// If it turns out this command's timeout is too long, we might want to change this to first
	// call `docker-compose stop --timeout $NUM`, to do the presumably slow part under a smaller
	// timeout.
```

internal/engine/dcwatch/disable_watcher.go
landism: neat!

internal/engine/dcwatch/disable_watcher.go
landism: changing this to pass the logger as `out`

Just poking around now, I'm regularly hitting the 10-second timeout when running `docker-compose rm -sf fortune`, and having some streaming output will be a better experience than clicking the button and nothing happens.

internal/dockercompose/fake_client.go
landism: ha! none!

that should be setting it to nil, and it got copypasta'd throughout

internal/engine/dcwatch/disable_watcher.go
landism: oops, turns out nothing was actually populating that field

landism:
> the `UIResource` status gets updated and logged as `disabled` before the docker rm completes

Yeah, this is ugly.

It's at least a little better now that we're logging the `docker-compose rm` output, so the user can see there's still stuff happening, but that stuff is happening after the UI shows it as disabled.

We'll see if users care enough to make it worth fixing.

PR #5297, opened 2021-12-16
tiltfile: add image deps to k8s_custom_deploy

nicks:
Hello @milas,

Please review the following commits I made in branch nicks/kapp-imagemap2:

e7778b5c430cf2cb4c76e1407cdf683e44898fec (2021-12-16 16:52:19 -0500)
tiltfile: add image deps to k8s_custom_deploy

bd66090651852012ec96ff42f5df9871446ba176 (2021-12-16 16:52:00 -0500)
kubernetesapply: add images to custom commands

### DIFF
--- a/internal/tiltfile/k8s_custom_deploy.go
+++ b/internal/tiltfile/k8s_custom_deploy.go
@@ -28,6 +28,7 @@ func (s *tiltfileState) k8sCustomDeploy(thread *starlark.Thread, fn *starlark.Bu
 	var applyCmdEnv, deleteCmdEnv value.StringStringMap
 	var imageSelector, containerSelector string
 	var liveUpdateVal starlark.Value
+	var imageDeps value.ImageList
 
 	deps := value.NewLocalPathListUnpacker(thread)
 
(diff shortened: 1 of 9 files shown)


internal/tiltfile/value/image.go
```diff
@@ -0,0 +1,46 @@
...
+		return nil
+	}
+
+	var iter starlark.Iterator
+	switch x := v.(type) {
+	case *starlark.List:
+		iter = x.Iterate()
+	case starlark.Tuple:
+		iter = x.Iterate()
+	default:
```
milas, line 27: IIRC if the arg is omitted, `v` will be `nil`, while if `image_deps=None` gets passed explicitly, it'll be `starlark.None`, so it'd be nice to treat that as an empty list here

internal/tiltfile/value/image.go
```diff
@@ -0,0 +1,46 @@
+package value
+
+import (
+	"fmt"
+
+	"github.com/docker/distribution/reference"
+	"go.starlark.net/starlark"
+)
+
+type ImageList []reference.Named
```
milas, line 10: 👏

milas (approved):

internal/tiltfile/k8s_custom_deploy.go
```diff
@@ -45,6 +46,7 @@ func (s *tiltfileState) k8sCustomDeploy(thread *starlark.Thread, fn *starlark.Bu
 		"delete_env?", &deleteCmdEnv,
 		"delete_cmd_bat?", &deleteCmdBatVal,
 		"container_selector?", &containerSelector,
+		"image_deps", &imageDeps,
```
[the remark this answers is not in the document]
nicks: yep!

internal/tiltfile/value/image.go
nicks: good catch! fixed

PR #5306, opened 2021-12-17
api: use K8s connection from Cluster in KubernetesDiscovery reconcile

milas:
Inject the `cluster.ClientCache` and use that to select the Kubernetes
client to use.


internal/controllers/core/kubernetesdiscovery/reconciler.go
```diff
@@ -293,27 +308,29 @@ func (w *Reconciler) cleanupAbandonedNamespaces() {
...
 
 	ctx, cancel := context.WithCancel(ctx)
-	w.watchedNamespaces[ns] = nsWatch{
+	w.watchedNamespaces[nsKey] = nsWatch{
 		watchers: map[watcherID]bool{watcherKey: true},
 		cancel:   cancel,
 	}
 
-	go w.dispatchPodChangesLoop(ctx, ch)
+	ownerFetcher := k8s.ProvideOwnerFetcher(ctx, kCli)
```
milas, line 332: I need to refactor this - we really need to share this with the event + service watchers so we don't do multiple owner lookups for everything.

My current thinking is is to add `OwnerFetcher OwnerFetcher()` to the `K8sClient` interface, but I'm open to suggestions.

internal/controllers/core/cluster/cache_fake.go
```diff
@@ -0,0 +1,70 @@
...
+import (
+	"sync"
+
+	"k8s.io/apimachinery/pkg/types"
+
+	"github.com/tilt-dev/tilt/internal/k8s"
+	"github.com/tilt-dev/tilt/pkg/apis/core/v1alpha1"
+)
+
+type FakeClientCache struct {
```
nicks, line 12: this might be a dumb question, but why not use the real cache? i feel like it exposes the same interface

internal/controllers/core/kubernetesdiscovery/reconciler.go
```diff
@@ -211,19 +219,25 @@ func (w *Reconciler) addOrReplace(ctx context.Context, key watcherID, kd *store.
...
+	if _, ok := w.watchers[watcherKey]; ok {
 		// if a watcher already exists, just tear it down and we'll set it up from scratch so that
 		// we don't have to diff a bunch of different pieces
-		w.teardown(key)
+		w.teardown(watcherKey)
+	}
+
+	kCli, err := w.clients.GetK8sClient(types.NamespacedName{Namespace: kd.Namespace, Name: kd.Spec.Cluster})
+	if err != nil {
+		return fmt.Errorf("cluster %q connection error: %v", kd.Spec.Cluster, err)
```
nicks, line 230: i don't think this should be treated as a reconciler error? we don't want to re-reconcile if this happens.

if anything, it feels like it should be some sort of waiting status

internal/controllers/core/kubernetesdiscovery/reconciler.go
nicks: i think the reconciler is going to need some per-cluster state, so that the state can be reset if the cluster resets. maybe put this there?

milas: For clarity, within the KD reconciler, this will only get re-created if we stop and then later restart the watch for a namespace.

The complication at the moment is that we use the `OwnerFetcher` elsewhere and want to share the same instance [for a given `K8sClient`] to make use of its internal caching/dedupe logic.

I moved ownership of the `OwnerFetcher` to the `K8sClient` interface in #5310 to do that. If we want to keep them separate, the alternative is to add it to the client cache interface afaict. I'm not against that, it just seems a bit less ergonomic for usage (tho perhaps cleaner from a separation of concerns perspective)

internal/controllers/core/kubernetesdiscovery/reconciler.go
milas: You're right, this shouldn't retry since it requires the `Cluster` to change which will implicitly trigger a re-reconciliation.

Agreed on status, I'll update the `KubernetesDiscoveryStatus` to add those fields in a new PR and pull changes back in here

internal/controllers/core/kubernetesdiscovery/reconciler.go
nicks: ya, i think there are two separate questions:
- sharing state from the same k8s cluster
- clearing state when the k8s cluster changes

For cases where you want both (1) and (2), i think it's reasonable to move the state into the k8s client. but i'd worry if we started moving non-shared state into the k8s client.

internal/controllers/core/cluster/cache_fake.go
milas: The cluster package's implementation of `ClientCache` is `ConnectionManager`, which has the full `connection` object/state to avoid having to do book-keeping 2x (state + clients), so it's not usable as a generic implementation. That's definitely refactorable though!

Maybe this should just be called `ClientMap` or something, which is a generic/naive impl used for tests and then the cluster reconciler can still `wire.Bind` its specialized version for prod 🤔

nicks: re: "book-keeping 2x (state + clients)", what bookkeeping are you talking about exactly?

milas: The client-or-error vs everything else in `connection`:
```go
type connection struct {
	connType     connectionType
	spec         v1alpha1.ClusterSpec
	dockerClient docker.Client
	k8sClient    k8s.Client
	error        string
	createdAt    time.Time
	arch         string
}
```

Currently, the `Cluster` reconciler uses the `ConnectionManager` type to maintain its internal state (latest processed spec, created at, arch) + the client, and since that implements/is wire-bound to `ClientCache` it lets consumers get the client out without leaking other the other details, which should be read via API.

The "fake" only stores the client-or-error since it's not tied to the reconciler. Again, name is a sliiiight misnomer since it doesn't really have a lot of specialized test code in it (besides the ability to pre-seed a default client)

I can definitely split it though and have something more like:
```
type Reconciler struct {
   ...
   clients ClientCache
   state    map[types.NamespacedName]connection
}
```

nicks: i think we're talking past each other. here's a commit that demonstrates what i'm suggesting:
https://github.com/tilt-dev/tilt/commit/c05bac10c9f09fc32f8566d7ef617079c24d7b64

milas: Ahhhh ok yes now I understand what you meant!

I'll pull that over 👍

nicks (approved):

milas:
@nicks Requested a new review on this as it ended up with substantial changes to cover the original gaps. (FYI I rebased because the diff between earlier revisions wouldn't have been particularly useful, but the non-test changes are much smaller, 70% of this PR is a verbose test going through the refresh sequence 😛)

nicks (approved):

internal/controllers/core/kubernetesdiscovery/keys.go
```diff
@@ -0,0 +1,58 @@
...
+)
+
+// nsKey is a tuple of Cluster metadata and the K8s namespace being watched.
+//
+// If multiple clusters are are in use, it's possible to watch the same namespace
+// in both, so the namespace name alone is not sufficient. Similarly, a Cluster
+// object can change, so there might be multiple revisions active at once.
+type nsKey struct {
+	cluster   clusterKey
+	namespace k8s.Namespace
```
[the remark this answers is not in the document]
milas: Works for me!

(We'll definitely need to do a pass around the way all the flawed default namespace logic interacts with some of the Cluster pieces, so this is a good incremental step)

internal/controllers/apis/cluster/client_fake.go
```diff
@@ -67,7 +67,7 @@ func (f *FakeClientProvider) AddK8sClient(key types.NamespacedName, client k8s.C
 	defer f.mu.Unlock()
 
 	if _, ok := f.clients[key]; !ok {
-		now := time.Now()
+		now := time.Now().Truncate(0)
```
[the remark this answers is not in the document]
milas: Good call - added in 081204a44

PR #5309, opened 2021-12-18
ospath: add path canonicalization utilities

nicks:


nicks:
cc @lizzthabet - this is mostly to play around with in your other PR to see if it helps

PR #5310, opened 2021-12-20
k8s: make OwnerFetcher part of the K8sClient

milas:
Historically, we've injected (via wire) a single Kubernetes client,
so we also injected a single `OwnerFetcher` (which uses the K8s
client).

Now that we're moving to multiple Kubernetes cluster connections
being supported, that means there can be multiple `OwnerFetcher`
instances, BUT we only want 1x `OwnerFetcher` per cluster connection.
(There is a lot of internal caching and clever logic to avoid duplicate
lookups!)

To avoid adding extra book-keeping/logic for consumers to look these
up in addition to the actual Kubernetes client, this moves the
`OwnerFetcher` to be managed by the client, which will ensure there's
a 1:1:1 mapping between cluster connection, Kubernetes cilent, and
`OwnerFetcher`.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -2,3 +2,4 @@
 
 *.sh text eol=lf
 
+*_gen.go linguist-generated=true -diff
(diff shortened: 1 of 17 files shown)


nicks (approved):

nicksieger (approved):
Nice refactoring. I especially like the narrowing of K8sClient to MetaClient dependency in the owner fetcher. 💯

PR #5319, opened 2021-12-21
web: reset the confirmation state of `ApiToggleButton`s when navigating between resources

lizzthabet:
This PR fixes a bug in which the `confirming` state of an `ApiToggleButton` persisted across rendering different disable resources buttons. If you clicked "disable resource" on one resource (without confirming or cancelling), then navigated to another resource, that confirmation state would not reset.

The easiest solution I found is to track the button name in the component's formal state and then calculate the actual `confirming` state in the render based on whether or not the current button name matches the tracked state button name.

At first, this was a little bit goofy for me to conceptualize with React. Because, if the props change (i.e., the button name changes), then we want to reset the confirmation state of the component. However, we can't make any `setState` calls within the body of the functional component (which leads to an infinite loop!). I believe the solution I've implemented aligns with React patterns, but if others have suggestions for another approach (or find this one too confusing), I'm open to it.

Closes [ch13124](https://app.shortcut.com/windmill/story/13124/apitogglebutton-s-confirmation-state-does-not-reset-when-navigating-between-resources-in-detail-view)

### DIFF
--- a/web/src/ApiButton.test.tsx
+++ b/web/src/ApiButton.test.tsx
@@ -2,7 +2,7 @@ import { Button, Icon, TextField } from "@material-ui/core"
 import { mount } from "enzyme"
 import fetchMock from "fetch-mock"
 import { SnackbarProvider } from "notistack"
-import React from "react"
+import React, { PropsWithChildren } from "react"
 import { act } from "react-dom/test-utils"
 import { MemoryRouter } from "react-router"
 import { AnalyticsAction } from "./analytics"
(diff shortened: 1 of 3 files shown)


web/src/ApiButton.test.tsx
```diff
@@ -37,18 +37,23 @@ const buttonInputsAccessor = accessorsForTesting(
...
     </MemoryRouter>
   )
 }
 
-function mountButton(b: UIButton) {
-  return wrappedMount(<ApiButton uiButton={b} />)
+function mountButton(b: UIButton, wrapperProps?: {}) {
+  return mount(<ApiButton uiButton={b} />, {
+    wrappingComponent: Wrapper,
+    wrappingComponentProps: wrapperProps,
```
lizzthabet, line 55: fun fact: today i learned about the `wrappingComponent` option in enzyme and it's so convenient! it means we can mount the component we care about and change its props (which can only be done on the component that `mount()` returns) _and_ wrap it in the providers we need it to be wrapped in. 🧠

web/src/ApiButton.tsx
```diff
@@ -521,29 +521,41 @@ function ApiSubmitButton(props: PropsWithChildren<ApiButtonElementProps>) {
...
+  }>(`apibutton-${buttonName}`, {})
   const { enqueueSnackbar } = useSnackbar()
   const pb = usePathBuilder()
   const { setError } = useHudErrorContext()
 
   const [loading, setLoading] = useState(false)
-  const [confirming, setConfirming] = useState(false)
+  // Record the confirmation state with the current button's name,
+  // so if the button name changes, the confirmation state can be reset
+  const [confirmingByButton, setConfirmingByButton] = useState({
```
nicks, line 536: i...don't think this is how state is intended to be used (where you use the initial value to communicate props changes)

have you tried something like: 

```
useEffect(() => setConfirming(false), [buttonName])
```

That will reset the confirmation state when buttonName changes, which i think is closer to what you want.

nicks: (one of my big frustrations with react is that it has tons of undefined/underdetermined behavior if you don't use state in exactly the way it expects)

lizzthabet: i think `useLayoutEffect` might work instead. i stayed away from `useEffect` because it runs after the render is painted and i thought any state changes were discouraged from it (which i can't find again in the react docs, so perhaps i made that up!) i'll update in a moment.

nicks (approved):

PR #5335, opened 2021-12-23
build: refactor so that the builders read image dependencies from the API

nicks:
Hello @landism,

Please review the following commits I made in branch nicks/imagemap7:

bd576bab9a0a18275c37f5e001a332f58e560759 (2021-12-23 16:16:24 -0500)
build: refactor so that the builders read image dependencies from the API
There should be no functional changes in this PR.
This is preparation for custom_build to handle image dependencies
the same way DockerImage handles image dependencies.

### DIFF
--- a/internal/engine/buildcontrol/docker_compose_build_and_deployer.go
+++ b/internal/engine/buildcontrol/docker_compose_build_and_deployer.go
@@ -8,6 +8,7 @@ import (
 	"github.com/docker/distribution/reference"
 	"github.com/docker/docker/api/types"
 	"github.com/docker/go-connections/nat"
+	ktypes "k8s.io/apimachinery/pkg/types"
 	ctrlclient "sigs.k8s.io/controller-runtime/pkg/client"
 
 	"github.com/tilt-dev/tilt/internal/analytics"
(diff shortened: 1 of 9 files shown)


internal/engine/buildcontrol/docker_compose_build_and_deployer.go
```diff
@@ -138,18 +141,31 @@ func (bd *DockerComposeBuildAndDeployer) BuildAndDeploy(ctx context.Context, st
 		ps.EndPipelineStep(ctx)
 	}
 
-	iTargetMap := model.ImageTargetsByID(plan.tiltManagedImageTargets)
+	var cluster v1alpha1.Cluster
+	// If the cluster fetch fails, that's OK.
```
landism, line 145: FWIW, I found this confusing:
* if the cluster is an input to the build, does a failure here means we're building with incorrect parameters?
* if it's not, why call `Get` rather than passing a default Cluster?

landism (approved):

internal/engine/buildcontrol/docker_compose_build_and_deployer.go
nicks: ya, that's fair. i removed it.

in theory, you could have a DockerCompose cluster with a different arch than your host cluster, but it seems unlikely to happen in practice.

this will all go away once we move these into a reconciler, since the image builder won't know if it's building for docker compose or k8s

PR #5339, opened 2022-01-03
controllers: make the Cmd controller level-triggered rather than edge triggered

nicks:
Hello @milas, @landism,

Please review the following commits I made in branch nicks/cmd:

4411d4315bd14ed1ffa7e1ac9801be30ae4330e6 (2022-01-03 16:02:28 -0500)
controllers: make the Cmd controller level-triggered rather than edge triggered
Before this PR, the controller would listen for process
changes and update the apiserver on each change.

After this PR, the controller does all bookkeeping about the state of the
process in local memory, then updates the apiserver only during reconciliation.

This should be way more robust and eliminate a lot of race conditions.

### DIFF
--- a/internal/controllers/core/cmd/controller.go
+++ b/internal/controllers/core/cmd/controller.go
@@ -23,6 +23,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/source"
 
+	"github.com/tilt-dev/tilt/internal/controllers/apicmp"
 	"github.com/tilt-dev/tilt/internal/controllers/apis/configmap"
 	"github.com/tilt-dev/tilt/internal/controllers/apis/restarton"
 	"github.com/tilt-dev/tilt/internal/controllers/indexer"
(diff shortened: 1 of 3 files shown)


internal/controllers/core/cmd/controller.go
```diff
@@ -413,14 +431,20 @@ func (c *Controller) handleProbeResultFunc(ctx context.Context, name types.Names
...
-			// 	a terminated process to ready, for example; in practice, we
-			// 	should update internal state on any goroutine/async trackers
-			// 	and trigger a reconciliation, which can then evaluate the full
-			// 	state + current spec
+
+		// TODO(milas): this isn't quite right - we might end up setting
+		// 	a terminated process to ready, for example; in practice, we
+		// 	should update internal state on any goroutine/async trackers
+		// 	and trigger a reconciliation, which can then evaluate the full
+		// 	state + current spec
```
milas, line 439: I think we can resolve this now!

1. Check that `status.Running != nil` before setting `Ready` here
2. In `processStatuses`, for the `Running` branch, flip the `mutateStatus` and `initProbeWorker.Do` order call (so running is populated before probe starts so that the check here will work)

milas: (I can also do it in a PR later if you want to limit scope here)

milas (approved):

internal/controllers/core/cmd/controller.go
nicks: let's do this in a separate PR, since i think this is going to be more controversial

PR #5340, opened 2022-01-04
api: add ConfigHash to Cluster status

milas:
Consumers will use this to detect when the actual connection/client
has changed since an identical spec could produce an entirely
different connection if the underlying kubeconfig (or associated
files) change on disk.


milas:
See #5341 for how this will be used

pkg/apis/core/v1alpha1/cluster_types.go
```diff
@@ -158,6 +158,13 @@ type ClusterStatus struct {
 	//
 	// +optional
 	Error string `json:"error,omitempty" protobuf:"bytes,2,opt,name=error"`
+
+	// ConfigHash represents the state of the current cluster client config, so
```
nicks, line 162: have you considered using a ConnectedAt timestamp?

it feels a bit weird to expose a Hash in a public API (i know i did this in KubernetesApply but i kind of regretted it)

could also have a Connected metav1 condition, so that people can do `tilt wait --for=condition/Connected cluster/default`, though note that this isn't mutually exclusive (i think the other k8s api love to duplicate things in metav1.Condition)

nicks (approved):

pkg/apis/core/v1alpha1/cluster_types.go
milas: I did not! That could work as well and does solve other issues (namely that my hashing in #5342 is fairly arbitrary 😬).

Sidenote: I was thinking about starting to stamp the value (whatever it is) onto referencing objects via an annotation, e.g. `tilt.dev/active-cluster-config` which would be a helpful way to sanity check if/when things go awry. WDYT?

nicks: i think the k8s community has been moving towards using generation numbers for this (whereas they used to use hashes like you proposed above). here's some good docs on this:

https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#label-selector-and-annotation-conventions
https://alenkacz.medium.com/kubernetes-operator-best-practices-implementing-observedgeneration-250728868792 

(which even uses a Kind cluster as an example of when you might want to use this HA :grinning: )

milas: I've seen the generation bits, and it's definitely a useful pattern we should look into to avoid storing full specs all over the place 😬 

I _think_ in this case it'll be better to rely on the timestamp -- I'm concerned about unrelated updates triggering a cascade of refreshes TBH.

The `Cluster` object is odd/special - the spec/status alone aren't sufficient to determine equality but the cost of unnecessarily "cache-busting" is extremely high, so IMO relying on a specific, controlled field, i..e `ConnectedAt`, is the safest option.

nicks (approved):

PR #5341, opened 2022-01-04
cluster: version-aware client manager

milas:
This is an attempt at a middle-ground that doesn't totally sidestep
the apiserver but provides reasonable ergonomics to handle client
changes at runtime.

We don't want to wipe state if the `Cluster` changes at all because
that will inevitably result in unnecessarily rebuilding state, which
can be an expensive operation. Furthermore, the spec also (by nature)
isn't adequate - it is primarily a context/namespace tuple, with the
actual config being loaded as part of boostrapping the K8s client.

The way this is now broken down:
 * Cluster reconciler internally uses `ConnectionManager` to track
   its state including clients + their config hashes
 * Consumers use `ClientManager` which allows getting a client for
   a given `Cluster` object and handles "versioning" in a reconciler
   friendly way
    * Internally, this defers to a `ClientProvider` interface to
      handle refreshing on-demand (i.e. compare known version for
      `Cluster` object to canonical version); the runtime impl for
      this is `ConnectionManager` while a `FakeClientProvider` is
      offered for tests so they don't need to run the actual
      `Cluster` reconciler


milas:
This builds on #5340 and #5342 - I did some super quick rebasing to pull the easy parts into separate PRs for ease of review - if this is still a PITA I can split it further, but hopefully it's straightforward!

milas:
This is ready for review - rebased/refactored to use `ConnectedAt`.

Note that nothing actually uses this functionality as of this PR - consumers will be moved over individually in follow-on PRs.

milas:
See also: #5349 for the first consumers of this (the `k8swatch` engine subscribers for events/services)

nicks (approved):

PR #5342, opened 2022-01-04
api: set Cluster client config hash on status

milas:
Set a hash for the Cluster connection based on the client config
so that consumers have a single value to determine whether their
client is current based on the Cluster status.


milas:
Most likely going to change this to be a `ConnectedAt` timestamp instead of a hash - closing for now

PR #5343, opened 2022-01-04
api: report ConnectedAt on Cluster status

milas:
See #5340 for API change. The data here was already tracked
internally, so this just applies it to the `Cluster` status object.

There's some tweaks to the test assertion helper for API time types
so that it can handle `nil` types, as used in the `Cluster` status
`ConnectedAt` field.

### DIFF
--- a/internal/timecmp/assert.go
+++ b/internal/timecmp/assert.go
@@ -2,6 +2,7 @@ package timecmp
 
 import (
 	"fmt"
+	"reflect"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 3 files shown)


internal/timecmp/assert.go
```diff
@@ -16,12 +17,27 @@ type stringableTimeValue interface {
...
+		if !equal {
+			assert.Fail(t, fmt.Sprintf("Not equal: \n"+
+				"expected: %s\n"+
+				"actual  : %s", expected, actual))
+		}
+	}()
+
+	expectedNil := isNilValue(expected)
+	actualNil := isNilValue(actual)
+	if expectedNil && actualNil {
```
nicks, line 33: `expectedNil == actualNil`?

nicks (approved):

internal/timecmp/assert.go
milas: Hmmm, I don't think so? If they're both non-nil, that'd be `true`, in which case we want to proceed with the `Equal` call that will normalize them by granularity before comparison.

FWIW I'm actually about to open another PR that will make the real comparison functions work with `nil`, at which point this code will go away.

nicks: oops, you're right, i misread smth

PR #5344, opened 2022-01-04
tiltfile: include name of crd when image_json_path doesn't match

### LINKED ISSUE #5326: When k8s_kind's image_json_path doesn't exist, Tilt's error message doesn't tell you where.
### Expected Behavior
When a k8s_kind matches a CRD instance and specifies an image_json_path that doesn't exist in that CRD instance, the error message tells you which object had the issue.

### Current Behavior
```
Loading Tiltfile at: /Users/matt/go/src/github.com/tilt-dev/tilt/integration/crd/Tiltfile
Matching (json_path="{.spec.image}"): image is not found
```

### Steps to Reproduce
Define a k8s_kind w/ an image_json_path, and two CRD instances, only one of which actually has that image_json_path
or:
check out https://github.com/tilt-dev/tilt/tree/matt/missing_image_json_path
`cd integration/crd && tilt up`

#### About Your Use Case
A user reported this. They had a number of CRDs, some of which had an image at the specified image_json_path and some of which didn't. They spent nontrivial time trying to get their config right, thinking that a CRD instance w/ a valid image was causing the error, not realizing it was one of the others. If the error message specified which instance caused the error, that time would have been saved.

landism:
Fixes #5326

Old:
```
Loading Tiltfile at: /Users/matt/go/src/github.com/tilt-dev/tilt/integration/crd/Tiltfile
Matching (json_path="{.spec.image}"): image is not found
```

New:
```
Loading Tiltfile at: /Users/matt/go/src/github.com/tilt-dev/tilt/integration/crd/Tiltfile
Finding image in UselessMachine/bobo2: Matching (json_path="{.spec.image}"): image is not found
```

### DIFF
--- a/internal/tiltfile/k8s.go
+++ b/internal/tiltfile/k8s.go
@@ -129,7 +129,7 @@ func (r *k8sResource) addEntities(entities []k8s.K8sEntity,
 	for _, entity := range entities {
 		images, err := entity.FindImages(locators, envVarImages)
 		if err != nil {
-			return err
+			return errors.Wrapf(err, "finding image in %s/%s", entity.GVK().Kind, entity.Name())
 		}
 		for _, image := range images {
 			r.addImageDep(image, false)
--- a/internal/tiltfile/tiltfile_test.go
+++ b/internal/tiltfile/tiltfile_test.go
@@ -2500,6 +2500,25 @@ docker_build('tilt.dev/frontend', '.')
 		m.ImageTargets[0].Refs.LocalRef().String())
 }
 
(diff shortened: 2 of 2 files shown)


nicks (approved):

esselius:
This helped me out a lot today, thanks! :)

Apparently the first arg of `k8s_kind` is actually a regexp pattern and the pattern `KafkaConnect` was also matching `KafkaConnector` of the strimzi kafka operator, upgrading tilt 0.22.15 -> 0.25.1 and getting this helpful change enabled me to actually see the problem and fix it! :)

PR #5345, opened 2022-01-04
tiltfile: report enabled features

landism:
to allow us to track how many users have opted into experimental features

### DIFF
--- a/internal/engine/analytics/analytics_reporter.go
+++ b/internal/engine/analytics/analytics_reporter.go
@@ -2,11 +2,13 @@ package analytics
 
 import (
 	"context"
+	"fmt"
 	"strconv"
 	"time"
 
 	"github.com/tilt-dev/tilt/internal/analytics"
 	"github.com/tilt-dev/tilt/internal/controllers/apis/liveupdate"
+	"github.com/tilt-dev/tilt/internal/feature"
 	"github.com/tilt-dev/tilt/internal/k8s"
 	"github.com/tilt-dev/tilt/internal/store"
 )
@@ -15,11 +17,12 @@ import (
 const analyticsReportingInterval = time.Minute * 15
 
 type AnalyticsReporter struct {
-	a       *analytics.TiltAnalytics
-	store   store.RStore
-	kClient k8s.Client
-	env     k8s.Env
-	started bool
+	a               *analytics.TiltAnalytics
+	store           store.RStore
+	kClient         k8s.Client
+	env             k8s.Env
+	featureDefaults feature.Defaults
+	started         bool
 }
 
 func (ar *AnalyticsReporter) OnChange(ctx context.Context, st store.RStore, _ store.ChangeSummary) error {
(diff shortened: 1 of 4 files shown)


nicks:
quick q: is this going to report a lot of obsolete flags that we haven't cleaned up?

internal/tiltfile/tiltfile.go
```diff
@@ -243,6 +246,12 @@ func (tfl *tiltfileLoader) reportTiltfileLoaded(
 			tags[fmt.Sprintf("tiltfile.invoked.%s.arg.%s", builtinName, argName)] = strconv.Itoa(count)
 		}
 	}
+	for k, v := range features {
+		if v.Status == feature.Active && v.Enabled {
```
nicks, line 250: feels like we should only report this when it's different than the default (if the intention is to only report when people have opted in)

landism: I briefly considered that, but wasn't able to come up with any use cases for it. My thinking is that for questions like "of users who have this feature enabled, how many have interacted with it?", this saves us the step of checking the version number, at relatively close to launch. Once we change the flag to obsolete, we have to check the version number anyway, so maybe it doesn't buy us that much.

nicks: the pr description says "to track how many users have opted into experimental features"...so are you saying that's outdated?

if you're trying to answer the question "of users who have this feature enabled, how many have interacted with it?", it seems like this would make that harder? you'd have to do some aggregations and joins of ​the "interacted" event against the "tiltfile.loaded" event, right?

a pattern i've seen is to create a manually-curated fact table for which feature launched with which version, then use that? seems like it'd be simpler, since we already ship a version on every event. unless i'm missing something?

lizzthabet: i could see a version table with feature list being useful in the future!

fwiw, i think including a list of enabled features on tilt up (regardless of explicit user opt-in) is fine and useful. it would allow us to track how many people have access to a feature, whether they explicitly opted into it or it's turned on by default. 

this current approach may make it more complicated if we did want to analyze how many users opted into a specific feature before it was turned on by default.

though, if we only counted explicit opt-ins, i could see instances where a user opts into a feature, then upgrades tilt to a version where that's turned on by default, but they don't remove that call from their `Tiltfile`, so analyzing explicit opt-ins would still require version comparison. 

(hopefully this isn't too much bike-shedding. i'm outlining specific cases that i'm thinking through and how this might be useful for us!)

just for reference, here's an example event payload from my local copy of servantes, where i have labels and disable resources explicitly enabled:
```json
  ...
  "tiltfile.feature.disable_resources": "true",
  "tiltfile.feature.labels": "true",
  "tiltfile.feature.live_update_v2": "true",
  "tiltfile.feature.snapshots": "true",
  ...
  "tiltfile.invoked.enable_feature": "2",
  "tiltfile.invoked.enable_feature.arg.msg": "2",
```

nicks: i would be fine with adding feature tags to the 'up.running' event as in here:
https://github.com/tilt-dev/tilt/blob/master/internal/engine/analytics/analytics_reporter.go#L148
which is at as least consistent with existing patterns (i.e., the feature is a property of a running tilt instance rather than a property of the tiltfile load) and should be easier to query correctly

landism:
I've moved the reporting to AnalyticsReporter. I'll put up a separate PR to adjust features' obsolete status.

internal/engine/analytics/analytics_reporter.go
```diff
@@ -157,5 +162,11 @@ func (ar *AnalyticsReporter) report(ctx context.Context) {
 
 	stats["tiltfile.error"] = tiltfileIsInError
 
+	for k, v := range st.Features {
+		if ar.featureDefaults[k].Status == feature.Active && v != ar.featureDefaults[k].Enabled {
```
nicks, line 166: wait, i'm so confused, i thought @lizzthabet wanted us to have the tag attached even if the default was enabled?

landism: whoops, right!

nicks (approved):

PR #5352, opened 2022-01-05
api: make ClientManager resource aware

milas:
There's numerous competing concerns with managing K8s clients/their
operations, and a big one is that for efficiency, we only ever want
to set up a namespace-level watch for a specific cluster once and
then triage the incoming data.

In particular, this worked well with the engine subscribers, as
they'd always evaluate global state.

However, for reconcilers, this is trickier, as we operate on
individual objects. Behind the scenes, the reconciler tracks the
consumers for a given watch, so that it can de-duplicate and perform
refcounting to know when it's safe to cancel.

This complicates client refreshes, as if the underlying K8s client
for the `Cluster` object gets changed, we need to reset state for
each referencing object individually when its reconcile happens.

The changes here actually reduce the `ClientManager`'s surface area
by having it act purely as a coordinator on top of the
`ClientProvider` rather than an extra level of caching (i.e. it
no longer stores pointers to the clients themselves but always
delegates reads back to the provider).

As this coordination is now on a per-object basis, there's a slight
hack in the engine subscribers to use a single, virtual object for
keying so that it can continue acting on a global basis.

### DIFF
--- a/internal/controllers/apis/cluster/client.go
+++ b/internal/controllers/apis/cluster/client.go
@@ -26,11 +26,11 @@ var NotFoundError = errors.New("cluster client does not exist")
 // also ensuring that callers handle connection changes uniformly.
 //
 // See the expected usage/flow on ClientManager.
-var StaleClientError = errors.New("cluster hash is stale")
+var StaleClientError = errors.New("cluster revision is stale")
 
-// ClientProvider provides cached clients to the ClientManager.
+// ClientProvider provides client instances to the ClientManager.
 //
-// All clients are goroutine-safe.
+// All clients MUST be goroutine-safe.
 type ClientProvider interface {
 	// GetK8sClient returns the Kubernetes client for the cluster or an error for unknown clusters, connections
 	// in a transient error state, or if the connection is of a different type (i.e. Docker Compose).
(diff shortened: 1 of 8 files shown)


milas:
I thought I could get the original design to work with the reconcilers but it was getting really icky - I've got this working well with the `KubernetesDiscovery` reconciler now, so this **should** be it for refactoring this client refresh logic 🙄

nicks (approved):
LGTM

i think it's important to get this working end-to-end ASAP, so i think this is fine for now.

but i'm concerned that this API has a lot of implicit assumptions around state, and that it will be very difficult to use correctly in practice. 

i'm also not sure this approach will work in the case where you can change the "cluster" field on an API object, due to the multiplicative complexity

milas:
Yeah, I'm not totally thrilled with this either (in case that wasn't evident by the myriad of iterations it's gone through), so if you have specific suggestions re: design, I'm all ears.

My hope was to simplify + standardize this across the reconcilers since there's a lot of tedious, error-prone considerations to handling the cluster updates, so I definitely don't want to introduce a fragile API for that 😧 

RE: changing `.spec.cluster` on obj, you're right that it's imperfect - in practice it should probably track `(clusterNN, revisionTimestamp)` rather than just `revisionTimestamp`, but since it's not feasible for 2x different clusters to end up with the same exact timestamp, I believe it'll still work in its current state. I made note of it and will revisit + make sure there's test coverage for that.

nicks:
haha exactly - i'm not sure i have a better idea. onwards!

PR #5359, opened 2022-01-07
api: add retry + cleanup to PodSource

milas:
Currently, if the `PodSource` fails to watch a namespace, an error
will be logged, but there's no way to get out of that state because
the watch remains, so subsequent attempts to watch will short circuit.

Now, the watch logic uses a retry loop with backoff. Errors are logged
but still not propagated to `PodLogStream` objects currently. In
practice, this should be mostly fine because if the K8s client is
inoperable, the reconciler will also fail when it tries to actually
set up the log streaming. In the future, it'd probably be more robust
to track the error state and propagate it via `handleReconcileRequest`
so that the reconciler can mark the `PodLogStream` status accordingly
and then requeue itself.

Additionally, if there's no more watchers for a namespace, the watch
will be canceled. This will become more important with multi-cluster
work to ensure we dispose of resources for old clients.

### DIFF
--- a/internal/controllers/core/podlogstream/podsource.go
+++ b/internal/controllers/core/podlogstream/podsource.go
@@ -3,6 +3,7 @@ package podlogstream
 import (
 	"context"
 	"sync"
+	"time"
 
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
@@ -18,6 +19,8 @@ import (
 	"github.com/tilt-dev/tilt/pkg/apis/core/v1alpha1"
 	"github.com/tilt-dev/tilt/pkg/logger"
 )
+
+const maxRestartBackoff = 5 * time.Minute
 
 var podGVK = schema.GroupVersionKind{Version: "v1", Kind: "Pod"}
 
@@ -38,6 +41,7 @@ type podWatch struct {
 	ctx       context.Context
 	cancel    func()
 	namespace string
+	watchers  map[types.NamespacedName]bool
 }
 
 var _ source.Source = &PodSource{}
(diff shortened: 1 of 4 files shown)


internal/controllers/core/podlogstream/podsource.go
```diff
@@ -72,45 +80,101 @@ func (s *PodSource) TearDown() {
...
+			s.maybeCleanupWatch(nsKey)
 		}
 	}
 }
 
-// Process pod events and make sure they trigger a reconcile.
-func (s *PodSource) doWatch(pw podWatch) {
-	defer pw.cancel()
+// maybeCleanupWatch removes a namespace watch if there are no more active watchers.
+func (s *PodSource) maybeCleanupWatch(nsKey string) {
```
nicks, line 116: i'm a little bit worried about the flow:
- user restarts a pod
- pod is deleted
- watches are deleted
- pods are created
- watches are recreated

and wonder if we should have some sort of debouncer on the cleanup (e.g., only debounce if it's been unused for 10s)

maybe this is a premature optimization though and it's not worth the extra complexity.

wdyt?

internal/controllers/core/podlogstream/podsource.go
```diff
@@ -72,45 +80,101 @@ func (s *PodSource) TearDown() {
...
+	if len(watch.watchers) != 0 {
+		// there's still active watchers
 		return
 	}
 
+	watch.cancel()
+	delete(s.watchesByNamespace, nsKey)
+}
+
+func (s *PodSource) doWatchWithRetry(pw podWatch) {
```
nicks, line 131: oof, how far are you down this approach?

i'm not a fan of having an internal restart loop like this, rather than letting things bubble up to the main reconciler and let the reconciler handle the retry (so that it can also publish the error state to the api)

i have a branch that i think might handle this (in the spirit of the other controller-cleanup prs)

nicks (approved):

internal/controllers/core/podlogstream/podsource.go
milas: I can back this out! I'll pause on PLS changes for now while you get those changes in.

internal/controllers/core/podlogstream/podsource.go
milas: I had the same thought - I can add that, I don't think it'll be a ton of extra code.

(I think in most cases you won't end up hitting that because you'll still have at least one other PLS from a different resource holding the namespace watch open, but it still probably merits avoiding it if possible)

milas:
#5361 covers the important retry bits here - will add cleanup if necessary in a new PR when doing `Cluster` work

PR #5360, opened 2022-01-07
features: update some feature flags to obsolete

landism:


internal/feature/flags.go
```diff
@@ -50,19 +50,19 @@ var MainDefaults = Defaults{
...
+		Status:  Obsolete,
 	},
 	Facets: Value{
 		Enabled: true,
 		Status:  Obsolete,
 	},
 	Labels: Value{
 		Enabled: true,
-		Status:  Active,
+		Status:  Obsolete,
```
lizzthabet, line 65: we're still referencing snapshots and labels in the frontend, so i think we should remove those references before marking the flags as obsolete. 

snapshots is only referenced in one line, but label logic is probably more complex and repeated in a couple places. if you want to remove the snapshot logic here and leave the label flag as-is, i could tackle the label logic is a separate pr.

nicks (approved):

internal/feature/flags.go
landism: > we should remove those references before marking the flags as obsolete

My understanding is the opposite! This PR does not remove the flags--it just makes them uneditable from the Tiltfile. If we remove the frontend code before making this change, then users who have edited the flags via the Tiltfile will have their enable/disable_feature calls silently ignored. Once we've marked the flags obsolete, then we can remove the flag checks in the frontend, assured that all users have the feature enabled.

lizzthabet: gotcha, ok! my interpretation was that "obsolete" means that we're no longer referencing this flag in our codebase, but that's not accurate and your explanation makes sense.

PR #5362, opened 2022-01-07
api: track Cluster connection events as a metric

milas:
Report a metric about Cluster connections. The primary goal here
is to allow us to determine how often users are running into
Cluster connection issues. Currently, there's no granularity beyond
successful or in an error state.

### DIFF
--- a/internal/controllers/core/cluster/reconciler.go
+++ b/internal/controllers/core/cluster/reconciler.go
@@ -13,6 +13,7 @@ import (
 	ctrlclient "sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
+	"github.com/tilt-dev/tilt/internal/analytics"
 	"github.com/tilt-dev/tilt/internal/controllers/apicmp"
 	"github.com/tilt-dev/tilt/internal/docker"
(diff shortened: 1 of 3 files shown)


internal/controllers/core/cluster/reconciler.go
```diff
@@ -103,6 +104,10 @@ func (r *Reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
 		}
 	}
 
+	if !hasConnection {
+		r.reportConnectionEvent(ctx, conn)
```
nicks, line 108: this is going to report on every re-reconcile.

can you put this in `maybeUpdateStatus` and only update on state change?

internal/controllers/core/cluster/reconciler.go
```diff
@@ -103,6 +104,10 @@ func (r *Reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
 		}
 	}
 
+	if !hasConnection {
```
nicksieger, line 107: Maybe rather than reporting on `if !hasConnection` you track a separate boolean `connectionChanged` and report on when that is true?

milas:
Moved the report into `maybeUpdateStatus` in ee348ec to ensure it only happens on status change.

PR #5363, opened 2022-01-07
docker: track connect events as a metric

milas:
Similar to K8s cluster connections from #5362, report Docker client
connections via analytics.

The goal here is to determine how frequently users are running into
Docker issues at startup.

### DIFF
--- a/internal/docker/client.go
+++ b/internal/docker/client.go
@@ -32,6 +32,7 @@ import (
 	"github.com/moby/buildkit/session/filesync"
 	"github.com/pkg/errors"
 
+	"github.com/tilt-dev/tilt/internal/analytics"
 	"github.com/tilt-dev/tilt/internal/container"
 	"github.com/tilt-dev/tilt/internal/docker/buildkit"
 	"github.com/tilt-dev/tilt/pkg/logger"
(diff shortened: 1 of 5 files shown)


milas:
I'll be honest, this feels VERY hacky, so if someone's got ideas for a better place to handle this, I'll happily refactor it 🙈

internal/docker/client.go
```diff
@@ -134,32 +135,45 @@ type Cli struct {
 }
 
 func NewDockerClient(ctx context.Context, env Env) Client {
+	var failureReason string
+	var serverVersion types.Version
+	defer func() {
+		reportDockerClientConnect(ctx, failureReason, env, serverVersion)
```
nicks, line 141: i understand what you're trying to do here, but this is going to give you bogus and unreliable metrics.

in particular, lots of people 
- use tilt with pure local services and docker turned off
- use tilt with minikube's docker socket (and no local docker socket)

the reason we have to use the exploding client pattern here is because we don't know until much later whether we're connecting to the right docker socket.

i think this logic probably needs to live in the tiltfile reconciler (see the SetOrchestrator call, which is where we determine which docker socket is the right one to connect to)

milas:
See #5369.

PR #5366, opened 2022-01-10
Introduces a generic Tiltfile to get first time Tilters started

lianmakesthings:
Hi @nicksieger @milas
could you please review the following changes:
- Added a generic Tiltfile to quickly get users started with their first Tiltfile


Tiltfile.starter
```diff
@@ -0,0 +1,31 @@
...
+# Happy Tilting!
+
+
+# use a custom remote registry
+# more info on https://docs.tilt.dev/personal_registry.html
+# default_registry('gcr.io/my-public-registry')
+
+# run a one-time command e.g. to customize your cluster
+# more info on https://docs.tilt.dev/local_resource.html
+# local_resource('my-command', cmd='./scripts/my-command.sh', deps=['scripts/my-command.sh'], allow_parallel=True, auto_init=False)
```
nicksieger, line 13: I was trying to think of useful cross-platform local commands to run, but I'm coming up empty :). I think these are good for now.

Tiltfile.starter
```diff
@@ -0,0 +1,31 @@
...
+
+# build docker image for a resource
+# more info on https://docs.tilt.dev/api.html#api.docker_build
+# docker_build('my-resource', '.', dockerfile='deploy/docs.dockerfile',
```
nicksieger, line 17:
```suggestion
# docker_build('my-resource-image-name', '.', dockerfile='deploy/docs.dockerfile',
```

Tiltfile.starter
nicksieger: ❓ 
```suggestion
local_resource('hello-local', cmd='echo Hello World')
```

nicksieger:
Adding this snippet here for completeness. We could include some code to run tilt-avatars from the default Tiltfile:

```starlark
# You can package related Starlark statements in a function. Here we
# add a local resource that downloads and loads the tilt-avatars demo app.
def tilt_demo():
    if os.path.exists('tilt-avatars-main'):
        load_dynamic('tilt-avatars-main/Tiltfile')
    watch_file('tilt-avatars-main/Tiltfile')
    url      = 'https://github.com/tilt-dev/tilt-avatars/archive/refs/heads/main'
    curl_cmd = 'curl -L {}.tar.gz | tar zxvf -'.format(url)
    ps_cmd   = ('PowerShell -Command "Invoke-WebRequest -Uri {}.zip -UseBasicParsing -OutFile tilt-avatars-main.zip; ' +
                'Expand-Archive tilt-avatars-main.zip -DestinationPath .; Remove-Item tilt-avatars-main.zip').format(url)
    local_resource('get-tilt-avatars', cmd=curl_cmd, cmd_bat=ps_cmd, labels=['demo'])

# Uncomment the tilt_demo() invocation to see it in action.
#tilt_demo()
```

Tiltfile.starter
```diff
@@ -0,0 +1,31 @@
+# Welcome to Tilt!
+# To get you started as quickly as possible, we have created this generic Tiltfile. 
+# Modify and comment out any commands to make them work for you.
+# Happy Tilting!
+
+
+# use a custom remote registry
+# more info on https://docs.tilt.dev/personal_registry.html
+# default_registry('gcr.io/my-public-registry')
```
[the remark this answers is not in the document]
lianmakesthings: Ya, I think you're right. I think our assumption was that we are dealing with the very basic case, so we would assume the user has something like Docker desktop, minikube or kind set up, for which they wouldn't need a remote registry. Will remove

Tiltfile.starter
lianmakesthings: One idea I had: install some cluster tooling that is used on prod, and you want to create the same conditions on your local cluster, for example OPA Gatekeeper

Tiltfile.starter
```diff
@@ -0,0 +1,55 @@
+# Welcome to Tilt!
+# To get you started as quickly as possible, we have created this generic Tiltfile. 
+# Modify and comment out any commands to make them work for you.
+# Happy Tilting!
+print("Hello Tilt! This appears in the (Tiltfile) pane whenever Tilt evaluates this file.")
+
+# Run a one-time command e.g. to install prerequisites
+# More info: https://docs.tilt.dev/local_resource.html
+# `cmd_bat`, when present, is used instead of `cmd` on Windows.
```
nicksieger, line 9:
```suggestion
# `cmd_bat`, when present, is used instead of `cmd` on Windows.
# More info: https://docs.tilt.dev/local_resource.html
```

milas (approved):
(lol I merged my changes from #5383 into this so it feels weird to approve but let's goooo :shipit: )

milas:
@lianmakesthings I think you're good to merge this! (at which point I'll rebase #5380 and merge it as well)

PR #5375, opened 2022-01-12
Search docs from UI

lianmakesthings:
Hi @hyu @SurbhiSGupta @lizzthabet 

could you please review the following changes:

- renamed `ShortcutsDialog` to `HelpDialog`
- restyled `FloatDialog` everywhere in the Tilt UI
- replaced button to docs with search bar in `HelpDialog`
- added `utm_source` param to search request from `HelpDialog` for setting up metrics in GA
- added tests

Screen Recording 2022-01-14 at 21 32 02 mov


SurbhiSGupta (approved):

web/src/HelpSearchBar.tsx
```diff
@@ -0,0 +1,114 @@
...
+} from "./style-helpers"
+
+export function searchDocs(query: string) {
+  window.open(
+    `https://docs.tilt.dev/search?q=${encodeURI(query)}&utm_source=tiltui`
+  )
+}
+
+export const HelpSearchBarTextField = styled(InstrumentedTextField)`
+  & .MuiOutlinedInput-root {
```
lizzthabet, line 26: (optional design suggestion) wdyt about giving the text field a `100%` width? that way, a user has more room to see their text input if it's a little longer. cc: @SurbhiSGupta 

Screen Shot 2022-01-18 at 12 05 22 PM

web/src/HelpSearchBar.tsx
```diff
@@ -0,0 +1,114 @@
...
+  Font,
+  FontSize,
+  mixinResetButtonStyle,
+  SizeUnit,
+} from "./style-helpers"
+
+export function searchDocs(query: string) {
+  window.open(
+    `https://docs.tilt.dev/search?q=${encodeURI(query)}&utm_source=tiltui`
+  )
```
lizzthabet, line 22: let's replace the `encodeURI` call with the `URL` constructor, so the query string will be formatted correctly! this is my bad for making the wrong suggestion when we were pairing. 😅 

```suggestion
  const docsSearchUrl = new URL("https://docs.tilt.dev/search")
  docsSearchUrl.searchParams.set("q", query)
  docsSearchUrl.searchParams.set("utm_source", "tiltui")
  window.open(docsSearchUrl)
```

lizzthabet (commented):
this looks great! 🎉 i noted two minor things and then had two suggestions for react best practices and changing how the search url is formatted.

web/src/HelpSearchBar.tsx
```diff
@@ -0,0 +1,114 @@
...
+  }
+`
+
+export const ClearHelpSearchBarButton = styled(InstrumentedButton)`
+  ${mixinResetButtonStyle};
+  display: flex;
+  align-items: center;
+`
+export let searchValue: string
+let setSearchValue: React.Dispatch<React.SetStateAction<string>>
```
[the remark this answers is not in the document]
lianmakesthings: Yes, that's precisely why. Would definitely appreciate the alternative approach because I don't feel good about this 😅

web/src/HelpSearchBar.tsx
```diff
@@ -0,0 +1,114 @@
...
+
+  return (
+    <HelpSearchBarTextField
+      aria-label="Search Tilt Docs"
+      className={props.className}
+      value={searchValue}
+      placeholder="Search Tilt Docs..."
+      InputProps={inputProps}
+      variant="outlined"
+      analyticsName="ui.web.HelpSearchBar"
```
lianmakesthings, line 110:
```suggestion
      analyticsName="ui.web.helpSearchBar"
```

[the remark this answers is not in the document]
lianmakesthings: That's that the UTM_source param is for so we can look for that in GA

web/src/HelpSearchBar.tsx
SurbhiSGupta: Visually the close search bar icon X may be too close to the dialogue close icon. Actual interaction wise, they may not conflict at all !

web/src/HelpSearchBar.tsx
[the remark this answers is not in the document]
lizzthabet: oh, doih! 🤦🏻‍♂️ just checking to make sure that we don't need to also track those analytics here.

web/src/HelpSearchBar.tsx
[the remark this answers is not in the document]
lizzthabet: for sure! there are two ways you could do this:

```ts
// Grab the `input` DOM node and check its `value` property
const searchFieldAfterInput = root.find("input").at(0)
expect((searchFieldAfterInput.getDOMNode() as HTMLInputElement).value).toEqual(searchTerm)

// Grab the text field component and check its `value` prop
const searchComponentAfterInput = root.find(HelpSearchBarTextField).at(0)
expect(searchComponentAfterInput.prop("value")).toEqual(searchTerm)
```

for both approaches, you need to `find` the elements in the wrapper that you want to access after the component has updated. i've forgotten about this in the past. 😅 

```ts
// Fetch the element
const searchField = root.find("input")
// Make the change
searchField.simulate("change", { target: { value: searchTerm } })
// To check something in the component affected by the change, fetch it again
const searchFieldAfterInput = root.find("input")
```

(i usually have the [enzyme docs](https://enzymejs.github.io/enzyme/docs/api/mount.html) bookmarked because i forget all of the helpers it has to access component / node properties.)

web/src/HelpSearchBar.tsx
lianmakesthings: I will leave it as is, then.

lianmakesthings: I will leave this as is, then.

PR #5377, opened 2022-01-12
web: add checkbox column on Table View behind feature flag

lizzthabet:
Screen Shot 2022-01-12 at 4 38 40 PM

* Add a `ResourceSelection` context that tracks what resources are selected by a list of their names
* Add a checkbox column that displays when the bulk feature flag is enabled and only for resources that are disable-able
* Add support for bulk disable resources feature in Storybook stories

(This PR doesn't cover resource group selection or enable/disable menu display yet.)

The easiest way to test is through Storybook (through looking at any of the `OverviewTable` stories and toggling the flag) or spinning up any `Tiltfile` with the `bulk_disable_resources` flag enabled.

Closes [ch12973](https://app.shortcut.com/windmill/story/12973/add-checkbox-selection-to-each-row-in-table-view).

### DIFF
--- a/web/src/HUD.tsx
+++ b/web/src/HUD.tsx
@@ -23,6 +23,7 @@ import PathBuilder, { PathBuilderProvider } from "./PathBuilder"
 import { ResourceGroupsContextProvider } from "./ResourceGroupsContext"
 import { ResourceListOptionsProvider } from "./ResourceListOptionsContext"
 import { ResourceNavProvider } from "./ResourceNav"
+import { ResourceSelectionProvider } from "./ResourceSelectionContext"
 import ShareSnapshotModal from "./ShareSnapshotModal"
 import { TiltSnackbarProvider } from "./Snackbar"
 import { SnapshotActionProvider } from "./snapshot"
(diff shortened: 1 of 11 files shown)


web/src/HUD.tsx
```diff
@@ -259,19 +260,21 @@ export default class HUD extends Component<HudProps, HudState> {
...
-                          <OverviewResourcePane view={this.state.view} />
-                        )}
-                      />
-                      <Route
-                        render={() => (
-                          <OverviewTablePane view={this.state.view} />
-                        )}
-                      />
-                    </Switch>
+                    <ResourceSelectionProvider>
```
nicks, line 263: i feel like at some point we should look into how we write + organize these providers so that we don't have unnecessary re-renders (for example, there's not really any reason this needs to be re-rendered on every overview render). but this is fine for now

nicks (approved):

web/src/HUD.tsx
lizzthabet: totally, that's fair!

PR #5379, opened 2022-01-13
podlogstream: fix a regression where the reconciler would thrash on pods with long teardown times

nicks:
Hello @milas, @landism,

Please review the following commits I made in branch nicks/pls2:

bdd85656fd53a9f6151eed867eebe38f12554d7f (2022-01-13 15:37:48 -0500)
podlogstream: fix a regression where the reconciler would thrash on pods with long teardown times
Also unifies the pod deletion handling with container termination handling, so
that we stream logs exactly once on containers that terminate quickly.

### DIFF
--- a/internal/controllers/core/podlogstream/podlogstreamcontroller.go
+++ b/internal/controllers/core/podlogstream/podlogstreamcontroller.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"strings"
 	"sync"
 	"time"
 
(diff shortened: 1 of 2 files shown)


internal/controllers/core/podlogstream/podlogstreamcontroller.go
```diff
@@ -158,16 +161,17 @@ func (c *Controller) Reconcile(ctx context.Context, req reconcile.Request) (reco
...
 			c.deleteStreams(streamName)
+			deleting = true
 		}
 
-		if err != nil {
-			c.setErrorStatus(streamName, fmt.Errorf("streaming logs: %v", err))
-		} else if pod == nil {
-			c.setErrorStatus(streamName, fmt.Errorf("streaming logs: pod not found: %s", podNN))
-		} else {
+		if pod == nil || apierrors.IsNotFound(err) {
```
milas, line 170: I assume `pod == nil` whenever _any_ error is returned, no?
```suggestion
		if apierrors.IsNotFound(err) || (err == nil && pod == nil) {
```
(TBH the latter half seems unnecessary, as I'd assume it's an invariant of `PodFromInformerCache` that the Pod isn't `nil` if there's no error, so maybe `apierrors.IsNotFound(err)` is sufficient)

milas (approved):

internal/controllers/core/podlogstream/podlogstreamcontroller.go
nicks: yeah, you're right, but i think i can clean this up one better and merge it with the branch above. done!

PR #5380, opened 2022-01-13
cli: prompt to generate Tiltfile

milas:
When launching `tilt up`, if there is no Tiltfile (either via the
implicit default of `./Tiltfile` or the path passed via `-f`), the
CLI will offer to create a starter Tiltfile from a baked-in
template.

demo of running tilt up against a non-existent Tiltfile
^ OMG there's a typo in that, I fixed it but didn't re-record the GIF 🤦 😅


hyu (approved):
I'm so delighted by this and very excited to ship it!

- Love your solution for giving feedback that Tiltfile has been generated!
- Very impressive writing a brief, not-overwhelming summary of what Tiltfile is 👏🏼

(Caveat: I spent 0 moments reviewing generate_tiltfile.go because I'm frightened of and have little to add re: the logic for implementing this blocking prompt in CLI! 😅)

hyu:
Re: your typo, @milas:

image

internal/cli/generate_tiltfile.go
```diff
@@ -0,0 +1,175 @@
...
+		return generateTiltfileError, nil
+	}
+
+	lineCount := 0
+	var postFinishMessage string
+	defer func() {
+		_ = restoreTerm()
+
+		if postFinishMessage != "" {
+			// NOTE: pre-win10, there's no support for ANSI escape codes, and
```
nicksieger, line 65: 😂

internal/cli/generate_tiltfile.go
```diff
@@ -0,0 +1,175 @@
...
+				// since we have the terminal in raw mode, no signal will be fired
+				// on Ctrl-C, so we manually propagate it here (sending ourselves
+				// a SIGINT signal is not practical)
+				return generateTiltfileUserExit, userExitError
+			}
+			return generateTiltfileError, err
+		}
+
+		line = strings.ToLower(strings.TrimSpace(line))
+		if line == "y" || line == "yes" {
```
nicksieger, line 115: Up to you if it's worth just checking for 'Y'. Or can we read a single character instead of a line and use that?
```suggestion
		if strings.HasPrefix(line, "y") {
```

internal/cli/up.go
```diff
@@ -114,22 +121,34 @@ func (c *upCmd) initialTermMode(isTerminal bool) store.TerminalMode {
...
+	log.SetFlags(log.Flags() &^ (log.Ldate | log.Ltime))
+	isTTY := isatty.IsTerminal(os.Stdout.Fd())
+	termMode := c.initialTermMode(isTTY)
 
 	cmdUpTags := engineanalytics.CmdTags(map[string]string{
 		"update_mode": updateModeFlag, // before 7/8/20 this was just called "mode"
 		"term_mode":   strconv.Itoa(int(termMode)),
 	})
+
+	generateTiltfileResult, err := maybeGenerateTiltfile(c.fileName)
```
nicksieger, line 139: Should we skip asking to generate a Tiltfile `if !isatty.IsTerminal(os.Stdin.Fd())`? If Tilt is run in the background without access to one, for example.

lianmakesthings (approved):

internal/cli/up.go
milas: Yep! There's a check at the beginning of the method - the `IsTerminal` check is cheap so it was easier to do redundantly to make the logic a bit more hermetic (and simplify the analytic logic)

https://github.com/tilt-dev/tilt/blob/ccb0908011bff82975193d7498ffde340cf5d6d4/internal/cli/generate_tiltfile.go#L33-L36

internal/cli/generate_tiltfile.go
milas: Ha! I meant to add some more color to the PR description about some of my logic/decisions but I'll give a quick TL;DR here:
* The low-level term lib we already have is `tcell` - this basically gives you a character grid, so not going near that directly for this feature 🙅 
* Tilt has its own internal package called `rty` (built on top of `tcell`) which is what powers the current TUI - it's obviously very powerful, and I did give it a shot, but was hitting some issues, and since we're planning for this to be temporary and extreme interactivity/in-place updates weren't critical, it didn't seem worth spending more time on
* I really didn't want to pull in another dependency for this (same logic as in previous point)

Visually, it's not quite as spiffy in Windows as a result, but it still behaves totally reasonably IMO, so this was a compromise I felt okay with. Otherwise, I definitely wouldn't have been going near raw ANSI escape codes 🙀

nicksieger (approved):

PR #5387, opened 2022-01-18
Tiltfile Contents API

nicksieger:
- Expose Tiltfile contents through the API, and keep file and api in sync.
- Tiltfile edit/update via API is gated by feature flag `tiltfile_edit_api` (off by default for now). 
- Also allow feature flags to be defaulted via environment variables.


milas (commented):
Apologies for being slow to review this, I've been thinking a lot about it because it's definitely a very tricky challenge!

At a high-level, I think it'd be more resilient/idiomatic to handle the Tiltfile contents parts early in the reconcile and return early if an action is performed:
* If `.spec.contents` -> disk, FW will see it and a new reconcile will get queued
* If disk -> `.spec.contents`, apiserver will see `Tiltfile` update and reconcile will get queued

When the next reconcile runs, during the happy path, things will now be in sync, so the contents logic wouldn't result in an early return and we'll go through the normal run flow.

If one of them has changed, the process will continue until things are aligned.

This might require a bit more/different internal state to make a decision about which way to go (`.spec.contents` -> disk or disk -> `.spec.contents`). In particular, I'd probably shy away from `mtime` beyond being informational, both [because it's got insane semantics on its own](https://apenwarr.ca/log/20181113) and because we don't have the equivalent of a modified time for `.spec.contents`. If we track `lastWriteSHA256`, then we can do conflict resolution:
* If `hash(file) != lastWriteSHA256`, disk always wins, so overwrite `.spec.contents`
* Else, if `hash(.spec.contents) != lastWriteSHA256`, write to disk

Happy to talk further, I know this is a wall of text but I wanted to write it out to get @nicks  take as well if possible

nicksieger:
Yeah, I've been mulling ways of attacking this for a bit too. I like the idea of reconciling the API early in the cycle and requeueing/returning to do the Tiltfile execution afterwards.

nicksieger:
I went through essentially a second try/rewrite of the functionality, and I think this is cleaner and closer to ready. It actually turned out to work well to do the contents reconciliation at the end of the reconciler, so that if the API needs to be updated we update it at the same time we update the status.

All that remains are a couple of tests for the edit functionality.

nicks:
I've read this PR over several times now. honestly, I don't understand what it's doing. This might just be me missing some context.

I think there might be fundamental mismatches between what the expected data flow (as described here: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md) and the data flow this PR is trying to accomplish. Like, as far as I can tell, it's comparing the spec to the status and then modifying the spec?? which is unusual.

is there a longer explanation I can read of the expected behavior?

nicks:
i think the problem might be that you're trying to use the `Contents` field in two different ways:

- as a "desired" state of the file (i.e., "please execute this code and then write it to the file at the end")
- as an "actual" state of the file (i.e., "i want to be able to read what the file contents currently are")

and then trying to use the checksum to divine which meaning was intended. but it'd be simpler with two different fields with different meanings

internal/controllers/core/tiltfile/reconciler.go
```diff
@@ -184,6 +203,54 @@ func (r *Reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
...
+	}
+
+	// Short-circuit here when feature is disabled to prevent modifying disk file
+	if (tlr != nil && !tlr.FeatureFlags[feature.TiltfileEditAPI]) ||
+		(tlr == nil && !r.defaults[feature.TiltfileEditAPI].Enabled) {
+		return result, nil
+	}
+
+	// File contents changed since last hash
+	if fileSha256 != tf.Status.ContentsSHA256 {
```
milas, line 240: I still think the "prior" state should be read off the internal `run` object rather than the previous status - the apiserver reconciler guidelines heavily discourage reading from your own object status except in a couple of very niche cases.

Specifically, you could have a scenario where you get a cached version of the TF object so the status is outdated, and now some comparisons are happening based on "real world state" (good/`TiltfileResult`) while some are happening off potentially stale data (bad/`tf.Status`).

internal/controllers/core/tiltfile/reconciler.go
```diff
@@ -184,6 +203,54 @@ func (r *Reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
...
+		return result, nil
+	}
+
+	// API contents changed since last hash, update file on disk
+	if contentsSha256 != tf.Status.ContentsSHA256 {
+		result.sha256 = contentsSha256
+		result.contents = []byte{}
+		err = os.WriteFile(tf.Spec.Path, []byte(tf.Spec.Contents), 0644)
+	}
```
milas, line 249: I'm confused by this - `Reconcile()` uses this return result and updates `tf.Spec` if `contents` is set and `tf.Status` if `sha256` is set.

So you're writing a new file to disk, which will trigger a FW event and cause re-reconcile: that makes sense. But you're going to preemptively set the `tf.Status.ContentsSHA256` right now, even though we haven't actually used that yet.

milas (commented):
I wrote up some pseudocode that hopefully demonstrates what I'm thinking about across the comments - hopefully it makes sense but I'm also happy to jump on a call if useful.

I could also be _totally_ missing out on some edge cases - there's so much subtlety here, particularly when trying to reconcile (ha!) semantics between filesystem and apiserver that I think the control loop model iteratively driving towards desired state is going to be useful to reason about it and avoid getting into inconsistent states or stuck in a loop 🙃 

```go
func Reconcile() (ctrl.Result, error) {
	if updated, err := reconcileTiltfileContents(tf, lastRun); err != nil {
		// mark the Tiltfile as failed, user will need to manually retry it
		// after e.g. resolving permissions errors
		tf.Status.Terminated = ...
		client.Status().Update(&tf)
		return ctrl.Result{}, nil
	} else if updated {
		// contents were updated, nothing more to do for now:
		//  * Tiltfile changed on disk since last run, we overwrote .Spec.Contents
		//  * .Spec.Contents changed since last run, we updated file on disk
		// in first case, this TF object changed, so we'll get re-reconciled
		// in the second case, the associated FW object will see the event and
		// 	  result in us getting re-reonciled
		return ctrl.Result{}, nil
	}

	// execute run - write sha256 to run + runResult and use for updating status
	...
}

func reconcileTiltfileContents(tf *v1alpha1.Tiltfile, lastRun *run) (bool, error) {
	currentOnDiskSHA256 := sha256File(tf.Spec.Path)
	currentContentsSHA256 := sha256String(tf.Spec.Contents)
	if (lastRun != nil && lastRun.sha256 != currentOnDiskSHA256) ||
		(lastRun == nil && currentContentsSHA256 != currentOnDiskSHA256) {
		// either we have run before and the file on-disk has changed since then
		// OR we have not run before but disk and contents are mismatched
		//
		// the disk is considered source of truth, so it always wins in cases
		// where there's ambiguity -> unconditionally update spec with it
		tf.Spec.Contents = readFile(tf.Spec.Path)
		if err := client.Update(&tf); err != nil {
			return false, err
		}
		return true, nil
	}
	
	
	if currentContentsSHA256 != currentOnDiskSHA256 {
		// .Spec.Contents changed since the last run but disk did not, so
		// we can safely update the file on-disk with the value from the
		// apiserver
		if err := os.WriteFile(tf.Spec.Path, tf.Spec.Contents); err != nil {
			return false, err
		}
		return true, nil
	}

	// file on-disk and .Spec.Contents are in sync
	return false, nil
}
```

nicks:
one very naive take - L often had an interesting philosophy around smaller reconcilers / API objects that were more composable. and it feels like a lot of the logic you're adding isn't related to the tiltfile-execution part. would it make sense to have an independent object for "File edit sessions" that would work for any file?

nicksieger:
I had a similar thought about whether to just make it an additional reconciler for changes to the Tiltfile model if that's allowed? Otherwise making the current reconciler more composable would definitely be an improvement, and would make it easier to opt out of the contents changes in the model.

nicksieger:
Closing, may revisit this idea later when we have a UI feature in progress that needs access to the contents of the Tiltfile.

PR #5388, opened 2022-01-18
cli: fix `tilt docker-prune` command

milas:
As we move to the apiserver being the source of truth, a greater
degree of objects and their fields are managed by the reconciler
based on watches for other objects.

To facilitate this transition, there are several `InferXXX()`
helpers on `Manifest`, which are called during reconciliation for
a `Tiltfile`.

However, this means that the objects were an inconsistent (at
best) state before reconciliation stabilized. This causes issues
in some cases, the most notable being that `tilt docker-prune`
stopped working because image name inference hasn't occurred
(it does not rely on a full apiserver + reconciliation).

Now, Live Update + image properties are both inferred during the
Tiltfile load (based on the values in the Tiltfile) in addition to
the reconciler, which resolves issues with Docker pruning.
Furthermore, an integration test that explicitly invokes
`tilt docker-prune` via CLI has been added, as the behavior of it
running standalone vs in the background during `tilt up` is quite
different and led to this regression not being caught originally.


internal/tiltfile/tiltfile_state.go
```diff
@@ -1480,6 +1485,14 @@ func (s *tiltfileState) imgTargetsForDepsHelper(mn model.ManifestName, imageMapD
 			WithTiltFilename(image.tiltfilePath).
 			WithImageMapDeps(image.imageMapDeps)
 
+		// once loaded to apiserver, the reconciler will take care of keeping
+		// this up-to-date, but we ensure the properties are initialized based
+		// on the values we've got so far
+		iTarget, err = iTarget.InferImagePropertiesFromCluster(s.defaultReg)
```
nicks, line 1491: this won't work correctly. 

`defaultReg` is the result of the default_registry function

in a real `docker-prune`, you need to read this off the cluster.

nicks: (e.g., you want to prune `localhost:5000/my-image` on a kind cluster)

nicks:
re: the high-level comments - ya, i think in the long-term, a lot of these commands are going to look like `tilt dump openapi` - they'll start up a headless server with only a few reconcilers hooked up, and wait for the objects to meet some condition.

re: the short-term - i think you'll probably have to move the registry inference out into cli/docker_prune.go

milas:
Superseded by #5392.

PR #5389, opened 2022-01-18
web: demonstrate how to use react-testing-library

nicks:
Hello @lizzthabet, @hyu,

Please review the following commits I made in branch nicks/testing:

650295439fa02a7c7438493a9d9399a8eaa0d1ea (2022-01-18 18:24:30 -0500)
web: demonstrate how to use react-testing-library
see discussion here: https://dev.to/wojtekmaj/enzyme-is-dead-now-what-ekl

i don't think it's worth stopping everything and dropping enzyme right now,
but i think it's worthwhile to write new tests in this format.


lizzthabet (approved):
lgtm! i'll spend some time looking at react testing library and use it with the bulk disable work i'm doing. 

i have a couple larger questions around how we want to migrate over and how to support everyone writing frontend code on our team. the outline the author gives might be a good starting point. do you have any additional thoughts?

> * Start familiarizing yourself with React Testing Library, an officially recommended library for React components.
> * Make a rule to write new tests using RTL only.
> * Consider making a rule to rewrite tests to RTL whenever you need to touch them and/or the component they are testing.
> * In your designated time for repaying technical debt (you have designated time for repaying technical debt, right? …right?), rewrite your remaining Enzyme-based tests to RTL.

PR #5392, opened 2022-01-19
cli: fix `tilt docker-prune` command panic

### LINKED ISSUE #5370: panic on `tilt docker-prune`
### Steps to Reproduce
In tilt.build, run `tilt docker-prune`
(also repros in servantes and oneup)

### Current Behavior
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x2 addr=0x18 pc=0x102adc9ac]

goroutine 1 [running]:
github.com/docker/distribution/reference.SplitHostname({0x0, 0x0})
	/Users/matt/go/src/github.com/tilt-dev/tilt/vendor/github.com/docker/distribution/reference/reference.go:183 +0x4c
github.com/docker/distribution/reference.TrimNamed(...)
	/Users/matt/go/src/github.com/tilt-dev/tilt/vendor/github.com/docker/distribution/reference/reference.go:323
github.com/tilt-dev/tilt/internal/container.RefSelector.AsNamedOnly(...)
	/Users/matt/go/src/github.com/tilt-dev/tilt/internal/container/selector.go:127
github.com/tilt-dev/tilt/internal/container.RefSet.LocalRef({{{0x0, 0x0}, 0x0}, {{0x0, 0x0}, {0x0, 0x0}, {0x0, 0x0}}})
	/Users/matt/go/src/github.com/tilt-dev/tilt/internal/container/reference.go:87 +0x60
github.com/tilt-dev/tilt/pkg/model.LocalRefSelectorsForManifests({0x14000b45500, 0x6, 0x8})
	/Users/matt/go/src/github.com/tilt-dev/tilt/pkg/model/manifest.go:392 +0x230
github.com/tilt-dev/tilt/internal/cli.(*dockerPruneCmd).run(0x1400068ae50, {0x1051ad668, 0x140000606c0}, {0x106974e90, 0x0, 0x0})
	/Users/matt/go/src/github.com/tilt-dev/tilt/internal/cli/docker_prune.go:67 +0x294
github.com/tilt-dev/tilt/internal/cli.addCommand.func1(0x140004bb400, {0x106974e90, 0x0, 0x0})
	/Users/matt/go/src/github.com/tilt-dev/tilt/internal/cli/cli.go:142 +0xac
github.com/spf13/cobra.(*Command).execute(0x140004bb400, {0x106974e90, 0x0, 0x0})
	/Users/matt/go/src/github.com/tilt-dev/tilt/vendor/github.com/spf13/cobra/command.go:860 +0x640
github.com/spf13/cobra.(*Command).ExecuteC(0x1400080f900)
	/Users/matt/go/src/github.com/tilt-dev/tilt/vendor/github.com/spf13/cobra/command.go:974 +0x410
github.com/spf13/cobra.(*Command).Execute(...)
	/Users/matt/go/src/github.com/tilt-dev/tilt/vendor/github.com/spf13/cobra/command.go:902
github.com/tilt-dev/tilt/internal/cli.Execute()
	/Users/matt/go/src/github.com/tilt-dev/tilt/internal/cli/cli.go:87 +0xcfc
main.main()
	/Users/matt/go/src/github.com/tilt-dev/tilt/cmd/tilt/main.go:19 +0x114
```

### Context
* I added some debug logging and confirmed `iTarg.Refs` is empty [here](https://github.com/tilt-dev/tilt/blob/d79b05e6a6f667e0a34599878a36394590fc8932/pkg/model/manifest.go#L389)
* I haven't dug too much further, but I'm guessing this is due to #5272 - `tilt docker-prune` uses the tiltfile load result directly, afaict and there are no reconcilers running, so iTarg.Refs is never populated
#### `tilt doctor` Output
```
$ tilt doctor
Tilt: v0.23.4-dev, built 2022-01-11
System: darwin-arm64
---
Docker
- Host: [default]
- Server Version: 20.10.11
- API Version: 1.41
- Builder: 2
- Compose Version: v2.2.1
---
Kubernetes
- Env: docker-for-desktop
- Context: docker-desktop
- Namespace: default
- Container Runtime: docker
- Version: v1.22.4
---

---
Analytics Settings
- User Mode: opt-in
- Machine: b8b381be9003750ae6fe9bec80567735
- Repo: Rh53WceNSl4o/GBV8Qf6bQ==
```

milas:
Re-use the logic from the reconciler directly to determine the
registry (based on the cluster + `default_registry` in Tiltfile)
in the CLI code.

This addresses a crash resulting from the image properties not being
present, as they're inferred + populated during Tiltfile
reconciliation, which does not occur here.

In the future, we'll likely want this (and other similar commands
such as `tilt alpha tiltfile-result`) to launch a headless apiserver
and wait for a condition, i.e. Tiltfile loaded, before reading the
objects from the apiserver to do work.

Fixes #5370.

### DIFF
--- a/internal/cli/docker_prune.go
+++ b/internal/cli/docker_prune.go
@@ -2,6 +2,7 @@ package cli
 
 import (
 	"context"
+	"os"
 	"strings"
 	"time"
 
@@ -55,9 +56,15 @@ func (c *dockerPruneCmd) run(ctx context.Context, args []string) error {
 	a.Incr("cmd.dockerPrune", nil)
 	defer a.Flush(time.Second)
 
-	// // (Most relevant output from dockerpruner is at the `debug` level)
-	// l := logger.NewLogger(logger.DebugLvl, os.Stdout)
-	// ctx = logger.WithLogger(ctx, l)
+	if !logger.Get(ctx).Level().ShouldDisplay(logger.VerboseLvl) {
+		// Docker Pruner filters output when nothing is pruned if not in verbose
+		// logging mode, which is suitable for when it runs in the background
+		// during `tilt up`, but we always want to include that for the CLI cmd
+		// N.B. we only override if we're not already showing verbose so that
+		// 	`--debug` flag isn't impacted
+		l := logger.NewLogger(logger.VerboseLvl, os.Stdout)
+		ctx = logger.WithLogger(ctx, l)
+	}
 
 	deps, err := wireDockerPrune(ctx, a, "docker-prune")
 	if err != nil {
--- a/internal/cli/docker_prune.go
+++ b/internal/cli/docker_prune.go
@@ -2,15 +2,18 @@ package cli
 
 import (
 	"context"
-	"os"
+	"strings"
 	"time"
 
 	"github.com/spf13/cobra"
 
 	"github.com/tilt-dev/tilt/internal/analytics"
+	"github.com/tilt-dev/tilt/internal/container"
 	ctrltiltfile "github.com/tilt-dev/tilt/internal/controllers/apis/tiltfile"
+	apitiltfile "github.com/tilt-dev/tilt/internal/controllers/core/tiltfile"
 	"github.com/tilt-dev/tilt/internal/docker"
 	"github.com/tilt-dev/tilt/internal/engine/dockerprune"
+	"github.com/tilt-dev/tilt/internal/k8s"
 	"github.com/tilt-dev/tilt/internal/tiltfile"
 	"github.com/tilt-dev/tilt/pkg/logger"
 	"github.com/tilt-dev/tilt/pkg/model"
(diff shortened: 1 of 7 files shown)


internal/cli/docker_prune.go
```diff
@@ -50,9 +55,9 @@ func (c *dockerPruneCmd) run(ctx context.Context, args []string) error {
 	a.Incr("cmd.dockerPrune", nil)
 	defer a.Flush(time.Second)
 
-	// (Most relevant output from dockerpruner is at the `debug` level)
-	l := logger.NewLogger(logger.DebugLvl, os.Stdout)
-	ctx = logger.WithLogger(ctx, l)
+	// // (Most relevant output from dockerpruner is at the `debug` level)
+	// l := logger.NewLogger(logger.DebugLvl, os.Stdout)
```
nicks, line 59: ?

was this intentional?

nicks (approved):

nicks:
yay for the integration test!

internal/cli/docker_prune.go
milas: Fixed in 6404e57e2, logic now is:
* `tilt docker-prune` forces a _minimum_ of VERBOSE for logger (won't override `--debug` if passed)
* Pruner only writes out messages when _nothing_ for {cache,container,image} pruned (@ info lvl) if verbose enabled (previously was debug - slight behavior change here but aligns better since these are user messages and not Tilt internals)
* `tilt docker-prune` has NEW logging @ DEBUG which shows the resolved images it's going to prune on, this isn't always output because registry rewriting might confuse some people, but it can be useful extra context if you think it's not pruning something it should be

(That sounds much more complicated than it really is 😛 )

PR #5396, opened 2022-01-20
engine: store manifest enabled state in ManifestState

landism:
A few places in the engine (the ones changed in this PR) need to know if a Manifest is enabled and were doing this by checking UIResource.DisableStatus.

As part of the tilt args / enable/disable work, I'm changing things so that all manifests are considered disabled when created and then become enabled when the api objects shake out. For the engine bits in this PR, that'd mean tests need to ensure enabled resources have both a UIResource and a ConfigMap.
It also felt kind of weird that, e.g., BuildController was coupled to UIResource.

This almost certainly isn't how we want to do this long term, but it seems like a better interim solution than what's already there and will make some upcoming changes less invasive.

### DIFF
--- a/internal/engine/buildcontrol/build_control.go
+++ b/internal/engine/buildcontrol/build_control.go
@@ -264,10 +264,8 @@ func HoldTargetsWaitingOnDependencies(state store.EngineState, mts []*store.Mani
 
 func HoldDisabledTargets(state store.EngineState, mts []*store.ManifestTarget, holds HoldSet) {
 	for _, mt := range mts {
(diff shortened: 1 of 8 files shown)


milas (approved):
Reasoning makes sense to me!

Only other possibility I could think of to avoid refactoring all the tests would be to pass around something like `isEnabled func(state store.EngineState, mt *store.ManifestTarget)` and use the UIResource logic at runtime and a hardcoded `return true` for tests that don't want to construct the whole world.

Issue #5394, opened 2022-01-20
tilt ci never exits when you create both deployments and jobs in a single resource


nicks:
Repro case:

```
ingress_yaml = 'https://raw.githubusercontent.com/kubernetes/ingress-nginx/main/deploy/static/provider/kind/deploy.yaml'
k8s_custom_deploy(
  'ingress-nginx',
  apply_cmd="kubectl apply -f %s -o yaml" % ingress_yaml,
  delete_cmd='kubectl delete -f %s -o yaml' % ingress_yaml,
  deps=[])
```

Then run:

```
tilt ci
```

Expectation: Tilt should exit as soon as the Deployment pod becomes ready

Actual: Tilt hangs forever

The problem is this manifest contains both deployments and jobs. Tilt gets super confused. Generally, the Tilt UI will use Deployment readiness for visual state, but the CI controller will see the Job and wait for all pods to complete :grimacing: 

## Related Context

I feel like we've made some investment recently into how Tilt handles jobs, in particular:
https://github.com/tilt-dev/tilt/issues/4809
and the stuff around PodReadinessMode (in pkg/model) and around job detection in the Session API

But I'm not quite sure what the right logic should be in this case.

cc @milas @landism what do you think?

milas:
Oh, that's interesting!

I'd say we should wait for all Jobs to be done + all non-job Pods ready.

Longer-term, I could see expanding the readiness options so you could do something like:
```python
k8s_resource(
    "foo",
    readiness={
        "job:ingress-nginx-admission-create": "ignore",   # don't even look at this object
        "deployment:ingress-nginx-controller": "started",  # ignore readiness probe but make sure it launches
    },
)
```

---

Sidenote: If we make any change, I'd also suggest a version bump and call it out as a potentially breaking change - during Session controller impl, there were a couple instances of behavior changes smaller than this that I had to revert to avoid impacting existing projects.

PR #5399, opened 2022-01-20
session: use RuntimeStatus and PodReadinessMode to derive CI state

nicks:
Hello @milas, @landism,

Please review the following commits I made in branch nicks/status2:

d85eb969cb8510c4081dc1fc6bc3766cf97294fa (2022-01-20 18:40:29 -0500)
session: use RuntimeStatus and PodReadinessMode to derive CI state
This leans more on RuntimeStatus as the "source of truth"
and the logic in the tiltfile about inferring whether something
is a job or a server.

fixes https://github.com/tilt-dev/tilt/issues/5394

### DIFF
--- a/internal/engine/session/controller.go
+++ b/internal/engine/session/controller.go
@@ -5,17 +5,14 @@ import (
 	"fmt"
 	"os"
 	"sort"
-	"strings"
 	"time"
 
 	"github.com/tilt-dev/tilt/pkg/apis"
 	"github.com/tilt-dev/tilt/pkg/model"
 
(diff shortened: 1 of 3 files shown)


nicks:
oh i also used this as an excuse to clean up the reconcilation a bit with our current understanding

nicks:
oh and also - i agree with Milas' comment on the issue that we should PROBABLY be smarter about detecting when Pods come from different ancestors, and using different readiness conditions depending on the ancestor type. 

But I also think that logic should live in RuntimeStatus() rather than be something that the session controller determines on its own.

milas (approved):

PR #5405, opened 2022-01-25
web: add bulk action buttons for disabling resources in Table View

lizzthabet:
This PR adds a new `BulkApiButton` component and an `OverviewBulkActions` component, so users can enable and disable multiple resources at a time from Table View.

Because the data model and styling are pretty different for bulk buttons, I opted to fork the `ApiButton` code and customize  it based on what functionality is needed for bulk actions now. In the future, I could see the `BulkApiButton` supporting other `UIButton` properties, like inputs and custom icons, or even supporting non-`UIButton` tasks, like starring resources or triggering updates.

The **`BulkApiButton`** renders a configurable html button that, when clicked, updates a list of `UIButton`s in the Tilt API. The bulk button can be configured to require confirmation, as well as only update `UIButton`s that can be toggled to a specific state (either `"off"` or `"on"`). If there's no target/desired toggle state specified, a bulk button will update all the `UIButton`s passed to it.

The **`OverviewBulkActions`** menu displays enable and disable buttons, as well as a selection count, whenever resources are selected in Table View. It will only display when the bulk disable feature flag is on.

The unit tests for both new components use `react-testing-library`, based on feedback in #5389. (Overall, I really like the new testing library!) I did some minor refactoring to the `ApiButton` components and added helper functions to make sharing code easier. I also updated references to the same string values, so they use enums or string constants.

The best way to test is a `Tiltfile` with the bulk feature flag enabled and spinning up the web UI. Storybook will display the buttons, but the button states won't update on button press.

ps. This PR is larger than I originally intended! Let me know if it's hard to review and I'll see what I can split it into.

[Closes ch12974](https://app.shortcut.com/windmill/story/12974/implement-enable-and-disable-button-controls).

### DIFF
--- a/web/src/ApiButton.tsx
+++ b/web/src/ApiButton.tsx
@@ -49,7 +49,7 @@ type ApiButtonProps = ButtonProps & {
   uiButton: UIButton
 }
 
-type ApiIconProps = { iconName?: string; iconSVG?: string }
+type ApiIconProps = { iconName?: string; iconSvg?: string }
 
 type ApiButtonInputProps = {
   spec: UIInputSpec
@@ -59,10 +59,14 @@ type ApiButtonInputProps = {
   analyticsTags: Tags
 }
 
-type ApiButtonElementProps = ApiButtonProps & {
+type ApiButtonElementProps = ButtonProps & {
+  text: string
   confirming: boolean
   disabled: boolean
-  tags: Tags
+  iconName?: string
+  iconSvg?: string
+  analyticsTags: Tags
+  analyticsName: string
 }
 
 // UIButtons for a location, sorted into types
@@ -72,11 +76,11 @@ export type ButtonSet = {
 }
 
 export enum ApiButtonType {
-  Nav = "Global",
+  Global = "Global",
   Resource = "Resource",
 }
 
-enum ApiButtonToggleState {
+export enum ApiButtonToggleState {
   On = "on",
   Off = "off",
 }
@@ -84,8 +88,9 @@ enum ApiButtonToggleState {
 // Constants
 export const UIBUTTON_SPEC_HASH = "uibuttonspec-hash"
 export const UIBUTTON_ANNOTATION_TYPE = "tilt.dev/uibutton-type"
+export const UIBUTTON_NAV_COMPONENT_ID = "nav"
 export const UIBUTTON_TOGGLE_DISABLE_TYPE = "DisableToggle"
-const UIBUTTON_TOGGLE_INPUT_NAME = "action"
+export const UIBUTTON_TOGGLE_INPUT_NAME = "action"
 
 // Styles
 const ApiButtonFormRoot = styled.div`
@@ -155,6 +160,12 @@ export const ApiButtonInputsToggleButton = styled(InstrumentedButton)`
     padding: 0 0;
   }
 `
+
+function isDisableToggleButton(b: UIButton) {
+  return (
+    annotations(b)[UIBUTTON_ANNOTATION_TYPE] === UIBUTTON_TOGGLE_DISABLE_TYPE
+  )
+}
 
 const svgElement = (src: string): React.ReactElement => {
   const node = convertFromString(src, {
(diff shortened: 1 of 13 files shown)


web/src/ApiButton.tsx
```diff
@@ -72,20 +76,21 @@ export type ButtonSet = {
...
-enum ApiButtonToggleState {
+export enum ApiButtonToggleState {
   On = "on",
   Off = "off",
 }
 
 // Constants
 export const UIBUTTON_SPEC_HASH = "uibuttonspec-hash"
 export const UIBUTTON_ANNOTATION_TYPE = "tilt.dev/uibutton-type"
+export const UIBUTTON_NAV_COMPONENT_ID = "nav"
```
milas, line 91: Nit: it looks like other places in this PR are changing `Nav` -> `Global`, should this be `UIBUTTON_GLOBAL_COMPONENT_ID`?

web/src/ApiButton.tsx
```diff
@@ -279,20 +307,20 @@ function ApiButtonWithOptions(props: ApiButtonWithOptionsProps & ButtonProps) {
...
       >
         <ApiButtonForm {...props} />
       </FloatDialog>
     </>
   )
 }
 
-export const ApiIcon = (props: ApiIconProps) => {
-  if (props.iconSVG) {
+export const ApiIcon = ({ iconName, iconSVG }: ApiIconProps) => {
```
milas, line 318: I always forget about all the fun destructuring things you can do in modern JS ✨

web/src/BulkApiButton.test.tsx
```diff
@@ -0,0 +1,341 @@
...
+const DISABLE_BUTTON_DB = disableButton("database", true)
+const DISABLE_BUTTON_FRONTEND = disableButton("frontend", true)
+const ENABLE_BUTTON_BACKEND = disableButton("backend", false)
+const TEST_UIBUTTONS = [
+  DISABLE_BUTTON_DB,
+  DISABLE_BUTTON_FRONTEND,
+  ENABLE_BUTTON_BACKEND,
+]
+
+describe("BulkApiButton", () => {
```
milas, line 34: These tests are super readable and seem to have great coverage 👏 ✨

web/src/BulkApiButton.tsx
```diff
@@ -0,0 +1,363 @@
...
+  }
+
+  return false
+}
+
+async function bulkUpdateButtonStatus(uiButtons: UIButton[]) {
+  try {
+    await Promise.all(uiButtons.map((button) => updateButtonStatus(button, {})))
+  } catch (err) {
+    throw err
```
milas, line 179: Any reason for the try-catch if we're re-throwing as-is?

milas (commented):
Some nice cleanup and the new stuff LGTM 🙃

landism (approved):
code looks good, ui looks good!

web/src/BulkApiButton.tsx
lizzthabet: that's a great question! my native promise knowledge is a little rusty, but i thought that it's best practice to try/catch here, even though i'm expecting the parent component to do the error handling. (if nothing else, it does explicitly document that expectation. 🤷🏻‍♂️ happy to be corrected if i'm wrong!)

web/src/BulkApiButton.tsx
```diff
@@ -0,0 +1,363 @@
...
+  const individualButtonsCanBeToggled = uiButtons.map((b) =>
+    canButtonBeToggled(b, targetToggleState)
+  )
+
+  return individualButtonsCanBeToggled.some(
+    (canBeToggled) => canBeToggled === true
+  )
+}
+
+function isBulkButtonDisabled(
```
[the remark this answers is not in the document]
lizzthabet: sure! i separated out the logic bc i thought conflating toggle-able <> disabled state may be confusing since all bulk buttons aren't toggle buttons, but it sounds like that extra layer of abstraction isn't working. i'll refactor.

PR #5411, opened 2022-01-25
engine: improve behavior prior to DisableStatus reconciliation

landism:
The gist of this PR is pretty much [this comment](https://github.com/tilt-dev/tilt/blob/96080b3a945de45f7c39ae017b305b08b85d1ba9/internal/engine/buildcontrol/build_control.go#L42-L48)

### Problem

1. The Tiltfile reconciler creates a bunch of API objects, some of which have DisableSources, and the ConfigMaps at which those DisableSources point. It's possible for one of those objects to run its reconciler prior to the ConfigMap existing in the API. If a ConfigMap is absent, we treat the object as enabled. For objects that ought to be disabled from the get-go, this means we briefly start their behavior (e.g., applying k8s yaml or starting a local process) and then very quickly cancel it once the ConfigMap exists in the API server. This is undesirable for anything that produces logs or otherwise has side effects.
2. Similarly, if NextTargetToBuild starts to build a manifest that should be disabled, or skips building a manifest that should be enabled, it's easy to break the build heuristics we've put in place.

This is only an issue for a very short period after the first Tiltfile execution completes.

### Solution
A lot of this diff consists of uninteresting changes to manipulating configmaps in tests, but this should cover the interesting stuff:
1. Treat a missing DisableSource as enabled (i.e., assume the object can never be disabled)
2. Change ManifestState.Enabled to ManifestState.EnabledStatus, which can be enabled/disabled/pending (one bit should be enough for anybody), so that consumers can distinguish between *actually* enabled/disabled, and not yet reconciled.
3. Change NextTargetToBuild to hold all targets if any manifests still have a pending EnabledStatus
4. Change API objects to treat errors resolving the DisableSource as disabled instead of enabled


internal/engine/buildcontrol/build_control.go
```diff
@@ -39,6 +39,21 @@ func NextTargetToBuild(state store.EngineState) (*store.ManifestTarget, HoldSet)
...
+	// We do not know whether targets are enabled or disabled until their configmaps + uiresources are synced
+	// and reconciled. This happens very quickly after the first Tiltfile execution.
+	// If any targets have an unknown EnableStatus, then we don't have enough information to schedule builds:
+	// - If we treat an unknown as disabled but it is actually enabled, then we break our heuristic prioritization
+	//   (e.g., we might schedule k8s resources before local resources).
+	// - If we treat an unknown as enabled but it is actually disabled, then we start logging + side-effecting
+	//   a build that might immediately be canceled.
+	if pending := TargetsWithPendingEnableStatus(targets); len(pending) > 0 {
+		holds.Fill(targets, store.Hold{
+			Reason: store.HoldReasonResourceInitializing,
```
landism, line 51: I'm not in love with this name and open to alternatives.
Another option is to not even hold, just return nil - is it better to briefly show a potentially confusing hold reason, or just show nothing and risk silently holding forever? Maybe hiding this reason would be better after all. If things work as expected, it should only ever flash for an unreadably short amount of time.

internal/controllers/apis/configmap/disable.go
```diff
@@ -50,10 +65,15 @@ func MaybeNewDisableStatus(ctx context.Context, client client.Client, disableSou
 		return cm, err
 	}
 
-	isDisabled, reason, err := DisableStatus(getCM, disableSource)
+	result, reason, err := DisableStatus(getCM, disableSource)
```
nicks, line 68: i feel like having both a DisableResultError state and an `err` state - but they have totally different behavior - is a recipe for problems

internal/controllers/apis/configmap/disable.go
```diff
@@ -50,10 +65,15 @@ func MaybeNewDisableStatus(ctx context.Context, client client.Client, disableSou
...
 	if err != nil {
 		return nil, err
 	}
+	// we treat pending as disabled
+	// eventually we should probably represent isDisabled by an enum in the API, but for now
```
nicks, line 73: is there some reason not to change from isDisabled -> a state enum now?

i think that by introducing this translation layer, we may be introducing more problems (see my above comment), and that this would be simpler and easier to reason about if DisableStatus() really returned a DisableStatus

internal/controllers/apis/configmap/disable.go
nicks: nm this comment, i was wrong, i misread where the error was coming from

internal/controllers/apis/configmap/disable_test_helper.go
```diff
@@ -0,0 +1,28 @@
...
+	"strconv"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	ctrlclient "sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+
+	"github.com/tilt-dev/tilt/pkg/apis/core/v1alpha1"
+)
+
+func UpsertDisableConfigMap(ctx context.Context, client ctrlclient.Client, name string, key string, isDisabled bool) error {
```
nicks, line 14: i think it'd be good to either put this in a testing package, or name it `ForTesting` or smth to make it clear this is only for testing

internal/engine/buildcontrol/build_control.go
nicks: i'd probably use `HoldReasonTiltfileReload`...that's a good user-level description of what's actually going on (we're still processing the tiltfile execution)...we don't need to have detailed steps of how the tiltfile execution registers objects with the apiserver.

pkg/apis/core/v1alpha1/uiresource_types.go
```diff
@@ -115,6 +115,8 @@ type DisableResourceStatus struct {
 	// How many of the resource's sources are disabled.
 	DisabledCount int32 `json:"disabledCount" protobuf:"varint,2,opt,name=disabledCount"`
 
+	PendingCount int32 `json:"pendingCount" protobuf:"varint,4,opt,name=pendingCount"`
```
nicks, line 118: this needs a documentation comment (since they get autogenerated everywhere)

nicks, line 118: what do you think of having a single State enum that uses the same Disabled | Pending | Enabled | Error you described above, then using that in DisableResourceStatus and DisableStatus and ManifestState?

it feels weird to have so many different ways to represent the same state (in this PR alone we have: a DisableResult enum, a EnableStatus enum, a EnabledCount/DisabledCount field, and an IsDisabled boolean)

landism: yeah, that seems better

pkg/apis/core/v1alpha1/disable_types.go
```diff
@@ -18,10 +18,27 @@ type ConfigMapDisableSource struct {
 }
 
 type DisableStatus struct {
-	// Whether this is currently disabled.
+	// Whether this is currently disabled. Deprecated in favor of `State`.
 	Disabled bool `json:"disabled" protobuf:"varint,1,opt,name=disabled"`
```
landism, line 22: I'm guessing it's not worth the effort of v1alpha2ing everything to remove this

pkg/apis/core/v1alpha1/disable_types.go
```diff
@@ -18,10 +18,27 @@ type ConfigMapDisableSource struct {
...
 	// The last time this status was updated.
 	LastUpdateTime metav1.Time `json:"lastUpdateTime" protobuf:"bytes,2,opt,name=lastUpdateTime"`
 	// The reason this status was updated.
 	Reason string `json:"reason" protobuf:"bytes,3,opt,name=reason"`
+	// Whether this is currently disabled (if known)
+	State DisableState `json:"state" protobuf:"bytes,4,opt,name=state,casttype=DisableState"`
```
landism, line 28: `DisableState` vs `DisableStatus` are not very discriminating names. I haven't come up with a better option here.

pkg/apis/core/v1alpha1/disable_types.go
nicks: ya, let's just keep them both for now

nicks (approved):

PR #5414, opened 2022-01-27
web: show error state of websocket while connecting

nicks:
Hello @lizzthabet,

Please review the following commits I made in branch nicks/issue5403:

52e826907986a918e872ca796a56a6c70b4ede70 (2022-01-27 11:22:49 -0500)
web: show error state of websocket while connecting


web/src/HeroScreen.stories.tsx
```diff
@@ -1,8 +1,17 @@
 import React from "react"
 import HeroScreen from "./HeroScreen"
+import SocketBar from "./SocketBar"
+import { SocketState } from "./types"
 
 export default {
-  title: "New UI/_To Review/HeroScreen",
+  title: "HeroScreen",
```
lizzthabet, line 7: i think we want to keep this under the `New UI` category, otherwise it's grouped outside the rest of the stories.

lizzthabet (approved):

web/src/HeroScreen.stories.tsx
nicks: sgtm. tho i feel like we need to redo the organization here - it's kind of a mess. always think it's sus to call something "New" because that will always be inaccurate someday.

lizzthabet: yeah, i won't argue with you there! +1 on reorganizing (and refreshing) the storybook entries.

PR #5418, opened 2022-01-28
tiltfile: set disabled state instead of dropping disabled resources

landism:
Implements [this spec](https://github.com/tilt-dev/tilt.specs/blob/master/enable_disable_persistence.md), minus persistence.

TL;DR, when a Tiltfile executes:
- if its args differ from its previous run, it will reset the list of enabled/disabled resources using existing logic (i.e., empty string = all enabled, otherwise only listed resources are enabled)
- if its args are the same as its previous run, it will not change the enabled/disabled state of any resources


internal/engine/buildcontroller_test.go
```diff
@@ -732,7 +732,7 @@ func TestBuildControllerManualTriggerWithFileChangesSinceLastSuccessfulBuildButB
 
 	f.podEvent(basePB.Build())
 
-	f.b.nextBuildError = errors.New("build failure!")
+	f.SetNextBuildError(errors.New("build failure!"))
```
landism, line 735: while debugging an issue, I saw this comment, and figured I might as well act on it:
```
	// Set this to simulate a build with no results and an error.
	// Do not set this directly, use fixture.SetNextBuildError
	nextBuildError error
```

internal/engine/buildcontroller_test.go
```diff
@@ -1201,15 +1201,10 @@ func TestBuildControllerResourceDepTrumpsPendingBuild(t *testing.T) {
...
 
-	f.b.nextBuildError = errors.New("failure")
 	f.fsWatcher.Events <- watch.NewFileEvent(f.JoinPath("foo", "main.go"))
 	call = f.nextCall()
 	require.Equal(t, "foo", call.local().Name.String())
 
-	f.fsWatcher.Events <- watch.NewFileEvent(f.JoinPath("foo", "main2.go"))
-	call = f.nextCall()
-	require.Equal(t, "foo", call.local().Name.String())
-
```
landism, line 1212: AFAICT, there's not currently a good way to get the fake builder to return errors on successive builds, and between the choices of adding more configuration to the fake builder to allow that or just removing the seemingly useless third build in this test, I went for the latter

landism:
It just occurred to me that this probably requires further changes so that `tilt ci` doesn't wait on disabled resources, but I don't suspect that will have much impact on the changes already made.

landism:
> It just occurred to me that this probably requires further changes so that `tilt ci` doesn't wait on disabled resources, but I don't suspect that will have much impact on the changes already made.

done!

internal/tiltfile/config/resources.go
```diff
@@ -100,10 +104,11 @@ func match(manifests []model.Manifest, requestedManifests []model.ManifestName)
 		addManifestAndDeps(manifestsToRun, manifestsByName, m)
 	}
 
-	var result []model.Manifest
+	var result []model.ManifestName
 	for _, m := range manifests {
-		if manifestsToRun[m.Name] {
-			result = append(result, m)
+		// Default to including UnresourcedYAML ("Uncategorized") to match historical behavior.
```
landism, line 109: I'm not sure this is the right long-term behavior but didn't want to unnecessarily expand the scope of this change.

nicks (approved):

PR #5420, opened 2022-01-28
web: add disabled resource status count to summary 

lizzthabet:
This PR adds a count of disabled resources to the resource status summary component. The total resource count doesn't include the disabled resources, and the status summary will display if the (enabled) resource count or the disabled count is > 0.

Screen Shot 2022-01-28 at 4 18 33 PM

I also added more semantic html (like replacing `div`s with `li`s) and aria markup, so it's more accessible to assistive tech. (There are more improvements we can make in the future to make sure the visual information is also accessible over text and audio, but I didn't want the scope of this PR to creep.)

[Closes ch12978](https://app.shortcut.com/windmill/story/12978/display-disabled-resource-count-in-status-summary).


landism (approved):
LGTM

PR #5422, opened 2022-01-29
web: let anser colorize ansi logs

ApocDev:
This PR simply disables the `use_classes` option when converting ANSI logs to HTML.

While previously only the ANSI classes in https://github.com/tilt-dev/tilt/blob/master/web/src/AnsiLine.scss had any form of colorization support, disabling the `use_classes` option allows any ANSI log to be colorized properly.

image

### DIFF
--- a/web/src/OverviewLogPane.tsx
+++ b/web/src/OverviewLogPane.tsx
@@ -114,7 +114,8 @@ function newLineEl(
   let spacer = "\n"
   code.innerHTML = anser.linkify(
     anser.ansiToHtml(anser.escapeForHtml(line.text) + spacer, {
-      use_classes: true,
+      // Let anser colorize the html as it appears from various consoles
+      use_classes: false,
     })
   )
   span.appendChild(code)


nicks:
AFAIR, the reason we did it this way originally is because we wanted the "standard" ANSI named colors (green, blue, red, etc) to match the colors in the rest of the Tilt UI.

I wonder if there's a way to get Tilt colors for the main named colors but defer to the full color codes when they're provided.

cc @lizzthabet @SurbhiSGupta what do you think?

lizzthabet:
@nicks, i'll let others make the call on the important of keeping with the tilt color scheme! 

off the top of my head, i can't think of an efficient way for the log pane to preserve the main colors but render all the other full colors. with the class name option, anser does add a standard class `.ansi-truecolor-fg` and puts the rgb values in a `data-ansi-truecolor-fg` property for the full color codes, but styling those correctly with pure css [isn't supported](https://developer.mozilla.org/en-US/docs/Web/CSS/attr()) (ie, `.color: attr(data-ansi-truecolor-fg color)`). 

maybe there are other ideas i'm not thinking of!

ApocDev:
@nicks @lizzthabet Potentially best of both worlds? https://codepen.io/apocdev/pen/PoONqqP?editors=0100 (Please don't mind the ugliness there, I threw it together quickly. It probably can be optimized quite a bit)

We can just generate all the ANSI colors by default, then override any that need branded colors.

nicks:
I'm game to do something like that! LOL

landism:
@ApocDev After some discussion, it sounds like we're OK with what's in this PR already. Would you prefer we merge this as-is or are you working on the approach you proposed?

nicks:
let's merge it!! LGTM

PR #5424, opened 2022-01-31
engine: support shutting down multiple DC resources with one DC invocation

landism:
### Problem

Bulk actions now allow stopping multiple DC resources, but 1) we only pass one a at a time to `docker-compose rm`, 2) it often takes 10 seconds (on my machine, at least). 

### Solution

Accumulate multiple DC resources in need of stopping, and stop them all with a single invocation.

### Notes

1. Most of the complexity / concurrency in this PR is accumulating the resources in need of a disable, so that for a bulk stop, we don't call `dc rm` once for the first resource and a second time for the rest. It's more complex than I would have preferred, but I'm not aware of a simpler pattern for this using an engine subscriber.
2. FWIW, it turns out the previous implementation is buggy in that it only acts on UIResources that are marked changed in the ChangeSummary, but we actually care about changes to the ManifestState.


internal/engine/dcwatch/disabler.go
```diff
@@ -0,0 +1,126 @@
...
+	// the service's order wrt other services, to ensure logging in consistent order
+	Order int
+}
+
+func (d *disabler) Update(ctx context.Context, e disableQueueEntry) {
+	d.loopOnce.Do(func() {
+		go d.Loop(ctx)
+	})
+
+	d.updates <- e
```
nicks, line 48: if 'docker rm' really takes 10s, aren't we worried about race conditions between when an 'rm' enters the queue and when it gets scheduled? e.g., it seems like if a user is toggling a service on and off, bizarre things could happen. where the 'rm' could get schedules 10s later.

internal/engine/dcwatch/disabler.go
```diff
@@ -0,0 +1,126 @@
...
+		}
+
+		sort.Slice(toDisable, func(i, j int) bool {
+			return entries[toDisable[i].Service].Order < entries[toDisable[j].Service].Order
+		})
+
+		// Upon disabling, the DC event watcher will notice the container has stopped and update
+		// the resource's RuntimeStatus, preventing it from being re-added to specsToDisable.
+
+		// NB: For now, DC output only goes to the global log
```
nicks, line 104: can we restore the old behavior in the simple case where there's only 1 docker-compose object getting removed?

nicks (commented):
some quick highlevel questions:

nicks (commented):
(i'm also a bit nervous about how a lot of this code is hand-rolling a reconciling debouncing loop, which might be wasted effort if we're going to move this to a real reconciler? but maybe i'm misunderstanding and it will still be useful)

internal/engine/dcwatch/disabler.go
milas: Would it be nuts to make an `io.MultiWriter` to duplicate the output across all [affected] resources?

landism: We could! I haven't come up with a compelling reason to go one way or the other. I leaned this way since I figured that would set an expectation that the `docker-compose rm` logs would show up in the resource log, which could then confuse the user when they do a bulk disable and don't see the logs show up there. WDYT?

milas (commented):
I feel like there's two things at play here:
1. Compose can take up to 10 seconds
2. We don't want to repeatedly attempt to take down the same resource(s) while waiting 

On Compose stop slowness: my understanding is that `docker-compose rm` is _mostly_ slow because there's a (hardcoded*) 10 second grace period for the container(s) to handle the signal and exit cleanly. With Compose v1, maybe add 100-200ms of fixed overhead for Python.

Given that, would it be unreasonable to still rm 1:1 but do it in a non-blocking goroutine with some de-duplication logic (similar to what you've got)? That might make it easier to handle the "rapid enable/disable" case as well?

\* It's possible to call `docker-compose stop -t X foo && docker-compose rm foo` to shorten but `rm --stop` doesn't directly take a timeout arg because _reasons_ and at the end of the day, unless we're going to `-t 0`, it's still gonna be an issue

internal/engine/dcwatch/disabler.go
landism: > aren't we worried about race conditions between when an 'rm' enters the queue and when it gets scheduled?

yes! IIUC, that's handled correctly, but maybe I missed something? The loop maintains the current state of which resources are enabled and disabled, and when it calls `rm`, only passes resources that are currently disabled (i.e., if a resource is reenabled during those 10 seconds, a new update will come in with `NeedsCleanup: false`, and get filtered out before `rm` is called). Is that the concern you describe?

(Obviously, it's possible that the user reenables the resources after the last state update but before we've invoked `rm`, but that section of code should be ~instant, and that's effectively the same case as enabling the resource while it's in the process of disabling--the behavior in that case is reasonable - it finishes disabling, then reenables)

nicks: hmmm...maybe? i dunno the scheduling seems sus to me here.

if you use a blocking channel here, then will we have any race conditions where we're holding onto the EngineState Lock while the channel blocks? which means tilt will lock up?

it's also possible that i'm just more used to the pattern that kubernetes uses, where they try to keep the state caching completely separate from the scheduling logic, though i appreciate that approach has pros/cons too.

landism:
> (i'm also a bit nervous about how a lot of this code is hand-rolling a reconciling debouncing loop, which might be wasted effort if we're going to move this to a real reconciler? but maybe i'm misunderstanding and it will still be useful)

I totally expect the debouncing to be thrown out when we have a real reconciler, and while writing this, kept wishing we had a real reconciler for DC so that I wouldn't have to write it. My sense is it's not worth tackling DC API-ification as part of bulk disable.

I hadn't considered until now that we could have a subscriber write DC status to the API and have a reconciler that does nothing but stop things that need stopping. The code to populate that status would be similarly wasted once we move DC specs to the API, but at least it'd be simpler?

Is this comment a "whoaaa we shouldn't commit this code" or "let's keep an eye on this and if we keep hitting problems, let's switch to a reconciler" or "in the future, bias more strongly toward putting code in reconcilers", or something else?

landism:
> Given that, would it be unreasonable to still rm 1:1 but do it in a non-blocking goroutine with some de-duplication logic (similar to what you've got)? 

Maybe! This would certainly make things a lot simpler! The main thing (and really the reason for all of the complexity in this PR) is that our DC client currently assumes we can't trust multiple parallel DC processes, so we mutex almost everything. AFAIK, the only tangible problems I could find are with `up` and `run` (as in, there are GH issues--I have not seen any actual documentation on the safety or unsafety of parallel DC processes).

It seems intuitive `rm` would be safe to do in parallel, and a brief inspection of the code looks like it's probably safe at the moment. Given other comments on this PR, that route seems less risky, and we can reconsider if problems arise.

> (EDIT: added connecting thought ->) On repeatedly attempting to take down the same resource, that's currently handled by subscriber blocking, so I agree that's definitely a problem.

Not sure if it's important, but I haven't been able to make sense of this sentence!

nicks:
re: throwing the code out - nah, i expect we're going to still need most of this when we have a real reconcile. docker-compose isn't really safe to run in parallel because it needs to create and delete networks.

mostly asking this question to get a sense of what comments would be useful. i do think that using a sync.Map and a separate debouncer that doesn't handle state would make it a bit easier to move (and more consistent with how we handle reconciler state when there are goroutines involved)

landism:
re: logging per-manifest, just for context, the resource logs still get this:
```
Resource "fortune" disabled. To enable/disable it, use the Tilt Web UI.
```

the issue is just about whether any of this also shows up in the resource log, or only in the global log:
```
Stopping servantes_spoonerisms_1 ... 
Stopping servantes_snack_1       ... 
Stopping servantes_vigoda_1      ... 
Stopping servantes_fortune_1     ... 
Stopping servantes_vigoda_1      ... done
Stopping servantes_snack_1       ... done
Stopping servantes_fortune_1     ... done
Stopping servantes_spoonerisms_1 ... done
Removing servantes_spoonerisms_1 ... 
Removing servantes_snack_1       ... 
Removing servantes_vigoda_1      ... 
Removing servantes_fortune_1     ... 
Removing servantes_vigoda_1      ... done
Removing servantes_fortune_1     ... done
Removing servantes_snack_1       ... done
Removing servantes_spoonerisms_1 ... done
Going to remove servantes_spoonerisms_1, servantes_snack_1, servantes_vigoda_1, servantes_fortune_1
```

landism:
I've gotten rid of the chan and things are much simpler now.

milas:
> It seems intuitive rm would be safe to do in parallel, and a brief inspection of the code looks like it's probably safe at the moment

I also looked at Compose v2 earlier 😂 Since we're doing `--stop --force`, it's definitely safe BUT it doesn't handle not found, so you could conceivably try to delete the same container 2x and one of the Compose execs would fail.

It might be reasonable to limit the `rm` concurrency by Compose service to prevent that, but I also don't know how much it'd be an issue in practice (you'd need 2x Compose CLIs that query the Docker API simultaneously and then both issue deletes), and short of some potentially confusing output in the Tilt logs, nothing unexpected should happen.

internal/engine/dcwatch/disable_watcher.go
```diff
@@ -26,64 +53,98 @@ func (w *DisableSubscriber) OnChange(ctx context.Context, st store.RStore, summa
...
-		// a down service is a no-op.
-
-		ctx, err := store.WithObjectLogHandler(ctx, st, uirToDisable)
-		if err != nil {
-			logger.Get(ctx).Errorf("error creating logger for %s: %v", uirToDisable.Name, err)
-			// continue anyway:
-			// probably better to try stopping the service instead of re-logging error infinitely
-		}
-		out := logger.Get(ctx).Writer(logger.InfoLvl)
+	go func() {
```
nicks, line 91: it would be good to do a quick check here and only schedule the goroutine if at least one service needs cleanup.

creating a new goroutine and Sleep() on every OnChange() is going to create problems for the goroutine scheduler; it's why the Store is careful about batching https://github.com/tilt-dev/tilt/blob/master/internal/store/subscriber.go#L124

nicks (approved):

PR #5425, opened 2022-01-31
demo: fix perpetual "Waiting for cluster" in some setups

milas:
This should always use the _local_ Docker client rather than the
cluster client because we're creating our own ephemeral cluster
and doing that e.g. inside of minikube's Docker is unlikely to
be what anyone wants beyond resulting in the UI getting perpetually
stuck at "Waiting for cluster" because it's launched an inaccessible
ephemeral cluster with comical levels of nested virtualization/
containerization.


milas:
That Slack thread about local vs cluster Docker clients was really timely - I hit this coincidentally today and knew exactly what the problem was right away 😂

nicksieger (approved):

nicks (approved):

PR #5428, opened 2022-02-01
Kustomize: specify custom executable path

### LINKED ISSUE #5412: Allow custom path to kustomize
Allow custom path to kustomize

In new versions of kubebuilder it downloads it's own version of kustomize into a local `bin` directory. This is so all developers on an operator can use the same version of the required tools.

It would be nice if tilt could be given a custom path for the location of kustomize.

### Current Behavior

Tilt uses kustomize from the $PATH

### Why Do You Want This?

Currently my developers need to have the correct version of kustomize installed in their $PATH. The wrong version can produce invalid results as versions have made breaking changes.

This change to tilt will allow our operator projects to dictate which version of kustomize to use.

nicksieger:
Enable kustomize executable path to be kustomized. Fixes #5412.

### DIFF
--- a/internal/tiltfile/files.go
+++ b/internal/tiltfile/files.go
@@ -125,8 +125,8 @@ func (s *tiltfileState) execLocalCmd(t *starlark.Thread, cmd model.Cmd, options 
 }
 
 func (s *tiltfileState) kustomize(thread *starlark.Thread, fn *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
-	var path starlark.Value
-	err := s.unpackArgs(fn.Name(), args, kwargs, "paths", &path)
+	var path, kustomizeBin starlark.Value
+	err := s.unpackArgs(fn.Name(), args, kwargs, "paths", &path, "kustomize_bin?", &kustomizeBin)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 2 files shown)


internal/tiltfile/files.go
```diff
@@ -125,8 +125,8 @@ func (s *tiltfileState) execLocalCmd(t *starlark.Thread, cmd model.Cmd, options
 }
 
 func (s *tiltfileState) kustomize(thread *starlark.Thread, fn *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
-	var path starlark.Value
-	err := s.unpackArgs(fn.Name(), args, kwargs, "paths", &path)
+	var path, kustomizeBin starlark.Value
```
nicks, line 128: can you make these both a `NewLocalPathUnpacker()`?

Since this code was written, starlark supports custom unpackers, so the deserialization is built into the type object.

internal/tiltfile/files.go
```diff
@@ -136,6 +136,24 @@ func (s *tiltfileState) kustomize(thread *starlark.Thread, fn *starlark.Builtin,
...
+		binStr, ok := value.AsString(kustomizeBin)
+		if !ok {
+			return nil, fmt.Errorf("Could not convert kustomize_bin to string")
+		}
+		kustomizeArgs[0] = binStr
+	}
+
+	_, err = exec.LookPath(kustomizeArgs[0])
+	if err != nil {
+		s.logger.Infof("Falling back to `kubectl kustomize` since `%s` was not found in PATH", kustomizeArgs[0])
```
nicks, line 153: we probably want this to be an error if `kustomizeBin` was specified?

nicks (approved):

PR #5432, opened 2022-02-01
repo: reorganize update() to be more level-triggered

nicks:
Hello @milas, @landism,

Please review the following commits I made in branch nicks/reconcile:

915c3bc8c3f2e736e09fc1ed1455860f9d1d9de9 (2022-02-01 16:35:40 -0500)
repo: reorganize update() to be more level-triggered
This will make it easier to re-use the extensionrepo reconciler from within
the tiltfile loader.


internal/controllers/core/extensionrepo/reconciler.go
```diff
@@ -100,59 +103,70 @@ func (r *Reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
...
-		status.Path = filePath
-	})
+	state.status = v1alpha1.ExtensionRepoStatus{LastFetchedAt: timeFetched, Path: filePath}
 }
 
 // Reconcile a repo that we need to fetch remotely, and store
 // under ~/.tilt-dev.
-func (r *Reconciler) reconcileDownloaderRepo(ctx context.Context, nn types.NamespacedName,
-	repo *v1alpha1.ExtensionRepo, importPath string, state *repoState) (reconcile.Result, error) {
+func (r *Reconciler) reconcileDownloaderRepo(ctx context.Context, state *repoState, importPath string) (reconcile.Result, error) {
```
milas, line 169: It feels weird that this still returns `error` in a couple places (not sure if coincidental that they're both `stat` calls) instead of updating `state.status` accordingly

milas (approved):

internal/controllers/core/extensionrepo/reconciler.go
nicks: i think you're right! i think this should be a status update instead of a retry

PR #5436, opened 2022-02-02
cli: tilt args output indicates when there is no change

landism:
### Problem

If `tilt args` specifies the existing args value, then nothing happens. The Tiltfile is not re-executed, and the set of enabled resources is not changed. This might make a user think their command didn't work.

### Solution

Explicitly tell the user the command didn't do anything.

### DIFF
--- a/internal/cli/args.go
+++ b/internal/cli/args.go
@@ -116,8 +116,7 @@ func parseEditResult(b []byte) ([]string, error) {
 }
 
 func (c *argsCmd) run(ctx context.Context, args []string) error {
-	logLvl := logger.Get(ctx).Level()
-	ctx = logger.WithLogger(ctx, logger.NewLogger(logLvl, c.streams.ErrOut))
+	ctx = logger.WithLogger(ctx, logger.NewLogger(logger.Get(ctx).Level(), c.streams.ErrOut))
 
 	ctrlclient, err := newClient(ctx)
 	if err != nil {
@@ -159,7 +158,7 @@ func (c *argsCmd) run(ctx context.Context, args []string) error {
 	defer a.Flush(time.Second)
 
 	if sliceutils.StringSliceEquals(tf.Spec.Args, args) {
-		logger.Get(ctx).Infof("Tilt is already running with those args. No action taken.")
+		logger.Get(ctx).Infof("Tilt is already running with those args -- no action taken")
 		return nil
 	}
 	tf.Spec.Args = args
@@ -169,7 +168,7 @@ func (c *argsCmd) run(ctx context.Context, args []string) error {
 		return err
 	}
 
-	logger.Get(ctx).Infof("changed config args for Tilt running at %s to %v", apiHost(), args)
+	logger.Get(ctx).Infof("Changed config args for Tilt running at %s to %v", apiHost(), args)
 
 	return nil
 }
--- a/internal/cli/args.go
+++ b/internal/cli/args.go
@@ -5,6 +5,7 @@ import (
 	"bytes"
 	"context"
 	"fmt"
+	"os"
 	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


lizzthabet (approved):
fantastic!

internal/cli/args.go
```diff
@@ -5,28 +5,35 @@ import (
...
 type argsCmd struct {
-	clear bool
+	streams genericclioptions.IOStreams
+	clear   bool
 }
 
 func newArgsCmd() *argsCmd {
-	return &argsCmd{}
+	return &argsCmd{
+		streams: genericclioptions.IOStreams{Out: os.Stdout, ErrOut: os.Stderr, In: os.Stdin},
```
kensipe, line 35: looking over the code... many of these commands do exactly this.. making the pattern consistent.   
observations:
* it seems like a small amount of repeating code which isn't necessary
* it seems like a decision that should only be made once... the decision of what defines the IOStreams.

It would be much better to have the `cli.go` define this once and pass to all the `newXYZCmd`.   

In this case, I recommend a 

```
func newArgsCmd(streams genericclioptions.IOStreams) *argsCmd {
	return &argsCmd{
		streams: streams,
	}
}
```

internal/cli/args.go
```diff
@@ -109,6 +116,9 @@ func parseEditResult(b []byte) ([]string, error) {
 }
 
 func (c *argsCmd) run(ctx context.Context, args []string) error {
+	logLvl := logger.Get(ctx).Level()
+	ctx = logger.WithLogger(ctx, logger.NewLogger(logLvl, c.streams.ErrOut))
```
kensipe, line 120:
```suggestion
	level := logger.Get(ctx).Level()
	ctx = logger.WithLogger(ctx, logger.NewLogger(level, c.streams.ErrOut))
```

internal/cli/args.go
kensipe: I should add that this could help with testing as well... as the test could check an Out which is a buffer defined by the Test and no directed to stdio

internal/cli/args_test.go
```diff
@@ -71,6 +72,24 @@ func TestArgsClearAndNewValue(t *testing.T) {
...
+func TestArgsNoChange(t *testing.T) {
+	f := newServerFixture(t)
+	defer f.TearDown()
+
+	createTiltfile(f, []string{"foo", "bar"})
+
+	cmd := newArgsCmd()
+	out := &bytes.Buffer{}
+	cmd.streams.Out = out
+	cmd.streams.ErrOut = out
```
kensipe, line 84: this necessary because the current design... it is the creation of a default IOStream object then setting each property for a test.  It would be much better to not create the defaults and to pass in the IOStream to the cmd IMO.

internal/cli/args.go
```diff
@@ -148,14 +158,18 @@ func (c *argsCmd) run(ctx context.Context, args []string) error {
 	a.Incr("cmd.args", tags.AsMap())
 	defer a.Flush(time.Second)
 
+	if sliceutils.StringSliceEquals(tf.Spec.Args, args) {
+		logger.Get(ctx).Infof("Tilt is already running with those args. No action taken.")
```
kensipe, line 162: It seems concerning to have capitalization and punctuation.   Below log message don't have it... this has it.  It is unclear when it is appropriate or not but seems at this point inconsistent.

kensipe (commented):
not a project member (yet)... just your everyday fly by neighborhood OSS review.   Nothing blocking, just some things to consider.  Nice contribution!

internal/cli/args_test.go
landism: same as previous stream comment

internal/cli/args.go
landism: I just got rid of the variable entirely, instead

PR #5437, opened 2022-02-02
tiltfile: add config.set_disabled_resources

landism:
Add a way to specify that all resources should be disabled, giving a "resource catalog" experience.

`config.set_enabled_resources([])` currently acts as `tilt up` with `argv == []`, i.e., it enables all resources. We probably don't want to change this since it'll break basic use cases like the first [tiltfile config example](https://docs.tilt.dev/tiltfile_config.html#run-only-some-services-reimplement-default-tilt-behavior).

`config.set_enabled_resources(None)` was previously an error, but will now indicate that all resources should be disabled.

### DIFF
--- a/internal/tiltfile/config/resources.go
+++ b/internal/tiltfile/config/resources.go
@@ -14,7 +14,7 @@ import (
 )
 
 func setEnabledResources(thread *starlark.Thread, fn *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
-	var slResources starlark.Sequence
+	var slResources starlark.Value
 	err := starkit.UnpackArgs(thread, fn.Name(), args, kwargs,
 		"resources",
 		&slResources,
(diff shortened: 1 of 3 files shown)


nicks:
oof. i understand why you want to do this. but i think making `config.set_enabled_resources(None)` and `config.set_enabled_resources([])` do different things is going to be so deeply weird that it will make the problem worse.

i could think of a few different options

what if we had a separate method for this, like:

```
config.clear_resources()
```

that just worked exactly like your current `config.set_enabled_resources(None)`?

another alternative would be to have different modes for what servers to start with:

```
# Starts with 0 resources
config.catalog_mode('none')

# This is the default mode. Uncomment to start with all resources
# config.catalog_mode('all')
```

to express different modes for how people start tilt? we could even use the docker-compose 'profiles' terminology (https://docs.docker.com/compose/profiles/)

```
# Starts with 0 resources
config.profile('none')

# This is the default mode. Uncomment to start with all resources
# config.profile('all')
```

which might open up the door to let the user define different profiles (for sets of services that should start in edit mode vs run mode)

landism:
> but i think making config.set_enabled_resources(None) and config.set_enabled_resources([]) do different things is going to be so deeply weird that it will make the problem worse

fair!

> which might open up the door to let the user define different profiles (for sets of services that should start in edit mode vs run mode)

I like this! But I'm inclined to stick with the cheaper option for the moment.

internal/tiltfile/config/resources.go
```diff
@@ -44,8 +45,20 @@ func setEnabledResources(thread *starlark.Thread, fn *starlark.Builtin, args sta
 	return starlark.None, nil
 }
 
+func clearEnabledResources(thread *starlark.Thread, fn *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
+	err := starkit.SetState(thread, func(settings Settings) Settings {
+		settings.disableAll = true
```
nicks, line 50: should this also set enabledResources to nil? i guess it doesn't really matter.

internal/tiltfile/config/resources.go
```diff
@@ -44,8 +45,20 @@ func setEnabledResources(thread *starlark.Thread, fn *starlark.Builtin, args sta
 	return starlark.None, nil
 }
 
+func clearEnabledResources(thread *starlark.Thread, fn *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
+	err := starkit.SetState(thread, func(settings Settings) Settings {
```
nicks, line 49: we probably should still do an Unpack() here, so that it errors out if the user passes any arguments.

nicks (approved):

internal/tiltfile/config/resources.go
landism: Yeah, shouldn't affect functionality at the moment, but probably better to have it, to make surprises less likely when changing related code.

internal/tiltfile/config/resources.go
landism: oops, yes!

PR #5438, opened 2022-02-02
features: turn on Disable Resources feature flag by default

lizzthabet:
This PR turns on the `DisableResources` flag by default and deprecates `BulkDisableResources` flag since it is no longer needed. All the frontend code that uses the bulk feature flag now uses the disable resources flag.

Disable resources will launch for everyone on the next release. 🎉


landism (approved):
🎉

PR #5439, opened 2022-02-03
web: add resource selection checkbox for table groups

lizzthabet:
This PR adds a checkbox to the resource selection column header for each table displayed in Table View. The header checkbox will select or deselect all resources in its table. It will display as unchecked, indeterminate (partially checked), or checked, depending on the resources in its table.

Here's a sample of what those states look like:
Screen Shot 2022-02-02 at 6 36 06 PM

I'd like to refactor the `ResourceSelectionContext` tests and possibly the `OverviewTable` tests to use React Testing Library in a separate PR, so that this one can be reviewed and potentially make into the next release in time.

[Closes ch12975](https://app.shortcut.com/windmill/story/12975/add-checkbox-for-resource-groups).

### DIFF
--- a/web/src/OverviewTableColumns.tsx
+++ b/web/src/OverviewTableColumns.tsx
@@ -1,5 +1,5 @@
-import React, { ChangeEvent, useCallback, useState } from "react"
-import { CellProps, Column, Row } from "react-table"
+import React, { ChangeEvent, useCallback, useMemo, useState } from "react"
+import { CellProps, Column, HeaderProps, Row } from "react-table"
 import TimeAgo from "react-timeago"
 import styled from "styled-components"
 import { AnalyticsAction, AnalyticsType, incr, Tags } from "./analytics"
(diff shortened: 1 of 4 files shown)


web/src/OverviewTableColumns.tsx
```diff
@@ -533,7 +617,6 @@ const DEFAULT_COLUMNS: Column<RowValues>[] = [
   {
     Header: () => <TableHeaderStarIcon title="Starred" />,
     id: "starred",
-    accessor: "name", // Note: this accessor is meaningless but required when `Header` returns JSX. The starred column gets its data directly from the StarredResources context and sort on this column is disabled.
```
landism, line 536: huh! did this comment turn out to be wrong, or was there some other change I missed?

(if only I'd also commented what I observed to make me think it was required)

web/src/OverviewTableColumns.tsx
```diff
@@ -209,6 +209,89 @@ function statusSortKey(row: RowValues): string {
...
+
+    for (let i = 0; i < resourcesInTable.length; i++) {
+      const selected = isSelected(resourcesInTable[i])
+      if (selected) {
+        selectionCount++
+      }
+
+      // If at least one item has been selected, but the
+      // current item isn't, we can exit the loop early
+      if (!selected && selectionCount > 0) {
```
landism, line 246: As long as we're doing small optimizations, this stops early for [selected, unselected], but not [unselected, selected] :)

totally fine as-is, but I think this would be easier to understand without counting, e.g.:
```
let anySelected = false
let anyUnselected = false
resourcesInTable.forEach(r => {
  if (isSelected(r)) {
    anySelected = true
  } else {
    anyUnselected = true
  }

  if (anySelected && anyUnselected) {
    break
  }
}

return {
  indeterminate: anySelected !== anyUnselected
  checked: !anyUnselected
}
```

landism (approved):
LGTM!

web/src/OverviewTableColumns.tsx
lizzthabet: that's a great question! what i remember (when adding that comment) is that there was a typescript compilation error, saying that `accessor` was required, even though [react-table's docs](https://react-table.tanstack.com/docs/api/useTable#column-options) say that you can omit it when you have an id. (i've reread these docs a couple times and my take-away is a little bit different each time. 😅) 

i didn't get the same ts complaint this time around, which is why i removed the field. because both the starred and selection columns get data from context, rather than table data, i think it's okay to omit the `accessor` property. 

(however, this is making me consider that we should be mapping the context data to the table row data, instead of grabbing the data from whatever column component, and then that data would be sortable, etc. definitely a thought for another pr though!)

web/src/OverviewTableColumns.tsx
lizzthabet: ah, great suggestion! i'll update.

PR #5440, opened 2022-02-03
Linting fix:  gocritic for replaceall and commentFormatting

kensipe:
There is a preference to use `replaceAll` over `replace` with a `-1`.  `replaceAll` is more readable and expressive and is simply a wrapper around the latter func.  It is easier when reading to reason about.

gocritic also wants `//comment` to be `// comment` which does look better.  Additionally this change makes the comments consistent with all other comments in the code base.

### DIFF
--- a/integration/fixture_test.go
+++ b/integration/fixture_test.go
@@ -287,7 +287,7 @@ func (f *fixture) ReplaceContents(fileBaseName, original, replacement string) {
 		f.originalFiles[file] = contents
 	}
 
-	newContents := strings.Replace(contents, original, replacement, -1)
+	newContents := strings.ReplaceAll(contents, original, replacement)
 	if newContents == contents {
 		f.t.Fatalf("Could not find contents %q to replace in file %s: %s", original, fileBaseName, contents)
 	}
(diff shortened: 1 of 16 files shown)


kensipe:
Looked deep to under if this was PR related and it doesn't look to be.  All tests run successful locally.  Setting up circleci for local execution didn't work, I believe that is an issue with the right arch of the linux image (I'm on an M1 processor).  Reviewing code, I can't see a reason this would fail.  I don't have rights to hit the retest button on circleci for this project.  any help @nicksieger ?   

happy to fix if I can reproduce or had the rights to understand

kensipe:
oh wow... that failure is a flaky failure.  it fails on some frequency locally.  So...  It seems my changes are not the cause (yay).  I'll write up an issue.  I would guess a bad test design usually around eventual consistency... but just a guess.  Flaky tests are dangerous... we will need to address (on the issue and future PR)

lizzthabet (approved):
lgtm! if you have another issue with a test failing in ci, but not locally, you're welcome to note it in the pr and someone from the tilt team can take a look and rerun the test. circleci also has a [list of "flaky" test suites](https://app.circleci.com/insights/github/tilt-dev/tilt/workflows/build/tests?branch=master&reporting-window=last-30-days) (which i believe is public!) that includes tests that failed and succeeded on the same commit. sometimes i find this view helpful when i have test failures i can't explain.

kensipe:
thanks @lizzthabet !

PR #5443, opened 2022-02-03
cli: add tilt enable/disable

landism:
Some CLI ergonomics that should feel a lot nicer than `tilt args` for use cases that don't involve Tiltfile arg parsing.

```
# enables the resources named 'frontend' and 'backend'
tilt enable frontend backend

# enables frontend and backend and disables all others
tilt enable --only frontend backend

# enables all resources
tilt enable --all

# disables the resources named 'frontend' and 'backend'
tilt disable frontend backend

# disables all resources
tilt disable --all
```

### DIFF
--- a/internal/cli/disable.go
+++ b/internal/cli/disable.go
@@ -1,0 +1,84 @@
+package cli
+
+import (
+	"context"
+	"strconv"
+	"time"
+
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+
+	"github.com/tilt-dev/tilt/internal/analytics"
+	engineanalytics "github.com/tilt-dev/tilt/internal/engine/analytics"
+	"github.com/tilt-dev/tilt/pkg/apis/core/v1alpha1"
+	"github.com/tilt-dev/tilt/pkg/model"
+)
+
+type disableCmd struct {
+	all bool
+}
+
+func newDisableCmd() *disableCmd {
+	return &disableCmd{}
+}
+
+func (c *disableCmd) name() model.TiltSubcommand { return "disable" }
+
+func (c *disableCmd) register() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:                   "disable {-all | <resource>...}",
+		DisableFlagsInUseLine: true,
+		Short:                 "Disables resources",
+		Long: `Disables the specified resources in Tilt.
+
+# disables the resources named 'frontend' and 'backend'
+tilt disable frontend backend
+
+# disables all resources
+tilt disable --all`,
+	}
+
+	cmd.Flags().BoolVar(&c.all, "all", false, "Disable all resources")
+
+	addConnectServerFlags(cmd)
+
+	return cmd
+}
+
+func (c *disableCmd) run(ctx context.Context, args []string) error {
+	ctrlclient, err := newClient(ctx)
+	if err != nil {
+		return err
+	}
+
+	if c.all {
+		if len(args) > 0 {
+			return errors.New("cannot use --all with resource names")
+		}
+	} else if len(args) == 0 {
+		return errors.New("must specify at least one resource")
+	}
+
+	a := analytics.Get(ctx)
+	cmdTags := engineanalytics.CmdTags(map[string]string{})
+	cmdTags["all"] = strconv.FormatBool(c.all)
+	a.Incr("cmd.disable", cmdTags.AsMap())
+	defer a.Flush(time.Second)
+
+	names := make(map[string]bool)
+	for _, name := range args {
+		names[name] = true
+	}
+
+	unselectedState := v1alpha1.DisableStatePending
+	if c.all {
+		unselectedState = v1alpha1.DisableStateDisabled
+	}
+
+	err = changeEnabledResources(ctx, ctrlclient, args, v1alpha1.DisableStateDisabled, unselectedState)
+	if err != nil {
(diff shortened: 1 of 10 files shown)


internal/cli/disable.go
```diff
@@ -0,0 +1,84 @@
...
+	cmdTags["all"] = strconv.FormatBool(c.all)
+	a.Incr("cmd.disable", cmdTags.AsMap())
+	defer a.Flush(time.Second)
+
+	names := make(map[string]bool)
+	for _, name := range args {
+		names[name] = true
+	}
+
+	unselectedState := v1alpha1.DisableStatePending
```
nicks, line 73: i feel like it'd be more idiomatic to have some sort of struct

EnableOptions{All: boolean, Only: boolean}

and just pass it on to the implementation function? sort of like ListOptions https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1#ListOptions

i found the use of "DisableStatePending" as a magic value here kind of inscrutable.

nicks (approved):
LGTM!

internal/cli/disable.go
```diff
@@ -0,0 +1,84 @@
...
+}
+
+func (c *disableCmd) run(ctx context.Context, args []string) error {
+	ctrlclient, err := newClient(ctx)
+	if err != nil {
+		return err
+	}
+
+	if c.all {
+		if len(args) > 0 {
```
kensipe, line 55: could we consider combining the if conditions  to reduce the nesting?
```suggestion
	if c.all && len(args) > 0 {
```

internal/cli/disable.go
```diff
@@ -0,0 +1,84 @@
...
+			return errors.New("cannot use --all with resource names")
+		}
+	} else if len(args) == 0 {
```
kensipe, line 58: we can collapse the `else` it isn't necessary.
```suggestion
	} 
	if len(args) == 0 {
```

internal/cli/enable.go
```diff
@@ -0,0 +1,174 @@
...
+func (c *enableCmd) run(ctx context.Context, args []string) error {
+	ctrlclient, err := newClient(ctx)
+	if err != nil {
+		return err
+	}
+
+	if c.all {
+		if c.only {
+			return errors.New("cannot use --all with --only")
+		} else if len(args) > 0 {
```
kensipe, line 66: we can drop the `else`
```suggestion
		}
		 if len(args) > 0 {
```

internal/cli/enable.go
```diff
@@ -0,0 +1,174 @@
...
+			return errors.New("cannot use --all with resource names")
+		}
+	} else if len(args) == 0 {
```
kensipe, line 69: we can drop the `else`
```suggestion
	} 
	if len(args) == 0 {
```

internal/cli/enable.go
```diff
@@ -0,0 +1,174 @@
...
+		unselectedState = v1alpha1.DisableStateDisabled
+	} else if c.all {
+		unselectedState = v1alpha1.DisableStateEnabled
+	}
+	err = changeEnabledResources(ctx, ctrlclient, args, v1alpha1.DisableStateEnabled, unselectedState)
+	if err != nil {
+		return err
+	}
+
+	return nil
```
kensipe, line 96: this can be simplified to
```suggestion
	return changeEnabledResources(ctx, ctrlclient, args, v1alpha1.DisableStateEnabled, unselectedState)	
```

internal/cli/enable.go
```diff
@@ -0,0 +1,174 @@
...
+// For all other resources:
+// - if `unselectedState` is Enabled or Disabled, ensure their disabled state is that
+// - otherwise, ignore them
+func changeEnabledResources(
+	ctx context.Context,
+	cli client.Client,
+	selectedResources []string,
+	selectedState v1alpha1.DisableState,
+	unselectedState v1alpha1.DisableState,
+) error {
```
kensipe, line 110:
```suggestion
	unselectedState v1alpha1.DisableState) error {
```
this seems easier to read

kensipe (commented):
some suggestions to consider

internal/cli/disable.go
landism: yeah, that's much nicer. thanks!

internal/cli/disable.go
landism: then the else will trigger when `c.all` is `false` and `len(args) == 0`, which is not what we want

internal/cli/enable.go
landism: really? that leaves us without anything visually breaking up the params and the code:
image

internal/cli/enable.go
landism: It sure can be!
I tend to prefer it this way in a longer function like this -  I don't have to wonder about what's being returned (how many values, what are they).

landism:
Hi @kensipe! Thanks for taking a look at this PR.

I think the comments I've left unresolved are effectively "I prefer it this way" style comments without a rationale -- I happen to prefer them the way they're written as-is! One might be better than the other, but I think the difference is unlikely to be of much importance. These feel like the sort of style issue that many programmers love getting into debates over but (generally) aren't worth the time to worry about.

I'd like to get this change in for today's release, so I'm going to merge as-is, but if you disagree, we can talk about making these changes in another commit.

Issue #5446, opened 2022-02-03
Virtualised microk8s registry settings not correctly inferred.


dspasojevic:
### Expected Behavior
Tilt correctly infers both the `Host` and `hostFromCluster` values of a registry when using microk8s in a vm. The `Host` should probably be the address of the registry, as addressable from the host machine (i.e. the IP address of the VM) and `hostFromCluster` should be `localhost`.

### Current Behavior
The `Host` value is hard-coded to be `localhost`: https://github.com/tilt-dev/tilt/blob/77ae37f8f66c4b2d6df5eb526bd71954be73ecf2/internal/k8s/registry.go#L101 

#### `tilt doctor` Output
$ tilt doctor
Tilt: v0.23.9, built 2022-01-27
System: darwin-arm64
---
Docker
- Host: [default]
- Server Version: 20.10.12
- API Version: 1.41
- Builder: 2
- Compose Version: v2.2.3
---
Kubernetes
- Env: microk8s
- Context: microk8s
- Cluster Name: microk8s-cluster
- Namespace: default
- Container Runtime: containerd
- Version: v1.22.5-3+2cf820a97417ce
- Cluster Local Registry: {Host:localhost:32000 hostFromCluster: SingleName:}
---
```

#### About Your Use Case
I use macos. Docker is running in a VM using Docker Desktop. microk8s is running in a different one using multipass. I have enabled microk8s local registry. In the Docker VM, `localhost` (the inferred registry host) is referring to the docker vm, not the host vm. `default_registry` (as discussed in #4125) can not be used to override the inferred registry values.

When I hack `DecideRegistry` to always use the `default_registry` value, things work, so I think my issue is only related to the inference of the registry settings for microk8s.

**Comment by nicksieger** [comment]
Did you run `microk8s install` to build your cluster, or did you do any additional configuration? I want to make sure I know how to recreate the issue.

**Comment by nicks** [comment]
fwiw, it's hard-coded to localhost in microk8s too: https://github.com/ubuntu/microk8s/blob/c70f5f25b1ecb5a7d75718ee3169e967610265c1/microk8s-resources/actions/registry.yaml#L77

if you do:

```
kubectl edit -n kube-public configmap local-registry-hosting
```

and change it to the VM IP, I think Tilt will pick it up automatically?

dspasojevic:
@nicksieger I installed a vanilla version of microk8s and enabled the registry add-on.

dspasojevic:
[...]

PR #5449, opened 2022-02-04
tiltfile: warn if default registry has been set but will be ignored

dspasojevic:
This suggested change addresses the possible lack of feedback discussed in #5446

### DIFF
--- a/internal/controllers/core/tiltfile/reconciler.go
+++ b/internal/controllers/core/tiltfile/reconciler.go
@@ -429,7 +429,7 @@ func DecideRegistry(ctx context.Context, kCli k8s.Client, tlr *tiltfile.Tiltfile
 
 		if !tlr.DefaultRegistry.Empty() {
 			// The user has specified a default registry in their Tiltfile, but it will be ignored.
-			logger.Get(ctx).Warnf("Default registry specified, but will be ignored in favour of auto-detected registry.")
+			logger.Get(ctx).Warnf("Default registry specified, but will be ignored in favor of auto-detected registry.")
 		}
 
 		return registry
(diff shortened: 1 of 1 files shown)


internal/controllers/core/tiltfile/reconciler.go
```diff
@@ -426,6 +426,12 @@ func DecideRegistry(ctx context.Context, kCli k8s.Client, tlr *tiltfile.Tiltfile
 		// If we've found a local registry in the cluster at run-time, use that
 		// instead of the default_registry (if any) declared in the Tiltfile
 		logger.Get(ctx).Infof("Auto-detected local registry from environment: %s", registry)
+
+		if !tlr.DefaultRegistry.Empty() {
+			// The user has specified a default registry in their Tiltfile, but it will be ignored.
+			logger.Get(ctx).Warnf("Default registry specified, but will be ignored in favor of auto-detected registry.")
```
nicks, line 432: this should be an info log rather than a warning log

warnings mean "this code is incorrect 100% of the time and we can't do what you want here"

There are legitimate reasons to specify `default_registry` but still prefer a local cluster, as described in this example: https://docs.tilt.dev/personal_registry.html#the-medium-easy-way-for-4-of-users

dspasojevic: Changed.

For an individual, setting default registry never has an effect if they have a local registry. So it can still be useful for a team to have it in their Tiltfile for cases where some individuals may not have a cluster with a local registry?

nicks: yep! many teams use both remote clusters and local clusters depending on what they're doing.

nicks (approved):
thank you! LGTM

PR #5453, opened 2022-02-04
Fix local-prefixes on linters imports

kensipe:
Fix local-prefixes on linters imports then resolved imports to be as expected.

I've made this configuration mistake in the past and quite familiar with it.    The order of import expectations is that each group should be alphabetized and the groups are:

1. go core deps
2. 3rd party deps
3. internal to project deps

I was going to make an entry for the contributes but didn't see a logical place to drop that information.  It is fairly well adopted in OSS go projects but may add value in the contrib MD.  With some guidance, I'm happy to add to this PR or another import expectations.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -42,7 +42,7 @@ linters-settings:
           - Debugf
           - PrintColorf
   goimports:
-    local-prefixes: github.com/tilt-dev/tilt
+    local-prefixes: github.com/tilt-dev
   misspell:
     # Correct spellings using locale preferences for US or UK.
     # Default is to use a neutral variety of English.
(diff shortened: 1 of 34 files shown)


lizzthabet:
thanks, @kensipe! i don't remember if you were having trouble with seeing the circleci failures on an earlier pr, so i'll post the failure here just in case:

Screen Shot 2022-02-04 at 4 39 34 PM

kensipe:
thank you @lizzthabet !   I am able to see circleci... just haven't had a chance to investigate yet... but looking now

kensipe:
hmm.. that is strange.  it passes locally.   Continuing investigate

kensipe:
ah.. makes sense... it has a build tag which was missed locally.  on it!

PR #5454, opened 2022-02-04
engine: fix build graph structure error on LU-only targets

### LINKED ISSUE #5447: Bug with Tilt 23.9 and Live Update
### Expected Behavior
Live update to start.

### Current Behavior
Live Update failed with unexpected error: Cannot extract live updates on this build graph structure error. I've attached a screenshot of the error. Reverting back to 23.2, I no longer encounter the error.

### Steps to Reproduce
1. Start live update with Tiltfile.
2. Terminal output will show error.

### Context
My Tilt file(redacted):
```
SOURCE_IMAGE = os.getenv("SOURCE_IMAGE", default='your-registry.io/project/image')
LOCAL_PATH = os.getenv("LOCAL_PATH", default='.')
NAMESPACE = os.getenv("NAMESPACE", default='default')

k8s_custom_deploy(
    'app-name',
    apply_cmd="my-cli apps workload apply -f config/workload.yaml --live-update" +
               " --local-path " + LOCAL_PATH +
               " --source-image " + SOURCE_IMAGE +
               " --namespace " + NAMESPACE +
               " --yes >/dev/null" +
               " && kubectl get workload app-name --namespace " + NAMESPACE + " -o yaml",
    delete_cmd="my-cli apps workload delete -f config/workload.yaml --namespace " + NAMESPACE + " --yes",
    deps=['pom.xml', './target/classes'],
    container_selector='workload',
    live_update=[
      sync('./target/classes', '/workspace/BOOT-INF/classes')
    ]
)

allow_k8s_contexts('my-cluster')
k8s_resource('app-name', port_forwards=["8080:8080"],
            extra_pod_selectors=[{'serving.knative.dev/service': 'app-name'}])
```

milas:
Historically, the engine builds a lightweight DAG for the targets
in a manifest. A sanity check is done up-front to ensure there's
a single root node (the "deploy" target, i.e. `K8sTarget` or
`DockerComposeTarget` for the cases relevant to this bug). That
node can then have children, which will be `ImageTarget`s, and
for completeness sake, those can recursively have further child
`ImageTarget`s for base images.

Live Update-only targets are currently modeled in the engine as
`ImageTarget`s with a special `IsLiveUpdateOnly` flag which causes
the engine to skip the (non-existant) build. Because there is no
build, there is no `ImageMap`.

As part of API migration, the deploy targets mentioned above are
now driven by their API specs as the source of truth, and both
`KubernetesApplySpec` and `DockerComposeUpSpec` include an
`ImageMaps` field, which is used when building the DAG.

This means that LU-only targets were no longer represented as a
dependency of the deploy target. In reality, this is correct!
~~However, the DAG sanity check now fails as a result. The changes
here are to add (engine-only) tracking of the LU-only targets,
which are then merged with the `ImageMap` dependencies for DAG
assembly.~~

~~While this is less than ideal because we've been working hard to
eliminate "secret"/special bits of state like this and drive all
the things from the API models, I'm not sure what a good alternative
is right now, as removing the single root-node invariant will
require pretty extensive testing to ensure that things behave
correctly.~~

However, the DAG sanity check now fails as a result. The change
here is to temporarily remove the LU-only targets from the set
before building the DAG, and then add them back in to the list
of deployed images.

Once API migration work is further along (e.g. Docker Compose
Live Update happens via reconciler), the DAG logic and this hack
will go away.

Fixes #5447.

### DIFF
--- a/internal/engine/buildcontrol/extractors.go
+++ b/internal/engine/buildcontrol/extractors.go
@@ -31,7 +31,31 @@ func extractImageAndK8sTargets(specs []model.TargetSpec) (iTargets []model.Image
 // If there are images that can be updated in-place in a container, return
 // a state tree of what needs to be updated.
 func extractImageTargetsForLiveUpdates(specs []model.TargetSpec, stateSet store.BuildStateSet) ([]liveUpdateStateTree, error) {
-	g, err := model.NewTargetGraph(specs)
+	// HACK(milas): LU-only targets are NOT dependencies of the deploy target
+	// 	(there is no build); this breaks the Live Update DAG logic which wants
+	// 	the deploy target (K8sTarget/DockerComposeTarget) as the root node with
+	// 	ImageTargets as dependencies, which can then themselves have more
+	// 	ImageTargets as dependencies for base images. As a workaround, these
+	// 	are _removed_ from the spec list for the graph and then always included
+	//  in the collection of deployed images considered for Live Update ops.
+	// 	Moving forward, this will be a non-issue, as the API reconcilers are
+	// 	designed to work independently, but this keeps compatibility for the
+	// 	v1 Live Update (currently always used for Docker Compose and only if
+	// 	a feature flag is explicitly disabled for K8s) and prevents a scary
+	// 	error in the v2 case.
+	var luOnlyTargets []model.ImageTarget
+	specsForGraph := make([]model.TargetSpec, 0, len(specs))
+	for _, spec := range specs {
+		if iTarget, ok := spec.(model.ImageTarget); ok && iTarget.IsLiveUpdateOnly {
+			luOnlyTargets = append(luOnlyTargets, iTarget)
+		} else {
+			// * any image targets that have a build (i.e. not LU-only)
+			// * all other targets (e.g. K8sTarget, DockerComposeTarget)
+			specsForGraph = append(specsForGraph, spec)
+		}
+	}
+
+	g, err := model.NewTargetGraph(specsForGraph)
 	if err != nil {
 		return nil, errors.Wrap(err, "extractImageTargetsForLiveUpdates")
 	}
@@ -43,6 +67,8 @@ func extractImageTargetsForLiveUpdates(specs []model.TargetSpec, stateSet store.
 	result := make([]liveUpdateStateTree, 0)
 
 	deployedImages := g.DeployedImages()
+	deployedImages = append(deployedImages, luOnlyTargets...)
+
 	for _, iTarget := range deployedImages {
 		state := stateSet[iTarget.ID()]
 		// If this is a normal image build, it must have info about the deployed image.


milas:
One additional thing that's worth mentioning is despite the warning in the UI, LU actually continues to work in these cases from what I can see ⬇️ 

Our integration test for k8s_custom_deploy actually tests that and still works:
https://github.com/tilt-dev/tilt/blob/ecb9daaaea8728786617007f83ee7f3fd7882a91/integration/k8s_custom_deploy_test.go#L59-L64

I am not intimately familiar with all the nuances of the build/engine filtering logic, but I'm _guessing_ what's happening is:
* Initial build -> error triggered because both deploy `K8sTarget` + LU-only `ImageTarget` passed which aren't parent->child as expected
  * This build would have been rejected by the LU BaDer regardless because it's an initial build! Though there might be some sketchy special casing there for LU-only targets?
* LU watched file changes -> only the LU-only `ImageTarget` submitted to build queue, so there's only a single node in the DAG and it's happy

pkg/model/k8s_target.go
```diff
@@ -55,6 +55,15 @@ type K8sTarget struct {
 	// re-evaluated.
 	pathDependencies []string
 	localRepos       []LocalGitRepo
+
+	// HACK(milas): Live Update-only targets don't have an ImageMap because
+	// 	there's no Tilt-managed build despite being modeled as an ImageTarget
+	// 	in the engine. From the API perspective, this is fine: there is not
+	// 	actually a dependency between KubernetesApply<>LiveUpdate API types,
+	// 	but this breaks engine invariants because it expects the build DAG to
```
nicks, line 63: not blocking on this PR but - how hard would it be to skip the DAG check on Kubernetes objects if live-update-v2 is on?

my goal is that as soon as docker-compose is on live-update-v2, all the old live-update-v1 checks will go away

nicks (approved):

nicks:
re: your comment above - ya, that's right, seems like the warning is coming from the live-update-v1 codepath, which ends up getting skipped anyway, and everything works ok in practice. that's probably why no one has reported this....

pkg/model/k8s_target.go
milas: That should be trivial - can just check that all the `ImageTarget`s have `LiveUpdateReconciler == true`!

I think we'll still need this hack in `DockerComposeTarget` since it's always v1 reconciler, right? (Hopefully no K8s users are turning it off)

milas: (the above diff _should_ be sufficient to replace this PR and work with both v1/v2)

nicks: ya that's fine too

milas: Rebased this PR to use that technique, way less code and easier to get rid of in the future 🗑️

nicks (approved):

Issue #5455, opened 2022-02-05
Tilt up `--hud` looks broken


kensipe:
### Expected Behavior
the cmd line docs for `tilt up --hud` seem to indicate it is default true... which should be the default `tilt up`.  however it looks identical to `tilt up --legacy`.    In addition, I would expect a validation error for `tilt up --legacy --hud` as those should be mutually exclusive.

### Current Behavior
Currently `--hud` looks like `--legacy` and not like the default.

### Steps to Reproduce
1. `tilt up --hud`

### Context
This is off HEAD of master... sha at this moment:  057bf28ad6f4e00a80addbe304f128d1439d4513

#### `tilt doctor` Output
```
Tilt: v0.24.0-dev, built 2022-02-05
System: darwin-arm64
---
Docker
- Host: [default]
- Server Version: 20.10.12
- API Version: 1.41
- Builder: 2
- Compose Version: v1.29.2 (build 5becea4c)
---
Kubernetes
- Env: kind-0.6+
- Context: kind-kind
- Cluster Name: kind-kind
- Namespace: default
- Container Runtime: containerd
- Version: v1.21.1
- Cluster Local Registry: none
---
...
```

kensipe:
perhaps this isn't a bug... not sure but it seems awkward.  `--hud=false` gives the default behavior.  `--hud` and `--legacy`  seem to be the same display.    It is really unclear why there is a distinction for `hudFlagExplicitlySet`.      The code is written such that `store.TerminalModeHUD` is used for `--hud` and `--legacy` which is really confusing.  The fact the `hud` is by default true but is for some reason treated different if set explicit doesn't seem like a great user experience.    The fact that hud and legacy open the same terminal is very confusing.   Looking for thoughts and input... happy to fix if we have clear expectations.

landism:
Oy, yes, that is a confusing mess!

1. `--hud` should default to `false`, and we should indicate that it's deprecated and replaced by `--legacy`.
2. The `hudFlagExplicitlySet` was (supposedly temporary) code to allow testing when we were changing the default behavior away from hud, and should be removed.

Does that clear things up?

kensipe:
@landism I believe so...  coding it make it more clear.   I will mod on this PR and remove  the `explicit` check.  on it!

kensipe:
oops... this PR == #5456

kensipe:
[...]

PR #5456, opened 2022-02-05
Flags --hud and --legacy mutually exclusive

kensipe:
Improving UX by not allowing nonsensical combination of flags

As reported in #5455, it would be nice to remove the notion of 2 different modes for `--hud=true`.   It is true by default and it can be true by setting. Those 2 different conditions of being true have different experiences.   I would love to fix that if we can get some direction.   It is also worth pointing out that the terminal has 2 letters to go to legacy.   one that is explicitly displayed as an option `t` -- legacy terminal... but also one that looks to be historical as written in the [blog](https://blog.tilt.dev/2020/06/19/the-right-display-for-now.html) `h` which goes to hud.   This increases the confusion as hud and legacy seem to be called out separately but also seem to mean the same thing.   Clarity could make this PR meaningless.


kensipe:
failure is a timeout on linting

kensipe:
taking a different approach based on feedback... closing this PR and will create another based on the guidance provided

PR #5457, opened 2022-02-07
disable: DC: don't spawn goroutines for work that's already in progress

landism:
### Problem

When a DC resource is running and configured to be disabled, the DC disable watcher will spawn a goroutine to `docker-compose rm` it. This command can take 5-10 seconds. In the meantime, every call to the watcher's `OnChange` will spawn a new goroutine.

### Solution

Track which resources currently have an `rm` in progress, and only start cleanup goroutines when there are resources that both need an `rm` and do not already have one.

### DIFF
--- a/internal/engine/dcwatch/disable_watcher.go
+++ b/internal/engine/dcwatch/disable_watcher.go
@@ -28,11 +28,16 @@ type DisableSubscriber struct {
 	// (the subscriber will continue reporting that the resource needs cleanup until we successfully kill the
 	// container and the DC event watcher notices and updates EngineState)
 	lastDisableStartTimes map[string]time.Time
+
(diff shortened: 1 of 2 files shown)


internal/engine/dcwatch/disable_watcher.go
```diff
@@ -77,29 +83,43 @@ func (w *DisableSubscriber) OnChange(ctx context.Context, st store.RStore, summa
...
 		w.mu.Unlock()
 	}
 
 	st.RUnlockState()
 
-	if anyNeedCleanup {
+	if len(toCleanup) > 0 {
 		go func() {
+			w.mu.Lock()
+			w.goroutinesSpawnedForTesting += 1
```
nicks, line 107: you probably want this to be outside the goroutine? then you don't need the lock here

PR #5461, opened 2022-02-07
backend: add cancel update buttons

landism:
1. For each resource created by a Tiltfile, the Tiltfile reconciler adds a UIButton to cancel builds for that resource.
2. The build controller watches that button and cancels builds as appropriate.

This is hidden behind a feature flag pending a frontend PR for styling. Until that's done, this works fine and displays as a normal, user-added UIButton in the web ui. We could probably just ship with a bit of icon/text tweaking, but the frontend work should be pretty quick.

### Notes
* I'm kind of on the fence between a) one cancel button per manifest (like in this PR), b) one cancel button per build (have the build controller create the UIButton when starting a build and remove it when it's done). The latter hadn't occurred to me until I'd already implemented the former, and I haven't come up with any super compelling arguments either way. I think it'd be pretty easy to change, if anyone wants to make the case.
* We'll probably want a `CancelOnSpec` at some point, but I'm not sure it's worth it before we put this stuff in the API.


internal/controllers/core/tiltfile/api.go
```diff
@@ -342,6 +344,34 @@ func toToggleButtons(tlr *tiltfile.TiltfileLoadResult, disableSources disableSou
...
+				Annotations: map[string]string{
+					v1alpha1.AnnotationButtonType: v1alpha1.ButtonTypeCancelUpdate,
+				},
+			},
+			Spec: v1alpha1.UIButtonSpec{
+				Location: v1alpha1.UIComponentLocation{
+					ComponentID:   m.Name.String(),
+					ComponentType: v1alpha1.ComponentTypeResource,
+				},
+				Text:     "cancel",
```
milas, line 366: cancel -> Cancel?

internal/controllers/core/uibutton/reconciler.go
```diff
@@ -4,28 +4,31 @@ import (
...
 	"github.com/tilt-dev/tilt/internal/hud/server"
+	"github.com/tilt-dev/tilt/internal/store"
+	"github.com/tilt-dev/tilt/internal/store/uibuttons"
 	"github.com/tilt-dev/tilt/pkg/apis/core/v1alpha1"
 )
 
 type Reconciler struct {
 	client ctrlclient.Client
 	wsList *server.WebsocketList
+	store  store.RStore
```
milas, line 22: FYI there's a `store.Dispatcher` iface that's bound with wire - not critical, but I've been using it to avoid the temptation for a reconciler to start actually mutating the store 🤪 

https://github.com/tilt-dev/tilt/blob/449fe2dbcc77c7bfacea82d1b0e14f93c4b0194a/internal/store/logger.go#L71-L73

internal/engine/buildcontroller.go
```diff
@@ -143,15 +149,21 @@ func (c *BuildController) buildAndDeploy(ctx context.Context, st store.RStore, e
...
 	state := st.RLockState()
 	defer st.RUnlockState()
 
 	for _, ms := range state.ManifestStates() {
-		if ms.DisableState == v1alpha1.DisableStateDisabled {
+		disabled := ms.DisableState == v1alpha1.DisableStateDisabled
+		canceled := false
+		if cancelButton, ok := state.UIButtons[uibutton.CancelButtonName(ms.Name.String())]; ok {
+			lastCancelClick := cancelButton.Status.LastClickedAt
+			canceled = !lastCancelClick.Time.Before(ms.CurrentBuild.StartTime)
```
milas, line 164: Also not critical given how it's used here - just sharing in case you haven't seen before, the `timecmp` package can be useful when dealing with the API times, particularly if comparing to stdlib times (see package doc for long-winded explanation)

https://github.com/tilt-dev/tilt/blob/449fe2dbcc77c7bfacea82d1b0e14f93c4b0194a/internal/timecmp/cmp.go#L58-L66

internal/controllers/core/uibutton/reconciler.go
```diff
@@ -44,6 +47,9 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 		return ctrl.Result{}, nil
```
milas, line 47: I think you're missing a dispatch for the `UIButtonDeleteAction`  in this block

milas (approved):
RE: Manifest vs per-build, your approach makes sense to me. I'm struggling to think of a valid use case for canceling an individual part of the "build pipeline" for a manifest, so not sure if there'd be a tangible benefit to that. (Also, from a purely UX perspective, having buttons popping in and out seems icky)

internal/engine/buildcontroller.go
landism: oh, nice

internal/controllers/core/uibutton/reconciler.go
landism: 😬

PR #5462, opened 2022-02-07
dockercompose: accept `env_file` argument

### LINKED ISSUE #5265: --env-file support for docker-compose projects
Docker Compose lets you manually specify the env-file on the CLI
https://docs.docker.com/compose/environment-variables/#using-the---env-file--option

The docker_compose function should support this as well

**Additional context**
Related issue (but very different): https://github.com/tilt-dev/tilt/issues/2643 (this is more about having ways to integrate the .env file with tilt configuration)

nicksieger:
Fixes #5265.

### DIFF
--- a/internal/dockercompose/client.go
+++ b/internal/dockercompose/client.go
@@ -39,7 +39,6 @@ var dcProjectOptions = []compose.ProjectOptionsFn{
 	compose.WithResolvedPaths(true),
 	compose.WithNormalization(true),
 	compose.WithOsEnv,
-	compose.WithDotEnv,
 }
 
 type DockerComposeClient interface {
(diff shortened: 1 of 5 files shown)


internal/dockercompose/client.go
```diff
@@ -39,7 +39,6 @@ var dcProjectOptions = []compose.ProjectOptionsFn{
 	compose.WithResolvedPaths(true),
 	compose.WithNormalization(true),
 	compose.WithOsEnv,
-	compose.WithDotEnv,
```
nicks, line 42: mostly for my curiosity - why did this change? since you seem to add it down below

nicks (approved):

internal/dockercompose/client.go
nicksieger: The `WithDotEnv` function is what actually applies and reads the file and loads it into the `Environment` option as key-value pairs. The default filename is `.env` in the current directory. `WithEnvFile` changes the filename, but the `WithDotEnv` needs to follow specifying the filename in the slice of option functions to read it.

nicksieger: See
- https://github.com/compose-spec/compose-go/blob/master/cli/options.go#L175
- https://github.com/compose-spec/compose-go/blob/master/cli/options.go#L183

PR #5466, opened 2022-02-08
web: unify overview + sidebar trigger button logic

landism:
Our overview + sidebar trigger button implementations were pretty much completely separate.

When implementing the cancel build button, I started continuing this duplication between overview + sidebar, but felt like it was getting too messy.

This PR combines the logic of the two trigger buttons, but keeps the styling separate (we could probably unify that a bit as well, but that can be a separate PR, and isn't as annoying right now).


web/src/OverviewTableColumns.tsx
```diff
@@ -366,7 +366,6 @@ export function TableTriggerColumn({ row }: CellProps<RowValues>) {
       isBuilding={trigger.isBuilding}
       triggerMode={row.values.triggerMode}
       isQueued={trigger.isQueued}
-      resourceName={row.values.name}
```
landism, line 369: it turns out this was only used to implement `onClick`, but we already have `onTrigger` a few lines down for that purpose, and we were ignoring it!

web/src/SidebarItemView.tsx
```diff
@@ -366,7 +365,6 @@ export function EnabledSidebarItemView(props: SidebarItemViewProps) {
               {hasSuccessfullyDeployed ? timeAgo : "—"}
             </SidebarItemTimeAgo>
             <SidebarTriggerButton
-              isTiltfile={item.isTiltfile}
```
landism, line 369: this was unused - a relic from when the tiltfile didn't have a trigger button

web/src/OverviewTableTriggerButton.tsx
```diff
@@ -46,63 +41,3 @@ export let TriggerButtonRoot = styled(InstrumentedButton)`
...
-    if (props.hasPendingChanges && isManual) {
-      isEmphasized = true
-    } else if (!props.hasBuilt && !isAutoInit) {
-      isEmphasized = true
-    }
-  }
-
-  let classes = []
-  if (!clickable) {
-    classes.push("is-disabled")
```
nicks, line 86: hmmmm...what about these css classes? looks like they don't exist in the new class, but they're still used in the styled component above

lizzthabet (approved):
overall lgtm! i'm happy to see the code condensed. 👏🏻  

i noticed the same thing as nick, which is the missing classes. when i tested locally and compared to two small inconsistencies on table view:
* the style for manual, auto-init false resources that have yet to be triggered is different
Screen Shot 2022-02-10 at 3 38 04 PM
* the disabled styling (with the `not-allowed` cursor and title on the icon) are missing when a resource has been triggered and is building
what the trigger buttons look like now:
Screen Shot 2022-02-10 at 3 30 38 PM
what they should look like:
Screen Shot 2022-02-10 at 3 52 10 PM

surbhi is welcome to chime in too, if we just want to align the detail view and table view manual + auto-init false + not yet triggered styles. if we do that, my only recommendation would be making sure that the padding is the same between those two icons, so they align in the row.

web/src/OverviewTableTriggerButton.tsx
landism: fixed, thanks

landism:
I've cleaned up the border issue (yay mui--we should find a more systemic way of interacting with mui's styles) and classnames.

The other changes are because the old buttons had different mouseover behavior when disabled: the table button got a not-allowed cursor and a tooltip, but still sent http requests on clicks!, and the sidebar button got `pointer-events: none`, which stopped all that. Setting `disabled` and stopping the actual clicks seems desirable, so I've added cursor and tooltip to all the trigger buttons via a wrapper element instead (also I prefer real tooltips to button titles anyway).

lizzthabet (approved):
tested locally and it looks great! i think the tooltips are a better way to communicate the build/trigger status information.

the only thing i noticed is that detail view colors the svg blue after a manual, auto-init=false resource gets triggered for the first time, but table view does not. so there are three different (non-active/hover/focus) states for the trigger button on detail view, but only two states on table view.

Screen Shot 2022-02-14 at 5 20 59 PM

Screen Shot 2022-02-14 at 5 21 31 PM

definitely not a merge blocker, but would be nice to fix!

PR #5470, opened 2022-02-09
Deprecate --hud in favor of --legacy

kensipe:
Improving UX and reducing confusion around `hud` and `legacy` in favor of `legacy` as noted in  Issue #5455 
Internally, it is still referenced as hud or legacy hud, but it exposed to users as `legacy`.  

The change is much easier to reason about on startup of `up` or `demo`.
`hud` is marked as deprecated which requires a sleep pause in order to display.  1 sec seemed too fast to notice and guessed 3 was reasonable but looking for feedback.
Additionally I guess that 3 major release versions was a reasonable deprecation window, looking for guidance there.

This PR does not handle consolidation of key navigation from prompts yet.

### DIFF
--- a/Tiltfile
+++ b/Tiltfile
@@ -34,7 +34,7 @@ def go_vendor():
 
 all_go_files = get_all_go_files(".")
 
-go("Tilt", "cmd/tilt/main.go", all_go_files, srv="cd /tmp/ && ./tilt up --hud=false --web-mode=prod --port=9765")
+go("Tilt", "cmd/tilt/main.go", all_go_files, srv="cd /tmp/ && ./tilt up --legacy=false --web-mode=prod --port=9765")
 go_test_changes(all_go_files)
 go_lint(all_go_files)
 go_vendor()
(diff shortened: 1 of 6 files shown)


kensipe:
didn't run integration tests... looking at the failure and will update

kensipe:
may need some help understand the integration env.   The first failure was definitely related to the change... I don't see how the current failure is related.   I'm not sure what is need to run the integration test locally and will deep later tonight

nicksieger:
@kensipe we could probably expound upon what [this line means](https://github.com/tilt-dev/tilt/blob/master/integration/README.md?plain=1#L5) in the README.

AFAIK the requirements can be satisfied by:
1. Kind, Ctlptl, Kubectl installed locally
2. `ctlptl create cluster kind --registry=ctlptl-registry`
3. `make integration` or `go test ./integration/...` the specific tests you want to run

kensipe:
Thank you!   long time k8s eng, first time `ctlptl` user...   methinks melike

kensipe:
Seems like all check pass and this is ready to go if the team sees value in it.

kensipe:
For clarity for others that may be reading this... to run integration without make:  `go test --tags 'integration' ./integration/...`

nicks (approved):
LGTM, thanks!

PR #5471, opened 2022-02-09
Add Tiltfile.loaded contents.changed tag

nicksieger:
Adds a `contents.changed` tag to the `tiltfile.loaded` metric that reports true if any of the executed files' contents changed since the last execution.

- tiltfile: pre-read file and pass contents to exec plugins
- tiltfile: hasher plugin to sum sha256 of exec'd files
- apis: add TiltfileHashes to TiltfileStatus
- tiltfile: emit contents.changed tag and report updated status

cc: @milas @nicks


internal/controllers/core/tiltfile/reconciler.go
```diff
@@ -493,12 +493,17 @@ type runStatus struct {
 }
 
 func (rs *runStatus) TiltfileStatus() v1alpha1.TiltfileStatus {
+	var hashes *v1alpha1.TiltfileHashes
+	if rs.tlr != nil && rs.tlr.Hashes.AllFilesSHA256 != "" {
```
nicks, line 497: is there a reason to publish this to the apiserver? seems like it's internal state to the Tiltfile reconciler, and not something other reconcilers need to know about.

i would really like to avoid passing TiltfileStatus to the tiltfile loader at all

nicksieger: No, there isn't. I was looking for a clean way to pass previous state into `TiltfileLoader.Load(...)`. I can modify the signature there and store the state in the reconciler runStatus instead with an internal-only struct, would that work for you?

nicks: ya, i'd be fine with passing some sort of `lastTiltfileLoadResult` struct or `lastLoadInfo` struct

nicksieger: Or maybe just the previous TiltfileLoadResult. I think that would work well.

internal/controllers/core/tiltfile/reconciler.go
```diff
@@ -304,11 +311,8 @@ func (r *Reconciler) run(ctx context.Context, nn types.NamespacedName, tf *v1alp
 	}
 
 	r.mu.Lock()
```
nicksieger, line 313: Is this mutex lock still needed now that the map lookup is gone?

nicks (approved):

PR #5473, opened 2022-02-09
api: make tiltfile builds cancelable

landism:
Tiltfile cancelation is implemented via a new `CancelOnSpec` type. This is highly similar to `StartOnSpec`/`RestartOnSpec`, but sticking that on top of `StartOn` and `RestartOn` in the "restarton" package started feeling a little ugly, so:

* The first commit renames the "restarton" stuff (i.e., the stuff that handles StartOnSpec and RestartOnSpec, but not RestartOnSpec itself) to "trigger" and moves some things inside structs to reduce the impact of adding CancelOnSpec as a third friend in that group. Unfortunately github didn't recognize this as a rename. I made a throwaway branch without the file renames [here](https://github.com/tilt-dev/tilt/commit/5c35018f1652743ce02c1ea1279575efc495bbd1) to generate an actual diff for those files.
  * I'm not super excited to the name "trigger"--it's just the best I came up with after some struggle. (as in, it handles checking trigger conditions for events)
* The second commit just adds `CancelOn`.
* The third commit does all the actual Tiltfile work (creating and watching the button).


pkg/apis/core/v1alpha1/trigger_types.go
```diff
@@ -23,3 +23,13 @@ type StartOnSpec struct {
 	// UIButtons that can trigger a start/restart.
 	UIButtons []string `json:"uiButtons" protobuf:"bytes,2,rep,name=uiButtons"`
 }
+
+type CancelOnSpec struct {
+	// StartAfter indicates that events before this time should be ignored.
+	//
+	// +optional
+	CancelAfter v1.Time `json:"cancelAfter,omitempty" protobuf:"bytes,1,opt,name=cancelAfter"`
```
nicks, line 31: are you sure we need this? it wasn't necessary for RestartOn, and i thought we rejiggered the StartOn implementation to make this obsolete

pkg/apis/core/v1alpha1/trigger_types.go
```diff
@@ -23,3 +23,13 @@ type StartOnSpec struct {
 	// UIButtons that can trigger a start/restart.
 	UIButtons []string `json:"uiButtons" protobuf:"bytes,2,rep,name=uiButtons"`
 }
+
+type CancelOnSpec struct {
```
nicks, line 27: maybe we should call this StopOn, to mirror StartOn?

pkg/apis/core/v1alpha1/trigger_types.go
landism: nope, we don't!

pkg/apis/core/v1alpha1/trigger_types.go
landism: 👍

internal/controllers/apis/restarton/restarton.go
```diff
@@ -138,15 +138,15 @@ func FileWatches(ctx context.Context, client client.Reader, restartOn *v1alpha1.
 //
 // Returns the most recent trigger time. If the most recent trigger is a button,
 // return the button. Some consumers use the button for text inputs.
-func LastStartEvent(startOn *v1alpha1.StartOnSpec, restartObjs Objects) (time.Time, *v1alpha1.UIButton) {
+func LastStartEvent(startOn *v1alpha1.StartOnSpec, triggerObjs Objects) (time.Time, *v1alpha1.UIButton) {
```
milas, line 141: Should this file/package be deleted?

milas (approved):
Nice refactor and fwiw I actually like the new `trigger` name 🙃

landism:
I've broken StartOn/RestartOn/StopOn apart.
I've also combined FetchObjects with Last{Start,Restart,Stop}EventTime, since the latter was never called without the former.

nicks (approved):

PR #5475, opened 2022-02-10
web: add a show/hide toggle for disabled resources

lizzthabet:
This PR adds a "Show disabled resources" option to Detail View and Table View, so people can opt into viewing disabled resources. Like the alerts and resource name filter, it persists in session storage. The option will only display if the `"disable_resources"` feature is enabled.

I also refactored how Detail View and Table View loop over resources to apply options and feature flags to minimize how many iterations the component is doing.

_Side note:_ I've found writing tests with react-testing-library is much easier than enzyme! But refactoring entire test suites takes time. This PR has a combo of new tests written with react-testing-library and new enzyme tests that were easy to copy-paste from existing tests. If this is confusing or undesirable, I'm happy to address, but for speed, I think this is okay.

Screen Shot 2022-02-09 at 7 08 45 PM

### DIFF
--- a/web/src/OverviewSidebarOptions.test.tsx
+++ b/web/src/OverviewSidebarOptions.test.tsx
@@ -11,7 +11,7 @@ import {
   TwoResourcesTwoTests,
 } from "./OverviewResourceSidebar.stories"
 import {
-  AlertsOnTopToggle,
+  CheckboxToggle,
   OverviewSidebarOptions,
 } from "./OverviewSidebarOptions"
 import {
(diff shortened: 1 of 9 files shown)


web/src/OverviewSidebarOptions.test.tsx
```diff
@@ -70,37 +74,6 @@ describe("overview sidebar options", () => {
...
-    )
-
-    assertSidebarItemsAndOptions(
-      root,
-      ["beep", "boop"],
-      DEFAULT_OPTIONS.alertsOnTop,
-      "B p"
-    )
-  })
-
```
lizzthabet, line 103: these tests cases should already be covered in `SidebarResources.test.tsx`

web/src/OverviewTable.test.tsx
```diff
@@ -746,6 +758,39 @@ describe("when disable resources feature is NOT enabled", () => {
   })
 })
 
+describe("when disable resources feature is enabled, but `showDisabledResources` option is false", () => {
+  let view: TestDataView
+  let wrapper: ReactWrapper<OverviewTableProps, typeof OverviewTable>
+
+  beforeEach(() => {
```
nicks, line 765: seems odd to use `beforeEach` when there's only 1 test

PR #5479, opened 2022-02-10
Update vendor

nicksieger:
- vendor: containerd/continuity/fs is gone as of b509b25e3
- golangci-lint: update due to TFL signature change


nicksieger:
@landism our PRs merged past each other in the night and I left you with a signature change that's needed for TiltfileLoader's. Also includes a go mod vendor update that has been around since b509b25e3.

PR #5483, opened 2022-02-11
web: proposal - add a small show more button in each section with more than 20 items

nicks:
Hello @lizzthabet,

Please review the following commits I made in branch nicks/window:

5ed07433fef40c85a44f5c48440d1b2a173ea7e6 (2022-02-11 11:12:23 -0500)
web: proposal - add a small show more button in each section with more than 20 items

### DIFF
--- a/web/src/SidebarResources.tsx
+++ b/web/src/SidebarResources.tsx
@@ -34,6 +34,7 @@ import SidebarItemView, {
   SidebarItemRoot,
 } from "./SidebarItemView"
 import SidebarKeyboardShortcuts from "./SidebarKeyboardShortcuts"
+import SrOnly from "./SrOnly"
 import {
   AnimDuration,
   Color,
@@ -203,7 +204,7 @@ const defaultMaxItems = 20
 
 const ShowMoreRow = styled.li`
   margin: ${SizeUnit(0.5)} ${SizeUnit(0.5)} 0 ${SizeUnit(0.5)};
-  color: ${Color.grayLight};
+  color: ${Color.gray7};
   font-size: ${FontSize.small};
   display: flex;
   align-items: center;
@@ -214,7 +215,7 @@ const ShowMoreRow = styled.li`
 const ShowMoreButton = styled(InstrumentedButton)`
   ${mixinResetButtonStyle};
   font-size: ${FontSize.small};
-  color: ${Color.offWhite};
+  color: ${Color.gray6};
   transition: color ${AnimDuration.default} ease;
   cursor: pointer;
   padding: 0 0.5em;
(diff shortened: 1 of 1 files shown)


nicks:
cc @SurbhiSGupta 

Screenshot from 2022-02-11 11-10-41

This is very much a "Minimum Viable Fix" UI - something that we can push out quickly without changing the "default" experience or introducing much complexity. If a section of the sidebar has more than 20 elements, cap it and add a little button. It might make sense to make the cap even smaller - like 10 elements.

wdyt?

SurbhiSGupta:
Love the quick fix !
Here is an option to make it even lesser wordy
image

nicks:
Screenshot from 2022-02-11 14-28-56

nicks:
Screenshot from 2022-02-11 14-30-07

web/src/SidebarResources.tsx
```diff
@@ -190,10 +199,65 @@ export function SidebarListSection(props: SidebarSectionProps): JSX.Element {
...
+  padding: 0 0.5em;
+
+  &:hover {
+    color: ${Color.blue};
+  }
+`
+
 function SidebarListSectionItems(props: SidebarSectionProps) {
+  let [maxItems, setMaxItems] = useState(defaultMaxItems)
+  let items = props.items
```
lizzthabet, line 229: (tiny nit) could you rename this var so it doesn't echo the prop name, something like `itemsToDisplay`? it took me a second to realize this was a var that's potentially getting reassigned elsewhere, rather than the prop data.

lizzthabet (approved):
this looks good to me! i tested with my "hundreds of resources" example, with and without labels, and it worked well.

do you feel strongly about the 20 max items? it's an easy threshhold to change if people complain about it, though i'm mindful that the "show more" display doesn't persist if you switch views, so i could see someone getting annoyed if they want to view a resource that doesn't make the cut off. 

i was just looking at the "Resource counts in last 28 days for known teams" chart we have to get a sense of how that 20 threshhold relates. about half of the teams run more than 20 resources at once on average, but because that doesn't consider how many resources are in each label group, that may not be a helpful metric!

these are more general thoughts than specific / actionable feedback.

web/src/SidebarResources.tsx
```diff
@@ -203,6 +267,7 @@ function SidebarListSectionItems(props: SidebarSectionProps) {
           resourceView={props.resourceView}
         />
       ))}
+      {showMoreItemsButton}
```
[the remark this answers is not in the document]
nicks: i don't think that would be an accurate use of describedby?

updated the SrOnly text and colors!

web/src/SidebarResources.tsx
nicks: done

nicks (commented):
To speak to the high-level question - nope, I don't have a strong attachment to max=20

for what it's worth - as the set of resources gets more heterogenous (with clusters and registries and k8s secrets and k8s servers and local jobs all being presented as "sidebar items"), I suspect that the sidebar won't even make sense anymore as a way to navigate them...

Issue #5482, opened 2022-02-11
Use both Docker Compose and K8s at the same time


nicksieger:
## Describe the Feature You Want
To support a gradual migration from DC to K8s, Tilt should allow both DC-deployed containers and K8s-deployed containers to coexist.

### Current Behavior
Tilt currently forbids mixing DC and K8s in the same session

### Why Do You Want This?
[People on K8s slack have asked for this functionality](https://kubernetes.slack.com/archives/CESBL84MV/p1644460651626619). Perhaps they can provide more context. The assumption so far is that big-bang conversion from DC to K8s [as described in this blog post](https://blog.tilt.dev/2019/09/16/tips-on-moving-your-dev-env-from-docker-compose-to-kubernetes.html) is less desirable.

**Additional context**
This might be more of a feasibility study of whether this makes sense to do with the current state of Tilt's implementation, as well as whether future work on clusters might make this easier to do later, or make it obsolete.

PR #5484, opened 2022-02-11
docker: allow simultaneous k8s and compose orchestration

nicksieger:
This is an experimental/spike on supporting K8s and DC orchestration simultaneously for #5482. I took a naive approach and am probably overlooking some issues, but I thought I'd put it up for review in any case.

With these changes, I was able to get simultaneous deploys working for the tilt-example-docker-compose project and a Kind cluster. I suspect using Minikube/Microk8s with a separate cluster Docker daemon might still cause issues.
https://github.com/tilt-dev/tilt-example-docker-compose/blob/nicksieger/dc-and-k8s/Tiltfile

### DIFF
--- a/internal/controllers/core/tiltfile/reconciler.go
+++ b/internal/controllers/core/tiltfile/reconciler.go
@@ -309,8 +309,10 @@ func (r *Reconciler) run(ctx context.Context, nn types.NamespacedName, tf *v1alp
 		tlr.Error = fmt.Errorf("No resources found. Check out https://docs.tilt.dev/tutorial.html to get started!")
 	}
 
-	if tlr.Orchestrator() != model.OrchestratorUnknown {
-		r.dockerClient.SetOrchestrator(tlr.Orchestrator())
+	if tlr.HasOrchestrator(model.OrchestratorK8s) {
+		r.dockerClient.SetOrchestrator(model.OrchestratorK8s)
+	} else if tlr.HasOrchestrator(model.OrchestratorDC) {
+		r.dockerClient.SetOrchestrator(model.OrchestratorDC)
 	}
 
 	if requiresDocker(tlr) {
@@ -435,7 +437,7 @@ func (r *Reconciler) enqueueTriggerQueue(obj client.Object) []reconcile.Request 
 // Otherwise, we'll return the zero value of `s.defaultReg`, which is an empty registry.
 // It has side-effects (a log line) and so should only be called once.
 func DecideRegistry(ctx context.Context, kCli k8s.Client, tlr *tiltfile.TiltfileLoadResult) container.Registry {
-	if tlr.Orchestrator() != model.OrchestratorK8s {
+	if !tlr.HasOrchestrator(model.OrchestratorK8s) {
 		return tlr.DefaultRegistry
 	}
 
@@ -466,7 +468,7 @@ func (r *Reconciler) defaultK8sConnection() *v1alpha1.KubernetesClusterConnectio
 }
 
 func requiresDocker(tlr tiltfile.TiltfileLoadResult) bool {
-	if tlr.Orchestrator() == model.OrchestratorDC {
+	if tlr.HasOrchestrator(model.OrchestratorDC) {
 		return true
 	}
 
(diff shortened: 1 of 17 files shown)


internal/controllers/core/tiltfile/reconciler.go
```diff
@@ -309,8 +309,10 @@ func (r *Reconciler) run(ctx context.Context, nn types.NamespacedName, tf *v1alp
 		tlr.Error = fmt.Errorf("No resources found. Check out https://docs.tilt.dev/tutorial.html to get started!")
 	}
 
-	if tlr.Orchestrator() != model.OrchestratorUnknown {
-		r.dockerClient.SetOrchestrator(tlr.Orchestrator())
+	if tlr.HasOrchestrator(model.OrchestratorK8s) {
```
nicks, line 312: have you seen this code?
https://github.com/tilt-dev/tilt/blob/master/internal/cli/doctor.go#L54

what might be good is to have a little bit of code to map orchestrators to docker clients.

if both orchestrators use the same docker client, then it's ok to use them both at the same time.

does that make sense?

nicksieger: Nope, haven't seen it! Thanks for the pointer.

My changes to the "switch" docker client were an attempt to deal with the multiple docker clients without fully understanding the consequences there. It also seems like with multiple Docker clients the image builder needs to be aware of where the image is going to be deployed in order to choose the correct client, and I haven't chased that down yet other than forcing anything going through the DockerComposeBuildAndDeployer to use the local docker client.

This could also end up being another learning PR for me, as a way of exploring obstacles to correctly using multiple docker clients.

nicks: fwiw - i think we should try to move forward with this PR. 

allowing both DC+K8s when there's a single docker client is a good goal!

have you tested manually if we push to the right registry when you have both DC+K8s in-play? I _think_ it should work (bc DC image builds are a different codepath)

nicksieger: Yes! The example above works, the image is pushed to the Kind registry.

nicks (approved):

internal/docker/client.go
```diff
@@ -109,6 +110,15 @@ type Client interface {
 	ContainersPrune(ctx context.Context, pruneFilters filters.Args) (types.ContainersPruneReport, error)
 }
 
+// Add-on interface for a client that manages multiple clients transparently.
+type CompositeClient interface {
+	Client
+	DefaultLocalClient() Client
+	DefaultClusterClient() Client
+	ClientFor(cluster v1alpha1.Cluster) Client
```
nicksieger, line 118: Leaving this as a hint for multi-cluster. I can envision other possible methods like:
```go
AddClient(client Client, cluster v1alpha1.Cluster)
```

nicksieger:
Also, I tested this with docker-desktop, kind, and minikube clusters and all three work with the mixed-mode deployment!

internal/engine/buildcontrol/docker_compose_build_and_deployer.go
```diff
@@ -141,6 +143,10 @@ func (bd *DockerComposeBuildAndDeployer) BuildAndDeploy(ctx context.Context, st
 		ps.EndPipelineStep(ctx)
 	}
 
+	var cluster v1alpha1.Cluster
+	// If the cluster fetch fails, that's OK.
+	_ = bd.ctrlClient.Get(ctx, ktypes.NamespacedName{Name: v1alpha1.ClusterNameDocker}, &cluster)
```
milas, line 148: What actually happens if this fails?

milas (approved):
I'm sure there's going to be lots more adventures in multi-cluster / multi-docker land in the future, and I think this is going in a good direction!

internal/engine/buildcontrol/docker_compose_build_and_deployer.go
nicksieger: I copied that from [image build and deployer](https://github.com/tilt-dev/tilt/blob/master/internal/engine/buildcontrol/image_build_and_deployer.go#L157-L158), it looks like currently the cluster is only used for platform detection/injection and the builder can continue without fetching the actual cluster info, but this may be more of an error condition in the future that needs to be handled. Do you want me to mark as such?

PR #5488, opened 2022-02-11
tiltfile: unknown resource_dep is a warning instead of error

landism:
### Problem

You run your Tiltfile and get an error like "resource X specified a dependency on unknown resource Y". You know you created a resource named "Y". How do you validate that? At the moment, you have to remove your resource_deps and see what resources come out of Tilt. You're also might suspect a bug in Tilt (and you very well might be right! but there's frustratingly little to go on).

### Solution

When we generate this error, also include the list of resources Tilt knows about, so that the user can see what resources Tilt is expecting to see there and narrow down the problem.

### DIFF
--- a/internal/tiltfile/tiltfile_state.go
+++ b/internal/tiltfile/tiltfile_state.go
@@ -318,7 +318,7 @@ to your Tiltfile. Otherwise, switch k8s contexts and restart Tilt.`, kubeContext
 		manifests = append(manifests, yamlManifest)
 	}
 
-	err = validateResourceDependencies(manifests)
+	err = s.validateResourceDependencies(manifests)
 	if err != nil {
 		return nil, starkit.Model{}, err
 	}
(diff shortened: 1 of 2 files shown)


landism:
> have you considered using the ClosestN function to limit the number of resources to print? (https://github.com/tilt-dev/tilt/blob/master/internal/tiltfile/build_index.go#L106). Seems like this is going to get chatty if there are a lot of resources.

I'd thought about it, but we have the same behavior if you `tilt up foo` and foo doesn't exist, and hadn't noticed that being a problem, and that seems *way* more likely to occur (though that also seems like the kind of annoyance users wouldn't bother to report). If we've got it handy, though, we might as well.

> what if we made this a warning? then you could use tilt get uiresources to find the resource name, as we suggest here: https://docs.tilt.dev/resource_dependencies.html

Oh, that's much better!

nicks (approved):

PR #5497, opened 2022-02-15
web: Rename TriggerButton to BuildButton

landism:
1. Rename TriggerButton to BuildButton (mostly- in the next PR, BuildButton will display either a StartBuildButton or StopBuildButton, depending on the build state, so in some places I've renamed "Trigger" to "StartBuild", as appropriate)
2. Rename the SidebarTriggerButton tests to just BuildButton tests
3. Get all the SidebarResources out of BuildButton tests (and they probably shouldn't have been in there in the first place) (the PR didn't detect the rename, but you can see this diff [here](https://github.com/tilt-dev/tilt/commit/5b5168d265b80e0194acce9eeb24df77d9d57377)


lizzthabet (approved):
lgtm! especially moving sidebar items out of the trigger button tests. 🎉 

(warning: potential bike shedding!) i don't have any opinions about what the now-build button is called in the code, but i'm not 100% convinced that the `"Build"` table header makes more sense than `"Trigger"`. from what i understand, this button could trigger a build, sync changes, or run a local resource. "trigger" does capture more of those cases in my mind, and it relates to "trigger mode." i'll defer to you and others to make this call, but at least wanted to note my thoughts. definitely not blocking for this pr.

landism:
> but i'm not 100% convinced that the `"Build"` table header makes more sense than `"Trigger"`. from what i understand, this button could trigger a build, sync changes, or run a local resource. "trigger" does capture more of those cases in my mind, and it relates to "trigger mode." i'll defer to you and others to make this call, but at least wanted to note my thoughts. definitely not blocking for this pr.

Oh, good point! I still had my head stuck in "'trigger' is the verb we use to indicate starting a build", but that's not where we want to be! I'll change that back for now and we can discuss separately. It's outside the scope of this PR, anyway.

PR #5504, opened 2022-02-15
web: add a small show more button in each section with more than 20 items (try 2)

nicks:
Hello @lizzthabet,

Please review the following commits I made in branch nicks/revert2:

b2d6f0e4a9d863b91e64a0f5b1ff7cdfa9ee60d3 (2022-02-15 18:24:09 -0500)
web: add a small show more button in each section with more than 20 items (try 2)
I had to remove the SrOnly screenreader text because it was causing major layout problems

### DIFF
--- a/web/src/SidebarResources.tsx
+++ b/web/src/SidebarResources.tsx
@@ -3,10 +3,11 @@ import {
   AccordionDetails,
   AccordionSummary,
 } from "@material-ui/core"
-import React, { ChangeEvent, useMemo } from "react"
+import React, { ChangeEvent, useCallback, useMemo, useState } from "react"
 import styled from "styled-components"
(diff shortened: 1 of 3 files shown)


nicks:
Here's what it looked like when I added the `SrOnly` element:

Screenshot from 2022-02-15 18-26-09

Note the big gap at the bottom.

I didn't spent a lot of time investigating why the SrOnly element behaves this way. But poked around at some a11y articles that recommended using aria-label instead these days.

lizzthabet (approved):
this makes sense to me, good catch! the mui screen-reader only class does do `position: absolute; overflow: hidden;` and perhaps that behaves strangely with the relative positioning of the `.Pane`. 🤔

PR #5505, opened 2022-02-16
lint: use consistent goimports local everywhere

milas:
Follow-up on #5453 that uses the same `-local` arg a few other
places where we have `goimports` tasks.

It also adds `integration/` to the `make goimports` task, which
was missing previously.

Lastly, this runs the latest (v0.1.9) goimports against the repo,
as there were some lingering unformatted files around.

See: https://pkg.go.dev/golang.org/x/tools@v0.1.9/cmd/goimports


nicksieger (approved):

kensipe (approved):
great catch and update!

PR #5506, opened 2022-02-16
Consolidate STDIO  choice of streaming to cli and not independently by all cmd

kensipe:
Consolidating STIO creation and choice for multiple reasons:
1. easier to reason about
2. not independently created for each need
3. more logical for tests

* branch name inspired by ghostbusters as this is dealing with streams


milas (approved):
Thanks for the cleanup/consolidation!

kensipe:
woot!  thanks @milas !

PR #5509, opened 2022-02-16
api: ensure API times used for trigger comparisons

milas:
The `trigger` package now returns `metav1.MicroTime` instead of
stdlib `time.Time` for trigger (start/restart/stop) timestamps.

There's a ripple effect for some internal state types on reconcilers
which went through the same conversion.

Comparisons are also all now made via the `timecmp` package using
the API types.

Primarily, this fixes flakiness for tests on Windows; it's fairly
unlikely that the cases here would cause unexpected behavior at
runtime because real world events (e.g. clicking a button) are not
performed instantaneously.

#### Testing
Previously, I'd pretty consistently see failures with just a few (~10) test runs for `TestCancelButton` (in `internal/engine`) on a Windows 11 machine.

After these changes, 500 runs passes:
```
PS C:\Users\User\tilt> go test ./internal/engine -run '^TestCancelButton$' -failfast -count 500
ok      github.com/tilt-dev/tilt/internal/engine        98.211s
```

### DIFF
--- a/internal/controllers/apis/trigger/trigger.go
+++ b/internal/controllers/apis/trigger/trigger.go
@@ -167,7 +167,7 @@ func LastStartEvent(ctx context.Context, cli client.Reader, startOn *v1alpha1.St
 		}
 
 		lastEventTime := b.Status.LastClickedAt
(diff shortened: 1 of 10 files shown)


landism (approved):
Thanks!

Issue #5495, opened 2022-02-15
displaying many (disabled) resources makes the ui slow


lizzthabet:
### Expected Behavior
The UI should not be affected when there are many resources.

### Current Behavior
When displaying many resources, the UI can become slow and unresponsive. Table View is noticeably slowing than Detail View. The first time I tilted up with 300 resources (10-12 enabled, rest disabled), the UI was unusably slow, and the majority of the resources showed as “enabled” while Tilt was still in the process of marking them as disabled. After the UI became unresponsive for ~3 minutes, I quit and restarted Tilt.

Once Tilt is up and running and there are log-only changes (from what I can tell), the UI is responsive, but it can quickly become sluggish again if there are non-log resource updates. Enabling or disabling a resource can sometimes take 30+ seconds, even though the uibutton update request succeeded quickly.

### Steps to Reproduce

1. Check out [debug-many-resources](https://github.com/lizzthabet/debug-many-resources) and adjust Tiltfile to generate anywhere between 75 - 300+ resources
2. `tilt up` (and if on Tilt v0.25.0+, check the "Show disabled resources" option on Detail View or Table View)
3. Interacting with the UI, like clicking on links, triggering/enabling/disabling resources, etc will be sluggish

lizzthabet:
as of [v0.25.0](https://github.com/tilt-dev/tilt/releases/tag/v0.25.0), disabled resources are hidden by default in the ui, which should address the immediate pain point here! we're working on performance improvements throughout the ui and adding display logic that'll improve how many resources are shown. 

related work includes #5475, #5477, #5478 (will go out in next release), and #5483 (will go out in next release).

lizzthabet:
closing this issue in addition to #5496, since the major performance issues have been mitigated in recent releases.

PR #5511, opened 2022-02-16
web: add pagination to resource list tables in Table View to mirror Detail View

lizzthabet:
This PR adds a `...Show More` button to each table in Table View when the resource list exceeds twenty items. It pulls out the shared code from Detail View into its own component. These changes will help mitigate the performance issues we're seeing with displaying many resources (#5495) while we investigate root causes.

Screen Shot 2022-02-16 at 3 26 28 PM

Screen Shot 2022-02-16 at 3 14 07 PM

There may be some subtleties in how we want the list truncation to work, but for now, I took the easiest approach. A couple examples I can think of:
* When you collapse and expand a resource group, does list truncation reset? (It currently does not.)
* If you view disabled resources, click "show more" in a table with many resources, hide disabled resources, and then show them again, does the list truncation reset? (It currently does not.)

[Closes ch13307](https://app.shortcut.com/windmill/story/13307/perf-mitigation-add-pagination-for-resources-in-table-view)


web/src/SidebarResources.tsx
```diff
@@ -199,58 +196,35 @@ export function SidebarListSection(props: SidebarSectionProps): JSX.Element {
...
       <ShowMoreRow>
         <ShowMoreButton
           onClick={showMore}
-          analyticsName="ui.web.sidebarShowMore"
-          analyticsTags={emptyTags}
-        >
-          …Show More
-        </ShowMoreButton>
-        <span aria-label={`${remaining} hidden`}>{`(${remaining})`}</span>
+          analyticsTags={{ type: AnalyticsType.Detail }}
```
nicks, line 224: i would put this in a global var, so it memoizes a bit better

web/src/OverviewTable.tsx
```diff
@@ -539,21 +551,63 @@ export function ResourceTableHeadRow({
...
+
+  return (
+    <ResourceTableRow style={{ height: "45px" }}>
+      <ResourceTableData colSpan={colSpan - 2} />
+      <ResourceTableData className="alignRight" colSpan={2}>
+        <ShowMoreButton
+          itemCount={itemCount}
+          currentListSize={pageSize}
+          onClick={onClick}
+          analyticsTags={{ type: AnalyticsType.Grid }}
```
nicks, line 577: i would put this in a global var, so it memoizes a bit better

web/src/OverviewTable.tsx
```diff
@@ -539,21 +551,63 @@ export function ResourceTableHeadRow({
...
+  itemCount: number
+  pageSize: number
+  onClick: (e: MouseEvent) => void
+}) {
+  if (itemCount <= pageSize) {
+    return null
+  }
+
+  return (
+    <ResourceTableRow style={{ height: "45px" }}>
```
nicks, line 570: i'd slightly prefer we have something like 

```
&.isNarrow {
  height: 45px
}
```

in the styled component, so it's a bit easier to see all the style combinations in one place

nicks (approved):

web/src/OverviewTable.tsx
lizzthabet: good point!

PR #5512, opened 2022-02-16
tests: make test fixtures responsible for tearing themselves down

landism:
TLDR - eliminate the need for all the `defer f.teardown()`s

This PR is basically the result of searching for `f.teardown()`s, making sure the underlying `f` passes its own `teardown` to `t.Cleanup`, and removing the `f.teardown`.

[here](https://gist.github.com/landism/389f6f8dee14e078d9f2ba18d9625780)'s a gist of the diff with (most of) the "defer f.teardown"s filtered out (generated via `git diff HEAD^ -I'defer t?f.[tT]ear[dD]own'`, if you want to run it locally to get proper colors, though you might have to upgrade `git` to get `-I`)

### DIFF
--- a/integration/analytics_test.go
+++ b/integration/analytics_test.go
@@ -35,6 +35,7 @@ func newAnalyticsFixture(t *testing.T) *analyticsFixture {
 
 	af.SetupAnalyticsServer()
 
+	t.Cleanup(af.TearDown)
 	return af
 }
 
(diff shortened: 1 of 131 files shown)


milas (approved):
Nice!

It might be nice to fully eliminate the `TearDown` methods so they don't get unintentionally called twice (was gonna suggest unexporting them but realized that won't make a difference since they're only used package-local to begin with 🙃), buuuuut that seems like it might be a pain, so I'm pretty indifferent as presumably the tear down methods are either idempotent and thus harmless or will 💥 if called 2x so you'd notice pretty quick

PR #5513, opened 2022-02-16
k8s: detect Rancher Desktop in-cluster builder

### LINKED ISSUE #5367: Tilt should detect Rancher Desktop to skip Image push
When Tilt can not find a local registry and no remote registry is defined, Tilt should be able to detect that Rancher Desktop is being used and skip the push step.
Since 0.7.0 of Rancher Desktop, the user is now able to decide between using containerd and dockerd as container runtime.
Currently, Tilt is able to detect Docker Desktop and some minikube configs to skip pushing images built with `docker_build` to a remote registry. We want the same behavior for Rancher Desktop.

### Current Behavior
Currently, Tilt cannot detect that the user is running Rancher Desktop and tries to push an image built with `docker_build` to docker.io

### Why Do You Want This?
Users running Rancher Desktop want to be able to use images built in their container runtime without having to go past an image registry, to reduce complexity (i.e. having to set up an image registry)

`When using Rancher Desktop and building with the RD docker context, you can deploy images directly to k8s (with imagePullPolicy: IfNotPresent) without needing to have a registry. It just uses the image in the docker cache. I could use a local registry, but I'd rather not set one up if I don't have to.`

**Additional context**

milas:
Newer versions of Rancher Desktop support running with different
container runtimes. The default is containerd, but it also supports
moby (aka OSS dockerd).

When running in the latter mode, Tilt will now appropriately detect
it, similar to some minikube configs + Docker Desktop w/ K8s, and
skip image pushes since the builds are happening directly on the
cluster container runtime socket. As a result, the images will be
already present/visible to K8s Pods.

Fixes #5367.


internal/k8s/rancher/rancher_desktop.go
```diff
@@ -0,0 +1,135 @@
...
+	ContainerRuntimeContainerd
+	ContainerRuntimeDocker
+)
+
+// openFileFunc is a hack for testing.
+type openFileFunc func(name string) (io.ReadCloser, error)
+
+var openFile = func(name string) (io.ReadCloser, error) {
+	return os.Open(name)
+}
```
nicksieger, line 31: Do you need the wrapper?
```suggestion
var openFile = os.Open
```

nicksieger (approved):
LGTM, sucks that we have to copy rancher logic for config, but hopefully they stabilize a bit more?

internal/k8s/rancher/rancher_desktop.go
milas: Ooo thanks, Go was being annoying about types here originally

milas:
> LGTM, sucks that we have to copy rancher logic for config, but hopefully they stabilize a bit more?

I think the config should remain pretty stable! I was excessively defensive here since I didn't want to fail Tilt startup in the event that it _does_ change though. If they introduce a management CLI (vs just GUI) in the future, we could switch to shelling out to that like we do with minikube. Maybe worth a convo at some point 🙃 

Sidenote: It's also been forever on the TODO list for us to migrate a bunch of this logic to https://github.com/tilt-dev/cluster so we can share it with `ctlptl` and the 🌍

internal/docker/env.go
```diff
@@ -139,6 +147,11 @@ func ProvideClusterEnv(ctx context.Context, kubeContext k8s.KubeContext, env k8s
 			// If we're running Microk8s with a docker runtime, talk to Microk8s's docker socket.
 			result.Host = microK8sDockerHost
 			result.BuildToKubeContexts = append(result.BuildToKubeContexts, string(kubeContext))
+		} else if env == k8s.EnvRancherDesktop {
+			rancherRuntime := rancher.DetermineContainerRuntime(ctx)
```
nicks, line 151: are you aware that we're already injecting  `container.Runtime` everywhere (with it "unknown" if the cluster is down)?

i'm fine with this, just seems a bit odd to have to maintain rancher-specific logic on top of the generic detection logic. or is there some advantage to having rancher-specific logic?

nicks (approved):

internal/docker/env.go
milas: Oh wow, I was very much NOT aware of that! I saw the minikube logic and wrongly assumed we needed something comparable.

Guess I can dramatically simplify this PR now 😂

nicks: ya, it used to be possible to run `microk8s` in docker or containerd mode, so the logic is vestigial from those days (they've been containerd only for a while)

PR #5514, opened 2022-02-16
web: stop build button replaces start build button when building

landism:
old: the "cancel build" button shows up as a normal user-added api button
new: the "cancel build" button replaces the "start build" button if the resource is building

image

image

This PR mostly consists of:
1) Changing some backend names from "cancel update" to "stop build" (since we're calling its complement "start build")
2) Changing ButtonSet / buttonsByComponent around a little so that we can get all the cancel buttons on table view
3) Changing BuildButton to use either the StartBuildButton or the stopBuild UIButton, depending on whether the resource is building
4) getting rid of the fixed height and width in the start-build-button svgs so that we can scale them in the code and don't have to eyeball padding to get them to line up


web/src/BuildButton.test.tsx
```diff
@@ -219,4 +232,18 @@ describe("SidebarBuildButton", () => {
       expectWithTooltip(button, BuildButtonTooltip.AlreadyQueued)
     })
   })
+
+  describe("stop builds", () => {
+    it("renders a stop button when the build is in progress", () => {
+      const root = mount(<BuildButtonTestWrapper isBuilding={true} />)
```
landism, line 238: 1) whoops, none of the existing tests had isBuilding=true, so this change didn't break any existing tests
2) I feel like there should be more tests here, but I didn't come up with any. All the underlying stuff is just standard ApiButton logic that we probably shouldn't re-test.

web/src/OverviewTableBuildButton.tsx
```diff
@@ -15,8 +20,10 @@ export const OverviewTableBuildButton = styled(BuildButton)`
   &:hover .fillStd {
     fill: ${Color.white};
   }
-  & > svg {
+  & .icon {
```
landism, line 23: The & > svg does not seem to be matching anything in master, and I failed to come up with a selector that would match it within 15 minutes, so I went with a class.

web/src/assets/svg/start-build-button.svg
```diff
@@ -1,3 +1,3 @@
-<svg width="16" height="14" viewBox="0 0 16 14" fill="none" xmlns="http://www.w3.org/2000/svg">
+<svg viewBox="-4 -4 20 20" fill="none" xmlns="http://www.w3.org/2000/svg">
```
landism, line 1: 1. if an svg sets a height / width, you can't scale it by setting height/width in css
2. Change the viewbox size to match the viewbox size of the other icons that go in the same place. This makes the icon keep appropximately the same size as it previously had, even though we're now giving it a larger height/width. (not quite, since the viewbox is now 24x24 instead of 16x14, but probably close enough?)

web/src/OverviewResourceSidebar.tsx
```diff
@@ -28,7 +29,11 @@ export default function OverviewResourceSidebar(
   let logAlertIndex = useLogAlertIndex()
   let pathBuilder = usePathBuilder()
   let resources = props.view.uiResources || []
-  let items = resources.map((res) => new SidebarItem(res, logAlertIndex))
+  let buttons = buttonsByComponent(props.view.uiButtons)
```
lizzthabet, line 32: do you think it's worth memoizing the button map?

web/src/trigger.tsx
```diff
@@ -6,6 +6,7 @@ export const BuildButtonTooltip = {
   AlreadyQueued: "Resource already queued!",
   NeedsManualTrigger: "Trigger update to sync changes",
   UpdateInProgOrPending: "Resource already updating!",
+  Stop: "Stop this update",
```
lizzthabet, line 9: (grammar nit) i'd prefer `Stop update` to align with `Trigger update` 😅

lizzthabet (approved):
this looks good! 

from the ux perspective, it'd be nice to:
* clean up the state changes when a resources is triggered, so there isn't a "flash" of the "build in-progress" state before the cancel button is rendered. (is this because the cancel button is created when a build is in progress?)


* remove the snackbar notification from cancel build button, since we don't have one for start build (and i don't think the notification is necessary)

web/src/trigger.tsx
landism: hm, maybe?

I put the "this" to make it less likely to be interpreted as "stop updates for this resource" (like, pause auto trigger), but there's a lot of context to help with that anyway.

web/src/OverviewResourceSidebar.tsx
landism: Yeah, that's probably a good idea.

landism:
> remove the snackbar notification from cancel build button, since we don't have one for start build (and i don't think the notification is necessary)

Thanks! I'd been tempted to do this but hadn't decided it was important enough. I've removed them for all special buttons (which, at the moment, is this one and enable/disable). It feels much nicer now!

> clean up the state changes when a resources is triggered, so there isn't a "flash" of the "build in-progress" state before the cancel button is rendered. (is this because the cancel button is created when a build is in progress?)

The cancel button (as in, the UIButton) always exists, but isn't shown until a build is in progress.
I assume the flash is because there's an instant when the resource is in the "trigger queue" (list of resources that have been manually triggered) before the build controller pulls it out, starts building, and that information makes it back to the frontend.
I'm not sure there's a simple state change cleanup here - the resource is legitimately transitioning from idle to queued to building, and the UI is reflecting the state it gets from the backend.

I've set `enterNextDelay` on the tooltip, which at least makes it so we only get the "queued" tooltip if we're in that state for >500ms, which seems to make it a fair bit better.

PR #5518, opened 2022-02-17
cli: better `tilt trigger` error messages

landism:
old:
```
$ tilt trigger this_resource_is_disabled
Successfully triggered update for resource: "this_resource_is_disabled"
$ tilt trigger this_resource_doesnt_exist
Request to http://localhost:10350/api/trigger failed with status "400 Bad Request": no manifest found with name 'this_resource_doesnt_exist'
```

new:
```
$ tilt trigger this_resource_is_disabled
Error: resource "this_resource_is_disabled" is currently disabled
$ tilt trigger this_resource_doesnt_exist
Error: resource "this_resource_doesnt_exist" does not exist
```

also added analytics for `tilt trigger` invocations

I didn't clean up the other errors (e.g., invalid json) since they'd indicate an internal bug rather than user error.


internal/hud/server/server.go
```diff
@@ -239,6 +240,10 @@ func (s *HeadsUpServer) HandleSetTiltfileArgs(w http.ResponseWriter, req *http.R
 	}
 }
 
+// Responds with:
+// * 200/empty body on success
+// * 200/error message in body on well-formed, unservicable requests (e.g. resource is disabled or doesn't exist)
```
nicks, line 245: seems like we should do 404 in this case? that's probably what k8s would do (e.g., trying to exec on a pod that didn't exist). (404s can still have error messages)

nicks (approved):

PR #5519, opened 2022-02-17
dc: hide 'going to remove...' message when disabling dc services

landism:
### Problem

For some reason, when we filter `docker-compose rm` through a Tilt logger, we get:
```
Stopping servantes_fortune_1 ... 
Stopping servantes_fortune_1 ... done
Removing servantes_fortune_1 ... 
Removing servantes_fortune_1 ... done
Going to remove servantes_fortune_1
```

The last line says "Going to...", which leads one to expect it's going to do something else, but nothing ever comes.

When running docker-compose directly, it shows up in a more sensible place:
```
Stopping servantes_vigoda_1 ... done
Going to remove servantes_vigoda_1
Removing servantes_vigoda_1 ... done
```

(you also might note there's only one line each for "stopping" and "removing" - they do tty tricks to go back and update the existing line, which is probably also why it looks weird in Tilt)

### Solution
I couldn't find a clean way to fix this, and the message doesn't add much value anyway, so just filter it out of the DC output:
```
Stopping servantes_fortune_1 ... 
Stopping servantes_fortune_1 ... done
Removing servantes_fortune_1 ... 
Removing servantes_fortune_1 ... done
```

Ideally we could find a way to get out of their TTY tricks, which might also save us from their duplicate "stopping" and "removing" messages, but I couldn't find one in 20 minutes, and `docker build` has a bad history of "you either get tty or non-human-readable output", which doesn't give me hope.

(do we already have a less home-rolled filtering io.Writer?)

### DIFF
--- a/internal/cli/klog.go
+++ b/internal/cli/klog.go
@@ -8,52 +8,13 @@ import (
 	"os"
 	"regexp"
 
+	"github.com/tilt-dev/tilt/internal/filteredwriter"
+
 	"k8s.io/klog/v2"
 )
 
 // https://kubernetes.io/docs/reference/kubectl/cheatsheet/#kubectl-output-verbosity-and-debugging
 var klogLevel = 0
-
-type filteredWriter struct {
-	underlying io.Writer
-	filterFunc func(s string) bool
-	leftover   []byte
-}
-
-func (fw *filteredWriter) Write(buf []byte) (int, error) {
-	buf = append(fw.leftover, buf...)
-	start := 0
-	written := 0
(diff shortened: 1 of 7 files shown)


nicks (approved):

nicks:
fwiw, it seems like what's happening is that docker-compose prints "Going to remove" on stdout and the other messages on stderr. It's probably a race condition in how the go exec library merges stdout and stderr.

Maybe we should /dev/null stdout instead? or file a bug upstream complaining about this.

PR #5525, opened 2022-02-18
analytics: report counts of enabled, local resources

landism:
### Problem

A number of UI conversations have revolved around how many resources our users are looking at. The introduction of disabled resources means that when we report the resource count to analytics, any number of those resources might be disabled, and we don't know how many the user is actually looking at.

### Solution

Explicitly report how many resources are enabled.

(also, while I was here, I noticed we were reporting counts for k8s and DC resources, but not local, and figured I might as well)


lizzthabet (approved):
lgtm! 🎉

PR #5527, opened 2022-02-22
api: fix confusing DEBUG logging for image injection

milas:
This was inside of the entity loop, so it was possible to get both
"no images injected" along with a list of injections for a single
resource if not all entities had an image injected, which was very
confusing. (Thanks @landism for pointing this out since evidently
I've forgotten how to read code 😬)

Thankfully, this is DEBUG logging so not particularly visible, but
the fix is easy enough to move the logging outside the loop. (Note
that `injectResults` is _exclusively_ used for logging, so this is
fine.)


landism (approved):
> Thanks @landism for pointing this out since evidently
I've forgotten how to read code

FWIW I was pretty far into logging + debugging + checking ridiculous hypotheses before realizing this!

PR #5535, opened 2022-02-23
web: adjust log buffering behavior to avoid screen jiggling

nicks:
Hello @hyu, @lizzthabet,

Please review the following commits I made in branch nicks/shuffle:

fe4c96119035f475758668bdec96c8164db275f9 (2022-02-23 17:08:27 -0500)
web: adjust log buffering behavior to avoid screen jiggling

### DIFF
--- a/web/src/OverviewLogPane.tsx
+++ b/web/src/OverviewLogPane.tsx
@@ -310,14 +310,20 @@ export class OverviewLogComponent extends Component<OverviewLogComponentProps> {
 
     // If we're scrolled horizontally, cancel the autoscroll.
     if (rootEl.scrollLeft > 0) {
-      this.autoscroll = false
+      if (this.autoscroll) {
+        this.autoscroll = false
+        this.maybeScheduleRender()
+      }
       return
     }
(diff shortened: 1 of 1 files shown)


nicks:
this has bothered me for a while!

lizzthabet (approved):
lgtm! i tested with storybook tiltfile + resource, since it has such noisy output. i do see some jiggle when navigating back to a resource with a ton of logs (which i also see in the main branch, so it's not a regression). viewing a resource with a ton of active new log output does look less jiggly.

nicks:
hmmm -- in your screenshot, it says you're testing v0.25.0, but this PR is on top of v0.25.1. so i don't think you're testing the right branch?

lizzthabet:
@nicks, i should be on the right branch! just to clarify -- are there any backend changes that would affect the log display? (i did compile tilt while testing the build logs pr, though i tested this first and hadn't compiled tilt since the last release.)

nicks:
hmmm i don't know why you're seeing the old behavior then

PR #5538, opened 2022-02-24
tiltextension: fix a regression where we stopped pulling the default repo

nicks:
Hello @landism, @milas,

Please review the following commits I made in branch nicks/extensions:

308a574504cb4023f98fd347f06a2f493e4d3127 (2022-02-24 11:28:17 -0500)
tiltextension: if a regression where we stopped pulling the default repo
also some minor usability fixes

### DIFF
--- a/internal/controllers/core/extensionrepo/reconciler.go
+++ b/internal/controllers/core/extensionrepo/reconciler.go
@@ -230,8 +230,11 @@ func (r *Reconciler) reconcileDownloaderRepo(ctx context.Context, state *repoSta
 	state.lastFetch = time.Now()
 
 	needsDownload := true
-	if exists && state.spec.Ref != "" {
-		// If an explicit ref is specified, there's no reason to pull a new version.
+	if exists && state.spec.Ref != "" && state.spec.Ref != "HEAD" {
+		// If an explicit ref is specified, we assume there's no reason to pull a new version.
+		//
+		// TODO(nick): Should we try to support cases where the ref can change server-side?
+		// e.g., a "stable" tag.
 		err := r.dlr.RefSync(importPath, state.spec.Ref)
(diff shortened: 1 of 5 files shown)


internal/controllers/core/extensionrepo/reconciler.go
```diff
@@ -230,8 +230,11 @@ func (r *Reconciler) reconcileDownloaderRepo(ctx context.Context, state *repoSta
 	state.lastFetch = time.Now()
 
 	needsDownload := true
-	if exists && state.spec.Ref != "" {
-		// If an explicit ref is specified, there's no reason to pull a new version.
+	if exists && state.spec.Ref != "" && state.spec.Ref != "HEAD" {
```
nicks, line 233: i'm actually not totally sure what the right API is here for tags that can change server-side...but this seems like a good stopgap until we figure out what we want.

milas: Agreed this is probably a reasonable stop-gap, and custom extension repos can still take advantage of rolling updates by ensuring their default branch is `stable` so that it'd be `HEAD`, right?

Longer-term, we could probably inspect the _local_ ref to determine the type, and if it doesn't exist yet or is a branch, we could fetch updates but if it exists & is a tag/commit SHA, skip the sync.

milas (approved):

internal/controllers/core/extensionrepo/reconciler.go
nicks: fwiw, i think the correct behavior here is to do something like:
- always fetch the remote origin
- resolve the `Ref` in the remote origin to a git hash
- checkout the git hash
but it also feels like a bit of a rabbithole that's not worth spending time optimizing

PR #5539, opened 2022-02-25
web: make build event logs sticky

nicks:
Hello @hyu, @lizzthabet,

Please review the following commits I made in branch nicks/ch13284:

fc1623a022e386281baf124e1d99283ba92ba0ab (2022-02-24 19:38:56 -0500)
web: make build event logs sticky

### DIFF
--- a/web/src/LogLine.scss
+++ b/web/src/LogLine.scss
@@ -25,6 +25,12 @@
     padding-right: $spacing-unit;
     border-top: 1px solid rgba($color-gray-light, $translucent);
     border-bottom: 1px solid $color-blue-dark;
+    
+    &.is-sticky {
+      position: sticky;
+      top: 0;
+      z-index: 1;
+    }
   }
   &.is-buildEvent-fallback {
     background-color: $color-gray-darker;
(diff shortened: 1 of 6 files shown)


web/src/OverviewLogPane.tsx
```diff
@@ -85,6 +85,14 @@ function newLineEl(
   if (buildEvent === "init") {
     classes.push("is-buildEvent")
     classes.push("is-buildEvent-init")
+
+    if (showManifestPrefix) {
+      text += ` • ${line.manifestName}`
```
lizzthabet, line 90: (nit) it'd be nice to have a short comment here that explains the reason for suffixing with the manifest name.

web/src/LogLine.scss
```diff
@@ -40,10 +46,17 @@
...
+.LogLine:not(.is-buildEvent) + .LogLine.is-buildEvent {
   margin-top: 8px;
 }
+.LogLine.is-buildEvent + .LogLine:not(.is-buildEvent) > .LogLine-content,
+.LogLine.is-buildEvent + .LogLine:not(.is-buildEvent) > .logLinePrefix {
+  padding-top: 8px;
+}
+.LogLine:not(.is-buildEvent) + .LogLine.is-buildEvent.is-sticky,
+.LogLine.is-buildEvent.is-sticky:first-child {
+  top: -8px; // For position:sticky with padding on LogPane
```
lizzthabet, line 58: definitely not a merge-blocker, but i still see the 8px of padding when i scroll all the way to the top of the log pane. 



fiddling around in the browser, i was able to remove it by setting the log pane padding to `0 0 8px 0` and then this can be set to `0`.

lizzthabet (approved):
🎉

web/src/LogLine.scss
nicks: oh that's intentional! that's what it looks like at head (go here https://storybook.tilt.dev/?path=/story/new-ui-overviewresourcepane--ten-resources-with-log-store and click "all resources")

lizzthabet: ahh, interesting! nevermind then.

web/src/OverviewLogPane.tsx
nicks: done

nicks (commented):
i fiddled with the top padding in a way that i hope will make this look a bit nicer!

PR #5547, opened 2022-02-28
apis: move DockerComposeUpSpec to the API model

nicks:
Hello @milas, @nicksieger,

743100c322be171681bf340c8819ee44b4b73d53 (2022-02-28 16:04:31 -0500)
apis: move DockerComposeUpSpec to the API model


nicks:
one high-level question: The Docker Compose team has been pushing "Compose" as the general API, and "Docker Compose" as a particular implementation.

https://github.com/compose-spec/compose-spec

e.g., you can also do `nerdctl compose`

what do you all think of how this is reflected in our API?

i kind of go back and forth on this -- on the one hand, we could probably implement a "ComposeService" with different underlying orchestrators. But I think users might be more confused by this.

nicksieger (approved):
I'm inclined to leave the naming as-is, since we only support (and require) docker-compose in order to work. Agree that talking about "compose" without Docker could confuse. Since the api namespace has "alpha" in it, a future change to drop the "docker" part of the name seems like it would still be acceptable. LGTM :shipit:

PR #5548, opened 2022-02-28
args: clarify long description and add examples

lizzthabet:
There are the changes I originally attempted to make [here](https://github.com/tilt-dev/tilt.build/pull/1056/files#diff-9196fb42514c5c98aa44fc9372700ed5b542218e7d229881830618ad63e99176) before realizing the source code lived here! 

The formatting for headings is a little off on the [`tilt args` docs](https://docs.tilt.dev/cli/tilt_args.html), so this PR attempts to fix the heading hierarchy and moves the code examples into a separate `Examples` section with comments.

### DIFF
--- a/internal/cli/args.go
+++ b/internal/cli/args.go
@@ -42,25 +42,25 @@ func (c *argsCmd) register() *cobra.Command {
 		DisableFlagsInUseLine: true,
 		Short:                 "Changes the Tiltfile args in use by a running Tilt",
 		Long: `Changes the Tiltfile args in use by a running Tilt.
-If no args are specified, (i.e., just "tilt args"), opens the current args for editing in
+
+The "args" can be used to specify what resources Tilt enables. They can also 
+be used to define custom key-value pairs that can be accessed in a Tiltfile
+by calling "config.parse()".
+
+If no args are explicitly specified, (i.e., just "tilt args"), it opens the current args for editing in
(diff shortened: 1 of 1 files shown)


internal/cli/args.go
```diff
@@ -42,25 +42,25 @@ func (c *argsCmd) register() *cobra.Command {
...
+The "args" can be used to specify what resources Tilt enables. They can also 
+be used to define custom key-value pairs that can be accessed in a Tiltfile
+by calling "config.parse()".
+
+If no args are explicitly specified, (i.e., just "tilt args"), it opens the current args for editing in
 the editor defined by your TILT_EDITOR or EDITOR environment variables, or fall back to
 an OS-appropriate default.
 
-The editor can be skipped by providing new args on the command line, e.g.: "tilt args svc1 svc2".
+Note that Tiltfile arguments do not affect built-in Tilt args (e.g. "--legacy", "--port"), and they
```
landism, line 54:
```suggestion
Note that Tiltfile arguments do not affect built-in Tilt args (i.e., the things that show up in "tilt up --help", such as "--legacy", "--port"), and they
```

landism: just since it's feasible most users have no idea --legacy or --port even exist

landism (approved):
huzzah!

PR #5553, opened 2022-03-02
web: scroll when a sidebar item is selected

nicks:
Hello @lizzthabet,

Please review the following commits I made in branch nicks/ch13353:

920c9a54406d503c5657b54cefc865aa5260cea3 (2022-03-02 10:27:19 -0500)
web: scroll when a sidebar item is selected

### DIFF
--- a/web/src/SidebarItemView.tsx
+++ b/web/src/SidebarItemView.tsx
@@ -1,4 +1,4 @@
-import React from "react"
+import React, { useEffect, useRef } from "react"
 import TimeAgo from "react-timeago"
 import styled from "styled-components"
 import { Flag, useFeatures } from "./feature"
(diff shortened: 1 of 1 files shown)


web/src/SidebarItemView.tsx
```diff
@@ -336,9 +336,17 @@ export function EnabledSidebarItemView(props: SidebarItemViewProps) {
   let onStartBuild = startBuild.bind(null, item.name)
   const groupViewIndentClass = props.groupView ? "groupViewIndent" : ""
   let analyticsTags = { target: item.targetType }
+  let ref: any = useRef(null)
```
lizzthabet, line 339: could this be typed as `MutableRefObject<HTMLLIElement | null>`?

lizzthabet (approved):
one type suggestion, but otherwise lgtm!

web/src/SidebarItemView.tsx
nicks: oooh nice suggestion! done

PR #5557, opened 2022-03-03
k8s: improve error messages during Pod exec failures

milas:
When running a command in a Pod, `client-go` attempts to handle
the failure case of a non-zero exit code by inspecting a
`metav1.Status` object.

For other failure cases, it returns `fmt.Errorf(status.Message)`,
and `Message` might be blank, resulting in cryptic error messages
from Tilt.

There's not a lot we can do without upstream `client-go` changes
on the Tilt side, so some extra context is added around where the
exec failed, and for the stream error case, a generic fallback
error message is used if we get an empty error from `client-go`.

See `remotecommand/v4.go` for the problematic `client-go` code:
https://github.com/kubernetes/client-go/blob/8e46da3fd12b54a2d1726fc9a9438ff8fc38494c/tools/remotecommand/v4.go#L91-L118

### DIFF
--- a/internal/k8s/exec.go
+++ b/internal/k8s/exec.go
@@ -30,7 +30,7 @@ func (k *K8sClient) Exec(_ context.Context, podID PodID, cName container.Name, n
 
 	exec, err := remotecommand.NewSPDYExecutor(k.restConfig, "POST", req.URL())
 	if err != nil {
-		return fmt.Errorf("establishing connection: %v", err)
+		return fmt.Errorf("establishing connection: %w", err)
 	}
 
 	err = exec.Stream(remotecommand.StreamOptions{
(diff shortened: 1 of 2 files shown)


internal/k8s/exec.go
```diff
@@ -28,12 +30,25 @@ func (k *K8sClient) Exec(ctx context.Context, podID PodID, cName container.Name,
 
 	exec, err := remotecommand.NewSPDYExecutor(k.restConfig, "POST", req.URL())
 	if err != nil {
-		return err
+		return fmt.Errorf("failed to establish connection: %v", err)
```
nicks, line 33: fwiw, i would probably shorted this to "establishing connection: %v", on the assumption that callers may wrap this in their  own error message (and so we don't need to say "failure" each time)

nicks (approved):

nicksieger (approved):

PR #5560, opened 2022-03-04
Parallel uploads

pnovotnak:
The current behavior is to upload files serially, to each container where an image is in use. This can make developing services on remote clusters slow when the same container is used multiple times.

This is a quick & dirty change that enables n-parallelism for uploads where n is the number of containers using the given build. Output becomes a little muddied:

```
Will copy 1 file(s) to container: [my-big-fat-pod/container-1]
- '/Users/pj/Code/my-app/server.py' --> '/src/server.py'
Will copy 1 file(s) to container: [my-big-fat-pod/container-2]
- '/Users/pj/Code/my-app/server.py' --> '/src/server.py'
Will copy 1 file(s) to container: [my-big-fat-pod/container-3]
- '/Users/pj/Code/my-app/server.py' --> '/src/server.py'
Will copy 1 file(s) to container: [my-big-fat-pod/container-3]
- '/Users/pj/Code/my-app/server.py' --> '/src/server.py'
Will copy 1 file(s) to container: [my-big-fat-pod/container-4]
- '/Users/pj/Code/my-app/server.py' --> '/src/server.py'
  → Container my-big-fat-pod/container-3 updated!
  → Container my-big-fat-pod/container-4 updated!
  → Container my-big-fat-pod/container-1 updated!
  → Container my-big-fat-pod/container-3 updated!
```

I'm willing to invest some time in cleaning that up as well if it's felt that's required for a merge.

I've also bumped the .nvmrc since the version listed there doesn't seem to be usable by the js toolchain. I bumped it to the newest version that the toolchain claims support for.


internal/controllers/core/liveupdate/reconciler.go
```diff
@@ -925,63 +925,87 @@ func (r *Reconciler) applyInternal(
...
-		err = cu.UpdateContainer(ctx, cInfo, archive,
-			build.PathMappingsToContainerPaths(toRemove), boiledSteps, hotReload)
-		_ = archive.Close()
-
-		lastFileTimeSynced := input.LastFileTimeSynced
-		if lastFileTimeSynced.IsZero() {
-			lastFileTimeSynced = apis.NowMicro()
-		}
+		go func(cInfo liveupdates.Container) {
+			wg.Add(1)
```
nicks, line 933: this needs to be outside the goroutine (i think CI has a lint check for this mistake)

though i would probably switch to using an errgroup, as in this example: 
https://pkg.go.dev/golang.org/x/sync/errgroup#example-Group-Parallel
which is a bit easier to reason about

internal/controllers/core/liveupdate/reconciler.go
```diff
@@ -925,63 +925,87 @@ func (r *Reconciler) applyInternal(
...
+				logger.Get(copierContext).Infof("  → Container %s updated!", cInfo.DisplayName())
 			}
-		} else {
-			logger.Get(ctx).Infof("  → Container %s updated!", cInfo.DisplayName())
-			if lastExecErrorStatus != nil {
-				// This build succeeded, but previously at least one failed due to user error.
-				// We may have inconsistent state--bail, and fall back to full build.
-				result.Failed = &v1alpha1.LiveUpdateStateFailed{
-					Reason: "PodsInconsistent",
-					Message: fmt.Sprintf("Pods in inconsistent state. Success: pod %s. Failure: pod %s. Error: %v",
```
nicks, line 975: i think you lost this bit

nicks (commented):
i like this change!

internal/controllers/core/liveupdate/reconciler.go
pnovotnak: So, I'm not sure I fully understand the purpose of this or how to implement in the parallelized version. In the serialized version it would be invoked like:

```
update -> update failure
update -> abort
```

In the new version that logic can't work, since we are doing everything in parallel. However, we do ofc catch errors in the new `select` statement, so we could update the `Reason` field there?...

internal/controllers/core/liveupdate/reconciler.go
pnovotnak: Cool, I didn't know about this! This use-case doesn't seem to fit perfectly, but it does clean things up a bit.

internal/controllers/core/liveupdate/reconciler.go
nicks: sure!

think about it like this:

- if all `run('yarn install')`s succeed, that's OK
- if all `run('yarn install`)`s fail, that's also OK - it just means you introduced a syntax error
- if some of them succeeded and some of them failed, that's not something we can easily recover from.

an important part of this is there's a distinction between `run("yarn install")` reported some errors in the code being installed, vs `run("yarn install")` failed because of some underlying kubernetes failure.

internal/controllers/core/liveupdate/reconciler.go
```diff
@@ -925,63 +936,77 @@ func (r *Reconciler) applyInternal(
...
+		select {
+		case res := <-results:
+			if res.Failed != nil {
+				cancel()
+				result.Failed = res.Failed
+			}
+			result.Containers = append(result.Containers, res.Containers...)
+		case <-gErrorC(g):
+		case <-ctx.Done():
+			cancel()
```
nicks, line 1006: this is going to create a memory leak i think ... the channel send will block forever if the outer function returns early without returning all values.

you'll need to drain the channel here.

though honestly, i think this would be a lot simpler if you used a slice to accumulate the results, and didn't use a channel at all (as the errgroup example code does)

nicks:
Hi @pnovotnak ! Is there anything that you're waiting on for us in this PR? I appreciate the problem you're trying to solve here, though it seems like you're having some issues getting all the test failures / bug fixes out.

We can also re-file as a feature request if you've run out of time to work on this.

nicks:
going to close this, since it's stale and we could never get the tests to pass. we can re-open if anyone wants to pick it up.

PR #5568, opened 2022-03-07
liveupdate: provide more user-friendly errors

milas:
Follow up to #5527.

The big addition here is to handle the case where `tar` is not
found in `$PATH`, which is very common when users are using a
`scratch`-built image or very minimal base image like the `distroless`
family.

How these manifest between the Docker updater (used for Compose +
K8s exposed in-cluster Docker socket) and exec updater (used for all
other K8s configs, generally meaning containerd) are slightly
different.

### Permission Denied
dockerd_permissions

### No `tar` in PATH
dockerd_notfound

### Permission Denied
containerd_permissions

### No `tar` in PATH
containerd_notfound


internal/containerupdate/exec_updater.go
```diff
@@ -50,9 +52,18 @@ func (cu *ExecUpdater) UpdateContainer(ctx context.Context, cInfo liveupdates.Co
...
 	err := cu.kCli.Exec(ctx, cInfo.PodID, cInfo.ContainerName, cInfo.Namespace,
-		tarArgv(), archiveToCopy, tarWriter, tarWriter)
+		tarCmd.Argv, archiveToCopy, tarWriter, tarWriter)
 	if err != nil {
+		if exitCodeErr, ok := build.WrapCodeExitError(err, tarCmd); ok {
+			switch exitCodeErr.ExitCode {
+			case TarExitCodePermissionDenied:
+				return permissionDeniedErr(err)
+			case GenericExitCodeCannotExec:
+				return cannotExecErr(err)
```
nicksieger, line 64: Is it worth extracting this switch to a new function? Or moving to build.WrapCodeExitError()?

nicksieger (approved):

milas:
@nicksieger I pushed a (shockingly chonky) commit (5a5afebd6) with some clean up

* Split concept of container exec errors from `RunStepFailure`
  * `RunStepFailure` is a generic wrapper error based on _where_ the failure occurred so that LU reconciler can react differently
  * This allows consistent/uniform handling of exec errors during things like the `tar` step to sync files and the actual `run()` steps execution
* Consolidated some error-handling logic in `containerupdate`
  * There's still a lot of duplication here because the K8s exec API + `client-go` doesn't give us as much as Docker, so we still have to peek at error messages in some cases
* Added a new test case for "not found in PATH" as well
* Swapped some logic to new stdlib error unwrapping and fix a couple issues here
* Remove redundant parts of errors (in some cases you could see the container ID or command multiple times)
* Fix mismatched capitalization in (wrapped) error messages

internal/containerupdate/exec_updater.go
milas: (see top-level comment - I tried to refactor for this + some other improvements)

nicksieger (approved):
Sweet! Shows how error handling can be pretty verbose at times but it definitely makes the error handling logic clearer.

PR #5574, opened 2022-03-08
tiltfile+cli: embed api stub files and add `tilt dump api-stubs`

nicksieger:
- tiltfile: import and go:embed api stubs
- cli: add `tilt dump api-stubs` command


internal/tiltfile/api.go
```diff
@@ -0,0 +1,41 @@
+package tiltfile
+
+import (
+	"embed"
+	"io/fs"
+	"os"
+	"path/filepath"
+)
+
+//go:embed api/*.py api/config/*.py api/os/*.py api/shlex/*.py api/sys/*.py api/v1alpha1/*.py
```
landism, line 10:
```suggestion
//go:embed api/*.py api/*/*.py
```

landism: if only to make it less likely that someone adds a module and is confused when / doesn't notice that there are no docs

landism (approved):

PR #5583, opened 2022-03-10
web: make navigation and table options sticky on Table View, plus minor style fixes

lizzthabet:
This PR moves the resource name filter, bulk actions, starred bar, and display options into the `OverviewTablePane` component and wraps them in a `sticky` container. They'll be visible as a user scrolls through a long list of resources.

I also included a couple minor adjustments so:
* disabled resource status links are clickable
* tables have a minimum width, since the fixed table layout can make content spill outside of the table container on medium and small screens

[Closes ch12976](https://app.shortcut.com/windmill/story/12976/nice-to-have-make-header-and-options-bar-sticky-to-top-on-table-view)


nicks (approved):
LGTM!

PR #5590, opened 2022-03-14
ci: fix cross-compilation toolchain + bump to Go 1.17.8

milas:
This pulls in changes from tilt-dev/golang-cross#2 and makes the
necessary changes on the Tilt side:
 * Use the updated `golang-cross` base image
   * Minor cleanup to `tilt-releaser` image to reduce layer sizes
 * Split `.goreleaser.yml` into more "builds" by OS/arch and pass
   the proper `CC` / `CXX` flags now that we need cgo for every
   target
 * Re-build generic Tilt CI Linux image to pick up Go 1.17.8 to
   match the releaser
   * Minor cleanup to reduce layer sizes + fix build issue due to
     image having cached IP for apt source
 * Update image references for CircleCI jobs
 * Update Go versions for macOS/Windows CircleCI jobs
   * Linux Go version determined by generic Tilt CI image above
 * Add new `release-verify` job that runs a `goreleaser build` +
   `yarn build` but doesn't publish anything
   * Does not run by default! A pipeline can be manually kicked
     off for a branch with the `verify_release` parameter set to
     `true`, which will only run the verify workflow and not the
     standard build/test jobs
   * Can be used to safely verify build changes pre-release on
     feature branches (e.g. after upgrading deps in `go.mod` or
     messing with Webpack configs)

### DIFF
--- a/.circleci/Dockerfile
+++ b/.circleci/Dockerfile
@@ -6,13 +6,16 @@ RUN go get -u github.com/google/wire/cmd/wire
 
 RUN go get -u golang.org/x/tools/cmd/goimports
 
-RUN sudo apt-get install default-jdk
-
 USER root
 
-RUN curl -sL https://deb.nodesource.com/setup_16.x | bash -
-
-RUN apt-get install -y nodejs
+RUN curl -sL https://dl.yarnpkg.com/debian/pubkey.gpg | apt-key add - && \
+    echo "deb https://dl.yarnpkg.com/debian/ stable main" | tee /etc/apt/sources.list.d/yarn.list && \
+    curl -sL https://deb.nodesource.com/setup_16.x | bash - && \
+    apt install -y -q --no-install-recommends \
+      default-jdk \
+      nodejs \
+      yarn \
+    && rm -rf /var/lib/apt/lists/*
 
 RUN mv /usr/bin/docker-compose /usr/bin/docker-compose-v1
 ARG DOCKER_COMPOSE_V2_VERSION=v2.2.2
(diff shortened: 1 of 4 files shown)


.circleci/Dockerfile
```diff
@@ -6,13 +6,16 @@ RUN go get -u github.com/google/wire/cmd/wire
 
 RUN go get -u golang.org/x/tools/cmd/goimports
 
-RUN sudo apt-get install default-jdk
```
milas, line 9: Since there was no `apt-get update` before, this was now failing; I went ahead and combined it with the NodeJS + Yarn installs with cleanup in the process of fixing

.goreleaser.yml
```diff
@@ -11,30 +11,61 @@ builds:
...
   - arm64
   env:
-  # Compiles against the MacOS FSEvents system libraries.
+  # clang is a cross-compiler by default, so the same CC/CXX works
+  # for arm64 (M1) & amd64 (Intel) builds
   - CGO_ENABLED=1
-  - CC=/osxcross/target/bin/x86_64h-apple-darwin20.2-clang
-  - CXX=/osxcross/target/bin/x86_64h-apple-darwin20.2-clang++
+  - CC=o64h-clang
+  - CXX=o64h-clang++
```
milas, line 18: The version changed to `darwin20.4` but these are generic symlinks (and exist in `PATH` so don't need to be qualified), which should make future base image updates less painful

milas:
New (manual) `ci/circleci: release-verify` job/workflow: https://app.circleci.com/pipelines/github/tilt-dev/tilt/11239/workflows/e4e5f8e0-0151-4775-8e5a-8cb8633824b9/jobs/86641

(Also listed above in the status checks)

.circleci/config.yml
```diff
@@ -166,15 +176,22 @@ jobs:
...
 
+parameters:
+  verify_release:
+    default: false
+    type: boolean
+
 workflows:
   version: 2
   shellcheck:
+    unless: << pipeline.parameters.verify_release >>
```
nicksieger, line 187: Is this valid yaml to have those angle brackets unescaped?

nicks (approved):

nicksieger (approved):

.circleci/config.yml
milas: Yep! (I think `>` will cause problems at string start, but `<` is fine)

https://support.circleci.com/hc/en-us/articles/360043638052-Conditional-steps-in-jobs-and-conditional-workflows

PR #5594, opened 2022-03-15
goreleaser: add -static to LDFLAGS for windows build

nicksieger:


milas (approved):

.goreleaser.yml
```diff
@@ -64,6 +64,7 @@ builds:
   - amd64
   env:
     - CGO_ENABLED=1
+    - CGO_LDFLAGS=-static
```
nicks, line 67: maybe add it to every build to be consistent? i'd hate if we had a bug that only affected static builds. i guess in theory macos should always have glibc from xcode

milas: That said, might be worth adding `CGO_LDFLAGS=-static` to the CircleCI `build-windows` job:
https://github.com/tilt-dev/tilt/blob/b1ac8297050e588939b7ee5793ddb24a966293f2/.circleci/config.yml#L73

nicks: ah ok!

PR #5606, opened 2022-03-21
ci: upgrade to Go 1.18

milas:
See also:
* tilt-dev/golang-cross#3 (`goreleaser` cross-compile base image aka `gcr.io/windmill-public-containers/golang-cross:1.18.0-1`)
* tilt-dev/tilt-extensions#366 (extensions CI image, used here to re-build Tilt from `HEAD` instead of using latest release)

I cut out some unnecessary stuff from the image and made some changes to shrink the size, which should help speed up CI a bit since there'll be less fetching of massive images.


.circleci/Dockerfile.integration
```diff
@@ -21,31 +36,21 @@ RUN curl -fL "https://github.com/docker/compose/releases/download/${DOCKER_COMPO
...
+RUN go install gotest.tools/gotestsum@latest \
+    && gotestsum --version
 
 # install Kind (Kubernetes in Docker)
 ENV KIND_VERSION=v0.12.0
 RUN curl -fLo ./kind-linux-amd64 https://github.com/kubernetes-sigs/kind/releases/download/${KIND_VERSION}/kind-linux-amd64 \
   && chmod +x ./kind-linux-amd64 \
-  && mv ./kind-linux-amd64 /go/bin/kind
-
-RUN apt update && apt install -y curl ca-certificates liblz4-tool rsync socat
```
nicks, line 48: i think you still need socat and friends for the remote docker stuff

nicks (approved):

.circleci/Dockerfile.integration
milas: I moved them above to be part of one mega `apt install`!

PR #5622, opened 2022-03-23
assets: embedded asset server

nicksieger:
During release builds, copy js assets to pkg/assets/build for embedding in the Tilt executable using go:embed. When in `--web-mode=prod`, prefer embedded asset server to prod asset server if embedded assets are available, identified by index.html being present.

Also, `PrecompiledWebMode` is replaced by this embedded server as well, since both run from the result of `make build-js`.


internal/cli/up.go
```diff
@@ -245,20 +245,21 @@ func provideWebURL(webHost model.WebHost, webPort model.WebPort) (model.WebURL,
 }
 
 func provideAssetServer(mode model.WebMode, version model.WebVersion) (assets.Server, error) {
+	s, err := assets.NewEmbeddedServer()
```
nicks, line 248: this error is getting dropped on the floor

pkg/assets/embedded.go
```diff
@@ -0,0 +1,59 @@
...
+	<-ctx.Done()
+	return nil
+}
+
+func (s embeddedServer) TearDown(ctx context.Context) {
+}
+
+func handleRoutingURLs(handler http.Handler, index []byte) http.HandlerFunc {
+	return func(w http.ResponseWriter, req *http.Request) {
+		if strings.HasPrefix(req.URL.Path, "/r/") {
```
nicks, line 49: hmm...

the prod and old precompiled servers redirect everything to index.html, and whitelist the set of urls that are supposed to be served from the filesystem. 

your new server does the opposite - it serves everything from the embed FS, and whitelists the set of urls that should redirect to index.html. 

is there a reason you decided to flip the decision logic? seems error prone to have two very different decision sets of decision logic and likely to lead to problems.

pkg/assets/embedded.go
```diff
@@ -0,0 +1,59 @@
...
+)
+
+//go:embed build
+var build embed.FS
+
+type embeddedServer struct {
+	http.Handler
+}
+
+func NewEmbeddedServer() (Server, error) {
```
nicks, line 22: i think the more idiomatic API is something like:

GetEmbeddedServer() (Server, bool)

since "the server doesn't exist" isn't something that you're treating as an error. here's an example of an API that does this: https://pkg.go.dev/os#LookupEnv

internal/cli/up.go
```diff
@@ -245,20 +245,21 @@ func provideWebURL(webHost model.WebHost, webPort model.WebPort) (model.WebURL,
...
-		if err != nil {
-			return nil, err
-		}
-		pkgDir := assets.PackageDir(path)
-		if mode == model.PrecompiledWebMode {
-			return assets.NewPrecompiledServer(pkgDir), nil
-		} else {
-			return assets.NewDevServer(pkgDir, model.WebDevPort(webDevPort))
-		}
+	if mode == model.PrecompiledWebMode && err == nil {
```
nicks, line 255: if you ask for precompiled mode, and there are no precompiled assets, seems like that should be an error

nicks (commented):
yessssssssss!!!!

internal/cli/up.go
nicksieger: I think based on your comment below that this would work better as `s, ok := ...` instead

pkg/assets/embedded.go
nicksieger: That wasn't intentional. I can change this to wrap the response from the file server and on 404 serve up index.html. That's basically what you're saying right?

PR #5623, opened 2022-03-23
cluster: monitor K8s livez/readyz after initialization

milas:
After initializing a connection, start a monitor goroutine to
poll the `/livez` and `/readyz` endpoints of the cluster and
update the Tilt `Cluster` object status accordingly.

Additionally, if the client cannot be initialized in the first
place, the reconciler will requeue itself in the future and
attempt re-initialization.


nicksieger (approved):

internal/controllers/core/tiltfile/api.go
```diff
@@ -557,11 +557,19 @@ func toClusterObjects(nn types.NamespacedName, tlr *tiltfile.TiltfileLoadResult,
 		return result
 	}
 
+	var annotations map[string]string
+	if tlr.FeatureFlags[feature.ClusterRefresh] {
+		annotations = map[string]string{
+			"features.tilt.dev/cluster-refresh": "true",
+		}
+	}
```
milas, line 565: This felt like a reasonably clean way to propagate feature flag information to an individual resource, and there's sorta precedent here because we conditionally create disable buttons in a sibling function based on feature flags

milas:
Reason I'd been holding off on merging this is because of the issues it exposes that I mentioned during demo in company meeting.

Requested a new review (cc @nicksieger) because I added a feature flag to gate the problematic part of the functionality (existing logic remains as-is), so we can continue to address the underlying issues safely

nicksieger (approved):
Seems like a reasonable use of annotations to me.

PR #5632, opened 2022-03-24
cluster: populate local registry information

lizzthabet:
This PR populates the local registry info in the cluster API object. (Once #5631 is merged in, I'll point this PR to the main branch.)

### DIFF
--- a/internal/controllers/core/cluster/reconciler.go
+++ b/internal/controllers/core/cluster/reconciler.go
@@ -113,6 +113,11 @@ func (r *Reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
 		}
 	}
 
+	if conn.error == "" && conn.connType == connectionTypeK8s && conn.registry == nil {
+		reg := conn.k8sClient.LocalRegistry(ctx)
+		conn.registry = &reg
+	}
+
 	r.connManager.store(nn, conn)
 
 	status := conn.toStatus()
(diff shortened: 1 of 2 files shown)


internal/controllers/core/cluster/reconciler.go
```diff
@@ -239,9 +244,21 @@ func (c *connection) toStatus() v1alpha1.ClusterStatus {
 		connectedAt = &t
 	}
 
+	var reg *v1alpha1.RegistryHosting
+	if c.registry != nil {
+		reg = &v1alpha1.RegistryHosting{
+			Host:                   c.registry.Host,
+			HostFromClusterNetwork: c.registry.HostFromCluster(),
```
nicks, line 251: ah, you reversed it! In the tilt internal data structure. "HostFromCluster" means the container runtime. see here:
https://github.com/tilt-dev/tilt/blob/master/internal/k8s/registry.go#L161

nicks: (there was originally a single "host from cluster", but after a long conversation with the k8s sig-cluster-lifecycle team about it, we realized that there were two separate networks in the cluster - the network of the container runtime pulling the image, and the network inside the containers)

lizzthabet: ah, ok! updated.

PR #5634, opened 2022-03-28
cluster: add a connection status field

nicks:
Hello @milas,

Please review the following commits I made in branch nicks/productid2:

d8103242ab7999ea4007b375763dda02e57189ec (2022-03-28 14:26:17 -0400)
cluster: add a connection status field


pkg/apis/core/v1alpha1/cluster_types.go
```diff
@@ -231,3 +236,24 @@ var _ resource.StatusSubResource = &ClusterStatus{}
...
 }
+
+// Connection spec for an existing cluster.
+type ClusterConnectionStatus struct {
+	// Defines connection to a Kubernetes cluster.
+	Kubernetes *KubernetesClusterConnectionStatus `json:"kubernetes,omitempty" protobuf:"bytes,1,opt,name=kubernetes"`
+}
+
+// Kubernetes-specific fields for connection status
+type KubernetesClusterConnectionStatus struct {
```
nicks, line 247: i struggled a lot with what the name of this should be...ContextStatus? something else?

i expect we'll want to do a similar thing with Docker (where we publish the resolved docker host)

milas: I think this makes sense!

internal/k8s/exploding_client.go
```diff
@@ -102,3 +104,11 @@ func (ec *explodingClient) OwnerFetcher() OwnerFetcher {
 func (ec *explodingClient) ClusterHealth(_ context.Context, _ bool) (ClusterHealth, error) {
 	return ClusterHealth{}, errors.Wrap(ec.err, "could not set up kubernetes client")
 }
+
+func (ec *explodingClient) ConnectionStatus() *v1alpha1.KubernetesClusterConnectionStatus {
+	return &v1alpha1.KubernetesClusterConnectionStatus{
+		Context:   "default",
+		Namespace: "default",
+		Product:   string(clusterid.ProductUnknown),
+	}
```
milas, line 113: I wonder if this should be `nil` or empty object?

internal/k8s/client.go
```diff
@@ -135,6 +136,8 @@ type Client interface {
 	OwnerFetcher() OwnerFetcher
 
 	ClusterHealth(ctx context.Context, verbose bool) (ClusterHealth, error)
+
+	ConnectionStatus() *v1alpha1.KubernetesClusterConnectionStatus
```
milas, line 140: Ok so in the context of the API `KubernetesClusterConnectionStatus` makes sense BUT here I wonder if the method should be something like `ConnectionConfig()`?

milas (approved):
Thanks for adding these!

internal/k8s/exploding_client.go
nicks: i made context and namespace empty

internal/k8s/client.go
nicks: good idea! done

nicks:
i actually needed this for something unrelated! :innocent:

PR #5637, opened 2022-03-29
web: simplify how the global nav manages dialog display state

lizzthabet:
This PR refactors how the `GlobalNav` manages the display state for menu item dialogs, and it'll simplify the code I'll write to add a cluster connection menu item.

* It now uses a single `openDialog` state, instead of tracking separate (mutually exclusive) state for each dialog. 
* It also removes references to analytics `close` events, since there wasn't a code path that would log those events. (I think it's sufficient for our analytics to continue to log `click` events on the menu buttons, and separate `close` events don't add additional information, but chime in if I'm wrong!)

### DIFF
--- a/web/src/GlobalNav.tsx
+++ b/web/src/GlobalNav.tsx
@@ -154,52 +154,38 @@ type GlobalNavProps = {
   runningBuild: TiltBuild | undefined
 }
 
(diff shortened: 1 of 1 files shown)


milas (approved):
✨

Issue #3442, opened 2020-06-11
local_resource should have a way to expose stding


nicks:

emschwartz:
Just ran into this issue trying to get Tilt to run the webpack dev server. In case anyone else searches for the error, it prints `Server exited with exit code 0`. Adding `ail -f /dev/null | yarn start` resolved the issue.

lizzthabet:
@nicks, does #5641 solve this issue? i'm able to run the tilt webpack dev server without any problems.

nicks:
@lizzthabet nope! that's just for local() in the tiltfile. 

i wouldn't be surprised if webpack got fixed to handle this mode, but there are still lots of tools that don't handle stdin correctly.

PR #5641, opened 2022-03-29
tiltfile: local: add stdin param

landism:
This is mostly to make it easier to get a starlark variable into a command (e.g., `kubectl apply -f -` or `cat > foo.yaml`).

This is currently possible by escaping the var into the command line and piping, but 1) escaping feels finicky and 2) that risks hitting ARG_MAX.


internal/tiltfile/api/__init__.py
```diff
@@ -675,6 +676,7 @@ def local(command: Union[str, List[str]],
     echo_off: If set to True, skips printing command to log.
     env: Environment variables to pass to the executed ``command``. Values specified here will override any variables passed to the Tilt parent process.
     dir: Working directory for ``command``. Defaults to the Tiltfile's location.
+    stdin: If non-empty, will be written to ``command``'s stdin.
```
landism, line 679: Or maybe #3442 will want an empty stdin so this should be "if non-None"

nicks: ya, we should handle None and empty string differently? i think there are some subtle differences in how the underlying process APIs handle it but I'm not 100% sure.

milas (approved):
Agree it's probably advisable to distinguish `None` from `""` (or provide some sort of special built-in value like `DUMMY_STDIN`), but otherwise 👍

landism:
redid it using `None` instead

and generics!

nicksieger (approved):
❤️

PR #5647, opened 2022-03-31
apis: add cluster name to the image spec

nicks:
Hello @milas,

Please review the following commits I made in branch nicks/cluster6:

bba51b1592b840c328f7c42035e42091ac017685 (2022-03-31 16:17:00 -0400)
apis: add cluster name to the image spec


nicks:
fwiw - i struggled a lot with how to model this, but i think this is probably good enough to start porting some of the build/push code into the reconciler.

PR #5653, opened 2022-04-01
tilt: bump releaser image to go 1.18

landism:


nicks:
i don't think this is right. it only updates the release-verify container, not the release contianer

landism:
thanks!

PR #5659, opened 2022-04-04
docker: support Colima for building direct to cluster

### LINKED ISSUE #5625: Tilt should detect Colima as a safe context and also skip Image push, if Colima is using Docker
Colima is `Container runtimes on macOS (and Linux) with minimal setup.` It could use docker or containerd as container engine and could also host kubernetes. Colima is based on Lima (the same LinuxOnMac engine, which is used by Rancher Desktop). So it would be nice for Tilt to detect Colima as a safe environment and also not push images in case of docker usage.

### Current Behavior
Currently, Tilt cannot detect that the user is running Colima, Tilt also does not know that Colima is safe. So the user needs to `allow_k8s_contexts` and moreover Tilt tries to push an image built with docker_build to docker.io

### Why Do You Want This?
Colima is widely used replacement for Docker Desktop on Mac, which provide almost drop-in replacement experience for it - so it would be nice if Tilt supports it out of the box.

**Additional context**
I'll try to make a PR for this feature...

milas:
Colima is a macOS/Linux local cluster tool built on top of Lima.
It creates lightweight VMs with Docker and (optionally) Kubernetes
with Docker being used as the Kubernetes container runtime and the
socket exposed for building images directly to its context.

Fixes #5625.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -46,7 +46,7 @@ require (
 	github.com/spf13/cobra v1.3.0
 	github.com/spf13/pflag v1.0.5
 	github.com/stretchr/testify v1.7.0
-	github.com/tilt-dev/clusterid v0.1.2
+	github.com/tilt-dev/clusterid v0.1.3
 	github.com/tilt-dev/dockerignore v0.1.1
 	github.com/tilt-dev/fsevents v0.0.0-20200515134857-2efe37af20de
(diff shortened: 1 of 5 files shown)


milas:
FYI logic changes are 759149780198a2a0eee763bdc96cec7ef4c22927 and I made some variable names less confusing/more consistent in 2f7c476c65fac02ce10bfd94e71c402d4cb8b521 🙃

PR #5666, opened 2022-04-06
api: add DefaultRegistry to Cluster spec

milas:
Default registry to use for a Cluster if no local registry is
available. If nil, no registry rewriting should occur.


pkg/apis/core/v1alpha1/cluster_types.go
```diff
@@ -79,12 +79,58 @@ type KubernetesClusterConnection struct {
...
 	Context string `json:"context,omitempty" protobuf:"bytes,1,opt,name=context"`
 
 	// The default namespace to use.
 	//
 	// If not specified, will use the namespace in the kubeconfig.
+	//
+	// +optional
 	Namespace string `json:"namespace,omitempty" protobuf:"bytes,2,opt,name=namespace"`
+
+	// DefaultRegistryOptions determines where images for this Cluster should
```
nicks, line 93: it might make more sense to put this on the Cluster.Spec? this doesn't really seem kubernetes-specific.

pkg/apis/core/v1alpha1/cluster_types.go
```diff
@@ -79,12 +79,58 @@ type KubernetesClusterConnection struct {
...
+	//
+	// If not specified, no registry rewriting will occur, and the images will
+	// be pushed/pulled to from the registry specified by the corresponding
+	// image build directive (e.g. `docker_build` or `custom_build`).
+	//
+	// +optional
+	DefaultRegistryOptions *DefaultRegistryOptions `json:"defaultRegistryOptions,omitempty" protobuf:"bytes,3,opt,name=defaultRegistryOptions"`
+}
+
+type DefaultRegistryOptions struct {
```
nicks, line 105: i think we probably want to use a single RegistryHosting type, and add SameName to it? i don't see any real data-modelling reason why they would be different, do you?

nicks (commented):
just some quick high-level qs on the API model

pkg/apis/core/v1alpha1/cluster_types.go
milas: I think safe to use `RegistryHosting` directly (w/ addition of `SingleName`), and that also opens up some future opportunities to configure a local registry **including the different host{fromClusterNetwork,fromContainerRuntime} values** so makes sense! Done in 2f4a0b5e5

pkg/apis/core/v1alpha1/cluster_types.go
milas: Moved it to be a top-level field of `ClusterSpec` in 2f4a0b5e5

nicks (approved):

PR #5668, opened 2022-04-06
Build dependency graph for k8s deletion in tilt down

### LINKED ISSUE #4431: `tilt down` doesn't account for resource_deps
### Steps to Reproduce
1. define a Tiltfile with two resources. "m2", which has a resource_dep on "m1", and "m1".
2. `tilt down`

### Expected Behavior
`tilt down` deletes m2's k8s entities before m1's

### Current Behavior
`tilt down` deletes m1's k8s entities before m2's (at least, it can--depends on declaration order)

#### About Your Use Case
`tilt down`'s current behavior is basically to delete resources in the order of `reverse(k8ssort(allManifests.flatMap(m => m.entities)))`, where k8ssort is a function that simply orders entities [by kinds](https://github.com/tilt-dev/tilt/blob/334baabc141cfe72b2f49a35de5d46230d090991/internal/kustomize/ordering.go#L32), where, e.g. deployments always come after namespaces and configmaps. This is because we want to ensure we, e.g., delete namespaces after the deployments that contain them, or configmaps after the deployments that depend on them.
If there is a dependency not covered by that entity order (e.g., the argo crd EventBus and its controller), then Tilt might try to delete them in the wrong order, leading to a failed deletion.

grahambrereton-form3:
When running tilt down, the order of deletion isn't dependent on the
resource_deps specified. These changes introduce a dependency graph to
the down command to determine the correct deletion order.

Fixes #4431

### DIFF
--- a/internal/cli/down.go
+++ b/internal/cli/down.go
@@ -25,6 +25,12 @@ type downCmd struct {
 	downDepsProvider func(ctx context.Context, tiltAnalytics *analytics.TiltAnalytics, subcommand model.TiltSubcommand) (DownDeps, error)
 }
 
+type dependencyNode struct {
+	manifest   model.Manifest
+	dependents []*dependencyNode
+	processed  bool
+}
+
 func newDownCmd() *downCmd {
 	return &downCmd{downDepsProvider: wireDownDeps}
 }
@@ -77,7 +83,7 @@ func (c *downCmd) down(ctx context.Context, downDeps DownDeps, args []string) er
 		return err
 	}
 
-	if err := deleteK8sEntities(ctx, tlr.Manifests, tlr.UpdateSettings, downDeps, c.deleteNamespaces); err != nil {
+	if err := deleteK8sManifests(ctx, tlr.Manifests, tlr.UpdateSettings, downDeps, c.deleteNamespaces); err != nil {
 		return err
 	}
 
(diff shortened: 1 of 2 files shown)


nicks:
ya, there was a long circleci outage today and some of the infra errors are still clearing out

grahambrereton-form3:
It looks like the tests are failing due to an auth error with pulumi:
```
++ pulumi whoami --non-interactive
warning: A new version of Pulumi is available. To upgrade from version '3.26.1' to '3.28.0', run 
   $ curl -sSL https://get.pulumi.com | sh
or visit https://pulumi.com/docs/reference/install/ for manual instructions and release notes.
error: PULUMI_ACCESS_TOKEN must be set for login during non-interactive CLI sessions
+ PULUMI_USER=
make: *** [Makefile:92: test-extensions] Error 255
```

@nicks This logging in one of your PRs looks different:
```
running pulumi/test/test.sh
++ pulumi whoami --non-interactive
Logging in using access token from PULUMI_ACCESS_TOKEN
warning: A new version of Pulumi is available. To upgrade from version '3.26.1' to '3.28.0', run 
   $ curl -sSL https://get.pulumi.com | sh
or visit https://pulumi.com/docs/reference/install/ for manual instructions and release notes.
+ PULUMI_USER=nicks
+ [[ nicks == '' ]]
+ yarn install
```

Are the tests just failing on this PR because there are no credentials associated with my account or something like that?

internal/cli/down.go
```diff
@@ -100,14 +106,59 @@ func (c *downCmd) down(ctx context.Context, downDeps DownDeps, args []string) er
 	return nil
 }
 
+func deleteK8sManifests(ctx context.Context, manifests []model.Manifest, updateSettings model.UpdateSettings, downDeps DownDeps, deleteNamespaces bool) error {
+	nodes := []*dependencyNode{}
```
nicks, line 110: the api here is a bit odd. why only sort when deleting kubernetes resources? seems like docker-compose would benefit from this as well.

can you factor this out into a function that sorts the list of manifests in reverse order and returns it, then sort the list before of deleteK8sManifests / delete dc manifests?

internal/cli/down.go
```diff
@@ -100,14 +106,59 @@ func (c *downCmd) down(ctx context.Context, downDeps DownDeps, args []string) er
...
+
+		nodes = append(nodes, &dependencyNode{
+			manifest:   manifest,
+			dependents: []*dependencyNode{},
+		})
+	}
+
+	for _, node := range nodes {
+		for _, resourceDep := range node.manifest.ResourceDependencies {
+			for _, other := range nodes {
```
nicks, line 123: would it make sense to have a map of nodes by name? simply to make the lookup here easier to read / more ergonomic

nicks (approved):
LGTM, thanks!

I left some small clarification questions / possible suggestions for improvements. will merge next week either way before the next release.

sorry about the pulumi tests -- i added them recently, and the circleci rules for login credentials in CI jobs are incrutable. should be ok now

internal/cli/down.go
grahambrereton-form3: I think I've addressed all your comments now. This is now a function `sortManifestsForDeletion` and I call it in the `down` function before k8s and dc deletion. I'm not 100% sure I understand the dc case, since it looks like it assumes there's a single dc project, so let me know if you'd like something different from what I just pushed.

nicks (approved):

internal/cli/down.go
nicks: we're working on supporting multiple dc projects :)

PR #5674, opened 2022-04-07
web: add Cluster Status to global nav in storybook

lizzthabet:
This PR adds a cluster button to the global navigation menu that displays information from the cluster API object. Since we're still adding information to the cluster API object, I've only passed the cluster info to the nav in storybook components, so the cluster button won't display for any users yet. When we're ready to have cluster information visible in the dashboard, I'll put up a separate PR.

The best way to test is to run `web/storybook.tiltfile` and look at `Shared > Header Bar > Nav` (with and without error) stories. You can manipulate the `clusterConnection()` method in `testdata.tsx` to change the cluster API object.

Cluster button and dialog in a healthy state:
Screen Shot 2022-04-07 at 11 00 03 AM

Cluster button and dialog with an error:
Screen Shot 2022-04-07 at 11 00 16 AM


web/src/ClusterStatusDialog.test.tsx
```diff
@@ -0,0 +1,136 @@
...
+
+const DEFAULT_TEST_PROPS: ClusterStatusDialogProps = {
+  open: true,
+  onClose: () => console.log("closing dialog"),
+  anchorEl: TEST_BUTTON,
+}
+const HEALTHY_CLUSTER = clusterConnection()
+const UNHEALTHY_CLUSTER = clusterConnection("Yikes, something went wrong.")
+
+describe("ClusterStatusDialog", () => {
```
milas, line 22: These tests are positively delightful ✨

web/src/FloatDialog.tsx
```diff
@@ -104,8 +106,8 @@ export default function FloatDialog(props: props) {
     >
       <TitleBar>
         {titleEl}
-        <CloseButton onClick={props.onClose}>
-          <CloseSvg />
+        <CloseButton aria-label="Close dialog" onClick={props.onClose}>
+          <CloseSvg role="presentation" />
```
milas, line 110: 🙌

web/src/HeaderBar.stories.tsx
```diff
@@ -46,3 +54,49 @@ export const UpgradeAvailable = () => {
   status!.versionSettings = { checkUpdates: true }
   return <HeaderBar view={view} currentPage={AnalyticsType.Detail} />
 }
+
+// TODO (lizz): Use HeaderBar component instead of GlobalNav
+// when design & implementation are finalized
```
milas, line 59: Just checking that this is still a valid `TODO` (i.e. you're not ready to move to `HeaderBar` here yet)

milas (approved):

web/src/HeaderBar.stories.tsx
lizzthabet: yup! i'll do this when we're ready to ship the cluster status info to users. right now, the cluster info is only passed to the `GlobalNav` component in storybook.

PR #5680, opened 2022-04-08
engine: remove Refs from ImageTarget

milas:
This is another place where we need to use the `Cluster` object
to determine the registry to resolve image references on the fly.

The brunt of the fallout here is actually throughout tests; many
of these now rely on the `.ImageMapSpec.Selector` directly, but
some that are testing specific registry-rewriting functionality
have been adapted to work.

Marking this as a draft because it depends upon #5677.

### DIFF
--- a/internal/build/image_builder.go
+++ b/internal/build/image_builder.go
@@ -186,7 +186,7 @@ func (ib *ImageBuilder) shouldUseKINDLoad(refs container.TaggedRefs, cluster *v1
 	// if we're using KIND and the image has a separate ref by which it's referred to
 	// in the cluster, that implies that we have a local registry in place, and should
 	// push to that instead of using KIND load.
-	if refs.LocalRef != refs.ClusterRef {
+	if refs.LocalRef.String() != refs.ClusterRef.String() {
 		return false
 	}
 
--- a/internal/build/image_builder.go
+++ b/internal/build/image_builder.go
@@ -41,7 +41,7 @@ func (ib *ImageBuilder) CanReuseRef(ctx context.Context, iTarget model.ImageTarg
 		return true, nil
 	}
 	return false, fmt.Errorf("image %q has no valid buildDetails (neither "+
-		"DockerBuild nor CustomBuild)", iTarget.Refs.ConfigurationRef)
+		"DockerBuild nor CustomBuild)", iTarget.ImageMapSpec.Selector)
 }
 
 // Build the image, and push it if necessary.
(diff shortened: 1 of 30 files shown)


milas:
Ahhh integration test failures are legit - I definitely broke Docker Pruning from the CLI 🙃

internal/build/image_builder.go
```diff
@@ -180,7 +186,7 @@ func (ib *ImageBuilder) shouldUseKINDLoad(ctx context.Context, iTarg model.Image
 	// if we're using KIND and the image has a separate ref by which it's referred to
 	// in the cluster, that implies that we have a local registry in place, and should
 	// push to that instead of using KIND load.
-	if iTarg.HasDistinctClusterRef() {
+	if refs.LocalRef != refs.ClusterRef {
```
nicks, line 189: i don't think this equality check is right? i think you need to convert to strings before comparing

internal/controllers/apis/cluster/registry.go
```diff
@@ -0,0 +1,18 @@
+package cluster
+
+import (
+	"github.com/tilt-dev/tilt/internal/container"
+	"github.com/tilt-dev/tilt/pkg/apis/core/v1alpha1"
+)
+
+func RegistryHosting(registry *container.Registry) *v1alpha1.RegistryHosting {
```
nicks, line 8: fwiw, we should eventually move towards getting rid of container.Registry and use the api struct everywhere

pkg/model/manifest.go
```diff
@@ -422,11 +422,23 @@ func (m Manifest) ManifestName() ManifestName {
 	return m.Name
 }
 
-func LocalRefSelectorsForManifests(manifests []Manifest) []container.RefSelector {
+func LocalRefSelectorsForManifests(manifests []Manifest, clusters map[string]*v1alpha1.Cluster) []container.RefSelector {
 	var res []container.RefSelector
 	for _, m := range manifests {
+		cluster := clusters[m.ClusterName()]
+		if m.ClusterName() != "" && cluster == nil {
```
nicks, line 429: this 'if' branch is kind of confusing, is it necessary? seems like everything should be ok if the cluster is nil

nicks (approved):

pkg/model/manifest.go
milas: I got rid of it - it was trying to skip any targets where we were missing the `Cluster` object since we otherwise might produce the wrong ref (i.e. w/o local registry). That said, it's harmless if it attempts and arguably not any less confusing than silently skipping it, so I don't feel bad about nixing it 🙃

PR #5682, opened 2022-04-11
api: add Cluster field to PortForward spec

milas:
This will be used to allow the PortForward reconciler to dynamically
react to Kubernetes cluster changes.


pkg/apis/core/v1alpha1/portforward_types.go
```diff
@@ -169,6 +176,12 @@ func (in *PortForward) Validate(_ context.Context) field.ErrorList {
 	return fieldErrors
 }
 
+func (in *PortForward) Default() {
```
nicks, line 179: i would assign this to the Defaulter interface, just to make sure they didn't add extra methods or smth

var _ resourcestrategy.Defaulter = &PortForward{}

nicks (approved):

PR #5683, opened 2022-04-11
portforward: add indexing for Cluster

milas:
This is very standard except that I added some extra logic to
the fake client to apply defaults as they are in the real API
server.


internal/controllers/fake/client.go
```diff
@@ -81,3 +93,24 @@ func (f fakeStatusWriter) Patch(ctx context.Context, obj ctrlclient.Object, patc
...
+
+// simulateMarshalRoundtrip makes API types behave closer to reality for the
+// fake client.
+//
+// NOTE(nick): We've had a lot of bugs due to the way the apiserver
+// modifies and truncates objects on update. We simulate this behavior
+// to catch this class of bug.
+//
+// The `apiResp` argument is marshaled and then unmarshalled into `dest` arg.
+// (This allows the fake client to mutate *copy* of the user-provided obj, and
```
lizzthabet, line 105: (tiny nit) should this be:
```suggestion
// (This allows the fake client to mutate a *copy* of the user-provided obj, and
```
or maybe:
```suggestion
// (This allows the fake client to mutate the *copy* of the user-provided obj, and
```
?

lizzthabet (approved):
lgtm! i have less knowledge about the testing side of things, but all the indexing looks good.

internal/controllers/core/portforward/reconciler.go
```diff
@@ -71,7 +81,11 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 func (r *Reconciler) reconcile(ctx context.Context, name types.NamespacedName) error {
 	pf := &PortForward{}
 	err := r.ctrlClient.Get(ctx, name, pf)
+	if ctrlclient.IgnoreNotFound(err) != nil {
```
nicks, line 84: usually you'd do the IsNotFound version

nicks (approved):

PR #5684, opened 2022-04-12
web: centralize button test helpers

lizzthabet:
This PR combines the existing `oneButton` and `makeUIButton` helpers into a single `oneUIButton` function that's used in storybook and tests. It also moves the button input spec helpers into `testdata.tsx`, so that all the data-generating helpers live in one place.


milas (approved):

web/src/testdata.tsx
```diff
@@ -66,52 +82,9 @@ const TEST_DATA_LABELS = [
   "javascript",
 ]
 
-// NOTE(dmiller) 4-02-19 this function is currently unused but I'm going to keep it around.
-// I have a feeling that it will come in handy later.
```
landism, line 70: hehehe

landism (approved):
👐

PR #5687, opened 2022-04-13
web: display cluster connection status in global nav

lizzthabet:
This PR passes the cluster information from the `view` to the `GlobalNav` component, so that cluster status will be visible to users. (Previously the cluster info was only added to storybook components, so we had flexibility to do more API work before showing that information explicitly to users.)

I'll keep this PR in draft state till we're ready to launch. 

Closes [ch13397](https://app.shortcut.com/windmill/story/13397/as-user-i-see-cluster-status-with-dropdown-menu-that-shows-only-k8s-info-no-buttons).

### DIFF
--- a/web/src/GlobalNav.tsx
+++ b/web/src/GlobalNav.tsx
@@ -167,7 +167,7 @@ export function MenuButtonLabeled(
   )
 }
 
-type GlobalNavProps = {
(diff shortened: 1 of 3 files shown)


milas (approved):
🙌

PR #5689, opened 2022-04-13
cluster: improve unrecoverable error message

milas:
Until the `cluster_refresh` feature is done, initialization errors
for the cluster are unrecoverable, as not all parts of Tilt know
how to re-initialize the client/themsleves properly on change.

Tilt web UI showing a cluster error message

The error message is now wrapped with some additional context and
a notice that Tilt needs to be restarted after fixing the issue.


lizzthabet (approved):
🎉

PR #5691, opened 2022-04-13
web: refactor ApiButton tests to use react-testing-library instead of enzyme

lizzthabet:
I'm sorry this diff is massively large. 😅 All the tests should be the same, both in title and what they're testing! Some are shorter because RTL has more concise interfaces for stuff like user events, and some are longer because I used more robust assertions, `expect().toBeInDocument()` and `expect().toHaveValue()`, and it can be wordier to query by label text.

### DIFF
--- a/web/src/ApiButton.tsx
+++ b/web/src/ApiButton.tsx
@@ -293,6 +293,7 @@ function ApiButtonWithOptions(props: ApiButtonWithOptionsProps & ButtonProps) {
       >
         {props.submit}
         <ApiButtonInputsToggleButton
+          {...buttonProps}
           size="small"
           onClick={() => {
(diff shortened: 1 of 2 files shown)


web/src/ApiButton.tsx
```diff
@@ -293,14 +293,14 @@ function ApiButtonWithOptions(props: ApiButtonWithOptionsProps & ButtonProps) {
       >
         {props.submit}
         <ApiButtonInputsToggleButton
+          {...buttonProps}
```
lizzthabet, line 296: the `aria-label` that gets set on the parent component is unintentionally passed down to this `ToggleButton` in `buttonProps`, so moving the props up here allows the correct label to be specified on this component.

milas (approved):
✨

PR #5694, opened 2022-04-14
k8s_custom_deploy: fix Live Update rebuild

milas:
The `k8s_custom_deploy` is unusual: it's possible to use with
no Tilt-managed image build while still getting Live Update
functionality. This is useful for scenarios where you have an
all-in-one build&deploy tool (e.g. Knative) you want to use
with Tilt.

However, this meant that the value of the `deps` arg was being
passed both to the deploy target (K8s/`apply_cmd`) as well as
the image target (Live Update), which introduced a race between
them. If the Live Update reconciler saw a (supported) file change
first, it would clear it from the pending file list on the deploy
target once done, and so the deploy target would be skipped, as
it no longer had any pending changs. On the other hand, if the
deploy target saw it first, it'd start running because deploy
target changes can't be Live Update'd. At the same time, the Live
Update reconciler would still run, as it does not have a concept
of blocking/waiting for other targets.

To address this, the `deps` arg is now filtered using the generated
`LiveUpdateSpec` and any paths matching it (from `sync()` and
`fall_back_on` rules) are added as ignore rules to the deploy target.

### DIFF
--- a/integration/k8s_custom_deploy_test.go
+++ b/integration/k8s_custom_deploy_test.go
@@ -62,7 +62,7 @@ func TestK8sCustomDeploy(t *testing.T) {
 	defer cancel()
 	f.CurlUntil(ctx, "http://localhost:54871", "Greetings from Live Update!")
 
-	f.Touch(filepath.Join("web", "fallback.txt"))
+	f.Touch("fallback.txt")
 	ctx, cancel = context.WithTimeout(f.ctx, time.Minute)
 	defer cancel()
 	f.CurlUntil(ctx, "http://localhost:54871", "Welcome to nginx!")
--- a/internal/tiltfile/k8s_custom_deploy.go
+++ b/internal/tiltfile/k8s_custom_deploy.go
@@ -19,6 +19,7 @@ type k8sCustomDeploy struct {
 	applyCmd  model.Cmd
 	deleteCmd model.Cmd
 	deps      []string
+	ignores   []model.Dockerignore
 }
 
 func (s *tiltfileState) k8sCustomDeploy(thread *starlark.Thread, fn *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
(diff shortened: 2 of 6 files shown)


integration/k8s_custom_deploy_test.go
```diff
@@ -62,7 +62,7 @@ func TestK8sCustomDeploy(t *testing.T) {
 	defer cancel()
 	f.CurlUntil(ctx, "http://localhost:54871", "Greetings from Live Update!")
 
-	f.Touch(filepath.Join("web", "fallback.txt"))
+	f.Touch("fallback.txt")
```
nicks, line 65: maybe we can also add a check here that the pod ID didn't change / an image build didn't start?

internal/tiltfile/k8s_custom_deploy.go
```diff
@@ -149,3 +151,24 @@ func (s *tiltfileState) k8sCustomDeploy(thread *starlark.Thread, fn *starlark.Bu
 
 	return starlark.None, nil
 }
+
+func filterDepsForCustomDeploy(deps []string, spec v1alpha1.LiveUpdateSpec) (deployPaths []string, luPaths []string) {
+	luMatcher := pathMatcherFromLiveUpdateSpec(spec)
+	for _, dep := range deps {
+		if match, _ := luMatcher.Matches(dep); match {
```
nicks, line 158: i don't think this is quite right, since you could have:

k8s_custom_deploy(deps=['.'], live_update=[sync('./src', ...)])

i think you probably need to add the live update paths as an ignore path?

milas: Ahhh great point. Changed things around in 98cf10ea9, integration test now uses `deps=['.']` and passes 👍

integration/k8s_custom_deploy_test.go
milas: Hmm yeah let me see what we've got handy here 🔨

nicks (approved):

integration/k8s_custom_deploy_test.go
milas: Ah you know what, this test is actually already set up in a way that it'll fail if the pod is (or isn't) replaced based on what it's going.

It's kind of a "file sync only" setup in that it's using `nginx` and then live updates in a file, and the `apply_cmd` is `kubectl delete && kubectl create` so that we always destroy things.

PR #5695, opened 2022-04-14
web: use react testing library instead of enzyme in select tests

lizzthabet:
I'm refactoring tests to replace Enzyme with React Testing Library and breaking them into a couple PRs since it'll be easier to review. 📋 These tests were pretty straightforward to tackle, and I'll work on updating the more complex ones (like `HUD.test.tsx`) later.


web/src/HelpDialog.test.tsx
```diff
@@ -1,12 +0,0 @@
...
-import { MemoryRouter } from "react-router-dom"
-import { DialogOverview } from "./HelpDialog.stories"
-
-it("renders overview dialog", () => {
-  mount(
-    <MemoryRouter initialEntries={["/"]}>
-      <DialogOverview />
-    </MemoryRouter>
-  )
-})
```
lizzthabet, line 12: this test didn't have any assertions, so i deleted it. (plus the component itself is mostly just rendering data with the shared dialog component, so there's not really functionality to test.)

nicks (approved):

Issue #5567, opened 2022-03-05
Keep log regex filter on a resource when switching resources


adamf:
## Describe the Feature You Want
I want log filters to persist on a per-resource basis even when switching to and from resources.

### Current Behavior
If I'm looking at resource A, and put in a log filtering regex, then switch to resource B, and switch back to resource A, the log filter is gone.

### Why Do You Want This?
I often run a filter like `/^((?!(kube)).)*$/` to filter out logs from kubernetes probes. It's annoying that when I switch off the resource I'm filtering to check another resource, then flip back, I lose the filter.

**Additional context**

lianmakesthings:
Hi @adamf,
thanks for filing this issue.
The filter string is part of the URL when you apply it. So, if you're really just switching back and forth between two resources, you could navigate back in the browser, which would take you to resource A with the regex filled in.
Would the work for you as a temporary workaround?

PR #5706, opened 2022-04-19
web: preserve filters across resource views

nicks:
Hello @lizzthabet,

Please review the following commits I made in branch nicks/issue5567:

d12c91ea2455b7a930256cc423881f60185335d2 (2022-04-19 12:22:55 -0400)
web: preserve filters across resource views
fixes https://github.com/tilt-dev/tilt/issues/5567

### DIFF
--- a/web/src/ResourceNav.tsx
+++ b/web/src/ResourceNav.tsx
@@ -1,4 +1,10 @@
-import React, { useCallback, useContext, useMemo } from "react"
+import React, {
+  useCallback,
+  useContext,
+  useEffect,
+  useMemo,
+  useState,
+} from "react"
 import { matchPath, useHistory, useLocation } from "react-router-dom"
 import { usePathBuilder } from "./PathBuilder"
 import { ResourceName } from "./types"
@@ -48,6 +54,9 @@ export function ResourceNavProvider(
   let location = useLocation()
   let pb = usePathBuilder()
   let selectedResource = ""
+  let [filterByResource, setFilterByResource] = useState(
+    {} as { [key: string]: string }
+  )
   let invalidResource = ""
 
   let matchResource = matchPath(location.pathname, {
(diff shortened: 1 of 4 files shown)


nicks:
to be clear - i'm not sure if this is a good idea or not, but seems safe to try!

lizzthabet (approved):
yeah, that sounds good to me! this is a little different than the original request (which was to preserve filters on a per-resource basis), but i think it's fine to try. 

would you mind adding a test before merging in? i refactored the resource nav tests to use rtl already in a separate branch, but i can rebase and update whatever test you add.

nicks:
oh! good catch! I missed that part. ya, let me try to do something closer to that.

nicks:
ok, updated it with different behavior that more closely matches the feature request

web/src/ResourceNav.tsx
```diff
@@ -62,14 +71,34 @@ export function ResourceNavProvider(
...
+  useEffect(() => {
+    let existing = filterByResource[selectedResource] || ""
+    if (existing != search) {
+      let obj = {} as { [key: string]: string }
+      Object.assign(obj, filterByResource)
+      obj[selectedResource] = search
+      setFilterByResource(obj)
+    }
+  }, [selectedResource, search])
+
```
lizzthabet, line 85: (nit) one thing i noticed is that if there's a filter term that's saved on a resource and then the filter state is reset (that is, no filters are applied), this is still saving the empty query string `level=&source=&term=`.

lizzthabet, line 85: (nit) one thing i noticed is that if there's a filter term that's saved on a resource and then the filter state is reset (that is, no filters are applied), this is still saving the empty query string `level=&source=&term=`.

lizzthabet (approved):
i like this behavior more!

web/src/ResourceNav.tsx
nicks: added a fix for that!

PR #5707, opened 2022-04-19
web: switch HUD tests to use react test-utils instead of enzyme

nicks:
Hello @lizzthabet,

Please review the following commits I made in branch nicks/reacttesting:

a729e8594ffe49a65515beb5ffa2e323e97c2624 (2022-04-19 14:31:46 -0400)
web: switch HUD tests to use react test-utils instead of enzyme

### DIFF
--- a/web/src/HUD.test.tsx
+++ b/web/src/HUD.test.tsx
@@ -50,8 +50,8 @@ beforeEach(() => {
 
 it("renders reconnecting bar", async () => {
   let root = emptyHUD()
-  let rootTree = renderIntoDocument(root)
-  let container = ReactDOM.findDOMNode(rootTree)
+  let rootTree = renderIntoDocument(root) as any
+  let container = ReactDOM.findDOMNode(rootTree)! as HTMLElement
   expect(container.textContent).toEqual(expect.stringContaining("Loading"))
 
   const hud = findRenderedComponentWithType(rootTree, HUD)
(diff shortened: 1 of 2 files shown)


nicks:
react-testing-library doesn't have a good way to get the component instance, but this lower-level library does, so we'll just use that for tests that need it.

lizzthabet (approved):
🎉

web/src/HUD.test.tsx
```diff
@@ -40,44 +43,36 @@ const emptyHUD = () => {
...
-  ReactDOM.render(emptyHUD(), div)
-  ReactDOM.unmountComponentAtNode(div)
-})
-
 it("renders reconnecting bar", async () => {
-  const root = mount(emptyHUD())
-  const hud = root.find(HUD)
-  expect(hud.text()).toEqual(expect.stringContaining("Loading"))
+  let root = emptyHUD()
+  let rootTree = renderIntoDocument(root)
```
lizzthabet, line 53: oh! it looks like this may not return the tree. i'm seeing type errors locally and in ci.

lizzthabet, line 53: oh! it looks like this may not return the tree. i'm seeing type errors locally and in ci.

nicks: fixed! tbh i don't know why react thinks this returns void

PR #5711, opened 2022-04-20
podlogstream: fix flaky test

### LINKED ISSUE #5441: Flaky Test:  podlogstreamcontroller_test.go::TestLogsFailed
### Expected Behavior
Consistency and deterministic test results

### Current Behavior
with no change in the environment, occasionally this test fails.  It was first noticed by me as the output of https://github.com/tilt-dev/tilt/pull/5440 shown on circleci here:  https://app.circleci.com/pipelines/github/tilt-dev/tilt/10689/workflows/c546b656-d6bb-42e9-ac1e-fabfd5c9efa8/jobs/83675

After a review, it seemed clear the code wasn't the issue.  It was by chance that I ran testing several times locally and eventually saw the same error.

### Steps to Reproduce

Well... that's the hard part... isn't it:)   The steps are to run the tests.   Run, Test Run!  multiple times!   It is unclear what the alignment of the planets must be... or if there was a rampant neutrino passing thru my machine.  or some kind of crazy quantum particle entanglement with circleci.

My experience has been this is usually the result of:
1. eventual consistency / timeout 
2. poor test design with no wait on what should be eventually consistent

I've tried `for i in {1..100}; do go test -run TestLogsFailed ./internal/controllers/core/podlogstream; done` but with no success... err failure.

#### `tilt doctor` Output

Not needed... this is a bug on test not on tilt itself

milas:
Reconciliation is async, so use `require.Eventually`. Other tests
doing similar things here were already using this pattern.

Fixes #5441.

### DIFF
--- a/internal/controllers/core/podlogstream/podlogstreamcontroller_test.go
+++ b/internal/controllers/core/podlogstream/podlogstreamcontroller_test.go
@@ -11,6 +11,7 @@ import (
 	"github.com/jonboulle/clockwork"
 	"github.com/stretchr/testify/require"
 
+	"github.com/tilt-dev/tilt/internal/controllers/apicmp"
 	"github.com/tilt-dev/tilt/internal/timecmp"
 	"github.com/tilt-dev/tilt/pkg/apis"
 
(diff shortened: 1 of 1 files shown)


milas:
FWIW I could reproduce the failure pretty consistently with `-count 1000` previously but cannot anymore

nicks (approved):

PR #5712, opened 2022-04-20
web: refactor tests to use react testing library instead of enzyme (part two!)

lizzthabet:
* Refactor more tests that use Enzyme to use React Testing Library
* Add ARIA labels to improve markup and make testing easier

Again, apologies for the large diff! The tests should be exactly the same (except for a couple i broke into smaller tests and one or two i removed), but the way that the diff generates, it doesn't make the comparison easy.


web/src/OverviewLogPane.test.tsx
```diff
@@ -22,152 +23,133 @@ import {
...
-  expect(el.querySelectorAll(".LogLine a")).toHaveLength(2)
-  expect(el.querySelectorAll(".LogLine button")).toHaveLength(0)
-})
+  it("escapes html and linkifies", () => {
+    customRender(<StyledLines />)
+    expect(screen.getAllByRole("link")).toHaveLength(2)
+    expect(screen.queryByRole("button")).toBeNull()
+  })
 
-it("filters by source", () => {
```
lizzthabet, line 52: i broke these larger filter tests with multiple `expect` scenarios into smaller tests.

web/src/OverviewLogPane.test.tsx
```diff
@@ -22,152 +23,133 @@ import {
...
 })
 
+/**
+ * The following tests rely on testing React component state directly,
+ * which is not possible to do with React Testing Library. They'll need to
+ * either use React's test utilities (which involve some funky type manipulation)
+ * or perhaps modify/wrap the component to render its state to the DOM instead.
+ */
+
 it("engages autoscrolls on scroll down", () => {
```
lizzthabet, line 153: i'll refactor these in a separate pr.

web/src/OverviewLogPane.tsx
```diff
@@ -491,8 +489,6 @@ export class OverviewLogComponent extends Component<OverviewLogComponentProps> {
       ? logStore.manifestLogPatchSet(mn, startCheckpoint)
       : logStore.allLogPatchSet(startCheckpoint)
 
-    let { source, level } = this.props.filterSet
-
```
lizzthabet, line 495: these deleted vars were unused!

web/src/ResourceStatusSummary.tsx
```diff
@@ -109,29 +108,24 @@ type ResourceGroupStatusItemProps = {
...
             {props.countOutOf}
           </ResourceGroupStatusSummaryItemCount>
         </>
       )}
     </>
   )
 
-  const descriptiveCount = (
-    <>
-      {count} <SrOnly>&nbsp; {props.label}</SrOnly>
```
lizzthabet, line 128: i changed the a11y markup here from additional inline text to using aria labels, since it's easier to test. i'm not sure if there's a meaningful difference between the two approaches for screen readers and other assistive tech.

PR #5714, opened 2022-04-21
web: refactor AnalyticsNudge component and tests for react testing library

lizzthabet:
* Refactor `AnalyticsNudge` component tests to use React Testing Library instead of Enzyme and remove snapshots
* Fix an issue where the nudge is always visible/accessible to screen readers, regardless of when it should be displayed
* Fix HTML warning that `pre` shouldn't be a descendant of `p` elements

There's one test I couldn't figure out how to pass without throwing a React warning. I left the test in (but not running) with a comment. If there are any suggestions, lmk!


web/src/AnalyticsNudge.test.tsx
```diff
@@ -1,67 +1,120 @@
...
+
+    userEvent.click(screen.getByRole("button", { name: /I'm in/i }))
+
+    await waitFor(() => {
+      expect(screen.getByRole("alert")).toBeInTheDocument()
+    })
+  })
 
-  expect(component).toMatchSnapshot()
+  // TODO: Fix this test, if possible! There's a timing issue with the
```
nicks, line 95: ya, this is a legit bug!

in OverviewLogPane, we handle this by keeping track of the timer ids, and cancelling them on unmount. we should do something similar here

https://github.com/tilt-dev/tilt/blob/master/web/src/OverviewLogPane.tsx#L301

lizzthabet: ah, yes! thanks for the pointer. i fixed this and the test is now passing. :tada:

nicks (approved):

PR #5715, opened 2022-04-21
apis: add an api for the dockercompose log streamer

nicks:
Hello @milas,

Please review the following commits I made in branch nicks/dclogs3:

f3c90c08d3f7f87ad8564e831ecbf3f527ebef13 (2022-04-21 13:10:43 -0400)
apis: add an api for the dockercompose log streamer


nicks:
this is almost all boilerplate and copying the podlogstream

pkg/apis/core/v1alpha1/dockercomposelogstream_types.go
```diff
@@ -0,0 +1,132 @@
...
+
+// DockerComposeLogStreamSpec defines the desired state of DockerComposeLogStream
+type DockerComposeLogStreamSpec struct {
+	// The name of the service to stream from.
+	Service string `json:"service" protobuf:"bytes,1,opt,name=service"`
+
+	// A specification of the project the service belongs to.
+	//
+	// Each service spec keeps its own copy of the project spec.
+	Project DockerComposeProject `json:"project" protobuf:"bytes,2,opt,name=project"`
```
milas, line 62: Just curious, what's the reason for embedding the project vs having something like `DockerComposeService string` that points to the main API object?

milas (approved):

pkg/apis/core/v1alpha1/dockercomposelogstream_types.go
nicks: ya, i went back and forth on this

my main thinking is that we should have the log streamers be independent objects that you can use without deploying anything, much like PodLogStream, and trying to keep the general pattern as much like PodLogStream as possible.

PR #5727, opened 2022-04-25
feat(helm): fix #5726, add kube-version flag support to helm command

### LINKED ISSUE #5726: Add support for `--kube-version` in helm command
* when using `helm template` helm take the version of the kubectl installed to check Kube API 
* you can use `--kube-version` ( ex: `helm --kube-version 1.17` ) to override this behavior
* would be cool if tilt can support this flag.

### Current Behavior
* tilt does not pass the flag, so `helm template` default to kubectl client version

### Why Do You Want This?
* I have cluster with different version of k8s ( from 1.17 to 1.22 ) which have moved API to different namespaces.

olivierchatry:
This add a parameter to `helm` command (kube-version) to support `--kube-version` `helm template` flag.

### DIFF
--- a/internal/tiltfile/api/__init__.py
+++ b/internal/tiltfile/api/__init__.py
@@ -711,7 +711,7 @@ def kustomize(pathToDir: str, kustomize_bin: str = None) -> Blob:
     kustomize_bin: Custom path to the ``kustomize`` binary executable. Defaults to searching $PATH for kustomize."""
   pass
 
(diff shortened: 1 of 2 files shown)


nicks:
thanks @olivierchatry !

can you add this to the api documentation as well https://github.com/tilt-dev/tilt/blob/master/internal/tiltfile/api/__init__.py#L714 ?

that way the doc site and IDE plugins will get the new parameter on the next release

olivierchatry:
Implement https://github.com/tilt-dev/tilt/issues/5726

PR #5733, opened 2022-04-25
web: migrate more tests from Enzyme to React Testing Library (part three!)

lizzthabet:
This is the final batch of test migration to React Testing Library! The changes / diffs should be pretty straightforward and similar to the last two PRs with an additional note:

There are a several tests in `HUD` and `OverviewLogPane` that rely on changing and asserting directly on the component state. Following @nicks' example in `HUD.test.tsx`, I added a helper and rudimentary type guarding to make it easier to do these tests. 

(I'm not sure why but the [return type for `render`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react-dom/index.d.ts#L93-L97) is like _maybe you'll get something back, maybe you won't_, but at least passing it a type will give you a return type other than `void`!)


nicks (approved):

nicks:
woooooo

PR #5735, opened 2022-04-26
portforward: deprecate implicit container port behavior

### LINKED ISSUE #5728: Confusing behavior with port forwards when no container port specified
### Expected Behavior
* If a container has multiple exposed ports, and there's not an unambiguous mapping, warn or error

### Current Behavior
* Tilt picks one of the exposed ports (at random? first in list?)

### Steps to Reproduce
1. Add `k8s_yaml` for a resource that has multiple exposed ports
2. Add a `k8s_resource` call with `port_forwards=['9000']` (only a singular port)

### Context
My 2c about how to handle this:
* ~If there is exactly one exposed container port, map the local port to it regardless of if they match~
* ~If there are multiple exposed container ports and one of them matches the local port number, forward that one~
* ~If there are multiple exposed container ports and none of them match the local port number, emit a warning~
  * ~In a future version, this could become an error~
~(Note: in all these cases, the assumption is the user has passed a single port number, not a tuple)~

If no container port is specified, map to the same port number as the local port. For example, `port_forwards=8000` would mean `8000:8000` regardless of if `8000` (or any other ports) are exposed by the container spec.

#### About Your Use Case
This was raised in [Slack by Adam](https://kubernetes.slack.com/archives/CESBL84MV/p1650567370910339), who mentioned losing a few hours to debugging this when the wrong container port got picked.

milas:
Tilt supports a bunch of variations when defining a port forward in
a Tiltfile.

It's possible to specify a single value (a local port), and in this
case, Tilt will look for a matching exposed port to use. If none
is found, it'll use the first exposed port.

This behavior is confusing and doesn't match existing functionality
in e.g. `kubectl`.

Allowing a single port that will be used identically for local and
container is reasonable short-hand, but picking an implicit port
to map to adds extra complexity for questionable value.

In this latter case, a warning is now emitted along with the mapping
and instructions for the user to update their `Tiltfile` to make it
explicit before it breaks in a future release of Tilt.

To fully deprecate this, we can change the logic to always default
the container port to local port. However, as this will change the
behavior in the aforementioned case, it's a breaking change, so
let's provide a heads up first :)

Fixes #5728.

### DIFF
--- a/internal/controllers/core/kubernetesdiscovery/portforwards.go
+++ b/internal/controllers/core/kubernetesdiscovery/portforwards.go
@@ -95,7 +95,8 @@ func (r *Reconciler) toDesiredPortForward(kd *v1alpha1.KubernetesDiscovery) (*v1
 
 	pf := &v1alpha1.PortForward{
 		ObjectMeta: metav1.ObjectMeta{
-			Name: fmt.Sprintf("%s-%s", kd.Name, pod.Name),
+			Name:      fmt.Sprintf("%s-%s", kd.Name, pod.Name),
+			Namespace: kd.Namespace,
 			Annotations: map[string]string{
 				v1alpha1.AnnotationManifest: kd.Annotations[v1alpha1.AnnotationManifest],
 				v1alpha1.AnnotationSpanID:   kd.Annotations[v1alpha1.AnnotationSpanID],
--- a/internal/controllers/core/kubernetesdiscovery/portforwards.go
+++ b/internal/controllers/core/kubernetesdiscovery/portforwards.go
@@ -3,7 +3,6 @@ package kubernetesdiscovery
 import (
 	"context"
 	"fmt"
-	"strings"
 
 	v1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
@@ -195,8 +194,11 @@ func warnDeprecatedImplicitForwards(ctx context.Context, kd *v1alpha1.Kubernetes
 	if kd == nil || pf == nil {
 		return
 	}
+	resourceName := kd.Annotations[v1alpha1.AnnotationManifest]
+	if resourceName == "" {
+		return
+	}
 
-	var implicit []string
 	for _, pft := range kd.Spec.PortForwardTemplateSpec.Forwards {
 		if pft.ContainerPort != 0 {
 			continue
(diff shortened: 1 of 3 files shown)


nicksieger (approved):

internal/controllers/core/kubernetesdiscovery/portforwards.go
```diff
@@ -180,3 +189,32 @@ func isBetterPortForwardPod(podA, podB *v1alpha1.Pod) bool {
...
+			if pft.LocalPort == f.LocalPort && f.LocalPort != f.ContainerPort {
+				implicit = append(implicit,
+					fmt.Sprintf("  – %d ⇢ %d:%d", f.LocalPort, f.LocalPort, f.ContainerPort))
+			}
+		}
+	}
+
+	if len(implicit) != 0 {
+		logger.Get(ctx).Warnf(
+			"One or more port forwards did not specify a container port, and no exposed container port matched the local port.\n"+
```
nicks, line 214: a couple independent questions:

- maybe we should _always_ print the port forward mapping in the logs? i think that would have helped with the original complaint.

- I'm not sure most users are going to understand what to do with this error message...i would be much more explicit about what needs to change. Something like:

`k8s_resource(name=x, port_forward='8000')` currently maps `localhost:8000` to port 8001 in your container.

Tilt vX.Y.Z will change this default! k8s_resource(port_forward='8000')` will map `localhost:8000` to port 8000 in your container.

To keep your project working, change your Tiltfile to `k8s_resource(name=x, port_forward='8000:8001')`

internal/controllers/core/kubernetesdiscovery/portforwards.go
```diff
@@ -116,9 +121,12 @@ func (r *Reconciler) toDesiredPortForward(kd *v1alpha1.KubernetesDiscovery) (*v1
 //
 // TODO(nick): This is old legacy behavior, and I'm not totally sure it even
 // makes sense. I wonder if we should just insist that ContainerPort is populated.
-func populateContainerPorts(pf *v1alpha1.PortForward, pod *v1alpha1.Pod) {
+func populateContainerPorts(pft *v1alpha1.PortForwardTemplateSpec, pod *v1alpha1.Pod) []v1alpha1.Forward {
```
nicks, line 124: do you expect that we will still want to print this warning after we change the default? is there any reason not to change the default now?

nicks (commented):
a test would be nice, because this can be hard to reason about!

internal/controllers/core/kubernetesdiscovery/portforwards.go
milas: Good question, I was taking a fairly cautious approach here and adding the warning for now before changing the default. At that point, I could see an argument for keeping it to let users know that Tilt is now picking a different port than it would have in the past in this case, but that also might be overkill, e.g. when we **remove** deprecated functionality, we don't _generally_ leave in a warning if someone tries to use it, we just error out.

PR #5736, opened 2022-04-26
cli: remove deprecated --hud flag in favor of --legacy

milas:
The `--hud` flag has been deprecated for a bit and thrown a warning
on start (with a forced multi-second wait).

Users who wish to continue using the TUI may do so for now by running
`tilt up --legacy`.


milas:
I'm just going to keep this queued up for a bit, as it's a breaking change to the CLI, but I don't want a version bump just for this because it's really quite a silly breaking change!

PR #5741, opened 2022-04-27
tiltfile: improve k8s_yaml error messages

### LINKED ISSUE #5640: misleading / confusing "illegal base64 data at input byte 0" on invalid k8s secret values
### Expected Behavior
Tilt indicates what field in what object is invalid.

### Current Behavior
The output looks like this:
```
ERROR: Traceback (most recent call last):
  /Users/matt/go/src/github.com/tilt-dev/servantes/deploy/Tiltfile:1:9: in <toplevel>
  <builtin>: in k8s_yaml
Error: illegal base64 data at input byte 0
```
On first glance, it looks like it's trying to read the yaml as base64 and finding a problem at the first byte, which is very confusing! And then even if one is aware of the problem described in this issue, Tilt provides no help narrowing down what part of the yaml is problematic.

When trying to create the secret directly via `kubectl`, the output pinpoints the problem:
```
Error from server (BadRequest): error when creating "secrets.yaml": Secret in version "v1" cannot be handled as a Secret: v1.Secret.Data: decode base64: illegal base64 data at input byte 0
```

### Steps to Reproduce
Try to load any secret with a non-base64 value, e.g.:
```
k8s_yaml(blob('''apiVersion: v1
kind: Secret
metadata:
  name: mysecret
type: Opaque
data:
  stuff: "!"'''))
```
`tilt up`

#### `tilt doctor` Output

```
$ tilt doctor
Tilt: v0.26.3-dev, built 2022-03-28
System: darwin-arm64
---
Docker
- Host: [default]
- Server Version: 20.10.11
- API Version: 1.41
- Builder: 2
---
Kubernetes
- Env: kind
- Context: kind-kind
- Cluster Name: kind-kind
- Namespace: default
- Version: v1.21.1
- Cluster Local Registry: {Host:localhost:58059 hostFromCluster: SingleName:}```

### PR DESCRIPTION
**Pull request by milas**
If possible, contextualize the error with the object kind and
name.

Tiltfile error for a Kubernetes Secret with invalid base64 data

This is useful when the object is syntactically valid YAML/JSON
but fails parsing due to e.g. mismatched types on a field.

See #5640.

### DIFF
--- a/internal/k8s/serialize.go
+++ b/internal/k8s/serialize.go
@@ -3,6 +3,7 @@ package k8s
 import (
 	"bufio"
 	"bytes"
+	"fmt"
 	"io"
 
 	v1 "k8s.io/api/core/v1"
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: internal/k8s/serialize.go [hunk 1]
```diff
@@ -43,23 +44,47 @@ func decodeToRuntimeObj(ext runtime.RawExtension) (runtime.Object, error) {
...
-	if !runtime.IsNotRegisteredError(err) {
+	// decode as unstructured - if the _original_ decode error was due to it
+	// being a non-standard type, the unstructured object will be returned;
+	// otherwise, it'll be used to provide additional context to the error if
+	// possible
+	var unst unstructured.Unstructured
+	_, gvk, err :=
+		unstructured.UnstructuredJSONScheme.Decode(ext.Raw, nil, &unst)
+	if err != nil {
+		if decodeErr != nil {
```
landism, line 61: this is guaranteed to be true after line 49, yeah?

landism (approved):
this is a definite improvement, but I don't know if I'd close 5640 with it: the error message is still easy to read as saying the yaml is invalid base64 since it doesn't specify the problematic field

internal/k8s/serialize.go
milas: Yup 🤦 

Tightened this up in [8b9be03](https://github.com/tilt-dev/tilt/pull/5741/commits/8b9be03e8e759fd127466964241e82017e5be314)

milas:
> I don't know if I'd close 5640 with it: the error message is still easy to read as saying the yaml is invalid base64 since it doesn't specify the problematic field

Fair, I changed this to just be a "see" so it's linked but won't close it

PR #5742, opened 2022-04-27
docker: fix spurious double builds on Dockerfile change

### LINKED ISSUE #5586: Change to a `Dockerfile` sporadically results in erroneous re-use of cached image
## Expected Behavior
* A Docker build is triggered whenever the `Dockerfile` changes

## Current Behavior
* Sometimes, despite seeing the `Dockerfile` change, the Docker build will be skipped and the previously cached image will be used

## Steps to Reproduce
1. Create a project with a `docker_build` deployed to K8s
2. Change the `Dockerfile` (add/delete a commented line) <-- intermittent, so might have to try many times 😢 

## Context
I believe I know what's going on:
* The `(Tiltfile)` resource watches the `Dockerfile`s because it reads the file contents into memory to use for the image target (let's call it `image:foo`)
 * The `image:foo` target might _also_ be watching `Dockerfile` if not explicitly ignored by the file watch rules
 * Order these get seen is non-deterministic and can cause problems

### Tiltfile -> image (Happy Path)
1. `(Tiltfile)` is processed first, new Dockerfile contents are read in and it will clear the image build status as part of `ConfigsReloadedAction` reducer since the specs no longer match (as Dockerfile contents have changed)
2. `image:foo` is processed, sees that there is not prior build and executes, using the updated Dockerfile contents from its spec

### Image -> Tiltfile (Broken)
1. `image:foo` is processed, sees that `len(PendingFileChanges) != 0` (because `Dockerfile` is in there), but the re-executed build still has the old Dockerfile contents in its spec, so every layer is cached
2. _While the prior build is still occurring_, `(Tiltfile)` is processed, which clears out the image build status (see happy path) and sets `PendingManifestChange` to now
3. `image:foo` build finishes, storing the build status
4. Another `image:foo` build is kicked off due to `PendingManifestChange`, it sees that it has a prior build and no file changes, so it re-uses the (incorrect) cached result

## Solutions
* Ignore `Dockerfile` from the image target (I think we might want this regardless)
* Avoid adding build statuses for builds that started _before_ the pending manifest change

milas:
Do not _watch_ the Dockerfile from the image target, as this is
handled by the Tiltfile, which provides the full contents to the
target in-memory (i.e. the on-disk Dockerfile isn't actually used
when we execute the build).

This prevents race conditions from them both watching it, and
ensures that the flow always goes:
    (Dockerfile change) -> Tiltfile execute -> image build

Fixes #5586.

### DIFF
--- a/internal/tiltfile/tiltfile_state.go
+++ b/internal/tiltfile/tiltfile_state.go
@@ -1684,8 +1684,6 @@ func (s *tiltfileState) ignoresForImage(image *dockerImage) (contextIgnores []v1
 		return nil, nil, fmt.Errorf("reading dockerignore for %s: %v", image.configurationRef.RefFamiliarString(), err)
 	}
 	if image.tiltfilePath != "" {
-		// TODO(milas): technically, this should be a FW ignore, not a context
-		// 	ignore, but historically it has been this way
 		contextIgnores = append(contextIgnores, v1alpha1.IgnoreDef{BasePath: image.tiltfilePath})
 	}
 	contextIgnores = append(contextIgnores, s.repoIgnoresForImage(image)...)
--- a/internal/tiltfile/tiltfile_state.go
+++ b/internal/tiltfile/tiltfile_state.go
@@ -1421,19 +1421,10 @@ func (s *tiltfileState) imgTargetsForDepsHelper(mn model.ManifestName, imageMapD
 			iTarget.LiveUpdateName = liveupdate.GetName(mn, iTarget.ID())
 		}
 
-		dockerignores, err := s.dockerignoresForImage(image)
+		contextIgnores, fileWatchIgnores, err := s.ignoresForImage(image)
 		if err != nil {
-			return nil, fmt.Errorf("Reading dockerignore for %s: %v", image.configurationRef.RefFamiliarString(), err)
+			return nil, err
 		}
-
-		repoIgnores := s.repoIgnoresForImage(image)
-		fileWatchIgnores := []v1alpha1.IgnoreDef{}
-		if image.tiltfilePath != "" {
-			fileWatchIgnores = append(fileWatchIgnores, v1alpha1.IgnoreDef{BasePath: image.tiltfilePath})
-		}
-		fileWatchIgnores = append(
-			append(fileWatchIgnores, repoIgnores...),
-			model.DockerignoresToIgnores(dockerignores)...)
 
 		switch image.Type() {
 		case DockerBuild:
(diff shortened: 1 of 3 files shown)


internal/tiltfile/tiltfile_state.go
```diff
@@ -1692,6 +1678,53 @@ func toKubernetesApplyCmd(cmd model.Cmd) *v1alpha1.KubernetesApplyCmd {
...
+	if image.tiltfilePath != "" {
+		// TODO(milas): technically, this should be a FW ignore, not a context
+		// 	ignore, but historically it has been this way
+		contextIgnores = append(contextIgnores, v1alpha1.IgnoreDef{BasePath: image.tiltfilePath})
+	}
+	contextIgnores = append(contextIgnores, s.repoIgnoresForImage(image)...)
+	contextIgnores = append(contextIgnores, model.DockerignoresToIgnores(dockerignores)...)
+
+	for i := range contextIgnores {
+		fileWatchIgnores = append(fileWatchIgnores, *contextIgnores[i].DeepCopy())
```
nicks, line 1695: fyi - i think the k8s api convention is to only do a copy when you know you're going to modify the object

internal/tiltfile/tiltfile_state.go
```diff
@@ -1692,6 +1678,53 @@ func toKubernetesApplyCmd(cmd model.Cmd) *v1alpha1.KubernetesApplyCmd {
 	}
 }
 
+func (s *tiltfileState) ignoresForImage(image *dockerImage) (contextIgnores []v1alpha1.IgnoreDef, fileWatchIgnores []v1alpha1.IgnoreDef, err error) {
+	dockerignores, err := s.dockerignoresForImage(image)
+	if err != nil {
+		return nil, nil, fmt.Errorf("reading dockerignore for %s: %v", image.configurationRef.RefFamiliarString(), err)
+	}
+	if image.tiltfilePath != "" {
+		// TODO(milas): technically, this should be a FW ignore, not a context
```
nicks, line 1687: fwiw, i think we'd want to do this anyway, so that changes to the tiltfile don't accidentally break the docker cache

nicks (approved):
LGTM! nice fix!

internal/tiltfile/tiltfile_state.go
milas: Ahhh fair, let me remove that TODO

internal/tiltfile/tiltfile_state.go
milas: Yeah, this is probably extra defensive but given the lifetime of these items and how much they get passed around (+ the fact that we've got two intersecting sets of them originating from the same objects), I figured an extra copy was worth saving potential future debugging headache down the road

PR #5743, opened 2022-04-28
custom_build: add EXPECTED_{IMAGE,TAG,REGISTRY} env vars

### LINKED ISSUE #5433: `custom_build`: provide EXPECTED_REPOSITORY and EXPECTED_TAG addition to EXPECTED_REF

For `custom_build` I would like to have `EXPECTED_REPOSITORY` and `EXPECTED_TAG` in addition to `EXPECTED_REF`. Essentially `EXPECTED_REF` should equal `EXPECTED_REPOSITORY:EXPECTED_TAG`.

### Current Behavior

Currently for `custom_build`, Tilt injects the `EXPECTED_REF` environment variable which contains the full image name + tag reference, like `repository:tilt-build-1643758949`.

### Why Do You Want This?

This is a nice to have. It is common to allow configuring the image repository and tag separately and that is the way our build tooling is currently set up. This means we have to prefix every `custom_build` command with `EXPECTED_REPOSITORY="${EXPECTED_REF%%:*}" EXPECTED_TAG="${EXPECTED_REF#*:}"` to refer to them separately in our build tooling.

milas:
Currently, we pass `EXPECTED_REF`, which is the full image reference
including registry, path, and tag.

Some build scripts want to consume these independently, so in addition
to passing the full reference, this adds the individual components as
well.

There's a slight behavior change here in that previously, the registry
was passed as `REGISTRY_HOST` only in the case that
`outputs_image_ref_to` mode was being used. Now, it's _always_ passed
(if there is a local registry!) regardless of custom build mode along
with `EXPECTED_REGISTRY`, which is new and duplicative but consistent
with the other names. This should be harmless, however.

Closes #5433.

### DIFF
--- a/internal/build/custom_builder.go
+++ b/internal/build/custom_builder.go
@@ -8,6 +8,7 @@ import (
 	"os/exec"
 	"strings"
 
+	"github.com/docker/distribution/reference"
 	"github.com/opencontainers/go-digest"
 	"github.com/pkg/errors"
 	ktypes "k8s.io/apimachinery/pkg/types"
@@ -37,9 +38,9 @@ func (b *CustomBuilder) Build(ctx context.Context, refs container.RefSet,
 	imageMaps map[ktypes.NamespacedName]*v1alpha1.ImageMap) (container.TaggedRefs, error) {
 	expectedTag := spec.OutputTag
 	outputsImageRefTo := spec.OutputsImageRefTo
+	registryHost := refs.Registry().Host
 
 	var expectedBuildRefs container.TaggedRefs
-	var registryHost string
 	var err error
 
 	// There are 3 modes for determining the output tag.
@@ -49,10 +50,6 @@ func (b *CustomBuilder) Build(ctx context.Context, refs container.RefSet,
 
 		// Remove the output file, ignoring any errors.
 		_ = os.Remove(outputsImageRefTo)
-
-		// Inform the user script about the registry host
-		registryHost = refs.Registry().Host
-
 	} else if expectedTag != "" {
 		// If the tag is coming from the user script, we expect that the user script
 		// also doesn't know about the local registry. So we have to strip off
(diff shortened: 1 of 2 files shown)


internal/build/custom_builder.go
```diff
@@ -83,10 +80,18 @@ func (b *CustomBuilder) Build(ctx context.Context, refs container.RefSet,
 	if expectedBuildResult != nil {
 		extraEnvVars = append(extraEnvVars,
 			fmt.Sprintf("EXPECTED_REF=%s", container.FamiliarString(expectedBuildResult)))
+		extraEnvVars = append(extraEnvVars,
+			fmt.Sprintf("EXPECTED_IMAGE=%s", reference.Path(expectedBuildResult)))
```
milas, line 84: Is `EXPECTED_IMAGE` misleading?

Technically, in `distribution/reference` parlance, this is the "path", but I think `EXPECTED_PATH` is extremely vague/confusing.

I almost went with `EXPECTED_IMAGE_NAME` but wasn't sure if that was just redundantly & long 🤷

nicksieger: I think `EXPECTED_IMAGE` is fine. Is the path always part of the image name though? I'm thinking of the ECR case (https://docs.tilt.dev/personal_registry.html#elastic-container-registrys-repository-dance).

nicksieger (approved):

landism (approved):

internal/build/custom_builder_test.go
```diff
@@ -223,6 +223,33 @@ func TestCustomBuildImageDep(t *testing.T) {
...
 
+func TestEnvVars(t *testing.T) {
+	if runtime.GOOS == "windows" {
+		t.Skip("no sh on windows")
+	}
+
+	expectedVars := map[string]string{
+		"EXPECTED_REF":      "localhost:1234/foo_bar:tilt-build-1551202573",
+		"EXPECTED_REGISTRY": "localhost:1234",
+		"EXPECTED_IMAGE":    "foo_bar",
```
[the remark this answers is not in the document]
milas: Added in [4219735](https://github.com/tilt-dev/tilt/pull/5743/commits/42197353556b6e951bf713f5caa8518b822d0f98)

internal/build/custom_builder.go
milas: Thanks for all the research & thoughts! Will keep it as `EXPECTED_IMAGE` - I'll make sure when adding to custom build docs that I give an examples of values for each of the env vars, which should help

Issue #5751, opened 2022-05-01
Tilt v0.28.0 with Rancher Desktop is trying to push images to remote repo


iatzmon:
### Expected Behavior
Tilt should detect Rancher Desktop (using dockerd) and should skip pushing images to remote repo, as described in https://github.com/tilt-dev/tilt/issues/5367

### Current Behavior
Tilt is trying to push built images to docker.io

### Steps to Reproduce
Build and deploy an image to Rancher Desktop

#### `tilt doctor` Output
From version v0.28:
```
Tilt: v0.28.0, built 2022-04-29
System: darwin-arm64
---
Docker
- Host: unix:///var/run/docker.sock
- Server Version: 20.10.11
- API Version: 1.41
- Builder: 2
- Compose Version: v1.29.2 (build 5becea4c)
---
Kubernetes
- Env: rancher-desktop
- Context: rancher-desktop
- Cluster Name: rancher-desktop
- Namespace: default
- Container Runtime: docker
- Version: v1.20.10+k3s1
- Cluster Local Registry: none
---
Thanks for seeing the Tilt Doctor!
Please send the info above when filing bug reports. 💗

The info below helps us understand how you're using Tilt so we can improve,
but is not required to ask for help.
---
Analytics Settings
--> (These results reflect your personal opt in/out status and may be overridden by an `analytics_settings` call in your Tiltfile)
- User Mode: opt-in
- Machine: 4d2d8601baacf43ca72de3d91185cc6b
- Repo: 7Lb63VaGpdiANTKM1WL2GA==
```

From version 0.27.3 (which works properly):
```
Tilt: v0.27.3, built 2022-04-22
System: darwin-arm64
---
Docker
- Host: [default]
- Server Version: 20.10.11
- API Version: 1.41
- Builder: 2
- Compose Version: v1.29.2 (build 5becea4c)
---
Kubernetes
- Env: rancher-desktop
- Context: rancher-desktop
- Cluster Name: rancher-desktop
- Namespace: default
- Container Runtime: docker
- Version: v1.20.10+k3s1
- Cluster Local Registry: none
---
Thanks for seeing the Tilt Doctor!
Please send the info above when filing bug reports. 💗

The info below helps us understand how you're using Tilt so we can improve,
but is not required to ask for help.
---
Analytics Settings
--> (These results reflect your personal opt in/out status and may be overridden by an `analytics_settings` call in your Tiltfile)
- User Mode: opt-in
[...]

PR #5752, opened 2022-05-01
docker: whitelist default host strings

nicks:
Hello @milas, @nicksieger,

Please review the following commits I made in branch nicks/rancher-desktop:

54602f1fe5be7eba28966b82cec1e462e04e83bd (2022-05-01 12:58:59 -0400)
docker: whitelist default host strings
fixes https://github.com/tilt-dev/tilt/issues/5751


nicks:
sometimes there can be a mismatch between what DaemonHost() reports and what the CLI parsing library returns. This is a bad enough regression that we should probably do a bugfix release for it.

PR #5582, opened 2022-03-10
web: fix the Mode tooltip, and make sure the tooltip icons don't shrink

nicks:
Hello @lizzthabet,

Please review the following commits I made in branch nicks/tooltip2:

4df5549b5b0be72ca8024f0d1b26e414f8fca8e7 (2022-03-10 15:01:41 -0500)
web: fix the Mode tooltip, and make sure the tooltip icons don't shrink

Code review reminders, by giving a LGTM you attest that:

* Commits are adequately tested
* Code is easy to understand and conforms to style guides
* Incomplete code is marked with TODOs
* Code is suitably instrumented with logging and metrics

PR #5757, opened 2022-05-02
web: fix table view trigger mode button

### LINKED ISSUE #5754: Toggle button for Trigger mode not working
### Expected Behavior
When I click on Toggle Auto (or Toggle Manual) it should switch.

### Current Behavior
Nothing is happening. In order to change the trigger mode I have to change it on the Tiltfile. Even so the change isn't showed on the UI.

### Steps to Reproduce
1. Start a resource on Tilt
2. Go to UI Page
3. Table page (http://localhost:10350/overview)
4. Click on Switch button (Mode)

#### `tilt doctor` Output

```
$ tilt doctor
Tilt: v0.28.1, built 2022-05-01
System: linux-amd64
---
Docker
- Server Version: 20.10.12
- API Version: 1.41
- Builder: 2
---
Kubernetes
- Env: kind
- Context: kind-kind
- Cluster Name: kind-kind
- Namespace: default
- Version: v1.21.1
- Cluster Local Registry: {Host:localhost:5005 hostFromCluster: SingleName:}
---

---
Analytics Settings
- User Mode: opt-in
- Machine: c85df380dc70a009beacb572ba0f612d
- Repo: WCAKS1KGD+Ba+i4rUIYVmA==

```Tilt: v0.28.1, built 2022-05-01
System: linux-amd64
---
Docker
- Server Version: 20.10.12
- API Version: 1.41
- Builder: 2
---
Kubernetes
- Env: kind
- Context: kind-kind
- Cluster Name: kind-kind
- Namespace: default
- Version: v1.21.1
- Cluster Local Registry: {Host:localhost:5005 hostFromCluster: SingleName:}
---

---
Analytics Settings
- User Mode: opt-in
- Machine: c85df380dc70a009beacb572ba0f612d
- Repo: WCAKS1KGD+Ba+i4rUIYVmA==


#### About Your Use Case
Switching contexts for live update development

landism:
fixes #5754

This was broken by #5582 - the addition of `id` to the "mode" column definition caused the trigger mode to be written to RowValues as "mode" rather than "triggerMode", and this somehow didn't cause a compilation failure. The react-table docs around id vs accessor are not very clarifying, and I'm not sure it's worth getting to the bottom of that at the moment.

I'm happy to hear if there's a more idiomatic way to write the test.

### DIFF
--- a/web/src/OverviewTable.test.tsx
+++ b/web/src/OverviewTable.test.tsx
@@ -23,6 +23,7 @@ import OverviewTable, {
   TableGroupedByLabels,
 } from "./OverviewTable"
 import { Name, RowValues, SelectionCheckbox } from "./OverviewTableColumns"
+import { ToggleTriggerModeTooltip } from "./OverviewTableTriggerModeToggle"
 import {
   DEFAULT_GROUP_STATE,
   GroupsState,
@@ -43,7 +44,7 @@ import {
   oneUIButton,
   TestDataView,
 } from "./testdata"
-import { RuntimeStatus, UpdateStatus } from "./types"
+import { RuntimeStatus, TriggerMode, UpdateStatus } from "./types"
 
 // Helpers
 const tableViewWithSettings = ({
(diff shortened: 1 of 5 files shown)


nicks (approved):
we should probably also have a test that clicks the button and makes sure the right requests are sent?

PR #5766, opened 2022-05-04
registry: use API types everywhere

milas:
Eliminate the `container.Registry` helper and rely on the API type
of `v1alpha1.RegistryHosting`, which 1:1 maps to the local registry
spec.

A couple helpers methods are now functions, but the biggest change
is probably around default value semantics. For more consistency with
how the API types are actually used, the registry is `nil` rather
than empty.

The naming is also a bit more explicit for the changed areas, so that
"cluster ref" is now called "container runtime ref" to align with the
spec and avoid confusion between "cluster network ref", which is not
used by Tilt (aside from passing it along as metadata in the `Cluster`
API). I left a couple TODOs here for additional mechanical renames we
can do here but omitted them from this commit to keep it from getting
too unwieldy.


pkg/apis/core/v1alpha1/registry_types_test.go
```diff
@@ -0,0 +1,62 @@
+package v1alpha1_test
```
milas, line 1: I moved these tests from `internal/container/default_registry_test.go`

These are the first _explicit_ tests we have for `Validate()` methods on API types (😅). I think it makes sense for them to live here along with the types, but calling it out because there's currently no other `_test.go` files here and the API types package is kind of special with all the codgen stuff

internal/container/registry.go
```diff
@@ -0,0 +1,123 @@
...
+	}
+
+	return newRef, nil
+}
+
+// ReplaceRegistryForLocalRef returns a new reference using the target local
+// registry (as seen from the user).
+func ReplaceRegistryForLocalRef(rs RefSelector, reg *v1alpha1.RegistryHosting) (reference.Named, error) {
+	var host, singleName string
+	if reg != nil {
```
nicks, line 104: should we treat reg == nil and reg.Host == "" the same?

internal/container/reference.go
```diff
@@ -67,11 +68,14 @@ func (rs RefSet) WithoutRegistry() RefSet {
 	return MustSimpleRefSet(rs.ConfigurationRef)
 }
 
-func (rs RefSet) Registry() Registry {
-	return rs.registry
+func (rs RefSet) Registry() *v1alpha1.RegistryHosting {
+	if rs.registry == nil {
+		return nil
+	}
+	return rs.registry.DeepCopy()
```
nicks, line 75: TBH i'm surprised we're doing a copy so often. was mutating registries a problem, or is this just being defensive?

(one of my big complains about go is the lack of mutability protection, but i do like the Kubernetes convention of "always copy before mutating")

nicks (approved):

nicksieger (approved):

internal/container/reference.go
milas: Well, previously it was always used as a value type in conjunction with the `Registry::Empty()` method, so we weren't handing out mutable references here

FWIW, in cases like this, `DeepCopy()` allocates and then does `*out = *in`, so minus the unavoidable heap usage, it shouldn't be worse

internal/container/registry.go
milas: Good call, changed in a0c6aa5f2

Issue #5745, opened 2022-04-28
`docker_build` is not respecting `.dockerignore`


garyng:
### Expected Behavior

`docker_build` should respect `.dockerignore`.

### Current Behavior

`docker_build` seems to ignore `.dockerignore` causing very long build if there are a lot of ignored files (eg: `node_modules`). I did some search on GitHub and found this: https://github.com/tilt-dev/tilt/pull/5416, which might be relevant?

### Steps to Reproduce

1. Clone https://github.com/tilt-dev/tilt-avatars
1. Run `tilt up`
1. Observe the time for `[background] read source files`, in this case it was `[background] read source files 61.67kB [done: 2.789s]`
```
STEP 1/3 — Building Dockerfile: [tilt-avatar-web]
Building Dockerfile for platform linux/amd64:
  FROM node:16-bullseye-slim
  
  ENV YARN_CACHE_FOLDER=/cache/yarn
  
  WORKDIR /app
  
  # install dependencies
  COPY web/package.json web/yarn.lock ./
  RUN --mount=type=cache,target=/cache/yarn \
    yarn install
  
  # copy app source
  COPY web/ ./
  
  CMD ["yarn", "dev"]

     Building image
     [stage-0 1/5] FROM docker.io/library/node:16-bullseye-slim@sha256:9ae59bd83f46c6e6b83c0552fdfb0da00b3df50542b80ec818c658438cc61af1 68.68MB / 68.68MB [done: 58.791s]
     [background] read source files 61.67kB [done: 2.789s]
     [stage-0 2/5] WORKDIR /app [done: 10.288s]
     [stage-0 3/5] COPY web/package.json web/yarn.lock ./ [done: 5.699s]
     [stage-0 4/5] RUN --mount=type=cache,target=/cache/yarn   yarn install
       → yarn install v1.22.18
       → [1/4] Resolving packages...
       → [2/4] Fetching packages...
       → [3/4] Linking dependencies...
       → [4/4] Building fresh packages...
       → Done in 27.73s.
     [stage-0 4/5] RUN --mount=type=cache,target=/cache/yarn   yarn install [done: 38.546s]
     [stage-0 5/5] COPY web/ ./ [done: 4.13s]
     exporting to image [done: 13.678s]
```

1. Run `npm install` inside `web`
[...]

PR #5768, opened 2022-05-04
build: don't stat directories that we don't plan to upload

nicks:
Hello @milas, @landism,

Please review the following commits I made in branch nicks/walk:

c1a8373b5f4db3b324ad66cf690280b2f4884f06 (2022-05-04 19:06:56 -0400)
build: don't stat directories that we don't plan to upload
creates some temporary patches on buildkit until we can upstream them

fixes https://github.com/tilt-dev/tilt/issues/5745


nicks:
I'm hoping the upstream patches will be accepted, because they're pretty minimal/non-invasive (which you can see in the vendor diff). But figure we'll have to patch anyway because we're on an older version of buildkit.

milas:
> But figure we'll have to patch anyway because we're on an older version of buildkit.

FWIW if the fork(s) become onerous to maintain, we could probably get _most_ of the way there by using the existing `Excludes` field on `SyncedDir`. We'd have to populate from `v1alpha1.IgnoreDef`, which would require some care to avoid it kicking right back down the slow path, so it's not ideal for a variety of reasons

milas (approved):

nicks:
ya, it's tricky because ExcludePatterns MUST be relative to the context, so would need to do a fair bit of re-formatting

PR #5778, opened 2022-05-05
dockercomposeservice: create logs object

nicks:
Hello @milas,

d3992e3f912468db69504e17c3fd4419b805b552 (2022-05-05 17:16:43 -0400)
dockercomposeservice: create logs object

### DIFF
--- a/internal/controllers/core/dockercomposeservice/logs.go
+++ b/internal/controllers/core/dockercomposeservice/logs.go
@@ -1,0 +1,128 @@
+package dockercomposeservice
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+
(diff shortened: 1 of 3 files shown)


internal/controllers/core/dockercomposeservice/logs.go
```diff
@@ -0,0 +1,128 @@
...
+//
+// I think the better way to model this would be to more clearly separate "spinning up a service"
+// from "watching a service", so we'd actually have:
+//
+// 1) A service object with a Spec that tells us how to create the service and
+//    a Status that tells us whether the creation succeeded.
+// 2) A monitor object with a status that contains both the container state and the log state.
+//
+// But moving to this system will be easier once everything is in the API server.
+func (r *Reconciler) manageOwnedLogStream(ctx context.Context, nn types.NamespacedName, obj *v1alpha1.DockerComposeService) (reconcile.Result, error) {
```
nicks, line 34: most of this logic is loosely adapted from KubernetesApply. The important bit is the comment on this function...not sure how strongly to go about changing this now, but seems like the better way to model this.

milas (approved):

internal/controllers/core/dockercomposeservice/logs.go
milas: Seems reasonable to me to hold off on for now - the split is very useful/important in K8s where there's multiple different ways to deploy but we want to share the monitoring logic across them. I can't imagine much of a demand for e.g. `dc_custom_deploy` though, so having them be a bit tighter coupled seems safe enough

Issue #5729, opened 2022-04-25
Cluster health check failure can get stuck


milas:
### Expected Behavior
* If the cluster becomes unhealthy and then healthy again, Tilt reflects that both in the cluster pop-up in the UI and by "unholding" any resources waiting for cluster

### Current Behavior
* Possible for health check to get stuck in a failing state

### Steps to Reproduce
This is a recent feature and we've only had this [reported once via Slack](https://kubernetes.slack.com/archives/CESBL84MV/p1650646820306519), but the error was showing an error on the `/livez` check, and the user reported that request was succeeding via `curl` at that point.

![Screen Shot 2022-04-22 at 9 47 08 AM](https://user-images.githubusercontent.com/841263/165095767-be52a00b-876a-4292-ac3a-01d1412005ea.png)

They'd mentioned getting into the state after having put their laptop to sleep for the day and returning the next morning.

andymartin-sch:
a few more developers of ours saw this recently

milas:
@andymartin-sch Thanks for the extra reports - agreed this is not the experience we want here; I'm hoping to include at least some form of remediation in our release today.

In the cases you've seen, has the error shown in the Tilt UI been similar to that in the issue above? If so, do you know if anyone tried manually accessing the endpoint (e.g. `curl https://..../livez`) and whether that was successful?

andymartin-sch:
ah, one developer just said:
> When I hit this, I went to that endpoint in my browser and it returned “ok”

milas:
A couple improvements/fixes went into v0.29.0 (released May 6) - please let me know if you still see the issue after upgrading!

PR #5781, opened 2022-05-06
cluster: refactor health monitoring to avoid getting stuck

milas:
I'm not 100% sure this is the root cause of the cluster health
monitoring getting stuck, but it's definitely not totally correct
in its current state.

The `ConnectionManager` is a fancy wrapper over a `sync.Map` and
expects the `connection` objects to be immutable — we replace them
entirely on change. Within the context of the reconciliation loop,
this is good. However, the health status monitor also needs to store
the result of the health checks and was doing so on the same object,
which resulted in a race condition.

Now, the cluster health state is stored independently and used within
reconciliation. The main `Reconcile()` loop still replaces objects
in their entirety in the `ConnectionManager`, but as its the only writer
now, there's no potential for stale data/races.

See #5729.

### DIFF
--- a/internal/controllers/core/cluster/monitor.go
+++ b/internal/controllers/core/cluster/monitor.go
@@ -91,6 +91,12 @@ func (c *clusterHealthMonitor) run(ctx context.Context, clusterNN types.Namespac
 	defer ticker.Stop()
 	for {
 		err := doKubernetesHealthCheck(ctx, conn.k8sClient)
+		if ctx.Err() != nil {
+			// if the context as canceled while the health check was running,
+			// it'll be the cause of the error, which isn't actually a health
+			// check failure
+			return
+		}
 		if err != nil {
 			c.UpdateStatus(clusterNN, err.Error())
 		} else {
--- a/internal/controllers/core/cluster/monitor.go
+++ b/internal/controllers/core/cluster/monitor.go
@@ -47,9 +47,16 @@ func (c *clusterHealthMonitor) GetStatus(clusterNN types.NamespacedName) string 
 	return c.monitors[clusterNN].error
 }
 
-func (c *clusterHealthMonitor) UpdateStatus(clusterNN types.NamespacedName, error string) {
+func (c *clusterHealthMonitor) UpdateStatus(ctx context.Context, clusterNN types.NamespacedName, error string) {
 	c.mu.Lock()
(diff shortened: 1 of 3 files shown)


internal/controllers/core/cluster/monitor.go
```diff
@@ -3,33 +3,98 @@ package cluster
...
+	defer c.mu.Unlock()
+
+	c.cleanup(clusterNN)
+	delete(c.monitors, clusterNN)
+}
+
+func (c *clusterHealthMonitor) cleanup(clusterNN types.NamespacedName) {
+	m := c.monitors[clusterNN]
+	if m.cancel != nil {
+		m.cancel()
```
nicks, line 75: i think you have a slight race condition here -- `cancel()` assumes that it stops the healthcheck loop immediately, but in your code, the loop can keep going and recording results.

there's two patterns that we usually do to deal with this:
1) Add some sort of done channel so that your healthcheck loop can tell everyone when it's done. (i.e.,have the controller code synchronize before starting a new loop)
2) Change UpdateStatus() to do nothing if ctx.Err() != nil (i.e., have the healthcheck loop synchronize before sending data out to the controller)

nicks: alternatively, the healthcheck loop could have a pointer to its own monitor and only update that one; creating a new loop creates a new monitor

milas: Nice catch - went with (2) in [dbadc7f](https://github.com/tilt-dev/tilt/pull/5781/commits/dbadc7fa5d37b29573764f3f2322a0703d1a0023) as it was the most straightforward for now!

internal/controllers/core/cluster/monitor.go
```diff
@@ -3,33 +3,104 @@ package cluster
...
 	}
 
-	ticker := r.clock.NewTicker(clientHealthPollInterval)
+	ticker := c.clock.NewTicker(clientHealthPollInterval)
 	defer ticker.Stop()
 	for {
-		lastErr := conn.statusError
-
 		err := doKubernetesHealthCheck(ctx, conn.k8sClient)
+		if ctx.Err() != nil {
```
nicks, line 94: i would probably do this in UpdateStatus() (when you have the mutex lock)

nicks (approved):

PR #5783, opened 2022-05-06
web: add download snapshot dialog behind feature flag

lizzthabet:
* Add a new snapshot dialog for downloading a snapshot file directly, only visible behind a feature flag; this dialog is styled like the other menu item dialogs.
* Add a new feature flag named `offline_snapshot_creation`
* Add a dialog anchor to the HUD state for the float dialog to display. (Because snapshot creation grabs data from the HUD state, I opted for this change, rather than a larger refactor of the `SnapshotProvider` state to move data there so that the float dialog could live in the `GlobalNav`.)

Thanks to @landism for doing the majority of this PR! ~Leaving this in draft mode until I have a couple tests added, but otherwise nothing else should change.~

Screen Shot 2022-05-06 at 12 59 14 PM

Closes [ch13474](https://app.shortcut.com/windmill/story/13474/as-user-download-snapshot-json-via-tilt-web).


web/src/ShareSnapshotModal.tsx
```diff
@@ -1,22 +1,50 @@
...
   tiltCloudTeamID: string | null
   isOpen: boolean
   highlightedLines: number | null
+  dialogAnchor: HTMLElement | null
 }
 
-export default class ShareSnapshotModal extends PureComponent<props> {
+// TODO (lizz): When offline snapshots are the default method for snapshots
+// generation, cloud snapshot code can be refactored
+export default function ShareSnapshotModal(props: ShareSnapshotModalProps) {
```
milas, line 31: Pure bikeshed, but `{Upload,Download}SnapshotModal` are kinda confusing in this context, maybe `LocalSnapshotModal` & `CloudSnapshotModal`?

milas (approved):

web/src/ShareSnapshotModal.tsx
landism: ha, yeah. I named DownloadSnapshotModal without thinking and thought it was cutely chaotic (in that the names make sense in isolation but pretty misleading in combination) but had meant to change them before merging.

landism:
I haven't figured out why, but this breaks online snapshot creation for me:
image

(and apparently this was true in my original branch as well).

I'll update if I figure it out before EOD.

landism:
> I'll update if I figure it out before EOD.

pushed a [fix](https://github.com/tilt-dev/tilt/pull/5783/commits/cba7e70ad9f872209e6f8d8d37c0349cfb6209d4)!

web/src/ShareSnapshotModal.tsx
```diff
@@ -215,3 +243,81 @@ export default class ShareSnapshotModal extends PureComponent<props> {
...
+  background-color: ${Color.offWhite};
+  padding: 0 5px;
+  white-space: nowrap;
+`
+
+function downloadSnapshot(snapshot: Proto.webviewSnapshot) {
+  const data = new Blob([JSON.stringify(snapshot)], {
+    type: "application/json",
+  })
+  saveAs(data, "snapshot.json")
```
nicks, line 280: small nit: it might be good to put the current datetime in the file name, similar to how OS screenshotting tools do it, so you don't have snapshots overwriting each other

web/src/ShareSnapshotModal.test.tsx
```diff
@@ -22,98 +29,166 @@ describe("ShareSnapshotModal", () => {
...
+            highlightedLines={null}
+            dialogAnchor={null}
+            getSnapshot={fakeGetSnapshot}
+          />
+        )
+        .toJSON()
 
-    expect(tree).toMatchSnapshot()
+      expect(tree).toMatchSnapshot()
+    })
```
lizzthabet, line 136: these are all whitespace changes. all the tests are the same.

web/src/ShareSnapshotModal.tsx
lizzthabet: done!

web/package.json
```diff
@@ -71,6 +73,7 @@
     "@typescript-eslint/eslint-plugin": "^5.3.1",
     "@typescript-eslint/parser": "^5.3.1",
     "babel-loader": "8.1.0",
+    "blob-polyfill": "^7.0.20220408",
```
lizzthabet, line 76: this was needed to read and test the snapshot blob.

nicks (approved):

PR #5786, opened 2022-05-06
cluster: write a frozen kubeconfig when we connect

nicks:
Hello @milas, @lizzthabet,

Please review the following commits I made in branch nicks/minify2:

01f0473accbe5a95cc3514db257935a7ba20b16f (2022-05-06 15:20:37 -0400)
cluster: write a frozen kubeconfig when we connect

### DIFF
--- a/internal/controllers/core/cluster/reconciler.go
+++ b/internal/controllers/core/cluster/reconciler.go
@@ -46,13 +46,14 @@ const (
 )
 
 type Reconciler struct {
-	globalCtx   context.Context
-	ctrlClient  ctrlclient.Client
-	store       store.RStore
-	requeuer    *indexer.Requeuer
-	clock       clockwork.Clock
-	connManager *ConnectionManager
-	base        xdg.Base
+	globalCtx     context.Context
+	ctrlClient    ctrlclient.Client
+	store         store.RStore
+	requeuer      *indexer.Requeuer
+	clock         clockwork.Clock
+	connManager   *ConnectionManager
+	base          xdg.Base
+	apiServerName model.APIServerName
 
 	localDockerEnv      docker.LocalEnv
 	dockerClientFactory DockerClientFactory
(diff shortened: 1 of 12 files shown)


nicks:
a lot of this PR is just getting the right API objects to the right places

internal/controllers/core/cluster/reconciler.go
```diff
@@ -357,6 +377,50 @@ func (r *Reconciler) populateK8sMetadata(ctx context.Context, clusterNN types.Na
...
+	}
+
+	obj, err := latest.Scheme.ConvertToVersion(config, latest.ExternalVersion)
+	if err != nil {
+		logger.Get(ctx).Warnf("Converting Kubernetes config: %v", err)
+		return ""
+	}
+
+	printer := printers.YAMLPrinter{}
+	path, err := r.base.ConfigFile(filepath.Join("cluster", nn.Name))
```
milas, line 401: Nit: `RuntimeFile` seems more appropriate for this

internal/k8s/env.go
```diff
@@ -12,33 +10,52 @@ import (
...
+		context, ok := config.Contexts[config.CurrentContext]
+		if ok {
+			context.Namespace = string(namespaceOverride)
 		}
 	}
 
-	return &config, nil
+	err = clientcmd.ConfirmUsable(config, config.CurrentContext)
+	if err != nil {
+		return APIConfigOrError{Error: errors.Wrap(err, "Overriding Kubernetes context")}
```
milas, line 48: Minor: this is always happening now, so a more generic message like "Loading Kubernetes config" might be more applicable

milas (approved):
Woooo

internal/controllers/core/cluster/reconciler.go
nicks: done

internal/k8s/env.go
nicks: done

PR #5796, opened 2022-05-10
script: add local release build `Makefile` task

milas:
Our builds have gotten more complicated recently with the new
requirement of CGO on _all_ platforms (due to Tree-sitter for
the LSP) as well as the embedded assets mode.

As a result, I added an alternate CI flow to build (but not
publish) binaries to ensure things were working as expected.
However, this requires having sufficient CircleCI access, has
a confusing UX, and complicates the `.circleci/config.yml` file
pretty significantly.

This commit eliminates that in favor of a `Makefile` task that
can be executed locally and uses Toast to do a containerized build
that will include embedde JS assets and produce cross-compiled
binaries for all supported OS/arch combinations using the same
underlying builder image as CI.

(Users who wish to build Tilt from source can also take advantage
of this to avoid having a complex build setup locally, though we
will continue to recommend using our binary releases.)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -196,3 +196,6 @@ update-codegen-starlark:
 
 update-codegen-ts:
 	toast proto-ts
+
+release-build:
+	toast -f build.toast.yml
--- a/scripts/release.Dockerfile
+++ b/scripts/release.Dockerfile
@@ -51,6 +51,3 @@ RUN git clone https://github.com/Homebrew/brew /home/linuxbrew/.linuxbrew
 ENV PATH=/home/linuxbrew/.linuxbrew/bin:$PATH
 
 RUN mkdir -p ~/.windmill
-
-ENTRYPOINT ["goreleaser"]
-CMD ["-h"]
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -150,20 +150,11 @@ jobs:
       - slack/notify-on-failure:
           only_for_branches: master
 
-  release-verify:
-    resource_class: medium+
-    docker:
-      - image: gcr.io/windmill-public-containers/tilt-releaser@sha256:a17867de17e04c8a8ea2150f9b48f9fa8588617cebe0bb36dc8521d1546414a3
-    steps:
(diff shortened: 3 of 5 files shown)


scripts/release.Dockerfile
```diff
@@ -51,6 +51,3 @@ RUN git clone https://github.com/Homebrew/brew /home/linuxbrew/.linuxbrew
 ENV PATH=/home/linuxbrew/.linuxbrew/bin:$PATH
 
 RUN mkdir -p ~/.windmill
-
-ENTRYPOINT ["goreleaser"]
```
nicks, line 55: what's with this change? i don't think we want to use the upstream entrypoint

Makefile
```diff
@@ -196,3 +196,6 @@ update-codegen-starlark:
 
 update-codegen-ts:
 	toast proto-ts
+
+release-build:
+	toast -f build.toast.yml
```
nicks, line 201: fwiw, i've been wondering if we should move to using Earthly rather than Toast for these kinds of containerized tasks:
https://github.com/earthly/earthly

i've found Toast to be pretty finicky and has a lot of non-intuitive semantics

nicks (approved):

scripts/release.Dockerfile
milas: This was a fun adventure:
* Toast does not have support for overriding entrypoint (it expects to be dropped into some sort of shell)
* You CAN pass arbitrary flags, so I tried to hack around with `--entrypoint=""`
  * Toast uses `--init` flag with Docker to get a Tini init to avoid orphaned processes since not all shells handle this
  * Tini complains it can't exec `""` 😬 

The upstream `ENTRYPOINT` is `bash` and there's no `CMD`.

It's irrelevant to CircleCI since they override the entrypoint (presumably due to situations like this, which I've had to battle before with GitLab CI 😓)

nicks: ya, i think this just speaks to my earthly comment haha

PR #5798, opened 2022-05-10
web: update document HTML title based on snapshot status

lizzthabet:
This PR fixes the misleading `... disconnected | Tilt` document title that displays for snapshots when there's no socket connection. Instead, snapshots will have `Snapshot:` prepended to the usual (connected) resource status title. 

This PR also moves the document metadata calculations into a separate helper function, so it's easier to test.

You can test this PR by creating and viewing snapshots using the CLI or web UI with `offline_snapshot_creation` flag enabled.

[Closes ch13500](https://app.shortcut.com/windmill/story/13500/fix-snapshots-json-disconnected-html-title)

### DIFF
--- a/web/src/ResourceStatusSummary.test.tsx
+++ b/web/src/ResourceStatusSummary.test.tsx
@@ -1,7 +1,11 @@
 import { render, screen } from "@testing-library/react"
 import React from "react"
 import { MemoryRouter } from "react-router"
-import { ResourceGroupStatus, StatusCounts } from "./ResourceStatusSummary"
+import {
+  getDocumentTitle,
+  ResourceGroupStatus,
+  StatusCounts,
+} from "./ResourceStatusSummary"
 
 function expectStatusCounts(expected: { label: string; counts: number[] }[]) {
   const actual = expected.map(({ label, counts }) => {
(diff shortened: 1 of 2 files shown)


web/src/ResourceStatusSummary.test.tsx
```diff
@@ -102,3 +142,85 @@ it("does NOT link to warning and unhealthy resources when `linkToLogFilters` is
...
+    ],
+    [
+      "all resources are disabled",
+      {
+        counts: testCountsAllDisabled,
+        isSnapshot: false,
+        isSocketConnected: true,
+      },
+      "… 0/0 ┊ Tilt",
+      "gray",
```
lizzthabet, line 209: outside the scope of this pr, but i'll mention in case anyone has better suggestions! the "all resources are disabled" document title could probably be improved, so it doesn't seem like this is a "pending" / "disconnected" state.

milas: Perhaps `⃠`? (`&nbsp;&#x20E0;`)

Alternatively, a gray `✔︎` with `0/0` seems reasonable?

milas (approved):

web/src/ResourceStatusSummary.test.tsx
lizzthabet: good suggestion! i like `✔︎ 0/0`.

PR #5799, opened 2022-05-10
web: don't display the account menu item when offline snapshots are enabled

lizzthabet:
This PR hides the account menu item and dialog when offline snapshot creation is enabled. Since people won't need an account to create an offline snapshot, account information is no longer relevant.

[Closes 13514](https://app.shortcut.com/windmill/story/13514)

### DIFF
--- a/web/src/AccountMenu.tsx
+++ b/web/src/AccountMenu.tsx
@@ -4,6 +4,8 @@ import styled from "styled-components"
 import { ReactComponent as TiltCloudLogoSvg } from "./assets/svg/logo-Tilt-Cloud.svg"
 import ButtonInput from "./ButtonInput"
 import ButtonLink from "./ButtonLink"
+import { Flag, useFeatures } from "./feature"
+import FloatDialog from "./FloatDialog"
 import { AnimDuration, Color, Font, FontSize, SizeUnit } from "./style-helpers"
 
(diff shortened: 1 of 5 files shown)


web/src/AccountMenu.tsx
```diff
@@ -166,3 +167,33 @@ export function AccountMenuHeader(props: AccountMenuProps) {
...
+  onClose: () => void
+} & AccountMenuProps
+
+export function AccountMenuDialog(props: AccountMenuDialogProps) {
+  const { open, anchorEl, onClose, ...accountMenuProps } = props
+  const features = useFeatures()
+
+  const hideAccountDialog = features.isEnabled(Flag.OfflineSnapshotCreation)
+  if (hideAccountDialog) {
+    return null
```
nicks, line 183: this might be a silly question...but shouldn't it be impossible to hit this codepath? because disabling the dialog happens further upstream.

nicks (approved):

web/src/AccountMenu.tsx
lizzthabet: not a silly question! i'm being extra-defensive in this code path and mostly added this case due to my local testing. you can open the account dialog, enable the feature flag, and the dialog will stick around even though the menu item is gone.

PR #5801, opened 2022-05-10
circleci: upgrade kind

nicks:
Hello @nicks,

Please review the following commits I made in branch nicks/kinds:

17aa68ed810d54fb98057d1c63a79e54b0b40a45 (2022-05-10 19:06:00 -0400)
circleci: upgrade kind


nicks:
hmmm...i think the failure here is something about the way that the test creates a new k8s account w/ rbac...

nicks:
ah, i see the problem -
https://kubernetes.io/docs/concepts/configuration/secret/#service-account-token-secrets
https://github.com/kubernetes/kubernetes/pull/108309

we're getting hit by the above change in k8s 1.24, which changes how service account auth works by default. Need to read up on what's changing and why...

milas (approved):
🥳

PR #5802, opened 2022-05-11
snapshot: include creation timestamp

milas:
Historically, we've relied on the Tilt Cloud server timestamp for
snapshots. With the new offline flow, this isn't available, so a
new field is added with the create time.


nicks (approved):

lizzthabet (approved):
lgtm!

PR #5805, opened 2022-05-11
web: add banner with timestamp when viewing snapshot

lizzthabet:
This PR adds a small banner above the header bar that makes it clearer when a user is viewing a snapshot. The data flow / data access for snapshots is a little convoluted, so let me know if you have any better suggestions than the `AppController` > `HUD` > `SnapshotProvider` flow. I ended up rendering the `SnapshotBar` alongside the `HeaderBar` to avoid any layout shifts.

Screen Shot 2022-05-11 at 3 29 41 PM

You can test these changes by looking at the new Storybook story or by navigating to `snapshot/local` with any Tiltfile when running this branch.

[Closes ch13503](https://app.shortcut.com/windmill/story/13503).

### DIFF
--- a/web/src/ShareSnapshotModal.tsx
+++ b/web/src/ShareSnapshotModal.tsx
@@ -275,7 +275,7 @@ const CodeSnippet = styled.code`
 `
 
 function downloadSnapshot(snapshot: Proto.webviewSnapshot) {
-  const timestamp = moment().format("YYYY-MM-DD_HHmmss")
+  const timestamp = snapshot.createdAt ?? moment().format("YYYY-MM-DD_HHmmss")
   const data = new Blob([JSON.stringify(snapshot)], {
     type: "application/json",
   })
--- a/web/src/ShareSnapshotModal.tsx
+++ b/web/src/ShareSnapshotModal.tsx
@@ -274,13 +274,17 @@ const CodeSnippet = styled.code`
   white-space: nowrap;
 `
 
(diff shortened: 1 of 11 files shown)


web/src/ShareSnapshotModal.tsx
```diff
@@ -275,7 +275,7 @@ const CodeSnippet = styled.code`
 `
 
 function downloadSnapshot(snapshot: Proto.webviewSnapshot) {
-  const timestamp = moment().format("YYYY-MM-DD_HHmmss")
+  const timestamp = snapshot.createdAt ?? moment().format("YYYY-MM-DD_HHmmss")
```
milas, line 278: `snapshot.createdAt` is going to be a full ISO8601 string, so I think it should be passed _to_ `moment()` so we always format or we'll get filenames like `tilt-snapshot_2022-05-11T20:10:17.882742.json` instead of `tilt-snapshot_2022-05-11_201017.json`

e.g.
```javascript
const createdAt = snapshot.createdAt ? moment(snapshot.createdAt) : moment();
const timestamp = createdAt.format("YYYY-MM-DD_HHmmss")
```
(`moment()` is cranky and doesn't like `null` or `""`)

milas (approved):

web/src/ShareSnapshotModal.tsx
lizzthabet: ah, good point! i've fixed the formatting.

Issue #1105, opened 2019-02-03
Streamling getting started with `tilt new`


hausdorff:
Various projects (_e.g._, npm) have an `init` or `new` command that asks you a series of questions and then generates the relevant files (_e.g._, `package.json`) in response.

It would be very nice if something similar was available with `tilt`.

nicks:
Yes!! This is on the [roadmap](https://github.com/windmilleng/tilt/blob/master/ROADMAP.md#tilt-init-p1-c2). does that description sound right to you?

jazzdan:
FWIW it looks like @nicks just removed this from the ROADMAP. Was that in error or is this not a feature that will be added to Tilt?

lizzthabet:
we've launched a couple features that i think address this issue! when you `tilt up` in a directory without a `Tiltfile`, tilt will offer to generate a sample one for you to run. there's also [`tilt demo`](https://docs.tilt.dev/cli/tilt_demo.html), which allows you to run a sample tilt project and have tilt create a temporary k8s cluster for you, and a library of [`Tiltfile` snippets](https://docs.tilt.dev/snippets.html) that demonstrate common setup and tasks.

PR #5809, opened 2022-05-12
feature: mark the `disable_resources` flag as obsolete

milas:
This has been enabled by default for a few releases now, so time
for it to go!


lizzthabet (approved):
i also linked [#1105](https://github.com/tilt-dev/tilt.build/pull/1105) for the docs update.

PR #5810, opened 2022-05-12
web: fix snapshot bar date format

milas:
The formatted version for the filename was also being used for
display in the header. Switched it to use `lll` format, which
is friendlier/more readable.

### Before
Screen Shot 2022-05-12 at 1 05 06 PM

### After
Screen Shot 2022-05-12 at 1 05 12 PM


lizzthabet (approved):
thanks for this improvement!

PR #5823, opened 2022-05-17
docker: recognize docker desktop for linux socket location

### LINKED ISSUE #5812: Tilt fails to detect Docker Desktop for Linux and attempts to push images
### Expected Behavior
Tilt should realize I'm using a local Docker Desktop skip pushing the image. Such as
```
Skipping push: building on cluster's container runtime
```

### Current Behavior
Tilt attempts to push my image to docker.io and fails.
```
STEP 2/3 — Pushing pigeon:tilt-f7346993891b395b
     Pushing with Docker client
     Authenticating to image repo: docker.io
     Sending image data

Build Failed: docker push: pushing image "docker.io/library/pigeon": denied: requested access to the resource is denied
```

### Steps to Reproduce
While I can't provide steps to an exact repro, it starts off with.
1. Install [Docker Desktop for Linux](https://docs.docker.com/desktop/linux/install/) on Debian 11.
2. For good measure, stop Docker engine.
    ```
    sudo systemctl stop docker docker.socket containerd
    ```
3. Open Docker Desktop from the desktop environment and in the settings enable Kubernetes.
4. `kubectl config use-context docker-desktop`
5. `tilt up ...`.  The project I'm doing this on ends up failing 2 resources by attempting to push images.

### Context

This looks very similar to  #5034 on Windows, but I suspect the root cause could be different.  Having `DOCKER_HOST` unset does _not_ work around the problem for me, like the other issue mentions.

I do not encounter this problem when using Minikube instead of Docker Desktop for Linux.

#### `tilt doctor` Output
```
$ tilt doctor
Tilt: v0.29.0, built 2022-05-06
System: linux-amd64
---
Docker
- Host: unix:///home/jmay/.docker/desktop/docker.sock
- Server Version: 20.10.14
- API Version: 1.41
- Builder: 2
---
Kubernetes
- Env: docker-desktop
- Context: docker-desktop
- Namespace: default
- Container Runtime: docker
- Version: v1.24.0
---

---
Analytics Settings
- User Mode: opt-out
- Machine: cf920a31c2760bcfd0af8a1d35fb070f
- Repo: GMRpQPMQ7lumf8n4GPEFgg==
```

nicksieger:
Fixes #5812.

### DIFF
--- a/internal/docker/env.go
+++ b/internal/docker/env.go
@@ -300,7 +300,11 @@ func isDefaultHost(e Env) bool {
 			host == "npipe:////./pipe/docker_engine" ||
 
 			// https://github.com/moby/moby/blob/master/client/client_unix.go#L6
-			host == "unix:///var/run/docker.sock"
+			host == "unix:///var/run/docker.sock" ||
+
+			// Docker Desktop for Linux - socket is in ~/.docker/desktop/docker.sock
+			(strings.HasPrefix(host, "unix://") && strings.HasSuffix(host, "/.docker/desktop/docker.sock"))
+
 	if isStandardHost {
 		return true
 	}
--- a/internal/docker/client_test.go
+++ b/internal/docker/client_test.go
@@ -312,6 +312,22 @@ func TestProvideClusterProduct(t *testing.T) {
 			},
 		},
 		{
+			env:     clusterid.ProductDockerDesktop,
+			runtime: container.RuntimeDocker,
+			// Set DOCKER_HOST here to mimic docker client discovering socket from desktop-linux context
+			osEnv: map[string]string{
+				"DOCKER_HOST": "unix:///home/tilt/.docker/desktop/docker.sock",
+			},
+			expectedCluster: Env{
+				Client:              hostClient{Host: "unix:///home/tilt/.docker/desktop/docker.sock"},
+				BuildToKubeContexts: []string{"docker-desktop-me"},
+			},
+			expectedLocal: Env{
+				Client:              hostClient{Host: "unix:///home/tilt/.docker/desktop/docker.sock"},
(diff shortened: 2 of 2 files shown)


internal/docker/env.go
```diff
@@ -300,7 +303,11 @@ func isDefaultHost(e Env) bool {
 			host == "npipe:////./pipe/docker_engine" ||
 
 			// https://github.com/moby/moby/blob/master/client/client_unix.go#L6
-			host == "unix:///var/run/docker.sock"
+			host == "unix:///var/run/docker.sock" ||
+
+			// Docker Desktop for Linux - socket is in ~/.docker/desktop/docker.sock
+			ddLinux.Match([]byte(host))
```
nicks, line 309: `MatchString` would be more idiomatic.

internal/docker/env.go
```diff
@@ -284,6 +285,8 @@ func isOldMinikube(ctx context.Context, minikubeClient k8s.MinikubeClient) bool
 	return minMinikubeVersionBuildkit.GTE(vParsed)
 }
 
+var ddLinux *regexp.Regexp = regexp.MustCompile(`^unix:///home/[^/]+/.docker/desktop/docker.sock$`)
```
nicks, line 288: can we be a bit more permissive here? e.g., `unix:///.*/.docker/desktop/docker.sock`?

i have definitely worked on unix systems where the home directories did not match this regexp, and the relevant manuals note that we shouldn't be assuming they're laid out this way (cf https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch03s08.html)

nicks (approved):

internal/docker/env.go
nicksieger: Right, thanks.

internal/docker/env.go
nicks: 👍

PR #5826, opened 2022-05-18
runtimelog: port docker compose log streamer to the reconciler

nicks:
Hello @milas,

467df7cf94c856c71eef2bc61a016a26b553005e (2022-05-18 15:00:43 -0400)
runtimelog: port docker compose log streamer to the reconciler


nicks:
a lot of this is just moving code around and writing better tests for it. i'm hoping this will be a lot more robust in general when interoperating with docker-compose.

internal/controllers/core/dockercomposelogstream/reconciler.go
```diff
@@ -2,48 +2,251 @@ package dockercomposelogstream
...
 	}
 
 	if apierrors.IsNotFound(err) || obj.ObjectMeta.DeletionTimestamp != nil {
+		r.clearResult(nn)
+		r.manageOwnedProjectWatches(ctx)
 		return ctrl.Result{}, nil
 	}
 
+	ctx = store.MustObjectLogHandler(ctx, r.store, obj)
+	r.manageOwnedProjectWatches(ctx)
```
milas, line 78: I think this is going to result in potentially arbitrary assignment of project watch logs to API objects since the context logger is for a specific obj but `manageOwnedProjectWatches` will operate across everything

internal/controllers/core/dockercomposelogstream/reconciler.go
```diff
@@ -2,48 +2,251 @@ package dockercomposelogstream
...
+	}
+
+	serviceKey := result.serviceKey()
+	r.reconcileContainerState(ctx, obj, serviceKey)
+
+	containerState := r.containers[serviceKey]
+	if containerState == nil || containerState.StartedAt.IsZero() {
+		// wait for the container to start before attaching so that we can filter out old logs
+		// for job containers that can be re-used
+		return
```
milas, line 138: Should this result in a `RequeueAfter` to make sure we come back and start watching logs once it's started? (Or maybe it's taken care of by the event watcher triggering a reconcile when it sees a container start?)

milas (approved):
👏

internal/controllers/core/dockercomposelogstream/reconciler.go
nicks: good catch! it took me a while to respond to this comment, but i got it!!!

internal/controllers/core/dockercomposelogstream/reconciler.go
nicks: ya, that's what the event-based requeuer is for!

PR #5834, opened 2022-06-01
ci: update dependencies

milas:
* Docker Compose v2.6.0
* Helm v3.9.0
* kustomize v4.5.5
* golangci-lint v1.46.2

The only major bump here is for kustomize, but `kubectl` now ships
v4.x built-in, so this should be safe and keep things more in
alignment with what many users are likely using implicitly.


nicks:
ack, sorry, forgot to push. updated this branch with the new image hash.

PR #5839, opened 2022-06-02
engine: don't reparse yaml on every KA/KD upsert

### LINKED ISSUE #5837: Performance issues when reconciling very large CRD manifests
I am currently running into a significant performance issue when applying large CRD manifests to a cluster using tilt. I have spent a little while debugging tilt's internals to pinpoint the issue, and here are my findings:
1. On every iteration of the Store's action loop, the upper reducer function processes a `KubernetesDiscoveryUpsertAction` dispatched by the `kubernetesdiscovery` reconciler.
2. The handler for this action will eventually call `k8sconv.NewKubernetesApplyFilter` which parses the yaml from `KubernetesApplyStatus.ResultYAML`. In my case, one of the large CRD manifests weighs in at over 6 MB. This manifest takes a long time to parse (couple seconds). Parsing YAML is the main contributor to the performance issues, as far as I can tell.
3. The frequency at which `KubernetesDiscoveryUpsertAction`s are dispatched will overwhelm the event queue. Additionally, the following code in Store ends up creating a lot of goroutines, increasing lock contention which causes additional slowdowns (though this seems to not be the main problem):
```go
func (s *Store) Dispatch(action Action) {
	s.actionQueue.add(action)
	go s.drainActions()
}
```

### Steps to Reproduce

1. Obtain a large CRD manifest (e.g. prometheus operator)
2. Apply it using `k8s_yaml`

#### `tilt doctor` Output

```
$ tilt doctor
Tilt: v0.30.1-dev, built 2022-06-01
System: linux-amd64
---
Docker
- Server Version: 20.10.16
- API Version: 1.41
- Builder: 2
---
Kubernetes
- Env: unknown
- Context: kubernetes-admin@kubernetes
- Cluster Name: opni-testing
- Namespace: default
- Version: v1.24.1
---

---
Analytics Settings
- User Mode: opt-in
- Machine: 794145422d20ad7eae9c13868585b957
- Repo: 30P//3m6QZhY3W8urCRPnQ==
```

### Additional Info
Here is a pprof cpu profile from running `tilt up` that may be helpful: [profile.gz](https://github.com/tilt-dev/tilt/files/8819124/profile.gz)
Try viewing it with `go tool pprof -web -show='github\.com\/tilt-dev\.*' profile`.

landism:
Fixes #5837

I described a few other potential solutions in a comment on the issue. This is probably my least favorite of them, but it's nice and cheap. I'm happy to entertain alternatives.

### DIFF
--- a/internal/store/kubernetesdiscoverys/reducers.go
+++ b/internal/store/kubernetesdiscoverys/reducers.go
@@ -42,6 +42,23 @@ func HandleKubernetesDiscoveryDeleteAction(state *store.EngineState, action Kube
 	}
 }
 
+func filterForResource(state *store.EngineState, name string) (*k8sconv.KubernetesApplyFilter, error) {
+	a := state.KubernetesApplys[name]
+	if a == nil {
+		return nil, nil
+	}
+
+	// if the yaml matches the existing resource, use its filter to save re-parsing
+	// (https://github.com/tilt-dev/tilt/issues/5837)
+	if prevResource, ok := state.KubernetesResources[name]; ok {
+		if a.Status.ResultYAML == prevResource.ApplyStatus.ResultYAML {
+			return prevResource.ApplyFilter, nil
+		}
+	}
+
+	return k8sconv.NewKubernetesApplyFilter(a.Status.ResultYAML)
+}
+
 func RefreshKubernetesResource(state *store.EngineState, name string) {
 	var aStatus *v1alpha1.KubernetesApplyStatus
 	a := state.KubernetesApplys[name]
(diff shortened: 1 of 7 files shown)


internal/store/kubernetesdiscoverys/reducers.go
```diff
@@ -50,10 +67,11 @@ func RefreshKubernetesResource(state *store.EngineState, name string) {
 	}
 
 	d := state.KubernetesDiscoverys[name]
-	r, err := k8sconv.NewKubernetesResource(d, aStatus)
+	filter, err := filterForResource(state, name)
```
nicks, line 70: would it work to do an apicmp.DeepEqual() on both the discovery and the apply, and skip the whole refresh branch if they're the same?

this is supposed to be idempotent.

landism: In the problematic code path, [we already do](https://github.com/tilt-dev/tilt/blob/0f1dcaa3eec0e99e3b23883d6ff0546b1fc147ea/internal/store/kubernetesdiscoverys/reducers.go#L27-L32).

My (unverified) understanding of the issue is that there are a lot of pod status changes in the cluster, which generate KD status changes, which generate KD upsert actions, and since the KD status has changed, they generate calls to RefreshKubernetesResource, which reparse yaml that hasn't changed. I'll see if the user can confirm this.

nicks (approved):
ah, ok, that makes sense

milas (approved):
Maybe your least favorite of the options, but all-in-all a pretty solid fix given the constraints IMO 👏

PR #5846, opened 2022-06-03
web: replace node-sass with sass

milas:
The `node-sass` package has been deprecated and superseded by the
`sass` package.

The `node-sass` package used a C library via `node-gyp`, which itself
requires Python to wrangle the compilation. The version of `node-sass`
we were using was old and thus relied on an old version of `node-gyp`,
which only supported <= Python 3.8.

This caused problems with Homebrew builds with embedded assets, as
Python 3.9 is used there, so `node-gyp` would error out.

While it's theoretically possible to upgrade `node-sass` to a newer
version to transitively get a newer `node-gyp` that will not bail on
Python 3.9, that did NOT work out of the box, and the recommended
solution for the problem was...switch to `sass`, which does not have
any native code requirements and is actively supported.

As part of this, the `sass` compiler emitted warnings on deprecated
division outside of `calc()`, so those have all been updated to avoid
producing a bunch of noise in the build log.

I did a visual pass after this change and didn't see any differences
in styling in Chrome on macOS.

tl;dr: This will dramatically simplify producing Homebrew builds with
       embedded assets by sidestepping the need for a Python/`node-gyp`
       build chain just for bundling the UI assets.

### DIFF
--- a/web/package.json
+++ b/web/package.json
@@ -80,11 +80,11 @@
     "fetch-mock": "^9.11.0",
     "jest-junit": "^13.0.0",
     "jsdom": "^18.1.0",
-    "node-sass": "^6.0.1",
     "prettier": "2.4.1",
     "prettier-plugin-organize-imports": "^2.3.4",
     "react-scripts": "^4.0.3",
     "react-test-renderer": "^16.9.0",
+    "sass": "^1.52.2",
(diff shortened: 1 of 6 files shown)


landism (approved):
nice SHA

PR #5859, opened 2022-06-08
engine: consider apply conditions for runtime status

### LINKED ISSUE #5822: job never becomes ready when re-attaching to existing env
## Repro steps

Create the following Tiltfile:

```
load('ext://deployment', 'deployment_create', 'job_create')

docker_build('db', '.', dockerfile_contents="""
FROM busybox
ENTRYPOINT busybox httpd -f -p 8000
""")
deployment_create('db')

docker_build('db-init', '.', dockerfile_contents="""
FROM busybox
ENTRYPOINT echo INIT
""")
job_create('db-init', resource_deps=['db'])

docker_build('app', '.', dockerfile_contents="""
FROM busybox
ENTRYPOINT busybox httpd -f -p 8000
""")
deployment_create('app', resource_deps=['db-init'])
```

Then:
1) Run `tilt ci`
2) Manually delete the db-init pod
3) Run `tilt ci` a second time

### Current Behavior

Tilt reattaches to the existing db-init Job object.

There's no pod associated with the Job, so db-init never becomes ready.

App waits forever for db-init to become ready.

### Expected Behavior

Entire app should eventually become ready.

### Ideal Behavior Notes

In the original bug report, the way the pod gets deleted is that the node is drained.

One thing I haven't quite clarified in my mind: from a user perspective, what should `db-init` actually DO on the second tilt up? Possible answers:

- The `db-init` job should never re-run (because we're reconnecting to an existing env)

- The `db-init` job should always re-run (because we're re-initializing an existing env)

- The `db-init` job should only run if we created a fresh `db` Deployment (even if it already exists)

But that doesn't speak to how we would implement this in practice in terms of Kubernetes objects.

### Possible Implementation Notes

- Check completeness on apply

The simplest way to fix this is, after we apply the job, check if it's already Complete (in the status conditions API).

This feels kind of gross because it's special-casing one API kind (Job) and only really works because Jobs are immutable (so isn't generalizable to other API kinds). But it would fix the immediate problem

- Watch for completeness

We could set up API watches on Jobs, and use Job condition Complete=true (rather than monitoring pods). 

Setting up new watches is non-trivial, so this is a lot more complex to implement than the first solution.

But it aligns better with where the Kubernetes API is moving more generally (see: `kubectl wait`), so might be worth investing in.

See also: https://github.com/tilt-dev/tilt/issues/3301

milas:
If a Job ran to completion in the past, so the apply was a no-op,
and the Pod is now gone, consider the Job as being in an "OK"
state. See #5858 for API changes.

Fixes #5822.

### DIFF
--- a/internal/controllers/core/kubernetesapply/reconciler.go
+++ b/internal/controllers/core/kubernetesapply/reconciler.go
@@ -7,6 +7,7 @@ import (
 	"sync"
 
 	"github.com/pkg/errors"
+	batchv1 "k8s.io/api/batch/v1"
 	v1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -596,6 +597,30 @@ type applyResult struct {
 	Objects            []k8s.K8sEntity
 }
 
+func conditionsFromApply(result applyResult) []metav1.Condition {
+	if result.Error != "" || len(result.Objects) == 0 {
+		return nil
+	}
+
+	for _, e := range result.Objects {
+		if e.HasKind("Job") && e.ImmutableOnceCreated() {
+			job := e.Obj.(*batchv1.Job)
+			for _, cond := range job.Status.Conditions {
+				if cond.Type == batchv1.JobComplete && cond.Status == v1.ConditionTrue {
+					return []metav1.Condition{
+						{
+							Type:   v1alpha1.ApplyConditionJobComplete,
+							Status: metav1.ConditionTrue,
+						},
+					}
+				}
+			}
+		}
+	}
+
+	return nil
+}
+
 // Create a result object if necessary. Caller must hold the mutex.
 func (r *Reconciler) ensureResultExists(nn types.NamespacedName) *Result {
 	existing, hasExisting := r.results[nn]
@@ -631,6 +656,7 @@ func (r *Reconciler) recordApplyResult(
 	updatedStatus.LastApplyStartTime = applyResult.LastApplyStartTime
 	updatedStatus.LastApplyTime = applyResult.LastApplyTime
 	updatedStatus.AppliedInputHash = applyResult.AppliedInputHash
+	updatedStatus.Conditions = conditionsFromApply(applyResult)
 
 	result.Cluster = cluster
 	result.Spec = spec
--- a/internal/controllers/core/kubernetesapply/reconciler.go
+++ b/internal/controllers/core/kubernetesapply/reconciler.go
@@ -597,27 +597,31 @@ type applyResult struct {
 	Objects            []k8s.K8sEntity
 }
 
+// conditionsFromApply extracts any conditions based on the result.
+//
+// Currently, this is only used as part of special handling for Jobs, which
+// might have already completed successfully in the past.
 func conditionsFromApply(result applyResult) []metav1.Condition {
 	if result.Error != "" || len(result.Objects) == 0 {
 		return nil
 	}
 
 	for _, e := range result.Objects {
-		if e.HasKind("Job") && e.ImmutableOnceCreated() {
-			job := e.Obj.(*batchv1.Job)
-			for _, cond := range job.Status.Conditions {
-				if cond.Type == batchv1.JobComplete && cond.Status == v1.ConditionTrue {
-					return []metav1.Condition{
-						{
-							Type:   v1alpha1.ApplyConditionJobComplete,
-							Status: metav1.ConditionTrue,
-						},
-					}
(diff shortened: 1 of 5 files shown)


internal/controllers/core/kubernetesapply/reconciler_test.go
```diff
@@ -295,6 +297,55 @@ func TestBasicApplyCmd_MalformedYAML(t *testing.T) {
...
+			Status: metav1.ConditionTrue,
+		},
+	}
+
+	assert.Equal(f.T(), ka.Status.Conditions, expected,
+		"KubernetesApply status should reflect Job completion")
+
+	// Make sure that re-reconciling doesn't clear the conditions
+	f.MustReconcile(types.NamespacedName{Name: "a"})
+	assert.Equal(f.T(), ka.Status.Conditions, expected,
```
nicksieger, line 345: Should you re-get `ka` here?
```suggestion
	f.MustGet(types.NamespacedName{Name: "a"}, &ka)
	assert.Equal(f.T(), ka.Status.Conditions, expected,
```

internal/store/kubernetesdiscoverys/reducers.go
```diff
@@ -88,21 +89,28 @@ func RefreshKubernetesResource(state *store.EngineState, name string) {
...
+				for _, pod := range r.FilteredPods {
+					if krs.PodReadinessMode == model.PodReadinessSucceeded {
+						// for jobs, we don't care about whether it's ready, only whether it's succeeded
+						isReadyOrSucceeded = pod.Phase == string(v1.PodSucceeded)
+					} else {
+						isReadyOrSucceeded = len(pod.Containers) != 0 && store.AllPodContainersReady(pod)
+					}
 				}
+			} else {
+				isReadyOrSucceeded = meta.IsStatusConditionTrue(r.ApplyStatus.Conditions, v1alpha1.ApplyConditionJobComplete)
```
nicks, line 105: it should be ok to do this check even if FilteredPods != 0

nicksieger (approved):

internal/store/kubernetesdiscoverys/reducers.go
```diff
@@ -88,21 +89,28 @@ func RefreshKubernetesResource(state *store.EngineState, name string) {
...
-				}
-				if isReadyOrSucceeded {
-					// NOTE(nick): It doesn't seem right to update this timestamp everytime
-					// we get a new event, but it's what the old code did.
-					krs.LastReadyOrSucceededTime = time.Now()
+			if len(r.FilteredPods) != 0 {
+				for _, pod := range r.FilteredPods {
+					if krs.PodReadinessMode == model.PodReadinessSucceeded {
+						// for jobs, we don't care about whether it's ready, only whether it's succeeded
+						isReadyOrSucceeded = pod.Phase == string(v1.PodSucceeded)
```
nicks, line 99: unrelated to this pr, but i'm pretty sure this logic is wrong? this will use the readiness state of the last pod (and the order of pods is non-deterministic).

maybe break this out so that it's only ready if all pods are ready? (The FilteredPods will take care of filtering old pods)

nicks (approved):

internal/controllers/core/kubernetesapply/reconciler.go
```diff
@@ -596,6 +597,30 @@ type applyResult struct {
...
 
+func conditionsFromApply(result applyResult) []metav1.Condition {
+	if result.Error != "" || len(result.Objects) == 0 {
+		return nil
+	}
+
+	for _, e := range result.Objects {
+		if e.HasKind("Job") && e.ImmutableOnceCreated() {
+			job := e.Obj.(*batchv1.Job)
+			for _, cond := range job.Status.Conditions {
```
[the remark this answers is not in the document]
milas: Not on this one! It's a `batch/v1.JobCondition` and not a `meta/v1.Condition` womp womp

internal/controllers/core/kubernetesapply/reconciler.go
```diff
@@ -596,6 +597,30 @@ type applyResult struct {
...
 }
 
+func conditionsFromApply(result applyResult) []metav1.Condition {
+	if result.Error != "" || len(result.Objects) == 0 {
+		return nil
+	}
+
+	for _, e := range result.Objects {
+		if e.HasKind("Job") && e.ImmutableOnceCreated() {
+			job := e.Obj.(*batchv1.Job)
```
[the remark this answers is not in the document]
milas: This is true 😛  - done!

internal/store/kubernetesdiscoverys/reducers.go
milas: Updated, there's technically still a length check but it's _after_ the status condition check (needed to handle the other change that makes us look at _all_ valid Pods instead of just the final one during iteration)

PR #5864, opened 2022-06-10
Replace registry when using outputs_image_ref_to with different cluster host

nicksieger:
- chore: go mod tidy and use gotestsum for local `make test-go`
- build: outputsImageRefTo cluster ref uses HostFromContainerRuntime

### DIFF
--- a/internal/build/custom_builder.go
+++ b/internal/build/custom_builder.go
@@ -39,7 +39,8 @@ func (b *CustomBuilder) Build(ctx context.Context, refs container.RefSet,
 	expectedTag := spec.OutputTag
 	outputsImageRefTo := spec.OutputsImageRefTo
 	var registryHost string
-	if reg := refs.Registry(); reg != nil {
+	reg := refs.Registry()
+	if reg != nil {
 		registryHost = reg.Host
 	}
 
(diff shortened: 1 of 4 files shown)


internal/build/custom_builder.go
```diff
@@ -184,9 +185,20 @@ func (b *CustomBuilder) readImageRef(ctx context.Context, outputsImageRefTo stri
 			outputsImageRefTo, err)
 	}
 
-	// TODO(nick): Add support for separate local and cluster refs.
+	clusterRef := ref
+	if reg != nil && reg.HostFromContainerRuntime != "" {
+		replacedName, err := container.ParseNamed(strings.Replace(ref.Name(), reg.Host, reg.HostFromContainerRuntime, 1))
```
nicksieger, line 190: Is there a better way to do this?

milas: There is! It's not as discoverable since I flattened the Registry types recently, but it's here: https://github.com/tilt-dev/tilt/blob/33fdb81f7bbbf240edb35d17103f7b3936c39244/internal/container/registry.go#L111-L123

(And for completeness, to get the ref type it wants -> `container.NewRefSelector(ref)`)

milas (approved):

internal/build/custom_builder.go
nicksieger: Hmm, I think I tried this at one point. It would turn a ref like `localhost:500/foo-bar` into `remote:5000/localhost_5000_foo_bar`, which isn't the same since the custom builder is pushing the image to the registry.

milas: Ahhh d'oh, you're totally right!

nicksieger: Cool, thanks! Just wanted to make sure I didn't miss anything else.

PR #5890, opened 2022-07-12
cli: fix link to tiltfile_config doc

liskin:
<https://tilt.dev/user_config.html> is 404 and digging in <https://github.com/tilt-dev/tilt.build/> didn't reveal when it disappeared, so… no idea how long this has been broken, but now it is and should be fixed. :-/


nicks (approved):
LGTM, thanks!

Issue #5898, opened 2022-07-20
Feature Request: dropdown / choice input for cmd_button


jbadlato:
`text_input` gives us unlimited options, and `bool_input` give us the choice between two options, but it would be nice to have something in between.  For my use case, I want to give the user a list of supported values in a dropdown selection.

I tried to see if this would be something easy to extend myself, but it looks pretty involved? May be interested in giving it a try if I'm wrong though.

nicks:
This might be a good external contributor issue, TBH. Maybe not a FIRST contributor issue (it touches a lot), but it's basically adding new fields to existing wiring. Here's a sample PR:
https://github.com/tilt-dev/tilt/pull/4893

Don't be put off by the large diff! It's mostly generated code.
The important bits are:
1) Updating the button spec: https://github.com/tilt-dev/tilt/blob/master/pkg/apis/core/v1alpha1/uibutton_types.go
2) Running `make update-codegen` to generate all the new APIs from the button spec
3) Interpreting that spec in the React component: https://github.com/tilt-dev/tilt/blob/master/web/src/ApiButton.tsx

jbadlato:
@nicks Thanks, that was super helpful, and was definitely enough direction for me to hack together a PR!

PR #5900, opened 2022-07-22
UIButton: Add choice_input / dropdown selections

jbadlato:
Implements #5898 

I have pretty limited experience with typescript/react/go, so I'm sure there's some convention issues/bad practices here (including some `@ts-ignore`s that I don't know how/if to fix.  I cargo-culted pretty much everything here from #4893 & what else I could gather from the surrounding code.

I also considered implementing this as radio buttons, but that should probably be easy to expose as another parameter to `choice_input` as a future improvement?

choice_input

Smoke tested this by editing the oneup Tiltfile: 
```
diff --git a/integration/oneup/Tiltfile b/integration/oneup/Tiltfile
index 1cf5dda00..62b0e1c71 100644
--- a/integration/oneup/Tiltfile
+++ b/integration/oneup/Tiltfile
@@ -1,5 +1,6 @@
 # -*- mode: Python -*-
 
+load('ext://uibutton', 'cmd_button')
 include('../Tiltfile')
 
 # If you get push errors, you can change the default_registry.
@@ -11,3 +12,20 @@ default_registry(read_json('tilt_option.json', {})
 docker_build('oneup', '.')
 k8s_yaml('oneup.yaml')
 k8s_resource('oneup', port_forwards=8100)
+
+
+def choice_input(name, label='', default='', choices=[]):
+    result = {
+        "name": name,
+        "label": label,
+        "choice": {
+            "defaultValue": default,
+            "choices": choices
+        }
+    }
+    return result
+
+cmd_button(name='Test Choice',
+           resource='oneup',
+           argv=['sh', '-c', "echo choice: $MY_CHOICE_INPUT"],
+           inputs=[choice_input(name='MY_CHOICE_INPUT', label='input name label', default='choice2', choices=['choice1', 'choice2', 'choice3'])])
```


pkg/apis/core/v1alpha1/uibutton_types.go
```diff
@@ -253,6 +253,23 @@ type UIHiddenInputStatus struct {
 	Value string `json:"value" protobuf:"bytes,1,opt,name=value"`
 }
 
+// Describes a choice dropdown input field attached to a button
+type UIChoiceInputSpec struct {
+	// Initial value for this field.
+	//
+	// +optional
+	DefaultValue string `json:"defaultValue" protobuf:"bytes,1,opt,name=defaultValue"`
```
nicks, line 261: is it ok to have a DefaultValue that's not an element of the Choices array?

alternatively, what do you think of getting rid of DefaultValue, and making the default value implicitly the first Choice?

pkg/apis/core/v1alpha1/uibutton_types.go
```diff
@@ -292,6 +312,9 @@ func (in *UIInputSpec) Validate(_ context.Context, path *field.Path) field.Error
 	if in.Hidden != nil {
 		numInputTypes += 1
 	}
+	if in.Choice != nil {
+		numInputTypes += 1
```
nicks, line 316: we should also validate that Choices is non-empty if this field is defined

web/tilt_modules/uibutton/README.md
```diff
@@ -54,6 +54,17 @@ checkbox's state. By default, the variable will be set to the string `"true"` or
 | `true_string`  | `str`  | If not None, when the checkbox is checked, the environment variable will be set to this string instead of "true". |
 | `false_string` | `str`  | If not None, when the checkbox is checked, the environment variable will be set to this string instead of "false". |
 
+### `choice_input(name, label='', default='', choices=[])`
```
nicks, line 57: i think this tilt_modules directory is vestigial, lemme remove it in another pr

nicks (commented):
nice!! quick high-level question about the schema

pkg/apis/core/v1alpha1/uibutton_types.go
jbadlato: Yeah now that you mention it, that seems like a way better solution.  I'll get rid of the DefaultValue

web/tilt_modules/uibutton/README.md
jbadlato: Should I be making a separate PR to https://github.com/tilt-dev/tilt-extensions/?

nicks: ya, but you'll have to wait for a release

jbadlato: Sounds good!

jbadlato:
@nicks Alright I think I've cleaned up all the code with the suggestions so far.  I'm not sure how to resolve the DCO check, since the details only give remedy instructions if I don't have empty commits on my branch, which I do.  Should I just squash these to a new branch and create a new PR?

nicks:
@jbadlato thanks! ya, it's probably easiest to squash and push --force the pr

nicks (approved):
lgtm!

PR #5903, opened 2022-07-25
Update emacs ignore patterns

person142:
Currently the emacs ignore patterns include `**/.#*` (lock files), but
doesn't include `**/#*#` (autosave files;
https://www.emacswiki.org/emacs/AutoSave, not to be confused with
`**/*~` backup files, which are ignored.)

Add autosave files.


nicks (approved):
LGTM! Thanks!

PR #5935, opened 2022-09-09
feat: add k8s_namespace() to pull cached namespace

### LINKED ISSUE #5934: Make the namespace from Tilt's k8s client config available in the Tiltfile
A `k8s_namespace()` to match the existing [`k8s_context`](https://docs.tilt.dev/api.html#api.k8s_context)

### Current Behavior
Inside the Tiltfile, there is not a good way to tell what namespace Tilt is deploying to.

### Why Do You Want This?
Tilt loads the k8s client config into memory and freezes its value at startup, continuing to use that despite any changes that might occur after Tilt starts.
To be consistent with this behavior, any cluster interactions the user puts in their Tiltfile that don't go through Tilt's k8s client (e.g., invoking `kubectl apply` directly) should use `--namespace` and `--context` with appropriate values. Tilt currently provides `k8s_context` to get the name of the context, but does not provide a builtin to get the namespace Tilt is using.
Additionally, users who change their context / namespace elsewhere have reported being unaware why their other k8s tools are not showing things done by Tilt (e.g., Tilt says it's deploying pods, but `kubectl get pods` shows nothing). The ability to highlight this situation in Tilt could save a lot of debugging time.

**Additional context**
The namespace is available via `tilt get cluster -ojson`, but the cluster object doesn't exist until after the first Tiltfile execution.

nsheaps:
This will add a `k8s_namespace()` function to tiltfiles to allow them to fetch the namespace that tilt is configured to use, regardless of if the kubeconfig changes which namespace or context it's targeted at.

Still needs docs update

Closes #5934


internal/tiltfile/tiltfile_test.go
```diff
@@ -5765,7 +5766,7 @@ type fixture struct {
 func (f *fixture) newTiltfileLoader() TiltfileLoader {
 	dcc := dockercompose.NewDockerComposeClient(docker.LocalEnv{})
 
-	k8sContextPlugin := k8scontext.NewPlugin(f.k8sContext, f.k8sEnv)
```
kylelemons, line 5768: missing test coverage, see:
https://sourcegraph.com/github.com/tilt-dev/tilt/-/blob/internal/tiltfile/tiltfile_test.go?L3601:4

kylelemons (changes requested):

internal/tiltfile/tiltfile_test.go
nsheaps: Fixed

nicks:
lgtm, but the build is failing because you need to run `gofmt` or `goimports`?

kylelemons (approved):

nicks (approved):
LGTM! thanks!

PR #5936, opened 2022-09-12
Detect non-admin socket for Rancher Desktop on Mac

ewilkins-csi:
Rancher Desktop has an option to run in a mode that does not require `sudo` access on Linux and Mac.  In these cases, Rancher Desktop creates the docker socket at `~/.rd/docker.sock`.  This PR updates the docker socket connection logic to account for this case.

Related issues:
#5367 
#5751


nicks (approved):
LGTM, thank you!

PR #5952, opened 2022-10-10
tiltfile: load compose services from multiple projects/directories

### LINKED ISSUE #5170: support docker-compose configs from multiple directories
The ability to compose multiple docker-compose files from different projects, e.g. `docker_compose(['proj1/docker-compose.yaml', 'proj2/docker-compose.yaml'])`

### Current Behavior
We infer the DC project directory from the first config provided and then use that for all other DC configs.
This means, e.g., paths in `proj2` are interpreted relative to `proj1`, even though they are likely written to be interpreted relative to `proj2`.

### Why Do You Want This?
Onboarding to Tilt and wanting to run multiple DC projects at once

### Notes
* This actually doesn't work in DC - https://github.com/docker/compose/issues/3874
* Maybe this should actually be a bug instead of a feature request? Except we're matching the underlying DC behavior, so a user migrating from DC couldn't have had this working in the first place.

nicksieger:
- Load and track multiple compose projects
- Allow DC resources to be renamed in the event of collisions between multiple projects.
- Upgrade compose-go to v1.6.0

Fixes #5170.


internal/tiltfile/tiltfile_state.go
```diff
@@ -685,25 +696,28 @@ func (s *tiltfileState) assembleImages() error {
...
+			} else {
+				// create a DockerComposeBuild image target and consume it if this service has a build section in YAML
+				err := s.maybeAddDockerComposeImageBuilder(svc)
+				if err != nil {
+					return err
+				}
 			}
+			// TODO(maia): throw warning if
+			//  a. there is an img ref from config, and img ref from user doesn't match
+			//  b. there is no img ref from config, and img ref from user is not of form .*_<svc_name>
```
nicksieger, line 719: Not sure if these todos are valid anymore? Delete?

nicks: :+1:

internal/tiltfile/tiltfile_docker_compose_test.go
```diff
@@ -421,6 +442,30 @@ k8s_yaml('bar.yaml')`
 	assert.Equal(t, 2, len(f.loadResult.Manifests))
 }
 
+func TestDockerComposeAndK8sNameConflict(t *testing.T) {
```
nicks, line 445: can you also add a test for the case where the k8s_yaml comes before the docker_compose?

internal/tiltfile/tiltfile_docker_compose_test.go
```diff
@@ -421,6 +442,30 @@ k8s_yaml('bar.yaml')`
...
+	f.setupFooAndBar()
+	f.file("docker-compose.yml", simpleConfig)
+	tf := `docker_compose('docker-compose.yml')
+k8s_yaml('foo.yaml')`
+	f.file("Tiltfile", tf)
+
+	f.loadErrString(`k8s_resource named "foo" already exists`)
+}
+
+func TestDockerComposeAndLocalNameConflict(t *testing.T) {
```
nicks, line 457: ditto

internal/tiltfile/k8s.go
```diff
@@ -632,23 +632,35 @@ func makeWorkloadToResourceFunction(f *starlark.Function) (workloadToResourceFun
 	}, nil
 }
 
-func (s *tiltfileState) checkResourceConflict(name string) error {
-	if s.k8sByName[name] != nil {
-		return fmt.Errorf("k8s_resource named %q already exists", name)
+func (s *tiltfileState) checkResourceConflicts() error {
```
nicks, line 635: My big high-level concern with this approach is that users don't understand how resource names are generated. 

The old approach deliberately pointed them at the line of code where the resource name was generated. 

The new approach doesn't give them any hints. I don't think users will know how to fix this new error and we will get lots of complaints about this.

nicksieger: Yeah, doing the old approach is trickier now with the ability to rename both k8s and dc resources. And it shows up in a worse errors experience. I'll give another go at improving it. Maybe it will work to track the files and line numbers of all the resources as they are created and then use them here to report errors during conflict resolution.

nicks: for now, could we go back to the old way of checking resource name conflicts on creation rather than on assembly?

i understand that the assembly approach makes it possible to rename resources and move things around...but i don't think we actually have the debugging tools to support a good user experience for that.

nicks: (i.e., i don't think we want users to have to think really hard about the order that resource naming happens in. we should fail fast as soon as there's a conflict, rather than trying to optimize the case where the conflict is resolved further down)

nicks (approved):
LGTM!

nicks:
will merge this and do a release tomorrow morning

Issue #5956, opened 2022-10-20
Kustomize does not work with helm charts


Blinkuu:
### Expected Behavior
When specifying helm charts in `kustomization.yaml`, I expect Tilt to apply those, similarly to applying simple resources. Perhaps Tilt should call customize with `--enable-helm` option enabled by default.

### Current Behavior

Error (anonymized directories):
```
Error in kustomize: command "kubectl kustomize operations/kubernetes/development/tilt" failed.
error: exit status 1
stdout: ""
stderr: "error: accumulating resources: accumulation err='accumulating resources from 'fancy-project': '/Users/Dev/fancy-project/operations/kubernetes/development/tilt/fancy-project' must resolve to a file': recursed accumulation of path '/Users/Dev/fancy-project/operations/kubernetes/development/tilt/fancy-project': accumulating resources: accumulation err='accumulating resources from '../../../base/timescaledb': '/Users/Dev/fancy-project/operations/kubernetes/base/timescaledb' must resolve to a file': recursed accumulation of path '/Users/Dev/fancy-project/operations/kubernetes/base/timescaledb': trouble configuring builtin HelmChartInflationGenerator with config: `\nname: charts/timescaledb-single\nreleaseName: timescaledb\nrepo: [https://charts.timescale.com\n`:](https://charts.timescale.com/n%60:) must specify --enable-helm\n"
```

### Steps to Reproduce

1. Setup `kustomization.yaml` that references a helm chart using `helmCharts` field.
2. Run `tilt up`

nicks:
Thanks for reporting!

You can workaround this now by using a `local()`

```
local('kustomize build --enable-helm')
```

I'm not convinced that `kustomze()` should be a built-in anymore...we wrote it before we had a real extensions system, and increasingly feels like it should be an extension. But maybe there's a kustomize(flags=...) hack we can add in the meantime.

related issue: https://github.com/tilt-dev/tilt/issues/5007

nicks:
we just released a new version of tilt that allows you to add flags to `kustomize()`, see: https://docs.tilt.dev/api.html#api.kustomize

Blinkuu:
Sweet, thanks @nicks!

Issue #5007, opened 2021-10-04
support kustomize build --enable_alpha_plugins


landism:
## Describe the Feature You Want
The ability to use [kustomize plugins](https://kubectl.docs.kubernetes.io/guides/extending_kustomize/) with Tilt's [`kustomize builtin`](https://docs.tilt.dev/api.html#api.kustomize).
(alternatively, the ability to invoke `kustomize` without having to reimplement the builtin's [kustomize dependency calculation](https://github.com/tilt-dev/tilt/blob/bbc177213ee576959bea4efdca31215bb507e4b6/internal/kustomize/dependencies.go#L70))

### Current Behavior
The `kustomize builtin` does not really have any parameterization

**Additional context**
* A potential workaround is this:
```
yaml = local('kustomize build --alpha-plugins deployments')
k8s_yaml(yaml)
watch_file('deployments') # reload if anything in this directory changes
```
This isn't quite the same since it's just assuming the dependencies are the "deployments" directory, which might be both too narrow and too wide.
* Potential solutions:
  * Break out dependency calculations into a separate builtin / extension, so users can use that to calculate deps and `local` to invoke kustomize however they want.
  * Add an arg to the `kustomize` builtin specifically to enable plugins
  * Add an arg to the `kustomize` builtin to allow users to add arbitrary options to the `kustomize call`

albaquarto:
it would also be useful if Tilt built the plugin image watching files changes.

tedtramonte:
I'd also love to see this, as I wrote my own Kustomize transformer and can't really use it in my dev environment with Tilt or my prod environment with ArgoCD.

I know it's categorized as "alpha" but there's really no reason NOT to pass the flag by default. 

The flag wont negatively impact anyone, and I find the security concerns a bit overblown, considering we're either
- writing these deployments ourselves and therefore can trust what we're running
- we're deploying code we didn't write and it's already not fully trusted and therefore may as well run it all

nicks:
we just released a new version of tilt that allows you to add flags to `kustomize()`, see: https://docs.tilt.dev/api.html#api.kustomize

PR #5957, opened 2022-10-20
tiltfile: add kustomize(flags) for passing extra flags

nicks:
also print kustomize invocation like we do with helm

fixes https://github.com/tilt-dev/tilt/issues/5956
fixes https://github.com/tilt-dev/tilt/issues/5007


nicksieger (approved):
Nice!

milas (approved):

Issue #5926, opened 2022-08-19
custom_build not handling cancellation correctly


nicks:
### Steps to Reproduce

Use the following tiltfile:

```
custom_build('example-html-image', 'sleep 1000', deps=[])
k8s_yaml('kubernetes.yaml')
k8s_resource('example-html', port_forwards=8000)
```

(adapted from tilt-example-html)

Run `tilt up`

Hit the cancel button on `example-html`

### Expected Behavior
The sleep should get killed

### Current Behavior
The sleep is not killed properly.

PR #5971, opened 2022-11-11
build: support cancellation of custom image builds

nicks:
Custom image builds now create an explicit Cmd object in the API server. But for now these Cmds are kept up to date as part of the buildcontroller, rather than by the runtime controller.

Fixes https://github.com/tilt-dev/tilt/issues/5926


internal/build/custom_builder.go
```diff
@@ -107,20 +106,21 @@ func (b *CustomBuilder) Build(ctx context.Context, refs container.RefSet,
 			l.Infof("  %s", v)
 		}
 	}
-	cmd.Env = append(os.Environ(), extraEnvVars...)
-	err = imagemap.InjectIntoLocalEnv(cmd, spec.ImageMaps, imageMaps)
+	cmd.Spec.Env = append(cmd.Spec.Env, extraEnvVars...)
```
milas, line 109: Will the spec already have the `os.Environ()`?

AFAICT we don't add it before execing in `ForceRun` -> https://github.com/nicks/tilt/blob/cfca55cae493dce8616a0ebf986d10e9922eec0a/internal/controllers/core/cmd/controller.go#L389-L399

internal/engine/buildcontrol/docker_compose_build_and_deployer.go
```diff
@@ -166,8 +166,17 @@ func (bd *DockerComposeBuildAndDeployer) BuildAndDeploy(ctx context.Context, st
 			return store.ImageBuildResult{}, fmt.Errorf("Not an image target: %T", target)
 		}
 
+		var cmd v1alpha1.Cmd
+		if iTarget.CmdImageName != "" {
```
milas, line 170: `cmd` is going to always be non-nil, which might be unexpected within `Build()` - might be better to pass a nil pointer if there is no custom build command

milas (approved):

Issue #5949, opened 2022-10-07
`tilt ci` never completes if job goes away


nicks:
### Steps to Reproduce
Start with this project:
https://github.com/tilt-dev/tilt-example-html/tree/master/0-base

Change the tiltfile to:

```
docker_build('example-html-image', '.', entrypoint='sleep 1 && echo success')
k8s_yaml('kubernetes.yaml')
k8s_resource('example-html')
local_resource('hi', 'sleep 5', resource_deps=['example-html'])
```

Change the kubernetes yaml to:
```
apiVersion: batch/v1
kind: Job
metadata:
  name: example-html
  labels:
    app: example-html
spec:
  ttlSecondsAfterFinished: 1
  template:
    spec:
      restartPolicy: Never
      containers:
      - name: example-html
        image: example-html-image
```

### Expected Behavior
`tilt ci` should notice that both jobs have succeeded and finish successfully.

### Actual behavior
The job gets cleaned up and `tilt ci` never finishes.

#### About Your Use Case
I think we need to adjust the `tilt ci` semantics for jobs so that if the Job goes away after finishing, it doesn't revert back to pending UNLESS you redeploy the job for some reason.

nicks:
poking around at this a bit. i think there's a more fundamental problem with this `ttlSecondsAfterFinished` feature. Tilt uses a lot of the core k8s api infrastucture. The core k8s api infrastructure has an assumption deeply baked in: the api server is the source of truth, and it's OK to miss events, because you will eventually re-sync. So the code doesn't worry too much about dropped events.

`ttlSecondsAfterFinished` breaks this assumption. If you miss an event, the Job might get deleted before you get notified that it succeeded.

an idea: what if tilt removed `ttlSecondsAfterFinished` from any jobs it deploys?

I think I could convince myself that if you're using Tilt to deploy a job, then you want Tilt to control the Job's lifecycle, not the Job Controller. e.g., you want the job status to persist in the cluster until you run `tilt down`

PR #5972, opened 2022-11-14
kubernetesdiscovery: continue to track pods that are completed but deleted

nicks:
Fixes https://github.com/tilt-dev/tilt/issues/5949


nicksieger (approved):
LGTM

Issue #5970, opened 2022-11-10
Make it possible to run tilt up offline


roma-glushko:
## Describe the Feature You Want
Currently it doesn't seem to be possible to run tilt up offline if you have extensions used in your Tiltfile, e.g.

```
load('ext://dotenv', 'dotenv')

dotenv()
```

### Current Behavior
Tilt constantly tries to pull github repo with extensions on `tilt up` command. There is no way (afaik) to cache that requests to be able to run tilt locally offline.

### Why Do You Want This?
- To make it possible to work under bad internet/connection circumstances (when you are in a long flight, outside of the good connection)
- Local cluster is kind of imply that it would may be run locally e.g. without external dependencies. 

**Additional context**
N/A

nicks:
thanks for the report! i think this is just a bug, the repo fetcher is supposed to cache things in the global xdg-base cache.

PR #5973, opened 2022-11-14
extensionrepo: if the extensionrepo fails to download, fallback to offline

nicks:
Fixes https://github.com/tilt-dev/tilt/issues/5970


nicks:
i forgot what it was like to have a repo with lots of codegen!

landism (approved):

Issue #5913, opened 2022-08-04
force restart does not rebuild image


nhomble:
### Expected Behavior

Given a resource from `docker_compose` that points to a `Dockerfile`, I expected the restart button to rebuild my image.

### Current Behavior
The same image is restarted.

### Steps to Reproduce

1. used the button described in https://blog.tilt.dev/2019/11/14/force-update.html
2. found similar discussion in GH-1012

#### `tilt doctor` Output

```
$ tilt doctor
Tilt: v0.30.5, built 2022-07-14
System: darwin-arm64
---
Docker
- Host: unix:///var/run/docker.sock
- Server Version: 20.10.14
- API Version: 1.41
- Builder: 2
- Compose Version: v2.5.1
Thanks for seeing the Tilt Doctor!
Please send the info above when filing bug reports. 💗
```

#### About Your Use Case

As I was developing my image, I was adding new files to my image (passed as a volume). On a restart, I wasn't seeing the changes. 

My `Tiltfile` is minimal

```
docker_compose('docker-compose.yml')
```

Where the `docker-compose.yml` looks like

```
version: '3.9'
services:
  svc:
    build: ./thing
```

My workaround is to:
1. disable my resource
2. enable my resource
3. restart

nicks:
thanks for filing! ya, i think this might just be an oversight

nhomble:
Happy to help out as well! I'm a big fan of tilt @nicks

mleiter06:
This is a big one for me. 

Additionally, it seems like this is part of the same issue. When I click the force restart button, the logs do not reattach. When I click disable/enable like nhomble suggested, the logs DO reattach.

PR #5977, opened 2022-11-18
buildcontrol: restart should do a full delete of the resource

nicks:
This is similar to how it works for the kubernetes controllers; this is just porting the same logic to the compose controllers.

fixes https://github.com/tilt-dev/tilt/issues/5913


nicks:
i was working on a project using compose and really missed this feature!

nicksieger (approved):
👍

PR #5982, opened 2022-11-30
config: copy go's file lock and use it instead of client-go's

### LINKED ISSUE #4814: Error: writing tilt api configs: open /home/$USER/.tilt-dev/config.lock: file exists
Two users have reported getting this error when starting Tilt:
`Error: writing tilt api configs: open /home/$USER/.tilt-dev/config.lock: file exists`

We don't have a repro.

The issue can be worked around by deleting the config.lock file.

nicksieger:
Fixes #4814.

Signed-off-by: Nick Sieger <nick@nicksieger.com>

### DIFF
--- a/internal/filelock/with_lock.go
+++ b/internal/filelock/with_lock.go
@@ -1,0 +1,33 @@
+package filelock
+
+import (
+	"os"
+
+	"k8s.io/client-go/tools/clientcmd"
+)
+
+func init() {
+	// We're using our own file locking mechanism
+	clientcmd.UseModifyConfigLock = false
+}
+
+func WithLock(configAccess clientcmd.ConfigAccess, action func() error) error {
+	return withLock(configAccess.GetDefaultFilename()+".lock", writeLock, action)
+}
+
+func WithRLock(configAccess clientcmd.ConfigAccess, action func() error) error {
+	return withLock(configAccess.GetDefaultFilename()+".lock", readLock, action)
+}
+
+func withLock(filename string, lt lockType, action func() error) error {
+	lockfile, err := os.Create(filename)
+	if err != nil {
+		return err
+	}
+	err = lock(lockfile, lt)
+	if err != nil {
+		return err
+	}
+	defer func() { _ = Unlock(lockfile) }()
+	return action()
+}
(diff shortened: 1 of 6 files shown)


internal/filelock/with_lock.go
```diff
@@ -0,0 +1,33 @@
...
+
+import (
+	"os"
+
+	"k8s.io/client-go/tools/clientcmd"
+)
+
+func init() {
+	// We're using our own file locking mechanism
+	clientcmd.UseModifyConfigLock = false
```
nicks, line 11: i think my big concern with this is that it's global state.

it assumes you've updated all the call sites.

i'm not at all confident you've updated all the call sites (particularly the ones in our dependencies)

i'd honestly be more comfortable with a patched version of client-go that uses our locking mechanism. (we already patch apimachinery)

nicks (approved):
neat! were you able to reproduce the lock failure?

nicksieger:
I've seen it from time to time, never consistently. Recent report in slack made me realize that it's probably Tilt getting killed by some other process and the I/O of the HUD server de-registering itself is probably a convenient place for the process to pause, get killed and fail to delete the lock file.

internal/filelock/with_lock.go
nicksieger: Yeah, it's not ideal and could be easily overlooked if we started expanding the ways in which Tilt handles kubeconfigs. In this case, the only place in Tilt where we write config with client-go is the 2 places in this PR in the hud server controller package. I added read locks in places where we read the same config but client-go already didn't try to achieve read consistency, so those aren't strictly necessari. I could move all code dealing with reading and writing Tilt's kubeconfig to a separate package if you think that would help with the maintenance issues.

As it is, I consider this PR to be a temporary band-aid that should be fixed once Go has a public file locking API, at which point we or someone would send a PR to client-go to change the global. Realizing too that "temporary" fixes often become long-term ones, I feel that this change is worthwhile given the lower amount of maintenance we do on Tilt, the point above on how we only have two locations (in the same package) that do writes and the fact that write collisions on kubeconfigs in a dev environment are pretty unlikely.

nicksieger:
Still going to do some sanity check manual testing on Mac, Windows and Linux before merging this. Let me know if you have any more thoughts about the client-go global state issue.

nicks:
nope, no further thoughts, LGTM

Issue #5981, opened 2022-11-29
Port forwards fail to bind on macOS when `localhost` resolves to `fe80::1%lo0`


louismrose:
### Expected Behavior
Using `k8s_resource` with a port forward should work on macOS, when `localhost` resolves to `fe80::1%lo0`.

### Current Behavior
Port forwarding fails in Tilt with the following error:

```
Reconnecting... Error port-forwarding test-deploy (9090 -> 9090): error forwarding port: fe80::1%lo0 is not a valid IP
```

I believe this is an issue with Tilt because I can successfully configure a port forward via Kubernetes for the same service:
```
$ kubectl port-forward test-deploy-56469867cc-v852c 9090:9090
Forwarding from 127.0.0.1:9090 -> 9090
Forwarding from [::1]:9090 -> 9090
Handling connection for 9090
```

Interestingly if I try to port forward with Kubernetes via an address hardcoded to `fe80::1%lo0` then it fails with exactly the same error message seen in Tilt:
```
$ kubectl port-forward test-deploy-56469867cc-v852c --address fe80::1%lo0 9090:9090
error: fe80::1%lo0 is not a valid IP
```

I suspect this is due to `net.LookupHost` including `fe80::1%lo0` in the list of addresses via the default Go resolver on macOS:

```
$ cat localhost.go
package main

import "fmt"
import "net"

func main() {
        addr, _ := net.LookupHost("localhost")
	fmt.Println("Localhost addresses are:", addr)
}

$ go run localhost.go
Localhost addresses are: [::1 127.0.0.1 fe80::1%lo0]
```

`fe80::1%lo0` seems to appear in `/etc/hosts` [by default on some macOS installations](https://gist.github.com/andretw/e3946bd9c9b4571702e5). I believe the `%lo0` is intended to restrict traffic to a [particular network interface](https://en.wikipedia.org/wiki/IPv6_address#Scoped_literal_IPv6_addresses_(with_zone_index)) via a "zone index". It looks like Kubernetes doesn't support IP addresses with a zone index, so Tilt probably shouldn't pass addresses of this form to Kubernetes when setting up port forwards.

I had a quick look at `internal/k8s/portforward.go`, and it looks like `getListenableAddresses` might be the place to update? Let me know if that's the case, and I can attempt a PR to exclude IP addresses with a zone index.

### Steps to Reproduce

[...]

PR #5984, opened 2022-12-07
Ignores IPv6 addresses containing a zone index #5981

louismrose:
A possible fix for #5981. Ignores IP addresses that contain a zone index when trying to determine which IP address to use for port forwarding.

### DIFF
--- a/internal/k8s/portforward.go
+++ b/internal/k8s/portforward.go
@@ -168,6 +168,12 @@ func getListenableAddresses(host string) ([]string, error) {
 	for _, addr := range addresses {
 		var l net.Listener
 		if ipv6 := strings.Contains(addr, ":"); ipv6 {
+			// skip ipv6 addresses that include a zone index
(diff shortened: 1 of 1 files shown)


louismrose:
I don't have write perms here, so will wait for somebody else to merge. If there's more to do here, do let me know. Thanks!

nicks:
feels like this needs a unit test, to ensure someone doesn't break it in the future.

i'm really confused why you're hitting this codepath at all, since you're really only supposed to hit this path when you manually specify a host. the default behavior should be to use kubectl's built-in resolution logic. i'm not opposed to this pr, i just worry we're missing a deeper bug here.

nicksieger:
There's some context in the linked issue. Apparently some mac OS machines will return an ipv6 address with a zone index when looking up localhost.

nicks (approved):
OK, i think this is harmless, but posted a separate fix for the broader defaulting issue.

PR #5992, opened 2022-12-20
session: api for graceful recovery in 'tilt ci'

nicks:


nicks:
Note this doesn't add any of the implementation, just the tiltfile stuff

nicksieger (approved):

PR #5996, opened 2022-12-20
controllers: move Session into a reconciler

nicks:
A few reasons to do this:

- Now that we have session settings in the tiltfile, we need a way to propagate data from the tiltfile to the Session

- Reconcilers make it way easier to implement async and wait logic, because they have a bunch of nice queue scheduling underneath.


nicks:
Not much really changed, the controller was already structured a lot like a reconciler. the main bit is moving the Session creation to the tiltfile reconciler.

nicksieger (approved):

Issue #4194, opened 2021-02-12
tilt ci when there are services that crash loop


nicks:
In the #tilt channel, Zack writes:

> is there a way to make tilt run in non interactive mode? I.e. as if I’ve ran tilt up and CTRL+C

> We see tilt ci but we have a few k8s resources that will crash loop a couple times while they are spinning up. We can’t easily set dependencies in the Tiltfile because they have unique names generated on startup.

Currently, `tilt ci` assumes that all services come up cleanly. It exits with failure immediately if any of them crash, and exits with success if all of them are healthy.

But there are lots of teams where the services don't fit this model: maybe they crash and restart while they're waiting for the database to come up. For those cases, you want some way to set a more nuanced "fail-fast" condition than "fail on the first crash". e.g., "fail if it doesn't come up after N sec" or "fail after 3 crashes"

nicks:
For now, you can work around this with a bash script that puts tilt in the background, then use the bash script to wait on your own failure conditions. Here's an example:

https://github.com/tilt-dev/tilt-extensions/blob/master/file_sync_only/test/test.sh

andymartin-sch:
another scenario where something like this would be useful is if you only want `tilt ci` to report success if all services _stay_ healthy for `N` seconds

having `N` be passed in via a command-line argument would be nice

nicks:
I'm going to take a crack at something like:

```
ci_settings(k8s_grace_period='1m')
```

which means "wait at least 1m for a pod to become healthy before failing the ci job"

that will be a good start to let people play around with what values make sense for them.

PR #5997, opened 2022-12-21
session: implementation of ci_settings(k8s_grace_period=[duration])

nicks:
Fixes https://github.com/tilt-dev/tilt/issues/4194

I expect people might want other knobs on this, but this is a minimum viable one.


nicksieger (approved):
LGTM

PR #6001, opened 2022-12-23
prefer user specified registry over discovered

voprak:
Current implementation is checking config map to see if there is registry in the cluster of the current context. 
If you have vm (e.g. in azure) with microk8s cluster and registry addon enabled, then tilt will discover that registry as localhosts:32000 which is correct if you push from localhost, but pushing to the registry from your local dev machine wouldn't work unless you use  VM_IP:32000. This can be configured using default_registry() in tilt but tilt ignores this call.

### DIFF
--- a/internal/container/registry.go
+++ b/internal/container/registry.go
@@ -54,13 +54,14 @@ func RegistryFromCluster(cluster *v1alpha1.Cluster) (*v1alpha1.RegistryHosting, 
 		return nil, fmt.Errorf("cluster not ready: %s", cluster.Status.Error)
 	}
 
-	if cluster.Status.Registry != nil {
-		return cluster.Status.Registry.DeepCopy(), nil
+	if cluster.Spec.DefaultRegistry != nil {
+		// user specified registry
+		return cluster.Spec.DefaultRegistry.DeepCopy(), nil
 	}
 
-	if cluster.Spec.DefaultRegistry != nil {
-		// no local registry is configured for this cluster, so use the default
-		return cluster.Spec.DefaultRegistry.DeepCopy(), nil
+	if cluster.Status.Registry != nil {
+		// discovered registry
+		return cluster.Status.Registry.DeepCopy(), nil
 	}
 
 	return nil, nil
--- a/internal/controllers/core/cluster/reconciler.go
+++ b/internal/controllers/core/cluster/reconciler.go
@@ -334,7 +334,9 @@ func (r *Reconciler) populateK8sMetadata(ctx context.Context, clusterNN types.Na
 
 	if conn.registry == nil {
 		reg := conn.k8sClient.LocalRegistry(ctx)
-		if !container.IsEmptyRegistry(reg) {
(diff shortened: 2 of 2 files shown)


milas:
@voprak The current behavior is intentional - if the cluster is advertising a local registry, that is assumed to be the most performant option. The `default_registry()` function in the Tiltfile (& API via `Cluster.Spec.DefaultRegistry`) is meant as a fallback for clusters without local registries.

I'm not sure if it's currently possible to have Tilt ignore a cluster-advertised registry. Can you give some context on why you want this behavior? There might be a different way to tackle this.

voprak:
> @voprak The current behavior is intentional - if the cluster is advertising a local registry, that is assumed to be the most performant option. The `default_registry()` function in the Tiltfile (& API via `Cluster.Spec.DefaultRegistry`) is meant as a fallback for clusters without local registries.
> 
> https://github.com/tilt-dev/tilt/blob/9c00ce34c62ff0ed322bc923b113fef6ece4baca/internal/container/registry.go#L34-L40
> 
> I'm not sure if it's currently possible to have Tilt ignore a cluster-advertised registry. Can you give some context on why you want this behavior? There might be a different way to tackle this.

@milas, thank you for your reply. This is up to you guys to decide which way to go. In my opinion this should be fully configurable. I've updated description to give some more context around my issue. I might be able to use VM_IP:32000 from cluster and outside by changing configuration of microk8s ingress. Let me try it actually. Will let you know. :-)

nicks:
I'm going to mark this as a duplicate of https://github.com/tilt-dev/tilt/issues/5446. See this comment for how to configure the cluster for this use-case : https://github.com/tilt-dev/tilt/issues/5446#issuecomment-1029452723. 

i think it's important that the cluster config is the source of truth

milas:
For your issue specifically, Tilt detects cluster registries from the `local-registry-hosting` ConfigMap.

For microk8s, it appears that only one hostname is supplied, which is hardcoded as `localhost`: https://github.com/canonical/microk8s-core-addons/blob/1ba24ea3116247219473fe8f26f2ba3023c6f3b4/addons/registry/registry.yaml#L76-L85

There are actually 3 different potential hostname values:
 * `host` - registry host as seen by user when pushing images (this is what you'd like to override to your VM IP)
 * `hostFromContainerRuntime` - hostname as seen by e.g. containerd when pulling the image on a node (e.g. in your case if it's a single-node host, it could be localhost)
 * `hostFromClusterNetwork` - less commonly used, hostname to address registry from a Pod, useful for native, in-cluster building (Tilt does not utilize this)

You could try `kubectl -n kube-public edit configmap local-registry-hosting` and set `host` to the public IP:port, and add `hostFromContainerRuntime` as `localhost:32000`.

Ideally, microk8s could add some flags to `microk8s enable registry` for these to allow you to set them without having to patch the config manually.

voprak:
Thank you guys for your help. I made the changes in confit map and all works now. Much appreciated.

Issue #3783, opened 2020-09-16
default timeout for tilt ci


nicks:
Tilt ci should have a default, configurable timeout (30min)?

I think there should probably be other sensible timeout defaults in Tilt (e.g., if a pod doesn't appear in 5 minutes, it's probably never going to show up. if it's stuck in pending for 10 minutes, it's probably never going to be ready.).

But an overall tilt ci timeout is a good first step

nicks:
note that you can also do this with

```
timeout 30m tilt ci
```

PR #6004, opened 2023-01-05
session: implement ci timeouts, defaulting to 30m

nicks:
Fixes https://github.com/tilt-dev/tilt/issues/3783


nicks:
most of the logic is just getting the data into the right places

since this is a backwards incompatible change, I'm going to make this tilt v0.31.0

nicksieger (approved):

PR #6008, opened 2023-01-06
session: fix a performance regression

nicks:
firing lots of spurious events was leading to some problems


nicks:
grrr annoyed at myself i didn't catch it; thought it was the other stuff i was running that was slowing things down.

nicksieger (approved):

PR #6014, opened 2023-01-10
docker_compose: infer resource deps from compose project

nicksieger:
`depends_on` or other dependencies in the compose file implicitly add a resource dependency, so you won't have to declare duplicate information with `dc_resource(..., resource_deps=...)`.

### DIFF
--- a/internal/engine/upper_test.go
+++ b/internal/engine/upper_test.go
@@ -1979,9 +1979,9 @@ func TestDockerComposeUp(t *testing.T) {
 
 func TestDockerComposeRedeployFromFileChange(t *testing.T) {
 	f := newTestFixture(t)
-	_, m := f.setupDCFixture()
+	r, m := f.setupDCFixture()
 
-	f.Start([]model.Manifest{m})
+	f.Start([]model.Manifest{r, m})
 
 	// Initial build
 	call := f.nextCall()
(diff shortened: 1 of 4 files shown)


nicks (approved):

milas (approved):
🎉

heurisctics123:
Hi. Nice work. But just wanted to point out that this change might break the setup if you have not set up the dependencies properly. The following scenario happened to me today. I do "tilt up" for lets say for service 'x', 'y' and 'z'. In my docker-compose unintentionally someone had written that service 'z' depends on service 'k'. Before this feature my setup would just spin up services x,y,z but now it spins up x,y,z and k. This behavior was not intended or simply saying the behavior has changed and causes confusion.  Since the versioning you have used for the release gives the impression that it is only a patch I would highly recommend to point out this (breaking) change to the users. That's my recommendation. Thanks in advance.

rodonal:
Hi. Nice work. But just wanted to point out that this change might break the setup if you have not set up the dependencies properly. The following scenario happened to me today. I do "tilt up" for lets say for service 'x', 'y' and 'z'. In my docker-compose unintentionally someone had written that service 'z' depends on service 'k'. Before this feature my setup would just spin up services x,y,z but now it spins up x,y,z and k. This behavior was not intended or simply saying the behavior has changed and causes confusion. Since the versioning you have used for the release gives the impression that it is only a patch I would highly recommend to point out this (breaking) change to the users. That's my recommendation. Thanks in advance

Issue #6000, opened 2022-12-22
only attach namespaced labels


nicks:
Tilt's `docker_build` currently attaches builtby=tilt to images. This is used for image pruning.

The OCI spec recommends that all labels should be namespace by the organization that owns them. Ref: https://github.com/opencontainers/image-spec/blob/main/annotations.md

We should change this to some sort of tilt label (dev.tilt.*)

Related: https://github.com/tilt-dev/tilt/issues/4023

PR #6025, opened 2023-01-20
docker: add properly namespaced labels

nicks:
fixes https://github.com/tilt-dev/tilt/issues/6000

note that after this PR, tilt will only prune images with the new label. we could add backcompat code to prune the old images, but i'm not sure it's worth the ongoing performance hit.

### DIFF
--- a/internal/build/container_test.go
+++ b/internal/build/container_test.go
@@ -7,6 +7,7 @@ package build
 import (
 	"bytes"
 	"context"
+	"regexp"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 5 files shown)


milas (approved):
LGTM!

> i'm not sure it's worth the ongoing performance hit.

Is the performance hit here that we'd need to run 2x list queries each time? (Or is image filtering using label namespaces for indexing internally or something?)

[Agreed re: not worrying about old images, just curious here!]

nicks:
ya, i don't think we can do "OR" queries, so we'd have to do two separate queries.

PR #6026, opened 2023-01-20
circleci: update some deps and drop some deprecated linters

nicks:


nicksieger (approved):

nicks:
looking at the history, Go on circleci got a lot slower. maybe disk problems? i moved the 'install' command up-front, which should warm the build cache a bit.

Issue #2995, opened 2020-02-21
trigger resource rebuild from the TUI interface


nicks:
Some users only interact with Tilt through the TUI interface. The big thing that they've noted is missing is a keyboard interface for triggering builds.

Note that you can currently do this from an independent terminal with `tilt trigger`
https://docs.tilt.dev/cli/tilt_trigger.html

PR #6039, opened 2023-02-07
Adds Resource Trigger Hotkey to HUD/TUI

bshore:
## Notes

Aims to resolve https://github.com/tilt-dev/tilt/issues/2995 by introducing hotkey `t` to trigger rebuild on currently selected resource when in HUD / TUI mode.

image


internal/hud/hud.go
```diff
@@ -174,6 +178,22 @@ func (h *Hud) handleScreenEvent(ctx context.Context, dispatch func(action store.
 				escape()
 			case r == 'R': // hidden key for recovering from printf junk during demos
 				h.r.screen.Sync()
+			case r == 't': // [T]rigger resource update
+				_, selected := h.selectedResource()
+				h.recordInteraction("trigger_resource")
+				payload := []byte(fmt.Sprintf(`{"manifest_names":[%q], "build_reason": %d}`, selected.Name, model.BuildReasonFlagTriggerCLI))
+				resp, err := http.Post(fmt.Sprintf("http://%s/api/%s", h.webURL.Host, "trigger"), "application/json", bytes.NewBuffer(payload))
```
bshore, line 185: Note: I tried to use `dispatch()` passing this Action:

https://github.com/tilt-dev/tilt/blob/688f5dfb982ce68de84977dc26e3d12b7f486355/internal/hud/server/actions.go#L7-L10

Like how the server handles it:

https://github.com/tilt-dev/tilt/blob/688f5dfb982ce68de84977dc26e3d12b7f486355/internal/hud/server/server.go#L270

But I encountered `import cycle not allowed`, so I ended up following the example in the linked issue here - https://github.com/tilt-dev/tilt/issues/2894#issuecomment-585866835 to just send a POST

internal/hud/hud.go
```diff
@@ -174,6 +178,22 @@ func (h *Hud) handleScreenEvent(ctx context.Context, dispatch func(action store.
 				escape()
 			case r == 'R': // hidden key for recovering from printf junk during demos
 				h.r.screen.Sync()
+			case r == 't': // [T]rigger resource update
+				_, selected := h.selectedResource()
+				h.recordInteraction("trigger_resource")
+				payload := []byte(fmt.Sprintf(`{"manifest_names":[%q], "build_reason": %d}`, selected.Name, model.BuildReasonFlagTriggerHUD))
+				resp, err := http.Post(fmt.Sprintf("http://%s/api/trigger", h.webURL.Host), "application/json", bytes.NewBuffer(payload))
```
bshore, line 185: Note: I tried to use `dispatch()` passing this Action:

https://github.com/tilt-dev/tilt/blob/688f5dfb982ce68de84977dc26e3d12b7f486355/internal/hud/server/actions.go#L7-L10

Like how the server handles it:

https://github.com/tilt-dev/tilt/blob/688f5dfb982ce68de84977dc26e3d12b7f486355/internal/hud/server/server.go#L270

But I encountered `import cycle not allowed`, so I ended up following the example in the linked issue here - https://github.com/tilt-dev/tilt/issues/2894#issuecomment-585866835 to just send a POST

nicks: how hard would it be to move AppendToTriggerQueueAction to the `internal/store` package? then you could use it here. POST is more of the api for external binaries, and using it here makes this harder to test/maintain.

nicks (approved):
LGTM!

Issue #6055, opened 2023-02-28
tilt explain with no args panics


nicks:
### Current Behavior

```
 tilt explain
panic: runtime error: index out of range [0] with length 0

goroutine 1 [running]:
k8s.io/kubectl/pkg/cmd/explain.(*ExplainOptions).Run(0xc0001bd290)
        /root/project/vendor/k8s.io/kubectl/pkg/cmd/explain/explain.go:131 +0x73d
github.com/tilt-dev/tilt/internal/cli.(*explainCmd).run(0xc0003d0d50, {0x47dbb28, 0xc000537480}, {0x60b3168, 0x0, 0x0})
        /root/project/internal/cli/explain.go:107 +0x1af
github.com/tilt-dev/tilt/internal/cli.addCommand.func1(0xc00033db80?, {0x60b3168, 0x0, 0x0})
        /root/project/internal/cli/cli.go:156 +0xb1
github.com/spf13/cobra.(*Command).execute(0xc00033db80, {0x60b3168, 0x0, 0x0})
        /root/project/vendor/github.com/spf13/cobra/command.go:876 +0x67b
github.com/spf13/cobra.(*Command).ExecuteC(0xc0007f5680)
        /root/project/vendor/github.com/spf13/cobra/command.go:990 +0x3bd
github.com/spf13/cobra.(*Command).Execute(...)
        /root/project/vendor/github.com/spf13/cobra/command.go:918
github.com/spf13/cobra.(*Command).ExecuteContext(...)
        /root/project/vendor/github.com/spf13/cobra/command.go:911
github.com/tilt-dev/tilt/internal/cli.Execute()
        /root/project/internal/cli/cli.go:96 +0xef6
main.main()
        /root/project/cmd/tilt/main.go:19 +0xde

#### About Your Use Case
Note that `tilt explain --help` tells you that it expects 1 argument

PR #6059, opened 2023-03-04
cli: check for args in explain command

bshore:
Aims to resolve https://github.com/tilt-dev/tilt/issues/6055

### DIFF
--- a/internal/cli/explain.go
+++ b/internal/cli/explain.go
@@ -23,7 +23,6 @@ package cli
 
 import (
 	"context"
-	"fmt"
 	"time"
 
 	"github.com/spf13/cobra"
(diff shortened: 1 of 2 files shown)


internal/cli/explain.go
```diff
@@ -95,6 +96,10 @@ func (c *explainCmd) run(ctx context.Context, args []string) error {
 	a.Incr("cmd.explain", cmdTags.AsMap())
 	defer a.Flush(time.Second)
 
+	if len(args) == 0 {
```
nicks, line 99: this is a good fix! but we usually try to keep this consistent with what upstream kubectl is doing. in this case, i think it's missing a call to Validate() after the Complete() call a few lines below this. does that work?
https://github.com/kubernetes/kubectl/blob/master/pkg/cmd/explain/explain.go#L141

bshore: Yeah that's much better, thanks!

nicks (approved):
LGTM!

aw, i wish you had kept the test!

bshore:
> LGTM!
> 
> aw, i wish you had kept the test!

@nicks I wish I could have, too! Unfortunately the upstream `Validate()` function calls `os.Exit()` so there was no easy way to assert/require on any output 😞

Issue #5955, opened 2022-10-19
Ability to merge multiple logs in the UI


dfeyer:
First, thanks for this awesome tool. I have a feature request, when I have a few starred resources, it can be useful to select more than one (double clic, shift clic, or just toggle clic), to be able to have a merged log stream from the selected resources. 

In micro service architures it's nice to be able to watch stream accros a selected resources group, but most of the time "All resources" are too chatty.

gamerson:
I've tried myself to do this (select multiple resources in UI and was hoping for the logs to be multiplexed).  Would be pretty useful it seems.

eliyabar:
I'm using the CLI to get this exact functionality using `tilt logs service_a service_b -f`.
It would be awesome to get this from the UI.

PR #6064, opened 2023-03-08
web: starred resources log pane

bshore:
Aims to resolve https://github.com/tilt-dev/tilt/issues/5955

## Notes
- Adds ResourceName enum for `(starred)`
- When path is `/r/(starred)/overview`, only display logs matching the starred resources.

## Placement
I wasn't sure where to place a link/button for starred resource logs so I put it in two places:
- At the top, always to the left of starred resources
- Under the `All Resources` button/link

image


nicks (approved):
i like this idea! let's do it

Issue #6053, opened 2023-02-25
Failing to restart the container


besarthoxhaj:
I have a very simple app, composed of only two files:

```
.
├── Tiltfile
└── docker-compose.yml

1 directory, 2 files
```

After `tilt up`, I navigate to `http://localhost:10350/r/agent/overview` and manually trigger two updates. During the first update the container exists. Thus, subsequents updates don't execute properly the `command` in compose.

Am I doing something wrong or is this a bug?

![Screenshot 2023-02-25 at 17 43 13](https://user-images.githubusercontent.com/7887496/221371986-6582514c-084f-4a62-9fef-926559973dfa.png)
![Screenshot 2023-02-25 at 17 40 52](https://user-images.githubusercontent.com/7887496/221372152-a0025179-cfa8-421a-b17b-de96f20794f2.png)

#### `tilt doctor` Output
```
Tilt: v0.31.2, built 2023-02-10
System: darwin-amd64
---
Docker
- Host: unix:///var/run/docker.sock
- Server Version: 20.10.17
- API Version: 1.41
- Builder: 2
- Compose Version: v2.6.1
---
Kubernetes
- Env:
- Context:
- Cluster Name: Unknown
- Namespace: default
- Container Runtime: unknown
- Version: Error: invalid configuration: no configuration has been provided, try setting KUBERNETES_MASTER environment variable
- Cluster Local Registry: none
---
Thanks for seeing the Tilt Doctor!
Please send the info above when filing bug reports. 💗

The info below helps us understand how you're using Tilt so we can improve,
but is not required to ask for help.
---
Analytics Settings
--> (These results reflect your personal opt in/out status and may be overridden by an `analytics_settings` call in your Tiltfile)
- User Mode: opt-in
- Machine: 62c457b42b8358fde9219b4639140a8d
- Repo:
```

nicks:
thanks for the report! ya, i can repro this. i think there might be a race condition where the container dies before tilt sees the logs, since this container dies so quickly.

nicks:
I filed a bug upstream about this: https://github.com/docker/compose/issues/10351

i wonder if we should we should update tilt's log streamer to use the docker log api directly, rather than going through docker compose. The `docker compose logs` command has a lot of behavior that makes sense for humans, but has odd edge cases that make it hard for machines to use.

PR #6067, opened 2023-03-09
dockercompose: simplify the log streamer

nicks:
Fixes https://github.com/tilt-dev/tilt/issues/6053


nicks:
now that we have a reconciler, it ended up being easier to track the container id ourselves and stream directly from the docker api. i feel bad that we have so much code working around the subtleties of 'docker compose logs' :grimacing:

nicksieger (approved):
Looks great! Love that we can move this to the go client to fetch the logs.

Issue #5909, opened 2022-08-01
Failure to load extension from gitlab.com


ghost:
Hello :sunny:

1. **I'm trying to** load some extensions from a gitlab.com-repository: https://gitlab.com/opvasger/tilt-extension-experiment
```
v1alpha1.extension_repo(name='tilt-extension-experiment', url='https://gitlab.com/opvasger/tilt-extension-experiment', ref="main")
v1alpha1.extension(name='tilt_extension', repo_name='tilt-extension-experiment', repo_path='tilt_extension')
load('ext://tilt_extension', 'vue_build', 'dotnet_build')
```
**I'm expecting** the definitions to load successfully.

**What happens** is that Tilt fails to load with the following error:
```bash
Loading Tiltfile at: <REDACTED>\tiltfile
Traceback (most recent call last):
  <REDACTED>\tiltfile:3:1: in <toplevel>
Error: cannot load ext://tilt_extension: loading extension repo tilt-extension-experiment: download error: waiting 5s before retrying. Original error: dynamic lookup required to find mapping
```

I hope you can tell me what I'm doing wrong!

Thanks for your awesome work on Tilt - it's really cool, and I appreciate you taking at look at my issue

opvasger:
I'll keep the GitLab-repo open until the issue is resolved - If I can help in any way, let me know!

Thanks for working on tilt :sunny:

nicks:
forked the gitlab repo, ty

regiontog:
I had the same issue and did some digging. Seems like this error occurs when the `repoRootForImportPath` function in go-get fails to identify the kind of vcs system behind the URL. There are regexp cases for a few hosts like github.com but not one for gitlab. There is also a generic one if the URL ends in `.git`. I worked around my issues by using a `.git` URL and ensuring that the domain part includes a `.` since I had originally used a URL like `https://vcs/my/repo/` and fixed it using `https://vcs.my.domain/my/repo.git`.

You can see the [regexp](https://github.com/tilt-dev/go-get/blob/master/vcs.go#L815-L872) here.

@nicks FYI

PR #6070, opened 2023-03-10
vendor: update tilt-dev/go-get

nicks:
adds support for gitlab modules
Pulls in https://github.com/tilt-dev/go-get/pull/13 Fixes https://github.com/tilt-dev/tilt/issues/5909


bshore (approved):
😉

PR #6071, opened 2023-03-10
vendor: use upstream fsevents

nicks:


nicks:
based on a discussion where we determined all our patches have been upstreamed. once this is merged, i'll archive the fork.

nicks:
need to revisit this

milas:
FWIW it's possible we actually have more code in Tilt than necessary now

We were able to strip this when porting some code to Compose: https://github.com/tilt-dev/tilt/blob/ce8d0d674381f7be54bf04a46665c2c3173bc5aa/internal/watch/watcher_darwin.go#L41-L44

Everything else logic-wise is identical in the low-level Darwin watcher code, so I'm guessing that behavior was fixed (we don't ask for history so shouldn't have ever been getting any)

I'll test it out when I get a sec

Issue #5771, opened 2022-05-05
Ability to hide detail sidebar


kevin-lindsay-1:
## Describe the Feature You Want

Sometimes I need to use a single screen for both code and updates. Tilt is the perfect tool to have open alongside my code, but I can't read the logs for the app I'm working on because the real estate of the sidebar is too large.

### Current Behavior

Currently it seems to expect a wide viewing area.

### Why Do You Want This?

To be able to use tilt even when I don't give it a dedicated monitor.

**Additional context**

landism:
Thanks for filing! As a single-monitor user, I definitely empathize.

Have you tried `tilt logs` for this use case? How close does it come to meeting your needs?

kevin-lindsay-1:
I haven't, and I'm sure that would work much better, but I prefer the UI. Seems like a fairly easy button to add, since there's already a sidebar.

bshore:
Hi @nicks @landism @kevin-lindsay-1 !

I started hacking away at this for a bit today and thought I should reach out before going any further. I'm happy to cleanup & push the Proof of Concept example.

<details>
  <summary>Basic Proof of Concept Screenshots</summary>

<table>
  <thead>
    <tr>
      <td>Open</td>
      <td>Collapsed</td>
    </tr>
  </thead>
  <tbody>
    <tr>
      <td><img src="https://user-images.githubusercontent.com/14914028/224513194-4c57e610-6ab9-4f92-820d-99cca72e94a6.png" /></td>
      <td><img src="https://user-images.githubusercontent.com/14914028/224513203-2ee6ce07-d965-4f66-876e-418a3295bdfc.png" /></td>
    </tr>
  </tbody>
</table>
</details>

### Minibar?
I started implementing a _Minibar_ so users can still switch between resource logs and view status while the sidebar is collapsed.  I'd just like to see what everyone's thoughts were before I continue.

- Resources in the minibar show their name & status on hover.
- Group labels will probably just become arrow accordions with their label shown on hover.
- Search may open the sidebar onFocus to give users room to type?
  - onBlur re-collapses if previously collapsed?
  - maybe don't render search bar when collapsed?
<details>
  <summary>Example</summary>

![image](https://user-images.githubusercontent.com/14914028/224513342-60a27a38-159b-40b4-98b8-2ce341e02a45.png)
</details>

PR #6076, opened 2023-03-14
web: sidebar collapse/expand toggle

bshore:
Relates to https://github.com/tilt-dev/tilt/issues/5771#issuecomment-1465038918

This introduces the Proof of Concept collapsible Sidebar mentioned in the linked comment.

### DIFF
--- a/web/src/HUD.tsx
+++ b/web/src/HUD.tsx
@@ -25,6 +25,7 @@ import { ResourceListOptionsProvider } from "./ResourceListOptionsContext"
 import { ResourceNavProvider } from "./ResourceNav"
 import { ResourceSelectionProvider } from "./ResourceSelectionContext"
 import ShareSnapshotModal from "./ShareSnapshotModal"
+import { SidebarContextProvider } from "./SidebarContext"
 import { TiltSnackbarProvider } from "./Snackbar"
 import { SnapshotActionProvider, SnapshotProviderProps } from "./snapshot"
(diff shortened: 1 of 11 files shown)


nicks:
thanks for prototyping this up. a couple UI observations:
- the hamburger menu is pretty dark, usually our icons are white
- there's not any way to tell what service you're on when you're in this mode
- currently the sidebar is draggable, and this feature interacts weirdly with the dragger (see screenshot)

Screenshot from 2023-03-16 18-58-14

one idea we had played around with was using icons instead of names for each service (auto-assigning icons in some deterministic order)

what do you think of triggering the collapsed mode when you drag the sidebar "left" of a current point?

bshore:
* Fixes hamburger menu icon color (I use a dark reader extension so I didn't even notice 😅  )
* Adds a title row when sidebar is collapsed
* Adds a dragger breakpoint to fix the weirdness when expanding, and will auto-collapse when dragging to the left (upon releasing)
  * There's still a little bit of weirdness when expanding from fully collapsed, but once the breakpoint is hit, all resources come back
* Re: Icons - I'd love to give this a shot, maybe not in this PR but in a follow-up?

nicks (approved):
LGTM

i'm not so sure about the expand/collapse icons, but can think more about what they should be

ya, the service icons was kind of a silly idea that i thought would fit in well with some of your earlier explorations

PR #6082, opened 2023-03-24
web: clean up sidebar ux

nicks:
- adjusts the icon used to expand/collapse the sidebar
- makes it inline instead of taking up its own line


nicks:
Screenshot from 2023-03-24 17-57-40
Screenshot from 2023-03-24 17-57-25

Issue #6085, opened 2023-04-04
Extraneous characters on runtime logs with docker compose


klutzer:
### Expected Behavior
It seems Tilt is appending some characters in front of each runtime log line for all containers running with `docker_compose`/`dc_resource` commands. Like the image below:
![image](https://user-images.githubusercontent.com/10657645/229817336-77ff2a53-91d5-488f-b3e9-511351003575.png)

This happens for all containers, another example using localstack container:
![image](https://user-images.githubusercontent.com/10657645/229817652-0cd8ffe3-bfd9-482c-8013-d6f9ca28409b.png)

This is not a docker compose problem, as we can see the logs without those characters running `docker compose logs`:
![image](https://user-images.githubusercontent.com/10657645/229818919-69735f67-2436-4453-babd-c0bf2b9e1c3a.png)

I supposed this behavior is coming from some internal Tilt processing. It's important to say it started to happen after upgrading Tilt to the last version, so in versions lower than `v0.30.12` the problem didn't occur.

### Current Behavior
Described above.

### Steps to Reproduce
1. With the last version of Tilt, configure a Tiltfile with `docker_compose` and `dc_ resource` commands.
2. Go to the UI interface and see the Runtime logs of any container

#### `tilt doctor` Output

```
$ tilt doctor
Tilt: v0.32.0, built 2023-03-13
System: linux-amd64
---
Docker
- Host: unix:///var/run/docker.sock
- Server Version: 23.0.2
- API Version: 1.42
- Builder: 2
- Compose Version: v2.17.2
---
Kubernetes
- Env: 
- Context: 
- Cluster Name: Unknown
- Namespace: default
- Container Runtime: unknown
- Version: Error: invalid configuration: no configuration has been provided, try setting KUBERNETES_MASTER environment variable
- Cluster Local Registry: none
---
Thanks for seeing the Tilt Doctor!
Please send the info above when filing bug reports. 💗

The info below helps us understand how you're using Tilt so we can improve,
but is not required to ask for help.
---
Analytics Settings
--> (These results reflect your personal opt in/out status and may be overridden by an `analytics_settings` call in your Tiltfile)
- User Mode: opt-out
- Machine: 88a4b10099a8eca87f65c1df21c6c005
- Repo: FDpe9se1/r7m8/NJ7B+AhA==
```

#### About Your Use Case
This bug could mislead to false interpretations of containers outputs as we're seeing strange info in front of each log line

nicks:
[...]

PR #6087, opened 2023-04-04
dockercompose: fix bug in log decoding

nicks:
detects when logs are multiplexed and de-multiplexes them

fixes https://github.com/tilt-dev/tilt/issues/6085


nicks:
most of this pr is just threading through the right configs

milas (approved):

PR #6095, opened 2023-04-07
vendor: use upstream fsevents

milas:
See #6071 for original PR - I rebased the dependency update and then removed the code causing trouble on the Tilt side.


nicks (approved):
LGTM!

PR #6108, opened 2023-04-25
Adds additional exclusions for linkerd logs in addition to Istio

anas-domesticus:
Currently Istio container logs are ignored, this adds the same functionality for Linkerd


nicksieger (approved):
👍 Thanks for the contribution!

Issue #6048, opened 2023-02-22
Full update does not re create PVCs


huangzou70:
### Expected Behavior
Force refreshing a tilt resource should recreate everything in the tilt resource

### Current Behavior
Force refreshing a tilt resource with a PVC does not recreate the PVC, resulting in my app to wait indefinitely for the PVC to be recreated. A second force refresh fixes this.

### Steps to Reproduce
1. Set up a k8s deployment that mounts a PVC
2. Define, in the Tilt file something like this: `k8s_resource('myapp', objects=['myapp-pvc:persistentvolumeclaim'])`
3. tilt up
4. Perform a force refresh in the Tilt UI

#### `tilt doctor` Output

```
Huangs-MacBook-Pro-2:seurat zou$ tilt doctor
Tilt: v0.31.1, built 2023-01-09
System: darwin-amd64
---
Docker
- Host: [default]
- Server Version: 
- API Version: 
- Builder: 
- Compose Version: v2.3.3
---
Kubernetes
- Env: gke
- Context: gke_livedesign-development-31333_us-central1_shared-dev-cluster-11
- Cluster Name: gke_livedesign-development-31333_us-central1_shared-dev-cluster-11
- Namespace: dev-zou-darwin-04d442fb
- Container Runtime: containerd
- Version: v1.24.8-gke.2000
- Cluster Local Registry: none
---
Thanks for seeing the Tilt Doctor!
Please send the info above when filing bug reports. 💗

The info below helps us understand how you're using Tilt so we can improve,
but is not required to ask for help.
---
Analytics Settings
--> (These results reflect your personal opt in/out status and may be overridden by an `analytics_settings` call in your Tiltfile)
- User Mode: opt-in
- Machine: bbfb9c78f9450f4c6ace15cd5c20613d
- Repo: GuRVNB7cdKeumnQBjLUROQ==
...
```

#### About Your Use Case
This is made worse by https://github.com/tilt-dev/tilt/issues/5492, since if you want MANUAL_UPDATE resources, they can only be updated via full refreshes.

I suspect the issue here is when tilt refreshes the resource and deletes everything tracked by the tilt resource, the PVCs are deleted but will not go away until the pod referencing the PVC gets deleted. And if the pod takes too long to terminate (it takes about 30 seconds for my pod to terminate), then tilt would try to recreate the PVC while the pod is still going down, and fail at the recreate

I suspect this also affects other k8s objects that are not PVCs

nicks:
[...]

PR #6109, opened 2023-04-26
k8s: adjust how tilt handles two-step deletes

nicks:
before this change, force refresh would delete, then wait for 30s for the delete to finalize. if the delete timeout failed, the force refresh would ignore it and continue applying resources as if nothing had happened.

after this change, force refresh does an asynchronous deletion. When it applies, it checks if the resource is still in the delete phase, and waits for the delete to finish before continuing the apply.

I think this is likely to be a lot more robust and better for parallelization (since now, deletes can happen in parallel with image builds).

Fixes https://github.com/tilt-dev/tilt/issues/6048

### DIFF
--- a/internal/cli/down.go
+++ b/internal/cli/down.go
@@ -206,7 +206,7 @@ func deleteK8sEntities(ctx context.Context, manifests []model.Manifest, updateSe
 	errs := []error{}
 	if len(entities) > 0 {
 		dCtx, cancel := context.WithTimeout(ctx, updateSettings.K8sUpsertTimeout())
-		err = downDeps.kClient.Delete(dCtx, entities, false)
+		err = downDeps.kClient.Delete(dCtx, entities, 0)
 		cancel()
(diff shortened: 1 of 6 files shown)


nicksieger (approved):
What a wild bug. LGTM. I like that just throwing an error on timeout just naturally forces everything to be reconciled later (or so I assume).

andymartin-sch:
you all are amazing! thanks for working on this! 🎉

nicks:
@andymartin-sch don't jinx it, let's make sure it works for you all before celebrating!

andymartin-sch:
ha fair :) we'll test it and let you know

huangzou70:
Thanks for the quick turnaround! Amazing work!

I tried this out and can confirm now either
1) The Tilt resource fails if the PVC takes more than 1 minute to delete or
2) The PVC gets recreated correctly if the PVC is deleted within the 1 minute timeout

PR #6110, opened 2023-05-04
Update ReadRune to support Windows

Onewaysidewalks:
Currently, when using `ReadRune` on windows OSes, the input comes prefixed with 'empty' characters (as described here: https://github.com/mattn/go-tty/issues/22)


nicks:
hi! looks like the test failures are legit but shouldn't be too hard to fix, looks like you need to run `gofmt`:
https://app.circleci.com/pipelines/github/tilt-dev/tilt/12605/workflows/ba262aa2-8cdb-4c11-a3d3-c860245500c7/jobs/94080/tests
happy to accept this once they're fixed.

nicks:
going to close this, since it's stale and we could never get the tests to pass. we can re-open if anyone wants to pick it up.

Issue #6083, opened 2023-03-30
Dockerfile heredocs don't work when using tilt-built base image


lamdor:
### Expected Behavior
The docker build succeeds when using heredoc syntax in Dockerfiles as described in https://www.docker.com/blog/introduction-to-heredocs-in-dockerfiles/

### Current Behavior
When using a base image that is also built by tilt, the docker build fails with a `Build Failed: ImageBuild: unterminated heredoc
` message.

### Steps to Reproduce
I've forked the tilt-avatars project to show the bug:  https://github.com/lamdor/tilt-avatars/tree/tilt-heredoc-bug
Comparison at https://github.com/tilt-dev/tilt-avatars/compare/main...lamdor:tilt-avatars:tilt-heredoc-bug
1. run `tilt up`
2. `api` service fails to build the docker image with error:
```
STEP 1/4 — Loading cached images
     - localhost:5000/base:tilt-66e364f183dd19c8

STEP 2/4 — Building Dockerfile: [tilt-avatar-api]
Building Dockerfile for platform linux/amd64:
  # syntax = docker/dockerfile:1
  FROM localhost:5000/base:tilt-66e364f183dd19c8
  
  ENV PIP_DISABLE_PIP_VERSION_CHECK "1"
  
  WORKDIR /app
  COPY api/requirements.txt ./
  RUN --mount=type=cache,target=/root/.cache/pip pip install -r requirements.txt
  
  
  COPY api/ ./api/
  
  ENV FLASK_APP ./api/app.py
  ENV FLASK_ENV development
  
  RUN <<EOF
  ECHO 
  EOF 
  
  CMD ["flask", "run", "--host", "0.0.0.0"]


     Building image
     resolve image config for docker.io/docker/dockerfile:1 [done: 573ms]
     docker-image://docker.io/docker/dockerfile:1@sha256:39b85bbfa7536a5feceb7372a0817649ecb2724562a38360f4d6a7782a409b14 11.56MB / 11.56MB [done: 386ms]

Build Failed: ImageBuild: unterminated heredoc
```

#### `tilt doctor` Output

```
Tilt: v0.31.2, built
System: darwin-arm64
---
Docker
- Host: http://docker.example.com
- Server Version: 20.10.21
- API Version: 1.41
- Builder: 2
- Compose Version: v2.16.0
---
Kubernetes
- Env: kind
- Context: kind-kind
- Cluster Name: kind-kind
- Namespace: default
- Container Runtime: containerd
[...]

PR #6114, opened 2023-05-10
dockerfile: format heredocs correctly

nicks:
fixes https://github.com/tilt-dev/tilt/issues/6083

### DIFF
--- a/internal/dockerfile/ast.go
+++ b/internal/dockerfile/ast.go
@@ -187,22 +187,20 @@ func (a AST) Print() (Dockerfile, error) {
 			currentLine++
 		}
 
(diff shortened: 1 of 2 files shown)


milas (approved):

milas:
Somehow I don't think I ever realized Tilt was actually parsing `Dockerfile`s into ASTs before!

Btw at some point, we could actually get rid of all this [for BuildKit] and use extra named build contexts directly.)

nicks:
ya, we inject our immutable image tags into the dockerfile, similar to how we do it with kube manifests.

nicks:
we also have to read out what images a Dockerfile depends on so we know what order to build things in.

Issue #6133, opened 2023-06-04
ImagePullPolicy is always set to Never


realtimetodie:
### Expected Behavior
The ImagePullPolicy should never be set to Never

### Current Behavior
The ImagePullPolicy is always set to Never

### Steps to Reproduce
When using minikube with the registry addon enabled, the ImagePullPolicy is always set to Never when using custom builds with Bazel remote execution, where the image is pushed to the minikube registry without docker using [rules_oci](https://github.com/bazel-contrib/rules_oci).

https://github.com/tilt-dev/tilt/blob/63831a8cf315784b7248d58af38ec824fcfcb0cd/internal/controllers/core/kubernetesapply/reconciler.go#L504-L506

Tiltfile

```
custom_build(
    ...
    skips_local_docker=True,
```

Error

```
Error: ErrImageNeverPull
```

#### `tilt doctor` Output
```
$ tilt doctor
Tilt: v0.32.4, built 2023-05-24
System: linux-amd64
---
Docker (cluster)
- Host: tcp://192.168.49.2:2376
- Server Version: 23.0.2
- API Version: 1.42
- Builder: 2
---
Docker (local)
- Host: unix:///var/run/docker.sock
- Server Version: 24.0.2
- Version: 1.43
- Builder: 2
- Compose Version: v2.18.1
---
Kubernetes
- Env: minikube
- Context: minikube
- Cluster Name: minikube
- Namespace: default
- Container Runtime: docker
- Version: v1.26.5
- Cluster Local Registry: none
---
Thanks for seeing the Tilt Doctor!
Please send the info above when filing bug reports. 💗

The info below helps us understand how you're using Tilt so we can improve,
but is not required to ask for help.
---
Analytics Settings
--> (These results reflect your personal opt in/out status and may be overridden by an `analytics_settings` call in your Tiltfile)
- User Mode: opt-in
- Machine: 347b8107501235dc490b664d4356fcb9
- Repo:
...
```

nicks:
thanks for the report!

The way most people use Minikube with the Docker socket is that they build directly to the runtime, like this:
https://minikube.sigs.k8s.io/docs/commands/docker-env/
This is how Tilt's docker_build interacts with Minikube. In that case, most people want to set ImagePullPolicy: Never, to ensure you always get the locally built image.

In your case, you're NOT building to the runtime, you're using `skips_local_docker` to build to a registry. We probably need to thread that through the system, so that it knows you specifically configured your builder to build somewhere else.

[...]

PR #6134, opened 2023-06-04
fix: never set the image pull policy to never

realtimetodie:
Fixes an issue where the ImagePullPolicy was set to Never for no apparent reason. This was originally introduced in https://github.com/tilt-dev/tilt/pull/5786.

Related

- https://github.com/tilt-dev/tilt/issues/6133


nicks:
hmmm, this isn't quite right, let's discuss on the issue. 

also -- not sure why you think this was introduced in https://github.com/tilt-dev/tilt/pull/5786, ImagePullPolicy: Never was like one of the very first features we added to Tilt because it's unusable without it in many cases AFAIR; here's a reference to when we were talking about it in 2018! https://github.com/tilt-dev/tilt/pull/124

Issue #6129, opened 2023-06-01
Tilt incorrectly excludes some files with a whitelist-based dockerignore


jwhitaker-swiftnav:
### Expected Behavior
I have an app that looks like
```
├── apps
│   └── myapp
│       ├── build
│       │   └── Dockerfile
│       ├── internal
│       │   └── somemodule
│       │       └── somemodule.go
│       ├── k8s
│       │   ├── deployment.yaml
│       │   ├── kustomization.yaml
│       │   └── namespace.yaml
│       └── main.go
├── go.mod
├── go.sum
└── Tiltfile
└── .dockerignore
```

a .dockerignore that looks like
```.gitignore
*

!apps/myapp/**/*.go

!go.mod
!go.sum
```

and a Dockerfile that looks like
```Dockerfile
FROM docker.io/golang:1.20

RUN apt update && apt install -y tree

WORKDIR /usr/src/app

COPY go.mod go.sum .
RUN go mod download

COPY . .
RUN tree
RUN go build ./apps/myapp

RUN go run ./apps/myapp
```

When I run
` docker build -f apps/myapp/build/Dockerfile .`,  **the app builds its image and runs correctly,**

### Current Behavior

However, in Tilt, the image build fails because files in `apps/myapp/internal` are not included, even though they should be.
Tilt gives output (i know cached steps are in the build sorry)
```

     Building image
     [background] read source files 387B [done: 48ms]
     [1/9] FROM docker.io/library/golang:1.20@sha256:690e4135bf2a4571a572bfd5ddfa806b1cb9c3dea0446ebadaf32bc2ea09d4f9 [done: 39ms]
     [6/9] COPY . . [cached]
     [5/9] RUN go mod download [cached]
     [4/9] COPY go.mod go.sum . [cached]
     [3/9] WORKDIR /usr/src/app [cached]
     [2/9] RUN apt update && apt install -y tree [cached]
     [7/9] RUN tree -a
       → .
       → |-- apps
       → |   `-- myapp
       → |       `-- main.go
       → |-- go.mod
       → `-- go.sum
       → 
       → 2 directories, 3 files
     [7/9] RUN tree -a [done: 340ms]
     [8/9] RUN go build ./apps/myapp
       → apps/myapp/main.go:6:2: no required module provides package github.com/jwhitaker-swiftnav/tiltbug/apps/myapp/internal/somemodule; to add it:
[...]

PR #6140, opened 2023-06-09
dockerignore: tweak the directory-skipping optimization

nicks:
fixes https://github.com/tilt-dev/tilt/issues/6129


nicks:
we need to better reconcile this with how Compose + Buildkit do directory skipping upstream. it will hit Compose too, since skipping directories saves a lot of inotify nodes.

milas (approved):

Issue #6125, opened 2023-05-25
missing provenance for fuqamtlkfg8ef41sv5ivh9bn6


nicks:
We've received a report of buildkit on Docker 24 returning this error when running under tilt:

```
Build Failed: ImageBuild: missing provenance for fuqamtlkfg8ef41sv5ivh9bn6
```

we haven't been able to reproduce the problem. buildkit-team pointed us at https://github.com/moby/buildkit/issues/3562 but i can't think of a way you could even tell tilt to build multi-arch image.

filing for now to see if we get other reports and/or a repro case.

nicks:
OK, i think i understand the basic bug:
- The Docker CLI ensures that you pass in a directory as a context. If you try to pass in a directory that doesn't exist, you will get an error `ERROR: unable to prepare context: path X not found`
- Tilt does not work this way. If you try to pass in a directory that doesn't exist to tilt, it will build fine. It treats that as an non-existent context.
- As of Docker 24, Buildkit has a regression where if you pass in an non-existent build context, you get cryptic provenance errors.

In the short term, you can workaround this by making sure that docker_build(tag, context) always puts at a real directory.

in the medium term, I think we should change the tilt build behavior to more closely match the docker cli behavior:
- passing in a non-existent directory should be an error
- passing "-" as the context should give you an empty context

nicks:
filed upstream as https://github.com/moby/buildkit/issues/3928

Issue #3897, opened 2020-10-30
empty docker_build context


nicks:
Not all docker_build commands take a build context. 

For example, the restart_process wrapper in https://github.com/tilt-dev/tilt-extensions/tree/master/restart_process doesn't need any input files.

It would be nice if there was some way to specify this in docker_build

In the CLI, you'd do something like:

docker build - < Dockerfile

so maybe it should be

docker_build(img, '-', dockerfile=...)

gebinic:
@nicks I tried this with `docker_build('test', '-', dockerfile_contents='...')` but I got an error:

`Build Failed: reading build context: stat /home/blah/tilt/-: no such file or directory` <- note the hyphen between the absolute path and the colon.

The "dockerimages" api resource for "test" also has as context "/home/blah/tilt/-" and not the temp directory created by tilt.

I'm using tilt v0.37.0 on Debian Trixie. Is this really a bug in tilt or did I miss something?

PR #6141, opened 2023-06-09
build: handle missing build contexts

nicks:
Before this change, if you specified a directory that doesn't exist as your docker_build context, you would get an empty build context.

After this change, if you specify a directory that doesn't exist, you get a build-time error. You can use "-" to specify an empty build context.

this is a breaking change, but i think makes us more consistent with how the docker cli works.

fixes https://github.com/tilt-dev/tilt/issues/3897 
fixes https://github.com/tilt-dev/tilt/issues/6125


internal/build/docker_builder.go
```diff
@@ -206,6 +206,27 @@ func (d *DockerBuilder) BuildImage(ctx context.Context, ps *PipelineState, refs
 func (d *DockerBuilder) buildToDigest(ctx context.Context, spec v1alpha1.DockerImageSpec, filter model.PathMatcher, allowBuildkit bool) (digest.Digest, []v1alpha1.DockerImageStageStatus, error) {
 	var contextReader io.Reader
 
+	buildContext := spec.Context
+
+	// Treat context: "-" as an empty context.
+	if buildContext == "-" {
+		emptyContextDir, err := os.MkdirTemp("", "tilt-dockercontext-")
```
milas, line 213: Is there no direct way to tell this to the builder? 🙈

milas (approved):

internal/build/docker_builder.go
nicks: i think there are ways to do it but they're fairly complex to invoke? since the api is more optimized for the case where you're uploading a tarball.

PR #6144, opened 2023-06-15
compose: use --remove-orphans by default

nicks:
`--remove-orphans` ensures that if you remove a service from the compose file, tilt will clean it up
on the next `tilt up`. I think this better aligns the Compose+Tilt behavior with the Kubernetes+Tilt behavior.


internal/dockercompose/client.go
```diff
@@ -128,7 +128,7 @@ func (c *cmdDCClient) Up(ctx context.Context, spec v1alpha1.DockerComposeService
 	c.mu.Lock()
 	defer c.mu.Unlock()
 	runArgs := append([]string{}, genArgs...)
-	runArgs = append(runArgs, "up", "--no-deps")
+	runArgs = append(runArgs, "up", "--no-deps", "--remove-orphans")
 	// Omit --no-build for now to get v2 working.
 	// https://github.com/docker/compose/issues/8785
```
nicksieger, line 133: I wonder if this can be cleaned up now too?

nicksieger (approved):
LGTM

internal/dockercompose/client.go
nicks: ooh good point! done

PR #6147, opened 2023-06-23
vendor: update to k8s 1.27

nicks:
The most impactful change is how status fields are handled in the fake client, so i needed to do a bunch of test updates. i think the new behavior is more correct!


nicks:
ugh, i also need to upgrade to go 1.20 :(

milas (approved):

nicksieger:
maybe go 1.20 would be easier to do first?

nicksieger (approved):

PR #6148, opened 2023-06-23
circleci: use a go 1.20 release image

nicks:
Before this change, we did all our own cross-compilation.

After this change, we pull the cross-compiling logic from a shared image. Loosely based on the Compose approach.

Does not yet switch to go 1.20 by default.


nicksieger (approved):
🤞

PR #6161, opened 2023-07-05
web: show the build tooltip over the build bar, not just the status icon

nicks:
the tooltip often shows the names of resources that the build is waiting on, so it's useful to make the hover target a bit bigger.

Screenshot from 2023-07-05 16-57-40

### DIFF
--- a/web/src/SidebarIcon.tsx
+++ b/web/src/SidebarIcon.tsx
@@ -17,8 +17,7 @@ import { ResourceStatus } from "./types"
 
 type SidebarIconProps = {
   status: ResourceStatus
-  alertCount: number
-  tooltipText: string
+  tooltipText?: string
 }
 
 let SidebarIconRoot = styled.div`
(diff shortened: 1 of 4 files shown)


web/src/SidebarIcon.tsx
```diff
@@ -91,23 +90,16 @@ export default class SidebarIcon extends PureComponent<SidebarIconProps> {
...
+      <SidebarIconRoot
+        className={`${ClassNameFromResourceStatus(this.props.status)}`}
+      >
+        {icon}
+      </SidebarIconRoot>
     )
+    if (!this.props.tooltipText) {
+      return content
+    }
+    return <Tooltip title={this.props.tooltipText}>{content}</Tooltip>
```
landism, line 103: I'm pretty sure MUI's Tooltip bows out and just returns its children when `title` is empty, which would save us from this whole dance. I don't know if we failed to match that behavior in Tilt for some reason or were just unaware when we wrote this code.

This PR is an improvement either way, though.

landism (approved):

web/src/SidebarIcon.tsx
nicks: oops good point!

Issue #6160, opened 2023-07-03
docker_build - only parameter is not working as expected


matanagazan:
### Expected Behavior
`only` parameter should ignore all files except some files that do not change when I run `go generate`

### Current Behavior
when I run go generate, I get this error
`LiveUpdate "bionic-dashboard:localhost_dashboard" UpdateStopped: Found file(s) not matching any sync (files: [/Users/matanagazan/Desktop/bionic/bionicstork])`

`go generate` does not change any of files that I watch with only parametes

folder structure:
```
bionicstork/
├── config
├── relay-execution-files
├── scripts
└──tilt/
   ├── Tiltfile
   ├── build/
   ├── dockerfiles/
   ├── scripts/
   └── version.txt
```
docker_build declaration:
```
docker_build_with_restart(
    ref="localhost/dashboard".,
    dockerfile="dockerfiles/dashboard.dockerfile",
    context="../",
    build_args={"DLV_PORT": 35761},
    target="dashboard",
    entrypoint=["/startup.sh", "app"],
    only=[
      "tilt/build/migrate-cmd",
      "tilt/build/dashboard",
      "tilt/build/healthcheck",
      "scripts/startup/startup.sh",
      "relay-execution-files/HelmChart",
      "tilt/build/collector-binaries",
      "tilt/build/integration-tasks-executor",
      "config/certificate/csr.config",
      "tilt/version.txt",
    ],
    live_update=[
      sync("./build/migrate-cmd", "/migrate-cmd"),
      sync("./build/dashboard", "/app"),
      sync("./build/healthcheck", "/healthcheck"),
      sync("../scripts/startup/startup.sh", "/startup.sh"),
      sync("../relay-execution-files", "/build/"),
      sync("./build/collector-binaries", "/build/"),
      sync("./build/integration-tasks-executor", "/build/integration-tasks-executor-binary/"),
      sync("version.txt", "/version.txt"),
    ],
  )
```
dockerfile
```
# syntax=docker/dockerfile:1

# Stage for compiling go services before copying binaries into images
FROM golang:1.20-alpine AS compile-dashboard
RUN mkdir /app
WORKDIR /app

RUN mkdir certificate
WORKDIR /app/certificate
RUN apk update && apk add \
  openssl \
  curl
COPY config/certificate/csr.config ./csr.config
RUN openssl genrsa -des3 -passout pass:abdefghijk -out server.pass.key 2048 && \
[...]

PR #6163, opened 2023-07-07
watch: ignore the watched dir on all platforms

nicks:
before this change, we would ignore on linux and windows, but emit events on linux

fixes https://github.com/tilt-dev/tilt/issues/6160

### DIFF
--- a/internal/watch/watcher_darwin.go
+++ b/internal/watch/watcher_darwin.go
@@ -38,9 +38,9 @@ func (d *darwinNotify) loop() {
 				e.Path = filepath.Join("/", e.Path)
 
 				_, isPathWereWatching := d.pathsWereWatching[e.Path]
-				if e.Flags&fsevents.ItemIsDir == fsevents.ItemIsDir && e.Flags&fsevents.ItemCreated == fsevents.ItemCreated && isPathWereWatching {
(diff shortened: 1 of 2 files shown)


internal/watch/watcher_darwin.go
```diff
@@ -38,7 +38,7 @@ func (d *darwinNotify) loop() {
 				e.Path = filepath.Join("/", e.Path)
 
 				_, isPathWereWatching := d.pathsWereWatching[e.Path]
-				if e.Flags&fsevents.ItemIsDir == fsevents.ItemIsDir && e.Flags&fsevents.ItemCreated == fsevents.ItemCreated && isPathWereWatching {
+				if e.Flags&fsevents.ItemIsDir == fsevents.ItemIsDir && isPathWereWatching {
 					// This is the first create for the path that we're watching. We always get exactly one of these
 					// even after we get the HistoryDone event. Skip it.
```
milas, line 43: nit: this comment is misleading now, I'd delete it

milas (approved):

internal/watch/watcher_darwin.go
nicks: updated!

Issue #6177, opened 2023-07-23
Upgrading from 0.33.1 to 0.33.2 causes ImageBuild timeout 


saweber:
### Expected Behavior
[This sample repository] should successfully build and deploy my example application on v0.33.2

### Current Behavior
Attempting to build the container times out.

```STEP 1/5 — Building Dockerfile: [example-go-image-tilt_docker_build_with_restart_base]
Building Dockerfile for platform linux/arm64:
  FROM golang:alpine as dlv-env
  
  RUN go install github.com/go-delve/delve/cmd/dlv@latest
  
  FROM alpine:latest
  
  WORKDIR /app
  
  COPY --from=dlv-env /go/bin/dlv /go/bin/dlv
  
  ADD main main
  
  EXPOSE 8000 40000
  ENTRYPOINT ["/go/bin/dlv", "--listen=:40000", "--headless=true", "--api-version=2", "--accept-multiclient", "--log", "exec", "/app/main"]

     Building image
     
     ERROR IN: [internal] load .dockerignore
     
     ERROR IN: [internal] load build definition from Dockerfile

Build Failed: ImageBuild: failed to read dockerfile: no active session for k5b7eah03710pr8w1qql0dpbb: context deadline exceeded
```

### Steps to Reproduce

1. Run `tilt up` in the root of this sample repository: [This sample repository](https://github.com/saweber/go-tilt-hello-world)
2. With 0.33.2, I am seeing the above error.
3. Downgrading to 0.33.1 resolves the issue.
4. Building and running the container in docker with: `docker build -t example-go-image:test .` and `docker run -p 8000:8000 -p 40000:40000 example-go-image:test` also work.

#### `tilt doctor` Output

```
> tilt doctor
Tilt: v0.33.2, built 2023-07-19
System: darwin-arm64
---
Docker
- Host: unix:///Users/stevenweber/.docker/run/docker.sock
- Server Version: 24.0.2
- API Version: 1.43
- Builder: 2
- Compose Version: v2.19.1
---
Kubernetes
- Env: kind
- Context: kind-kind
- Cluster Name: kind-kind
- Namespace: default
- Container Runtime: containerd
- Version: v1.27.3
- Cluster Local Registry: none
---
Thanks for seeing the Tilt Doctor!
Please send the info above when filing bug reports. 💗

The info below helps us understand how you're using Tilt so we can improve,
but is not required to ask for help.
---
Analytics Settings
[...]

PR #6178, opened 2023-07-24
docker: use errgroup for build session error handling

nicks:
we really should convert this code to vendor buildx directly like Compose does, but for now this makes the implementations closer together.

### DIFF
--- a/internal/docker/client.go
+++ b/internal/docker/client.go
@@ -29,6 +29,7 @@ import (
 	"github.com/moby/buildkit/session/auth/authprovider"
 	"github.com/moby/buildkit/session/filesync"
 	"github.com/pkg/errors"
+	"golang.org/x/sync/errgroup"
 
 	"github.com/tilt-dev/tilt/internal/container"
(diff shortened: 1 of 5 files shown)


nicks:
related to https://github.com/tilt-dev/tilt/issues/6177, this makes sure that the error message gets propagated out when session creation fails

internal/docker/client.go
```diff
@@ -332,8 +333,8 @@ func (c *Cli) startBuildkitSession(ctx context.Context, key string, syncedDirs [
 		dialSession := func(ctx context.Context, proto string, meta map[string][]string) (net.Conn, error) {
 			return c.Client.DialHijack(ctx, "/session", proto, meta)
 		}
-		_ = session.Run(ctx, dialSession)
-	}()
+		return session.Run(ctx, dialSession)
```
nicksieger, line 336: All that errgroup plumbing for this one error capture!

nicksieger (approved):

internal/docker/client.go
nicks: yep :grimacing:

Issue #6181, opened 2023-07-25
Please support skip_crds options


kahirokunn:
## Describe the Feature You Want

add skip_crds options

### Current Behavior

Always install crds.
https://github.com/tilt-dev/tilt/blob/15349a7bf1d0aef77726bba4232e2289f5a30bc4/internal/tiltfile/files.go#L255

### Why Do You Want This?

Crds may be included in the helm chart and you often do not want to install it.

**Additional context**

kahirokunn:
I made PR.
https://github.com/tilt-dev/tilt/pull/6182

kahirokunn:
done

PR #6182, opened 2023-07-25
chore: add skip_crds parameter to helm functions

kahirokunn:
https://github.com/tilt-dev/tilt/issues/6181

### DIFF
--- a/internal/tiltfile/files.go
+++ b/internal/tiltfile/files.go
@@ -196,6 +196,7 @@ func (s *tiltfileState) helm(thread *starlark.Thread, fn *starlark.Builtin, args
 	var valueFiles value.StringOrStringList
 	var set value.StringOrStringList
 	var kubeVersion string
+	var skip_crds bool
 
 	err := s.unpackArgs(fn.Name(), args, kwargs,
 		"paths", &path,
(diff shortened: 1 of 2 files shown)


internal/tiltfile/files.go
```diff
@@ -196,6 +196,7 @@ func (s *tiltfileState) helm(thread *starlark.Thread, fn *starlark.Builtin, args
 	var valueFiles value.StringOrStringList
 	var set value.StringOrStringList
 	var kubeVersion string
+	var skip_crds string
```
nicks, line 199: shouldn't this be a boolean?

nicks:
thanks! can you also add this to the API documentation? it's here:
https://github.com/tilt-dev/tilt/blob/master/internal/tiltfile/api/__init__.py#L724

kahirokunn:
@nicks done. how about now?

internal/tiltfile/files.go
kahirokunn: done

nicks (approved):
LGTM!

kahirokunn:
@nicks hello! May I ask when you plan to release it?

kahirokunn:
Okay. Thx

Issue #2487, opened 2019-11-04
tilt errors with `watcher: too many open files`


nicks:
From Kubernetes slack:

Leaving Tilt up overnight leaves it busted with "too many open files" errors.

```
Init:0/1 • OK       (0.7s) •  20h ago 
    HISTORY: FIRST BUILD                                                       OK       (0.7s) •  20h ago 
    K8S POD: [redacted]-85677675dd-gq84w                                             1 restart • AGE 240h 
    URL: http://localhost:2525/                                                                           
    ERROR:                                                                                                
       watcher: too many open filesfailed to create fsnotify watcher: too many open filesfailed to create 
        fsnotify watcher: too many open filesfailed to create fsnotify watcher: too many open filesfailed 
        to create fsnotify watcher: too many open filesfailed to create fsnotify watcher: too many open   
       filesfailed to create fsnotify watcher: too many open filesfailed to create fsnotify watcher: too  
       many open filesfailed to create fsnotify watcher: too many open filesfailed to create fsnotify     
       watcher: too many open files  
```

I think we've seen this before but it seems very rare, and haven't come up with repro steps

landism:
I hit this by running yarn storybook in a k8s node overnight on GKE.

I downloaded inotify-consumers from [here](https://unix.stackexchange.com/a/502812) and sshed into the GKE node itself and saw:
```
# sysctl fs.inotify.max_user_instances
fs.inotify.max_user_instances = 128
# sysctl fs.inotify.max_user_watches
fs.inotify.max_user_watches = 12288
# sh ./inotify-consumers

   INOTIFY
   WATCHER
    COUNT     PID     CMD
----------------------------------------
    6843   146308  /usr/bin/node --max-old-space-size=2048 /app/snapshot_frontend/web/node_modules/fork-ts-checker-webpack-plugin/lib/servi
    4971   146144  /usr/bin/node /app/snapshot_frontend/web/node_modules/.bin/start-storybook -p 9009 -s public --quiet
     402     1008  /home/kubernetes/bin/kubelet --v=2 --cloud-provider=gce --experimental-check-node-capabilities-before-mount=true --allow
```

lianmakesthings:
Been running into the same issue now. I usually never leave Tilt running for extended periods of time. Only when I actively work on something (and I shut down my computer, so I'm also not leaving it running accidentally)
I run into it on the tilt.build project
```
[...]

PR #6193, opened 2023-08-15
podlogstream: handle kubelet fsnotify failures

nicks:
fixes https://github.com/tilt-dev/tilt/issues/2487


nicks:
i've been trying to track down this bug for years and i finally got it!!!

milas (approved):
> // Kubernetes has a bug where it will dump certain kinds of errors to the
> // pod log stream.

fuuuuun

Issue #6192, opened 2023-08-13
Accept env vars to `custom_build`


dnephin:
### Problem description

Hello! I work on a tilt setup that uses `custom_build` and is used across many teams. We've noticed that when the `custom_build` fails and someone reaches out for help they may share the wrong part of the output. I'd like to make it easier for users to debug custom build failures themselves, and also make it easier for them to identify and share the relevant part of the output when they do need help. 

### Current Behavior

Today, the output from our `custom_build` looks something like this when it fails:

>  ... many lines of build output from `docker buildx` ...
> ... one or more lines from `docker buildx` that identify the failure ...  
>
> error: failed to solve: executor failed running [/bin/sh ... *some build step here that may end at the line break*]: exit code: 1
> /Users/username/src/project/scripts/custom-build.zsh --flag-1 /Users/username/src/project2/file3 --dockerfile /tmp/asdf0001/Dockerfile --dest-image-repo registry.internal.net/project/imagename --flag2 more-long-values-here --flag-3 /Users/username/src/project3/.config/imagerefs/service2 --flag-4 yet-more-configuration --flag-5 true --flag-6 false exited with exit code 1
>
> Build Failed: Custom build "/Users/username/src/project/scripts/custom-build.zsh --flag-1 /Users/username/src/project2/file3 --dockerfile /tmp/asdf0001/Dockerfile --dest-image-repo registry.internal.net/project/imagename --flag2 more-long-values-here --flag-3 /Users/username/src/project3/.config/imagerefs/service2 --flag-4 yet-more-configuration --flag-5 true --flag-6 false" failed: exit status 1

It is common for Go programs to print the context about an error following an `error:` line. In the output above the error line from `docker buildx` is immediately followed by a log line from `tilt` that prints the command line. The custom build script includes many command line arguments with long values. Together this block of text can look more significant than the failure details above the error. In practice the values we pass to this script are rarely relevant to the failure.

### Proposed solution

I'm hoping to address the problem with two changes.

I'd like to use environment variables instead of command line flags to pass values to the custom build script. I have found that in general environment variables are easier when the command is being run by code (where as command line flags are easier for humans). The env vars are easier to consume in shell scripts, removing the need to parse command line flags. The environment variables are not printed in the failure message today, so `tilt` could continue to print the full command, but the output would be simply `custom-build.zsh exited with exit code 1`. Those variables can be printed at the start of the script for debugging.

[...]

PR #6202, opened 2023-08-20
Add env argument to custom_build

dnephin:
Implementation of proposal 1 from #6192

Adds an `env = {}` argument to `custom_build`. Adds test coverage to the starlark parsing and custom builder. I also did a manual test to show the env vars are passed from starlark to the script. I believe I've updated the reference docs as well.


pkg/apis/core/v1alpha1/cmdimage_types.go
```diff
@@ -69,6 +69,16 @@ type CmdImageSpec struct {
...
 
+	// Additional environment variables for the build process.
+	//
+	// Expressed as a C-style array of strings of the form ["KEY1=VALUE1", "KEY2=VALUE2", ...].
+	//
+	// Environment variables are layered on top of the environment variables
+	// that Tilt runs with.
+	//
+	// +optional
+	Env []string `json:"env,omitempty" protobuf:"bytes,10,rep,name=env"`
```
dnephin, line 80: Do I need to run any code generation for this new argument? I don't think I updated the docs for this new field.

internal/build/custom_builder_test.go
```diff
@@ -270,11 +271,12 @@ func TestEnvVars(t *testing.T) {
 	sha := digest.Digest("sha256:11cd0eb38bc3ceb958ffb2f9bd70be3fb317ce7d255c8a4c3f4af30e298aa1aab")
 	f.dCli.Images["localhost:1234/foo_bar:tilt-build-1551202573"] = types.ImageInspect{ID: string(sha)}
 	cb := f.customBuild(strings.Join(script, "\n"))
+	cb.Env = []string{"EXTRA=value"}
 	_, err := f.Build(refSetWithRegistryFromString("foo/bar", TwoURLRegistry), cb, nil)
 	require.NoError(t, err)
 }
 
-func TestEnvVars_ConfigRefWithLocalRegistry(t *testing.T) {
+func TestCustomBuildEnvVars_ConfigRefWithLocalRegistry(t *testing.T) {
```
dnephin, line 279: I renamed these two test cases to match the naming convention in the rest of the file.

pkg/apis/core/v1alpha1/cmdimage_types.go
dnephin: The CI check caught some missing code gen and pointed me at the right command to run, I think.

Issue #6145, opened 2023-06-21
Feat: Specify the working directory for a custom build


DWSR:
## Describe the Feature You Want
I'd like to be able to specify the working directory when using a `custom_build`, similar to `local`

### Current Behavior
I can work around the lack of this feature by doing `cd dir && foo`.

### Why Do You Want This?
This addition would help bring `custom_build` into parity with `local` and eases some of the pain felt when trying to share `custom_build` using Tilt extensions.

**Additional context**
N/A

PR #6210, opened 2023-08-30
tiltfile: add custom_build(dir)

nicks:
sets the current working directory for the custom_build command

fixes https://github.com/tilt-dev/tilt/issues/6145


nicks:
mostly added because we also accepted a contribution for custom_build(env)

landism (approved):

Issue #6201, opened 2023-08-18
Resource group collapsed by default 


nicks:
## Describe the Feature You Want
Collapse resource groups by default

### Why Do You Want This?

If you have a lot of resources, tilt let's you group them by label.

By default, all groups are expanded. But presumably if you added labels, it's because the resource list is getting unwieldy!

**Additional context**
Not totally sure this is a good idea, posting to collect feedback.

Also could make sense to make it some sort of an option. I like that solution less, because it might be confusing how options interact.

dnephin:
The main benefit for me would be in `tilt snapshot view`. We use the snapshots to debug CI runs, so the url of the snapshot is frequently different. I think because the URL is different anything saved in local storage (starred resources, etc) are not preserved each time a new snapshot is opened. It would save a bunch of time if every time we opened a snapshot it could open to the detail view (`/r/(all)/overview`) with everything collapsed. I think we can link to `/r/(all)/overview` directly, we're just missing some way to auto collapse the resource groups. If we could link to `/r/(all)/overview?collapsed=true` , or `/r/(all)/overview#collapsed=true` that would work great. If we could use that same query parameter or fragment with the `tilt up` UI that would be really nice as well.

oleghind:
Sorting and collapsing groups is really necessary

Yordi:
Hello,
Is it possible to access (`/r/(all)/overview`) by default when opening the UI? (For example, with a `default-path` variable...)

PR #6212, opened 2023-08-30
web: collapse sections by default

nicks:
If you have a lot of resources, tilt let's you group them by label.

By default, all groups are expanded.
But presumably if you added labels, it's because the resource list is getting unwieldy! Let's try collapsing by default and see if people complain.

https://github.com/tilt-dev/tilt/issues/6201


nicks:
@lizzthabet curious how this dovetails with other "organizing containers" work you've been doing.

nicks:
we talked about this a bit today. we're concerned that the current 'collapse' UI doesn't really give much indication to new users what to do or what parts of the interface is important. 

thinking about introducing some new "default" state that is more compact, but gives a stronger indication of what's happening.

PR #6214, opened 2023-08-30
cli: report an error if embedded assets aren't available

nicks:
prepare for deprecation of cloud assets

### DIFF
--- a/.circleci/Dockerfile.integration
+++ b/.circleci/Dockerfile.integration
@@ -1,4 +1,6 @@
-FROM golang:1.20
+FROM cimg/go:1.20-node
+
+USER root
 
 # Install dependencies from OS package manager sources
 RUN apt update && apt install -y --no-install-recommends apt-transport-https \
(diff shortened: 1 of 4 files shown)


landism (approved):

jgillich:
What needs to be done to embed the assets at build time? This error ("Error: running in prod mode, but JS/CSS files are not available") appears in v0.33.6 of the tilt [Nix package](https://github.com/NixOS/nixpkgs/blob/8cfdb77c1f6b8680061d2caa0f9e297d767555ea/pkgs/applications/networking/cluster/tilt/default.nix). IIRC go embed happens automatically.. :thinking:

nicks:
@jgillich oops I didn't know there was a nix package! Sorry for the breakage.

Ya, it needs to build the js assets. There's a good example in the homebrew formula:
https://github.com/Homebrew/homebrew-core/blob/faea31354e37ab5d21a9ce4ebea6bee209f3d9c3/Formula/t/tilt.rb

This will make everything more hermetic and faster overall so it's a good change anyway.

PR #6229, opened 2023-09-21
ci: switch to M1 runner for macOS builds

milas:
Per the banner in the CircleCI dashboard leading to: https://discuss.circleci.com/t/macos-intel-support-deprecation-in-january-2024/48718

Intel runners are going away, so opting into the M1/medium runner now, which will be the default as the others are retired.

Note that this is only used to verify compilation and run OS-specific integration tests (file system watching) via Go, so everything should continue working 🤞

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -130,6 +130,7 @@ jobs:
           only_for_branches: master
 
   build-macos:
+    resource_class: macos.m1.medium.gen1
     macos:
       xcode: "13.4.1"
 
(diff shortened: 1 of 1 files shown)


nicks (approved):

nicks:
LOL - i see the problem

We're currently on the free plan, coasting on a big stockpile of credits in the Docker circleci account.

According to this doc -
https://discuss.circleci.com/t/macos-intel-support-deprecation-in-january-2024/48718#timeline-5
They're not letting people on the free plan switch to the ARM runners until December 2023

nicks:
i updated to a gen2 intel mac in a separate pr.

looks like we'll have to wait until december when the m1s become available to open source projects.

let's close this for now and revisit.

PR #6230, opened 2023-09-22
test: clean up update spec vs status

nicks:
newer versions of the fake client libraries are more strict about this.

### DIFF
--- a/internal/controllers/fake/upsert.go
+++ b/internal/controllers/fake/upsert.go
@@ -25,7 +25,7 @@ func UpsertSpec(ctx context.Context, t testing.TB, ctrlClient ctrlclient.Client,
 		if apierrors.IsAlreadyExists(err) {
 			return
 		}
-		assert.NoError(t, err)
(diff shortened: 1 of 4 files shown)


internal/controllers/fake/upsert.go
```diff
@@ -0,0 +1,63 @@
...
+	require.True(t, obj.GetName() != "", "object has empty name")
+	err := ctrlClient.Create(ctx, obj)
+	if ctx.Err() != nil {
+		return
+	}
+	if err != nil {
+		if apierrors.IsAlreadyExists(err) {
+			return
+		}
+		assert.NoError(t, err)
```
nicksieger, line 28: This reads funny to me -- I know you're asserting no error here because there is an error and you want to raise it. I wonder if this works a bit better.
```suggestion
		assert.Fail(t, "create failed", "%+v %+v", obj, err)
```

nicksieger (approved):
LGTM

PR #6238, opened 2023-09-29
k8s: Support circular owner references

mterwill:
We introduced a circular owner reference (so one CRD gets cleaned up when another is deleted and vice versa) and found that Tilt got stuck in pending for all resources. I peeked at pprof and this looked suspicious:

```
25 @ 0x104a19594 0x1049e2d1c 0x1049e2884 0x106252da8 0x106252da9 0x106253908 0x106252d30 0x106253908 0x1062534c0 0x106d6c818 0x104a4f494
#	0x106252da7	github.com/tilt-dev/tilt/internal/k8s.(*objectTreePromise).wait+0x2e7						/root/project/internal/k8s/owner_fetcher.go:275
#	0x106252da8	github.com/tilt-dev/tilt/internal/k8s.OwnerFetcher.OwnerTreeOfRef+0x2e8						/root/project/internal/k8s/owner_fetcher.go:157
#	0x106253907	github.com/tilt-dev/tilt/internal/k8s.OwnerFetcher.ownerTreeOfHelper+0x217					/root/project/internal/k8s/owner_fetcher.go:222
#	0x106252d2f	github.com/tilt-dev/tilt/internal/k8s.OwnerFetcher.OwnerTreeOfRef+0x26f						/root/project/internal/k8s/owner_fetcher.go:175
#	0x106253907	github.com/tilt-dev/tilt/internal/k8s.OwnerFetcher.ownerTreeOfHelper+0x217					/root/project/internal/k8s/owner_fetcher.go:222
#	0x1062534bf	github.com/tilt-dev/tilt/internal/k8s.OwnerFetcher.OwnerTreeOf+0x26f						/root/project/internal/k8s/owner_fetcher.go:214
#	0x106d6c817	github.com/tilt-dev/tilt/internal/controllers/core/kubernetesdiscovery.(*Reconciler).handlePodChange+0xb7	/root/project/internal/controllers/core/kubernetesdiscovery/reconciler.go:628
```

Found that Tilt was deadlocked on these promises for resources that had owner references pointing to one another.

### DIFF
--- a/internal/k8s/owner_fetcher.go
+++ b/internal/k8s/owner_fetcher.go
@@ -6,6 +6,7 @@ import (
 	"strings"
 	"sync"
 
+	"golang.org/x/exp/slices"
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -147,6 +148,10 @@ func (v OwnerFetcher) getOrCreatePromise(id types.UID) (*objectTreePromise, bool
 }
 
 func (v OwnerFetcher) OwnerTreeOfRef(ctx context.Context, ref v1.ObjectReference) (result ObjectRefTree, err error) {
+	return v.ownerTreeOfRefHelper(ctx, ref, nil)
+}
+
+func (v OwnerFetcher) ownerTreeOfRefHelper(ctx context.Context, ref v1.ObjectReference, path []types.UID) (result ObjectRefTree, err error) {
 	uid := ref.UID
 	if uid == "" {
 		return ObjectRefTree{}, fmt.Errorf("Can only get owners of deployed entities")
@@ -172,7 +177,7 @@ func (v OwnerFetcher) OwnerTreeOfRef(ctx context.Context, ref v1.ObjectReference
 		}
 		return ObjectRefTree{}, err
 	}
-	return v.ownerTreeOfHelper(ctx, ref, meta)
+	return v.ownerTreeOfHelper(ctx, ref, meta, path)
 }
 
 func (v OwnerFetcher) getMetaByReference(ctx context.Context, ref v1.ObjectReference) (metav1.Object, error) {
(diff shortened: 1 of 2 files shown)


internal/k8s/owner_fetcher.go
```diff
@@ -211,15 +216,20 @@ func (v OwnerFetcher) OwnerTreeOf(ctx context.Context, entity K8sEntity) (result
...
-	return v.ownerTreeOfHelper(ctx, ref, meta)
+	return v.ownerTreeOfHelper(ctx, ref, meta, nil)
 }
 
-func (v OwnerFetcher) ownerTreeOfHelper(ctx context.Context, ref v1.ObjectReference, meta metav1.Object) (ObjectRefTree, error) {
+func (v OwnerFetcher) ownerTreeOfHelper(ctx context.Context, ref v1.ObjectReference, meta metav1.Object, path []types.UID) (ObjectRefTree, error) {
 	tree := ObjectRefTree{Ref: ref, CreationTimestamp: meta.GetCreationTimestamp()}
 	owners := meta.GetOwnerReferences()
 	for _, owner := range owners {
+		// TODO: Owner references can also exist at cluster scope, for which this incorrectly propagates the parent ref's Namespace.
```
mterwill, line 226: Not sure what you want to do here. I noticed this as I was testing but it doesn't seem to materially change how Tilt operates and as far as I could tell there's no easy way in this code path to determine if a resource exists at cluster scope. The namespace of the owner reference is not specified in its object ([ref](https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1#OwnerReference)).

> OwnerReference contains enough information to let you identify an owning object. An owning object must be in the same namespace as the dependent, or be cluster-scoped, so there is no namespace field. +structType=atomic

nicks: maybe it's ok? our client library wrappers all require a namespace argument, but i think they just drop it if the resource is cluster-scoped.

nicks (commented):
ha! today i learned that owner refs could be circular!

i think the test failure on ci looks legit?

mterwill:
Yes the test failure was definitely legit! I was originally using a slice of the entire owner reference (rather than just the UID) which is how I realized the namespace discrepancy. Was appending the wrong UID to `path` when I switched it, thanks test 😅

internal/k8s/owner_fetcher.go
mterwill: Want me to adjust to a `NOTE` or anything?

mterwill:
OK now I think this test failure might be a flake?

nicks:
yep! merging

nicks (approved):

PR #6241, opened 2023-10-06
assets: remove the GCP-based asset loader

nicks:
instead, we now use the embedded asset loader everywhere


nicksieger (approved):
lgtm, nice to unload this. This means that Tilt won't render really old snapshots, only ones created with a recent Tilt version or newer, right?

nicks:
No, it will still load old snapshots. It will just use the embedded js to view them. So in theory, if there are backwards incompatible charges to the file format, things might get weird.

PR #6243, opened 2023-10-09
portforward: fix a flakey test

nicks:
- adds redundant reconcilation
- injects a clock


nicks:
actually, i don't think this is the right solution

Issue #6250, opened 2023-10-16
0.33.6 breaks Docker Compose with profiles


Tadimsky:
### Expected Behavior
We have a docker compose file that disables one of our services by default using docker compose profiles.

```yaml
libpostal-http:
    profiles:
      - "include_libpostal"
```

Using `COMPOSE_PROFILES=include_libpostal docker compose up ...` will let you start the service if desired (outside of Tilt).

In our Tiltfile we have this:
```
dc_resource("libpostal-http", auto_init=False, labels=["dependencies"])
```

Before upgrade to 0.33.6 all of this worked totally fine. The service didn't start by default, but could be started in the Tilt UI.

### Current Behavior

Tilt 0.33.6 now has the following error when starting up:

```
Traceback (most recent call last):
  /Users/jonno/Source/rainbow/Tiltfile:16:12: in <toplevel>
Error in dc_resource: no Docker Compose service found with name "libpostal-http". Found these instead:
	db-dev-service; muskie-http; temporal-postgresql; temporal; temporal-ui
```

I'd guess this is related to the upgrade of the compose-go library: https://github.com/tilt-dev/tilt/pull/6225

### Steps to Reproduce
1. Have a docker compose stack with one service using a profile to disable starting up by default
2. Reference the docker compose service in the Tiltfile

#### `tilt doctor` Output

```
$ tilt doctor
Tilt: v0.33.6, built 2023-09-29
System: darwin-arm64
---
Docker
- Host: unix:///Users/jonno/.docker/run/docker.sock
- Server Version: 24.0.6
- API Version: 1.43
- Builder: 2
- Compose Version: v2.22.0-desktop.2
---
Kubernetes
- Env: eks
- Context: arn:aws:eks:us-west-2:535092081012:cluster/temporal
- Cluster Name: arn:aws:eks:us-west-2:535092081012:cluster/temporal
- Namespace: default
- Container Runtime: read-failure
- Version: Error: Get "https://B36932AA0B6F7BDA656BE82B6B587448.yl4.us-west-2.eks.amazonaws.com/version": dial tcp: lookup B36932AA0B6F7BDA656BE82B6B587448.yl4.us-west-2.eks.amazonaws.com: no such host
- Cluster Local Registry: none
---
Thanks for seeing the Tilt Doctor!
Please send the info above when filing bug reports. 💗

The info below helps us understand how you're using Tilt so we can improve,
but is not required to ask for help.
---
Analytics Settings
[...]

PR #6261, opened 2023-11-19
dockercompose: support compose profiles

coldfgirl:
Adds support for compose profiles to both provide a solution for https://github.com/tilt-dev/tilt/issues/6250 and to improve the integration with docker compose.

The PR which changed the previous functionality from compose-go between `v1.6.0` and `v1.18.4` is most likely https://github.com/compose-spec/compose-go/pull/359.

### DIFF
--- a/internal/tiltfile/docker_compose.go
+++ b/internal/tiltfile/docker_compose.go
@@ -57,12 +57,14 @@ func (dcm dcResourceMap) ServiceCount() int {
 func (s *tiltfileState) dockerCompose(thread *starlark.Thread, fn *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
 	var configPaths starlark.Value
 	var projectName string
+	var profiles value.StringOrStringList
 	envFile := value.NewLocalPathUnpacker(thread)
 
 	err := s.unpackArgs(fn.Name(), args, kwargs,
 		"configPaths", &configPaths,
 		"env_file?", &envFile,
 		"project_name?", &projectName,
+		"profiles?", &profiles,
 	)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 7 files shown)


internal/tiltfile/docker_compose.go
```diff
@@ -57,12 +57,14 @@ func (dcm dcResourceMap) ServiceCount() int {
 func (s *tiltfileState) dockerCompose(thread *starlark.Thread, fn *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
 	var configPaths starlark.Value
 	var projectName string
+	var profiles value.StringOrStringList
```
nicks, line 60: in this PR, the default behavior is always "no profiles". is that right?

maybe the default behavior should be "read the profiles from the COMPOSE_PROFILES env var"

what do you think?

coldfgirl: I think it depends on how much tilt wants the local environment to colour what tilt does versus what is defined in a `Tiltfile`. I'm not super familiar with the project so I don't know how much or if this already happens. From my perspective though the usage of `COMPOSE_PROFILES` or the `--profile` argument can be options of docker compose but not necessarily of tilt.

For my own setup I find value in the straightforwardness of my full definition being inside the `Tiltfile`. I think this also works especially well for when I have multiple `docker-compose.yml` files I'm loading and I want to be intentional about which services I need from each.

I have to say though that the setup I'm using is relatively simplistic so if there are arguments for using the `COMPOSE_PROFILES` environment variable I probably wouldn't have run into them anyways :sweat_smile:

coldfgirl: Ah I forgot, to answer your question: it is correct that the default behaviour is no profiles.

nicks (approved):
LGTM!

ok. this is a good start. let me ask the docker-compose people what they think about making it use the COMPOSE_PROFILES env var

Issue #6231, opened 2023-09-22
More details in timeout error from `tilt ci`


dnephin:
## Describe the Feature You Want

When `tilt ci` reaches the `ci_settings.timeout` the error it returns is not very informative (ex: "Error: Timeout after 30m0s"). Generally when our builds reach the timeout there are one or more pods that are in crash loop backoff.

I'd like to augment the timeout error and include something like the detail we get for non-timeout builds.

> Error: Timeout after 30m0s: Pod "podname-mb2cb" is in state CrashLoopBackOff

I guess in this case there may be more than one pod that is not ready. I'm not sure if that complicates things or not.

### Current Behavior

> Error: Timeout after 30m0s

### Why Do You Want This?

We're working on making it easier to quickly find the common flakes in our CI builds. When we get an error that includes the pod name it's easier to aggregate the errors. When we get errors like `Timeout after x` it's more difficult because we don't know which pods caused the timeout. There could be many different causes for those timeouts.

nicks:
i like this idea! i'm not sure how you'd thread the info through though.

nicks:
maybe a good place to start would be to check for some common cases :thinking: need to think on it though.

dnephin:
#6262 has been working great for this, so I'd be fine to close this issue now. I think there is still potential to include more details about why something is not ready when the timeout is reached, but I'm not sure if it's worth the effort.

nicks:
🎉

PR #6262, opened 2023-11-20
More verbose timeout reason

dnephin:
Related to #6231

## Background

We've been trying out a new setup where we pre-create environments ahead of time using `tilt up`. Then we pass the environment to `tilt ci` for CI runs. That's been working well, but we've noticed a few Timeout failures where every service in the snapshot has a `RuntimeStatus: ok`. 

It seems like there may be something stuck that isn't visible from the `UIResources` stored in the snapshot.

## Changes

This PR attempts to expose more detail about why `tilt ci` reached the timeout. I'll do some testing of this internally using this patch to see if we can isolate the problem a bit further.

This message could be too verbose in some cases. Maybe instead of printing all 3 categories it only includes the first category that includes resources? Any retrying resources, then not ready resource,  then finally waiting?

### DIFF
--- a/internal/controllers/core/session/reconciler_test.go
+++ b/internal/controllers/core/session/reconciler_test.go
@@ -173,7 +173,7 @@ func TestExitControlCI_Timeout(t *testing.T) {
 
 	f.clock.Advance(20 * time.Second)
 	f.MustReconcile(sessionKey)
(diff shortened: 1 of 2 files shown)


nicks (approved):
LGTM!

nicks:
ya, i think this is a good place to start

PR #6264, opened 2023-11-22
tiltfile: compose projects should default to COMPOSE_PROFILES env var

nicks:
fixes https://github.com/tilt-dev/tilt/issues/6250


nicks:
this builds on the work in https://github.com/tilt-dev/tilt/pull/6261

nicksieger (approved):

PR #6273, opened 2023-12-03
Handle jobs with deleted pods

### LINKED ISSUE #6272: Job never becomes ready when re-attaching to exiting env when pods for the job are deleted
### Background

I believe this is the same issue as described in #5822. That issue was closed over a year ago, so creating a new one.

We have a system that pre-creates tilt environments for `tilt ci`, so when `tilt ci` runs it resumes from an existing environment. We've noticed that occasionally those resumes will fail with a timeout. Using the new output from #6262 we noticed that when it happens it's always a job that never goes ready. Looking at logs we notice that these jobs had deleted pods.

#5859 added special handling for "completed jobs without pods" to [kubernetesapply/reconciler.go](https://github.com/tilt-dev/tilt/blob/38238ecfc9d008ed259ce881c3f78f4ce869bde0/internal/controllers/core/kubernetesapply/reconciler.go#L600-L622), but that does not appear to work for [processExitCondition](https://github.com/tilt-dev/tilt/blob/38238ecfc9d008ed259ce881c3f78f4ce869bde0/internal/controllers/core/session/status.go#L60-L141).

It appears that [session/conv.go:Reconciler.k8sRuntimeTarget](https://github.com/tilt-dev/tilt/blob/38238ecfc9d008ed259ce881c3f78f4ce869bde0/internal/controllers/core/session/conv.go#L92-L102) sees a job with `status == v1alpha1.RuntimeStatusOK`, but there are no pods, so `v1.PodSucceeded == phase` is false, and the resource is marked as `Active.Ready = true`.  In `processExitCondition`  
https://github.com/tilt-dev/tilt/blob/38238ecfc9d008ed259ce881c3f78f4ce869bde0/internal/controllers/core/session/status.go#L101

means that the ready == True is ignored for jobs, so they never appear ready when the pods are gone.

### Expected Behavior

Re-attaching to an environment with `tilt ci` recognizes a job is completed, even when the pods for that job are deleted.

### Current Behavior

Today it will timeout with something like:

```
Error: Timeout after 30m0s: 1 resources not ready (db-init:runtime)
```

### Steps to Reproduce
1. Using the example `Tiltfile` from #5822
2. `tilt ci` and wait for it to complete
3. `kubectl delete` the `pod/db-init-` pod
4. `tilt ci --timeout 1m` - this will timeout instead of going ready

### Implementation Questions

I think we understand the problem well enough, but I'm not entirely sure how to fix it yet.

I think in `k8sRuntimeTarget` we can handle when no pod is found from `krs.MostRecentPod`, and produce a state of `Terminated`.  It seems like the condition is available at that time.

dnephin:
Fixes #6272

Check for the condition where a job is completed but no longer has pods, and set those to `state.Terminated` to indicate they are ready.

Tested manually using the steps described in #6272, and added a unit test that fails without this change.

### DIFF
--- a/internal/controllers/core/session/conv.go
+++ b/internal/controllers/core/session/conv.go
@@ -6,6 +6,7 @@ import (
 	"time"
 
 	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	ctrl "sigs.k8s.io/controller-runtime"
 
 	"github.com/tilt-dev/tilt/internal/engine/buildcontrol"
@@ -89,7 +90,7 @@ func (r *Reconciler) k8sRuntimeTarget(mt *store.ManifestTarget, ci *v1alpha1.Ses
 	}
 
 	if status == v1alpha1.RuntimeStatusOK {
-		if v1.PodSucceeded == phase {
+		if v1.PodSucceeded == phase || hasConditionBatchJobComplete(krs.Conditions) {
 			target.State.Terminated = &session.TargetStateTerminated{
 				StartTime: createdAt,
 			}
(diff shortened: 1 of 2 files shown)


nicks:
thanks for investigating this!

let me poke around a bit before we merge this approach. historically, we try to keep "status interpretation" as low level as possible, or you get weird bugs where multiple layers are interpreting the status differently.

nicks:
i re-rolled this into https://github.com/tilt-dev/tilt/pull/6276, which has some additional integration tests. ya, i think removing the `|| res.Type == v1alpha1.TargetTypeJob` is the right change.

PR #6286, opened 2023-12-14
cli: add 'tilt alpha shell' for creating a shell into a container

nicks:
designed to interoperate with the new `kubectl shell`, but fallback to kubectl exec if the container supports it.

Future work:
- use with Docker Compose
- add a terminal to the tilt UI

### DIFF
--- a/internal/cli/shell.go
+++ b/internal/cli/shell.go
@@ -1,0 +1,203 @@
+package cli
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"io"
+	"os"
+	"os/exec"
+	"syscall"
+
+	"github.com/spf13/cobra"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/cli-runtime/pkg/genericclioptions"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/tilt-dev/tilt/internal/container"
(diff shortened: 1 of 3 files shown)


nicks:
cc @joe0BAB

nicksieger (approved):
Nice!

internal/cli/shell.go
```diff
@@ -0,0 +1,202 @@
...
+		Long: `Opens a shell into a container running in Tilt.
+
+Given a resource name, finds the Kubernetes Pod in that resource,
+and opens an interactive shell.
+
+By default, in order, we'll try:
+- kubectl shell
+- dld shell
+- docker exec -it <container> bash
+- docker exec -it <container> sh`,
```
errordeveloper, line 55: should there be a fallback to `kubectl debug` also?

nicks: i think `kubectl debug` is for a different use-case? i.e., it's more for "i'm debugging the infra and want to ssh to the Node where this pod lives and check the firewall rules" rather than tilt's "i'm debugging the container and want to poke around on its filesystem"

errordeveloper: It's interesting you thought that. You can attach to a node, but you can do the same with a pod. It's currently a bit short of functionality of enabling what 'kubectl shell' does on the client side, there is an open PR to add a privileged profile.

nicks:
i put this under `tilt alpha shell` for now, there's still some ways that the underlying tooling is evolving

PR #6288, opened 2023-12-15
cli: give up and suppress klog output by default

nicks:
i don't think these logs have ever caught a real problem, and are often confusing / inscrutable. they're really meant for people writing k8s controllers, and i don't think they make sense to show to tilt users.


nicksieger (approved):
👍

Issue #6054, opened 2023-02-28
silence live_update run() commands


nicks:
## Describe the Feature You Want
Michael writes in the tilt channel:

> when using the run option with LiveUpdates to run a command in a container after files are synced, is there a way to not have the command being run printed into the logs? We're running a fairly long curl command and we don't need it printed out in the logs if there's an easy way to suppress that.

i don't think there's a way to do this right now but might not be too hard to add.

muripic:
I'm also interested in this 🙏 
We need to run a command with a secret we have locally but the way it is now, it's printed out every time

nicks:
fixed in https://github.com/tilt-dev/tilt/pull/6351

PR #6351, opened 2024-04-08
tiltfile: add echo_off to live update run

muripic:
This PR aims to solve issue https://github.com/tilt-dev/tilt/issues/6054.

It adds an `echo_off` parameter (as in the `local` command, see https://docs.tilt.dev/api.html#api.local) to silence printing out the command that is being run.

Ran `make test` and also tested by running Tilt with the built binary:

Without `echo_off`:
Screenshot 2024-04-05 at 14 51 52

With `echo_off`:
Screenshot 2024-04-05 at 14 51 24


nicks (approved):
nice change! lgtm

muripic:
Thank you for your review @nicks 🙇‍♀️ ❤️ 
I see that the tests in master have failed due to a timeout 😞 I think a retry should work, right?
Also, do you think it'd be possible to have a release soon? 🙏

PR #6353, opened 2024-04-09
goreleaser: update asset packaging directory

nicks:


com6056 (approved):
Thanks for the quick fix! 🎉

Issue #6356, opened 2024-04-16
When specifying a deployment namespace with `tilt up --namespace <NAMESPACE>`, if `<NAMESPACE>` does not currently exist, create the namespace.


peter-mcclonski:
## Describe the Feature You Want
When specifying a deployment namespace with `tilt up --namespace <NAMESPACE>`, if `<NAMESPACE>` does not currently exist, create the namespace.

### Current Behavior
If the namespace does not already exist, deployments will fail.

### Why Do You Want This?
There is currently no good way to have a custom namespace created for tilt.  Based on the initial description of this flag on release (https://blog.tilt.dev/2021/06/30/namespace-flag.html), it's implied that previous uses of the `namespace` extension, including `namespace_create('test-default')` should be fully replaced by the flag.

nicks:
hmmm...i'm not sure we should do this.

The tilt `--namespace` flag is indented to map to `kubectl --namespace`. In kubectl, it sets the default namespace for API requests - https://pkg.go.dev/k8s.io/client-go@v0.29.3/tools/clientcmd/api/v1#Context

`kubectl --namespace=... apply` does not auto-create the namespace. 

Helm has a separate `--create-namespace` flag, which you can simulate in tilt as:

``
namespace_create(k8s_namespace())
```

peter-mcclonski:
@nicks Fair concerns.  Regardless, I would make the case that this is a reasonable quality-of-life improvement.  If tying it directly to the `--namespace` flag is undesirable, perhaps adding a separate tilt `--create-namespace` flag to be analogous to the underlying helm `--create-namespace` flag?   

I would make the argument that the current reliance `namespace_create(k8s_namespace())` is non-intuitive, and contrasts with the linked tilt feature announcement.  The `--namespace` flag being provided out of the box contrasts with the need to load a separate extension for `namespace_create(...)`.  At minimum, if specifying the `--namespace` flag and the namespace does not exist, I would expect to see the initial tilt up fail immediately.

nicks:
ya, i guess we'd accept a --create-namespace flag that mirrors the helm flag behavior

peter-mcclonski:
To make sure we're on the same page-- I'm thinking that this would be a tilt flag on the `up` command.  For instance, `tilt up --create-namespace`.  The flag would be meaningless (and provide an associated error) if not paired with the `--namespace` flag, and it would result in the namespace being created prior to the actual contents of the Tiltfile being processed.

PR #6357, opened 2024-04-16
If specifying --namespace, auto create the namespace if it does not e…

peter-mcclonski:
…xist.


peter-mcclonski:
Resolves #6356

nicks:
this modifies the `helm()` build-in tiltfile function, which i'm not sure is what you want?

PR #6360, opened 2024-04-25
circleci: update to go 1.22

nicks:


rafrafek:
Should https://github.com/tilt-dev/tilt/blob/master/.golangci.yml#L4 also be updated?

PR #6363, opened 2024-04-29
feat(extensions): support for nested extensions

avidal:
This patch introduces support for nested extensions. You can load a nested extension by supplying the child path in the extension module name at load time. This makes it simpler to namespace your extension without having to define multiple extensions or import every symbol into the main extension.

For instance, given an extension repo like such:

```
- tilt-lib
  - bazel
    - Tiltfile
    - watch
        - Tiltfile
  - helm
    - ...
```

You can load symbols from `tilt-lib/bazel/watch` via:

```
v1alpha1.extension_repo(name='custom', url='git://host/repo')
v1alpha1.extension(name='bazel', repo_name='custom', repo_path='tilt-lib/bazel')

load("ext://bazel/watch", "watch_files")
```

To take it further, given the same extension repo, you could:

```
v1alpha1.extension_repo(name='internal', url='git://host/repo')
v1alpha1.extension(name='internal', repo_name='internal', repo_path='tilt-lib')

load("ext://internal/bazel/watch", "watch_files")
load("ext://internal/helm/...", "...")
```

### DIFF
--- a/internal/tiltfile/tiltextension/plugin.go
+++ b/internal/tiltfile/tiltextension/plugin.go
@@ -7,6 +7,7 @@ package tiltextension
 import (
 	"context"
 	"fmt"
+	"path/filepath"
 	"strings"
 
 	"go.starlark.net/starlark"
@@ -23,8 +24,10 @@ import (
 	"github.com/tilt-dev/tilt/pkg/logger"
 )
 
-const extensionPrefix = "ext://"
-const defaultRepoName = "default"
+const (
+	extensionPrefix = "ext://"
+	defaultRepoName = "default"
+)
 
 type Plugin struct {
 	repoReconciler ExtRepoReconciler
@@ -126,9 +129,30 @@ func (e *Plugin) LocalPath(t *starlark.Thread, arg string) (localPath string, er
 //
 // If it has, returns the existing object (which should only have a spec).
 //
+// If it hasn't, but it's a nested extension name, and the root has been registered, register it and
+// return it.
+//
 // Otherwise, infers an extension object that points to the default repo.
 func (e *Plugin) ensureExtension(t *starlark.Thread, objSet apiset.ObjectSet, moduleName string) *v1alpha1.Extension {
+	// Copy the supplied module name (which may contain slashes) into the "repo path".
+	// This ensures that sanitizing or name munging does not effect the path where the extension
+	// will be loaded.
+	repoPath := moduleName
+
+	// Before sanitizing the name, split off anything after the first /
+	extRoot, nestedPath, isNested := strings.Cut(moduleName, "/")
+
+	// This is the name of the extension root. For regular (non-nested) extensions, this is the same
+	// as the sanitized moduleName.
+	// For nested extensions, this is the root extension name and it must already be registered.
 	extName := apis.SanitizeName(moduleName)
+	if isNested {
+		extName = apis.SanitizeName(extRoot)
+	}
+
+	// Create an extension spec that points to the default extension repository.
+	// This will be "registered" in place if the requested extension has not already been
+	// registered.
 	defaultExt := &v1alpha1.Extension{
 		ObjectMeta: metav1.ObjectMeta{
 			Name: extName,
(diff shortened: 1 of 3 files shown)


avidal:
I implemented this because of our extensive tilt library at Datadog. Without this patch in order to use any of our extensions developers have to register the extension before loading it. But with this patch, they can register a single "top-level" extension and load the nested extensions from there.

avidal:
I think that, in general, the extension registering mechanism for tilt could be improved. The way it is, you cannot define your own extension repository and have that "auto-register" every extension in the repository the way it's done for the default `tilt-dev/tilt-extensions`.

Perhaps a complementary patch would be to support an `extensions_dir` kwarg to the `extension_repo` API, followed by a patch to `ensureExtension` (as changed in this PR) that would register an extension against a custom repo if the "module name" prefix is the name of said extension repo.

With those two patches combined, and with the extension repo as described in the PR description, you could use it like so:

```
# Register the internal repo where all extensions are under ./tilt-lib
v1alpha1.extension_repo(name='internal', url='git://host/repo', extensions_dir='tilt-lib')

# Load bazel/watch from the internal repo
load("ext://internal/bazel/watch", "..")
```

pkg/apis/identifiers.go
```diff
@@ -51,15 +50,18 @@ func SanitizeLabel(name string) string {
...
 	sanitized := name
+
 	if len(path.IsValidPathSegmentName(name)) != 0 {
 		for _, invalidName := range path.NameMayNotBe {
 			if name == invalidName {
 				// the only strictly invalid names are `.` and `..` so this is sufficient
 				return strings.ReplaceAll(name, ".", "_")
 			}
 		}
-		sanitized = invalidPathCharacters.ReplaceAllString(sanitized, "_")
```
avidal, line 61: I realize after the fact that this alone is a pretty significant change. I don't believe it's inherently a *problem*, as a URL escape will still produce valid path names, but I think a less broad change would be to replace `%` with `_` as the original implementation did, and replace `/` with `%2F`.

internal/tiltfile/tiltextension/plugin.go
```diff
@@ -126,9 +129,30 @@ func (e *Plugin) LocalPath(t *starlark.Thread, arg string) (localPath string, er
...
+//
 // Otherwise, infers an extension object that points to the default repo.
 func (e *Plugin) ensureExtension(t *starlark.Thread, objSet apiset.ObjectSet, moduleName string) *v1alpha1.Extension {
+	// Copy the supplied module name (which may contain slashes) into the "repo path".
+	// This ensures that sanitizing or name munging does not effect the path where the extension
+	// will be loaded.
+	repoPath := moduleName
+
+	// Before sanitizing the name, split off anything after the first /
+	extRoot, nestedPath, isNested := strings.Cut(moduleName, "/")
```
nicks, line 143: i don't think it's correct to assume that if a module name has a slash, then it's nested in another extension, here's an extension this would break:
https://github.com/tilt-dev/tilt-extensions/tree/master/tests/golang#usage

nicks (commented):
ya...as you pointed out in a later comment, this PR feels like a half-measure that introduces a bunch of odd edge cases, and the end result still isn't that great.

i liked your suggestion further down. i think if you structured the api a bit like:

```
v1alpha1.extension_repo(name='internal', url='git://host/repo', load_name='internal')
```
where `load_name` would be an signal to the extension load system that it should load from the `internal` repo instead of the `defaultRepoName` repo when it goes down the regular extension load path:

https://github.com/tilt-dev/tilt/blob/b4c799391338d2ef5ff7b5a91984da5276dc8970/internal/tiltfile/tiltextension/plugin.go#L140

and i think that'd be much simpler implementation-wise overall? wdyt?

pkg/apis/identifiers.go
avidal: This function is called against extension module names and you lose context: `module/sub` turns into `module_sub` which could be the name of an actual extension.

Since `%2F` is valid in path names it seemed a reasonable compromise.

internal/tiltfile/tiltextension/plugin.go
avidal: How would that break? It'd load the nested `golang` extension underneath the `tests` extension.

Also, I hadn't even realized that there were already nested extensions in the default repo. I'll take a look at that and adjust.

avidal:
> i liked your suggestion further down about making extensions load implicitly like they do with the default repo. i think if you structured the api a bit like:

This is what I wanted in the first place (even brought it up with you a couple of years ago in Slack). The clue that `tests/golang` already works to load from the default repo is a bit of a revelation, because the *first* thing I did when working on this was to add a test to load a nested extension which failed. I don't have the test output but perhaps the signal was there and I missed it.

I'll try this again from a fresh perspective. If nested extensions work as described and the only thing we need to do is support a "subpath" in the extensions repo to register a prefix that'd be a better solution.

Also, not sure why I didn't see your review pop up in my email. Sorry for the delay and thanks for looking even though it's in draft.

avidal:
Closing in favor of #6371

Issue #6378, opened 2024-05-20
tty in local_resource kills tilt


nicks:
### Steps to Reproduce

Tiltfile:

```
local_resource('test', ['./test.sh'])
```

test.sh:

```
#!/bin/bash

zsh -i -l -c 'exit;'
```

1) Run `tilt up` and wait a little while
2) Hit `space`

### Expected Behavior
Tilt should successfully run the command

### Current Behavior
On macOS, tilt crashes with `suspended (tty input)`
On Linux, the terminal gets mysteriously detached from the tilt process

nicks:
i was able to get something working by attaching a pty to stdin of the process...

i'm not sure if that's a good _general_ solution (since some stuff may behave differently when it's interacting with a pty) but i guess we can make it an option

nicks:
the pr was rolled back

PR #6382, opened 2024-05-24
tiltfile: add stdin_mode='pty' to local_resouce()

nicks:
Allows users to attach a pty,
for interactive commands that try
to grab the "main" tty

fixes https://github.com/tilt-dev/tilt/issues/6378


internal/tiltfile/local_resource_test.go
```diff
@@ -10,3 +17,275 @@ test("test", "echo hi")
 `)
 	f.loadAssertWarnings(testDeprecationMsg)
 }
+
+func TestLocalResourceOnlyUpdateCmd(t *testing.T) {
```
nicks, line 21: this is just moving code around

Issue #6387, opened 2024-05-31
tty in node in local_resource kills tilt


nicks:
Related to https://github.com/tilt-dev/tilt/issues/6378, but the repro steps are slightly different :thinking: 

### Steps to Reproduce

Tiltfile:

```
local_resource(
    'test',
    ['node', './main.js'],
    stdin_mode='pty')
```

main.js:

```
const { spawn } = require('node:child_process');

let promise = new Promise((resolve, reject) => {
  let proc = spawn('/bin/zsh', [
    '-i',
    '-l',
    '-c',
    'echo exit',
  ]);

  proc.on('error', (err) => {
    console.log('ERROR PROCESS');
  });
    
  proc.on('close', (code, signal) => {
    console.log('CLOSE PROCESS');
  });
});
```

1) Run `tilt up` and wait a little while
2) Hit `space`

### Expected Behavior
Tilt should successfully run the command

### Current Behavior
On macOS, tilt crashes with `suspended (tty input)`
On Linux, the terminal gets mysteriously detached from the tilt process

PR #6388, opened 2024-05-31
tiltfile: revert pty code

nicks:
After playing with this a bit, and seeing lots of cases
where it falls down, i no longer think this is a good idea.


nicks:
see also https://github.com/tilt-dev/tilt/issues/6387

PR #6398, opened 2024-06-18
circleci: update build-macos to use new runners

nicks:
(the old runners are being deprecated)


nicks:
ah looks like circleci still doesn't have M1s ready for our plan - https://discuss.circleci.com/t/macos-intel-support-deprecation-in-january-2024/48718/31

PR #6408, opened 2024-07-03
tiltfile/helm: handle parsing version strings without `v` prefix

aybabtme:
`helm version` doesn't necessarily return a string with `v` in the prefix, as it depends how it's packaged and bundled. For instance when running helm with [pkgx.sh](https://pkgx.sh/) (for [pinning tool dependencies](https://docs.pkgx.sh/using-dev/dev)):

```
$ pkgx helm version --client --short
3.15.2
```

(due to https://github.com/pkgxdev/pantry/blob/main/projects/helm.sh/package.yml#L22)

Without this change, tilt fails to parse the version with error:

```
Error in helm: could not parse Helm version from string: "3.15.2\n"
```

This fixes that.

### DIFF
--- a/internal/tiltfile/helm.go
+++ b/internal/tiltfile/helm.go
@@ -64,7 +64,7 @@ func parseVersion(versionOutput string) (helmVersion, error) {
 
 	if strings.HasPrefix(version, "v3.0.") {
 		return helmV3_0, nil
-	} else if strings.HasPrefix(version, "v3.") {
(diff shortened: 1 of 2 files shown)


aybabtme:
Not sure how to get CI to run.

nicks:
ugh, circleci has been really awful the past couple weeks, let me kick it to see if it resolves.

nicks (approved):

PR #6414, opened 2024-07-22
Populate docker compose healthcheck status

### LINKED ISSUE #6413: Use docker compose healthcheck status as runtime status on the UI

Today tilt UI does not present information on [docker compose health-check status](https://docs.docker.com/compose/compose-file/#healthcheck). Propagating this information can be helpful at spotting any development issues that are run-time specific.

### Current Behavior

Currently, when container somehow starts up but it's broken in a way that the process does not exit, but the health-check of the container indicates problem, it will be reported on the tilt UI at all-green. Tilt UI already has means to present such state more precisely.

### Why Do You Want This?

Use docker container status information and propagate it into runtime status information so that tilt UI can render is nicely.

**Additional context**

It is at least part of #6037, but since that other ticker is broader, I think it's worth separating it.

remi00:
With this change docker compose healthcheck status details are populated into runtime status, making the tilt UI reflect detailed information.

Closes #6413

### DIFF
--- a/internal/dockercompose/state.go
+++ b/internal/dockercompose/state.go
@@ -43,6 +43,9 @@ func (s State) RuntimeStatus() v1alpha1.RuntimeStatus {
 	if s.ContainerState.Error != "" || s.ContainerState.ExitCode != 0 {
 		return v1alpha1.RuntimeStatusError
 	}
+	if s.ContainerState.Status == ContainerStatusRestarting {
+		return v1alpha1.RuntimeStatusPending
+	}
 	if s.ContainerState.Running ||
 		s.ContainerState.Status == ContainerStatusRunning ||
 		s.ContainerState.Status == ContainerStatusExited {
@@ -101,6 +104,34 @@ func (s State) HasEverBeenReadyOrSucceeded() bool {
 	return !s.LastReadyTime.IsZero()
 }
 
+func populateContainerStatus(state *types.ContainerState) string {
+	if state.Health == nil {
+		return state.Status
+	}
+	if state.Health.Status == types.Starting {
+		return ContainerStatusRestarting
+	}
+	return state.Status
+}
+
+func populateContainerError(state *types.ContainerState) string {
+	if state.Error != "" {
+		return state.Error
+	}
+	if state.Health == nil || state.Health.Status != types.Unhealthy {
+		return ""
+	}
+	result := "container is unhealthy"
+	log := state.Health.Log
+	if len(log) > 0 {
+		last := log[len(log)-1]
+		if last != nil && last.Output != "" {
+			result = result + ": " + last.Output
+		}
+	}
+	return result
+}
+
 // Convert ContainerState into an apiserver-compatible state model.
 func ToContainerState(state *types.ContainerState) *v1alpha1.DockerContainerState {
 	if state == nil {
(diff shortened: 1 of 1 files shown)


internal/dockercompose/state.go
```diff
@@ -101,6 +104,34 @@ func (s State) HasEverBeenReadyOrSucceeded() bool {
 	return !s.LastReadyTime.IsZero()
 }
 
+func populateContainerStatus(state *types.ContainerState) string {
+	if state.Health == nil {
+		return state.Status
+	}
+	if state.Health.Status == types.Starting {
+		return ContainerStatusRestarting
```
nicks, line 112: i wonder if this should be created?

nicks (approved):
thanks! let me test this a bit manually

internal/dockercompose/state.go
remi00: This, combined with a change in `RuntimeStatus()` method, improves the user experience on container rebuild in case of docker containers with healthcheck. It reflects that, there is additional "pending" state of healthcheck status.

Without it, in case when the healthcheck started failing after the rebuild, the runtime status would be first green and only after the healthcheck is considered as failing by docker, the runtime status would become red.

With this change the runtime status will be "pending" as long as the the healthcheck is "healthcheck: starting".

By the runtime status I mean the red element on Tilt UI:
image

nicks: do ContainerStatusRestarting and ContainerStatusCreated have different displays in the UI?

internal/dockercompose/state.go
```diff
@@ -101,6 +104,34 @@ func (s State) HasEverBeenReadyOrSucceeded() bool {
...
+}
+
+func populateContainerError(state *types.ContainerState) string {
+	if state.Error != "" {
+		return state.Error
+	}
+	if state.Health == nil || state.Health.Status != types.Unhealthy {
+		return ""
+	}
+	result := "container is unhealthy"
```
nicks, line 124: is this log displayed anywhere today? i didn't see it anywhere in the UI, only in the tilt cli output

internal/dockercompose/state.go
remi00: `ContainerStatusCreated` **->> green runtime status**
image

`ContainerStatusRestarting ` **->> below animated. At the end of animation is a transition to unhealthy state.**

https://github.com/user-attachments/assets/6371d470-11cf-40f4-b541-a00fdd71d619

Again, without the change in question when the healthcheck procedure is ongoing the runtime status would mistakenly indicate green status.

internal/dockercompose/state.go
remi00: I didn't find it on the UI either.

nicks:
OK, I played with this in a few projects, and now I remember why we did it this way.

The Compose Spec says this:

> Compose guarantees dependency services have been started before starting a dependent service. Compose waits for dependency services to be "ready" before starting a dependent service.

i.e., the default definition of readiness is "the service has started", NOT "the service is healthy"

if you define readiness this way, it will break startup order.

the problem here is that Tilt and Compose model dependencies differently -- in Tilt, if 2 resources depend on 'common', then both resources can start once 'common' is ready. In Compose, 2 resources can have dependencies on different parts of the 'common' lifecycle

i have some ideas on how to address your use-case a bit differently but need to play around with it a bit.

remi00:
@nicks Thanks for explaining. Let me know if I can help more. In the projects I am using tilt, this change drastically improves UX, but I understand that the readiness aspect is more important here than just displaying it on the UI.

remi00:
Just to add to my previous comment. The healthcheck reporting is an important capability for devs by itself. Considering that docker does support concept of a healthchecks, there should be a way to present the healthcheck status on the tilt UI if it supports docker-compose in the first place. Missing that is a huge gap. In my case current behavior was a huge confusion (tilt supports probing, supports docker-compose and it misses supporting combination of both). Even if there was no news on tilt-dev website that [Tilt has joined Docker](https://www.docker.com/blog/welcome-tilt-fixing-the-pains-of-microservice-development-for-kubernetes/).

nicks:
addressed this in another pr

Issue #6037, opened 2023-02-02
Readiness Probes with Docker Compose


saulshanabrook:
## Describe the Feature You Want

I am curious if Docker Compose can support some form of a readiness probe, either from the Docker HEALTHCHECK or a the probe you have defined.

### Current Behavior
I don't believe currently there is a probe for docker compose.

### Why Do You Want This?
I am trying to see if my HTTP server is up, and it would be great in the GUI if there was a way to show that.

nicks:
are you aware of compose's built-in healthcheck? https://docs.docker.com/compose/compose-file/#healthcheck

i don't think we currently display this in the tilt UI, but if we did, we'd probably have tilt hook into that (rather than have people specify it in the tiltfile)

saulshanabrook:
@nicks Yep that makes sense to me! I wasn't sure if this was currently displayed in the UI.

apaniel:
Any plans to show it in the UI? I'd be interested too

remi00:
Having this would be a great addition. Today in the setup based on docker compose we can get 🟢 **all green** and ready on Tilt dashboard whereas the service may be crashing miserably on run-time.

PR #6419, opened 2024-07-31
dockercompose: add healthcheck to tilt ui

nicks:
fixes https://github.com/tilt-dev/tilt/issues/6413
fixes https://github.com/tilt-dev/tilt/issues/6037


nicks:
2024-07-31T18:00:53,395028418-04:00

here's what it looks like

nicksieger (approved):

PR #6429, opened 2024-08-30
web: use dark color scheme to match UI

metonym:
I use Tilt daily.

This is an unsolicited suggestion for the web UI styling. Because the Tilt UI is a dark theme, https://developer.mozilla.org/en-US/docs/Web/CSS/color-scheme hints to browsers to apply similar adjustments to the interface. In this case, the change is most notable with the scrollbars.

## Before

By default, scrollbars rendered by Chrome for overflowing content are light-themed.

tilt-before

## After

By applying the declaration on the `html` root element, the browser applies a dark theme to the scrollbars on the `body` and the components list.

tilt-after

### DIFF
--- a/web/src/index.scss
+++ b/web/src/index.scss
@@ -3,6 +3,7 @@
 html {
   width: 100%;
   height: 100%;
+  color-scheme: dark;
 }
 body {
   margin: 0;


nicks (approved):
lgtm!

PR #6433, opened 2024-09-05
feat: Adds functionality to make CRD image replacement optional

anas-domesticus:
This PR adds an "optional" flag to the k8s_kind type, applicable to both json path image replacements & the object flavour.

This is required in some cases where an image may or may not exist in a CRD, such as Argo Rollouts' AnalysisTemplates.

As below:
```
k8s_kind(
    "AnalysisTemplate",
    api_version="argoproj.io/v1alpha1",
    image_optional=True,
    image_json_path="{.spec.metrics[*].provider.job.spec.template.spec.containers[*].image}",
    pod_readiness="ignore",
)
```

Previously this would fail if a metrics analysis template were configured alongside a job-based template (a smoke test in our use case). This is backwards compatible as default values for bools are `false`


anas-domesticus:
The file deletions occurred when I ran `make update-codegen`, is this correct?

nicks:
I thought jsonpath supported optional filters natively?

The syntax looks something like -

```
image_json_path="{.spec.metrics[*].provider.job.spec.template.spec.containers[?(@.image)].image}"
```

I'm afk right now but can check later if that doesn't work

nicks:
(I appreciate that the jsonpath syntax is kinda inscrutable 😔)

anas-domesticus:
You're right! It works with the following:

```
image_json_path="{.spec.metrics[?(@.provider.job)].provider.job.spec.template.spec.containers[*].image}",
```

Thanks for this!

Issue #3048, opened 2020-03-05
no way to express build dependencies in Tilt


nicks:
Currently, Tilt has resource_deps for expressing runtime initialization dependencies (e.g., "You need to start up the database before the frontend").

But Tilt has no way to express other types of dependencies, including:

1) A common pattern is to have a local_resource that generates files, and then a docker_build that syncs those files to a container. (e.g., generating API documentation in tilt.build, compiling Go static binaries). This workflow has a bunch of rough edges (logs aren't grouped like you would expect them, redundant builds, etc)

2) Another common pattern is to have a local_resource that solely exists to kick off another build (e.g., the "touch Tiltfile" script described in https://github.com/windmilleng/tilt/issues/2957, or the "measure deployment" script described in https://docs.tilt.dev/example_static_html.html). These scripts are prone to race conditions and infinite loop problems (https://github.com/windmilleng/tilt/issues/1930)

People often try to do this with resource_deps, and are confused about why it doesn't work. See also discussion on this PR: https://github.com/windmilleng/tilt/pull/2837#issuecomment-578315808

I don't know what the solution is to this problem, or even if (1) or (2) are the same problem. This issue is just to have a place to point people who ask about this while we think about what to do about it.

jmo-qap:
To bring some more commentary to the issue, it is is hitting us pretty hard in terms of development work hours and development speed slowness.

1) We have tests we want to run after a certain service is updated. We have to run the whole test suite twice (the second time triggered manually) to get a true result. Described in https://github.com/tilt-dev/tilt/issues/3667

2) We have the following steps that must happen in the following order (otherwise several services in the cluster will enter an error state and create crash loops): (a) database migration file generation, (b) execute database migration files, (c) restart/update service. For (a), we've moved it into a separate docker build stage which works but only without live update. Doing (a) locally as a separate command causes a double sync and will put the cluster into an error state first. For (b), we've added etcd as a lock service and the first service to start aquires the lock and does (b) while all the other services spin on the lock. Then, (c) can progress. The alternative is to configure all those services to update only on manual trigger and manually trigger in the right order. That is also slow.

Finally, pointing to the other issue on the topic to bring it all into one place:
https://github.com/tilt-dev/tilt/issues/3676

nicks:
@jmo-qap thanks for the context!

ya, just to make sure it's clearly stated, manual trigger should still work with live_update.

[...]

PR #6467, opened 2024-11-11
tiltfile: fix ignores if context dir appears later

liskin:
Consider the following (admittedly somewhat unusual) Tiltfile:

    trigger_mode(TRIGGER_MODE_MANUAL)

    clone_cmd = "test -e tilt-avatars || git clone https://github.com/tilt-dev/tilt-avatars"
    custom_build(
        "x-clone-repo",
        deps = [],
        skips_local_docker = True,
        command = clone_cmd,
    )
    docker_build(
        "x-build-image",
        context = "tilt-avatars",
        dockerfile_contents = "\n".join([
            "FROM x-clone-repo",
            "FROM scratch",
            "COPY ./ /opt/app/",
        ]),
        ignore = [
            "**/.git",
        ],
    )
    k8s_custom_deploy(
        name = "x-dummy-resource",
        apply_cmd = "true",
        delete_cmd = "true",
        deps = [],
        image_deps = ["x-build-image"],
    )

This implements on-demand git cloning — the clone happens during the build phase of the resource, and thus other builds can proceed concurrently and this may speed up overall tilt up (and it does in the quite complex tilt setup that we have).

In current Tilt, however, `ignore` and `only` are completely ignored if the context directory doesn't exist when the Tiltfile is loaded. This is quite a strange and seemingly pointless limitation, as everything else seems to work fine.

The fix is trivial — simply remove the IsDir check from `dockerignoresFromPathsAndContextFilters`. The code that follows the check handles the absence of the dockerignore files just fine, so it seems the IsDir check is simply forgotten there from previous iterations.

Note that a similar issue exists in the `repoIgnoresForPaths` function as well. The Tiltfile to reproduce that problem is the same, just without the `ignore = ["**/.git"]` bit. I'm not entirely sure if that `isGitRepoBase` check serves any real purpose — would there be any harm if ".git" was ignored unconditionally (well, we'd need to check for the empty path like `dockerignoresFromPathsAndContextFilters` does)?

Related: https://github.com/tilt-dev/tilt/issues/3048#issuecomment-836877339

### DIFF
--- a/internal/tiltfile/docker.go
+++ b/internal/tiltfile/docker.go
@@ -526,10 +526,6 @@ func (s *tiltfileState) dockerignoresFromPathsAndContextFilters(source string, p
 		}
 		dupeSet[path] = true
 
-		if !ospath.IsDir(path) {
-			continue
-		}
-
 		if len(ignorePatterns) != 0 {
 			result = append(result, model.Dockerignore{
 				LocalPath: path,


nicks:
hmmm...nice find! 

i'm not sure this is the right fix. note that the integration tests are failing.

AFAIR the problem we ran into is that `custom_build` lets you depend on lots of individual files. and having lots of individual files in the ignore spec really blew up the matching engine and made everything slow.

what if we changed the logic to something like

```
generateIgnores := false
if isDir(path) {
  generateIgnores = true
} else if !exists(path) && isDockerBuild {
  generateIgnores = true
}
```

i think that would fix your use-case a bit more narrowly.

nicks:
closing as stale

PR #6493, opened 2025-01-28
fix (Tiltfile) stuck in pending state

### LINKED ISSUE #5902: Bug report: Watched file tiltfile reload leaves (Tiltfile) pending until manual reload
### Expected Behavior
When a watched file triggers a tiltfile reload, the tiltfile shouldn't remain in pending after the file(s) is reloaded.

### Current Behavior
On occasion (unclear why), when a watched file triggers a reload of `(Tiltfile)` the tiltfile finishes reloading but the resource remains in pending, which blocks the reload of other services in tilt.

### Steps to Reproduce
1. Create a tiltfile that watches a file, and has resources that depend on the tiltfile (all)
2. Start tilt, let everything settle
3. `touch` the file being watched
4. Note the tiltfile reloads, and will sometimes still be marked as pending after (see image)

Screen Shot 2022-07-21 at 10 54 00 AM

#### `tilt doctor` Output
NOTE: I upgraded to v0.30.5 today and saw this happen again.

```
➜  ~ tilt doctor
Tilt: v0.30.3, built 2022-06-10
System: darwin-amd64
---
Docker
- Server Version: 20.10.17
- API Version: 1.41
- Builder: 2
- Compose Version: v2.6.1
---
Kubernetes
- Env: unknown
- Context: dev-3
- Cluster Name: dev-3
- Namespace: ws-nathan-heaps
- Container Runtime: cri-o
- Version: v1.21.8
---

---
Analytics Settings
- User Mode: opt-out
- Machine: 849283229ca9b586029a6d42301bc3c7
- Repo:
```

#### About Your Use Case
We use tilt heavily, many times >25 services enabled via tilt, and `(Tiltfile)` is all the way at the bottom. This being hit-or-miss tends to trend towards me seeing things hit `Waiting on (Tiltfile)` and not understanding why, if I notice at all.

knisbet:
Fixes #5902 

I was having the same issue described in #5902 where when using filesystem triggers to add or remove resources, it would sometimes race and leave the Tiltfile in the pending status without the most recent state.

I believe I successfully tracked this down to the filewatch controller when reconciled can replace the watcher object with a new configuration, but doesn't preserve previous file events when doing so. This can then race with the Reconcile on the tiltfile to miss that there is a pending filesystem change and instead see an empty list of filesystem changes (https://github.com/tilt-dev/tilt/blob/59f9aa24c86dc5302772c61905275fd90675596f/internal/controllers/core/tiltfile/reconciler.go#L235-L240). The Tiltfile reconciler then does not trigger a subsequent build.

### DIFF
--- a/internal/controllers/core/filewatch/controller.go
+++ b/internal/controllers/core/filewatch/controller.go
@@ -190,6 +190,10 @@ func (c *Controller) addOrReplace(ctx context.Context, name types.NamespacedName
 		w.restartBackoff = existing.restartBackoff
 		status.Error = existing.status.Error
 	}
+	if hasExisting {
+		status.FileEvents = existing.status.FileEvents
+		status.LastEventTime = existing.status.LastEventTime
+	}
 
 	ignoreMatcher := ignore.CreateFileChangeFilter(fw.Spec.Ignores)
 	startFileChangeLoop := false
--- a/internal/controllers/core/filewatch/controller_test.go
+++ b/internal/controllers/core/filewatch/controller_test.go
@@ -324,13 +324,14 @@ func TestController_Reconcile_Watches(t *testing.T) {
 	assert.Truef(t, updatedStart.After(originalStart),
 		"Monitor start time should be more recent after update, (original: %s, updated: %s)",
 		originalStart, updatedStart)
-	if assert.Equal(t, 2, len(updated.Status.FileEvents)) {
+	if assert.Equal(t, 3, len(updated.Status.FileEvents)) {
 		// ensure ONLY the expected files were seen
 		assert.NotZero(t, updated.Status.FileEvents[0].Time.Time)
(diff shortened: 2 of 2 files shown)


knisbet:
I will note I'm not that familiar with the code base, so Sorry if there is an implication I'm missing with this change, or a better way to address the issue.

I will also note I think there is another race still present, but only impacting the Status as reported to the UI. The WebUI will show the (Tiltfile) status as pending, even though it has actually processed all the pending filesystem events. I don't believe it's caused by the same problem as this PR, but wanted to mention it just in case I'm missing something.

nicks:
Thanks for sending this change - I think this is the right change conceptually, but is also a fairly low-level change, and need some time to test out some cases I'm worried about.

nicks (approved):
lgtm!

Issue #6445, opened 2024-10-08
websocket-based portforwards


nicks:
## Describe the Feature You Want
In Kubernetes 1.31, Websocket-based portforwards went GA
https://kubernetes.io/blog/2024/08/20/websockets-transition/

### Current Behavior
Tilt still uses SPDY-based portforwards

### Why Do You Want This?
websockets should have broader support in most proxies

PR #6496, opened 2025-02-04
k8s: add websocket support to exec

ChyZma:
In version 1.31, Kubernetes switched from SPDY to WebSockets for streaming ([blog post](https://kubernetes.io/blog/2024/08/20/websockets-transition/)).

For clusters where SPDY is unavailable (e.g., the Kubernetes API server is accessed through an edge proxy that doesn't support SPDY), the `exec` operation must use WebSockets.
A possible solution for this would be to use a WebSocket Stream Executor by default and fall back to a SPDY Stream Executor when WebSocket based streaming is unavailable. (this also aligns with what [`kubectl exec`](https://github.com/kubernetes/kubectl/blob/316ed014ab225ba07735ebc18e05dca602bcaf5c/pkg/cmd/exec/exec.go#L138) currently does)

I also replaced `Executor.Stream` with `Executor.StreamWithContext` because the former is deprecated.

(A similar solution could be used to solve https://github.com/tilt-dev/tilt/issues/6445. I might check it out over the weekend.)

### DIFF
--- a/internal/k8s/exec.go
+++ b/internal/k8s/exec.go
@@ -7,13 +7,14 @@ import (
 	"io"
 
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/util/httpstream"
 	"k8s.io/client-go/tools/remotecommand"
 	"k8s.io/kubectl/pkg/scheme"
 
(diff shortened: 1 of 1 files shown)


nicks (approved):
LGTM!

awesome, thanks!

PR #6499, opened 2025-02-07
tiltfile: Add disable_push option to build_docker function

### LINKED ISSUE #1730: docker_build configuration for no-push environments

cm226:
Added a disable_push flag to the docker_build function, we wanted this for our CI environment. 
We use minikube in our CI deployment, and share a docker socket with the build environment, so no need to push images. 

Went with a flag per image, because I could "sort-of" imagine cases where it might be useful to be able have different push behavior per image, although willing to be corrected on that

resolves #1730


nicks:
hello! i'm a little bit confused by the bug report -- tilt should automatically detect when minikube is exposing it's docker socket, and skip the push in this case. (note that the issue you linked to mentions this). can you share the output of `tilt doctor` and how your minikube env is set up?

for background -- 

the comment you linked to was written 5 years ago, so we've learned a lot since then! i think most of us agree that adding disable_push to custom_build was a mistake. it's difficult to use correctly, and then people have to add more workarounds to fix the problems it creates. so these days we lean more towards reading the configuration off the cluster, rather than adding it to the tiltfile.

nicks:
(i'm pretty sure tilt uses other ways of detecting that the cluster is minikube that will work even if the cluster is renamed, but changing the name will at least test if our hypothesis is correct)

what does `tilt doctor` say?

cm226:
Hi, Sorry for the delay, 

Ok, can confirm that prefixing the cluster name with "minikube" does indeed work. 

`tilt doctor` output for cluster name `cluster-1`


> Tilt: v0.33.22, built 2025-01-03
> System: linux-amd64
> ---
> Docker
> - Host: unix:///var/run/docker.sock
> - Server Version: 27.4.1
> - API Version: 1.47
> - Builder: 2
> - Compose Version: v2.32.4
> ---
> Kubernetes
> - Env: unknown
> - Context: cluster-1
> - Cluster Name: cluster-1
> - Namespace: default
> - Container Runtime: docker
> - Version: v1.25.0
> - Cluster Local Registry: none
> ---
> Thanks for seeing the Tilt Doctor!
> Please send the info above when filing bug reports. 💗
> 
> The info below helps us understand how you're using Tilt so we can improve,
> but is not required to ask for help.
> ---
> Analytics Settings
> --> (These results reflect your personal opt in/out status and may be overridden by an `analytics_settings` call in your Tiltfile)
> - User Mode: default
> - Machine: anon
> - Repo: bhXRsbPmrtJv2TLZmwb8sA==




And this is the output of `tilt doctor` when using `minikube-cluster-1` 


> Tilt: v0.33.22, built 2025-01-03
> System: linux-amd64
> ---
> Docker (cluster)
> - Host: tcp://192.168.49.2:2376
> - Server Version: 20.10.17
> - API Version: 1.41
> - Builder: 2
> ---
> Docker (local)
> - Host: unix:///var/run/docker.sock
> - Server Version: 27.4.1
> - Version: 1.47
> - Builder: 2
> - Compose Version: v2.32.4
> ---
> Kubernetes
> - Env: minikube
> - Context: minikube-cluster-1
> - Cluster Name: minikube-cluster-1
> - Namespace: default
> - Container Runtime: docker
> - Version: v1.25.0
> - Cluster Local Registry: none
> ---
> Thanks for seeing the Tilt Doctor!
> Please send the info above when filing bug reports. 💗
> 
> The info below helps us understand how you're using Tilt so we can improve,
> but is not required to ask for help.
> ---
> Analytics Settings
> --> (These results reflect your personal opt in/out status and may be overridden by an `analytics_settings` call in your Tiltfile)
> - User Mode: default
> - Machine: anon
> - Repo: bhXRsbPmrtJv2TLZmwb8sA==


I guess whatever we are doing with our config is messing with tilts other methods for auto detecting minikube. 

If your interested ill try briefly describe our setup

host machine is running 
    - docker 
        - minikube cluster
        - container
            - tilt ci
            - (host docker socket volume mapped into container)
            - configured to point at minikube running on host

Since prefixing the cluster with "minikube" works we can use that. 

Thanks

cm226:
Also I'm happy to do more digging into this if you want to try to understand why the automatic detection is not working. If you want to point me at the code that should be detecting minikube. 

Thanks

nicks:
@cm226 :+1: the bits are here if you want to poke around - 

minikube detection:
https://github.com/tilt-dev/clusterid/blob/4a1ff3259ff9caa1a30df03cf6353369de4236de/product.go#L58

socket detection:
https://github.com/tilt-dev/tilt/blob/d4bd91e24d5058d32f3b995c260371a45ea35194/internal/docker/env.go#L222

PR #6507, opened 2025-02-23
web: update some yarn deps

nicks:


nicks:
i also removed the old Tiltfile while i was here

PR #6513, opened 2025-03-03
feat(cli): allow specifying log filters

georgijd-form3:
Filter CLI logs by span prefix. Multiple filters can be specified and
exclusion takes precedence.

For example:

- to print only pod logs: `tilt up --log-filters 'pod'`
- to exclude pod logs:  `tilt up --log-filters '!pod'`
- to print only pod and build logs: `tilt up --log-filters 'pod' --log-filters 'build'` or `tilt up --log-filters 'pod,build'`
- to exclude pod and build logs: `tilt up --log-filters '!pod' --log-filters '!build'` or `tilt up --log-filters '!pod,!build'`
- to exclude all pod logs except one: `tilt up --log-filters '!pod' --log-filters 'pod:my-namespace:my-pod'` or `tilt up --log-filters '!pod,pod:my-namespace:my-pod'`

Also fix a golangci-lint (errcheck) error which was not introduced by this change.


mateusz-drzymala-form3:
+1 Great idea ! 
This is very useful especially combined with the `tilt ci` command. It's very hard to debug issues when there are many resources. Restricting the resources that can print logs to stdout  will make our life much easier !

nicks (commented):
hi! thanks for the contribution!

i like the overall motivation and direction

i think my major concern with this PR is that we already have a log filtering system in the UI. This system is similar but subtly different. 

What do you think of modifying it to work more like the UI's filtering system? in other words: filter by log type (all vs runtime vs build), filter by resource

span ids were always kind of an internal implementation detail that's not exposed anywhere else, so filtering by span id seems kinda weird to me.

nicks (approved):
lgtm!

PR #6534, opened 2025-04-08
When calculating Kustomize dependencies, include components

borntyping:
When calling `kustomize(<directory>)` in a Tiltfile, the directory is watched as well as any paths the Kustomization includes. This doesn't currently include components, so Kustomizations that include [components](https://github.com/kubernetes-sigs/kustomize/blob/master/examples/components.md) don't trigger a reload of the Tiltfile and the Kustomization when files in those components change.

This PR adds the `components:` field of a Kustomization to the list of paths `dependenciesForKustomization` will inspect.


nicks (approved):
LGTM!

PR #6569, opened 2025-06-13
portforward: Allow websockets to be used for port-forwarding

### LINKED ISSUE #6568: Websockets don't work when trying to port-forward
### Expected Behavior

A port-forward that requires websockets should be able to successfully connect and forward traffic.

Doing the same port-forward with the `kubectl port-forward` command works as expected.

### Current Behavior

Currently the connection fails with an "Upgrade connection request" error.

### Steps to Reproduce

1. 
2.
3.
4.

#### `tilt doctor` Output

```
$ tilt doctor
...
```

#### About Your Use Case

We are running vcluster to have multiple clusters running on a single physical cluster. We have placed an Envoy proxy in-front of the vcluster API proxies to route traffic over L7. The `kubectl port-forward` command works without issue and upgrades to a websockets connection. The tilt code is still trying to use spdy which Envoyproxy doesn't support so the code needs to be updated to fix this issue.

ahamilton55:
The current implementation of port-forwarding only supports the spdy protocol. Newer API servers can handle websockets and fall back to spdy if they aren't available.

This updates the code to use websockets for port forwarding when available and to fall back to spdy if it needs to.

Fixes #6568.


ahamilton55:
@nicks This test failed in CI but it doesn't really seem to be related to the change. I tried running it again but it failed again. I ran the tests locally and they all passed so I'm not sure what the issue is.

PR #6593, opened 2025-08-13
Add safe argument to `local` function

### LINKED ISSUE #3066: Question: Is there a way to skip the kubernetes context check?
Right now I have the following Tiltfile, which consists entirely of `local_resource` declarations to run go tests for certain packages

```python
# Tiltfile
local_resource('go tests', cmd='go test -v ./pkg/...', deps=['./pkg/'])

```

But when I `tilt up`, it notices I am using a GKE kube context and (rightly) prevents startup:

```
       gke_smart-proxy-839_us-central1-b_dex-kots-tester might be a production     
       kube context.                                                               
       If you're sure you want to continue add:                                    
       allow_k8s_contexts('gke_smart-proxy-839_us-central1-b_dex-kots-tester')     
       before this function call in your Tiltfile. Otherwise, switch k8s           
       contexts and restart Tilt.  
```

And while I can certainly do this, I'm wondering if there's a way to automatically skip the context check if all Tiltfile invocations are `local_resource` and there is no `k8s_yaml` or `docker_build`. 

(note -- I'm not at all blocked on this, I can certainly switch my context to `docker-for-desktop` temporarily while using this Tiltfile)

```
Tilt: v0.12.7, built 2020-03-09
System: darwin-amd64
---
Docker
- Host: [default]
- Version: 1.40
- Builder: 2
---
Kubernetes
- Env: gke
- Context: gke_smart-proxy-839_us-central1-b_dex-kots-tester
- Cluster Name: gke_smart-proxy-839_us-central1-b_dex-kots-tester
- Namespace: default
- Container Runtime: docker
- Version: v1.14.10-gke.17
---

---
Analytics Settings
- User Mode: opt-in
- Machine: 503519e6fa9251f8c1b296d57e80c78c
- Repo: j+D3yI+Wl4UG3HEcwDSrdw==
```

jtamagnan:
## Context

Fixes #3066

## Problem

There are some functions that use local that I would like to run when a file is first loaded so that I can cache the output. If I want to load this file I need to have first allowed the k8s context which feels off. Stylistically, I'd like to have all of my imports at the top of my files before needing to execute any code like `allow_k8s_contexts`. Right now something like this would fail:
```
# File: /Tiltfile
load("/subdir/Tiltfile", cached_value)
load(...)
load(...)

allow_k8s_contexts(...)
```

```
# File: /subdir/Tiltfile
cached_value = local(...)
```

## Solution

My proposal is to add a new `safe` argument to the `local` function. This argument pretty much means "Hey take my word for it, I know that I want to run this regardless of the k8s context that is active".

The second file could now look like:
```
# /subdir/Tiltfile
cached_value = local(..., safe=True)
```

## Implementation

I had a few ideas for this. 
1. Create a new local_safe function.
2. Modify potentiallyK8sUnsafeBuiltin to iterate through the args+kwargs to detect if `safe` was passed in
3. :star: This is the one I've chosen. Get rid of  `potentiallyK8sUnsafeBuiltin` and move that logic into `local` since it is the only function that uses this. Then let all the arg parsing happen there

## Testing Done

Existing tests pass and I've added some new tests that exercise the `safe` functionality


jtamagnan:
I've looked into the one failing tests I see and I can't reproduce it locally on MacOs. Should do a simple force push to retrigger that test? I'd be surprised if it were related to my changes

nicks:
ah...

so for background,`local()` runs on every execution of a tiltfile. the goal of `local` was to handle cases where a user's config is _mostly_ declarative, but they have some shell script that generates their yaml.

we've always been pretty strongly opposed to using the Tiltfile as a "poor man's bash", where we have procedural features that depend on the order of operations. The reconciliation model just really isn't well-suited for that. this doc talks about this in more detail - https://docs.tilt.dev/controlloop.html

So we probably wouldn't accept any API that "allowlisted" local() calls to interact differently with the tilt engine like this. 

it also seems like you already have a pretty good solution - put `allow_k8s_contexts(...)` above the `load()` calls

jtamagnan:
@nicks, thanks for the super quick review and feedback. I'll jot down some remaining ideas about this because I do still see value in it but understand not wanting this feature.

> so for background,local() runs on every execution of a tiltfile. the goal of local was to handle cases where a user's config is mostly declarative, but they have some shell script that generates their yaml.

Ahh okay, I didn't realize that that was the intent. I was using logic like what was shared in the [tilt-example-bazel](https://github.com/tilt-dev/tilt-example-bazel/blob/main/3-recommended/Tiltfile#L27-L39) repository. In this case we aren't generating yaml.

> we've always been pretty strongly opposed to using the Tiltfile as a "poor man's bash"

I thought that by following the provided examples we would be staying within the vision of the project. 

> where we have procedural features that depend on the order of operations.

In a way I feel like the addition of a `safe` parameter would resolve issues around order of execution. One wouldn't need to perform code like `allow_k8s_context()` before certain `load` calls. This feels nice to me probably because of the auto-formatters that I have and a general personal preference in reducing side effects from load calls; doesn't mean its right though.

> it also seems like you already have a pretty good solution - put allow_k8s_contexts(...) above the load() calls

Sure I can go along with that. I'll place the `allow_k8s_contexts()` calls in a different file and make sure that it is the first imported file. Having code act differently depending on import order feels off and this pull-request would alleviate that issue but maybe it is my expectations that are off.

Again thank you for taking the time to look at the code, read this reply, and all your work on this project.

PR #6594, opened 2025-08-13
Docker delete volumes

SeanMBrennan:
Add '--delete-volumes' option to `tilt down` that removes volumes for docker compose projects only.

`docker compose down` does not remove internal volumes without a '-v|--volumes' option. This change adds that option if `tilt down` is called with an additional '--delete-volumes' option.

Tested and working as expected.

### DIFF
--- a/internal/cli/down.go
+++ b/internal/cli/down.go
@@ -24,7 +24,7 @@ import (
 type downCmd struct {
 	fileName         string
 	deleteNamespaces bool
-	deleteVolumes bool
+	deleteVolumes    bool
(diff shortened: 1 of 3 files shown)


SeanMBrennan:
~~Looks like CI tests failed on a spurious timeout, but I cannot re-run the test pipeline.~~

nvm, force pushed another rebase to trigger the CI pipeline.

nicks:
i like this proposal but now everyone is going to complain that --delete-volumes=false will delete kubernetes volumes. but maybe someone can do that later :joy:

nicks:
LGTM!

PR #6599, opened 2025-08-16
liveupdate: add debug log when no containers match selector in maybeSync

### LINKED ISSUE #6598: Tilt stuck in pending status indefinitely due to external kubernetes image caching prefix
When Tilt cannot find any matching containers for a LiveUpdate resource during reconciliation, the UI will remain stuck in `"Pending"` indefinitely, and there is currently no log output indicating why.

After debugging, I discovered that Tilt was comparing the built image reference:

```
registry.com/image
```

against the running container’s image reference:

```
localhost:7879/registry.com/image
```

The extra `localhost:7879` prefix came from [Kuik](https://github.com/enix/kube-image-keeper) image rewriting in my cluster. Because the references didn’t match, Tilt silently continued without applying the update.

The only workaround was to trigger a full build instead of a live update. I eventually fixed it by disabling Kuik/caching in my Helm charts.

I understand this is designed this way to avoid the pending/restarting containers or containers that are out of liveupdate scope but the lack of any logging made diagnosis much slower; getting no hints through `tilt doctor` or `tilt dump`.

### Expected Behavior
Tilt should should have a debug log about the names mismatch during an attempted liveupdate instead of being stuck in "Pending"

### Current Behavior
Tilt UI will stay stuck in "Pending" with no logs or debug level logs showing why the live update may be stuck.

### Steps to Reproduce
1. Install a container caching system like [Kuik](https://github.com/enix/kube-image-keeper) to the k8s cluster
2. Deploy your resources using Tilt
3. Try doing any changes to trigger the live updates
4. Tilt UI will just be stuck in "Pending" due to the name mismatch

Image

eqladios:
When no containers match the selector during live update reconciliation, Tilt currently continues silently. This will result in the Tilt UI showing "Pending" indefinitely, with no feedback about why no update is happening.

This change adds a debug-level log in `maybeSync()` when its `visitSelectedContainers` pass never matches any container, so user can have a clue what might be happening.

The intent is to give users a clue that no containers matched, so they can investigate possible causes such as:

- Image reference mismatches (e.g., rewritten by a local registry)
- Containers not yet started or already terminated

This can be more detailed to show each comparison as debug logs but I kept it minimal to just have a debug log if `visitSelectedContainers` never "visited" any container.

fixes #6598

### DIFF
--- a/internal/controllers/core/liveupdate/reconciler.go
+++ b/internal/controllers/core/liveupdate/reconciler.go
@@ -658,7 +658,11 @@ func (r *Reconciler) maybeSync(ctx context.Context, lu *v1alpha1.LiveUpdate, mon
 	// Visit all containers, apply changes, and return their statuses.
 	terminatedContainerPodName := ""
 	hasAnyFilesToSync := false
+	// Track if the anonymous function in visitSelectedContainers was ever invoked - for debug logs.
+	visitedAny := false
 	resource.visitSelectedContainers(func(pod v1alpha1.Pod, cInfo v1alpha1.Container) bool {
+		visitedAny = true
+
 		c := liveupdates.Container{
 			ContainerID:   container.ID(cInfo.ID),
 			ContainerName: container.Name(cInfo.Name),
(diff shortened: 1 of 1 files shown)


nicks:
i'm not sure this is a good idea -- the reconcilation loop runs continuously, and there's lots of totally normal reasons why the container might not have shown up yet.

PR #6606, opened 2025-08-25
store: initialize BuildStatus at startup

nicks:
Before this change, the engine would initialize
BuildStatus lazily.

This was leading to race conditions.
The event that consumed a file change
would sometimes appear before the file change event,
and would get ignored because BuildStatus
hadn't been initialized yet.


nicksieger (approved):
LGTM

Issue #5266, opened 2021-12-09
set kubernetes context from tiltfile


nicks:
## Describe the Feature You Want

Your tiltfile should be able to specify which kubernetes context to switch to.

e.g.,

```
k8s_context('minikube')
```

If you edit the k8s_context call in the tiltfile during a tilt session, tilt should disconnect from that cluster and re-connect to the new one.

### Why Do You Want This?

This would be a genuinely useful feature on its own! But we also expect this to be a prerequisite for a number of other cluster-management features, like
- onboarding flows that help you setup a cluster
- diagnosing a broken cluster
- resetting a cluster

**Additional context**

Here are some related issues that get at the same thing:
https://github.com/tilt-dev/tilt/issues/1113 (Set the default namespace you're deploying to)
https://github.com/tilt-dev/tilt/issues/2123 (reload the kubeconfig when the context changes)
https://github.com/tilt-dev/tilt/issues/3791 (set up the kubeconfig during tiltfile initialization)
https://github.com/tilt-dev/tilt/issues/3066 (don't connect to kubernetes until you need it)

nzagorsky:
Up for this. That might help to have multiple tilt instances running at once (on top of TILT_PORT env var).  Some flexibility here would be greatly appreciated.

nicks:
@cmoulliard if you set the KUBECONFIG env variable, I believe tilt will respect that

lars-solberg:
I'm currently evaluating using Tilt for developing both applications spanning multiple clusters, but also the platform itself, things like how argocd should work and that kind.

Since Tilt is not context aware yet, I need to specify this manually and using local_resource's. What would work for my usecase is to have an additional context flag on example k8s_yaml, like

```
k8s_yaml('k8s/app.yaml', k8s_context='kind-cl2')
```

PR #6623, opened 2025-10-20
Implemented kubeconfig isolation and effective `k8s_context()` context change

hectorgarcia-cw:
Fixes [#5266](https://github.com/tilt-dev/tilt/issues/5266)

### Summary

This PR implements missing `k8s_context()` functionality and adds kubeconfig isolation. Previously, `k8s_context()` existed but didn't switch contexts only returned current k8s context. It implements three key improvements:

1. **Context Switching Implementation**: `k8s_context()` now can switch the context that Tilt uses for operations
2. **Kubeconfig Isolation**: Tilt operates on a private copy of kubeconfig, leaving system kubectl context unchanged
3. **Command Order Validation**: Static analysis prevents `k8s_context()` calls after Kubernetes operations have begun

### Implementation

**Kubeconfig Isolation:**
- Creates temporary kubeconfig copy using `KUBECONFIG` environment variable override
- Maintains thread-local context tracking for immediate response to successive calls
- Integrates with existing dependency injection system via new `ConnectionProvider` interface

**Static Validation:**
- Uses AST parsing with `go.starlark.net/syntax` to analyze Tiltfile structure before execution
- Validates command order across all included files via `OnExecPlugin` interface  
- Allows read-only `k8s_context()` calls (without parameters) anywhere in Tiltfiles

### Behavior Changes

**Before:**
- `k8s_context('production')` was accepted but had no effect - Tilt continued using the default context
- All Kubernetes operations used the system's current kubectl context
- No validation or error when context switching was attempted

```tiltfile
k8s_context('production')  # Silently ignored
k8s_yaml('app.yaml')       # Always deployed to default context
```

**After:**
- `k8s_context('production')` actually switches Tilt to use the specified context
- Tilt uses an isolated copy of kubeconfig, never modifying system kubectl context
- Context must be set before any Kubernetes operations
- Clear validation errors for incorrect usage

```tiltfile
k8s_context('production')  # Actually switches to production context
k8s_yaml('app.yaml')       # Deploys to production cluster
```

Invalid usage now fails with clear error:
```tiltfile
k8s_yaml('app.yaml')
k8s_context('production')  # ❌ Will fail validation
```

```
ERROR: k8s_context() with parameters cannot be called after Kubernetes-interacting commands. 
Set the Kubernetes context before using commands like local(), k8s_yaml(), helm(), kustomize(), etc.
Error at Tiltfile:2:1
```

### Compatibility

- **New Functionality**: `k8s_context()` now can modify context
- **Breaking Change**: Tiltfiles with `k8s_context()` using parameter called after `k8s_yaml()`, `helm()`, etc. will fail validation
- **System Impact**: User's kubectl context remains completely isolated from Tilt operations

### Files Modified

- `internal/tiltfile/k8scontext/k8scontext.go` - Core isolation implementation
- `internal/tiltfile/tiltfile_state.go` - Static validation system  
- `internal/tiltfile/k8scontext/cache.go` - Cache invalidation
- `internal/controllers/core/cluster/` - Dynamic client integration
- `internal/cli/wire.go` - Dependency injection updates
- Test files - Comprehensive validation test suite


nicks:
hmm...this pr doesn't work -- it doesn't even touch most of the places where tilt establishes a k8s connection.

the problem we hit is that a lot of the underlying core k8s libraries -- e.g., informers, discovery, etc -- really insist on a single k8s apiserver connection, and don't respond well to the connection parameters changing at runtime.

We determined a long time ago that trying to change the connection at runtime is technically infeasible to do robustly without a major rearchitecture of tilt. sorry

PR #6643, opened 2025-11-19
fix(tilt): restart pod watch on channel closure to prevent hanging po…

jld-adriano:
…rt-forwards

When the pod watch channel closes (due to network issues, pod crashes, or cluster events), the dispatchPodChangesLoop now:
1. Logs the closure with namespace and cluster info
2. Triggers reconciliation for all watchers in the namespace
3. Cleans up the watch entry so it can be recreated

This prevents port-forwards from hanging indefinitely when pods restart, as the watch will be recreated during reconciliation and pick up the new pod.

Fixes: GIGA-1010


jld-adriano:
this was meant to be on our fork, my bad

Issue #6659, opened 2026-01-11
Serialized Live Update for containers with different images


rafaelgallani:
### Expected Behavior
`tilt` should perform its Live Update in parallel across all containers affected by the changes.

### Current Behavior
`tilt` performs it in a serialized way, one by one.

### Steps to Reproduce

1. Clone this repo: https://github.com/rafaelgallani/tilt-live-update-serialized
2. Run the scripts according to its [README file](https://github.com/rafaelgallani/tilt-live-update-serialized/blob/main/README.md)
3. Or... If you want to do it manually:
  - Create a test project where you have lots of services (so you have a use-case similar to mine)
  - Then, make it so that all those services have common dependencies
  - Set up a `Tiltfile` with `live_update` on those common dependencies - similar to the example repo one
  - Once all services are built/done, modify any common dependency file, and ensure you see the behavior reported.

#### `tilt doctor` Output

```
$ tilt doctor
Tilt: v0.35.0, built 2025-06-13
System: darwin-arm64
---
Docker
- Host: unix:///Users/rafaelgallani/.docker/run/docker.sock
- Server Version: 28.0.4
- API Version: 1.48
- Builder: 2
- Compose Version: v2.34.0-desktop.1
---
Kubernetes
- Env: k3d
- Context: k3d-opusdns-api
- Cluster Name: k3d-opusdns-api
- Namespace: default
- Container Runtime: containerd
- Version: v1.31.5+k3s1
- Cluster Local Registry: &RegistryHosting{Host:127.0.0.1:51318,HostFromClusterNetwork:ctlptl-k3d-registry:5000,HostFromContainerRuntime:ctlptl-k3d-registry:5000,Help:https://k3d.io/stable/usage/registries/#using-a-local-registry,SingleName:,}
---
Thanks for seeing the Tilt Doctor!
Please send the info above when filing bug reports. 💗
```

#### About Your Use Case
I have a setup where all the services are going through a full `docker_build` every time. 
Most of those services share some dependencies (I know this can be discussed/modified, but still).

I was trying to modify the setup in such a way that, instead of running full `docker_build`s everytime, it performed [Live Updates](https://docs.tilt.dev/live_update_reference.html) instead.

Then I ran into what I'm reporting here - the update of all these services were serialized, and it was even slower than performing it as before.

rafaelgallani:
Somewhat similar to:
- #4511

but not all that much - since this issue refers to **different** images.

[...]

Issue #4511, opened 2021-05-04
feat: parallel sync with live update


MasayaAoyama:
## Describe the Feature You Want

tilt sync files in `parallel` using the same image.

### Current Behavior

tilt sync files in `serial` using the same image.

### Why Do You Want This?

In our environment, a container image is used for several deployments.
We use tilt with live sync, but tilt sync files in serial to several deployments using the same image.
Do you have any idea or plan to sync files in parallel using the same image?

## test result

* test files
    * https://github.com/MasayaAoyama/tilt/commit/4e7a0072c5963136fe231801992e8a0ec429a5f0

Run Deployments `test1` and `test2` with the same image `nginx:1.18` by tilt.

```
cd tilttest
tilt up &
```

And then, trigger live sync for both Deployments `test1` `test2`.

```
touch ./testdata
```

and then, show live sync progress from the logfile.
* tilt have synced to `test1` pod between 17:37:29 - 17:37:36.
* tilt have synced to `test2` pod between 17:37:36 - 17:37:43.

```
# kubectl exec -it test1-5786f6b9d6-vgwc5 -- cat /root/log
Tue May  4 17:37:28 UTC 2021
0
Tue May  4 17:37:29 UTC 2021
9
Tue May  4 17:37:30 UTC 2021
55
Tue May  4 17:37:31 UTC 2021
90
Tue May  4 17:37:32 UTC 2021
132
Tue May  4 17:37:33 UTC 2021
175
Tue May  4 17:37:34 UTC 2021
211
Tue May  4 17:37:35 UTC 2021
251
Tue May  4 17:37:36 UTC 2021
255
Tue May  4 17:37:37 UTC 2021
255
```

```
# kubectl exec -it test2-56ccc6b89f-bbjs6 -- cat /root/log
Tue May  4 17:37:28 UTC 2021
0
Tue May  4 17:37:29 UTC 2021
0
Tue May  4 17:37:30 UTC 2021
0
Tue May  4 17:37:31 UTC 2021
0
Tue May  4 17:37:32 UTC 2021
0
Tue May  4 17:37:33 UTC 2021
0
Tue May  4 17:37:34 UTC 2021
0
Tue May  4 17:37:35 UTC 2021
0
Tue May  4 17:37:36 UTC 2021
43
Tue May  4 17:37:37 UTC 2021
91
Tue May  4 17:37:39 UTC 2021
136
Tue May  4 17:37:40 UTC 2021
184
[...]

PR #6661, opened 2026-01-15
Allow parallel execution of live updates

rafaelgallani:
related to: 
- #4511
- #6659 

summary:
- removed the lock from `Reconcile` so we can run live updates in parallel
- added some descriptions around the newly added locks... 
  - used basically for functions where maps are used
  - locks those whose had to be moved to the functions instead, since now `Reconcile` doesn't have them anymore

### DIFF
--- a/internal/controllers/core/liveupdate/reconciler.go
+++ b/internal/controllers/core/liveupdate/reconciler.go
@@ -254,7 +254,8 @@ func (r *Reconciler) handleFailure(ctx context.Context, lu *v1alpha1.LiveUpdate,
 	return ctrl.Result{}, err
 }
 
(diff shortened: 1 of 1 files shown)


rafaelgallani:
Please lmk if that's good... I also made it so that it shares the same `max_parallel_updates` setting used elsewhere too - as initially I ran into this since I thought that this setting should be shared by both regular builds and the live update.

rafaelgallani:
idk if/how this should have tests for too... I mean. since we'd be changing the intended way of how it works. anyways.. just lmk anything really 😬

rafaelgallani:
still not fully working - will do some more work locally and make it ready for review once again

PR #6662, opened 2026-01-15
Fix service grouping across namespaces

### LINKED ISSUE #6311: service grouping across namespaces is wrong
Apply this yaml:

```
apiVersion: apps/v1
kind: StatefulSet
metadata:
  name: example-html
  labels:
    app: example-html
spec:
  selector:
    matchLabels:
      app: example-html
  template:
    metadata:
      labels:
        app: example-html
    spec:
      containers:
      - name: example-html
        image: example-html-image
        ports:
        - containerPort: 8000
---
apiVersion: v1
kind: Service
metadata:
  name: example-html
  labels:
    app: example-html
spec:
  selector:
    app: example-html
  ports:
    - protocol: TCP
      port: 8000
      targetPort: 8000
---
apiVersion: v1
kind: Namespace
metadata:
  name: ns2
---
apiVersion: apps/v1
kind: StatefulSet
metadata:
  name: example-html
  namespace: ns2
  labels:
    app: example-html
spec:
  selector:
    matchLabels:
      app: example-html
  template:
    metadata:
      labels:
        app: example-html
    spec:
      containers:
      - name: example-html
        image: example-html-image
        ports:
        - containerPort: 8000

---
apiVersion: v1
kind: Service
metadata:
  name: example-html
  namespace: ns2
  labels:
    app: example-html
spec:
  selector:
    app: example-html
  ports:
    - protocol: TCP
      port: 8000
      targetPort: 8000

```

Expected behavior:
Two different resources, each with 1 StatefulSet and 1 Service

Actual behavior:
Two resources, one with 2 services and one with 0

imxuyou:
## Summary

  Fix service grouping across namespaces (#6311)

  ## Problem

  When multiple namespaces contain same-named Services and Deployments (e.g., `shop-phoenix` in both `local` and `upfit` namespaces), Tilt incorrectly groups all Services under one workload, leaving other workloads without their Services deployed.

  **Root Cause:** `FilterByMatchesPodTemplateSpec` in `internal/k8s/entity.go` only checks if selector labels match, without considering namespace boundaries.

  ## Solution

  Modified `FilterByMatchesPodTemplateSpec` to also check that entities are in the same namespace before associating them:

  ```go
  // Before: only checks selector labels
  match, rest, err := FilterBySelectorMatchesLabels(remaining, template.Labels)

  // After: checks both selector labels AND same namespace
  match, rest, err := filterBySelectorMatchesLabelsAndNamespace(remaining, template.Labels, workloadNamespace)

  Testing

  - Tested with multiple namespaces containing same-named resources
  - Verified Services are now correctly associated with their respective Deployments in each namespace
  - HTTPRoute correctly resolves Service references after the fix

  Fixes #6311

### DIFF
--- a/internal/k8s/entity.go
+++ b/internal/k8s/entity.go
@@ -232,10 +232,6 @@ func FilterByImage(entities []K8sEntity, img container.RefSelector, locators []I
 	return Filter(entities, func(e K8sEntity) (bool, error) { return e.HasImage(img, locators, inEnvVars) })
 }
 
-func FilterBySelectorMatchesLabels(entities []K8sEntity, labels map[string]string) (passing, rest []K8sEntity, err error) {
-	return Filter(entities, func(e K8sEntity) (bool, error) { return e.SelectorMatchesLabels(labels), nil })
-}
-
 func FilterByMetadataLabels(entities []K8sEntity, labels map[string]string) (passing, rest []K8sEntity, err error) {
 	return Filter(entities, func(e K8sEntity) (bool, error) { return e.MatchesMetadataLabels(labels) })
 }
--- a/internal/k8s/entity_test.go
+++ b/internal/k8s/entity_test.go
@@ -311,3 +311,115 @@ func mustParseYAML(t *testing.T, yaml string) []K8sEntity {
 	}
 	return entities
 }
+
+// TestFilterByMatchesPodTemplateSpecNamespace tests that Services are only matched
+// to Deployments in the same namespace. This is a regression test for:
+// https://github.com/tilt-dev/tilt/issues/6311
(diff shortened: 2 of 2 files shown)


imxuyou:
> thanks! can you also add a test for this, to ensure it doesn't regress?

  Thanks for the review! I've added the changes:

  1. Added TestFilterByMatchesPodTemplateSpecNamespace with 4 test cases covering:
    - Service in different namespace should not match
    - Service in same namespace should match
    - Service without namespace should match (backward compatibility)
    - Mixed scenario selecting only matching namespace
  2. Removed the unused FilterBySelectorMatchesLabels function as suggested.

Issue #6664, opened 2026-01-17
Different readiness_timeout for long-running jobs?


mterwill:
My organization heavily relies on Tilt for both interactive development as well as automated testing using Tilt CI. The [readiness_timeout](https://docs.tilt.dev/api.html#api.ci_settings) setting has been very helpful to both shorten the time to detect a failure as well as accurately attribute what resource failed.

However, tuning it accurately has proven difficult – we'd like to keep it as low as possible, but want to allow integration test jobs (which we also run in Tilt) a longer time window without raising the global readiness timeout.

Any thoughts or recommendations? Thank you in advance!

dnephin:
In https://github.com/tilt-dev/tilt/issues/6560 we were looking for way to apply a timeout to deployments. I didn't realize it was also going to apply to jobs.  The behaviour of `k8s_grace_period` seems correct for jobs. Let the job run, but stop retrying after this period of time.

Could we have `readiness_timeout` only apply to deployments and satefulsets and use `k8s_grace_period` for jobs?  Or some way to opt-out jobs (or potentially any resource) from the `readiness_timeout` ?

nicks:
hmmm...ack, let me poke around a bit at how it works now.

PR #6673, opened 2026-01-28
session: don't apply readiness timeout to jobs

nicks:
fixes https://github.com/tilt-dev/tilt/issues/6664


dnephin (approved):
Thank you!

PR #6680, opened 2026-02-05
feat: enable server side apply for k8s resources

Jaakkonen:
Adds `k8s_server_side_apply` option to `update_settings` accepting values `"true"|"false"|"auto"` and passes the option to K8sClient similarly to how the timeout is passed currently.


nicks:
thanks for opening this pr!

my original hope was something like:
- kubectl would switch to use server-side apply by default
- tilt would switch to use server-side apply by default, without even making it an option
- client-side apply would get removed entirely

but sadly server-side-apply-default has kind of stalled out. i haven't really followed why it got stuck, have you? https://github.com/kubernetes/enhancements/issues/3805

so i'm fine moving forward with making it an option in tilt

BUT

i don't think it should be a property of k8s_yaml - it feels like a setting you'd want to turn on globally

instead, can you make it an option on update_settings()? something like `update_settings(k8s_server_side_apply = 'false' | 'true' | 'auto')` (to match the flags in the upstream design doc)

Jaakkonen:
Thanks for the comments! :)

> i don't think it should be a property of k8s_yaml - it feels like a setting you'd want to turn on globally
>
> instead, can you make it an option on update_settings()? something like update_settings(k8s_server_side_apply = 'false' | 'true' | 'auto') (to match the flags in the upstream design doc)

Agreed! I actually fixed this in a force-pushed change as I realized the relation of k8sEntity and k8sResource in the codebase. The current API change would only be to add a new option to `update_settings` of type `bool` defaulting to false. I'll make it match the 'false' | 'true' | 'auto', and update the PR description.

As this is now a global setting the setting could be checked `K8sClient` reducing the diff quite heavily (Originally passed it as option to do per-resource setting).

> but sadly server-side-apply-default has kind of stalled out. i haven't really followed why it got stuck, have you? https://github.com/kubernetes/enhancements/issues/3805

Don't know unfortunately :/

Jaakkonen:
@nicks Updated the PR and regenerated the generated files to pass CI. Requesting review :)

PR #6682, opened 2026-02-06
fix(dockercompose): respect health check status in RuntimeStatus

ChakshuGautam:
## Summary

- Fix RuntimeStatus() to consider Docker health check status before reporting container as OK
- Return RuntimeStatusPending when health check is "starting"
- Return RuntimeStatusError when health check is "unhealthy"
- Add comprehensive tests for health check status handling

## Problem

When using Tilt with docker-compose services that have health checks configured, Tilt would mark containers as "OK" (green in the Tilt UI) as soon as they entered the "running" state, without waiting for the health check to pass.

This caused issues with docker-compose's `depends_on` with `condition: service_healthy`:

```yaml
services:
  db:
    healthcheck:
      test: ["CMD", "pg_isready"]
      
  app:
    depends_on:
      db:
        condition: service_healthy
```

Even though docker-compose was configured to wait for health checks, Tilt's `resource_deps` would not honor this because Tilt only looks at whether the container is "running", not whether it's "healthy".

## Solution

Modified `RuntimeStatus()` in `internal/dockercompose/state.go` to check `HealthStatus`:

- If `HealthStatus` is empty (no health check configured): behavior unchanged
- If `HealthStatus` is "starting": return `RuntimeStatusPending` 
- If `HealthStatus` is "healthy": return `RuntimeStatusOK`
- If `HealthStatus` is "unhealthy": return `RuntimeStatusError`

The `HealthStatus` field was already being captured and stored in `DockerContainerState` (via `ToContainerState()`), it just wasn't being used in the status determination.

## Test plan

- [x] Added unit tests for all health check status scenarios
- [x] All existing tests pass
- [ ] Manual testing with docker-compose services that have health checks

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/internal/dockercompose/state.go
+++ b/internal/dockercompose/state.go
@@ -47,6 +47,15 @@ func (s State) RuntimeStatus() v1alpha1.RuntimeStatus {
 	if s.ContainerState.Running ||
 		s.ContainerState.Status == ContainerStatusRunning ||
 		s.ContainerState.Status == ContainerStatusExited {
+		// If a health check is configured, wait for it to pass before reporting OK.
+		// HealthStatus will be "starting", "healthy", or "unhealthy" when configured.
+		// An empty HealthStatus means no health check is configured.
+		if s.ContainerState.HealthStatus != "" && s.ContainerState.HealthStatus != "healthy" {
+			if s.ContainerState.HealthStatus == "unhealthy" {
+				return v1alpha1.RuntimeStatusError
+			}
+			return v1alpha1.RuntimeStatusPending
+		}
 		return v1alpha1.RuntimeStatusOK
 	}
 	if s.ContainerState.Status == "" {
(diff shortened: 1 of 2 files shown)


nicks (commented):
this is currently working as designed -- for compose resources, we use service_started: the default depends_on behavior of compose.

changing it to service_healthy just swaps one problem for a bigger problem -- now service_started conditions don't work correctly.

figuring out how to model depends_on conditions in tilt would be a pretty big change, and we'd expect a short spec on how to do it before opening a pr to implement it

ChakshuGautam:
Hey @nicks is Tilt looking for contribution on this front though? We are using Tilt in our organisation with over 30 services and would want this feature. Happy to start on the spec and get feedback, this was just a quick hack through Claude for a deadline but would want to do work on this in the right manner over the next couple of months. Thanks.

nicks:
sure! the place to start is figuring out how to model both depends_on conditions in tilt, and how to model runtime state in a way that downstream resources can have different conditions

Issue #6685, opened 2026-02-07
Copy Console Output Mini Feature


openmindednewby:
## Describe the Feature You Want
Just add a feature with 1 click to copy the console output. Ideal for pasting results to AI assistance for troubleshooting

### Current Behaviour
There is no copy console output button of functionality

### Why Do You Want This?
I regularly find me self-selecting the console, coping it and pasting it to Claude agents

**Additional context**
I recommend placing it here and make it look more or less like the 'Clear Logs'

<img width="1262" height="197" alt="Image" src="https://github.com/user-attachments/assets/43c2a708-5ff4-4025-a3b9-0a08c50c2b71" />

openmindednewby:
I will open a PR for this one :P

openmindednewby:
> I will open a PR for this one :P

Opend PR here https://github.com/tilt-dev/tilt/pull/6686, please review!

PR #6686, opened 2026-02-07
Copy-Console-Output-Mini-Feature

openmindednewby:
Tackles issue https://github.com/tilt-dev/tilt/issues/6685

### DIFF
--- a/web/src/CopyLogs.test.tsx
+++ b/web/src/CopyLogs.test.tsx
@@ -125,9 +125,7 @@ describe("CopyLogs", () => {
     })
 
     // After the timeout, the tooltip transitions out (opacity: 0)
-    expect(
-      screen.queryByText(`Copied ${lineCount} lines`)
-    ).not.toBeVisible()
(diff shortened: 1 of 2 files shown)


nicks:
i like it! i played around with it a bit locally and found two obvious problems --

1 - there's no feedback right now that the button did anything. maybe a simple tooltip like "Copied 3 lines" would work?

2 - the copied text has all the terminal color codes. we should probably remove those?

(as an aside -- if you're just looking for a way for the agent to grab the logs, the agent would probably be happier with the `tilt logs` cli)

openmindednewby:
> i like it! i played around with it a bit locally and found two obvious problems --
> 
> 1 - there's no feedback right now that the button did anything. maybe a simple tooltip like "Copied 3 lines" would work?
> 
> 2 - the copied text has all the terminal color codes. we should probably remove those?
> 
> (as an aside -- if you're just looking for a way for the agent to grab the logs, the agent would probably be happier with the `tilt logs` cli)

Thank you very much! On it now!

openmindednewby:
Please review! @nicks

PR #6690, opened 2026-02-11
webview: flesh out typescript gen

nicks:
- generate types for core api
- always use slices of non-nullable objects

Signed-off-by: Nick Santos <nicholas.j.santos@gmail.com>


nicks:
i also found a bunch of types that appeared to be dead so i deleted them

Issue #4776, opened 2021-07-20
live-update to a CRD without a pod


nicks:
## The Problem

We're seeing more and more frameworks in Kubernetes that use CRDs that don't have images in them. Maybe the CRD builds its own image. Or maybe it forces you to choose from a limited set of pre-existing images.

As an example, let's say you want to live-update files to the pods created by Prometheus CRD: https://github.com/tilt-dev/tilt-example-frameworks/blob/master/prometheus/prometheus.yaml

Tilt's deploy pipeline works like this:
- tilt organizes kubernetes objects into groups
- tilt builds and inject images into the spec yaml, then apply the yaml to the cluster
- tilt tracks which pods belong to which objects
- tilt live-updates the containers that matches the images

Image names are an essential part of all these steps - to select objects to track, then to select containers to inject into.

## Proposal

We mostly have a way to group and track objects without images. It looks like this:

```
k8s_resource(
  new_name='app',
  objects=[...], # object spec
  extra_pod_selectors=[{...}])
```

This mostly works, but is a bit manual (in a lot of cases, it's hard to figure out the labels in advance). It's also missing a way to select the containers to live-update. It maps roughly to the KubernetesDiscovery object (https://api.tilt.dev/kubernetes/kubernetes-discovery-v1alpha1.html). 

I think, in the short term, we should add an API spec and function for live_update objects that "connects" to the underlying pod discovery.

```
live_update(
  name='app-live-update',
  kubernetes_discovery_name='app',
  steps=[...],
  containers=[...])
```

where the `containers=` argument is a system for selecting the containers to live-update, by name or by image.

As an aside - in the long-term, I'd love to have a tiltfile API that more closely matches the REST API, so you could imagine something like:

```
kubernetes_apply(name=..., yaml=..., ,kubernetes_discovery_template=...)
live_update(...)
```

but i think we can punt on that for now.

nicks:
Related issue: https://github.com/tilt-dev/tilt/issues/3282 (though I think the overlap between this one and that one is very implementation-dependent and not obvious)

milas:
With #5248 it's possible to do this for the `k8s_custom_deploy` case by using the `container_selector` argument:  https://docs.tilt.dev/api.html#api.k8s_custom_deploy

The more general use case of being able to set up a Live Update outside of an image build / custom deploy is still valid, however.

marcstreeter:
@milas the more general case you're referring to would apply to scenarios like the following is my hope:

```
[...]

Issue #3961, opened 2020-12-03
reconnect to existing services


nicks:
When people are using a remote cluster, a pattern we've seen is that they don't WANT to restart all the services from scratch. They want Tilt to check if the server is already there, and if possible, try to reconcile with what they had locally.

In other cases, they'd be fine with Tilt reconnecting to a stale version of a service, and updating that service to the fresh version in the background (rather than blocking other builds on it). 

We've had trouble making progress on it because the product experience is deeply tied to the technical implementation, and there's a couple different ways you can solve it, between:

- More robust caching of image builds - https://github.com/tilt-dev/tilt/issues/3690
- A "live connect" feature that checks if the binary can be made consistent with the local workspace with live-update sync alone 
- A way to deliberately use a pre-built image when available
- A way to share images across teams (https://github.com/tilt-dev/tilt/issues/3629)
- Changing the ergonomics about updating on the first tilt up, so that Tilt will only do the initial build if you confirm that that you want to replace the existing one.

I think the sheer number of options here speaks to how painful it is for app devs to start Tilt once you have more than a handful of services, and different ways that people are willing to sacrifice correctness to speed it up.

Filing this issue just to collect thoughts and feedback on this and what the right trade-offs are here between possible solutions

Issue #2582, opened 2019-11-19
enable syncing to a prebuilt image with no docker push


landism:
sample use case: I have a docker image that's built by CI, and is close enough to the code I want to run. I want to `tilt up`, have that image deployed, and then be able to live_update to it.

This is soooort of achievable at the moment by specifying a `docker_build` whose Dockerfile is just a one-line "FROM base-image", but that comes with the potentially major caveat that I have to pull that image to my laptop and then push it back up, which might be annoyingly expensive.

digger18:
Another possible hack is to use custom_build and instead docker build use docker tag.

maiamcc:
Note that this isn't quiiite as simple as "for a deployed image that we didn't build, apply existing live_update logic: whenever a `sync`'d file changes on disk, copy it up to the running container and execute any `run` steps."

we'll need a way of syncing up current code state if e.g. the pod goes down and we lose in-memory changes, which will require us to be able to do a one-off run of ALL live_update steps, i.e. sync ALL sync'd directories, run ALL run steps.

I think this issues dovetails with a feature requested by @drubin, right? (Or is there another ticket for that somewhere?)

drubin:
Yes it is something I am interested in. Been following this waiting :)

arrrnas:
We have the same use-case as OP. I've implemented this as a `live_update` step called `initial_sync`:
```python
  live_update=[
      initial_sync(ignore=['node_modules', '.git', '**/test']),
      sync('./src', '/app/src'),
      run('npm install', trigger=['package.json']),
  ]
```

Our setup: We use `k8s_custom_deploy` to deploy pods using pre-built images (for performance). Images are built from commits on the main branch, so they differ from local files unless you're on the exact same commit with no changes.           

How it works:                                                                                                                                                                                                                                    
  - On container start or restart, all files from sync paths are collected and transferred to the container
  - run steps execute based on whether their trigger files are in the synced set — steps without triggers always run, steps with triggers that don't match are skipped (to avoid failures from missing files)
  - ignore accepts dockerignore-style glob patterns `(**/test, node_modules, *.log)`
  - An optional dockerignore parameter points to a directory with a .dockerignore to reuse
                              
Performance: For larger projects (~20k files, ~90MB) sync takes under 20s to a remote EKS cluster. I initially implemented gzip compression but dropped it — the CPU overhead of compressing many small files was greater than the network transfer savings.

[...]

PR #6737, opened 2026-04-04
Add support for using pre-built images

arrrnas:
Adds `initial_sync()` as a new `live_update` step that syncs all local files to the container on start or restart. This enables live_update with pre-built images that weren't built locally.

```python
live_update=[
    initial_sync(ignore=['node_modules', '.git', '**/test']),
    sync('./src', '/app/src'),
    run('npm install', trigger=['package.json']),
]
```

Addresses #2582. Partially addresses #3961 and #4776. Improves on `file_sync_only` extension (tilt-dev/tilt-extensions#77).

## What it does

- **`initial_sync()`** — new step, must be first in the list. On container start/restart, walks all `sync` paths and transfers files to the container
- **`ignore`** — accepts dockerignore-style glob patterns to exclude files
- **`dockerignore`** — optional path to a directory containing a `.dockerignore` to reuse its patterns
- **`run` steps** respect their triggers during initial sync — steps with unmatched triggers are skipped, steps without triggers always run

## What it doesn't do

- No compression — tested with gzip but CPU overhead exceeded network savings for typical workloads (~20k files, ~90MB in under 20s to remote EKS)
- No changes to existing live_update behavior — `initial_sync` is opt-in


arrrnas:
here's an example of using a generic `nginx:alpine` image with a static website that get's synced when pod starts:
```python
➜  tilt-initial-sync-example cat Tiltfile 
# Example: using initial_sync to serve local files with a stock nginx image.
# No Dockerfile needed — just sync local files into the running container.

allow_k8s_contexts(k8s_context())

k8s_custom_deploy(
    'nginx',
    apply_cmd='kubectl apply -f k8s.yaml -o yaml',
    delete_cmd='kubectl delete -f k8s.yaml --ignore-not-found',
    deps=['site'],
    container_selector='nginx',
    live_update=[
        initial_sync(),
        sync('site', '/usr/share/nginx/html'),
    ],
)

k8s_resource('nginx', port_forwards='8080:80')
```
```yaml
➜  tilt-initial-sync-example cat k8s.yaml 
apiVersion: apps/v1
kind: Deployment
metadata:
  name: nginx
spec:
  replicas: 1
  selector:
    matchLabels:
      app: nginx
  template:
    metadata:
      labels:
        app: nginx
    spec:
      containers:
      - name: nginx
        image: nginx:alpine
        ports:
        - containerPort: 80
```
```html
➜  tilt-initial-sync-example cat site/index.html 
<!DOCTYPE html>
<html>
<head><title>Hello from Tilt</title></head>
<body>
  <h1>Hello from Tilt initial_sync!</h1>
  <p>Edit this file and watch it update in the container.</p>
</body>
</html>
```

nicks:
high-level comments:

overall i like the approach!

i don't understand what problem you're trying to solve with the ignore arguments to initial_sync. in all the cases i can think of, this will put users in weird and inconsistent states? for example, `initial_sync(ignore=['node_modules']), sync('.', '/app')` will put you in states where directories get half-synced if they're modified post-deploy.

it looks like the tests are broken?

is that why the pr is still in draft? or are there other things you want to iterate on before review?

arrrnas:
hey @nicks and thanks for looking into this, apologies for the broken state of the PR, didn't have enough time to update it yet...

regarding the ignore filter - our use case might be a bit contrived but the reason to ignore `node_modules`: the pre-built image the we use already has the correct `node_modules` folder baked in (at least with the dependencies that existed at that commit), so there's 2 scenarios: 
* developer legitimately has changes to the packages and needs them installed
  * `live_update` trigger handles running `npm -i` (or equivalent command) and `node_modules` inside the pod is up-to-date
* developer has changes in `node_modules` due to working on the project locally/natively
  * we don't want to sync those just because there's a huge number of files to sync and our main bottle neck is number of files

this was the main use case for us with the ignore filter, but think there might be more - mono-repo where you might want just a subset of the repo to get synced when using a generic image (like in the nginx example).

There's one main optimization that's missing from this, reconciler.go changes can be optimized I think, to simplify collecting all of the files that need to be synced, otherwise the structure of the PR shouldn't change much.

arrrnas:
so the main change to figuring out which files to sync:
before:
  1. collectAllSyncedFiles — walk all sync dirs, collect every file path into a []string
  2. FilesToPathMappings — loop through each file × each sync rule to compute container paths
  3. MissingLocalPaths — os.Stat() every file again to check if it still exists
  4. TarArchiveForPaths(individualFileMappings, nil) — build tar from individual file PathMappings (which internally walks/stats each file again)

after:
  1. SyncsToPathMappings — build directory-level mappings directly from sync spec (no per-file work)
  2. TarArchiveForPaths(directoryMappings, ignoreFilter) — single directory walk builds the tar with filtering
  
benchmarked this to save ~14% time and ~37% memory at 14k files

arrrnas:
I think I'm done making changes, sorry for the messy PR :{

nicks:
i thought about it a while. i don't think i'm comfortable accepting the change with the ignore_paths/dockerignore arguments to initial_sync.

i understand the problem you're trying to solve. But the feature only works if you hold it in a very specific way. And will lead to weird/broken behavior otherwise.

I'd be comfortable accepting the initial_sync without the ignore bits. I'm also open to discussing other solutions to this problem.

arrrnas:
hey @nicks fair enough, thinking about this, the explicit `ignore` might not even be needed for us, if I recall correctly tilt already takes into account `.dockerignore` and doesn't watch paths that are ignored by it, this makes sense since the files won't end up in the image if built locally, so no need to sync if they change, in our case the images are pre-built, but from the same dockerfile that could be used locally, so `.dockerignore` works the same and we still get the same result of `node_modules` on host machine being ignored and pods rebuilding their own `node_modules` when triggered by changes to `packages.json`, etc.. 

I'll update the PR to omit the ignore param.

arrrnas:
looking at the code my claim
> if I recall correctly tilt already takes into account .dockerignore and doesn't watch paths that are ignored by it,

is not correct, that applies to the normal `live_update` file syncing flow but doesn't apply to `initial_sync`, this leads to a situation where `initial_sync` could be able to copy files to a pod that will not get live_updated/synced later due to being ignored because of `.dockerignore`, which sounds pretty bad..

My proposal would be: `initial_sync()` takes no arguments, but respects `.dockerignore` the same way `sync()` does so that a file that will not get synced by `sync()` would also not get copied by `initial_sync()`, does that make sense?

arrrnas:
I've modified the implementation to leverage the ignore paths that `live_update` already has in order to make `initial_sync` and `sync` behave the same way - files ignored by `sync()` will also be ignored by `initial_sync()`

nicks (approved):
lgtm!

PR #6738, opened 2026-04-04
Consider cronjob resources to be immediately ready

### LINKED ISSUE #2136: Long running cron jobs show pending before their first run
From Doug Sparling in #tilt:

> are CronJob resources deployed through Tilt intended to show as yellow/pending before their first run? My job is hourly so typically when I tilt up it's gonna be awhile until the first job runs. Not a big deal, just a bit annoying since I want everything to come up green on a clean run :slightly_smiling_face:

I think that for cron jobs it should be green if the cron job has been created successfully and should only show pending while the cron job is running.

fabiocarneiro:
Fixes #2136                                                                                                                                                                                                                                                                                                                                                                                                                                                                            
   
CronJobs were stuck in a pending state indefinitely after deployment because Tilt waited for pods to appear before marking them healthy. Since cronJobs run on a schedule, it might take hours, days, etc. for pods to be created. 

It is not enough to set `pod_readiness='ignore'` since it's not about pods being ready, but about them existing in the first place.                                                                                                                                                                                                                                                                                                                                                                                                                    
                                                                                                                                                                                                                                                                                                                                                                                                                                                                                       
CronJobs are now marked as healthy (`RuntimeStatusOK`) as soon as the manifest is successfully applied to the cluster, consistent with the expectation that Tilt's responsibility ends at deployment and not at execution time.                                                                                                                                                                                                                                                                                                                                                                                                                      
                                                                                                                                                                                                                                                                                                                                                                                                                                                                                       
### Changes                                                                                                                                                                                                                                                                                                                                                                                                                                                                            
                                                                                                                                                                                                                                                                                                                                                                                                                                                                                       
- Added `CronJob` to `InitialKinds()` with `PodReadinessIgnore` mode, so Tilt considers a CronJob ready once the resource is created, without waiting for any scheduled pod to run.                                                                                                                                                                                                                                                                                                                                                                                                                                                        
- Added `TestCronJobReadiness` to cover this behavior

### DIFF
--- a/internal/tiltfile/k8s/resource.go
+++ b/internal/tiltfile/k8s/resource.go
@@ -11,11 +11,16 @@ type KindInfo struct {
 }
 
 func InitialKinds() map[k8s.ObjectSelector]*KindInfo {
-	sel, err := k8s.NewPartialMatchObjectSelector("batch/v1", "Job", "", "")
+	jobSel, err := k8s.NewPartialMatchObjectSelector("batch/v1", "Job", "", "")
+	if err != nil {
+		panic(err)
+	}
+	cronJobSel, err := k8s.NewPartialMatchObjectSelector("batch/v1", "CronJob", "", "")
 	if err != nil {
 		panic(err)
 	}
 	return map[k8s.ObjectSelector]*KindInfo{
-		sel: {PodReadinessMode: model.PodReadinessSucceeded},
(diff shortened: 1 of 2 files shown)


nicks (approved):
lgtm!

haggrip:
Just came across this issue, and this PR fixed it when I upgraded Tilt. TYSM!

PR #6743, opened 2026-04-12
Fix: Copy only visible filtered logs from the UI

alonfaraj:
**Problem**

The `Copy` button in the log UI was copying the entire underlying log buffer for the current scope, even when the user had narrowed the view with search or other log filters. That meant the pasted result did not match what was actually visible on screen.

**Solution**

Updated the copy flow to use the active `filterSet` and the same display-filtering rules as the log pane, so `Copy` now includes only the currently shown log lines. This covers search term filtering, source filtering, level filtering, etc. 
Added regression tests for filtered copying, while preserving existing behavior for unfiltered copy cases.

### DIFF
--- a/web/src/logs.ts
+++ b/web/src/logs.ts
@@ -1,6 +1,14 @@
 // Helper functions for dealing with logs
 
+import {
+  FilterLevel,
+  FilterSet,
+  FilterSource,
+  TermState,
+} from "./logfilters"
 import { LogLine, ResourceName } from "./types"
+
+export const DISPLAY_LOG_PROLOGUE_LENGTH = 5
 
 export function logLinesFromString(
   log: string,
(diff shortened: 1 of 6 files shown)


web/src/logs.ts
```diff
@@ -58,3 +61,92 @@ export function sourcePrefix(n: string) {
...
+    return false
+  }
+
+  return (
+    logLineMatchesLevelFilter(line, filterSet) &&
+    logLineMatchesTermFilter(line, filterSet)
+  )
+}
+
+export function filterLogLinesForDisplay(
```
nicks: it looks like all this code is copy-and-pasted from OverviewLogPane. can you factor this out?

possible implementation: create a LogDisplay class that tracks the prologues and filterSet as fields.

nicks (changes requested):
thanks for taking this on!

alonfaraj:
@nicks LMKWYT :slightly_smiling_face:

nicks (approved):
lgtm!

PR #6745, opened 2026-04-16
feat: enable image discovery for image volumes

Jaakkonen:
Enables discovery and replacement of image volumes which have been stabilized in Kubernetes v1.33.

In revisions prior to this the behavior can be had with
```
k8s_image_json_path('{.spec.volumes[*].image.reference}', kind='Pod')
k8s_image_json_path('{.spec.template.spec.volumes[*].image.reference}', kind='Deployment')
k8s_image_json_path('{.spec.template.spec.volumes[*].image.reference}', kind='StatefulSet')
```


nicks (approved):
lgtm!

PR #6751, opened 2026-04-24
adding jsonpath dict filtering

benjamin-wright:
## What?

Update the JsonPath implementation to support filtering in dicts as well as lists

## Why?

Suppose we have a CRD with this kind of structure:

```yaml
spec:
  sql:
    migrations: sql-hello-migrations
    users:
      - name: writer
        permissions:
          - tables: [greetings]
            grant: [ALL]
      - name: reader
        permissions:
          - tables: [greetings]
            grant: [SELECT]
  functions:
    - name: setup
      module: sql-hello-setup
      sqlUser: writer
      trigger:
        http:
          path: /sql-hello/setup
          methods:
            - POST
    - name: query
      module: sql-hello-query
      sqlUser: reader
      trigger:
        http:
          path: /sql-hello
          methods:
            - GET
    - name: insert-test
      module: sql-hello-insert-test
      sqlUser: reader
      trigger:
        http:
          path: /sql-hello/insert
          methods:
            - POST
```

The `functions` section is required, and has to include at least one executable `module` that tilt needs to track. The `sql` section is optional, there's no database if this section isn't provided. If it is provided, you can also include a `migrations` property, which the operator will interpret as an image containing migrations data.

We could construct a k8s_kind to represent this CRD like this:

```starlark
    k8s_kind('Application',
             image_json_path = ['{.spec.functions[*].module}', '{.spec[?(@.migrations)].migrations}'],
             pod_readiness = 'ignore')
```

The `.spec.functions[*].module` jsonpath query covers the easy case, but  `.spec[?(@.migrations)].migrations` uses a filter to optionally select the migrations image if it exists. JsonPath supports this, but tilt currently doesn't.

### DIFF
--- a/internal/k8s/jsonpath/jsonpath_test.go
+++ b/internal/k8s/jsonpath/jsonpath_test.go
@@ -171,6 +171,7 @@ func TestStructInput(t *testing.T) {
 		{"dict/", "{$.Employees.jason}", storeData, "manager", false},
 		{"dict/", "{$.Employees.dan}", storeData, "clerk", false},
 		{"dict-", "{.Labels.k8s-app}", storeData, "20", false},
+		{"dictfilter", "{.Book[?(@.Category == 'reference')].Title}", storeData, "Sayings of the Centurey", false},
 		{"nest", "{.Bicycle[*].Color}", storeData, "red green", false},
 		{"allarray", "{.Book[*].Author}", storeData, "Nigel Rees Evelyn Waugh Herman Melville", false},
 		{"allfileds", "{.Bicycle.*}", storeData, "{red 19.95 true} {green 20.01 false}", false},
--- a/internal/k8s/jsonpath/jsonpath_test.go
+++ b/internal/k8s/jsonpath/jsonpath_test.go
@@ -235,8 +235,7 @@ func TestKubernetes(t *testing.T) {
 		  "status":{
 			"capacity":{"cpu":"4"},
 			"ready": true,
-			"addresses":[{"type": "LegacyHostIP", "address":"127.0.0.1"}],
-			"unique": true
+			"addresses":[{"type": "LegacyHostIP", "address":"127.0.0.1"}]
 		  }
 		},
 		{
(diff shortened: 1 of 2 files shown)


internal/k8s/jsonpath/jsonpath_test.go
```diff
@@ -171,6 +171,7 @@ func TestStructInput(t *testing.T) {
 		{"dict/", "{$.Employees.jason}", storeData, "manager", false},
 		{"dict/", "{$.Employees.dan}", storeData, "clerk", false},
 		{"dict-", "{.Labels.k8s-app}", storeData, "20", false},
+		{"dictfilter", "{.Book[?(@.Category == 'reference')].Title}", storeData, "Sayings of the Centurey", false},
```
nicks: as far as i can tell, this test succeeds today, without your changes to jsonpath.go...it filters on the book array

nicks (commented):
hmmm...i'm a little bit confused by this change? i don't think the jsonpath spec lets you filter by map keys. can you give an example that fails today in the tests?

https://jsonpath.com/ might also help (i personally find jsonpath a bit obtuse and frequently forget bits, so maybe i'm mistaken)

nicks:
Ah no I think you're right, jsonpath does let you filter on keys ... Though I think my comment about adding a proper test is still right

benjamin-wright:
Oh, my bad, yeah, that was still a list filter! I've added an entirely separate little test function as I couldn't find a good place to put a variable object structure without messing with an existing test, does that look okay? Replicating e.g. [this](https://jsonpath.com/#eJx9jLEKwjAURX-lXBwUgijSwSz6Cd19DrV9bQM1qckLoiX_LimIm-s5954ZgUWM7aFnsO2NZWj4rjmWhxIKLsoUpaplCNBdPQZOCo_I_gWN1eW0Pm-DefPmum3c6DwUWtfEO1uBxky2KAhPI0Pl3UTQxYIyXOaZEG5jZIL6mtzLYr_LJKlfxEX51_HcEpYP2YT0AXTuRXQ=)

nicks (approved):
lgtm! thanks

note that this code is forked from https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/client-go/util/jsonpath/jsonpath.go, so if you want this to work in kubectl etc, you could also try to upstream it? but i think this is independently useful in tilt's fork

PR #6754, opened 2026-04-26
feat: limit resource deletion on trigger

Jaakkonen:
Currently if `tilt trigger` is run on a resource with RBACs, admission webhooks or ServiceAccounts those are removed causing cascading failures on other resources with `resource_deps` to them.

This commit changes the `tilt trigger` logic to not call ForceDelete on this kind of resources while keeping for example `Job`s and `Pod`s in the scope of ForceDelete for faster development loops.

### DIFF
--- a/internal/controllers/core/kubernetesapply/reconciler.go
+++ b/internal/controllers/core/kubernetesapply/reconciler.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"fmt"
 	"sync"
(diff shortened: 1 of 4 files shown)


nicks:
i would consider a change that, when running a trigger, skipped the delete step on serviceaccounts and namespaces.

this change is WAY more invasive that that though, and changes the reload semantics entirely.

PR #6755, opened 2026-04-27
handle duplicate objects in parallel upsert batching

### LINKED ISSUE #6756: Parallel Upsert doesn't handle duplicates
### Expected Behavior

Tilt should handle duplicate Kubernetes YAML objects the same way it did before Kubernetes upserts were parallelized.

If a Tiltfile intentionally uses `k8s_yaml(..., allow_duplicates=True)`, and the same Kubernetes object appears more than once in a resource’s YAML, a reload/full update should not fail because duplicate copies of that object race each other during apply.

### Current Behavior

After the parallel Kubernetes upsert changes, Tilt can apply duplicate YAML objects concurrently.

On a reload that performs a force update, Tilt deletes the existing objects and then reapplies the resource YAML. If the same Kubernetes object appears multiple times, two concurrent apply operations can both observe that the object is missing and attempt to create it. One succeeds, and the other fails with an error like:

```
podmonitors.monitoring.coreos.com "reddit-service-comments-go-podmonitor" already exists
```

The failure is intermittent. Reapplying the failed resource usually succeeds, because the object has already been recreated by the first apply attempt, so later applies become updates.

### Steps to Reproduce

1. Create a Tiltfile that applies duplicate YAML for the same Kubernetes object with `allow_duplicates=True`.
2. Start Tilt.
3. Trigger a reload/full update that deletes and reapplies that resource.
4. Observe an intermittent Kubernetes `already exists` error from one of the duplicate objects.

Minimal shape of the repro:

```
yaml = '''
apiVersion: monitoring.coreos.com/v1
kind: PodMonitor
metadata:
  name: duplicate-podmonitor
  namespace: default
spec:
  selector:
    matchLabels:
      app: example
  podMetricsEndpoints:
  - port: metrics
---
apiVersion: monitoring.coreos.com/v1
kind: PodMonitor
metadata:
  name: duplicate-podmonitor
  namespace: default
spec:
  selector:
    matchLabels:
      app: example
  podMetricsEndpoints:
  - port: metrics
'''

k8s_yaml(yaml, allow_duplicates=True)
```

#### About Your Use Case

We have Tiltfiles that intentionally use `allow_duplicates=True` because some generated YAML can contain duplicate objects. Before parallel upserts, this was effectively serialized and did not race during apply.

After the parallel upsert changes, duplicate objects in the same resource can be applied concurrently, producing intermittent `already exists` errors during local development reloads. This blocks the reload until the resource is manually reapplied.

glightfoot:
## Summary

Fix a race in parallel Kubernetes upsert when the input YAML contains duplicate references to the same Kubernetes object.

## Problem

After Kubernetes upserts were parallelized, Tilt could apply duplicate YAML objects concurrently. This shows up with Tiltfiles that use `k8s_yaml(..., allow_duplicates=True)`.

On a reload that performs a force update, Tilt deletes the old objects and then reapplies the target YAML. If the same object appears multiple times in the YAML, two concurrent applies can both observe that the object is missing and attempt to create it. One succeeds, and the other fails with an error like:

```text
podmonitors.monitoring.coreos.com "reddit-service-comments-go-podmonitor" already exists
```

Reapplying the failed resource works because the first attempt has already recreated the object, so later applies become updates instead of competing creates.

## Fix
Keep the parallel upsert behavior for distinct objects, but prevent duplicate Kubernetes object references from being placed in the same parallel batch.

Object identity is based on Kubernetes create/update identity: apiVersion/kind/name/namespace, ignoring server-assigned bookkeeping like UID and resourceVersion.

Fixes https://github.com/tilt-dev/tilt/issues/6756

### DIFF
--- a/internal/k8s/client_test.go
+++ b/internal/k8s/client_test.go
@@ -214,6 +214,58 @@ func TestUpsertIsParallelized(t *testing.T) {
 	})
 }
 
+func TestUpsertSerializesDuplicateObjectRefs(t *testing.T) {
+	synctest.Test(t, func(t *testing.T) {
+		f := newClientTestFixture(t)
+		entities := mustParseYAML(t, `
+apiVersion: monitoring.coreos.com/v1
+kind: PodMonitor
+metadata:
+  name: same
+  namespace: default
+---
+apiVersion: monitoring.coreos.com/v1
+kind: PodMonitor
+metadata:
+  name: other
+  namespace: default
+---
+apiVersion: monitoring.coreos.com/v1
+kind: PodMonitor
+metadata:
+  name: same
+  namespace: default
+`)
+
+		var mu sync.Mutex
+		inFlight := map[v1.ObjectReference]bool{}
+		applyFn := func(target kube.ResourceList, ssa SSAOptions) (*kube.Result, error) {
+			require.Len(t, target, 1)
+			ref := NewK8sEntity(target[0].Object).ToObjectReference()
+
+			mu.Lock()
+			if inFlight[ref] {
+				mu.Unlock()
+				return nil, fmt.Errorf("%s %q already exists", ref.Kind, ref.Name)
+			}
+			inFlight[ref] = true
+			mu.Unlock()
+
+			time.Sleep(50 * time.Millisecond)
+
+			mu.Lock()
+			delete(inFlight, ref)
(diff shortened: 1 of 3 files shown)


nicks (approved):
lgtm! thanks

PR #6762, opened 2026-05-09
fix(k8s): include entity kind/name in kubectl timeout error message

### LINKED ISSUE #3686: kubectl apply timeout error message needs more info
In the k8s tilt channel, Bob asks:

> I've been seeing this fairly often lately. What's it mean and which build failed?
Uncategorized: Build Failed: Killed kubectl. Hit timeout of 10s.

It would be nice of this error message had more info on what, exactly was failing. maybe it should print out the complete yaml that failed to apply

breezy-devs:
Fixes #3686

## Problem

When `kubectl apply` timed out, the error message only reported the duration:

```
Killed kubectl. Hit timeout of 10s.
```

This gave no indication of which resource caused the timeout, making it hard to diagnose in deployments with many resources.

Additionally, the condition that triggered this error path was checking `wgCtx.Err()` for `context.DeadlineExceeded`, but `wgCtx` is an errgroup context that only ever emits `context.Canceled` — never `context.DeadlineExceeded`. This made the timeout error message completely unreachable.

## Fix

1. Fix the condition to check `err` instead of `wgCtx.Err()`, so the timeout error is actually returned when `escalatingUpdate` times out
2. Include the entity's kind and name in the error message

The message now reads:
```
Killed kubectl. Hit timeout of 10s applying Deployment/my-app.
```

## Test

Added `TestTimeoutErrorMessage` to verify the error message includes the timeout duration and the entity kind/name.

### DIFF
--- a/internal/k8s/client.go
+++ b/internal/k8s/client.go
@@ -313,8 +313,8 @@ func ServiceURL(service *v1.Service, ip NodeIP) (*url.URL, error) {
 	return nil, nil
 }
 
-func timeoutError(timeout time.Duration) error {
-	return errors.New(fmt.Sprintf("Killed kubectl. Hit timeout of %v.", timeout))
+func timeoutError(timeout time.Duration, entity K8sEntity) error {
+	return errors.New(fmt.Sprintf("Killed kubectl. Hit timeout of %v applying %s/%s.", timeout, entity.GVK().Kind, entity.Name()))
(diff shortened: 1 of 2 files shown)


breezy-devs:
@nicks would you mind reviewing my PR?

nicks (approved):
lgtm

PR #6763, opened 2026-05-09
feat(web): add per-container log filtering to the log pane

### LINKED ISSUE #3814: Feature Request: Include/Exclude containers from logs
Hi!

Would it be possible to add an option to optionally include/exclude container logs from the UI? Specifically for us, we use Istio and the sidecar proxies can be fairly noisy. It would be nice to filter out logs from the sidecar container. Sort of related to #3744, but slightly different.

Thanks!

breezy-devs:
Closes #3814

## Summary

- Adds a `container` structured field to every pod log line in the streaming controller, enabling frontend filtering without text parsing
- Adds a container filter dropdown to the log pane action bar, shown when a resource has logs from 1+ distinct containers: disabled at exactly 1 (layout stays stable once any container logs arrive), interactive at 2+
- Resources with no container logs (e.g. `local_resource`) never render the dropdown
- Filter selection is URL-persisted via the existing search param pattern and integrates with the existing `FilterSet` pipeline

### Multi-Container

image

### Single Container

image

### No container

image

## Test plan

- [x] Run `go test ./internal/controllers/core/podlogstream/...` to verify the new `TestLogContainerField` test passes
- [x] Run `yarn test` in `web/` to verify all frontend tests pass
- [x] Deploy a multi-container resource (e.g. app + sidecar) and confirm the container dropdown appears and is interactive
- [x] Verify selecting a container filters logs correctly and the selection round-trips through the URL
- [x] Verify a single-container resource shows the dropdown disabled with the container name in the right pill
- [x] Verify a resource with no container logs (e.g. `local_resource`) shows no dropdown

### DIFF
--- a/internal/controllers/core/podlogstream/podlogstreamcontroller.go
+++ b/internal/controllers/core/podlogstream/podlogstreamcontroller.go
@@ -353,6 +353,13 @@ func (c *Controller) consumeLogs(watch *podLogWatch, st store.RStore) {
 
 	ns := watch.namespace
 	startReadTime := watch.startWatchTime
+
+	// Stamp every log line with the container name as a structured field so the
+	// frontend can filter by container without parsing log text.
+	ctx = logger.WithLogger(ctx, logger.Get(ctx).WithFields(logger.Fields{
+		logger.FieldNameContainer: string(watch.cName),
+	}))
+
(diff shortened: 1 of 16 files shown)


breezy-devs:
@nicks would you mind reviewing my PR?

PR #6765, opened 2026-05-13
Light mode

### LINKED ISSUE #1813: Web UI Light Mode
Dark mode is all the rage these days and while I can get behind that at least most of the time, one area where I've been having trouble accepting it is on the web. I have trouble focusing with dark user interfaces that are too saturated on the web for some reason. I usually refer to those color schemes as inky and solarized dark (which I heard the current web UI theme is based on) is one of those "inky" color schemes. If possible, I would love to have something that uses straight up white as the background. :smile:

acls:
Closes #1813

### DIFF
--- a/web/src/AnalyticsNudge.scss
+++ b/web/src/AnalyticsNudge.scss
@@ -11,7 +11,7 @@
   position: fixed;
   z-index: $z-analyticsNudge;
   border: 1px solid $color-gray-50;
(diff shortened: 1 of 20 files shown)


asyncjake:
As a fellow light theme enthusiast, I love this, thank you for the system-preferences check! I hope the maintainers dig this, nice work 💯

PR #6766, opened 2026-05-15
feat(magic): add python's magic variable __name__

yyewolf:
Technically this is already replicable with the following : 

```py
if __file__ = config.main_path:
  main()
```

But since Starlark is really welcoming for Python users, I feel like having `__name__` be either `__main__` or the name of the Tiltfile makes the entry really easy.

(To be honest, I've been using Tilt for quite a bit now, and only discovered `__file__` while trying to add `__name__` 😅)

This PR would allow such patterns : 

```py
def main():
  do_something()

if __name__ = "__main__":
  main()
```

Not sure what the workflow is for the documentation but if this is accepted I'd also need to [tilt.build](https://github.com/tilt-dev/tilt.build) in order to add __name__ workaround like __file__ ?


nicks (approved):
lgtm!

PR #6771, opened 2026-05-29
Update local registry warning documentation link 

jarpoole:
The repository originally pointed to just redirects here anyways so I thought it might worth cleaning up


nicks (approved):
thanks!

PR #6782, opened 2026-06-17
Debounce transient Kubernetes cluster-health failures

mterwill:
[#6742](https://github.com/tilt-dev/tilt/pull/6742) changed `tilt ci` so that any cluster error immediately fails the CI session. The cluster error it relies on, however, is produced by a single health-check poll and cleared by the next successful poll.

Kubernetes evaluates health over a series of probes rather than treating one failed check as terminal. Its probe model uses a [`failureThreshold`](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#configure-probes) and only considers the overall check failed after successive failures. Tilt should apply similar tolerance before turning a point-in-time health observation into an irreversible CI result.

We see this most often on busy clusters with many concurrent CI jobs, where one brief control-plane interruption can cascade into failures across every active Tilt session.

This change requires health checks to fail continuously for one minute—five successive failed polls at the current 15-second interval—before reporting a cluster error. A successful check resets the timer immediately. Sustained failures still terminate CI, but one failed poll no longer amplifies a brief interruption into many permanent job failures.

### DIFF
--- a/internal/controllers/core/cluster/monitor.go
+++ b/internal/controllers/core/cluster/monitor.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"sync"
+	"time"
 
 	"github.com/jonboulle/clockwork"
 	"k8s.io/apimachinery/pkg/types"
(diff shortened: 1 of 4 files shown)


nicks:
this pr is marked as draft -- is this something you're still working on?

mterwill:
Thanks @nicks! The original branch was meant to add observability into a failure we were seeing internally. I’ve investigated further and reworked it around what I believe is the underlying issue.

internal/controllers/core/cluster/monitor.go
```diff
@@ -96,11 +97,19 @@ func (c *clusterHealthMonitor) run(ctx context.Context, clusterNN types.Namespac
...
+	var unhealthySince time.Time
 	for {
 		err := doKubernetesHealthCheck(ctx, conn.k8sClient)
 		if err != nil {
-			c.UpdateStatus(ctx, clusterNN, err.Error())
+			now := c.clock.Now()
+			if unhealthySince.IsZero() {
+				unhealthySince = now
+			}
+			if now.Sub(unhealthySince) >= clientHealthFailureGracePeriod {
```
nicks, line 108: i'm going to change this so that the failure threshold is a count rather than a duration. i think that's more consistent with how failure thresholds are usually configured. doing it by duration may have weird effects if the health check takes a long time.

nicks (approved):
thanks! ya, agreed that this seems like the right fix

Issue #4816, opened 2021-07-29
port forwards disappearing


landism:
A user on slack [reports](https://kubernetes.slack.com/archives/CESBL84MV/p1627489859376200) that their port forwards occasionally stop working.

1. Tilt shows their resources as healthy.
2. `tilt get portforwards` says "No resources found"
3. If they comment out a `port_forward` and uncomment it, then the portforward starts working again.

We do not have an independent repro.

I've asked for the contents of their kubernetesdiscovery object the next time they observe this problem.

landism:
I'm assuming this is distinct from #4801 - in that issue, the portforwards exist but do not work, while in this issue they do not exist.

Issue #4801, opened 2021-07-27
portforward gets wedged if fails too many times


nicks:
We've received reports where if you have an istio sidecar that takes too long to set up networking, Tilt's portforwarding never connects.

I have not spent any time trying to independently reproduce this, or diagnose why it might be happening.

Possibilities include:
- the portforward retry logic gets wedged
- the backoff limit is too high, and simply gets bumped up to an unreasonable time

The best workaround right now is to run

```
tilt delete portforwards --all
```

That will delete all the active portforwards, so that Tilt re-creates them from the KubernetesDiscovery object

nicks:
additional info:

> just sending curls in a tight loop to the port seems to be enough to trigger this

> when i swap in gunicorn over a more direct flask runserver I see pretty consistently better behavior

another hypothesis - maybe there's some more complicated interaction between the port-forwarding code and the container network listener? like there's an open connection but it gets wedged, so we need to be restarting the connection

nicks:
i think i have a repro case for this

PR #6790, opened 2026-07-02
k8s: fix port-forward wedges when the pod connection or API server hangs

yarinzirlin:
## Problem

Two failure modes leave a PortForward permanently broken while its resource stays green in the Tilt UI, and both survive the reconciler's infinite retry loop (added for #4801):

**1. A hung `streamConn.Close()` leaks the local port forever.**

`ForwardPorts` defers `pf.streamConn.Close()` *after* `pf.Close()` (LIFO: the stream conn closes first, listeners last). spdystream's `Connection.Close()` synchronously writes a GOAWAY frame with no write deadline ([connection.go#L785](https://github.com/moby/spdystream/blob/v0.5.0/connection.go#L785)). On a half-open connection whose kernel send buffer is already saturated — the peer died mid-traffic — that write never returns, so the listeners are never closed:

- the local port stays bound with nobody accepting (connects hang on SYN until timeout), and
- when the pod is recreated, the successor PortForward fails every retry with `bind: address already in use` — we observed one stuck like this for **5 hours**, while `tilt get uiresources` showed the resource healthy. Only restarting Tilt cleared it.

**2. The dial has no deadline.**

`pf.dialer.Dial(...)` performs the SPDY/websocket upgrade with no timeout anywhere in the chain (the `http.Client` built in `ProvidePortForwardClient` has no `Timeout`). Against an API server that accepts TCP but never answers — overloaded, or its VM is unhealthy — `Dial` blocks forever. The reconciler's retry loop is wedged *inside* `onePortForward`, so the forward never recovers **even after the API server does**. Observed as a forward stuck on `lost connection to pod` (status never updated again) after an API-server brownout.

Both were hit in the wild against a kind cluster whose API server degraded under host disk pressure; the diagnosis above is from reading the wedged state live (`tilt get portforwards` error statuses, SYN-timeout behavior on the leaked port, nothing else bound to the port).

## Fix

- `ForwardPorts` now releases the local listeners **before** touching the stream connection, and bounds the stream-conn close with a 5s timeout (`closeStreamConn`). A wedged close abandons the already-dead connection instead of holding the port and the retry loop hostage.
- `Close()` takes a mutex and is idempotent; listener registration locks the same mutex.
- Dialing moved to `dial()`: runs the dialer in a goroutine, gives up after 30s or when `stopChan` closes, and reaps a late-arriving connection.

## Tests

- `TestForwardPortsReleasesListenersWhenStreamConnCloseHangs` — stream conn whose `Close` blocks forever; asserts the local port becomes bindable after stop anyway (fails on master).
- `TestForwardPortsTimesOutWhenDialHangs` — dialer that never returns; `ForwardPorts` errors with a timeout (hangs forever on master).
- `TestForwardPortsAbortsDialWhenStopChanIsClosed` — stop mid-dial returns promptly (hangs forever on master).

`go test ./internal/k8s/... ./internal/controllers/core/portforward/ -race` passes.

Related: #4801 (same symptom family; its `tilt delete portforwards` workaround is what we'd been running as an external watchdog), #4816.

### DIFF
--- a/internal/k8s/portforward/portforward.go
+++ b/internal/k8s/portforward/portforward.go
@@ -26,6 +26,7 @@ import (
 	"strconv"
 	"strings"
 	"sync"
+	"time"
 
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/util/httpstream"
@@ -44,6 +45,11 @@ var (
 	networkClosedError = "use of closed network connection"
 )
 
+const (
+	defaultDialTimeout     = 30 * time.Second
+	streamConnCloseTimeout = 5 * time.Second
+)
+
 // PortForwarder knows how to listen for local connections and forward them to
 // a remote pod via an upgraded HTTP request.
 type PortForwarder struct {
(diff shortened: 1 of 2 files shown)


nicks:
this pr is marked draft -- is this something you're still working on?

PR #6793, opened 2026-07-09
feat(UI): focus resource filter keyboard shortcut

corbanpro:
Adds a keyboard shortcut `/` that focuses the resource filter input


nicks (approved):
lgtm!

PR #6794, opened 2026-07-09
fix(UI): search bar target url

corbanpro:
When using the "Search Tilt docs..." search bar in the Tilt UI, the target URL was incorrect. It navigates users to <https://docs.tilt.dev/search?q=search&utm_source=tiltui> instead of <https://docs.tilt.dev/search.html?q=search&utm_source=tiltui>. Notice that ".html" is missing on the first link.


nicks (approved):
lgtm, thanks!

PR #6799, opened 2026-07-16
Add disable-port-forwards flag

glightfoot:
We have started running multiple `tilt up` instances per host. While this works well, we run into lots of port-forward conflict warnings that spam the logs and result in an uncertain set of port-forwards.

This adds a flag to `tilt up --disable-port-forwards` that prevents these port forwards from being created, allowing us to control which instance can use port-forwards.

I have not used wire before, so I wasn't sure if I should just use a bare bool, or the typed one I added here.

### DIFF
--- a/internal/cli/up.go
+++ b/internal/cli/up.go
@@ -50,6 +50,7 @@ var starterTiltfile []byte
 type upCmd struct {
 	fileName             string
 	outputSnapshotOnExit string
+	disablePortForwards  bool
 
 	legacy bool
 	stream bool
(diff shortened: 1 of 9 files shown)


nicks (approved):
lgtm

seems reasonable to me. i'm probably going to change the interface so it's a --port-forwards flag that defaults to true, rather than a --disable-port-forwards flag.

i wish unix networking had evolved so that applications could create arbitrary loopback addresses, like you can for ports. maybe someday when ipv4 has been retired... 👴

PR #6802, opened 2026-07-21
engine: skip exact log-only store notifications in PodMonitor

alleneubank:
## Summary

Under a sustained Kubernetes log storm, `PodMonitor` was spending a large
fraction of engine CPU and allocations on work that cannot affect rollout
status: every store notification (including exact log-only updates) walked
every manifest target and ran reflection-based pod status comparison.

This PR:

- Returns early from `PodMonitor.OnChange` for **exactly** log-only
  `ChangeSummary`s, matching the pattern already used by other store
  subscribers.
- Replaces `cmp.Equal` in `podStatusesEqual` with explicit field comparisons
  (same semantics, no reflection allocations).
- Adds `ChangeSummary.IsLogOnly` and fixes `ChangeSummary.Add` so `UIButtons`
  is merged like every other `ChangeSet` (a coalesced button change was
  previously droppable when mixed with a pending log-only summary).
- Adds field-count drift tests so hand-rolled enumerations fail closed when
  structs gain fields.

## Motivation

Profiled on a noisy log-only workload: `PodMonitor.diff` /
`podStatusesEqual` were ~18% cumulative CPU and ~17% cumulative allocations
in the engine. Runtime logs cannot change rollout state; re-diffing pods on
every log line is pure waste.

## Evidence

- Package tests for log-only skip, mixed summaries, field merge, and drift
  guards.
- Microbench: log-only `OnChange` ≈ 1.15ms / 6210 allocs → ≈ 5ns / 0 allocs.
- Hermetic log-only soak (k3d, clean tree including this series through
  `4c1d05bf6`): pass — hot path absent from steady-state profiles, memory
  under 2 GiB floor, rollout still observed, log rate healthy.

## Test plan

- [x] `go test ./internal/engine/k8srollout/ ./internal/store/`
- [ ] CI on this PR
- [ ] Manual: `tilt up` against a chatty workload still shows real pod
      restarts in resource logs; pure log spam does not re-churn rollout
      monitoring

### DIFF
--- a/internal/engine/k8srollout/podmonitor.go
+++ b/internal/engine/k8srollout/podmonitor.go
@@ -6,7 +6,6 @@ import (
 	"strings"
 	"time"
 
-	"github.com/google/go-cmp/cmp"
 	"github.com/jonboulle/clockwork"
 	v1 "k8s.io/api/core/v1"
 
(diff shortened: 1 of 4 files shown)


nicks:
i'm fine with the IsLogOnly change. we probably wouldn't accept a change to have hand-written cmp functions, sorry

nicks (approved):
On thinking on this - a lot of this diffing code is gradually going away in favor of the k8s generated code, so I guess I'm fine with this -- it won't really grow. Lgtm!

PR #6803, opened 2026-07-21
logstore: cut allocation churn on the incremental log hot path

alleneubank:
## Summary

After reducing subscriber work on log-only notifications, logstore was still
a major source of steady-state GC under a multi-span log storm:

- empty checkpoint windows still cloned span maps
- incremental `ToLogList` re-sent every span
- `toLogLines` allocated heavily per rendered line

This PR:

- Returns early from `ContinuingLinesWithOptions` when the checkpoint window
  is empty.
- Clones only spans referenced by the checkpoint window (drops full
  `cloneSpanMap` for that path).
- Scopes incremental `ToLogList` span payloads to segments actually sent
  (safe for `tilt logs` and the web client; covered by
  `TestLogIncrementalSpansScopedToSegments`).
- Preallocates segments from newline counts; reuses a single line builder
  and cuts per-line prefix/builder churn in `toLogLines` / prefix helpers.

## Motivation

On a ~200-span session shape after the PodMonitor skip, span cloning and
line materialization dominated remaining engine alloc profiles during
steady log ingress.

## Evidence

- Unit tests for scoped incremental spans and prefix characterization.
- New benches in `logstore_bench_test.go` (empty window, narrow window,
  full-store render, continuing lines).
- Reported microbench deltas (same shapes as in commit messages), e.g.
  full-store render ≈ 950µs / 16k allocs → ≈ 305µs / 2k allocs.
- Hermetic log-only soak floors green with log rate within the ≥80% baseline
  band while validating the firehose path.

## Test plan

- [x] `go test ./pkg/model/logstore/`
- [ ] CI on this PR
- [ ] Optional: `go test -bench=. ./pkg/model/logstore/` if reviewing numbers

### DIFF
--- a/pkg/model/logstore/logline.go
+++ b/pkg/model/logstore/logline.go
@@ -26,6 +26,9 @@ type LogLine struct {
 	Time time.Time
 }
 
+// Accumulates the segments of one line. A single builder is reused across
+// all lines of a toLogLines call (start/reset recycle the segment buffer),
+// so it must not retain segment state past reset.
 type logLineBuilder struct {
 	span        *Span
 	segments    []LogSegment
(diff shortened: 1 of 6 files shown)


nicks (approved):
lgtm! thanks

PR #6804, opened 2026-07-21
logs: avoid no-op filter copies and clamp Compose log replay to tilt start

alleneubank:
## Summary

Three small, related fixes on the log ingress / presentation path:

1. **Default log filter no longer copies every line.**
   The default `tilt up` terminal filter constrains nothing, but
   `ApplyWithOptions` still allocated a fresh `[]LogLine` on every batch.
   Constraint-free filters now return the input (tail as a re-slice). Selective
   filters keep grow-by-append (presizing regressed sparse filters in benches).
   Contract test ensures `matchesAllLines` cannot false-positive.

2. **Docker Compose log streams clamp `Since` to Tilt start.**
   Warm Compose containers previously replayed the full docker json-file
   backlog from container start into the store action queue. Kubernetes pod
   logs already stream from Tilt start; Compose now uses
   `max(container start − 1s, TiltStartTime)`. Live lines after `tilt up` are
   unchanged.

3. **Websocket reader test fixture is race-safe.**
   Test-only mutex around shared fake-handler state so
   `go test -race ./internal/hud/client/` passes (pre-existing race; unrelated
   to the filter change).

## Motivation

- Filter copy: largest optimizable allocator left in a log-storm profile after
  logstore work (~6% cumulative allocs / tens of MB flat on the no-op path).
- Compose replay: warm-restart against a multi-GB docker log backlog drove RSS
  into multi-GB within seconds and OOM under a memory cap, even though Tilt
  truncates retained spans to ~2MB — the damage is the transient replay.

## Evidence

- Filter unit + contract tests; benches: no-op 2000-line batch ≈ 170µs / 838KB
  → ≈ 3.5ns / 0 allocs.
- New reconciler test fails without the Compose `Since` clamp; fake Docker
  client records the requested window.
- Package tests pass with `-race` for the websocket fixture change.
- Compose warm-restart observation (local): multi-GB RSS / OOM without clamp →
  ~154–306MB over a 10‑minute soak with clamp.

## Test plan

- [x] `go test ./internal/hud/client/ ./internal/controllers/core/dockercomposelogstream/`
- [ ] `go test -race ./internal/hud/client/`
- [ ] CI on this PR
- [ ] Manual (Compose): restart Tilt against long-lived containers; confirm
      backlog is not fully replayed and new lines still stream

## Notes

If preferred, the race-only websocket commit can be split out; it is kept here
because it is the same package as the filter change and unblocks `-race` CI
locally.

### DIFF
--- a/internal/controllers/core/dockercomposelogstream/reconciler.go
+++ b/internal/controllers/core/dockercomposelogstream/reconciler.go
@@ -152,6 +152,16 @@ func (r *Reconciler) manageLogWatch(ctx context.Context, nn types.NamespacedName
 	// the first log timestamps (also reported by Docker), so we pad it by a second to reduce the
 	// number of potentially duplicative logs
 	startWatchTime := containerState.StartedAt.Time.Add(-time.Second)
+	state := r.store.RLockState()
+	tiltStartTime := state.TiltStartTime
+	r.store.RUnlockState()
+
+	// Tilt retains about 2MB per log span, while a warm Compose container can hold GBs of json-file
+	// logs. Requesting that full history would flood the store's unbounded action queue only to discard
+	// it, so mirror the pod log stream invariant and only stream logs since Tilt started.
+	if !tiltStartTime.IsZero() && startWatchTime.Before(tiltStartTime) {
+		startWatchTime = tiltStartTime
+	}
 	if result.watch != nil {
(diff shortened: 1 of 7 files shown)


nicks (approved):
lgtm!

PR #6819, opened 2026-08-06
cli: tweak --disable-port-forwards flag

nicks:
- it's not --port-forwards=false, not --disable-port-forwards
- we implement it in the tiltfile assembler
Signed-off-by: Nick Santos <nicholas.j.santos@gmail.com>


nicks:
oops there's a problem with this...need to update it so the port-forwards don't show up in the ui

PR #6823, opened 2026-08-13
Fix initial_sync being cancelled for syncing fall_back_on files

arrrnas:
Ran into an issue where `initial_sync` would silently fail due to files triggering `fall_back_on` and cancelling the file sync silently. Conceptually this shouldn't happen because `fall_back_on` is intended to fire when the user modifies a file while `initial_sync` is the system itself ensuring file state is the same between local and inside the pod.

context: in #6737 I introduced `initial_sync` which preemptively syncs all the watched files to the pod, the main use case is to ensure consistent file state between slightly outdated pre-built images and host machine file system.


nicks (approved):
lgtm!

PR #6834, opened 2026-09-07
`tilt down` deletes a concurrently-running instance's frozen kubeconfig, breaking its k8s_custom_deploy/helm_resource de

### LINKED ISSUE #6826: `tilt down` deletes a concurrently-running instance's frozen kubeconfig, breaking its k8s_custom_deploy/helm_resource deploys (localhost:8080)
### Expected Behavior
Running `tilt down` for one env should not delete or modify runtime state that a **different, concurrently-running** `tilt up` instance depends on.

The "frozen" kubeconfig that Tilt injects into `k8s_custom_deploy` shell-outs (from #5703 / #5790) should be isolated per instance, and `tilt down` should either be targetable at a specific instance or leave other instances' frozen kubeconfigs untouched.

### Current Behavior
The frozen kubeconfig is written to `$XDG_RUNTIME_DIR/tilt-dev/<identity>/cluster/default.yml`, where `<identity>` is `tilt-default` for the default web port and `tilt-<port>` when `tilt up --port <port>` is used.

`tilt down` has **no `--port` / instance-identity flag** — it always runs under the default `tilt-default` identity. On exit it cleans up `tilt-default/cluster/default.yml`. If a primary `tilt up` on the default port is still running, that primary's `helm_resource` / `k8s_custom_deploy` apply commands have the now-deleted path injected as `KUBECONFIG`. helm/kubectl then build an empty config and fall back to the default host:

```
Error: Kubernetes cluster unreachable: Get "http://localhost:8080/version": dial tcp 127.0.0.1:8080: connect: connection refused
```

So every deploy in the still-running primary instance fails. It does not self-recover: `tilt trigger <resource>` re-runs the same command with the same dead path. Only a full Tiltfile re-eval / restart rewrites the frozen kubeconfig and fixes it. The `localhost:8080` symptom is very misleading — it reads as a kubeconfig/auth problem on the primary, with nothing pointing at the cross-instance file deletion.

### Steps to Reproduce
1. `tilt up` on the default port against a Tiltfile that uses `helm_resource` / `k8s_custom_deploy` (env A). Let it deploy. → writes `$XDG_RUNTIME_DIR/tilt-dev/tilt-default/cluster/default.yml`.
2. In another terminal, start a second instance for a second env: `tilt up --port 10351 -- <env B args>`. → writes `$XDG_RUNTIME_DIR/tilt-dev/tilt-10351/cluster/default.yml`.
3. Tear down env B: `tilt down -- <env B args>` (note `tilt down` takes no `--port`, so it acts as `tilt-default`). This correctly deletes env B's k8s resources, but also removes `tilt-default/cluster/default.yml`.
4. Back in instance A, trigger any `helm_resource` / `k8s_custom_deploy` resource (or wait for a re-apply). It fails with `Kubernetes cluster unreachable: ... http://localhost:8080/version`, and stays failed until instance A's Tiltfile is fully re-evaluated. Note `tilt-10351/cluster/default.yml` (env B's own dir) is left as an orphan — `tilt down` never touched it.

#### `tilt doctor` Output
```
$ tilt doctor
Tilt: v0.37.4, built 2026-06-16
System: linux-amd64
---
Docker
- Server Version: 29.7.2
- API Version: 1.55
- Builder: 2
- Compose Version: v5.4.0
---
Kubernetes
- Env: unknown
- Context: development-aks
- Cluster Name: development-aks
- Namespace: ray-platform-djw2
- Version: v1.32.3
```

Deploys use the `helm_resource` extension (which is built on `k8s_custom_deploy`).

#### About Your Use Case
I run a primary dev environment via `tilt up` and occasionally stand up a second, short-lived environment in parallel via `tilt up --port <port>` (the port flag is required because a second `tilt up` on the same port just errors out and exits). When I'm done with the second env, I want to `tilt down` **only** that env to free cluster resources before auto-cleanup.

Because `tilt down` can't be pointed at the second instance, running it silently broke my primary instance's deploys — every resource failed with `localhost:8080`, `tilt trigger` didn't help, and it took a long investigation to trace it to `tilt down` having deleted the primary's frozen kubeconfig out from under it. Ctrl+C on the second instance isn't an alternative, since that leaves its deployed resources running in the cluster.

Possible fixes: scope the frozen kubeconfig per running instance so `tilt down` can't cross-delete it, and/or give `tilt down` a way to target a specific instance (e.g. `--port`/identity), and/or guard cleanup so an instance never removes a frozen kubeconfig that another live instance is using.

victorwon2001:
## Summary

- Allow `tilt down` to accept the existing `--port` and `--host` instance-selection flags, so the existing down dependency wiring derives its frozen kubeconfig identity from the targeted port.
- Add regression coverage showing that `tilt down --port 10351 -- env-b` uses and removes the `tilt-10351` frozen kubeconfig while leaving `tilt-default` untouched and preserving Tiltfile arguments.

Fixes #6826

## Test plan

All down command tests pass, as does the rest of the `internal/cli` package when two unrelated tree-view ANSI color tests are skipped. The unfiltered package run fails only those two color-output assertions because ANSI escape sequences are absent in this test environment.


nicks:
hmmm...we appreciate the contribution, but this seems like the wrong fix...

this updates tilt down so that it can delete a different running kubeconfig. but it's not clear to me why it should delete the kubeconfig at all....gmas...

nicks:
thanks again for the contribution! i decided to go with a different approach ... it felt a bit too magical to make people remember to pass --port correctly on tilt down.

